PR #1649, opened 2026-08-06
refactor: declare selector resolution policy as data

thymikee:
Part of #1630 — the policy matrix and its wiring. The remaining criterion (mutating find must not chain two resolution engines) is split out as #1654, with the reasoning in that issue.

## What

Five native consumers of "resolve a selector against the screen" each hand-declared their ambiguity contract as inline `requireUnique` / `disambiguateAmbiguous` literals, so the repo's real policy matrix was only discoverable by reading four files and diffing them in your head. `SELECTOR_RESOLUTION_POLICIES` (`packages/selectors/src/internal/resolution-policy.ts`) now declares **one row per caller**, and `selectorResolutionKnobs` turns a row into the engine knobs it stands for.

| row | caller | ambiguity | rect |---|
| `act` | click/press/fill/focus/longPress/drag/scroll | disambiguate | ✓ |
| `actCoveredDiagnosis` | the post-miss "covered?" probe | first-match | ✓ |
| `readText` | `get text` | disambiguate | — |
| `readUnique` | `is` non-exists, `get attrs` | fail-closed | — |
| `readAny` | `exists`, find read actions | first-match | — |
| `wait` | `wait` | first-match | — |
| `findAct` | mutating `find` | reject-candidates | ✓ |

**Zero ambiguity literals remain in `src`.** The matrix declares only the ambiguity contract and the rect requirement — the fields it actually enforces; the structural pipeline stages stay with their callers and are tracked in #1656 (an earlier revision declared them as columns nothing consumed). Semantics are unchanged by construction — each row was read off its call site; the matrix names what was previously implicit.

`reject-candidates` is declaration-only and **rejected by `selectorResolutionKnobs` at the type level**: mutating find enforces #1625's contract through its own narrowing logic, not through engine knobs, and the types now prevent anyone wiring it up as if it were a knob.

## The matrix is gate-tested, not trusted

`resolution-policy-parity.test.ts` follows ADR 0011's declared-plus-gate-tested pattern, because a matrix that merely *claims* things about callers rots silently:

- knobs must match the ambiguity contract each row names;
- every claimed structural column must appear in that caller's source (occlusion/off-screen/promotion/poll each have marker sets);
- the inverse direction too — read and wait pipelines must *genuinely lack* the machinery they disclaim, so the matrix can't under-report;
- no caller may reintroduce an inline `requireUnique`/`disambiguateAmbiguous` literal.

[...]

PR #1744, opened 2026-08-11
refactor: route selector-resolution structural stages into typed policy

### LINKED ISSUE #1656: Route selector-resolution structural stages (occlusion, off-screen, promotion, poll) into typed policy
Split out of #1630 (see PR #1649 review).

## Context

PR #1649 landed `SELECTOR_RESOLUTION_POLICIES` covering the **ambiguity contract** and the **rect requirement** — both consumed by `resolveSelectorChainWithPolicy` and pinned behaviorally.

An earlier revision of that matrix also carried four structural columns — `occlusion`, `offscreenGuard`, `promotion`, `poll` — describing which pipeline stages each caller runs. They were removed before merge because **nothing consumed them**: changing any of them left behavior and the tests green, so they were unverifiable claims that read as truth. (A source-sniffing test "verified" them by grepping caller files for marker strings, which is not behavioral coverage and stayed green when a row was disconnected entirely.) The suite now fails if such a field reappears without enforcement.

## What remains

The stages themselves are still per-caller pipeline code:

| stage | today |
|---|---|
| occlusion | `isSnapshotNodeInteractionBlocked` in `resolution.ts` (3 sites) + `find.ts`; `is`/`get`/`wait` never consult it |
| off-screen guard | `throwIfOffscreenInteractionTarget` in `resolution.ts`, with the iOS rescue probe |
| hittable-ancestor promotion | `resolveActionableTouchResolution` in `core/interaction-targeting.ts` |
| poll budget | `createWaitPolling` in `wait-polling.ts`, wait only |

Routing these into the policy would make the whole pipeline shape declarative rather than half-declared.

## Constraints

- Each stage must be **consumed** by the resolution interface, not merely declared — the point of this issue is to avoid re-introducing unverifiable columns.
- Each must be pinned by a test that fails when its row changes (fixture-tree behavior, not source sniffing).
- Semantics must not change: `is`/`get attrs` still fail closed, `wait` still skips occlusion and off-screen, acting paths keep promotion.
- Maestro stays out (ADR 0015).

## Acceptance criteria

- [ ] Structural stages consumed through the resolution interface, not caller-local branching
- [ ] Every declared field behaviorally pinned; flipping it changes an assertion
- [ ] The "declare only what is enforced" guard extended to the new fields
- [ ] No semantic change to any existing caller

thymikee:
Closes #1656; prework for #1739 (waves 4-5).

## What

#1649 landed the per-caller **ambiguity** matrix and deliberately dropped four structural columns — `occlusion`, `offscreenGuard`, `promotion`, `poll` — because nothing consumed them: flipping any of them left behavior and the suite green, so they read as truth while being free to drift. This adds the missing half **with an owner that executes it**.

`SELECTOR_PIPELINE_POLICIES` (`src/core/selector-pipeline-policy.ts`) gives each caller ONE row: the ambiguity row it resolves under, plus its four structural stages. `src/core/selector-pipeline.ts` is the only code that runs them, and the only code that may touch the matching engine:

| entry | for | runs |
|---|---|---|
| `resolveSelectorPipeline` | single-target rows (reads, `wait`, covered-diagnosis) | candidacy → ambiguity → replay-guard hook → promotion → occlusion → off-screen |
| `listSelectorPipelineMatches` | `reject-candidates` rows (`click`/`fill`, mutating `find`, `find list`) | candidacy → the candidate set **plus the tree the row sees** |
| `runNodePipelineStages` | targets from a non-chain matcher (`@ref`, find's fuzzy locator, a narrowed set) | every node stage the row declares |

The stage functions are private to the owner, so a caller cannot run three stages and forget the fourth, and cannot pair one row's candidate set with another row's ambiguity contract — the row enters once, at the top.

The table lives in `src/core/` rather than `packages/selectors`: R2 forbids `daemon/` importing `commands/`, and the selectors package is deliberately blind to occlusion annotations, backend probes and the wait clock, so a column there would be a claim nothing in that package could enforce.

## Review fix: the rows were bypassable (P1)

The first revision had read/wait routes composing `selectorPipelineCandidates(row, nodes)` with the raw `resolveSelectorChainWithPolicy(..., row.resolution)`. They never entered promotion or off-screen, so flipping a read row's cell changed only the policy unit tests — production `get`/`is`/`wait` were untouched. That is the same unverifiable-column failure one level down. Fixed in three parts:

1. **Owning interface** (above): every route now hands the owner a row and gets every stage, skips included.
2. **A refusal a row declares must exist.** A row with `offscreen: 'refuse'` must supply a refusal shape; flipping an observation row to `refuse` therefore fails on its real route instead of silently observing.
3. **R19 selector-pipeline-ownership** (`scripts/layering/`): the two engine entries moved to their own package subpath, `@agent-device/selectors/engine`, and only the owner may import it.

`find list` was an acknowledged bypass and is closed: it names a `readList` row (`reject-candidates`, no rect requirement — every match is the answer) instead of calling the engine.

### The engine door is a specifier, not a symbol

A name-shaped check cannot see the bypasses that matter: a namespace import, a re-export, and a deferred `import()` never mention the symbol it matches. R19 enforces on the SPECIFIER over the **resolved import graph**, where all of those are the same edge. Each form was planted into a shipped route and rejected by the repo-wide scan:

| planted into `selector-read.ts` | R19 |
|---|---|
| `import * as selectorEngine from '@agent-device/selectors/engine'` | red |
| `await import('@agent-device/selectors/engine')` inside a function | red |
| `export * from '@agent-device/selectors/engine'` in a sibling module | red |

`resolveImportEdges` **drops** an edge whose specifier resolves to nothing, so a specifier-keyed rule goes quiet rather than red if the subpath is ever retired. The gate now reports that as its own failure instead of scanning clean.

### Listing rows declare only what a listing can run

`find <q> list` resolves to a candidate SET: promotion, the off-screen guard and a poll budget have nothing to apply to, because a listing has no single element to retarget, keep on screen, or wait for. `readList` therefore declares two stages, not five (`SelectorListPolicy` = resolution + occlusion), and the narrow shape is load-bearing rather than documentation — `runNodePipelineStages` and `selectorPollBudget` take the full row, so a listing row cannot reach them:

```ts
// @ts-expect-error a listing row declares no promotion or off-screen stage.
const nodeStages: Parameters<typeof runNodePipelineStages>[0] = SELECTOR_PIPELINE_POLICIES.readList;
```

Widening `readList` back to five columns makes those directives unused and fails the typecheck (verified). Its one live cell, `occlusion`, is pinned on the shipped route: flipping it fails `find … list`.

### Rule id: R19, and a gate so the next collision is caught

R17/R18 belong to #1750. R19 verified free against `origin/main` and that PR's diff, then validated by **real merges in both directions** (not a file-level union): `#1750 → #1744` and `#1744 → #1750` both merge cleanly, the uniqueness gate passes in both, and the ids stay distinct (`R17 device-inventory-cutover`, `R18 contracts-implementation-authority`, `R19 selector-pipeline-ownership`).

Nothing caught R17 being taken twice, because two branches claiming one free number do not conflict in git — the files differ. `scripts/layering/rule-ids.ts` closes that: it collects rule-id declarations and fails when one id names two rules. Two notes on its construction:

- It matches **whole string literals**, which is what separates a declaration from prose that also names a rule (`'R17 holds the devices command to…'`). The first version keyed on `rule:` and missed #1750's `const RULE = '…'` shape — it would have been a gate that could not see the collision it exists for.
- `main` already has two collisions (R11 and R13, both named twice), which #1750 is renaming away. Those are allowed **transitionally, and each allowance expires with its collision**: an entry whose collision is absent from the scan fails as a stale allowance, so it has to be deleted in the same change that removes the collision and the list burns down to empty. A list that merely filtered those strings would fail open forever — after the rename, reintroducing either collision would be waved straight through.

  Verified against a scratch tree carrying #1750's rename: leaving the list untouched reports both entries stale, deleting them is clean, and reintroducing `R11 names contracts-implementation-authority and package-boundaries` afterwards is rejected. That last case is pinned as a unit regression using the exact old string.

## Flips fail through real command routes

Each cell was flipped one at a time and the failure observed on the **command** route, not on a runner call:

| flip | fails on |
|---|---|
| `readUnique.occlusion` → `exclude-and-refuse` | `get attrs` of a covered element |
| `readUnique.offscreen` → `refuse` | `get attrs` (covered + nested-text cases) |
| `readUnique.promotion` → `hittable-ancestor` | `get attrs` answers about the hittable ancestor instead of the matched text |
| `wait.occlusion` → `refuse` | `wait` on a covered match |
| `readAny.offscreen` → `refuse` | `is exists`, `find exists`, `find get_attrs` |
| `promotedTarget.promotion` → `none` | `runtime click still promotes non-touchable nodes` |
| `readList.occlusion` → `exclude-and-refuse` | `find … list` |
| an unconsumed column added to the table | "declares a stage its kind cannot run" |
| `readList` widened to declare node stages | typecheck (unused `@ts-expect-error`) |
| the engine imported outside the owner, in any import form | R19, on the repo-wide scan |

The `wait` route test needed an advancing clock first: with the frozen fixture clock a refused wait spun instead of reaching its deadline, so the flip hung rather than asserting — a test that could not have caught the regression it was written for.

## Live device validation

Repo CLI (`node bin/agent-device.mjs`), dedicated `--state-dir`, local iOS simulator (iPhone 16 Pro, Settings) and Android emulator (emulator-5554, Settings).

| route | row | command | outcome |
|---|---|---|---|
| `get text` | readText | `get text 'label="General"'` | `General` — disambiguated across the cell/text duplicates |
| `get attrs` | readUnique | `get attrs 'label="General"'` | `AMBIGUOUS_MATCH` — fail-closed by design |
| `get attrs` | readUnique | `get attrs 'role=cell label="General"'` | returns the **cell** `e28`, `hittable:false` — `promotion: none` live: a promoting row would have retargeted to a hittable ancestor |
| `is` | readUnique | `is visible 'role=cell label="General"'` | `Passed: is visible` |
| `find … exists` | readAny | `find 'role=cell label="Accessibility"' exists` | `Found: true` |
| `find … list` | readList | `find 'label="General"' list` | 4 matches (cell / other / button / text) — whole candidate set |
| `wait` | wait + poll budget | `wait 'role=cell label="Camera"' --json` | `waitedMs: 255` |
| `click` | promotedTarget | `click 'role=cell label="General"' --settle` | navigated into General (settle diff) |
| `fill` | resolvedTarget | `fill 'label="Search" editable=true' 'accessib' --settle` | search field became `accessib` (`@e69 [search] "accessib"`) |
| `find … click` | findAct → leaf | `find 'role=cell label="Privacy & Security"' click` | `Ref @e65 is off-screen and not safe to click` — the delegation `findAct` declares as `offscreen: 'ignore'`, refused by the acting row in the leaf |
| `find … click` | findAct | `find 'label="Continue"' click` | `Tapped @e8 (201, 829)` |

Android added the occlusion cell's decisive pair — **same screen, same selector, same covered element**:

| route | row | command | outcome |
|---|---|---|---|
| `get attrs` | readUnique (`occlusion: ignore`) | `get attrs 'label="Adaptive Battery"'` | answers, disclosing `"interactionBlocked": "covered"` |
| `click` | promotedTarget (`occlusion: exclude-and-refuse`) | `click 'label="Adaptive Battery"'` | `Selector … is covered by another visible element and cannot be tapped safely` |

Plus on Android: `find list` (1 match), `is visible`, `get attrs`, `wait` (`waitedMs: 743`), `click` (text promoted to its tappable ancestor, opened search), `fill` (`battery` → results). Both sessions closed cleanly.

## Behavior identity

- Acting refusals keep their exact message, label, `ref` and `interactionBlocked` in every branch; promotion declines to retarget away from a covered node, so the "both covered" case names the node it always did.
- The covered-diagnosis probe still names the matched **alternative**, not the whole expression.
- `find` carries its occlusion verdict to the focus/type seam: `find click`/`fill` re-enter the interaction leaf, which owns that refusal's shape (observed live above).
- The native-ref preflight names `resolvedTarget` (promotion `none`) — how ADR 0011's "the preflight never changes which element the backend acts on" becomes a declared property.
- `DEFAULT_WAIT_TIMEOUT_MS` and the 300 ms interval moved into the `wait` row; both wait loops read the values they did.
- ADR 0011's `occlusion`/`nonHittable` `via` pointers now name the runner that makes the decision rather than the predicate it applies.

## Validation

`pnpm check:affected --run`, `pnpm check:layering` (now including R19 and the rule-id uniqueness gate), `pnpm check:fallow`, `pnpm check:production-exports` (21 findings, unchanged from `origin/main`), typecheck, lint, format — all green, rebased on `origin/main`. No SessionState changes. No CLI/user-facing surface change, so no docs or skills updates.

## Not routed

`requireInteractive` stays a per-call-site boolean: it selects which **capture** to take, not one of the four stages, and folding it in would change what this PR is about.

### DIFF
--- a/packages/contracts/src/interaction-guarantees.ts
+++ b/packages/contracts/src/interaction-guarantees.ts
@@ -150,9 +150,12 @@ const SHARED_RESPONSE_CONSTRUCTION: GuaranteeEnforcement = {
 // guard/observation implementations; only how the target is found
 // (disambiguation) and how failures are described (errorTaxonomy) differ.
 const RUNTIME_TREE_SHARED_GUARANTEES = {
+  // #1656: the decision point is the pipeline runner, which reads the acting
+  // row's occlusion stage; isSnapshotNodeInteractionBlocked stays the
+  // predicate it applies (and the annotation contract it reads).
   occlusion: {
     kind: 'runtime',
-    via: 'src/snapshot/snapshot-occlusion.ts#isSnapshotNodeInteractionBlocked',
+    via: 'src/core/selector-pipeline-policy.ts#resolveSelectorPipelineTarget',
   },
   parentOwnedTouchPoint: {
     kind: 'runtime',
@@ -171,9 +174,11 @@ const RUNTIME_TREE_SHARED_GUARANTEES = {
     kind: 'runtime',
     via: 'src/commands/interaction/runtime/resolution.ts#throwIfOffscreenInteractionTarget',
   },
+  // Promotion runs only for rows that declare it (#1656); the retarget itself
+  // is still resolveActionableTouchResolution.
   nonHittable: {
     kind: 'runtime',
-    via: 'src/core/interaction-targeting.ts#resolveActionableTouchResolution',
+    via: 'src/core/selector-pipeline-policy.ts#resolveSelectorPipelineTarget',
   },
   responseConstruction: SHARED_RESPONSE_CONSTRUCTION,
   responseIdentity: {
@@ -249,7 +254,7 @@ export const INTERACTION_DISPATCH_PATHS: Record<InteractionPathId, InteractionPa
       },
       occlusion: {
         kind: 'runtime',
-        via: 'src/snapshot/snapshot-occlusion.ts#isSnapshotNodeInteractionBlocked',
+        via: 'src/core/selector-pipeline-policy.ts#resolveSelectorPipelineTarget',
       },
       parentOwnedTouchPoint: {
         kind: 'runtime',
@@ -370,7 +375,7 @@ export const INTERACTION_DISPATCH_PATHS: Record<InteractionPathId, InteractionPa
       },
       occlusion: {
         kind: 'runtime',
-        via: 'src/snapshot/snapshot-occlusion.ts#isSnapshotNodeInteractionBlocked',
+        via: 'src/core/selector-pipeline-policy.ts#resolveSelectorPipelineTarget',
       },
       parentOwnedTouchPoint: {
         kind: 'inapplicable',
--- a/packages/contracts/src/interaction-guarantees.ts
+++ b/packages/contracts/src/interaction-guarantees.ts
@@ -155,7 +155,7 @@ const RUNTIME_TREE_SHARED_GUARANTEES = {
   // predicate it applies (and the annotation contract it reads).
   occlusion: {
     kind: 'runtime',
-    via: 'src/core/selector-pipeline-policy.ts#resolveSelectorPipelineTarget',
+    via: 'src/core/selector-pipeline.ts#runNodePipelineStages',
   },
   parentOwnedTouchPoint: {
     kind: 'runtime',
@@ -178,7 +178,7 @@ const RUNTIME_TREE_SHARED_GUARANTEES = {
   // is still resolveActionableTouchResolution.
   nonHittable: {
     kind: 'runtime',
-    via: 'src/core/selector-pipeline-policy.ts#resolveSelectorPipelineTarget',
+    via: 'src/core/selector-pipeline.ts#runNodePipelineStages',
   },
   responseConstruction: SHARED_RESPONSE_CONSTRUCTION,
   responseIdentity: {
@@ -254,7 +254,7 @@ export const INTERACTION_DISPATCH_PATHS: Record<InteractionPathId, InteractionPa
       },
       occlusion: {
         kind: 'runtime',
-        via: 'src/core/selector-pipeline-policy.ts#resolveSelectorPipelineTarget',
+        via: 'src/core/selector-pipeline.ts#runNodePipelineStages',
       },
       parentOwnedTouchPoint: {
         kind: 'runtime',
@@ -375,7 +375,7 @@ export const INTERACTION_DISPATCH_PATHS: Record<InteractionPathId, InteractionPa
       },
       occlusion: {
         kind: 'runtime',
-        via: 'src/core/selector-pipeline-policy.ts#resolveSelectorPipelineTarget',
+        via: 'src/core/selector-pipeline.ts#runNodePipelineStages',
       },
       parentOwnedTouchPoint: {
         kind: 'inapplicable',
--- a/packages/selectors/package.json
+++ b/packages/selectors/package.json
@@ -4,7 +4,7 @@
   "private": true,
   "sideEffects": false,
   "type": "module",
-  "description": "Shared selector matching, argument, and replay semantics for agent-device. `.` is string-only; `./ast` is the published parser surface behind `agent-device/selectors`.",
+  "description": "Shared selector matching, argument, and replay semantics for agent-device. `.` is string-only; `./ast` is the published parser surface behind `agent-device/selectors`; `./engine` is the resolve/list surface reserved for the selector-pipeline owner (R19).",
   "dependencies": {
     "@agent-device/ad-script": "workspace:*",
     "@agent-device/contracts": "workspace:*",
@@ -18,6 +18,10 @@
     "./ast": {
       "types": "./src/ast.ts",
       "default": "./src/ast.ts"
+    },
+    "./engine": {
+      "types": "./src/engine.ts",
+      "default": "./src/engine.ts"
     }
   },
   "devDependencies": {
--- a/packages/selectors/src/engine.ts
+++ b/packages/selectors/src/engine.ts
@@ -1,0 +1,24 @@
+/**
+ * The matching engine's own door (#1656).
+ *
+ * "Resolve a selector against a screen" and "list what it matched" are the two
+ * decisions a pipeline policy row exists to qualify: which nodes may be
+ * candidates, whether several matches refuse or collapse, and which structural
+ * stages run around the answer. A caller that reaches these directly gets the
+ * ambiguity contract and silently skips every structural stage — the failure
+ * #1649 caught in the first policy matrix and #1656's review caught in the
+ * second.
+ *
+ * So they live behind a subpath of their own rather than on the root façade,
+ * and R19 selector-pipeline-ownership admits exactly one importer:
+ * `src/core/selector-pipeline.ts`. A specifier is what the import graph
+ * resolves, so namespace imports, dynamic imports, and re-exports are all the
+ * same edge and all equally refused — which a name-shaped check could not say.
+ *
+ * Everything else selectors publishes (parsing, matching, formatting, replay)
+ * stays on the root façade: none of it decides what a row decides.
+ */
+export {
+  listSelectorChainMatches,
+  resolveSelectorChainWithPolicy,
+} from './internal/facade-engine.ts';
--- a/packages/selectors/src/index.test.ts
+++ b/packages/selectors/src/index.test.ts
@@ -1,14 +1,15 @@
 import assert from 'node:assert/strict';
+import { resolveSelectorChainWithPolicy } from './engine.ts';
 import { test } from 'vitest';
 import type { SnapshotNode } from '@agent-device/kernel/snapshot';
 import * as selectorsFacade from './index.ts';
+import * as selectorsEngine from './engine.ts';
 import {
   buildSelectorCandidates,
   readReplaySelectorDisplayValue,
   readSelectorExpression,
   resolveRecordedTarget,
   resolveReplaySuggestionCandidate,
-  resolveSelectorChainWithPolicy,
   SELECTOR_RESOLUTION_POLICIES,
 } from './index.ts';
 
@@ -201,16 +202,36 @@ test('replay suggestion resolution and display values stay string-only at the fa
  * preserved semantics — so no fixture-tree assertion can catch a revert
  * (#1715 review). The absence of the symbol is the only observable.
  */
-test('the façade exposes no resolver that bypasses the policy matrix', () => {
+test('the façade exposes no resolver at all, and the engine door exposes only the two entries', () => {
   const exported = Object.keys(selectorsFacade);
-  assert.ok(exported.includes('resolveSelectorChainWithPolicy'));
-  assert.ok(!exported.includes('resolveSelectorChain'), 'knob-taking resolver must stay private');
+  // #1656 moved both engine entries behind `./engine`, which R19 reserves for
+  // the selector-pipeline owner: a route that could reach a resolver from the
+  // root façade would get an ambiguity contract while skipping every
+  // structural stage its policy row declares.
   assert.ok(
-    !exported.includes('findSelectorChainMatch'),
-    'count-only existence lookup must stay private; `is exists` names the readAny row',
+    !exported.includes('resolveSelectorChainWithPolicy'),
+    'resolution belongs to the engine subpath, behind the pipeline owner',
   );
   assert.ok(
-    !exported.includes('selectorResolutionKnobs'),
-    'knob derivation must stay private so a call site cannot rebuild a contract from knobs',
+    !exported.includes('listSelectorChainMatches'),
+    'enumeration belongs to the engine subpath, behind the pipeline owner',
   );
+  assert.deepEqual(Object.keys(selectorsEngine).sort(), [
+    'listSelectorChainMatches',
+    'resolveSelectorChainWithPolicy',
+  ]);
+
+  // Unchanged since #1630, on both surfaces: neither door hands out a knob
+  // resolver a call site could rebuild its contract from.
+  for (const [surface, names] of [
+    ['facade', exported],
+    ['engine', Object.keys(selectorsEngine)],
+  ] as const) {
+    assert.ok(!names.includes('resolveSelectorChain'), `${surface}: knob-taking resolver`);
+    assert.ok(
+      !names.includes('findSelectorChainMatch'),
+      `${surface}: count-only existence lookup; \`is exists\` names the readAny row`,
+    );
+    assert.ok(!names.includes('selectorResolutionKnobs'), `${surface}: knob derivation`);
+  }
 });
--- a/packages/selectors/src/index.ts
+++ b/packages/selectors/src/index.ts
@@ -59,6 +59,7 @@ export type { IsPredicate } from './internal/predicates.ts';
 export type {
   PolicyResolutionOutcome,
   SelectorChainMatchList,
+  SelectorMatchOptions,
   SelectorResolution,
 } from './internal/public-resolution-types.ts';
 export { formatSelectorFailure } from './internal/resolve.ts';
--- a/packages/selectors/src/index.ts
+++ b/packages/selectors/src/index.ts
@@ -1,11 +1,4 @@
-import type { SnapshotState } from '@agent-device/kernel/snapshot';
 import type { Selector } from './internal/parse.ts';
-import type {
-  PolicyResolutionOutcome,
-  SelectorChainMatchList,
-  SelectorMatchOptions,
-} from './internal/public-resolution-types.ts';
-import { resolveSelectorChainWithPolicy as resolveSelectorChainWithPolicyAst } from './internal/resolve-with-policy.ts';
 import {
   checkElementTargetArgs,
   checkGetFormat,
@@ -30,11 +23,7 @@ import {
   IS_PREDICATE_USAGE_HINT,
   normalizeIsPositionals,
 } from './internal/predicates.ts';
-import {
-  listSelectorChainMatches as listSelectorChainMatchesAst,
-  selectorFailureHint,
-  STALE_REF_HINT,
-} from './internal/resolve.ts';
+import { selectorFailureHint, STALE_REF_HINT } from './internal/resolve.ts';
 import {
   findBestMatchesByLocator,
   checkFindArgs,
@@ -81,7 +70,6 @@ export {
   isRoleHintWord,
   isSelectorToken,
   isValidSelectorExpression,
-  listSelectorChainMatches,
   normalizeIsPositionals,
   normalizeSelectorText,
   parseFindArgs,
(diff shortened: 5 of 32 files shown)


thymikee:
Reviewed exact head `a09c67700b02923e0f29d9e96b617f436a96ba25`.

**P1 — the structural policy is still bypassable, so several declared cells are not enforced on shipped routes.** Acting resolution is row-bound, but read/wait callers manually compose `selectorPipelineCandidates(row, nodes)` with the still-public raw `resolveSelectorChainWithPolicy(..., row.resolution)`. They never enter `resolveSelectorPipelineTarget` or `throwIfOffscreenInteractionTarget`. Consequently, flipping a read row’s `promotion` or `offscreen` value changes only the synthetic policy/runner tests, not production `get`/`is`/`wait` behavior—the same unverifiable-column failure #1656 exists to remove. The current tests demonstrate `readUnique.offscreen` by directly calling the off-screen guard with that row even though no observation command route calls it. Callers can also mix one row’s candidates with another row’s resolution, and `find list` remains an acknowledged direct bypass.

Please provide one owning core interface that accepts one row and executes every stage, including explicit no-op stages, or add a structural route/import gate that makes bypass impossible. The planted flips must fail through real command routes, not only direct runner calls.

The routed behavior otherwise appears preserved and all completed CI/device smoke lanes are green. Readiness still needs practical device evidence for the changed selector paths (`click`/`fill`/`find`/`get`/`is`/`wait`); the PR body currently reports no exact live scenario, and the Swift compile job remains pending.

thymikee:
You were right, and the diagnosis was exact: the read/wait routes composed `selectorPipelineCandidates(row, nodes)` with the raw resolver and never entered promotion or off-screen, so flipping a read row's cell moved only my policy tests. That is the same unverifiable-column failure one level down.

Took the **owning-interface** shape you preferred. It does not fight the architecture as much as I first thought — the acting route's capture retry and replay guard were the parts I expected to resist, and both fit: the retry stays a capture concern above the owner (the retry condition is "nothing matched", so the stages never run on a discarded attempt), and the replay guard is a hook the owner calls on the resolution winner *before* promotion, which is exactly where it ran by hand.

`src/core/selector-pipeline.ts` now owns the engine and the stages, with the stage functions private to it:

- `resolveSelectorPipeline` — single-target rows: candidacy → ambiguity → replay-guard hook → promotion → occlusion → off-screen.
- `listSelectorPipelineMatches` — `reject-candidates` rows; returns the candidate set **and the tree the row sees**, because find's ranking and the acting equivalence classification have to judge the same nodes candidacy produced (passing the unfiltered tree there would have let an overlay-covered node come back as a competitor).
- `runNodePipelineStages` — the node stages for a target from a non-chain matcher (`@ref`, find's fuzzy locator, a narrowed candidate set).

Three further pieces closing the specific holes you named:

- **Mixing rows is gone**: the row enters once, at the top, and the later stages read it off that call.
- **A declared refusal must exist**: a row with `offscreen: 'refuse'` must supply a refusal shape, so an observation row flipped to `refuse` fails on its route instead of silently observing.
- **`find list`** no longer reaches the engine; it names a new `readList` row (`reject-candidates`, no rect requirement — every match is the answer).

And the structural backstop, since an interface is only owning while nothing can route around it: **R17 selector-pipeline-ownership** (`scripts/layering/`) restricts `resolveSelectorChainWithPolicy` / `listSelectorChainMatches` to the owner. Proven by planting the import back into `selector-read.ts` and watching the repo-wide scan reject it, not just its unit test.

Flips now fail through the command routes, verified one at a time: `readUnique.occlusion`/`offscreen`/`promotion` and `wait.occlusion` via `get attrs` / `is` / `wait`; `readAny.offscreen` via `is exists` and `find`; `readList.occlusion` via `find … list`; `promotedTarget.promotion` via runtime `click`. One finding worth flagging: the `wait` route test needed an advancing clock first — with the frozen fixture clock a refused wait spun instead of reaching its deadline, so the planted flip *hung* rather than asserting. A test that cannot fail the way the regression would is the same class of problem as the column itself.

Live validation is in the PR body (iOS simulator + Android emulator, repo CLI, dedicated `--state-dir`). The occlusion cell got the pair I was hoping for on Android — same screen, same selector, same covered element:

```
get attrs 'label="Adaptive Battery"'  →  answers, disclosing "interactionBlocked": "covered"
click     'label="Adaptive Battery"'  →  Selector … is covered by another visible element and cannot be tapped safely
```

and `find … click` on an off-screen row refused through the leaf, which is the delegation `findAct` declares as `offscreen: 'ignore'` rather than a stage it skips.

thymikee:
Reviewed exact head d0de2f71. The production selector routes are now substantially centralized, the device evidence is adequate, and all authoritative CI is green. Three blockers remain:

- **P1 — `readList` still makes unverifiable full-row claims.** The shipped `find list` route calls `listSelectorPipelineMatches`, which consumes only candidacy/occlusion and resolution; it never runs promotion, off-screen, or poll stages. The command regression proves covered-node candidacy only, while the remaining `readList` cells are exercised synthetically. Narrow candidate-list policy to the stages that actually apply, or make the command route execute every declared stage so planted flips fail the shipped route.
- **P2 — R17 does not structurally enforce ownership.** `selector-pipeline-ownership.ts` recognizes only static named imports. Namespace imports, re-exports, and dynamic imports can still reach the public raw engine and bypass the owner. Use a package/private construction seam or enforcement that covers the real import graph rather than this regex subset.
- **P1 — the new rule ID already collides with open PR #1750.** This head declares `R17 selector-pipeline-ownership`, while #1750 declares `R17 device-inventory-cutover`; merging both silently recreates the duplicate-rule collision that #1750 exists to prevent. Allocate a unique rule identity and validate the stacked/merge-order case.

Please address these before merge.

thymikee:
All three addressed at `ab7f050f0`.

**1 — `readList`: narrowed the shape (your recommendation).** Promotion, the off-screen guard and a poll budget genuinely cannot apply to a listing: it resolves to a candidate set, so there is no single element to retarget, keep on screen, or wait for. `readList` now declares the two stages a listing executes — `SelectorListPolicy` = `resolution` + `occlusion` — and the narrowness is load-bearing rather than documentation, because `runNodePipelineStages` and `selectorPollBudget` take the full row:

```ts
// @ts-expect-error a listing row declares no promotion or off-screen stage.
const nodeStages: Parameters<typeof runNodePipelineStages>[0] = SELECTOR_PIPELINE_POLICIES.readList;
```

Widening `readList` back to five columns makes those directives unused and fails the typecheck — verified by doing it. Its one live cell is pinned on the shipped route: flipping `readList.occlusion` fails `find … list`.

**2 — enforcement moved off names and onto the import graph.** You were right that the regex was the wrong instrument, and the reason is sharper than "regexes are brittle": a namespace import, a re-export and a deferred `import()` never mention the symbol it was matching, so the check was blind precisely where a bypass would be deliberate. The two engine entries now live behind their own subpath, `@agent-device/selectors/engine`, and R19 enforces on the SPECIFIER over resolved edges — where all of those forms are the same edge. Planted into a shipped route and rejected by the repo-wide scan, not just the unit test:

```
import * as selectorEngine from '@agent-device/selectors/engine'   → red
await import('@agent-device/selectors/engine')                     → red
export * from '@agent-device/selectors/engine'                     → red
```

One thing worth flagging, since it makes the specifier shape a trade rather than a free win: `resolveImportEdges` **drops** an edge whose specifier resolves to nothing. Retire the subpath and a specifier-keyed rule goes quiet rather than red — it would scan clean forever. The gate now reports that as its own failure instead.

I kept the root façade closed as well: it exposes no resolver at all now, and its surface test pins both doors.

**3 — R19, and the collision class itself.** Verified free against `origin/main` and #1750's diff, then validated by real merges in both directions rather than a file-level union (the union was misleading: `check.ts` is touched by both branches, so picking one side wholesale invented conflicts a merge would not have). `#1750 → #1744` and `#1744 → #1750` both merge cleanly, the uniqueness gate passes in both, and the ids stay distinct — `R17 device-inventory-cutover`, `R18 contracts-implementation-authority`, `R19 selector-pipeline-ownership`.

There was no uniqueness gate to validate, which is why the collision happened: two branches claiming one free number do not conflict in git, because the files differ. `scripts/layering/rule-ids.ts` adds one. Two notes on it:

- It matches **whole string literals**. My first version keyed on `rule:` and therefore missed your `const RULE = '…'` declarations — a gate that could not see the very collision it exists for. Whole-literal matching is also what separates a declaration from prose that names a rule (`'R17 holds the devices command to…'`).
- `main` already carries two collisions — R11 and R13 each name two rules — which #1750 is renaming away. They are listed as known rather than pinned by equality, so #1750 lands in either order without breaking this PR, and the entries fall away with the rename.

Head for re-review: **`ab7f050f0`**. Still a draft; not undrafting, per your gate on re-review.

thymikee:
Re-reviewed exact head `ab7f050f`. The two prior selector-policy findings are fixed correctly: `readList` is narrowed to the only stages a listing can execute (resolution + occlusion), with node-stage/poll entry points rejecting it at the type boundary; and the dedicated `@agent-device/selectors/engine` subpath plus resolved-edge R19 ownership closes named, namespace, dynamic-import, re-export, type-import, and side-effect-import bypasses. Keep R19 after #1750 allocates R17/R18.

**P1 — the transitional collision allowlist fails open forever.** `KNOWN_RULE_ID_COLLISIONS` filters the exact old R11/R13 collision strings. Calling stale entries “inert” is incorrect: after #1750 removes those collisions, reintroducing either exact collision later is silently accepted. Rebase after #1750, require zero current collisions and delete the transitional allowlist, or make an allowance itself fail once its collision is absent. Add a planted regression using the exact old collision string to prove it is rejected post-transition.

The branch is also currently DIRTY/CONFLICTING, and only security-analysis checks ran for this head. Resolve/rebase, inspect the conflict resolution (especially rule allocation, package exports, and layering wiring), then require the full Layering/Typecheck/Static/Integration/Coverage/smoke suite. Prior iOS/Android evidence can carry if the resolution remains purely structural; rerun the affected routes if selector production semantics change. This head is not ready and should not be labeled.

thymikee:
Head correction: **`ac05290af`**, not `ab7f050f0`. The branch was conflicting after #1746 landed, so I rebased onto `origin/main`.

The conflict was substantive rather than textual: #1746 replaced the array of rule spreads in `check.ts` with the `LAYERING_RULES` registry, so R19 is now registered as a keyed entry (`'selector-pipeline-ownership'` in `LAYERING_RULE_IDS` + the registry) instead of another spread — which is the shape that PR added precisely so a rule cannot be run and reported twice. `package.json` also took main's side: `check:layering` now globs `scripts/layering/*.test.ts`, so it picks up `selector-pipeline-ownership.test.ts` and `rule-ids.test.ts` without a hand-maintained list.

Re-validated after the rebase: `check:affected --run`, `check:layering` (R19 live, R11 now at 40 subpaths with `./engine` counted), and both merge orders against #1750 — still clean, ids still distinct.

thymikee:
Both addressed at **`ff57b0908`**.

**1 — allowances now expire with the collisions they allow (option (b); #1750 is still open).** You were right that "inert" was wrong: filtering the exact R11/R13 strings meant that once those rules are renamed apart, the entries would keep waving those very collisions through if anyone reintroduced them — fail-open, permanently.

`ruleIdCollisionFailures` now checks the transition from both sides:

- a collision nobody allowed fails, and
- **an allowance whose collision is absent from the scan fails as a stale allowance**.

So the entry has to be deleted in the same change that removes the collision, and the list burns down to empty — which admits nothing.

Verified against a scratch tree carrying #1750's rename rather than only synthetic strings:

```
allowance left behind after the rename : 2 failures (both entries reported stale)
after deleting the entries             : 0 failures
reintroducing the exact old collision  : 1 failure
  two rules answer to one id: R11 names contracts-implementation-authority
  and package-boundaries. Allocate the next free number.
```

That last case is also a unit regression using the exact old string, so the post-transition guarantee is pinned rather than argued.

**2 — rebase.** Already on latest `origin/main` (`057ab1c82`); the DIRTY/CONFLICTING snapshot predated my push, and GitHub now reports MERGEABLE. The conflict resolution, re-checked deliberately since you flagged it:

- **#1746's registry preserved.** R19 is a keyed entry — `'selector-pipeline-ownership'` in `LAYERING_RULE_IDS` plus its `LAYERING_RULES` function — not another spread. Confirmed no hand-spread rule array survived anywhere in `check.ts`.
- **`package.json` took main's side**: `check:layering` globs `scripts/layering/*.test.ts`, so both new test files are picked up without a hand-maintained list (strictly better than what I had).
- **Package exports**: `./engine` is declared in the selectors manifest and counted by R11 (40 subpaths), with the reviewed subpath pin updated to name it and why.

Full suite green locally: `pnpm check` (tooling → format/lint/typecheck/layering/depgraph/production-exports/mcp-metadata/build/package, plus fallow and the 818-file unit run) and `pnpm check:affected --run`. One honest note on the way there: `check:unit` failed once on the tmpdir-leak gate, which was correct — an earlier `check:affected` run of mine that I killed on a timeout had skipped its `globalTeardown` and left `/tmp/agent-device-test-run-87207-*` behind. My artifact, not the tree's; removed, and the gate went green.

Live evidence carries: the resolution is structural (rule registration, export map, test glob) with no change to selector production semantics. I did re-run the Android routes against the rebuilt CLI anyway, since the package-export change is the one thing that could break the shipped binary — `find … list`, `get attrs` on the covered node (answers, `interactionBlocked: "covered"`), and `click` on the same node (refuses) all behave as before.

On CI: the failure on the previous head was the Android emulator lane losing its host services — `Unable to connect to adb daemon on port: 5037`, a late step failing with `processExitError: true`, and cleanup failing too (`stop: Not implemented`). Not one of the named flake signatures, so I checked it rather than rerunning blind: the run reached late E2E steps, so the CLI started and resolved the new subpath fine, and nothing in this diff touches adb or process spawning. The new head is running now.

Still a draft; no readiness label until your exact-head re-review.

thymikee:
Reviewed exact head `ff57b090`. The prior P1 is fixed: collision allowances now expire with the collision they name, so an unallowed present collision fails and a stale allowance also fails; the exact old R11 regression proves it cannot be silently reintroduced after the transition. R19’s resolved-specifier engine ownership remains intact across named, namespace, dynamic, re-export, type, and side-effect imports, and `readList` remains correctly narrowed to resolution + occlusion and cannot enter node-stage or poll runners.

All authoritative CI is green, including Layering, Typecheck, Static, Coverage, Integration, Fallow/FreeRange, native compile/releases, and all smoke lanes. The prior live iOS/Android route evidence still applies because the rebase/follow-up did not change selector behavior. No actionable finding remains; ready-for-human.

Integration note: #1744 can land first safely. When #1750 rebases/lands, it must delete the now-stale transitional R11/R13 entries, preserve device R17 / contracts R18 / selector R19, and consolidate the parallel rule-ID gates into one derived namespace mechanism during conflict resolution.

PR #1745, opened 2026-08-11
refactor(layering): one parametrized runtime-command-cutover gate (ADR 0019 §8)

thymikee:
## Summary

Consolidate the four existing ADR 0019 cutover policies into one table-driven gate. Each migrated command adds one typed row; the shared policy enforces row completeness, request-scoped versus durable-tier claims, retired paths, admission, runtime narrowing, and singular execution.

Named runtime operations declare a lexical owner, and call counting is restricted to that owner subtree. An unrelated daemon call can neither satisfy a missing owner call nor make a correct owner call look duplicated; planted tests prove both arms.

This is structural-only: no production behavior or device path changes.

Key invariants:

- one parameterized gate and table; no per-command fifth policy
- discriminated, omission-safe row model with exact named-operation equality and duplicate rejection
- `devices` R17, `logs` R14, `network` R15, `record` R16; contracts R18 and selector ownership R19 are preserved
- typed `LAYERING_RULES` registry and glob-based test discovery are preserved
- 20 attributable changed files, 2,215 additions / 1,846 deletions versus the merged #1750 base

Stack order is now encoded in GitHub: merged #1750 → this PR (`e9785ebc`) → #1747. The final semantic replay onto #1750 is complete; the temporary R11 collision allowance is gone.

## Validation

- `pnpm check:affected --run && git push` passed before GitHub's native stack rewrite; the rewritten exact head is `e9785ebc`
- layering 168/168
- daemon wire surface unchanged from v0.20.8
- exact-head GitHub checks are green except the still-running iOS smoke lane
- no live-device run required for this structural-only gate change

Docs/skills were not changed because the public CLI, daemon behavior, and user workflow are unchanged.

Part of #1739.


thymikee:
Reviewed exact head `4a31e6d4`. I found no production behavior change, and the four existing cutover claim families appear preserved, but three readiness blockers remain.

- **P1 — incomplete rows can pass silently.** `MigratedCommandCutover` makes every enforcement column optional, while `rowViolations` simply skips absent columns. A row containing only `{ rule, command, subject }` is green, so ADR 0019 §8’s promise that a new unit “adds a row” and inherits the mechanism is not omission-safe. Please make the row model discriminated/validated with mandatory claims for its execution kind/tier (including mandatory singular execution proof), and add a planted incomplete-row test that fails. The current planted row proves only a fully populated row.
- **P1 — stacked dependency #1740 must fix occurrence attribution before rebase.** Its CLI-route coherence scan collapses dispatches to command-name sets, so one attributed `runtime` dispatch can mask a second, unrelated/unattributed `runtime` occurrence. Track dispatch occurrence identity/location, add the attributed-plus-duplicate-unattributed planted red, land #1740, then rebase this PR onto that fixed head.
- **P2 — test module remains over the extraction tripwire.** The newly created `runtime-command-cutover-policy.test.ts` is 608 LOC. Tests are not exempt; split the generic mechanism coverage from coherent per-row/table acceptance coverage into matching modules while preserving shared helpers and the ported cases. Production module sizes are otherwise within budget.

The PR is currently draft and MERGEABLE/UNSTABLE. Integration, Maestro, and release-selection lanes are green; Layering, Typecheck, Static, Coverage, Fallow, and several smoke/security lanes are still pending and remain authoritative merge gates.

thymikee:
Prerequisite stack update: #1740 advanced from 8ad1292b to corrected head 4c1c4297 after the gate finding. This PR is still based on the old #1740 head. Please replay #1745 onto 4c1c4297, resolve the line-sized registry/gate conflicts, and rerun pnpm check:affected --run before merge/readiness. The five downstream ADR 0019 workers are intentionally holding publication until this corrected head is available.

thymikee:
All three addressed; rebased onto the fixed #1740 head (`4c1c42979`) first so the occurrence-identity change is underneath this one.

**P1 incomplete rows.** You're right that all-optional columns made §8's "add a row, inherit the mechanism" untrue — `{rule, command, subject}` enforced nothing and read green. `MigratedCommandCutover` is now a union discriminated on `execution` (`inventory` | `device-runtime`) with mandatory claims per kind: an inventory row must name the gateway proof that is its singular-execution evidence, and a device-runtime row must name its runtime types, its operations (exact set or family), and its singular daemon route. Rows also declare a mandatory `tier`, and `durable-resource` requires a lifecycle proof while `request-scoped` forbids one. Non-empty tuple types catch "declared but empty" at compile time; `cutoverRowDefects` catches what types cannot — and reads defensively, since a row omitting whole claim objects is exactly its input. The gate runs it per row before it scans the repo, so an under-declared row reports `(<command> cutover row): cutover row declares no ...` instead of passing. The planted incomplete row is a permanent test, alongside four narrower defect plants (retires nothing, empty singular execution, durable without lifecycle proof, request-scoped claiming durable machinery) and an assertion that every shipped row is defect-free.

Two real gaps surfaced while filling the mandatory columns: `logs` had no singular-route proof at all, so it now claims `handleLogsCommand` (verified singular in `session-observability.ts`), and `record`'s daemon-mechanics scan became its lifecycle proof, moving out of check.ts into the row — I confirmed by planting a `setTimeout` in `record-runtime.ts` that it still fires exactly once, not twice. The registry-join scan stays its own module since it is a descriptor-shape check, not a cutover claim.

**P2 test split.** 608 LOC became `runtime-command-cutover-policy.test.ts` (166, generic mechanism + row-completeness plants), `runtime-command-cutover-table.test.ts` (481, per-row acceptance with every ported case intact), and `runtime-command-cutover-fixtures.ts` (51, shared helpers and the planted row). The table test is over the 300 target and under the 500 tripwire; splitting it further would separate ported cases from the rows they accept, so I left it whole and am flagging it rather than hiding it.

Validation: `pnpm check:layering` 147/147 with the guard green, `check:affected --run` clean, `tsc` clean.

thymikee:
Confirmed new head bf67324e is correctly replayed onto #1740 head 4c1c4297 (ahead by one, behind by zero). Fresh CI now has an Integration Tests failure while other lanes continue. Please independently inspect and fix or prove contention per docs/agents/testing.md, rerun pnpm check:affected --run before the next push, and keep this draft. Also note #1740 itself still has the unresolved unknown/computed-target scanner P2, so bf67324e remains provisional even after CI is green.

thymikee:
Re-reviewed exact head `bf67324e`. The prior whole-row fail-open and test-size findings are substantially fixed: incomplete top-level claims now fail, the policy/table tests are 166/481 LOC with shared fixtures, and the new logs-route and record-lifecycle proofs strengthen the migrated rows. Three blockers remain:

- **P1 — the branch recreates the R13 collision.** Its current base already renumbered device-inventory cutover to R17, but the new table and header still declare `R13 device-inventory-cutover`. Change the row and all local references to R17 so it does not collide again with `R13 platform-package-substrate`; keep the namespace uniqueness gate green.
- **P1 — exact named operations can still omit singular-operation enforcement.** `DeviceRuntimeCutover.singularExecution.operations` is optional even when `operations.names` is an exact nonempty list. A future row can name its runtime operations, omit their exact call-count proof, and pass validation. Require the singular operation set for named-operation claims (preferably the same set), while permitting route-only singularity only for pattern-only families such as logs, and add a planted incomplete-row red.
- **P1 — stacked #1740 still drops unknown dispatch targets.** Occurrence attribution is now correctly source-offset based and has a duplicate-target regression, but `dispatchTargetOf` filters through `knownCommand`; an unknown literal or catalog target returns `undefined`, never enters `allSites`, and cannot become `unattributed`. Capture unknown dispatch occurrences as explicit failures and add a planted routed-known plus stray-unknown case in #1740. Land that fix, then rebase this PR.

All current CI is green and GitHub is CLEAN/MERGEABLE, but these structural gaps still block readiness. No production behavior defect was found.

thymikee:
All three done; head for re-review is `4dcfa58ba`, rebased onto #1740's `5b4f1e599` (the unknown-target fix).

**1. Rule ID.** The row and every local reference now read `R17 device-inventory-cutover`; the check.ts rule-map bullet says `devices R17`, and no `R13 device-inventory` string remains anywhere. I validated the namespace in **both** merge orders with an audit that reads every `RULE` constant plus the table's per-row ids:

- **#1745 first (what CI sees now):** R13 resolves to `platform-package-substrate` alone — my half of your collision is gone. R11 still names both `package-boundaries` and `contracts-implementation-authority`, which is your other half and untouched by me; collisions go 2 → 1, never up.
- **#1750 first:** I test-rebased onto `6b02dfe0a` locally. It conflicts exactly where expected — you edit `device-inventory-cutover-policy.ts`, which this PR deletes (delete/modify), plus the check.ts rule-map bullet. Resolution is to keep the deletion and keep both bullets (your R18 contracts line, my parametrized-gate line, already numbered R17). After that the namespace is fully unique R1-R18 with zero collisions and `check:layering` is 150/150 green.

Either order lands on the same map, so the renumber does not depend on who merges first.

**2. Named-operation singularity.** You're right that knowing the exact operation set and enforcing only the route left the operations free to be called twice or never. `DeviceRuntimeCutover` is now a union of two variants: a **named-operation** row must declare `singularExecution.operations`, typed `NonEmpty<string>` so it cannot be omitted or empty; a **pattern-only** row (logs, whose app-log plan picks operations from the request) declares `operations?: undefined` and proves singularity through its route alone. `cutoverRowDefects` enforces the same claim for `as`-cast rows and additionally requires the enforced set to *cover* the named set, so declaring three operations and enforcing one is rejected by name. Three planted reds cover it: enforces-none, enforces-some (`names operations it does not enforce exactly once: plantedReattach`), and a positive test that the pattern-only logs row is legitimately route-only. `network` and `record` already enforced their full named sets, so no production claim changed.

**3. Rebase** onto `5b4f1e599` is in the pushed head.

Validation: `check:layering` 150/150 with the guard green, `check:affected --run` clean, `tsc` clean. Left in draft — not requesting readiness this round.

thymikee:
Reviewed exact head 4dcfa58b. This branch is currently DIRTY/CONFLICTING after #1746, and the exact head restores the old hand-spread rule array and manual layering-test list, including the duplicate `checkContractsImplementationAuthority(sources)` call. The required rebase must preserve #1746’s typed `LAYERING_RULES` registry and `scripts/layering/*.test.ts` glob, then register the consolidated cutover exactly once. Only CodeQL ran on this head; authoritative CI evidence is absent.

[P2] Named-operation completeness is still one-way. `namedOperationDefects()` rejects named operations missing from `singularExecution.operations`, but extra or duplicate enforced operations pass even though the model says the enforced set must be the named set. Require symmetric set equality plus uniqueness and plant an extra-operation/duplicate red.

This stack also inherits #1740’s confirmed variable-envelope/computed-callee fail-open, so it cannot be ready ahead of a corrected #1740. Preserve the devices R17 / contracts R18 allocation when reconciling #1750. The split test/engine topology sizes themselves are within the project limits. No readiness label.

thymikee:
Both fixed; head is `897e50553`, rebased onto #1740's `09808532b`, which itself sits on `057ab1c82` (#1746).

**Rebase repair.** You're right about what my previous head did, and I can see how: I resolved the earlier conflicts by taking my side of `check.ts` and `package.json` wholesale, which silently reverted #1746. Redone properly — #1746's artifacts are preserved and verified individually:

- `LAYERING_CONTEXT` / `LAYERING_RULE_IDS` / `LAYERING_RULES` keyed registry intact; `main()` still runs `Object.values(LAYERING_RULES).flatMap(...)`.
- `check:layering` is back to the `scripts/layering/*.test.ts` glob, so my three new suites are discovered without registration (157 tests, up from 146 on main).
- `checkContractsImplementationAuthority` appears exactly once — `grep -c` is 1.
- `check-wiring.test.ts` passes, so catalog and registry still describe the same set.

The consolidation is registered as **one** entry: `'runtime-command-cutover'` replaces the four keys `logs-runtime-cutover`, `network-runtime-cutover`, `record-runtime-cutover` and `device-inventory-cutover`, plus `'record-runtime-registry-join'` for the descriptor-shape scan that stays its own module. Reporting order and the success line are unchanged.

**Symmetric equality.** `namedOperationDefects` now compares the two sets in both directions and rejects duplicates on each side: an unenforced name leaves an operation free to be called twice or never, and an enforced name the row does not claim asserts singularity over an operation the command does not own. Four planted reds: enforces-none, enforces-some, enforces-an-unnamed-operation, and duplicates on both sides.

Worth reporting: the duplicate plant failed on first run and caught a real bug in my check — I had written `values.filter((value) => !seen.add(value))`, and since `Set.add` returns the set that predicate is always false, so duplicates were never detected. The test went red exactly as a planted red should, which is the only reason it isn't still there.

Rule IDs unchanged from the last round: `devices` is `R17`, and the namespace audit still gives R13 → `platform-package-substrate` alone. The remaining R11 collision is #1750's contracts half, untouched here; both merge orders were re-validated last round and the reconciliation is unchanged.

Validation: `check:layering` 157/157 green, `check:affected --run` clean, `tsc` clean. Still draft, queued behind #1740.

thymikee:
Adversarial re-review of exact head `897e50553`: the prior row-completeness, symmetric named-operation, R17 allocation, and #1746 registry-preservation findings are fixed; current CI is fully green/CLEAN. One new structural blocker remains.

[P1] `exactCallViolations()` counts every `runtime.operations.<operation>` call across all `src/daemon/**` files (`scripts/layering/runtime-command-cutover-policy.ts:429-441`) without associating the call with this row's handler or the binding narrowed by that handler. An unrelated handler can therefore satisfy a row's required sole call, while a correct later unit that reuses an operation (for example `screenRecordingCleanup`) becomes a false duplicate. This makes the gate non-compositional precisely when wave 2+ adds rows. Scope singular-operation proof to the declared handler/binding identity, and keep a planted unrelated-handler call that must not satisfy or duplicate the row.

Please address independently, rerun `pnpm check:affected --run`, and keep draft. The PR also remains provisional behind corrected #1740 and must preserve the rule catalog when reconciling #1750.

thymikee:
Reviewed exact head 897e5055. The prior local blockers are fixed: the typed LAYERING_RULES registry and scripts/layering/*.test.ts discovery are preserved, each rule is registered once, named/enforced operations are validated symmetrically with duplicate rejection, and the test topology remains within limits. The branch is CLEAN/MERGEABLE and all authoritative CI is green.

This is still not ready for human merge review:

- contracts-implementation-authority remains allocated to R11, colliding with package-boundaries. The intended R18 allocation exists only in open, DIRTY #1750, so dependency order/rebase remains unresolved.
- This stack inherits exact parent #1740 confirmed P2: injected targets are not constrained to the registered daemon-command union, and raw transport can escape through an arbitrary helper callsite.

Please land/fix #1740 and #1750, then rebase this stack while preserving the registry/glob and R17/R18 allocations. No ready-for-human label yet.

thymikee:
Rebased onto #1740's corrected head (`a63a99f5c`) and onto main including #1744. New head: `800b6da83`. No local blockers outstanding; still queued behind #1740 and #1750.

Two things the rebase surfaced, both handled:

- **R19 survives.** #1744's `selector-pipeline-ownership` is registered alongside the consolidated cutover in `LAYERING_RULES`; the registry, the glob, and the single `contracts-implementation-authority` entry are all intact (`grep -c` is 1). 166 layering tests, green.
- **The new collision gate caught my renumber's other half.** `KNOWN_RULE_ID_COLLISIONS` allowed `R13 names device-inventory-cutover and platform-package-substrate`, and moving devices to R17 made that allowance stale — the gate failed with *"no longer occurs, so the entry admits a collision nobody is fixing. Delete it."* Exactly the expire-on-contact behaviour it documents. I deleted the R13 entry and left the R11 one for #1750's R18 rename, so the list still admits only what is actually present.

Current namespace with this head: R13 → `platform-package-substrate` alone, R17 → `device-inventory-cutover`, R19 → `selector-pipeline-ownership`, and R11 still shared pending #1750. When #1750 lands I will rebase again, delete the remaining allowance if #1750 has not already, and report the head.

Validation: `check:layering` 166/166 with the guard green, `check:affected --run` clean, `tsc` clean.

thymikee:
Adversarial follow-up is fixed at exact head `6deff6845`. Named runtime operations now declare a lexical owner, and the cutover gate counts each operation only inside that owner subtree; an unrelated daemon call can neither satisfy a missing owner call nor false-duplicate the correct call. The planted test covers both arms. The branch is rebased onto current `main`, preserves selector ownership/R19, and `pnpm check:affected --run && git push` passed on the exact head (481 files / 3,967 tests; changed-line and changed-branch coverage 100%). No production or live-device behavior changed.

thymikee:
Exact head `6deff6845` is now fully green on GitHub. The final iOS smoke rerun passed end to end; its prior two attempts failed in two different untouched tests, confirming contention rather than an attributable regression. The owner-scoped cutover proof, planted unrelated-call coverage, exact-head affected gate, scope accounting, and review comments are all resolved.

#1745 is technically ready relative to prerequisite #1740. It remains draft only because #1740 is still open; after #1740 merges, perform the final replay onto `main` and retain green checks before marking ready.

thymikee:
Re-reviewed exact head `6deff684`. The attributable delta is locally clean: the table-driven consolidation preserves the typed `LAYERING_RULES` registry and test glob, lexical operation-owner mapping closes the prior unrelated-call/singularity gap, the planted proof is meaningful, and all authoritative checks are green on a CLEAN/MERGEABLE branch. It is not stack-ready yet: it inherits #1740's confirmed arbitrary `{ transport: sendToDaemon }` object-handoff escape. The new rule-ID registry is fail-closed, but this head intentionally retains the temporary R11 collision allowance for `contracts-implementation-authority` and `package-boundaries`; #1750 must integrate its R18 rename and the stale allowance must then be removed. No device evidence is required for this structural-only delta. No readiness label.

thymikee:
The linked stack finding is addressed at exact head `2ba50c502`: #1745 now inherits corrected #1740 `fac5294b7`, whose transport-handoff gate rejects arbitrary `{ transport: sendToDaemon }` object handoffs. The lexical-owner callsite proof remains intact. `pnpm check:affected --run && git push` passed (481 files / 3,970 tests; 100% changed-line coverage). The PR stays draft for the explicit final dependency: corrected #1750 `9c53b9186` must merge, then this branch will replay to remove the temporary R11 allowance before readiness.

thymikee:
Exact-head coordinator update at d8f7aee6d: the semantic replay onto #1750 is complete and code review is clean. The typed LAYERING_RULES registry/glob discovery are preserved; devices/logs/network/record remain R17/R14/R15/R16; #1750 supplies contracts R18; the temporary R11 allowance is gone. Local pnpm check:affected --run passed, including layering 168/168 and unchanged daemon wire surface. GitHub stack #1768 now encodes #1750 → #1745 → #1747 → #1756. Fresh exact-head CI is running; no live-device evidence is required for this structural-only gate change.

PR #1749, opened 2026-08-11
fix(ios): make custom-action reads atomically single-flight

thymikee:
## Summary

Keep custom-action AX reads truly single-flight under contention. Admission now atomically claims the only in-flight slot, and rejected reads return without queue dispatch, semaphore waiting, or synchronous per-refusal logging.

The regression asserts the production contract directly: five blocked admissions, one dispatched read, and one read still in flight. It no longer uses a wall-clock threshold.

## Root cause

The previous admission split its check and increment across two atomic operations, so concurrent callers could both observe an empty slot before either claimed it. The blocked branch also logged synchronously, making the XCTest's elapsed-time assertion measure runner scheduling and log delivery rather than the no-work containment invariant.

The fix uses one compare-and-exchange for admission and exposes the blocked-admission count alongside the existing dispatch and in-flight counters.

## Validation

- Planted red: the targeted regression failed with blocked admissions `0` instead of `5` before production incremented the new counter.
- Targeted `testHungCustomActionReadIsContainedAndRecovers` XCTest passed after the fix.
- All seven custom-action XCTest regressions from the iOS smoke step passed.
- `pnpm check:affected --run` passed; the Swift runner build/smoke lanes remain GitHub-authoritative.

Three files changed within the existing Apple runner snapshot module. No docs or skills changed because the public command behavior and guidance are unchanged.

### DIFF
--- a/apple/runner/AgentDeviceRunner/AgentDeviceRunnerUITests/RunnerAXSnapshotBridge.h
+++ b/apple/runner/AgentDeviceRunner/AgentDeviceRunnerUITests/RunnerAXSnapshotBridge.h
@@ -97,6 +97,9 @@ FOUNDATION_EXPORT NSString *const RunnerAXSnapshotCustomActionsBlockedKey;
 /// adds no work", which is only observable as this counter standing still.
 + (NSInteger)customActionReadDispatchCount;
 
+/// Total reads refused by single-flight admission.
++ (NSInteger)customActionReadBlockedCount;
+
 /// The shared AX client (`XCUIDevice.accessibilityInterface`), or nil when the
 /// private interface is unavailable.
 + (nullable id)accessibilityClient;
--- a/apple/runner/AgentDeviceRunner/AgentDeviceRunnerUITests/RunnerAXSnapshotBridge.m
+++ b/apple/runner/AgentDeviceRunner/AgentDeviceRunnerUITests/RunnerAXSnapshotBridge.m
@@ -63,6 +63,9 @@ static atomic_int RunnerAXCustomActionReadsInFlight = 0;
 /// staying put, so it is recorded rather than inferred.
 static atomic_long RunnerAXCustomActionReadDispatches = 0;
 
+/// Total single-flight admission refusals.
+static atomic_long RunnerAXCustomActionReadBlocked = 0;
+
 /// Output caps. The element budget bounds how many elements we read; these
 /// bound what any ONE element can put in the response, so a pathological app
 /// cannot spend the whole snapshot on one node's action list.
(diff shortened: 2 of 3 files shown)


thymikee:
Reviewed exact head `e7b02cad`. **P2 regression validity:** raising the repeat bound to 1.5s exceeds the production AX read deadline of 1.0s, so the test can now admit a full-deadline refusal regression while still claiming repeats “return immediately instead of paying the deadline.” A planted 1.0–1.4s stall before the atomic refusal would pass.

The exact in-flight and dispatch-count assertions remain strong for the real no-work containment invariant: removing the guard would enqueue additional reads and fail those counts. Please choose and test one contract deterministically:

- If the contract is “blocked captures add no AX work,” remove the misleading wall-clock/immediate claim and expose/assert an explicit blocked admission result or blocked counter alongside unchanged dispatch/in-flight state.
- If bounded refusal latency is itself required, expose an injectable refusal/logging seam and prove the blocked branch performs no semaphore wait or queue dispatch. Do not use an XCTest wall-clock threshold equal to or greater than the downstream timeout.

The observed 1.1s old red demonstrates that the 0.2s wall-clock assertion is contention-sensitive; it does not validate 1.5s as a latency regression bound. The current failed Android smoke is unrelated to this Swift-test-only diff: API 36 live E2E failed with `Active application interaction viewport is unavailable`, an Android helper/device-state flake that should be rerun. Remaining pending authoritative lanes still gate merge.

thymikee:
Addressed in `ae9a6c707` by replacing the timing contract with deterministic production bookkeeping.

- Admission now uses one atomic compare-and-exchange instead of a separate load and increment.
- Rejected admissions increment an explicit blocked counter and return without queue dispatch, semaphore wait, or synchronous per-refusal logging.
- The regression asserts five blocked admissions, one dispatch, and one in-flight read; the wall-clock assertion and its long explanatory comment are gone.

Red/green evidence: before wiring the production blocked counter, the targeted XCTest failed with `0` blocked admissions instead of `5`; after the fix it passes, as does the full seven-test custom-action subset.

I also checked #1767. It touches the fixture alert flow and live iOS alert smoke scenario only, so there is no file or behavior overlap with this AX custom-action single-flight fix.

thymikee:
Exact-head re-review at `ae9a6c70`: code-clean and merge-ready. The prior timing-test finding is fixed at the owning interface rather than relaxed: admission is now one atomic compare-and-exchange, blocked reads increment an explicit refusal counter, and the regression deterministically proves five refusals with exactly one dispatch and one read still in flight. The wall-clock assertion and misleading immediate-return contract are gone; synchronous per-refusal logging is also removed from the blocked path. Red-before evidence is load-bearing (`0` blocked admissions before production bookkeeping, `5` after), the full custom-action XCTest subset passes, and all authoritative exact-head checks—including iOS, Android, macOS, Linux, Coverage, Integration, Fallow, and Layering—are green. Branch is CLEAN/MERGEABLE. Applied `ready-for-human`.

PR #1750, opened 2026-08-11
refactor(layering): give each colliding rule id its own number

thymikee:
## Summary

Give each layering rule number one meaning and make future collisions fail at the guard boundary.

- Keep device inventory at R17, contracts implementation authority at R18, and selector-pipeline ownership at R19.
- Derive the rule catalog from the guard's own non-test policy sources and reject ambiguous IDs before findings are grouped or annotated.
- Remove duplicate contracts-authority reporting and retire the temporary collision allowances now that the allocation is clean.
- Correct ADR 0019's platform-package rule reference from R11 to R13.

This is a structural-only change: five files, 13 additions / 14 deletions versus its exact base, with no production bundle delta or live-device requirement.

## Validation

- exact head `d32f70298`
- full authoritative GitHub CI is green, including all four smoke lanes
- the derived catalog reports unique rule IDs through R19
- planted regressions cover the former R11/R13 collisions, syntax variants, comments/prose, and the exact post-fix allocation

Docs were updated only for ADR 0019's corrected rule reference. Skills were not changed because CLI behavior and workflow guidance are unchanged.

### DIFF
--- a/docs/adr/0019-request-bound-platform-runtime.md
+++ b/docs/adr/0019-request-bound-platform-runtime.md
@@ -111,7 +111,7 @@ only that composition module may import a concrete platform package; reusable ty
 through type-only platform imports. Platform packages may import contracts, kernel/domain packages,
 and explicitly injected host capabilities; they may not import daemon requests or responses, mutable
 session state, command catalogs/grammar, root implementation files, sibling platform packages, or raw
-process primitives outside the shared host-command port. R11 applies these rules to static, type-only,
+process primitives outside the shared host-command port. R13 applies these rules to static, type-only,
 dynamic, and re-export edges; package-owned tests may import their own public façade. Contracts may
 depend on kernel vocabulary but never on concrete platform packages or daemon implementation types.
 
--- a/scripts/layering/check.ts
+++ b/scripts/layering/check.ts
@@ -33,10 +33,12 @@
 //     owner" shape as R7's SessionState ownership, applied to bin.ts's `--help` fast path.
 //   - Over PLATFORM PACKAGE COMPOSITION: six private metadata façades meet at the exact root
 //     composition file; premature implementation loading and forbidden cross-boundary edges fail (R13).
-//   - Over the DEVICES COMMAND CUTOVER: the handler calls the neutral inventory gateway and no
-//     superseded inventory module, import, or identifier remains in production (R13).
 //   - Over COMMAND-ATOMIC RUNTIME CUTOVERS: retired logs, network, and record routes/admission cannot
 //     coexist with their operation-fact-derived descriptor and handler paths (R14-R16).
+//   - Over the DEVICES COMMAND CUTOVER: the handler calls the neutral inventory gateway and no
+//     superseded inventory module, import, or identifier remains in production (R17).
+//   - Over CONTRACTS PRODUCTION SOURCE: contracts owns vocabulary only — host, process, and timer
+//     mechanics belong in capture-kit or an adapter (R18).
 // Only `(root)` is unranked among src/ zones (see `UNRANKED_ZONES` in model.ts):
 // it holds entrypoints and composition roots. Extracted workspace package zones
 // are classified separately and held behind R11 instead of the src folder spine.
--- a/scripts/layering/contracts-implementation-policy.ts
+++ b/scripts/layering/contracts-implementation-policy.ts
@@ -3,7 +3,7 @@ import type { LayeringViolation } from './model.ts';
 
 export type ContractsProductionSource = Readonly<{ path: string; source: string }>;
 
-const RULE = 'R11 contracts-implementation-authority';
+const RULE = 'R18 contracts-implementation-authority';
 const FORBIDDEN_HOST_MODULES = /^(?:node:)?(?:child_process|fs|timers)(?:\/|$)/;
 const FORBIDDEN_TIMER_CALLS = new Set([
   'clearImmediate',
--- a/scripts/layering/device-inventory-cutover-policy.ts
+++ b/scripts/layering/device-inventory-cutover-policy.ts
@@ -1,7 +1,7 @@
 import { parseSync } from 'oxc-parser';
 import type { LayeringViolation } from './model.ts';
 
-const RULE = 'R13 device-inventory-cutover';
+const RULE = 'R17 device-inventory-cutover';
 const HANDLER_FILE = 'src/daemon/handlers/session-inventory.ts';
 const INVENTORY_IMPORT_SOURCES = new Set([
   '../../core/device-inventory-context.ts',
@@ -253,5 +253,5 @@ export function checkDeviceInventoryCutover(
 }
 
 export function deviceInventoryCutoverSummary(): string {
-  return 'the devices command has one gateway-owned inventory route and no legacy route';
+  return 'R17 holds the devices command to one gateway-owned inventory route with no legacy route';
 }
--- a/scripts/layering/rule-ids.ts
+++ b/scripts/layering/rule-ids.ts
@@ -54,15 +54,12 @@ export function duplicateRuleIds(declarations: readonly RuleDeclaration[]): stri
 }
 
 /**
- * Collisions that predate this gate: `main` reuses R11 and R13 for two rules
- * each, which #1750 is renaming to R17/R18. Transitional, and each entry
- * expires on contact — see `ruleIdCollisionFailures`. Empty is the end state,
- * and the gate gets there on its own.
+ * Empty, which is the end state: the R11 and R13 collisions this list was opened
+ * for are gone (contracts-implementation-authority moved to R18, device-inventory-cutover
+ * to R17), so their allowances expired and were deleted with them. Every id now names
+ * exactly one rule, and an entry here would admit a collision nobody is fixing.
  */
-export const KNOWN_RULE_ID_COLLISIONS: readonly string[] = [
-  'R11 names contracts-implementation-authority and package-boundaries',
-  'R13 names device-inventory-cutover and platform-package-substrate',
-];
+export const KNOWN_RULE_ID_COLLISIONS: readonly string[] = [];
 
 /**
  * Both halves of the transition, because an allowance that outlives the thing


thymikee:
Reviewed exact head `6b02dfe0`. The renumbering itself is coherent: the current namespace is one-to-one from R2 through R18, R11 package-boundaries and R13 platform-package-substrate are the right keepers, and R17 devices before R18 contracts is reasonable. All authoritative checks are green and GitHub is CLEAN/MERGEABLE. Three readiness items remain:

- **Make the collision class impossible.** There is no committed test or construction path enforcing that one numeric ID maps to exactly one rule name; the PR currently relies on a one-off grep, and the moved policy tests do not assert their IDs. Prefer one keyed rule catalog from which declarations/report labels derive, or add the smallest derived uniqueness gate over authoritative rule declarations. Prove it red on the pre-fix collision and green here.
- **R18 findings are still duplicated.** `check.ts` invokes `checkContractsImplementationAuthority(sources)` twice (around lines 663 and 666), so one contracts violation produces duplicate counts and annotations. Land the #1746 correction first and rebase, or remove the duplicate invocation here and prove one planted contracts violation reports once.
- **Correct the normative ADR reference.** ADR 0019 line 114 says R11 applies the platform-package import rules, while CONTEXT correctly assigns those rules to R13. Update that exact R11 → R13 reference as part of making the namespace unambiguous.

Also ensure stacked PR #1745 rebases its devices cutover row from R13 to R17; otherwise it recreates this collision after merge.

thymikee:
Dependency update for exact heads `6b02dfe0` (#1750) and `d0de2f71` (#1744): branch-local uniqueness is insufficient. #1744 newly declares `R17 selector-pipeline-ownership`, while this PR declares `R17 device-inventory-cutover`. The declarations live in different files and the `check.ts` edits occupy separate hunks, so both branches can merge without a textual conflict while silently recreating the numeric collision.

Please coordinate/rebase the allocation using the historical ordering this PR established: device inventory stays R17, contracts stays R18, and the newer selector-pipeline ownership rule becomes R19. The #1745 devices table row must likewise rebase to R17. Run the derived uniqueness enforcement on the combined/rebased tree—not only each branch in isolation—before declaring the namespace collision-free.

thymikee:
All three readiness items addressed in `ba53691` + `6c3ee64`.

**1. Collision class made impossible.** New `scripts/layering/rule-catalog.ts` derives the catalog from the guard's own sources instead of restating it in a list that can drift. Two deliberate choices:

- It matches **rule-shaped string literals in any syntactic form** rather than `const RULE =` / `rule:` call sites, so a new declaration form can't be invisible to it. The anchored pattern keeps summary prose that merely opens with a number (`R13 holds six private …`) out of the catalog, and AST literals keep commented-out ids out too — both covered by tests.
- It **reads the directory** rather than taking a file list, so the catalog can't fall out of step with the policies.

This is a gate, not only a test: `check.ts` now refuses to report through a colliding namespace *before* grouping anything by rule, as a guard-integrity fault rather than one more finding under an ambiguous id. Proved red on the real pre-fix tree:

```
Layering guard: rule namespace is not one-to-one

  R11 names 2 rules: contracts-implementation-authority, package-boundaries
  R13 names 2 rules: device-inventory-cutover, platform-package-substrate

  Give each rule its own number before the report can name what fired.
exit=1
```

Green here, with the OK line now reporting `the guard's own 17 rule ids each name exactly one rule (R2-R18)`. `rule-catalog.test.ts` plants both exact pre-fix collisions as red cases. 136 → 147 tests.

**2. R18 double-reporting fixed.** Removed the duplicate `checkContractsImplementationAuthority(sources)` spread. Proved by planting one contracts violation (a `setTimeout` in `packages/contracts/src/`, staged so the tracked-files-only scan reaches it) and counting annotations: **2 before, 1 after**. I fixed it here rather than waiting on #1746 — both branches delete the same line, so it merges cleanly, and #1746's `check-wiring.ts` remains the structural guard against re-introduction. I deliberately did not duplicate that guard.

**3. ADR reference corrected.** `docs/adr/0019:114` R11 → R13.

---

**Cross-branch allocation (your follow-up).** Verified by local trial merge, not in isolation — and the new gate catches both cases:

| Combined tree | Result |
|---|---|
| #1750 + #1744 as-is | ✗ `R17 names 2 rules: device-inventory-cutover, selector-pipeline-ownership` |
| #1750 + #1744, selector-pipeline → **R19** | ✓ clean, 150 tests, `18 rule ids … (R2-R19)` |
| #1750 + #1745 as-is | ✗ `R13 names 2 rules: device-inventory-cutover, platform-package-substrate` |
| #1750 + #1745, table row → **R17** | ✓ clean |

So the allocation you proposed holds. The two edits needed on the sibling branches:

- **#1744** — `scripts/layering/selector-pipeline-ownership.ts:61` `rule:` → `R19`, plus the `R17` in its header comment at line 2.
- **#1745** — `scripts/layering/runtime-command-cutover-table.ts:19` `rule:` → `R17`, plus the "R13 devices" mention in that file's header comment.

I haven't pushed to either branch — say the word if you want me to.

**One integration hazard found while doing this:** #1745 renames `record-runtime-policy-ast.ts` → `cutover-policy-ast.ts`, and my catalog was importing `visitAst` from it, so the merge failed to resolve the module. Fixed in `6c3ee64` by collecting string literals locally — a guard-integrity check that breaks when a neighbour is renamed is the exact fragility this rule exists to remove, and a rule catalog shouldn't depend on a cutover-specific module either way.

**Separately, out of scope:** `record-runtime-mechanics-policy.test.ts` and `record-runtime-registry-policy.test.ts` exist but nothing runs them — `check:layering` hand-lists 14 of the 16 layering test files. They pass (3 tests) and just need wiring; that hand-maintained list is the same drift shape as this PR's. Filed separately rather than folded in.

thymikee:
Re-reviewed exact head `6c3ee646`. The prior code findings are fixed correctly: the derived AST/string-literal catalog scans the guard’s non-test policy sources, rejects collisions before report grouping, and carries planted reds for the exact old R11/R13 collisions without a permanent allowlist; the duplicate contracts-authority invocation is removed and pinned to one report occurrence; and ADR 0019 now correctly attributes the platform-package import rules to R13.

This head is still **not ready** because GitHub is DIRTY/CONFLICTING and only CodeQL/security-analysis jobs ran. Rebase and resolve the conflicts, inspect the combined rule allocation and layering wiring, then require the full Layering, Typecheck, Static, Integration, Coverage, and smoke suite. Preserve device inventory as R17 and contracts authority as R18. #1744 should keep selector ownership as R19 and replace its parallel `rule-ids`/transitional-allowlist implementation with this catalog; #1745 must rebase its devices row to R17.

**P2 — the live PR body is materially stale.** It still says three files, “No doc changes needed,” no policy test asserts the moved IDs, and a one-off grep is the uniqueness verification. Exact head now changes docs, package/test wiring, adds the derived catalog and planted regressions, and removes duplicate reporting. Update Summary, Validation, and dependency notes after the rebase so reviewers see the current design and evidence.

thymikee:
Rebased onto current main and resolved the R19 selector integration at exact head `9c53b9186`. The obsolete transitional `rule-ids` allowance/test pair is deleted; the derived catalog is now the single collision guard. `pnpm check:affected --run && git push` passed in full. GitHub exact-head checks are now authoritative; #1745 remains draft until this dependency merges and is replayed.

thymikee:
Rebased onto `main` at `057ab1c` — conflicts resolved, now `MERGEABLE`. #1746 landing changed two things in this PR's favour:

- **The R18 dedup is no longer mine to carry.** #1746 replaced the inline spread array with the keyed `LAYERING_RULES` registry, and an object cannot hold a key twice, so the copy-pasted invocation cannot come back. I resolved the `check.ts` conflict to main's registry and dropped my one-line deletion; the review item is satisfied structurally rather than by my edit. Reworded the commit so it no longer claims otherwise.
- **My `package.json` change is gone entirely.** #1746 also replaced the hand-maintained `check:layering` test list with a glob over `scripts/layering/*.test.ts`, so `rule-catalog.test.ts` is picked up automatically. That also closes the orphaned-test gap I flagged above — `record-runtime-mechanics-policy.test.ts` and `record-runtime-registry-policy.test.ts` now run. No separate fix needed.

Net diff against main is 6 files: the three renumbered surfaces, the ADR line, and the two new catalog files. `pnpm check:layering` green at **152 tests**, namespace summary `the guard's own 17 rule ids each name exactly one rule (R2-R18)`; format and lint clean.

Worth noting the two registries are complementary, not redundant: `LAYERING_RULE_IDS` keys are unnumbered rule *names* and stop a rule running twice; `rule-catalog.ts` derives number→name and stops two rules sharing a number. Neither subsumes the other.

The cross-branch allocation from my previous comment is unaffected — #1744 still needs selector-pipeline at R19, #1745 still needs its table row at R17.

thymikee:
Fixed, and the shape of this PR changed with it. Rebased onto `97c87ee`.

**What the failure was.** `rule-ids.test.ts` landed on `main` after my last rebase — a rule-ID uniqueness gate with a transitional `KNOWN_RULE_ID_COLLISIONS` allowlist naming the exact two collisions this PR renames apart, and an allowance expires on contact. So renaming them made both entries stale and the gate failed, correctly and by design. Deleting them in the same change that removes the collisions is what it was asking for, and the list is now empty.

**I removed my own catalog.** `main`'s `rule-ids.ts` is the same derived-uniqueness gate I'd built in `rule-catalog.ts` — same directory scan, same test-file exclusion, same number→name check — plus the allowance-expiry half mine didn't have. Two registries answering the same question is the duplication this PR exists to remove, so I deleted mine (`rule-catalog.ts`, `rule-catalog.test.ts`, and the `check.ts` wiring and OK-line summary) and kept `main`'s. Your first readiness item is satisfied by the merged gate, not by me; I've stopped carrying a second one.

One difference worth knowing, not worth a change: `rule-ids.ts` matches single-quoted literals by regex, so a double-quoted declaration wouldn't be seen and a rule id quoted inside a comment would be. `oxfmt` normalises quotes, so the first is close to unreachable; I mention it only so the limit is on record.

**The cross-branch coordination landed.** #1744 merged with selector-pipeline-ownership at **R19** — the allocation verified by trial merge above. Combined namespace on this branch is now one-to-one across **R2-R19**:

```
R2 commands-floor            R11 package-boundaries          R16 record-runtime-cutover
R3 platforms-seam            R12 bin-alias-fast-path         R17 device-inventory-cutover
R4 value-import-cycle        R13 platform-package-substrate  R18 contracts-implementation-authority
R5 zero-back-edges           R14 logs-runtime-cutover        R19 selector-pipeline-ownership
...                          R15 network-runtime-cutover
```

no duplicates. #1745 is the last one outstanding — its table row still needs R13 → R17, and with the allowlist empty the gate will now reject it rather than let it merge quietly.

**Net diff is 5 files, 13 insertions / 14 deletions**: the three renumbered surfaces, the ADR line, and the retired allowances. I also rebuilt the branch into two commits, since adding a catalog and then deleting it was history noise. `pnpm check:layering` green at 150 tests; format and lint clean.

thymikee:
Final coordinator verdict at d32f70298: clean and merge-ready. The patch is five structural files (+13/-14), allocations are unique (devices R17, contracts R18, selector R19), temporary allowances are retired, and every authoritative check including all device smoke lanes is green. No live-device evidence is required for this layering-only change.

PR #1751, opened 2026-08-11
fix(daemon): make request-timeout cleanup and hints platform-aware

thymikee:
Part of #1739 (independent cleanups)

**Update (978dbfaa0):** maintainer review found a real correctness inversion
in the first pass (see review thread below) — this revision redesigns the
fix around a strict split between cleanup eligibility and hint wording. The
section below is rewritten to describe the current design; see PR history
for the superseded first pass.

## What

Every local request timeout ran the Apple-runner `xcodebuild` pkill cleanup
(`IOS_RUNNER_XCODEBUILD_KILL_PATTERNS`) and emitted Apple-specific hint text
("Apple runner work was aborted", "Timed-out Apple runner xcodebuild
processes were terminated") regardless of the session's actual platform —
misleading for Android/web/Harmony timeouts.

## Investigation: what platform context is available at this layer, and why it's not authoritative

`src/daemon/client/` is outside the R3 platforms-seam and stays import-free
of `src/platforms`. Both timeout call sites in `daemon-client-transport.ts`
hold the exact `req: DaemonRequest` they sent, and `req.flags?.platform` is
the client-declared `--platform` selector for that request.

The first pass of this PR used that flag directly to decide whether to run
the Apple cleanup at all. Maintainer review (thymikee) found this backwards
in two ways, traced to `src/daemon/request-lock-policy.ts`:

1. **`applyStripLockPolicy`'s existing-session branch** (reached via
   `--session-lock strip`) strips a conflicting declared selector and
   backfills `flags.platform` from the *session's real device*
   (`publicPlatformString(existingSession.device)`) — server-side, after the
   client has already sent the request. A request that declared
   `--platform android` against an existing Apple-bound session can
   therefore legitimately execute Apple work. Skipping the pkill cleanup
   because the client saw `android` would skip cleanup that was actually
   needed — the dangerous direction.
2. **The common session-bound request omits `--platform` entirely** (no
   conflict to strip, so `flags.platform` stays `undefined` all the way
   through normalization too). Treating `undefined` as "fail-safe toward
   Apple" — my first pass's choice — meant the hint kept its Apple wording
   on exactly this common route, so the *motivating* bug (misleading Apple
   hints on ordinary Android/web/Harmony session timeouts) stayed unfixed.

**Eligibility conclusion:** at the client layer, a session-bound request is
never provably non-rebindable to Apple from its own flags alone — there is
no session-state lookup available here, by design (no `src/platforms`
import, no new plumbing). So the redesign does not try to prove
non-rebindability at all.

## Fix — split eligibility from wording

- **Cleanup eligibility is unconditional again** for every local (non-remote)
  timeout — the same behavior as before this PR touched the file. The pkill
  patterns are Apple-process-name-specific, so sweeping them on a non-Apple
  host/session matches nothing and costs a few no-op subprocess spawns —
  never a wrong skip. This removes the dangerous-direction bug entirely: the
  sweep can no longer be skipped based on a stale/absent client-side flag.
- **The hint is evidence-based**, not flag-based:
  `appleCleanupEvidence = isAffirmativelyApplePlatform(platform) ||
  cleanup.terminated > 0` — true only when the request declared an
  affirmatively Apple platform (`apple`/`ios`/`macos`), or the cleanup
  itself actually terminated a matching process (proof positive regardless
  of what `--platform` claimed). Any other combination — undeclared
  platform, declared non-Apple platform, zero processes terminated — gets
  platform-neutral wording. This is what fixes the motivating bug on its
  common route: an undeclared-platform Android/web/Harmony timeout runs the
  (harmless, no-op) cleanup, terminates nothing, and the hint goes neutral.
- Both `sendSocketRequest` and `sendHttpRequest` in
  `daemon-client-transport.ts` still pass `req.flags?.platform` through —
  it now only feeds the affirmative-platform half of the hint's evidence
  check, never cleanup eligibility.

## Before/after hint matrix

| resetDaemon | appleCleanupEvidence | Hint |
| --- | --- | --- |
| false | true (explicit Apple selector, or cleanup terminated something) | *(unchanged historical wording)* "...The timed-out `<command>` request was canceled and Apple runner work was aborted when detected; the daemon was kept alive..." |
| false | false (no Apple selector and nothing terminated) | *(new)* "...The timed-out `<command>` request was canceled; the daemon was kept alive..." (no Apple claim) |
| true | true | *(unchanged)* "...Timed-out Apple runner xcodebuild processes were terminated when detected." |
| true | false | *(new)* "...The daemon was reset after the timeout." (no Apple claim) |
| n/a | remote | *(unchanged)* platform-independent remote wording |

`snapshot`'s iOS-prepare follow-up sentence is also gated on
`appleCleanupEvidence` and drops entirely when it's false.

## Testing

**Pure formatter matrix** (`src/utils/__tests__/daemon-client.test.ts`):
extended with a fixture test asserting the full before/after matrix above
against `resolveRequestTimeoutHint` directly, byte-for-byte.

**Production-seam route tests** (new file,
`src/daemon/client/__tests__/daemon-client-timeout-route.test.ts`) — added
because the pure formatter test cannot catch a bug in cleanup *eligibility*,
only in hint *wording*. This file spies on the real process-execution seam
(`runCmdSync`, `src/utils/exec.ts`, via `vi.mock`) and drives an actual
socket/HTTP timeout through the real `sendRequest` against a hanging local
server, asserting on the real rejected `AppError`. Four cases:

1. **Socket, rebound-session simulation**: declares `platform: 'android'`,
   pkill mock reports a termination → cleanup still runs (3 pkill calls
   asserted) AND the hint claims Apple, on the termination evidence — proves
   the dangerous-direction bug is gone.
2. **HTTP, unknown-session simulation**: undeclared platform, pkill
   terminates nothing → cleanup still runs (3 pkill calls asserted) but the
   hint stays neutral — proves the motivating bug is fixed on its common
   route.
3. **HTTP, explicit Apple platform**: declares `platform: 'ios'`, pkill
   terminates nothing → hint still claims Apple (affirmative-selector
   evidence, honors the existing "when detected" hedge).
4. **HTTP, remote** (`info.baseUrl` set): pkill never invoked at all;
   remote-specific hint.

**Proved red before the fix**, per repo convention: `git stash push --
src/daemon/client/daemon-client-timeout.ts` reverted just that file back to
this PR's first (already-pushed, now-superseded) commit `82aa5ade2`, then
reran the new route-test file. 2 of 4 cases failed exactly as predicted:
case 1 failed because the flawed code skipped cleanup on the declared
`android` flag (no Apple evidence at all despite the real termination —
the dangerous-direction proof); case 2 failed because the flawed code's
undeclared-platform fail-safe still emitted the Apple claim with zero
evidence (proves the original bug survived under the first pass). Restored
the fix (`git stash pop`) and reran — all 4 green.

`pnpm check:affected --run` (multiple full runs): the unit-test portion
itself has been 100% green across every run (807 test files / 6439 tests,
typecheck/lint/format/layering/build/daemon-wire-compat all pass). Local
runs on this shared host intermittently fail only at the
`check-tmpdir-leaks.ts` gate, flagging `/tmp/agent-device-test-run-*`
directories confirmed to belong to other concurrent sessions on the same
host (verified: the specific flagged directories no longer exist by the
time they're inspected, while several *other* `agent-device-test-run-*`
directories with timestamps spanning many hours sit in `/tmp` at the same
time — not produced by this diff's own test runs). Noting this honestly
rather than claiming a fully clean local run: CI runs `check:affected` on
this pushed head in an isolated runner where that host-sharing noise does
not exist, and is the authoritative signal here.

## Scope

Hint-wording + cleanup-eligibility logic only, no redesign of the timeout
system, no `src/platforms` import added to `src/daemon/client/`.

---

### DIFF
--- a/src/daemon/client/__tests__/daemon-client-timeout-route.test.ts
+++ b/src/daemon/client/__tests__/daemon-client-timeout-route.test.ts
@@ -1,0 +1,241 @@
+// Production-seam coverage for the real request-timeout route.
+//
+// src/utils/__tests__/daemon-client.test.ts covers `resolveRequestTimeoutHint`
+// as a pure formatter, but a pure-formatter test cannot catch a bug in
+// CLEANUP ELIGIBILITY: whether `cleanupTimedOutIosRunnerBuilds` (the Apple
+// xcodebuild pkill sweep) actually runs. This file spies on the real
+// process-execution seam (`runCmdSync`, src/utils/exec.ts) and drives an
+// actual socket/HTTP timeout through `sendRequest` so the assertions exercise
+// the same code path a real client does.
+//
+// Why cleanup eligibility must stay unconditional for local timeouts: the
+// client's declared --platform is not authoritative for session-bound
+// execution. `applyStripLockPolicy` (src/daemon/request-lock-policy.ts) lets
+// an existing session's real device platform silently override a conflicting
+// declared selector under --session-lock strip, and the common session-bound
+// request omits --platform entirely. So a request declaring `platform:
+// 'android'` can still legitimately execute against an Apple-bound session
+// (the "rebound-session" case below), and a request with no platform at all
+// (the "unknown-session" case) is the common route the original bug misled.
+// A design that skips the pkill sweep based on the declared flag alone would
+// skip real cleanup in the rebound case — the dangerous direction. This test
+// proves the sweep always fires for local timeouts, and that the HINT text
+// (not the cleanup) is what carries the platform-evidence gating.
+
+import net from 'node:net';
+import http from 'node:http';
+import os from 'node:os';
+import path from 'node:path';
+import assert from 'node:assert/strict';
+import { beforeEach, test, vi } from 'vitest';
+
+const { mockRunCmdSync } = vi.hoisted(() => ({ mockRunCmdSync: vi.fn() }));
+
+vi.mock('../../../utils/exec.ts', async () => {
+  const actual =
+    await vi.importActual<typeof import('../../../utils/exec.ts')>('../../../utils/exec.ts');
+  return { ...actual, runCmdSync: mockRunCmdSync };
+});
+
+import { AppError } from '@agent-device/kernel/errors';
+import { sendRequest } from '../daemon-client-transport.ts';
+import type { DaemonRequest } from '../../types.ts';
+import type { DaemonInfo } from '../daemon-client-metadata.ts';
+import type { DaemonPaths } from '../../config.ts';
+
+const TIMEOUT_MS = 120;
+
+// `snapshot`'s timeout policy preserves the daemon (onTimeout !==
+// 'reset-daemon'), so `handleRequestTimeout` never reaches
+// `resetDaemonAfterTimeout` (`process.kill`) here — keeping this suite
+// side-effect-free outside the mocked pkill sweep.
+const SNAPSHOT_COMMAND = 'snapshot';
+
+function dummyStatePaths(): DaemonPaths {
+  const baseDir = path.join(os.tmpdir(), 'agent-device-timeout-route-test');
+  return {
+    baseDir,
+    infoPath: path.join(baseDir, 'daemon.json'),
+    lockPath: path.join(baseDir, 'daemon.lock'),
+    logPath: path.join(baseDir, 'daemon.log'),
+    sessionsDir: path.join(baseDir, 'sessions'),
+  };
+}
+
+function buildRequest(platform: 'android' | 'ios' | undefined): DaemonRequest {
+  return {
+    token: 'test-token',
+    session: 'default',
+    command: SNAPSHOT_COMMAND,
+    positionals: [],
+    flags: platform ? { platform } : {},
+    meta: { requestId: 'req-timeout-route' },
+  };
+}
+
+function startHangingSocketServer(): Promise<{ server: net.Server; port: number }> {
+  return new Promise((resolve, reject) => {
+    const server = net.createServer((socket) => {
+      // Accept the connection but never write a response — forces the
+      // client's own request-timeout envelope to fire.
+      socket.on('error', () => {});
+    });
+    server.on('error', reject);
+    server.listen(0, '127.0.0.1', () => {
+      const address = server.address();
+      if (address && typeof address === 'object') {
+        resolve({ server, port: address.port });
+      } else {
+        reject(new Error('failed to bind hanging socket test server'));
+      }
+    });
+  });
+}
+
+function startHangingHttpServer(): Promise<{ server: http.Server; port: number }> {
+  return new Promise((resolve, reject) => {
+    const server = http.createServer((_req, res) => {
+      // Never call res.end() — forces the client's own request-timeout
+      // envelope to fire instead of a real response.
+      res.on('error', () => {});
+    });
+    server.on('clientError', (_err, socket) => socket.destroy());
+    server.on('error', reject);
+    server.listen(0, '127.0.0.1', () => {
+      const address = server.address();
+      if (address && typeof address === 'object') {
+        resolve({ server, port: address.port });
+      } else {
+        reject(new Error('failed to bind hanging http test server'));
+      }
+    });
+  });
+}
+
+beforeEach(() => {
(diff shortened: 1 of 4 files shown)


thymikee:
Reviewed exact head 82aa5ade.

[P1] `req.flags.platform` is the pre-normalization selector and is not authoritative for session-bound execution. In particular, `--session-lock strip` can replace a conflicting declared Android selector with an existing Apple session’s platform inside `prepareLockedRequestBinding()`. This patch therefore sees `android`, skips Apple xcodebuild cleanup, and emits a non-Apple hint even though the request validly executes Apple work. Conversely, the common session-bound request omits `--platform`; treating `undefined` as Apple preserves the misleading Apple wording for Android/web/Harmony timeouts, so the motivating route remains unfixed.

Please separate conservative cleanup eligibility from truthful hint wording: unknown or rebindable requests may retain the Apple cleanup for safety, but mention Apple only with affirmative platform evidence (or evidence that Apple cleanup actually acted); otherwise use neutral wording. Add production-seam executor-spy coverage proving pkill call/skip plus both socket and HTTP timeout propagation, including rebound/unknown-session cases, and show the actual route red before the fix. The current tests exercise only the pure formatter, so they cannot catch inverted or omitted cleanup side effects. No readiness label while this remains.

thymikee:
Addressed at 978dbfaa0.

Traced the mechanism you pointed at: `applyStripLockPolicy`'s existing-session branch (`src/daemon/request-lock-policy.ts`) strips the conflicting declared selector and backfills `flags.platform` from the *session's real device* server-side — after the client has already sent/timed out on the request — so `req.flags.platform` can never be trusted as "definitely not Apple" for a session-bound request. Separately, the common session-bound request never repeats `--platform` at all, so it's `undefined` through the whole normalization path with no session lookup available at this layer to resolve it. Conclusion: a request is never provably non-rebindable to Apple from the client's own flags — I stopped trying to derive that rule and split the two decisions instead, per your instruction.

- **Cleanup eligibility** is unconditional again for every local timeout (the pre-this-PR behavior). The pkill patterns are Apple-process-name-specific, so sweeping them on a non-Apple host/session is a no-op, never a wrong skip — there's no real cost to running it unconditionally, so there was nothing to gain from gating it on a flag that can't be trusted anyway.
- **Hint wording** is evidence-based: `appleCleanupEvidence = isAffirmativelyApplePlatform(platform) || cleanup.terminated > 0` — true only on an explicit `apple`/`ios`/`macos` selector or the cleanup itself having terminated a matching process. This is what fixes the motivating route: an undeclared-platform Android/web/Harmony timeout still runs the (harmless) sweep, terminates nothing, and the hint goes neutral — no session-state lookup needed.

Added `src/daemon/client/__tests__/daemon-client-timeout-route.test.ts`: spies on the real exec seam (`runCmdSync`) via `vi.mock` and drives actual socket/HTTP timeouts through the real `sendRequest` against hanging local servers. Four cases: (1) socket + declared `android` + pkill reports a termination (rebound-session simulation) → cleanup still runs, hint claims Apple on the termination evidence; (2) http + undeclared platform + pkill terminates nothing (unknown-session/common case) → cleanup still runs, hint stays neutral; (3) http + declared `ios` + terminates nothing → hint still claims Apple (affirmative-selector evidence); (4) http + remote (`baseUrl` set) → pkill never invoked, remote hint.

Proved red first: `git stash push -- src/daemon/client/daemon-client-timeout.ts` to revert just that file to the flawed 82aa5ade0 state, reran the new route tests — 2 of 4 failed exactly as you predicted (case 1: cleanup skipped on the declared `android` flag despite real evidence; case 2: undeclared-platform fail-safe still claimed Apple with zero evidence). `git stash pop` restored the fix; all 4 green.

PR body updated with the full eligibility rule, hint-evidence model, and matrix. `check:affected` unit/typecheck/lint/layering/build have been green across multiple local runs (807 files / 6439 tests); the only local failures were the tmpdir-leak gate flagging directories confirmed to belong to other concurrent sessions on this shared host (verified: flagged dirs no longer exist by inspection time, while other unrelated `agent-device-test-run-*` dirs sit in `/tmp` with timestamps spanning hours) — noted honestly in the PR body, with CI as the authoritative signal on the isolated runner.

Issue #1429, opened 2026-07-27
gate-of-gates: derived manifest proving every check is owned, wired, and reachable


thymikee:
Part of #1412 (Track E — gate integrity). From external review: the most dangerous failure mode is a perfectly designed gate that silently stops running. Nothing currently proves the gates themselves are wired.

## Verified holes this closes
- `CHECK_CATALOG` entries declare `ciJobs` as manual strings (scripts/check-affected/checks.ts); `assertCatalogComplete` covers only the check-ID universe — a renamed or deleted workflow job leaves a catalog entry pointing at nothing, silently.
- ci.yml `paths-ignore` excludes `website/**`, so a gate that conceptually owns docs paths never fires on docs-only PRs (see #1420).

## Deliverables
- A derived gate manifest test (unit lane): enumerate every vitest project (parse vitest.config.ts), every `test:*`/`check:*` package script, and every build lane; each must be owned by a PR workflow job, a scheduled workflow, or an explicit local-only declaration with a reason. Derived from the real sources (package.json, vitest config, .github/workflows/*.yml) — no hand-maintained second list.
- Model ownership as a transitive execution graph, not string presence: `workflow job → local/composite action → package script → aggregate script → vitest project or native test/build suite`. A lane owns a suite only when the graph proves the suite is reachable from the workflow job.
- Reuse the R8 workflow-parsing machinery in scripts/layering/ and resolve literal local-action references, package-script invocations, aggregate-script chains, and explicit vitest project selection. Unresolved dynamic `run` expressions, env-dependent dispatch, missing local actions/scripts, and opaque aggregate edges fail closed unless covered by an explicit owned waiver with a reason and tracking issue. Do not infer reachability from a command name merely appearing in workflow text.
- `CHECK_CATALOG.ciJobs` → parse workflow files and assert every named job exists. A job rename without a catalog update fails a PR.
- Path-category reachability: for each catalog path category, assert the owning workflow's trigger filters (`paths`/`paths-ignore`) do not exclude that category — the #1420/website hole, generalized.
- Keep the deterministic manifest/reachability gate offline and network-free. It must run on PRs and from a clean checkout without GitHub credentials.
- Branch-protection required-contexts audit is the only online part: run it as a scheduled, read-only GitHub API check. Required contexts must correspond to live job names. If token scope is an obstacle, land it inside the scheduled-lane health job (#1430) and say so; never make the deterministic PR gate depend on API availability.

## Acceptance
[...]

PR #1714, opened 2026-08-10
test(ci): canonical CheckId runner + derived gate manifest (#1429)

thymikee:
Closes #1429. A derived manifest proving every registered check is owned by a CI lane, wired on the paths that need it, and reachable when that lane runs.

Body current as of `5ee129a`.

## The one rule

> **Every shell block a qualifying lane reaches is either `pnpm gate <check-id>` or inventoried by digest.**

Nothing inspects what a command *does*. A step is allowed because of its **shape**; everything else is written down.

That is deliberate, and it was learned the hard way. A content-based classifier was defeated by `pnpm exec`, then by `pnpm exec --` and `npx --yes`, then by `node -e 'import("./scripts/layering/check.ts")'` — and after those, `eval`, heredocs, base64 and string concatenation are all still available. **"Does this text run project code?" is not answerable from text.** So the question is no longer asked: an unknown spelling does not need to be *recognised* to be *rejected*.

A step's digest is SHA-256 over `run` plus the keys that decide whether and how it executes — `shell`, `working-directory`, and `if:`.

## What the construction is

Five seams, each narrowed so the verifier does not have to police it.

**Local actions take a gate id, not a command.** `setup-apple-runner-build` used to take `build-command: pnpm gate swift-runner-ios`, so every caller's value was a command that had to be fingerprinted. It now takes `gate: swift-runner-ios` and its body is the constant `pnpm gate "$INPUT_GATE"`. The id is checked against `CHECK_CATALOG`, and the audit proves the body invokes the gate rather than trusting the declaration — swapping that step for `true` stops crediting callers.

**No local action interpolates `${{ inputs.* }}` into a `run:` body.** Values reach shell through `env: INPUT_*`, GitHub's own convention for the same job.

**Every input that reaches a shell body has a closed vocabulary.** `INPUT_*` passing through `env:` does not make a command-valued input inert — an action could bind `INPUT_COMMAND: ${{ inputs.command }}` and run the constant body `bash -c "$INPUT_COMMAND"`. Asking what a body *does* with a variable is the content analysis this design refuses; asking whether it *mentions* it is a substring scan. So the loader finds every dereferenced input and the constraint lands on the caller's **value**, which is a literal here. A command cannot be a member of a closed set of literals, so the attack is unrepresentable rather than policed. Nine inputs across three actions is the whole surface; `gate` is exempt because `GATE_ACTIONS` constrains it harder.

[...]

PR #1753, opened 2026-08-11
refactor(ci): make gate ownership structural (#1429)

thymikee:
## Summary

Part of #1429.

Make CI gate ownership structural and cheap:

- keep `CHECK_CATALOG` as the registered check universe;
- run owned checks through `uses: ./.github/actions/run-gate` with a literal `gate` input;
- derive ownership from workflow/composite-action YAML, never from `run:` shell text;
- prove every registered check has a qualifying owner, every affected path can reach its selected checks, and every test suite/Vitest project is registered.

This removes the shell execution classifier, condition lattice, waiver tables, action-body interpreter, and their tests. Raw `run:` steps can still exist, but they deliberately earn no manifest ownership credit.

The manifest catches the concrete defects behind this work: unowned registered checks, path-filter gaps, orphan suites/projects, unknown structural gate IDs, and unsupported reusable-workflow surfaces. `test:replay:android` is registered and its nightly owner runs the same package script instead of an inlined copy.

The PR is +1,865/−361 across 38 files (net +1,504). `scripts/gate` is 1,193 lines total, including tests. #1429 remains open for separately scoped work such as branch-protection context auditing; this PR does not claim to model shell execution.

## Validation

- `pnpm check:affected --run`
- `pnpm check:gate-manifest` — 48 checks wired across 33 lanes, 1 externally owned lane declared unprovable
- Fallow — no issues in 38 changed files
- planted regressions prove raw shell earns no credit, a typoed structural gate fails, composite ownership follows literal inputs through arbitrary nesting, cycles fail loudly, missing path ownership fails, and orphan suites/projects fail
- the canonical action regression executes its real Bash body with no optional arguments; the prior head failed on macOS Bash with `args[@]: unbound variable`, while the fixed head passes

No device validation is required: this changes CI wiring and its structural audit, not product or device behavior.


thymikee:
Reviewed exact head e26f7ae. This remains **not ready**.

- **P1 — ownership is inferred from dead text, not execution.** `GATE_INVOCATION` is a substring regex over `run:`, so `echo "pnpm gate fixture-cache"`, a comment, `false && pnpm gate fixture-cache`, or an unreachable shell branch credits the gate although it never runs. That directly violates #1429’s requirement not to infer reachability merely because a command name appears in workflow text.
- **P1 — conditions can false-credit gates.** Job-level `if:` is not modeled at all despite existing in the live mutation workflows. Inside composite actions, `guard[0] ?? nestedStep.condition` overwrites the nested step condition; an outer allowed condition such as `always()` can therefore hide an inner `if: false` and credit a gate that cannot execute.
- **P1 — this cannot close #1429 at the reduced scope currently described.** The issue requires the transitive execution graph, unresolved dynamic edges failing closed behind explicit owned waivers, job-rename/catalog semantics, and the scheduled branch-protection required-context audit. Those deliverables are deliberately omitted here. Please split or re-scope the residual work and keep it open before using `Closes #1429`; the current unprovable-owner declaration also tracks its unresolved device owner to the issue this PR would close.
- **P2 — evidence and size remain disproportionate.** Most regressions mutate the already-derived `Model` rather than removing/planting wiring through the shipped loader, and the body provides no quoted red-before result. At 1,198 implementation lines, 718 test lines, and +2,050 net—with 343-line `workflows.ts` and 303-line `audit.ts` already past project tripwires—the parallel YAML/shell/unit model is still large while missing basic GitHub semantics.

The PR body is candid about dropping bypass enforcement, but source comments in `run.ts` and `check-affected/model.ts` still falsely claim outside-runner project code fails the manifest. Most exact-head CI is green; iOS Smoke and kernel-error mutants were still pending at review time. No ready label.

thymikee:
All four addressed at `5a72c28`. Two of them were real bugs I introduced in the reduction, and P1c was right — I've stopped claiming `Closes`.

## P1 — dead text credited gates

Confirmed, and it is a live defect in this tree, not just a constructed one: `conformance-regenerate.yml`'s "Fail if regeneration changed anything" step names `pnpm gate maestro-regenerate` inside an `echo` telling a human to run it, and the substring scan credited the gate for that string.

Two of your four examples do **not** reproduce, for reasons worth recording: a YAML `#` comment never reaches the model (the parser strips it), and `echo "pnpm gate x"` is blocked by the quote, since the regex required a shell separator before `pnpm`. The two that do reproduce — `false && pnpm gate x` and an unreachable branch — are the ones that matter, and I found four more myself: `echo pnpm gate x` unquoted, a shell `#` comment, a heredoc body, and a non-first segment (`echo hi; pnpm gate x`).

Fixed by crediting on execution shape: a gate counts only as the **first command segment of a line**, and a body carrying shell structure (`&&`/`||`, `if`/`case`/loops, heredocs, `eval`, `exec`) earns nothing at all. Reachability inside a script is not decidable so I do not attempt it — unrecognised shape means no credit, and the check reports unowned until someone wires it as a plain step.

One form is read deliberately: `modules=$(pnpm --silent gate mutation-affected --list-affected | tail -n1)`. The assignment is unambiguous and the gate runs; without it `mutation-affected` would report unowned while genuinely executing.

Eight dead-text rows and eight live-form rows are checked in as a table.

## P1 — conditions

Both real, both mine.

**Job-level `if:` was not modelled at all.** Six live jobs carry one. Now threaded through every sighting a job produces, including gates reached via a composite action. Two conditions on the mutation lanes needed rulings and are declared with reasons: `needs.select.outputs.modules != '[]'` and `always() && needs.select.result == 'success'`.

**`guard[0] ?? nestedStep.condition`** — exactly as you describe. `LaneStep` now carries every guard as a list, so an outer `always()` cannot erase an inner `if: false`. Regression checked in.

## P1 — scope

You are right, and I have changed `Closes #1429` to `Part of #1429` with the residual work enumerated in the body. Reading the issue again, the substring scan violated a literal acceptance criterion — "Do not infer reachability from a command name merely appearing in workflow text" — so P1a was not a nitpick but a named requirement.

Still missing and now listed rather than implied: the scheduled branch-protection required-contexts audit; job-rename detection (deriving the job list means a rename cannot leave a stale claim, but it also cannot *fail*, which is what the acceptance asks for); and the per-edge waiver format for unresolved dynamic edges. `replay-ios-device`'s unprovable owner tracking to #1429 is a further reason not to close it here — thank you for catching that.

## P2 — evidence and size

Fair on both counts, partially addressed.

The new regressions plant real YAML through the shipped loader rather than mutating the derived `Model`; `audit-coverage.test.ts` still mutates the model, because its cases are about lane deletion and path filters where planting a whole tree would obscure what is being tested. Red-before results are now quoted in the body per witness.

Size went **up**, not down: 1,916 → 2,090 lines, +2,050 → +2,224 net. Fixing the crediting cost ~40 lines of implementation and ~90 of table. I am not going to claim that is comfortable — it is still over the 1,500 ceiling, and I would rather say so than trim the table to get under it.

## Also fixed

The stale comments in `run.ts` and `check-affected/model.ts` claiming outside-runner project code fails the manifest. It does not, and they now say what actually happens: no credit, so the check reports unowned.

## CI

Green at `e26f7ae` except iOS `Smoke Tests`, whose log tail shows only runner cleanup — the established base-branch flake signature. `5a72c28` re-runs it. The two `src/daemon` unit failures I flagged in the body reproduce on clean `main` at `057ab1c` and are not from this branch.

---

thymikee:
Reviewed exact head 5a72c280. The prior job-level and nested-action condition findings are fixed: guards are accumulated instead of overwritten. The PR now honestly presents itself as only part of #1429, leaves the residual issue work open, and corrects the stale outside-runner source claims. All authoritative CI is green.

This is still **not ready**:

- **P1 — line-by-line shell parsing still credits dead text.** `gateIdsIn()` splits the body by newline before classifying each line, so it loses multiline shell context. For example, `message="\n pnpm gate layering\n"` followed by `echo "$message"`, or an uncalled `f() {` body containing a standalone `pnpm gate layering` line, has no token matched by `SHELL_STRUCTURE`; the standalone line is accepted as `GATE_SEGMENT` although the gate never executes. The new regressions call the helper directly and cover selected single-line/control forms, not these cross-line cases through the shipped loader/audit. This is the repeated shell-compiler reconstruction pattern the project guidance says to redesign after exceptions recur. Make a YAML/action construction structurally name one gate instead of deriving execution from shell text.
- **P2 — the implementation remains disproportionate.** The gate subsystem is now 2,090 LOC and the PR +2,224 net; `workflows.ts` alone is 411 LOC and now contains a second shell parser, while the stated review ceiling was 1,500. The body also says any `&&`/`||` chain earns no credit, but the implementation credits a gate in the first segment (for example `pnpm gate x || true`), so the execution/failure-propagation claim is inaccurate.

No ready label.

thymikee:
Both confirmed at `5a72c28`. P2's factual point is a straight error in my body text and I am correcting it now; P1 I am **not** patching, because your structural recommendation is right and a fourth epicycle on the same class is the wrong answer.

## P1 reproduces, in every form you name

```
CREDITS ["layering"]  msg="\npnpm gate layering\n"          (multiline assignment)
CREDITS ["layering"]  f() {\npnpm gate layering\n}          (uncalled function)
CREDITS ["layering"]  (\npnpm gate layering\n)              (subshell block)
```

Worth recording how close I came to reporting the function case as not reproducing: my first probe ended in `echo done`, and `done` is a `SHELL_STRUCTURE` keyword, so the whole body was rejected for the wrong reason and printed "no credit". Re-running without that word credits the gate. The accident is itself evidence for your point — a keyword denylist over line-split text is not a model of shell, and I cannot reliably reason about it even when I wrote it.

## P2 — the body claim is wrong

`SHELL_STRUCTURE` lists keywords, not operators, and crediting takes the first segment of a line. So `pnpm gate x || true` and `pnpm gate x && echo ok` both credit `x`, while the body says any `&&`/`||` chain earns nothing. That sentence is inaccurate and I have removed it.

It also hides a real distinction you are pointing at: `pnpm gate x || true` **runs** the gate but swallows its exit code, so the gate cannot fail the lane. "Runs" and "gates" are not the same claim, and the current model only ever answers the first.

## Not patching this again

Three rounds of #1714 (`pnpm exec`, `pnpm exec --`/`npx --yes`, `node -e 'import(…)'`), then rounds 5–12 on input flow, now rounds 13–14 on crediting. Each fix was correct and each exposed the next hole one level out. Adding `{`, `(` and multiline-assignment detection to a denylist is round 15, and it would still not model quoting, `$'...'`, arrays, or `set -f`.

Your recommendation is the convergent one, and it is already in the tree as a working precedent: `setup-apple-runner-build` takes `gate: swift-runner-ios` as an **input**, and `GATE_ACTIONS` + `gateActionBodies` credit it from YAML structure with no shell reading at all. Generalising that — every gate step becoming a `uses:` of a `gate` action with an `id:` input — makes the whole class unrepresentable rather than policed: a `run:` body is never read for crediting, so no amount of dead text in one can credit anything.

It should also answer P2 rather than worsen it. `gateIdsIn`, `GATE_SEGMENT`, `GATE_CAPTURE`, `SHELL_STRUCTURE` and their tables all delete; `workflows.ts` loses its second shell parser. And it gives `|| true` a structural answer — a gate reached through the action is a step whose exit code the lane sees, so "runs" and "gates" stop diverging.

I am putting the rewiring to the repo owner before starting it, because it changes how every lane invokes a gate — roughly fifty steps across eighteen workflow and action files — and that is live CI behaviour on lanes I cannot run here. Two wrinkles I would want to settle in review rather than guess at: `modules=$(pnpm --silent gate mutation-affected --list-affected | tail -n1)` needs the action to expose an output, and the handful of gates taking arguments (`--base "$FALLOW_BASE"`) need an `args:` input, which is shell again — though only for arguments, never for **which** gate runs, which is the part that has to be structural.

Holding here rather than pushing another patch. CI is green at this head apart from the known iOS Smoke flake.

---

thymikee:
Diagnosed the two mutation failures at `690bcf7`. **Not a score regression, and not from this branch** — but it is a real finding, so recording it here rather than papering over it.

## Cause

Reproduced locally with `pnpm gate mutation --modules kernel-errors`:

```
INFO  Instrumenter  Instrumented 1 source file(s) with 185 mutant(s)
INFO  DryRunExecutor  Starting initial test run (vitest, "perTest" coverage analysis)
ERROR DryRunExecutor  One or more tests failed in the initial test run:
        daemon entrypoint publishes HTTP metadata and cleans up on shutdown
                Test timed out in 5000ms.
ERROR Stryker  There were failed tests in the initial test run.
ConfigError: There were failed tests in the initial test run.
mutation: Stryker produced no report at .tmp/mutation/mutation.json (exit 1)
Lane envelope: fail at stage stryker in 337s
```

Stryker runs the related suite once before mutating anything and aborts if any test fails, so no report is produced — which is why the CI shard uploaded a 620-byte artifact instead of a report, and why `Affected decision-kernel mutants` failed downstream. (GitHub's log API only returned the cleanup tail for both jobs, so the local reproduction is the evidence, not the CI log.)

## It is not this branch

`src/__tests__/daemon-entrypoint.test.ts` passes on its own at **1084 ms**, comfortably inside the 5 s limit:

```
✓ daemon entrypoint publishes HTTP metadata and cleans up on shutdown  1084ms
  Tests  5 passed (5)
```

It only exceeds the timeout inside the dry run, where Stryker starts 4 concurrent runner processes over the 667 related test files this module pulls in.

This branch touches **no `src/` file, no `vitest.config.ts`, and no `stryker.config.json`** — the entire diff is `scripts/`, `.github/`, `docs/`, `package.json` and one `test/` file. It cannot affect that test's timing.

## Why it showed up here

`kernel-errors` is `LANE_CANARY`, and `scripts/mutation/**` fails open into the full sweep, so editing `scripts/mutation/workflow.test.ts` (needed when the workflow command text moved) makes this branch run mutants that ordinary PRs never run. The other branches passing `mutation-affected` today are not evidence the shard is healthy — they do not touch `scripts/mutation/**`, so they select nothing.

That makes this the same species as the three orphan checks #1770 just fixed: **a lane that almost never runs, and does not work when it does.** Worth its own issue — the dry run is one slow test away from failing regardless of the mutants, and the failure mode is an envelope with no report rather than a legible message.

## Not fixing it here

The minimal fix is a `testTimeout` bump for that one daemon test, which is a file this PR otherwise does not touch, in an area outside its scope — and this PR is already under instruction to get materially smaller, not wider. Happy to do it as a separate one-liner if you would rather it land now.

#1753 therefore stays red for a pre-existing reason, which is consistent with its body: it is holding at the structural-rewrite decision and is not a merge candidate either way.

## Also in this push

`8a091e3` merges main after #1770. That produced duplication rather than a conflict — git merged cleanly because the edits sat in different parts of `ci.yml`, leaving two steps per check, each running twice. Kept your reviewed placement and comments from #1770 and changed only the `run:` line to `pnpm gate <id>`; dropped this branch's duplicates. Same three checks, same three lanes, once each. Manifest green: 47 checks across 33 lanes.

---

thymikee:
Reviewed exact head `8a091e30`. This remains **not ready**.

**P1 — the shell-derived ownership flaw is unchanged.** The new `run-gate` action is a sound seam in isolation, but it has zero callers and is absent from `GATE_ACTIONS`. `workflows.ts` still calls `gateIdsIn(run)`, and roughly 50 workflow/action steps still invoke `pnpm gate` through raw shell. Consequently multiline assignments, uncalled function/subshell bodies, and `pnpm gate x || true` can still false-credit a gate. The checked-in regressions continue to exercise the shell parser rather than proving YAML-only structural ownership. Wire every credited gate through the structural action/input and delete the shell-execution inference before claiming this class fixed.

**P2 — the parallel model remains disproportionate.** `scripts/gate` is still 2,090 LOC (including 411-line `workflows.ts`), and the exact PR is +2,232 net, now with an additional unused 32-line action. The rewiring has not reduced or replaced the shell/YAML/unit model.

CI also has two deterministic reds with one root cause: `Mutants (kernel-errors)` cannot resolve `@agent-device/selectors/engine` inside the Stryker sandbox; the affected-mutant ratchet then fails because no shard report exists. This came through the merged base rather than the gate-action delta, but it is not a rerun-shaped flake—fix the base/Stryker workspace alias, rebase, and get both jobs green.

Finally, update the materially stale body: it says the branch is held at `5a72c28`, says the structural action is not present, reports the old size/net, and presents the three orphan checks as current main defects even though #1770 already fixed them and this branch retained those landed steps. No readiness label applied.

thymikee:
All six addressed at `43cb6db`. Both bugs were real, and the first one found a second orphan you didn't name.

## [bug] `registered` was shape-only

Correct, and worse than reported. Making a `test:*` script a suite **by name** immediately surfaced two:

```
- package script "test:replay:android" runs script:test:replay:android, which no registered check covers.
- package script "test:integration:progress" runs script:test:integration:progress, which no registered check covers.
```

`test:replay:android` is yours. Registered it, and pointed the nightly at the script rather than re-listing the six `.ad` files, so the two can no longer drift.

One correction to the suggested wiring: `pnpm gate replay-android` there earns **no credit anyway**, because the nightly invokes it inside `reactivecircus/android-emulator-runner`'s `script:` input — shell handed to a third-party action this loader does not read. The suite executes; the manifest cannot see it. Recorded in `UNPROVABLE_OWNERS` with that reason rather than pretending it is proven. That is the emulator-routing item #1429 already tracks.

`test:integration:progress` is the second, and it's a false positive of the name rule — it prints a table and exits 0, and its `--check` sibling is the registered `integration-progress` gate. Declared in `REPORTING_SCRIPTS`, with an inert check so the declaration can't be a free silencer.

## [bug] `freerange` fail-open

Correct. `gate('freerange', …, false)`. Fail-open would have put a Bun binary on the pre-push path for every `scripts/**` edit.

## [suggestion] Vacuous skip-list

Correct. Now derived from `CHECK_CATALOG` (`!localRunnable`) rather than hand-listed, so a repointed check can't leave a name that cannot fail.

## [suggestion] `matchesGlob`

Correct. `**` matches zero directories, so joining splits with `.*` left the adjacent slash mandatory and `src/**/*.test.ts` missed `src/a.test.ts`. Rewritten and pinned against `packages/*/src/**/*.test.ts` plus the zero-directory and negative cases.

## [suggestion] The unwired action

Deleted. It had no callers, wasn't in `GATE_ACTIONS`, and its comment described a system that hadn't shipped — you're right that this was a dead seam plus a false claim. It comes back with the rewiring, not before.

## [suggestion] Comment archaeology

Collapsed the headers in `audit.ts` and `workflows.ts` to the invariant.

## Mutation — I fixed a real cause, but not the one you name

I reproduced a failure locally at exactly `8a091e30b264` and it was **not** a resolution error:

```
ERROR DryRunExecutor  One or more tests failed in the initial test run:
        daemon entrypoint publishes HTTP metadata and cleans up on shutdown
                Test timed out in 5000ms.
```

That test is the only one in its file spawning a real daemon process — 1084 ms standalone, over 5 s under Stryker's four concurrent runners. It aborts the sweep before a single mutant runs, which is why the shard uploads an empty envelope. Given 30 s.

I could **not** reproduce `@agent-device/selectors/engine` failing to resolve. Its exports map points at TypeScript source, so no build is involved, and locally it resolves. GitHub's log API returns only the cleanup tail for these jobs, so I can't read your symptom directly and I'm not going to claim I've fixed something I never saw.

So this push discriminates: if mutation goes green, the timeout was the whole story. If it stays red on resolution, that's confirmed as separate and base-owned, and I'll fix the alias next with the CI log to work from. I'd rather find out than guess between two causes.

## Body

Rewritten — it was materially stale, as you say. It no longer claims to be held at `5a72c28`, no longer describes the structural action as present, carries the real size (2,201 in `scripts/gate`, +2,562/−330), and states plainly that #1770 fixed the three orphan checks and this branch retains those landed steps rather than presenting them as live defects.

## Still true

The shell-derived crediting is unchanged and still the core objection. The conversion is the plan; it is ~60 steps across 18 files with three that change CI control flow, and I've put it to the repo owner rather than starting it mid-review round.

---

thymikee:
Superseding the earlier shell-parser findings: exact head e3cb6f33d implements the requested structural redesign. Gate ownership now comes only from uses: ./.github/actions/run-gate with a literal gate input; raw run: text is intentionally invisible. The shell classifier, condition lattice, action-body interpreter, and their exception tests are gone. The final shape is +1,840/-357 across 38 files (net +1,483), with scripts/gate at 1,193 lines including tests. pnpm check:affected --run passed before push; the manifest reports 48 checks across 33 lanes with one explicitly external Android replay owner. Code review is clean and this is ready for human merge.

thymikee:
Exact rerun found an attributable deterministic bug at `e3cb6f33`: `Swift Runner Unit Compile` fails inside the new `run-gate` composite action before invoking pnpm. With empty `INPUT_ARGS`, Bash 3.2 + `set -u` treats `${args[@]}` as unbound even after `args=()`: `action.yml` line `pnpm gate "$INPUT_GATE" "${args[@]}" | tee ...` exits `args[@]: unbound variable`. This is the structural-action rewrite, not a Swift/toolchain flake; another rerun will not help. Please make the empty-args invocation explicit/safe in the action and add an action-level regression for an omitted/empty `args` input before repushing. I am not importing this cross-unit workflow fix into #1758.

thymikee:
Fixed the exact-head Apple CI blocker in d946e2bdc. The shared run-gate action used an empty Bash array under `set -u`; macOS Bash treated `${args[@]}` as an unbound variable, so Swift compile and both Apple smoke lanes failed before any build began. The action now constructs one always-nonempty command array and cleans its temporary output file. A regression executes the real composite-action Bash body with empty optional args; it fails on the prior shape and passes now. Full `pnpm check:affected --run` passed, including clean-install package verification and all 28 gate-manifest tests. `ready-for-human` is restored; fresh GitHub checks are running.

thymikee:
Exact corrected-head CI classification at `d946e2bdc`: the empty-args Apple blocker is fixed (`Swift Runner Unit Compile` and macOS Smoke now pass; iOS is running), but the reviewed alias correction is still ineffective in the real Stryker sandbox. `Mutants (kernel-errors)` reaches the initial dry run and fails with `Cannot find package @agent-device/selectors/engine imported from .../.tmp/stryker/sandbox-*/src/core/selector-pipeline.ts`, then produces no report; the aggregate fails consequently. This is the same concrete shared blocker preventing #1758 from consuming the fix. Please reproduce the actual sandbox resolution route and plant the regression through that route before the next head; a rerun will not change this failure.

thymikee:
Fixed the mutation dry-run blocker at exact head `c1eb50895`. The workspace export catalog already contained `@agent-device/selectors/engine`; the defect was Vite alias precedence: the string alias for `@agent-device/selectors` prefix-matched `/engine` first and produced an invalid `index.ts/engine` path inside Stryker's sandbox.

The mutation config now derives exact-match regex aliases for every exported workspace specifier, with a regression proving each subpath resolves only to its own export and unexported subpaths remain closed. Planted red: weakening exact matching makes the new test fail on `@agent-device/selectors/ast` with both the root and subpath targets. Green evidence: 46/46 mutation tooling tests; focused real Stryker copied-sandbox dry run completed 11/11 tests; full `pnpm check:affected --run` passed before push. Exact-head Actions/mutation jobs are queued.

Issue #1752, opened 2026-08-11
refactor(snapshot): clean ownership before an earned package extraction


thymikee:
# Clean snapshot ownership before an earned package extraction

Related to #1739 Wave 4 as behavior-neutral snapshot-domain prework. This issue does not migrate the snapshot command and does not authorize a new workspace package.

## Decision

Clean up the ownership problems that are already demonstrated, then run the package/ownership checkpoint before #1739 Wave 4's snapshot unit is scoped.

Do **not** extract `@agent-device/snapshot-semantics` as part of this issue. The current proposal would turn a healthy, focused root zone into a broad namespace package whose facade mirrors unrelated callers. Packaging is justified when more than one implementation or a concrete `packages/platform-*` consumer needs the same pure operation—not by root fan-in alone.

After the two cleanup PRs land, the scheduled checkpoint must decide the owners of the root-only helpers Wave 4 needs. Its output is a required input to that unit:

- no cohesive package boundary: record the contracts/kernel/platform ownership moves and close the extraction track; or
- cohesive boundary with concrete Wave 4 callers: open a focused extraction issue for the smallest co-occurring operation set.

## Evidence

The current `src/snapshot` zone has 12 production files, 1,786 lines, and no implementation file above 422 lines. Its size and topology are not themselves architecture debt.

The main deliverable is dependency reduction. The repository dependency graph currently reports 35 distinct external non-test consumers of `src/snapshot`. After PRs 1 and 2 relocate the non-cohesive quality, processing, tree, disclosure, Android-overlay, and label-signal responsibilities, the remaining cohesive cluster is expected to have 15 consumers: a reduction of 20 consumers, or about 57%. The checkpoint must recompute and attach these figures from the landed tree.

The relevant cross-boundary evidence is narrower:

- `src/platforms/android/snapshot.ts` imports `deriveMobileSnapshotHiddenContentHints` from root snapshot code;
- Android recovery code imports `AndroidContentRecoveryReason` from generic snapshot quality code;
- `src/platforms/apple/interactor.ts` imports `readSnapshotQualityVerdict` from root snapshot code;
- current `packages/platform-*` and provider packages import no root snapshot implementation **today**;
- #1739 Wave 4 schedules `src/daemon/snapshot-presentation/ios/` to move behind `platform-apple` and explicitly requires ownership decisions for its root-only overlay, scroll-indicator, and tree helpers first;
[...]

PR #1754, opened 2026-08-11
refactor(snapshot): clean snapshot ownership

thymikee:
## Summary

Part of #1752.

Clean snapshot ownership before the planned Wave 4 package checkpoint:

- move Android content-recovery taxonomy into contracts/platform and split snapshot-quality parsing/warnings
- move root-only lookup/tree/disclosure ownership into core, use contracts ancestry/text semantics, and delete duplicate helpers
- restore shared contracts adapters for ancestor traversal, text extraction/meaningful-label semantics, and Android IME node ownership
- preserve capture ordering and add characterization coverage for Android recovery, capture, tree, lookup, text, disclosure, ancestor, and IME paths

This is behavior-neutral structural prework; it does not create a snapshot package or change command behavior. The post-landing ownership checkpoint remains tracked by #1752 and must run before Wave 4 is scoped.

Post-cleanup dependency graph: external production consumers reduced from 35 → 15; `src/platforms/android/snapshot.ts` is the sole remaining platform-to-snapshot import.

Scope: 66 files touched across contracts, core, daemon, platform integration, snapshot presentation, tests, and layering metadata; scope remains the snapshot ownership cleanup described by #1752, with no unrelated command-surface expansion.

## Validation

- `pnpm check` passes on exact head: all tooling/package checks, 827 unit/subprocess test files with 6,512 tests, the temp-directory leak check, and 38 smoke tests passed; 3 live-device smoke lanes were skipped because their opt-in environments were not configured.
- `pnpm check:affected --run` passes after rebasing onto current `main`.
- Full affected validation passed: 604 test files / 5,034 tests, changed-line coverage 103/104 (99.04%), layering, replay compatibility, and daemon-wire compatibility.
- Focused snapshot-quality, Android recovery/IME, capture, tree, lookup, text, disclosure, and overlay tests pass (72 tests).
- Required R3 red-before proof: temporarily adding a static value import from `src/commands/interaction/runtime/interactions.ts` to `src/platforms/android/snapshot.ts` made the real `pnpm check:layering` fail with `[R3 platforms-seam]` at the planted import; removing it restored the gate.
- Snapshot-quality DAG red-before proof: a temporary `src/snapshot-quality/verdict.ts → src/core/snapshot-node-lookup.ts` import made the real `pnpm check:layering` fail R5 with a `snapshot-quality → core` back-edge; the import was removed.
- Moved-test discovery proof: a temporary type error in relocated `src/snapshot-quality/__tests__/warnings.test.ts` made `pnpm typecheck` fail with TS2322/TS6133; the error was removed.
- Docs/skills unchanged: the issue scopes no command behavior, CLI, or skills changes.


thymikee:
Reviewed exact head `ec6b5e6e`. The behavior-preserving splits and ownership moves are sound: capture ordering is preserved, the new modules and test topology stay small and focused, there is no compatibility barrel, the Fallow change is narrowly limited to retiring the deleted finding, all authoritative CI and smoke checks are green, and package size is net negative.

Before readiness, please complete the P2 validation/body evidence required by the linked issue and repository rules: provide red-before proof for the new `snapshot-quality` DAG rank by planting a forbidden back-edge and showing the real layering gate fail; provide moved-test discovery proof (a planted type error or discovered-count delta); and update the PR body with the post-cleanup dependency graph (`35 → 15` external production consumers, with `src/platforms/android/snapshot.ts` as the sole remaining platform-to-snapshot import) plus touched-file and scope accounting. No production behavior finding.

thymikee:
## Code quality review

The direction is right and the deletions are real (−692/+706 with five modules genuinely gone, `resolution.ts` dropping 1024 → 999). But two of the "inline the delegation / reuse the canonical helper" moves **deleted adapters that were doing real work and pasted their bodies into the call sites**. That's complexity rearranged, not deleted — and in both cases the call sites got worse.

---

### 1. `findNearestAncestor` was already the canonical adapter. Deleting it duplicated it three times.

`main`'s `src/snapshot/snapshot-processing.ts` had:

```ts
/** Returns the nearest ancestor matching `predicate`; false means keep walking. */
export function findNearestAncestor(nodes, node, predicate) {
  const nodesByIndex = buildSnapshotNodeMap(nodes);
  return findSnapshotAncestor(nodes, node, nodesByIndex, (parent) => (predicate(parent) ? parent : null));
}
```

It was *already* a four-line adapter over the contracts tree walk. #1752's "generic ancestor traversal to the existing contracts tree walk" was satisfied before this PR started.

This PR deletes it and inlines its body at three sites — `src/core/interaction-targeting.ts:112`, `src/daemon/screenshot-overlay.ts:232` and `:238`:

```ts
findSnapshotAncestor(nodes, node, byIndex, (parent) => (parent.hittable === true ? parent : null))
findSnapshotAncestor(nodes, node, byIndex, (parent) => (isOverlayActionableNode(parent) ? parent : null))
```

Every caller now has to build the index map itself and hand-convert a predicate into a `T | null` mapper. Net effect: `buildSnapshotNodeMap` leaks as a new concept into `core/` and `daemon/`, the `? parent : null` boilerplate is copied three times, and nothing was simplified. All three call sites want a **predicate**; none of them want the mapper form.

Keep the adapter — and move it to `packages/contracts/src/snapshot-tree.ts` next to `findSnapshotAncestor`, since that's where the vocabulary already lives. All three sites go back to one line and stop knowing about index maps.

### 2. Inlining the IME delegation duplicated a shape adapter and put a lambda in the capture pipeline.

`isAndroidInputMethodSnapshotNode` was not a pointless 12-line delegation — it adapted `SnapshotNode → { packageName, resourceId }`. After inlining, both call sites repeat that mapping (`src/core/snapshot-chrome.ts:244`, `src/daemon/handlers/snapshot-capture.ts:180`), and the capture site regressed from a clean function reference:

```ts
isAdditionalOverlayNode: data?.backend === 'android' ? isAndroidInputMethodSnapshotNode : undefined,
```

to a six-line inline arrow nested inside the `annotateCoveredSnapshotNodes(...)` argument — in the exact expression #1752 says must stay legible and unreordered.

Contracts already owns the predicate now. Give it the node-shaped form too:

```ts
export function isAndroidInputMethodNode(node: Pick<RawSnapshotNode, 'bundleId' | 'identifier'>): boolean
```

Both call sites become one-liners, the capture pipeline keeps its function reference, and the root module still disappears. The adapter belongs in the layer that owns the vocabulary, not copy-pasted into two consumers.

### 3. Three fresh copies of `extractNodeText` in the two brand-new files.

The PR correctly deletes `displayNodeLabel` as a proven duplicate of the canonical `extractNodeText` — then ships three verbatim copies of the same reduction in code it just created:

- `src/core/snapshot-node-lookup.ts:19` (`resolveRefLabel`)
- `src/core/snapshot-node-lookup.ts:43` (`findNearestMeaningfulLabel`)
- `src/core/snapshot-tree-ingestion.ts:50` (`pruneGroupNodes`)

All three are `[node.label, node.value, node.identifier].map(trim).find(nonEmpty)` — exactly `extractNodeText`. They came over from the deleted file, but these are new modules with new tests, so the duplication is being re-blessed rather than retired.

The blocker is a needlessly narrow parameter: `extractNodeText` takes `SnapshotState['nodes'][number]`, i.e. `SnapshotNode`, which requires `ref` — so `pruneGroupNodes` can't pass a `RawSnapshotNode`. It only reads three fields. Widen it:

```ts
export function extractNodeText(node: Pick<RawSnapshotNode, 'label' | 'value' | 'identifier'>): string
```

One line in contracts, and all three copies collapse to `extractNodeText(node)`.

### 4. The two new "clean ownership" modules are coupled on day one.

`src/core/snapshot-tree-ingestion.ts:3` imports `isMeaningfulLabel` from `snapshot-node-lookup.ts`, which exports it *only* for that. The shared primitive ended up inside one of its two consumers, so lookup and ingestion aren't independently ownable.

`isMeaningfulLabel` is label vocabulary — the same family as `extractNodeText`. Move it to `packages/contracts/src/snapshot-text.ts` and the lookup↔ingestion edge disappears. Pairs naturally with #3: do both and neither new module owns label vocabulary at all.

### 5. Dead branch carried into a new file with new tests.

`src/core/snapshot-node-lookup.ts:20-24`:

```ts
if (primary && isMeaningfulLabel(primary)) return primary;
const fallback = findNearestMeaningfulLabel(node, nodes);
return fallback ?? (primary && isMeaningfulLabel(primary) ? primary : undefined);
```

The final ternary's condition is provably false at that point — the guard two lines up already returned when it was true. The `??` alternative can only ever evaluate to `undefined`. It reduces to:

```ts
return findNearestMeaningfulLabel(node, nodes);
```

Pre-existing, but this PR is the moment: it's a new file getting fresh characterization tests, and those tests will now pin a branch that can't execute.

### 6. Minor: the error classifier got more convoluted in the move.

`main`:

```ts
const details = (error as { details?: Record<string, unknown> }).details;
const reason = details?.androidSnapshotHelperFailureReason;
```

`packages/contracts/src/android-snapshot-quality.ts:31-35` now uses two casts plus an extra `typeof details !== 'object'` branch for the same result. The single-cast form was fine — the move should have carried it verbatim.

---

### Credit where due

- `extractNodeReadText` was a pure identity wrapper (`return extractReadableText(node)`); deleting it and calling through directly is exactly right.
- Android recovery taxonomy genuinely single-sourced in contracts, with the platform producer deriving rather than duplicating — this closes the divergence hazard #1752 was built around.
- Capture ordering preserved verbatim with characterization coverage added.
- `snapshot-quality` rank added alongside the existing `snapshot` zone rather than replacing it, so layering stays green mid-cleanup.
- No file pushed near 1k; `resolution.ts` actually drops below it.

### Verdict

Items 1 and 2 are structural regressions and should be fixed before merge — both make the code more tangled than `main` while claiming to simplify it, and both have a two-line fix in contracts. Items 3–5 are cheap and belong in the same PR, since they're all the same theme the PR is already executing: put the vocabulary in the layer that owns it and stop keeping private copies.

thymikee:
Reviewed exact head `2e842d145`. The code delta is clean and behavior-neutral: the contracts-owned predicate/text/IME adapters and ownership moves remove the reviewed duplication while preserving the capture pipeline ordering, and the focused adapter tests are meaningful. All exact-head checks are green; no additional live-device evidence is needed for this structural refactor.

Two issue-acceptance items still block readiness:

1. Change `Closes #1752` to `Related to #1752` or `Part of #1752`. The issue explicitly requires an ownership checkpoint after PRs 1 and 2 land; merging this PR must not auto-close the issue before that post-landing decision is recorded.
2. Attach the exact acceptance evidence #1752 asks for: plant a command → Android-platform static value import and show R3 fail, then revert it; and record a passing broad `pnpm check`. The current body documents an R5 snapshot-quality back-edge and `pnpm check:affected --run`, which are useful but different proofs.

No code finding remains and no readiness label was applied.

thymikee:
Finished the remaining readiness work on exact head 2e842d145. The PR now keeps #1752 open for its required post-landing checkpoint, records the planted R3 commands-to-Android import failure, and records a fully green broad deterministic gate (827 test files / 6,512 tests plus 38 smoke tests; 3 opt-in live lanes skipped). The branch is CLEAN/MERGEABLE with all GitHub checks green and is labeled ready-for-human.

PR #1758, opened 2026-08-11
refactor: route install commands through platform runtime

thymikee:
## Summary

Route the ADR 0019 install/deployment unit—`install`, `reinstall`, internal `install_source` (the target of public `install-from-source`), and `push`—through exact runtime facts and package/provider-owned operations.

Facts are the only support authority. Each handler inspects once and binds once: install/reinstall use `deployApp`; source installation uses readiness, materialization, and deployment; push uses readiness and notification delivery. Unsupported provider/platform cells fail closed without legacy capability or local fallback. Platform packages own sequencing and native command construction; root composition supplies focused lazy host ports for shared source materialization and fuzzy app resolution.

The accepted stack allocation at this head is R18 contracts, R19 selector, R20 boot, R21 apps, R22 appstate, R23 shutdown, and R24-R27 install/reinstall/install_source/push. The derived rule catalog rejects collisions.

Review repairs include end-to-end request cancellation for Apple/Android/Harmony/WebDriver work, request-owned draining for opaque Limrun mutations, extracted WebDriver transport and Android deployment mechanics, whole-operation reinstall cache invalidation, descriptor-derived capabilities projection, stale-provider fail-closed coverage, and ref-frame expiry immediately before the sole source/push dispatch.

Scope: 169 files, +8,539/-3,382 against #1757 (`ab459da0a36d`). This is an exceptional but cohesive cross-platform migration: it removes legacy capability/dispatch paths and moves mechanics into the family/provider packages rather than adding a second execution route.

## Validation

- Exact head: `7fd538f0e211`; base: #1757 `ab459da0a36d`.
- Every exact-head GitHub check is green: static/type/layering/Fallow/coverage/integration/package/released-surface, CodeQL, FreeRange, all nine affected mutation shards, Swift compile/releases, and Android/iOS/Linux/macOS/Web smoke lanes.
- The pre-push `pnpm check:affected --run` passed from a clean committed tree.
- Historical live iOS simulator evidence covers install, reinstall, public install-from-source/internal `install_source`, and push with cleanup. Per user direction, this final rebase did not run local device or simulator tests; Android/Harmony/WebDriver/Limrun and physical-device cells remain supported by production-composition, cancellation, cache, and fail-closed fact tests plus GitHub platform lanes.
- No command behavior docs or skills changed because the public CLI surface is unchanged.

PR #1756, opened 2026-08-11
refactor: route apps through request runtime

thymikee:
## Summary

Routes the canonical descriptor `apps` command through the request-scoped platform runtime as the provisional sibling of the other ADR 0019 fan-out units. Refs #1739.

- Declares the complete request-scoped use explicitly as `ensureReady + listApps`; readiness is a separate operation before app inventory.
- Performs one side-effect-free `inspectFacts` admission and one `bind` per apps/doctor runtime request, with provider-first fail-closed ownership.
- Keeps metadata eager and implementation lazy, preserves selector/session/filter and response semantics, and adds no durable-tier machinery.
- Projects capability support from runtime facts only. Unsupported Linux, Web, watchOS, XCTest physical, and WebDriver app cells remain unavailable rather than reconstructed from the retired capability matrix.
- Retires the apps capability bucket/legacy adapter and matching branches/imports/helpers, removes the stale Apple apps admission closure, and adds the apps row to the existing parametrized cutover table. No fifth policy file was added.
- Keeps generic `ensureReady` distinct from `bootTarget`/`bootTargetHeadless`. macOS generic readiness is an explicit no-op fact and macOS app inventory remains available; boot operations remain separate and unavailable.
- Keeps the HarmonyOS doctor target-app cell as the legacy informational unsupported result, while the canonical `apps` command supports HarmonyOS through its request-scoped facts and lazy HDC inventory host.

Facts cover the current platform-leaf/device-kind/provider denominator:

| Cell | ensureReady | listApps | Result |
| --- | --- | --- | --- |
| Apple iOS/iPadOS/tvOS/visionOS simulator | available | available | supported |
| Apple physical iOS CoreDevice | available | available | supported |
| Apple physical iOS XCTest backend | available | unavailable | fail closed: unsupported-device-backend |
| Apple macOS host | available no-op | available | supported; boot operations remain unavailable |
| Apple watchOS | unavailable | unavailable | fail closed |
| Android emulator | available | available | supported; headless boot is separate |
| Android physical/device | available | available | supported; headless boot is unavailable |
| HarmonyOS local provider | available no-op | available | supported through the HDC inventory host |
| Linux/Vega | unavailable | unavailable | fail closed |
| Web | unavailable | unavailable | fail closed |
| Limrun iOS simulator / Android emulator | available | available | supported provider cells |
| WebDriver iOS / Android | available | unavailable | fail closed: owner-capability-missing |

## Stack status

[...]

PR #1755, opened 2026-08-11
refactor: route appstate through platform runtime

thymikee:
## Summary

Part of #1739.

Migrate `appstate` to the ADR 0019 request-bound platform runtime:

- declare the exact `ensureReady + appState` use on the command descriptor;
- inspect platform/provider facts once, bind once, and execute readiness before app-state lookup;
- keep Android and HarmonyOS query/parsing semantics in their owning platform packages, with request cancellation propagated to ADB/HDC;
- preserve active Apple session ownership and fail closed for unsupported or stale provider cells;
- retain exact-owner Limrun app-log recovery without exposing ordinary appstate/readiness/apps authority when no live provider session exists;
- enforce the single appstate execution path and retirement boundary as R22.

The change touches 80 files (+1,891/−366). The broad surface is the cross-platform appstate denominator plus the typed runtime contracts, capability projection, provider bindings, cutover gate, and production-route tests required to keep every platform/provider cell explicit.

## Validation

- `pnpm check:affected --run`
- 897 test files / 6,794 tests passed
- changed-line coverage 143/159 (89.94%); changed-branch coverage 100/126 (79.37%)
- Layering Guard: one appstate execution path; R22 retirement and operation singularity pass
- clean-install package verification imported all 12 published entry points
- daemon wire compatibility unchanged against v0.20.8
- planted red: removing the recovery-only `listApps` fact override made the Limrun binding test fail; fixed head exposes only reattach/cleanup operations and refuses apps/appstate/readiness narrowing without a live session

Live route evidence on the reviewed head:

- HarmonyOS physical nova 14: session-backed and sessionless `appstate` succeeded; session cleaned up
- Limrun Android: `appstate` succeeded; session cleaned up and provider disconnected
- Apple iPhone 17 Pro simulator (iOS 26.2): active-session `appstate` succeeded; session cleaned up

No CLI behavior documentation or skill changes are needed; the command surface is unchanged. ADR 0019 records the completed migration unit.

PR #1747, opened 2026-08-11
refactor: route boot through readiness runtime

thymikee:
## Summary

Move `boot` onto ADR 0019's request-scoped device-runtime tier at exact head `fcb2f35cc`:

- add side-effect-free, per-owner lazy `inspectFacts`; admission inspects once and a supported request binds exactly once through the discriminated `bootTarget` / `bootTargetHeadless` plan
- preserve Android emulator boot, Apple simulator boot, and physical-device readiness; headless boot is a fact, not a flag-only admission check
- move stopped-AVD placeholder resolution into Android inventory
- move Apple and Android readiness sequencing into their owning private platform packages; inject only narrow command/process ports
- make Apple automation keep-hot platform-internal and remove boot's daemon-threaded prewarm dependency
- propagate request cancellation through Apple simulator/physical and Android emulator polling; clean up an emulator started by an aborted request
- delete the root cross-family readiness semantic adapter and enforce its retirement
- add the `boot` R20 row to the shared parameterized cutover gate

Evidence tier: **request-scoped**. No admission ledger, fence, durable descriptor, recovery adapter, or durable lifecycle state is added.

### Legacy-parity denominator

| Owner/cell | Normal | Headless | Result |
| --- | --- | --- | --- |
| local Android emulator, including stopped AVD placeholder | available | available | package-owned boot and polling |
| local Android physical device | available | unavailable | shared physical readiness |
| local Apple iOS/iPadOS/tvOS/visionOS simulator | available | unavailable | package-owned simulator readiness |
| local Apple iOS physical device, CoreDevice or XCTest | available | unavailable | signal-aware narrow physical port |
| macOS host and watchOS sentinel | unavailable | unavailable | explicit fail-closed facts; zero bind |
| HarmonyOS | unavailable | unavailable | `ensureReady` remains the legacy no-op needed by later units; boot stays unavailable |
| Linux, Vega, web transport-composed | unavailable | unavailable | explicit fail-closed facts |
| direct Limrun iOS simulator / Android emulator | available | unavailable | provider-owned facts and operation |
| direct WebDriver iOS / Android | available | unavailable | provider-owned facts and operation |

Facts tests enumerate the local platform leaves and both provider owners. Handler regressions prove unavailable cells inspect once/bind zero, while supported cells inspect once/bind once. In-flight abort tests cover Apple simulator, Apple physical readiness, and a request-started Android emulator.

[...]

PR #1759, opened 2026-08-11
refactor: route application lifecycle through runtime facts

thymikee:
## Summary

Route `open`, `prepare`, `close`, and internal `runtime` through ADR 0019 facts-first admission and package/provider-owned lifecycle bindings. Each route inspects exact-owner facts once, admits the complete operation set, and binds once; unsupported platform/provider cells fail closed without a local fallback.

The stacked catalog remains unique: R18 contracts, R19 selector, R20 boot, R21 apps, R22 appstate, R23 shutdown, R24-R27 deployment, and R28-R31 lifecycle. Legacy `close` dispatch is retired, the runtime port-reverse route supports an explicit provider selector before session open, and Android test-IME recovery remains fail-closed only after durable state can be unsafe.

Scope: 213 files, +14,608/-6,956 against #1758. This covers four descriptors across platform/provider runtimes, their contracts and cutover proofs, and removal of superseded lifecycle routes.

## Validation

- Exact head: `183ce1715248`; base: #1758 `7fd538f0e211`.
- `pnpm check:affected --run && git push` passed from the committed tree: 599 test files / 4,672 tests, 90.11% changed-line coverage, plus format, lint, typecheck, layering, Fallow, build/package, provider integration, replay compatibility, wire compatibility, and gate-model checks.
- Focused lifecycle validation passed: 17 structural cutover proofs, 15 runtime/CLI/IME unit regressions, provider port-reverse composition, lifecycle recovery, and application-lifecycle contract/platform tests.
- Red-before: the provider scenario failed before the port-reverse fix with `SESSION_NOT_FOUND`; the planted close descriptor with `dispatch` fails the cutover proof.
- Claude hosted review of the final repair slice verified two P2 cleanup findings; both are fixed: stale recovery `sessionName` was removed and duplicate IME persistence diagnostics were eliminated. A follow-up review covers the exact cleanup commit.
- No local device tests were run at the user's request while benchmarks were active. GitHub platform/device lanes remain authoritative.
- No docs or skills changed: this preserves the public command surface and moves ownership/cutover internals.


thymikee:
Reviewed exact head 4ee2025e. Changes requested.

[P1] This does not establish the ADR 0019 ownership seam. Platform/provider packages mostly classify facts and call `host.applicationLifecycle.bind(...)`; the one root host then lazy-loads the 545-line open module and 236-line close module, where Apple/Android/provider mechanics, readiness, runner, IME, perf, and macOS behavior still execute. Move concrete lifecycle operations behind their owning platform/provider adapters and keep root composition limited to narrow neutral host capabilities.

[P1] ADR 0019 defines one canonical descriptor as the abandonment-safe migration unit. This PR batches `open`, `prepare`, `close`, and `runtime` across 128 files, changes all four discriminators to `device-runtime`, and adds no lifecycle cutover rows. Split the descriptors and give each its parametrized cutover row plus a production-route planted-red witness. The reported Android fact flip proves a package fact test, not removal of legacy or parallel shipped execution.

[P1] The direct base #1747 and prerequisites #1740/#1745 remain draft. Live evidence covers one Android-emulator open/close/test-IME recovery scenario, but not Apple-only `prepare`, runtime-hint apply/clear, macOS/physical Apple, or provider paths. CI also still has jobs in progress, so this is not ready.

[P2] The immediate delta is +104,691 source bytes and +40,614 raw shipped JS (+13,881 gzip), with no pre-start per-descriptor budgets. New and expanded files exceed project tripwires, including the 545-line root open implementation and 1,961-line close-shutdown test. This exceptional growth is a symptom of the combined root-owned design; splitting ownership should materially reduce it.

The facts-first admission, plan-specific single binds, exact provider resolver capture, durable IME ordering/recovery, and fail-closed handler tests otherwise look coherent. No ready label until ownership, unit scope/gates, budget, live evidence, dependencies, and CI are resolved.

thymikee:
Adversarial review of new lifecycle draft exact head `4ee2025e`: **not clean**. Confirmed blockers:

- Add parametrized cutover rows for `open`, `prepare`, `close`, and `runtime`; declare close’s hint-dependent use in its descriptor plan.
- Bring full routes behind admitted bindings: `open --foreground` currently probes local iOS before facts; `close --shutdown` uses ambient provider/legacy shutdown; `runtime port-reverse` bypasses its declared use; open/close still wrap the legacy dispatcher.
- Remove ambient provider ownership from open claims/close shutdown; derive policy from the selected runtime owner and add a provider-binding/no-ambient regression.
- Make stale Limrun lifecycle facts require a live session before bind.
- Fence Android IME startup recovery against the fire-and-forget open interleaving.
- Expire the old ref frame immediately before any mutating preparation/hint-clear operation, not only final launch/close.
- Move lifecycle semantics into owning packages. The current root universal lifecycle host retains cross-family mechanics, while capture-kit was expanded into forbidden generic platform-common. Root production grows ~1,902 LOC.
- Correct tarball/unpacked evidence to authoritative CI values.
- Decompose the new 545-line lifecycle-open implementation and extract additions from 648-line session-close; split the +295-line shutdown test by concept.

Live evidence covers only Android emulator open/IME recovery; macOS, physical Apple, and live-provider lifecycle evidence remain absent. Keep draft, address independently, run `pnpm check:affected --run` before push, and request another exact-head review.

thymikee:
Exact-head adversarial rereview of `b382e3f2`: **not clean**. Confirmed blockers:

1. **P1 — legacy close execution still ships.** Close is `device-runtime` but retains `dispatch: {}` and `DISPATCH_HANDLERS.close` in the generic dispatcher. Delete the legacy close projection and ensure R21 enforces retirement.
2. **P1 — package ownership remains a repackaged root dispatcher.** Broad lifecycle host/tool bags and monolithic `Interactor` semantics still route package modules into root dynamic imports of legacy platform/daemon mechanics. Replace with narrow low-level ports and package-owned mechanics; preserve laziness.
3. **P1 — open performs platform-specific target/surface policy before facts inspection/bind.** Move all Apple/Android resolution/classification behind the admitted owner; facts must be first.
4. **Attributable CI — Integration/Coverage are red.** Migrated Android open always tries test-IME activation, but clean fixtures lack the bundled manifest, causing missing-artifact failures and leaked claims. Fix the production/fixture contract and prove cleanup; rerun full affected gate.
5. **Evidence/size:** refresh tarball/unpacked from authoritative CI and continue decomposing touched >300 LOC concepts where locality warrants it.

The four cutover rows, declared plans, single inspect/bind, fail-closed selection, Limrun liveness, IME lock, ref expiry, capture-kit cleanup, Linux/Vega modules, and R7/R9/R10 are fixed. Keep draft; independently classify/rerun both smoke failures and retain missing Apple/provider live-evidence blockers.

thymikee:
Reviewed exact head `b382e3f2`. This is not ready.

- **P1 — Android test-IME behavior regressed.** The migration lost both the provider-owned exclusion and the documented best-effort handling: emulator-default `enableTestIme` now unconditionally awaits local bundled-IME activation. Exact-head Integration and Coverage consequently report 30 failures across 7 files with “Android test IME … helper artifact … not found.” Restore provider exclusion and best-effort behavior at the typed owning boundary, and plant a provider/emulator regression that fails on this head.
- **P2 — lifecycle ownership remains inverted.** The package binders now own sequencing, but broad root `AppleApplicationTools`/`AndroidApplicationTools` implementations still dynamically import and own concrete runner, simulator/device-target, runtime-hint, and IME mechanics. Replace these service bags with focused platform-owned capabilities.
- **P2 — cutover is incomplete.** `close` still has `dispatch: {}` and `DISPATCH_HANDLERS.close` remains shipped; R21 only retires the deleted wrapper, not this legacy route. R19–R22 also collide with active-stack rule allocations, and `open` still performs platform/surface target policy before facts admission.

The branch is CLEAN/MERGEABLE but authoritative Integration, Coverage, Android Smoke, and iOS Smoke are red. Live evidence covers only an Android-local happy path, not provider or Apple/macOS/physical parity. The exceptional bundle/source growth and several near-tripwire modules also still need tightening. No readiness label.

thymikee:
Took this over at `b382e3f2`, ran a thermo-nuclear quality pass, and pushed `4cef1f36`. Production source is **-217 net lines** (+624/-841) across 47 files, and the whole `pnpm check:affected --run` gate passes from a clean committed tree.

## Attributable CI (blocker 4) — root cause found, fixed, and the scenarios now prove the real path

The migrated Android open dropped the `try/catch` that made test-IME activation best-effort, and the comment that replaced it justified the wrong thing:

> Durable marker persistence is a pre-mutation fence. A failed fence is not an optional optimization: abort the open rather than silently continuing after an unprotected switch.

`activateAndroidTestIme` never throws for a failed fence — it returns `persistFailed: true` without switching. The only thing the removed `catch` was suppressing is helper *acquisition* (missing/unbuilt bundled artifact, device-side install rejection), which is exactly the case that must fall open. Since test IME is default-on for emulators, that made every `open` fail on a checkout without the artifact. 7 provider-integration files were red; all 50 are green now.

Two of them (`android-lifecycle`, `android-find`) were red for a second reason. They assert the IME broadcast channel and explicitly assert `shell input text` is never used — but the IME helper had no provider-supplied artifact override, so activation could only ever succeed from a built npm bundle. It now follows the precedence the snapshot and touch helpers already have (`adbProvider.imeHelperArtifact`), and the scenarios supply a fixture artifact. Those assertions now pass for the right reason instead of being unreachable.

Regression coverage: `src/__tests__/platform-runtime-android-application-tools.test.ts` — an unobtainable helper resolves, a failed durable fence still rejects. Proven red against the pre-fix tool.

## Package ownership (blocker 2)

`createComposedPlatformRuntimeGateway` named `androidApplications.hasTestImeRecoveryEvidence`, `recoverTestImeStartup` and `appleApplications.detach/finalizeRunnerSessionsForShutdown` directly — a neutral gateway that knows Android IME. Which durable resources exist is now the host's composition (`platform-runtime-application-resources.ts`); the gateway keeps only the startup fence, the lazy host load and the once-per-process shape of each phase. The cutover gate was rewritten accordingly: it now **rejects** a gateway that names any platform durable owner, and proves the marker-evidence ordering at the new owner, with two planted-red tests.

## Duplication the review hadn't itemised

- **5 copies of the facts-admission trio.** `open`, `close`, `prepare`, `runtime` and the pre-existing `boot` each declared their own `requireXFacts` / `requireXBinding` / `xUnavailableResponse`. All five throw the identical `runtime-gateway-missing` error that `RequestExecutionScope` already throws. One `src/daemon/runtime-admission.ts` owns the single side-effect-free inspection and the bind. The gate no longer enumerates five hand-picked helper names: each route must make exactly one shared admission call, `inspectFacts`/`bindDevice` are forbidden call names in every lifecycle handler, and the one-inspection/one-bind promise is proven once at the shared module.
- **6 packages hand-rolling the same lifecycle binding.** harmonyos/linux/vega/web/webdriver/limrun each wrote their own open, close, target resolution and `unavailable()` — 548 lines that differed only in an owner label, whether the target carries a bundle id, and whether relaunch closes first. (limrun's target resolver was the same rule as harmony's, written as nested ternaries.) One `bindDirectApplicationLifecycle` factory in contracts; each package is now a ~28-line declaration.
- **3 copies of the transport-hint predicate** over the "deliberately neutral" `RuntimeHintValues`, each re-listing `metroHost`/`metroPort`/`bundleUrl` in platform-apple, platform-android and daemon recovery. The contract owns the key set; `hasRuntimeTransportHints` delegates to it.
- **`src/platforms/android/ime-recovery-fence.ts`** was three aliases of contract exports. Deleted; the gate keys on the real symbol.
- **The two session-finalization entry points** each duplicated the finalize/clear-hints pair (4 copies of the same effects). One body now; they differ only in how they bind and dispose. The daemon path binds lazily, so an unsupported cell still fails on facts alone instead of allocating an owner first.

No file crosses 1000 lines.

## Still open (unchanged by this pass)

The live-evidence gaps stand: macOS, physical Apple, Apple `prepare`/hints, and a live provider lifecycle target. Dependencies #1740/#1745/#1747 are still draft. Keep this draft until those close.

thymikee:
Reviewed exact head `4cef1f36`. The ownership/facts-first delta is directionally sound, but this remains not ready.

[P1] `createAndroidApplicationTools().activateTestIme` catches every rejection from `activateAndroidTestIme` while intending only helper acquisition/install failure to be best-effort. It therefore also swallows startup-recovery-fence rejection, recovery-lock failure, and transport/read failures after durable state work, allowing `open` to continue with recovery incomplete or mutation state unknown. The regression mocks one generic `AppError` and consequently blesses that overbroad catch. Represent helper unavailability/install refusal as a typed non-activation outcome (or catch a structured reason at the acquisition seam), let fence/lock/post-record failures propagate, and plant both cases through the shipped route.

[P1 integration] This sibling still assigns R20/R21/R22 to prepare/close/runtime while #1758 assigns those same IDs to install/reinstall/install_source on the same boot base. The row model does not reject duplicate rule IDs, so rebase/consolidate against one accepted catalog before either sibling merges. The PR is also `DIRTY` after its base advanced and requires full stack replay.

Exact-head Android Smoke is red at `wait text Automation lab`; because this delta changes Android open/IME behavior, it requires owner investigation and a green rerun rather than an infrastructure assumption. The PR body remains at `b382e3f2`, with stale size/validation evidence and no disclosure of the red lane or dirty base. Exceptional growth still needs explicit acceptance, and live provider/macOS/physical-Apple lifecycle evidence remains absent. No ready-for-human label.

thymikee:
Rebased onto `main` at `96e891e7` and addressed the review. `pnpm check:affected --run` passes from a clean committed tree (527 test files, 4265 tests); the PR is `MERGEABLE` again.

## Retargeted to main, which changes the dependency picture

`main` already carries the boot unit (#1747), the parametrized cutover gate (#1745) and the apps unit (#1756). Retargeting off `refactor/adr19-boot-unit` **removes the draft-prerequisite blocker entirely** — this no longer waits on #1740/#1745/#1747.

It also answers the standing P2 from upstream rather than from me: main moved readiness into `packages/platform-*/src/readiness/` behind three narrow ports (`applePhysical`/`appleAutomation`/`androidEmulator`). The lifecycle bindings now call `ensureAppleReady`/`ensureAndroidReady` instead of a root readiness bag, and `ensureAppleReady` gained an `onColdBootStart` hook so `open` keeps warming the runner cache during a cold boot.

History is one commit; the replay's WIP checkpoint is not in it.

## [P1] Overbroad catch — replaced with a typed outcome

You were right that catching every rejection also swallowed startup-fence, recovery-lock and post-durable-state failures, and that the old test blessed it. `activateAndroidTestIme` now returns a discriminated result, so the tool needs no catch at all:

- `outcome: 'helper-unavailable'` — the acquisition seam (`selectAndroidImeHelperArtifact` + `ensureAndroidImeHelper`) failed, nothing was mutated → warn, fall back to ordinary text entry.
- everything else propagates; `persistFailed` still rejects.

Three planted-red cases cover exactly what the old catch hid — fence failure, lock failure, post-record transport failure — and they fail against a re-planted overbroad catch.

## [P1] Rule-ID collision — fixed, and made unrepresentable

Rows take **R24–R27**, clear of the accepted catalog (R20 boot, R21 apps) and of #1758's R18/R20–R23, so either sibling can land first. `cutoverTableDefects` now rejects a duplicate rule id, so the row model no longer permits the collision.

## Android Smoke — this PR's bug, not infrastructure

Android is green on the base and red on both prior heads, so I traced it. `openNewSessionWithDeviceClaim` computed:

```js
openPositionals: resolvedOpenTarget ? [resolvedOpenTarget] : (req.positionals ?? [])
```

On a **new** session `openTarget` is `positionals[0]`, so `open <app> <url>` collapsed to `[app]` — **the deep link was silently dropped on every platform**. The smoke journey opens a deep-linked route, so it landed on the default screen and `wait text Automation lab` timed out. Fixed; positionals pass through and a target is only synthesized when there were none (`--foreground`). Regression test proven red against the pre-fix code — the PR had deleted `session-open-launch-url.test.ts`, which is why nothing caught it.

## Two more defects found while reviewing

- **The IME helper had no provider-supplied artifact override**, so activation could only ever succeed from a built npm bundle. It now follows the precedence the snapshot and touch helpers already had, which is what lets the Android scenarios assert the broadcast channel and `expect(no 'shell input text')` for the right reason.
- **Package-owned readiness cost a second `simctl list` per flow**, because the boot checks that follow read a root memo the package could not seed. A narrow `markBooted` port carries the fresh observation — back to one listing, which the tvOS scenario asserts.

## Size

From this PR's own report: **+45.2 kB raw / +15.4 kB gzip / +10.7 kB tarball**, startup unchanged (`--version` -0.3 ms, `--help` -0.5 ms). Production source is ~217 lines smaller than the pre-review head, from deleting duplication: five per-command facts-admission trios → one module; six hand-rolled package lifecycle bindings → one factory; three transport-hint predicates → one; four session-finalization branches → one; one identity-wrapper module deleted. `platform-linux` and `platform-vega` also dropped their `capture-kit` dependency.

## Still open

Live evidence covers the Android emulator path only. macOS, physical Apple, Apple `prepare`/runtime-hint apply-clear, and a live provider lifecycle target have **no** live run — those lanes are GitHub-authoritative and should be read from this run, not assumed.

thymikee:
Force-pushed `0b80def5b`. **Correcting my previous comment**: the head I pushed before it (`96e891e7`) silently reverted work already on `main`, and I want that on the record rather than quietly fixed.

## What went wrong

I replayed this unit as a patch against `07d528086`, then committed it onto a `main` that had advanced to `c7565cb1f`. The patch therefore carried **reversions of everything main had changed in the files it touched** — #1754's snapshot-ownership refactor, `.fallowrc.json` exemptions, a `vitest.config.ts` entry, and ~25 unrelated files (logcat, batch, app-log resource store, install-source).

The local affected gate passed on that tree, so nothing caught it. A gate validates the tree in front of it; it cannot know the tree lost something. I only found it by chasing the red mutant lanes.

**That also explains the iOS Smoke failure on `96e891e7`** (`is visible id="automation-longpress"`): iOS is green on main at both `07d528086` and `c7565cb1f`, and `is visible` depends directly on the snapshot code that head reverted.

## The fix

Rebuilt on `c7565cb1f` from the pre-squash work, which never contained the reversion, then:

- restored main's version of every file the unit does not own;
- migrated the unit's callers to main's `appLogResourceStore` API;
- dropped `src/platforms/android/__tests__/logcat.test.ts`, which main deleted;
- ran a **line-level** scan across all 203 changed files for lines main has that this tree lost. One real loss: `src/__tests__/test-utils/index.ts` had lost main's `mkdtempForTest` re-export while legitimately adding the IME fixture export — a file-level check could never have caught that. The other 21 hits are this unit's genuine restructuring (capability folding in `session-inventory.ts`, the close-teardown extraction, `settleDaemonTeardownStep`).

Diff is now **203 files** instead of 287; the 84-file difference was reverted work, not this unit's.

`pnpm check:affected --run` passes from a clean committed tree: 526 test files, 4267 tests.

## Mutant lanes are broken on `main`, not here

All ten `Mutants (*)` lanes plus `Affected decision-kernel mutants` fail, and they fail identically on a clean checkout of `origin/main`:

```
pnpm mutation:run --modules interaction-settle
ERROR DryRunExecutor One or more tests failed in the initial test run:
  Cannot find package '@agent-device/selectors/engine' imported from
  <repo>/.tmp/stryker/sandbox-XXXX/src/core/selector-pipeline.ts
```

The subpath resolves in a normal run; it does not resolve from inside Stryker's sandbox. Because the dry run fails, the lanes exit before mutating anything — **mutation scores are not being measured on main at all right now**, so the ratchet is currently unguarded. That needs its own fix; it is not this PR's to make.

## Where this head stands

Android Smoke passed on the previous head, which is the evidence that the dropped-deep-link fix works. Everything else should be read from this run — I am not carrying forward the earlier greens, because they were measured on a tree that had reverted main.

thymikee:
CI settled on `0b80def5b`. **All 31 non-mutant checks pass**, including every lane that was red when this review started:

| Lane | At review start | Now |
|---|---|---|
| Integration Tests | fail (30 tests / 7 files) | pass |
| Coverage | fail | pass |
| Android Smoke | fail (`wait text Automation lab`) | pass |
| iOS Smoke | fail (`is visible automation-longpress`) | pass |
| Linux / macOS / Web Smoke | pass | pass |
| Layering, Fallow, Typecheck, Lint, Bundle Size, Replay-Compat, Released-Surface, CodeQL, … | — | pass |

Two notes on how those greens were reached, so they can be weighed properly:

- **iOS Smoke** was collateral from the reverted-`main` head described above, not from the lifecycle unit. Removing the reversion fixed it; nothing in the unit changed to make it pass.
- **Coverage** first failed with `[vitest-pool]: Worker forks emitted error` — a worker *teardown* timeout on `src/platforms/apple/core/__tests__/runner-session.test.ts`, with no failed test recorded. This PR touches neither that test nor any Apple runner code, and it passes locally 73/73. CI's contention-retry classifier declined to auto-retry it ("a reason a rerun cannot re-check"), so I re-ran the job manually and it passed. Worth knowing that this failure shape currently escapes the retry classifier.

## The ten mutant lanes are broken on `main`

They fail identically on a clean checkout of `origin/main` **and** on this exact head:

```
pnpm mutation:run --modules interaction-settle
ERROR DryRunExecutor One or more tests failed in the initial test run:
  Cannot find package '@agent-device/selectors/engine' imported from
  <repo>/.tmp/stryker/sandbox-XXXX/src/core/selector-pipeline.ts
```

The subpath resolves in a normal run and not from inside Stryker's sandbox. Because the dry run fails, the lanes exit before generating a single mutant — **mutation scores are not being measured on main right now, so the ratchet is unguarded**. That is a separate fix and not this PR's to make; flagging it because it is currently invisible behind ten red checks that read like PR noise.

## State

Based on `main` (no draft prerequisites), `MERGEABLE`, one commit, 203 files. `pnpm check:affected --run` green from a clean committed tree (526 files, 4267 tests).

Live evidence still covers the Android emulator path only — macOS, physical Apple, Apple `prepare`/runtime-hint apply-clear, and a live provider lifecycle target have no live run. The CI lanes above are the authority for what they cover; they are not a substitute for that missing evidence.

thymikee:
Rebased onto `main` at `b59b4e5c9` and force-pushed `4e823165c`. **CI is clean: 40/40 checks pass**, and the PR is `MERGEABLE`.

`main` had moved onto #1755 (appstate through platform runtime) and #1751, which overlap this unit heavily — 28 files conflicted, and three of them needed a real decision rather than a merge:

**1. `capabilities` now always takes its one facts snapshot.** #1755 skipped the sessionless inspection whenever the session already owned appstate. With this unit, `open`/`close`/`prepare`/`runtime`/`apps` availability *is* a runtime fact, so skipping it would report all five unavailable for exactly the sessions that support them. The session short-circuit survives as an explicit `appstate` override, and #1755's test changed from asserting "must not inspect" to asserting the invariant that actually matters: a snapshot reporting appstate unavailable still leaves the command available. `apps` is a small correctness gain too — it was collaterally reported unavailable for those sessions before.

**2. Limrun liveness is #1755's `hasLiveSession`, not this unit's interactor probe.** Both sides had independently introduced a liveness signal; the explicit option wins, and this unit's lifecycle facts now hang off it. A stale session therefore keeps #1755's recovery-only binding, and the lifecycle test was rewritten to assert that shape: an ordinary bind rejects with `provider-session-unavailable`, and an exact-owner bind exposes `appLogReattach`/`appLogCleanup` and nothing else.

**3. `createUnavailablePlatformRuntimeOwner` is deleted.** Its only two production callers were the Linux and Vega runtimes, which now own a lifecycle denominator and construct their own owner. Nothing outside its own test referenced it.

The fallow gate also caught a duplication I introduced while resolving: hoisting a 105-line host fixture into the Limrun fixtures file cloned one already in `platform-apple`. Replaced with the narrow existing fixture, plus a `limrunOwnerOptions` factory that collapses nine repeated option literals — so the next option added to that owner lands in one place instead of nine.

## Verification

`pnpm check:affected --run` passes from a clean committed tree: format, lint, typecheck, build/package, layering (181/181), fallow (0 complexity findings, 0 clone groups), coverage (529 files / 4289 tests), provider integration, replay and daemon-wire compatibility. The layering guard now reports **eleven** migrated commands (appstate, boot, apps, devices, logs, network, record, open, prepare, close, runtime) keeping exactly one platform-execution path.

## One red lane, and why it was not this PR

macOS Smoke failed once with `macos device host-macos-local is owned by session "…:attempt-1"`. That message belongs to attempt 3 and only names attempt 1, so it is not the failure — attempt 1's request log is:

```
request_failed                          error: "swift timed out after 10000ms"
device_claim_open_effects_unconfirmed   deviceKey: local:apple:macos:host-macos-local
```

`swift --version` has a 10 s budget inside `computeMacOsHelperFingerprint`; a cold runner exceeded it. Because the failure landed after the preparation boundary, `rollbackNewSessionClaim` deliberately retained the claim, and attempts 2 and 3 then failed `DEVICE_IN_USE` without running any real code.

Attribution, checked rather than assumed — `rollbackNewSessionClaim`, the position of `effects.mayHaveStarted = true`, and the claim-clear on failed close are all byte-identical to `main`; `isLocalDeviceClaimTarget` changed signature (admitted owner instead of request metadata) but returns the same verdict for a local macOS device. `src/platforms/apple/os/macos/helper.ts` is untouched here, and the lane is green on `main`. Re-ran the job; it passed.

Two things that lane exposed, both pre-existing and out of scope for this PR:

- A retryable first-attempt failure consumes the entire retry budget, because the retry loop reuses one device and cannot reclaim its own prior attempt's retained claim. The retention policy itself is right — handing an unconfirmed device to the next session would be worse.
- 10 s is a tight budget for `swift --version` on a cold runner, and it is only feeding a fingerprint hash.

## Residual gaps unchanged

Live evidence still covers Android emulator open/close/test-IME recovery only. macOS, physical Apple, Apple `prepare`/runtime-hint apply-clear, and a live provider lifecycle target are GitHub-authoritative; this green run is the gate for them.

_🤖 Prepared with [Claude Code](https://claude.com/claude-code)_

thymikee:
Stack integration update: lifecycle rows are now uniquely allocated as R28-R31 after shutdown R23 and deployment R24-R27. The PR base is now #1758 so GitHub reflects the actual landing order. This branch still requires a deliberate semantic replay over the shared runtime contracts; a mechanical rebase produced 50+ real overlaps and was aborted rather than choosing one side and dropping deployment/lifecycle behavior. No local device tests or broad CPU-heavy gates were run while the host is reserved for benchmarks.

thymikee:
Reviewed exact head `2e82fe677`. **Not ready.**

- **P1 — preserve the landed rule-ID source of truth.** This branch changes `contracts-implementation-authority` from main’s R18 back to R11 and restores the transitional R11 collision allowance, recreating the collision that main already removed. Keep R18 and the empty allowlist, with the combined-tree uniqueness proof intact.
- The branch is currently **DIRTY/CONFLICTING** and has zero checks on this exact head. Resolve it with a semantic rebase onto current main, preserving landed behavior, then run the full authoritative CI set.
- **P2 — module/test topology.** This change newly pushes `runtime-command-cutover-extensions.ts` from 294→572 LOC, `platform-runtime-gateway.test.ts` from 439→599, and WebDriver `platform-runtime.test.ts` from 200→503. Tests are not exempt from the >500 extraction rule; split these by the production questions they prove while retaining shared fixtures.
- The body is materially stale: it still claims R24–R27/start-at-R24 although exact-head code correctly reserves R28–R31 after shutdown/install allocations; it omits the 196-file expanded scope, and its size/validation claims are not exact-head CI evidence.

The lifecycle direction otherwise looks sound and consistent with ADR 0019: exact-owner facts are inspected before admission, the complete selected operation set is checked, and only then is the package/provider runtime bound. Android live evidence covers open/close/IME; Apple prepare/runtime-hints and live provider lifecycle remain residual evidence gaps.

thymikee:
Exact-head readiness update for 183ce1715248: code review is clean and the branch is ready for human review. The final repairs retire the legacy close dispatch, allow exact-owner provider port reverse before session open, preserve safe Android IME fallback while keeping unsafe recovery states fail-closed, remove stale lifecycle input, and add planted cutover proof. The repository-required pnpm check:affected --run passed in the same command chain as the push (599 files / 4,672 tests; 90.11% changed-line coverage; build/package/layering/Fallow/provider/wire/tooling gates green). Claude hosted review found two P2 cleanup issues in the repair slice; both were fixed, and its follow-up review of the exact cleanup commit completed with no findings. No local device tests were run by request; GitHub platform/device lanes remain authoritative and are currently in progress.

PR #1761, opened 2026-08-12
fix(batch): preserve typed error recovery signals

cosin2077:
### Summary

Preserve structured recovery signals from a failing batch step.

`DaemonError` can carry `retriable` and `supportedOn`, but `runBatch()` rebuilt the failing step error without forwarding either field. That meant the same underlying failure lost machine-readable recovery guidance when invoked through `batch`.

This change:

- forwards `retriable` and `supportedOn` when present;
- keeps those keys absent when the underlying error does not classify them;
- uses the canonical `DaemonError` type instead of a narrower local copy;
- adds regression coverage for both preservation and absence semantics.

### Test plan

- Added unit coverage for preserving `retriable` / `supportedOn`.
- Added coverage that unknown recovery signals remain absent.
- Full repository gates are pending GitHub Actions approval for this fork PR.

### DIFF
--- a/src/core/__tests__/batch.test.ts
+++ b/src/core/__tests__/batch.test.ts
@@ -75,6 +75,21 @@ test('batch preserves typed error recovery signals from a failing step', async (
   assert.equal(response.error.supportedOn, 'android, web');
 });
 
+test('batch keeps unknown typed error recovery signals absent', async () => {
+  const response = await runBatch(batchRequest(['open']), 'session', async () => ({
+    ok: false,
+    error: {
+      code: 'INVALID_ARGS',
+      message: 'bad args',
+    },
+  }));
+
+  assert.equal(response.ok, false);
+  if (response.ok) return;
+  assert.equal('retriable' in response.error, false);
+  assert.equal('supportedOn' in response.error, false);
+});
+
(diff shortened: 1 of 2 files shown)


thymikee:
Reviewed exact head 6cc05f72. Using the canonical DaemonError shape and conditionally copying retriable/supportedOn fixes the real batch-wrapper truncation while keeping unknown signals absent. The production route remains compatible with generic request finalization, the regression is non-vacuous, touched-file topology is healthy, and this pure wire-composition fix does not require device evidence. No source issue remains.

This is not readiness-ready yet: the PR body is still the untouched template. Please document the motivating recovery-signal loss, concise summary, validation with red-before evidence, and the two-file scope. GitHub also reports zero checks for this exact head, so it cannot yet be called CI-green. Keep draft until the body and authoritative checks are complete.

No ready-for-human label applied.

thymikee:
Readiness update for unchanged exact head `6cc05f72`: the title, substantive body, and draft-state blockers are resolved. The reviewed change remains code-clean and ready for human review; no device evidence is needed for this pure error-wire composition fix.

The only material blocker is owner approval and completion of full authoritative CI: GitHub still reports zero checks. Please also finish the body evidence with explicit two-file scope accounting and the actual red-before result, rather than only stating that coverage was added.

PR #1762, opened 2026-08-12
fix(install-source): preserve URL credential validation error

cosin2077:
### Summary

Preserve the specific validation error for install-source URLs containing credentials.

`parseSourceUrl()` currently throws `Source URL credentials are not allowed`, but that error is immediately caught by the surrounding `catch` and replaced with the generic `Invalid source URL` error.

Move credential validation outside the URL parsing `try/catch` so malformed URLs still receive the generic error while valid URLs containing credentials retain the intended validation error.

### Test plan

- Extended the existing credential-redaction test to assert the specific credential validation error.
- The existing secret-redaction assertions remain in place.
- Full repository gates are pending GitHub Actions approval for this fork PR.

### DIFF
--- a/src/platforms/__tests__/install-source-download.test.ts
+++ b/src/platforms/__tests__/install-source-download.test.ts
@@ -71,6 +71,7 @@ test('download errors do not disclose URL credentials or query values', async ()
       headers: { authorization: 'private-header' },
       signal: new AbortController().signal,
     }).catch((caught: unknown) => caught);
+    assert.match((error as Error).message, /credentials are not allowed/i);
     const serialized = JSON.stringify(error);
     for (const secret of ['private-user', 'private-pass', 'private-query', 'private-header']) {
       assert.equal(serialized.includes(secret), false, secret);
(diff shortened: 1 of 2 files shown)


thymikee:
Reviewed exact head `4f306fe3`. The code change is clean: narrowing the `try`/`catch` to URL construction preserves the intentional `INVALID_ARGS` credential-validation error while malformed URLs retain the generic error. The real `install_from_source` URL route reaches this parser, and the focused regression is load-bearing: 3/3 tests pass at this head, while restoring the pre-fix block makes the new assertion fail with `Invalid source URL`. No live-device evidence is needed for this pre-network error-classification change.

This is not readiness-ready yet: the title is non-conventional, the body is still the untouched template and needs the summary/motivation, two-file scope accounting, validation plus red-before evidence, and GitHub reports zero authoritative checks. Please update those and run the full required CI before moving the draft toward readiness. No code finding and no readiness label applied.

thymikee:
Readiness update for unchanged exact head `4f306fe3`: the title, substantive body, and draft-state blockers are resolved. The reviewed change remains code-clean and ready for human review; no device evidence is needed for this pre-network error-classification fix.

The only material blocker is owner approval and completion of full authoritative CI: GitHub still reports zero checks. Please also finish the body evidence with explicit two-file scope accounting and the actual red-before result, and record the already-proven focused result of 3/3 passing.

PR #1764, opened 2026-08-12
feat(snapshot): attach a fallback screenshot to sparse captures

thymikee:
## Summary

A user-facing sparse `snapshot` now captures the screenshot its recovery guidance already requires and returns the client-visible path as `fallbackScreenshotPath`. The image uses the normal artifact channel, so local and remote clients receive a usable path; internal observations and healthy snapshots do not trigger screenshots.

Only a `sparse-tree` verdict adds the app accessibility warning. Capture-limit verdicts remain attributed to the tool. The public contract, SDK normalization, digest projection, text output, architecture context, and snapshot documentation all describe the same response.

Scope: 16 files in the snapshot response/public projection. No command-family expansion. The `agent-device` skill remains unchanged because command behavior belongs in versioned help and product documentation.

## Validation

- `pnpm check:affected --run` — passed all 7 selected gates; 263 test files and 2,240 tests passed.
- Production-route regression covers sparse screenshot capture, artifact publication/materialization, best-effort failure, and internal-observation exclusion.
- Red-before proof: disabling the sparse fallback made the production-route test fail with `expected 1 screenshot call, received 0`.
- Live iOS simulator: source-built daemon opened Settings, returned a healthy tree without `fallbackScreenshotPath`, and closed the isolated session cleanly.

Residual: the repository has no live app fixture that deliberately publishes a sparse accessibility tree, so the sparse-device branch is proven through the production-route regression rather than a live sparse capture.

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -520,7 +520,9 @@ the observable freshness and failure semantics below before any runtime refactor
   stabilization is active.
 - Sparse snapshot quality verdicts are observable failures. Sparse captures must not replace
   `session.snapshot`, and selector routes should report the sparse verdict instead of treating a
-  root-only or sparse tree as an empty UI.
+  root-only or sparse tree as an empty UI. The user-facing `snapshot` dispatch publishes a fallback
+  screenshot through the response artifact channel (`fallbackScreenshotPath`); internal observations
+  never do, so a polling wait cannot turn an unreadable screen into one screenshot per poll.
 - iOS sparse and AX failures are not proof of empty UI. Regular visible snapshots can recover through
   the capture plan; raw and strict paths preserve failure. `runnerFatal` invalidates the cached target
   and must never refresh healthy mutation recency.
--- a/packages/contracts/src/client-capture.ts
+++ b/packages/contracts/src/client-capture.ts
@@ -43,6 +43,11 @@ export type CaptureSnapshotResult = {
   visibility?: SnapshotVisibility;
   unchanged?: SnapshotUnchanged;
   snapshotDiagnostics?: SnapshotDiagnosticsSummary;
+  /**
+   * Screenshot captured automatically when the semantic snapshot was sparse.
+   * Remote clients receive a materialized local path through the daemon artifact channel.
+   */
+  fallbackScreenshotPath?: string;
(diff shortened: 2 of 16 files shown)


thymikee:
Reviewed exact head `63a1c078`. The fallback screenshot dispatch structure is sound: it sits on the public snapshot response seam, skips `observationOnly` polling captures, composes its warning, and cannot fail the requested snapshot.

**P1 — `sparse-tree` is not proof of an app accessibility defect.** The runner stores the first capture-plan failure in the terminal verdict. A tree backend can set `firstFailure` to `sparse-tree`, while later query/private-AX tiers fail, exhaust budget, are unavailable, or never run; the final sparse payload still carries that first reason code. Therefore the new warning can falsely claim every backend reached the screen and that assistive technologies see an empty tree. It also contradicts the adjacent CONTEXT invariant that iOS sparse/AX failures are not proof of empty UI. Remove this attribution, or introduce an explicit aggregate terminal fact proving every applicable independent semantic backend returned semantic emptiness. Add a mixed-outcome regression showing tree-sparse plus later failure/budget/unavailable does not blame the app.

Before readiness, provide exact-head live iOS evidence on a genuinely sparse screen that the linked PNG exists and is readable, while internal polling does not capture repeatedly; the current mocked test proves dispatch decisions but not a usable artifact. Remove the unrequested `skills/agent-device/SKILL.md` behavior edit per the repository hard rule—versioned CLI help owns command behavior guidance. Finally rerun iOS Smoke to green: the current `automation-longpress` visibility failure is a known unrelated flake, but red CI still gates readiness. No readiness label applied.

thymikee:
Rebased onto current `main` and reviewed the final head `c42e1642f`. Code review is clean: the sparse screenshot now uses the normal artifact/materialization path, healthy and internal snapshots stay side-effect free, and app-defect wording is limited to `sparse-tree`. Local affected validation passed all 7 gates (263 files / 2,240 tests), including recording-artifact regressions; live iOS healthy-path verification also passed. Marking ready for human review while the new-head GitHub checks complete. Residual: there is no live sparse-tree fixture, so that branch is covered by the production-route regression and red-before proof.

PR #1217, opened 2026-07-11
refactor: replace Maestro compat with typed direct engine

thymikee:
## Summary

Replace the production Maestro YAML lowering/replay trampoline with a source-preserving typed program, immutable replay plan, and direct runtime port.

The new engine owns Maestro control flow, variables, target policy, assertions, and lifecycle semantics while reusing shared agent-device gesture and platform backends. The old SessionAction conversion, private __maestro commands, positional decoding, hidden caches, recursive dispatch, and runtime fallback were removed. Android snapshots now fail closed on missing, invalid, stale, or system-chrome-only helper output; persistent-to-command-scoped fallback remains within the same bundled helper contract and never invokes the legacy snapshot engine.

iOS snapshot presentation also narrows React Native LogBox/RedBox Dismiss targets to the innermost action geometry for all consumers, preventing wrapper bounds from misrouting taps.

## Validation

- Android API 36 Pixel 9 Pro XL: react-native-pager-view 9/9 in 430.4s; react-navigation 38/38 in 493.3s. Snapshot evidence reported android-helper 0.19.3 with persistent-session transport.
- iOS 26.2 simulator: react-navigation 38/38 in 814.6s.
- iOS 26.2 simulator: react-native-pager-view 4/9. All five remaining failures reproduce unchanged with upstream Maestro 2.5.1 on the freshly rebuilt app, including the suite cleanup label mismatch; no agent-device-only divergence remains.
- Review fixes: focused Maestro, gesture geometry, PNG worker, and replay failure suites passed 182/182; build, format, typecheck, lint, and fallow passed.
- The aggregate unit and coverage runs were affected by host-load timeouts in unrelated Metro, runtime-hint, Apple cache, and recording fixtures. The review-related Maestro replay file passed 69/69 in isolation, and all changed tests passed in the coverage run.

PR #1766, opened 2026-08-12
test(cli-help): gate every help example against the CLI schema

### LINKED ISSUE #1765: help react-native recommends open --metro-host/--metro-port, but open rejects both flags
## Summary

`agent-device help react-native` in `agent-device 0.19.3` recommends passing `--metro-host` and `--metro-port` to `agent-device open`, but the same CLI rejects both flags before app or device resolution.

## Reproduction

```console
$ agent-device --version
0.19.3

$ agent-device help react-native
Multiple local worktrees can reuse one native iOS simulator build by running each worktree's dev server on a different port and opening the same installed app on different simulators with explicit runtime hints:
  agent-device open "React Navigation Example" --platform ios --device "iPhone 17" --session rn-a --metro-host 127.0.0.1 --metro-port 8081 --relaunch

$ agent-device open com.example.app --platform ios --device "iPhone 17" --session rn-a --metro-host 127.0.0.1 --metro-port 8081
Error (INVALID_ARGS): Flags --metro-host, --metro-port are not supported for command open.
```

`agent-device open --help` also omits both flags, so the command help and topic help disagree.

## Expected behavior

Every command printed by `help react-native` should be accepted by the CLI parser. Either:

1. `open` should accept and apply the documented runtime-hint flags; or
2. the topic help should stop recommending unsupported flags and point to the supported runtime setup flow.

## Suggested patch

If the current public contract intentionally excludes these flags from `open`, the minimum documentation patch is:

```diff
- agent-device open "React Navigation Example" --platform ios --device "iPhone 17" --session rn-a --metro-host 127.0.0.1 --metro-port 8081 --relaunch
- agent-device open "React Navigation Example" --platform ios --device "iPhone 17 Pro" --session rn-b --metro-host 127.0.0.1 --metro-port 8082 --relaunch
+ # Configure each simulator's Metro endpoint through the supported project/session runtime flow first.
+ agent-device open "React Navigation Example" --platform ios --device "iPhone 17" --session rn-a --relaunch
+ agent-device open "React Navigation Example" --platform ios --device "iPhone 17 Pro" --session rn-b --relaunch
```

For project-owned RN debug shells, that preceding flow can be the shell's per-simulator runtime config. If `--metro-host` / `--metro-port` are intended to remain the generic agent-device mechanism, the parser/schema should instead be repaired and the existing help kept.

A regression test that parses every command example emitted by `help react-native` would prevent the help surface from getting ahead of the released CLI schema again.

## Related

- #197 introduced session-scoped runtime hints and explicitly discussed flags on `open`.
- #1245 assumes `open --metro-host/--metro-port` is available while describing a separate Expo dev-client limitation.

thymikee:
## Summary

`agent-device help react-native` recommended `open --metro-host/--metro-port` while `open` rejected both flags with `INVALID_ARGS` — help was ahead of the CLI schema for four releases. The reported command works on `main` today (`open` gained the flags in #1217), but nothing stopped the drift from happening, and a second instance was still live.

What changed:

- **New conformance gate.** Every example the CLI prints — root usage, all 20 help topics, all 67 command helps — is read back out of the rendered text and run through the production parser (`parseArgs(..., { strictFlags: true })`) plus the command's CLI reader. 274 examples check today. Lines whose unquoted tokens carry placeholder syntax (`<value>`, `[optional]`, `a|b`) are synopsis shapes and are skipped; quoted arguments such as `batch --steps '[{"command":…}]'` stay in scope.
- **`help macos` fix.** It advertised `agent-device snapshot -i --platform macos --surface menubar`, which the CLI rejects — `--surface` is an `open` flag, and the session keeps it (`snapshot-capture.ts` reads `session.surface`). The example drops the flag and the Surfaces block now says so explicitly.
- **Two prose lines reworded** (`help tv`, `help web`) so no help line starts with `agent-device ` unless it is a command. That keeps the gate's rule total instead of needing an allowlist, and removes two sentences an agent could read as commands.

Before/after for the macOS case:

```
$ agent-device snapshot -i --platform macos --surface menubar
Error (INVALID_ARGS): Flag --surface is not supported for command snapshot.

$ agent-device help macos    # now prints
  agent-device open "Agent Device Tester Menu" --platform macos --surface menubar
  agent-device snapshot -i --platform macos
```

Closes #1765

## Validation

`pnpm check:affected --run` green (38 files / 499 tests), plus `check:layering` and `check:fallow`.

The gate was proven red before green, twice against real defects rather than a synthetic one:

- Reverting `open`'s `...METRO_RELOAD_FLAGS` (recreating the released 0.19.3 schema) fails it with the issue's verbatim error: `Flags --metro-host, --metro-port are not supported for command open.` on both `help react-native` example lines.
- Restoring `--surface menubar` on the macOS snapshot example fails it with `Flag --surface is not supported for command snapshot.`

Live CLI evidence for the reported command shape, against a deliberately nonexistent device so no real one was claimed:

```
$ agent-device open "React Navigation Example" --platform ios --device NoSuchDeviceXYZ \
    --session rn-probe-1765 --metro-host 127.0.0.1 --metro-port 8081 --relaunch
Error (DEVICE_NOT_FOUND): No device named NoSuchDeviceXYZ
```

It now fails at device resolution instead of flag validation, which is the issue's expected behavior. The probe session's daemon was stopped.

## Scope and gaps

Three files, one module group (CLI help/parser). No website docs change: `website/docs/docs/commands.md` already documents `--surface` on `open` only and shows a bare `snapshot -i` after it, which is what the help now matches.

Known boundary, stated in the test header: only line-leading examples are checked. Examples embedded mid-sentence in command detail prose (`… for example agent-device open "Expo Go" exp://127.0.0.1:8081 --platform ios.`) have no reliable end delimiter — `127.0.0.1`, `.log`, and a sentence-final period are indistinguishable — so guessing where the command stops would trade this class of bug for false failures. Example blocks, the copy-paste surface, are fully covered.

### DIFF
--- a/src/cli/parser/__tests__/cli-help-examples.test.ts
+++ b/src/cli/parser/__tests__/cli-help-examples.test.ts
@@ -1,0 +1,130 @@
+import { test } from 'vitest';
+import assert from 'node:assert/strict';
+import { parseArgs } from '../args.ts';
+import { buildCommandUsageText, buildUsageText, helpTopicIds } from '../cli-help.ts';
+import { listCliCommandNames } from '../../../command-catalog.ts';
+import { readInputFromCli } from '../../../commands/cli-grammar.ts';
+import { isCommandName } from '../../../commands/command-metadata.ts';
+import { readVersion } from '../../../utils/version.ts';
+
+// Help is the agent-facing contract, and agents copy its example lines verbatim, so an example the
+// CLI schema rejects is a broken command surface — not a docs nit (#1765: `help react-native`
+// advertised `open --metro-host/--metro-port` for four releases before `open` accepted them).
+// The example set cannot be enumerated from a hand-written list without drifting from the text that
+// actually ships, so it is read back out of the rendered help itself: every line that starts with
+// `agent-device ` is either a runnable example or a synopsis shape, and shapes are exactly the ones
+// carrying placeholder syntax in an unquoted token (`<value>`, `[optional]`, `a|b`). Runnable ones
+// go through the production parser plus the command's CLI reader.
+//
+// Boundary: only line-leading examples are checked. Examples embedded mid-sentence in a command's
+// detail prose have no reliable end delimiter (`127.0.0.1`, `flow.log` and a sentence-final period
+// are indistinguishable), so they stay out rather than guessing where the command stops.
+
+const PLACEHOLDER_SYNTAX = /[<>[\]|]/;
+
+type Token = { value: string; quoted: boolean };
+
+type HelpExample = { surface: string; line: string; argv: string[] };
+
+const TOKEN_PATTERN = /'([^']*)'|"([^"]*)"|(\S+)/g;
+
+/** Shell-style split that remembers quoting, so `--steps '[{"command":…}]'` stays one argument. */
+function tokenize(line: string): Token[] {
+  return [...line.matchAll(TOKEN_PATTERN)].map(([, single, double, bare]) => {
+    const quoted = single ?? double;
+    return quoted === undefined ? { value: bare!, quoted: false } : { value: quoted, quoted: true };
+  });
+}
+
+/** Every help text a `--help`/`help <topic>` call can print, keyed by the argument that prints it. */
+function helpSurfaces(): Array<{ name: string; lines: string[] }> {
+  const topics = new Set(helpTopicIds());
+  const surfaces = [{ name: 'help', lines: buildUsageText().split('\n') }];
+  for (const topic of topics) {
+    const text = buildCommandUsageText(topic);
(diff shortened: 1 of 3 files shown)


thymikee:
Exact-head review at `bf00625d`: code-clean and ready for human review. The new gate reads the rendered help surface, sends runnable examples through the production strict parser and command CLI reader, and has meaningful red-before evidence for both the reported Metro flags and the live macOS `--surface` mismatch. The stated line-leading boundary is explicit and avoids inventing a prose parser. All 30 authoritative checks are green; branch is CLEAN/MERGEABLE. Applied `ready-for-human`.

PR #1767, opened 2026-08-12
fix: harden iOS alert smoke scenario

thymikee:
## Summary

Harden the fixture-backed iOS simulator alert smoke scenario against UI timing races.

The scenario now waits for the alert trigger to be visible after scrolling, waits for an app-visible `opened` canary after the click, defers native alert presentation by one frame, and gives XCTest a bounded 10-second alert wait.

## Root cause

The [failed job](https://github.com/callstack/agent-device/actions/runs/31578993018/job/94057524632) reported a successful click, but XCTest never observed the native alert during the 5-second query window. The preceding UI was not independently synchronized with the React Native state transition, making this a timing-sensitive failure.

## Validation

- `pnpm format`
- `pnpm typecheck`
- `pnpm exec oxlint examples/test-app/src/screens/AutomationLabScreen.tsx test/integration/ios-simulator-e2e/live-automation-scenario.ts`
- `node --experimental-strip-types --test test/integration/smoke-ios-simulator-coverage.test.ts` (14/14)
- Expo test-app typecheck was not run locally because its isolated dependencies were unavailable and npm registry access failed.

### DIFF
--- a/examples/test-app/src/screens/AutomationLabScreen.tsx
+++ b/examples/test-app/src/screens/AutomationLabScreen.tsx
@@ -86,7 +86,8 @@ export function AutomationLabScreen(props: {
 
   function showAutomationAlert() {
     setAlertResult('opened');
-    requestAnimationFrame(() => {
+    setTimeout(() => {
+      // The opened canary proves JS ran; the smoke step separately waits for native presentation.
       Alert.alert('Automation confirmation', 'Choose either result to update the visible canary.', [
         {
           style: 'cancel',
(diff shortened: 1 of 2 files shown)


thymikee:
Exact-head review at `93a95bf0`: the fixture/smoke hardening is coherent and no code finding remains. The scenario now proves trigger visibility, proves the click reached React state, then waits for the native alert with a bounded deadline; all code, integration, and Android/iOS/macOS/Linux smoke lanes are green. The lone red native-fingerprint job failed because the trusted-artifact lookup returned `fetch failed`, an infrastructure failure unrelated to this two-file test-app delta. Rerun that job; apply `ready-for-human` once the exact head is fully green.

PR #1769, opened 2026-08-12
fix(ios): use current devicectl capture-screenshot syntax for physical devices

### LINKED ISSUE #1760: iOS physical screenshot fallback calls devicectl with unsupported --device flag
## Reproduction

On a connected physical iPhone, invoke:

```sh
agent-device screenshot /tmp/test.png --platform ios --device "x iPhone" --json --verbose
```

## Actual

The fallback path logs:

```text
devicectl device screenshot --device 00008150-12345 /tmp/test.png
Error: Unknown option --device
Usage: devicectl device ...
```

It then falls back to the XCTest runner. In this run the runner could not start because the attached iPhone was locked, but the `devicectl` invocation itself is invalid independently and prevents direct physical-device capture.

## Expected

Use the current `devicectl` screenshot syntax for a selected physical device, so direct capture succeeds when the handset is unlocked and trusted.

Observed with agent-device globally installed via pnpm on macOS, 2026-08-12.

thymikee:
## Summary
- The physical-iOS direct-capture path ran `devicectl device screenshot --device <id> <path>`, which current Xcode `devicectl` rejects with `Unknown option '--device'`, forcing every physical-device screenshot into the slower XCTest runner fallback.
- Xcode's `devicectl` now requires `devicectl device capture screenshot --device <id> --destination <path>`. Updated the invocation in [physical-device-screenshot.ts](src/platforms/apple/core/physical-device-screenshot.ts) accordingly.

Fixes #1760

## Test plan
- [x] Added a regression test asserting the exact `devicectl` args (`device capture screenshot --device <id> --destination <path>`) — [physical-device-screenshot.test.ts](src/platforms/apple/core/__tests__/physical-device-screenshot.test.ts)
- [x] `npx vitest run src/platforms/apple/core/__tests__/` — 504/504 passing
- [x] `npx tsc --noEmit` — clean
- [x] `npx oxlint` on changed files — clean
- [x] Live verification on a connected, unlocked physical iPhone (`thymikee-iphone`): `agent-device screenshot` produced a valid 1206x2622 PNG via the direct `devicectl` path with no XCTest fallback triggered

### DIFF
--- a/src/platforms/apple/core/__tests__/physical-device-screenshot.test.ts
+++ b/src/platforms/apple/core/__tests__/physical-device-screenshot.test.ts
@@ -1,7 +1,47 @@
 import { test } from 'vitest';
 import assert from 'node:assert/strict';
-import { shouldFallbackToRunnerForIosScreenshot } from '../physical-device-screenshot.ts';
+import { IOS_DEVICE } from '../../../../__tests__/test-utils/index.ts';
+import {
+  captureCoreDeviceScreenshot,
+  shouldFallbackToRunnerForIosScreenshot,
+} from '../physical-device-screenshot.ts';
+import { createLocalAppleToolProvider, withAppleToolProvider } from '../tool-provider.ts';
 import { AppError } from '@agent-device/kernel/errors';
+
+test('captureCoreDeviceScreenshot invokes devicectl with the current capture-screenshot syntax', async () => {
+  const calls: Array<{ cmd: string; args: string[] }> = [];
+
+  await withAppleToolProvider(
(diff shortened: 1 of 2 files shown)


thymikee:
Exact-head review at `9ac47ac5`: code-clean and ready for human review. The production CoreDevice path now uses the current `devicectl device capture screenshot --device … --destination …` syntax, the focused regression asserts the exact shipped invocation and prevents XCTest fallback, and live physical-iPhone evidence produced a valid 1206×2622 PNG through the direct path. Completed checks are green; remaining CI is still running. Applied `ready-for-human`.

PR #1593, opened 2026-08-04
fix: stop the unit suite from leaking temp directories

thymikee:
## Summary

Stop Vitest runs from leaking scratch directories into the host temp directory.

- Redirect every Vitest worker to one PID-tagged `/tmp/agent-device-test-run-*` directory and remove it in global teardown.
- Preserve the intentional Swift compiler cache outside that disposable directory, avoiding per-run AVFoundation recompilation and test-timeout regressions.
- Migrate the repository's raw test `mkdtemp` calls to the canonical `mkdtempForTest` helpers for one discoverable convention.
- Add a leak checker that ignores directories owned by live concurrent Vitest processes and reports abandoned runs.
- Add a real lifecycle regression that launches the configured Vitest project, verifies worker `TMPDIR` and Swift-cache routing, and proves teardown removed the run directory.

The branch is rebased onto current `main`. `node --test` lanes remain outside this Vitest-specific lifecycle and are tracked by #1595.

## Validation

- `pnpm check:affected --run` — all runnable checks passed; 5,620/5,620 Vitest tests passed.
- `pnpm check:tmpdir-leaks:test`
- `pnpm exec vitest run --project provider-integration test/integration/provider-scenarios/android-recording.test.ts`
- Counterfactuals: removing the configured global setup fails on the worker `TMPDIR` assertion; disabling teardown fails on the surviving run directory; removing Swift-cache preservation fails on the cache-boundary assertion and reproduced the Android recording timeout.

PR #1770, opened 2026-08-12
fix(ci): run the three checks no workflow was running

thymikee:
Step 1 of the landing plan agreed on #1753: the defect fix on its own, with none of the machinery.

**23 lines of `ci.yml`. No new code, no new scripts, no new dependencies.**

## The defect

Three real package scripts with real assertions that **no CI lane executes**:

| Script | What it asserts |
| --- | --- |
| `check:tmpdir-leaks` | no abandoned `agent-device-test-run-*` directory survives a run whose owning process died (#1593/#1595) |
| `check:tmpdir-leaks:test` | the redirection model itself — liveness detection, the Vitest globalSetup path, and the `node --test` wrapper |
| `test:fixture-cache` | the trusted-artifact contract the device lanes use to decide a cached fixture app matches this commit |

All three are reachable only through `check:unit`, an aggregate no workflow invokes. A regression in any of them could not fail a PR.

Verified by transitive script reachability over every workflow and composite action on `main`, plus direct grep: **zero references to any of the three anywhere in `.github`.** That is how they stayed hidden — a check that silently stops running looks exactly like a green build.

## Placement, and one deliberate exception

- **`check:tmpdir-leaks` → Coverage.** The lane whose instrumented suite is what would leak a run directory.
- **`test:fixture-cache` → Integration Tests.** Beside the fixture-app fallback smoke, which covers the same artifact contract.
- **`check:tmpdir-leaks:test` → Layering Guard**, *not* next to the check it covers. `vitest-tmpdir-global-setup.test.ts` proves the lifecycle by spawning a real nested `vitest run`, and the Coverage lane already loses runs to `[vitest-pool]: Worker forks emitted error` when a fork is slow to terminate. Starting a nested Vitest seconds before the full instrumented suite is a contention risk with nothing to gain.

## Evidence

**All three pass at this commit**, so this wires green gates rather than red ones:

```
check:tmpdir-leaks         PASS
check:tmpdir-leaks:test    PASS
test:fixture-cache         PASS
```

**Non-vacuity.** Planting an abandoned run directory makes the check fail and name it:

```
$ mkdir /tmp/agent-device-test-run-planted-99999999
$ pnpm check:tmpdir-leaks
Error: Found 1 abandoned agent-device-test-run-* directory in /tmp:
$ rmdir /tmp/agent-device-test-run-planted-99999999
$ pnpm check:tmpdir-leaks
No abandoned agent-device-test-run-* directories found in /tmp.
```

**Wiring verified structurally** rather than by eye — each target job parsed back out of the YAML and asserted to contain the exact command. That check caught a real mistake while writing this: my first attempt missed the Coverage step entirely, because an `env:` block sits between the step name and its `run:`.

## Scope

Part of #1429, which stays open. The ownership proof that would catch this class *automatically* is the subject of #1753 and is being redesigned after review — this PR deliberately carries none of it, so the immediate risk is removed now rather than waiting on that.

https://claude.ai/code/session_01SkS4S8XXrfkJ8TD1VBKkvJ

---
_Generated by [Claude Code](https://claude.ai/code/session_01SkS4S8XXrfkJ8TD1VBKkvJ)_

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -116,6 +116,17 @@ jobs:
         # rather than in its own job so the two can never be green independently.
         run: pnpm depgraph:test
 
+      # Tests for the TMPDIR redirection itself, hidden the same way as the check above.
+      #
+      # Deliberately NOT in Coverage next to `check:tmpdir-leaks`, where the subject matter
+      # would put it: vitest-tmpdir-global-setup.test.ts proves the lifecycle by spawning a
+      # real nested `vitest run`, and the Coverage lane already loses runs to
+      # `[vitest-pool]: Worker forks emitted error` when a fork is slow to terminate.
+      # Starting a nested Vitest seconds before the full instrumented suite is a contention
+      # risk with nothing to gain.
+      - name: Check the tmpdir redirection model
+        run: pnpm check:tmpdir-leaks:test
+
   affected-selector:
     name: Affected-check Selector
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


thymikee:
Exact-head review at `8f84232c`: code-clean and ready for human review. This is the right small first landing from #1753: it wires the three confirmed orphan checks directly into semantically appropriate existing jobs, adds no ownership parser or new machinery, and keeps #1429 open. Coverage owns the post-suite leak assertion; Layering runs the nested-Vitest redirection-model suite away from the already contention-sensitive coverage process; Integration owns the trusted fixture-artifact contract. The planted abandoned-directory evidence is non-vacuous. Completed checks are green and remaining exact-head CI is in progress. Applied `ready-for-human`.

Issue #1545, opened 2026-08-01
Per-worktree dev state-dir spawns a fresh daemon per CLI invocation


thymikee:
Found while producing #1535's live repair evidence; reproduces on current main, so unrelated to the #1478 stack.

In a git worktree using the default per-worktree dev state dir (`~/.agent-device/dev/<worktree>-<hash>`), each CLI invocation spawned a NEW daemon instead of connecting to the existing one — three coexisting `dist/src/internal/daemon.js` processes were observed against one state dir within two minutes (ages 2:35 / 2:18 / 1:28), each command seeing an empty session store. Practical consequence: any multi-command session flow silently breaks with SESSION_NOT_FOUND — a held repair session (`repairSessionHeld: true`) is unreachable from the very next command, which presents exactly like a held-session regression and cost real diagnosis time. The CLI also intermittently printed "Replacing daemon (pid …): code-signature mismatch" in these worktrees earlier the same day, suggesting the daemon identity/signature check misbehaves for worktree dev builds (mismatch on every connect → spawn-instead-of-connect, without killing the predecessor).

Workaround: pass an explicit shared `--state-dir` on every command of the flow (all such flows worked flawlessly).

Repro sketch: in a worktree, `pnpm build`, then run any two session commands seconds apart without `--state-dir` (e.g. `replay <flow> --save-script --session s` then `click … --session s`), and check `ps` for multiple daemon processes plus SESSION_NOT_FOUND on the second command.

thymikee:
Live data point on the severity of this leak: a single working session in a worktree (running CLI invocations without an explicit `--state-dir`) accumulated **13 daemon processes** over ~16 hours — one long-lived daemon held the shared iPhone 17 Pro simulator's runner lease the entire time (forcing three separate evidence runs on this host onto substitute simulators), plus 12 per-invocation daemons that never exited. All were orphans of a finished session and had stale metadata (`daemon stop --state-dir <their dir> --clean` reported "No running daemon found" while the processes lived on), so even deliberate cleanup required a manual `pkill`. Two aggravating factors worth fixing together: (1) the per-invocation spawn itself (this issue), and (2) daemons whose metadata can go stale while the process survives, making them invisible to their own lifecycle tooling.

_Generated by [Claude Code](https://claude.ai/code)_

PR #1771, opened 2026-08-12
fix: fingerprint the built daemon's real import graph (#1545)

thymikee:
## Summary

- `computeDaemonCodeSignature`'s import-detection regex required whitespace around `from` and after `import`/`export`. The daemon entry that actually ships/runs (`dist/src/internal/daemon.js`) is a minified bundle (tsdown/rolldown `minify: true`), so real statements have zero whitespace, e.g. `import{o as e}from"../foo.js"`. The regex never matched, so the graph walk silently degraded to fingerprinting *only the entry file* instead of its real dependency graph.
- Verified directly against this repo's own build: `computeDaemonCodeSignature` on `dist/src/internal/daemon.js` returned `graph:1:...` before this fix and `graph:50:...` (its real transitive chunk count) after. Source-mode (`--experimental-strip-types`, unminified) fingerprinting is unaffected (`graph:741:...` in both cases).
- This directly matches what #1545 flags as a suspected cause: *"the daemon identity/signature check misbehaves for worktree dev builds"* and the `Replacing daemon (pid …): code-signature mismatch` messages observed there — a signature that only ever reflects the entry file's own size/mtime is a much weaker (and differently-behaved) safety net than the one the check was designed to provide, and int the extreme lets a daemon started from one build look identical to a client on a materially different build purely by mtime/size coincidence on that one file.
- New regex keeps a boundary check (`(?![\w$])`) after the `import`/`export` keyword so it doesn't false-positive on identifiers like `importantValue` that merely start with those words — covered by a new test case alongside the existing (unminified) fingerprinting test.

## Scope note

This fixes a concrete, reproducible defect in the fingerprinting mechanism itself, and I verified end-to-end (rebuilding this worktree's `dist/`, forcing a stale signature, running concurrent cold-start invocations) that takeover + predecessor termination behave correctly with the fix in place. I was not able to locally reproduce the broader "daemon never dies, metadata goes stale" orphan pileup described in the issue's follow-up comment (13 daemons over 16h) — that may need separate investigation with `--debug` diagnostics from an environment where it's actively recurring, since I could not force it under either a forced-mismatch or concurrent-cold-start scenario in this sandbox.

## Test plan
- [x] `vitest run --project unit-core src/utils/__tests__/daemon-client.test.ts src/utils/__tests__/daemon-client-lifecycle.test.ts` (new test included)
- [x] `vitest run --project unit-core src/daemon` (full daemon suite, 2037 tests)
- [x] `oxlint . --deny-warnings` / `oxfmt --check` on touched files
- [x] Live: `pnpm build` + two consecutive `agent-device session` invocations reuse the same daemon pid with the new `graph:50:...` signature; a forced stale `codeSignature` correctly triggers "Replacing daemon" and cleanly terminates the predecessor

### DIFF
--- a/src/daemon/code-signature.ts
+++ b/src/daemon/code-signature.ts
@@ -3,8 +3,15 @@ import fs from 'node:fs';
 import path from 'node:path';
 import { findProjectRoot } from '../utils/version.ts';
 
+// `\s*` (not `\s+`) around each optional part: the built daemon entry is a
+// minified bundle (tsdown/rolldown `minify: true`), so real import/export
+// statements have zero whitespace around `from` and even after the keyword
+// itself (e.g. `import{o as e}from"../foo.js"`). `\s+` here would never match
+// minified output, silently degrading the whole graph walk to just the entry
+// file (see #1545). The `(?![\w$])` boundary after the keyword still rejects
+// identifiers like `importantValue` that merely start with "import".
 const STATIC_IMPORT_RE =
-  /(?:^|[^\w$.])(?:import|export)\s+(?:type\s+)?(?:[^'"`]*?\s+from\s+)?['"]([^'"]+)['"]/gm;
+  /(?:^|[^\w$.])(?:import|export)(?![\w$])\s*(?:type\s+)?(?:[^'"`]*?\s*from\s*)?['"]([^'"]+)['"]/gm;
 const DYNAMIC_IMPORT_RE = /import\(\s*['"]([^'"]+)['"]\s*\)/gm;
 const RESOLVABLE_EXTENSIONS = ['.ts', '.tsx', '.js', '.jsx', '.mjs', '.cjs'] as const;
 
--- a/src/daemon/code-signature.ts
+++ b/src/daemon/code-signature.ts
@@ -3,16 +3,18 @@ import fs from 'node:fs';
 import path from 'node:path';
 import { findProjectRoot } from '../utils/version.ts';
 
-// `\s*` (not `\s+`) around each optional part: the built daemon entry is a
-// minified bundle (tsdown/rolldown `minify: true`), so real import/export
-// statements have zero whitespace around `from` and even after the keyword
-// itself (e.g. `import{o as e}from"../foo.js"`). `\s+` here would never match
-// minified output, silently degrading the whole graph walk to just the entry
-// file (see #1545). The `(?![\w$])` boundary after the keyword still rejects
-// identifiers like `importantValue` that merely start with "import".
-const STATIC_IMPORT_RE =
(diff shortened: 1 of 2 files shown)


thymikee:
Pushed a follow-up commit after re-reviewing the first one against a stricter code-quality bar.

**What changed:** the original fix widened `\s+` to `\s*` and added a keyword-boundary lookahead to the existing `import`/`export`/`from`-anchored regex. That's correct, but it's still tracking the exact whitespace/keyword *shape* bundlers emit around a specifier — which is precisely the kind of assumption that just broke once (formatted source vs. minified build). Replaced it with a single regex that matches any quoted, relative-path-shaped string literal, with no `import`/`export`/`from` awareness at all: static imports, re-exports, dynamic `import()`, and minified output all put the specifier in the same place (a `'./...'`/`"../..."` string literal), so matching on that directly is simpler *and* immune to whatever the next formatting difference turns out to be. This also let me delete the separate dynamic-import regex and the two-regex collection helper entirely.

Safety check: an over-broad match (e.g. a `./`-looking string inside a comment) can't corrupt the graph — `resolveRelativeImportPath` just fails to resolve it to a real file and it's dropped. Verified this empirically against both this repo's actual `dist/` build and its full `src/` tree: identical file counts (`graph:50` / `graph:741`) before and after the simplification, with the only "extra" matches being harmless non-resolving strings (e.g. a `../foo.js` example inside a comment in `code-signature.ts` itself).

Net: `code-signature.ts` went from 105 → 100 lines despite the added explanatory comment, one regex instead of two, no dual-collection helper. New test added for the "harmless over-match" case specifically.

One thing I did *not* touch: `src/utils/__tests__/daemon-client.test.ts` is ~1790 lines, well past a healthy single-file size — but that's pre-existing and not something this diff's ~40 added lines should take on decomposing. Flagging it as a good candidate for a future split (e.g. pulling the `computeDaemonCodeSignature`/`stopDaemonProcessForTakeover` tests into a dedicated `code-signature.test.ts`), not blocking this PR.

thymikee:
Exact-head review at `6e3c1eec`: code-review clean. The change is narrowly scoped and fixes the concrete shipped-bundle defect: minified relative imports now contribute to the daemon signature (`graph:1` → the verified 50-file built graph), while source-mode graph size remains unchanged. The conservative relative-literal scan is acceptable here because candidates must resolve to supported module files, and the measured built/source graphs match the grammar-aware implementation; the regression covers minified import, re-export, side-effect import, dependency mutation, and an unresolved non-import literal. The PR correctly keeps #1545 open for the broader daemon-pileup investigation it could not reproduce. All completed authoritative checks are green; Android/iOS Smoke, Coverage, and FreeRange are still in progress. Code-ready pending those checks.

PR #1773, opened 2026-08-13
fix: preserve compact snapshot action labels

thymikee:
## Summary

Preserve app-owned action labels that compact snapshots previously dropped on iOS and Android.

- iOS promotes a cell’s descendant title when the cell exposes only an implementation or system scroll-indicator label, while preserving legitimate cell labels.
- Android retains passive Compose `View` semantics under hittable containers and promotes only labeled passive leaf views, avoiding independently actionable subtrees.
- Scope is intentionally cross-platform because the benchmark exposed the same public `snapshot -i` contract failure through different native accessibility shapes. Nine files are touched.

## Validation

- Red proof: focused regressions failed 2/78 on Android and returned no labeled iOS room cells before the fixes.
- `pnpm check:affected --run`: 316 test files / 2,745 tests passed; changed-line coverage 92.59%; format, lint, typecheck, layering, fallow, build, and provider checks passed.
- Live iOS Element Classic: `Book Club` and `Team Standup` are labeled `Cell` actions; clicking the promoted `Book Club` ref opened the room.
- Live Android Element X: compact helper output labels `Back`, `Start a call`, and `User menu` from app-owned Compose semantics.
- `pnpm build:android` passed.

Docs and skills are unchanged because this repairs existing snapshot semantics without changing commands, flags, or workflow guidance.

### DIFF
--- a/src/daemon/snapshot-presentation/ios/index.ts
+++ b/src/daemon/snapshot-presentation/ios/index.ts
@@ -13,9 +13,9 @@ const IOS_PRESENTATION_RULES: Array<
 > = [
   // Semantic representatives must be collected before duplicate-label suppression.
   collectIosWebSemanticPresentation,
-  collectIosPresentationNoiseSuppression,
   collectIosImplicitScrollableActions,
   collectIosRowPresentation,
+  collectIosPresentationNoiseSuppression,
 ];
 
 export function presentIosInteractiveSnapshot(nodes: RawSnapshotNode[]): RawSnapshotNode[] {
--- a/src/daemon/snapshot-presentation/ios/rows.fixtures.ts
+++ b/src/daemon/snapshot-presentation/ios/rows.fixtures.ts
@@ -1,0 +1,51 @@
+import type { RawSnapshotNode } from '@agent-device/kernel/snapshot';
+
+const roomRowRect = { x: 0, y: 400, width: 402, height: 96 };
+
+export const elementClassicRoomListNodes: RawSnapshotNode[] = [
+  { index: 0, depth: 0, type: 'Application', label: 'Element Classic' },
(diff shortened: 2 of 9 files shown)


thymikee:
isImplementationCellLabel() is too broad: any PascalCase app-owned label ending in Cell (for example StemCell or a product name) is treated as leaked implementation text, replaced with the first descendant title, and that descendant is suppressed. The new Stem Cell preservation test misses this because its space avoids the regex. Please constrain promotion to a reliable implementation signal (for example, the class-shaped label matching the cell identifier, or another narrowly evidenced UIKit pattern) and add a no-space legitimate-label regression with red-before proof. Also ensure the narrowed rule still covers the observed RecentTableViewCell + scroll-indicator case. Exact head 18ec27ec; all checks are green, but this behavior finding blocks ready-for-human. No label.

thymikee:
Addressed in 736013621. `isImplementationCellLabel` now requires the class-shaped label to exactly match the cell identifier, so arbitrary app labels such as `StemCell` are preserved. I added a red-before regression (the old implementation incorrectly produced `Research details`) and retained coverage for the observed `RecentTableViewCell` + `Vertical scroll bar, 1 page` shape. `pnpm check:affected --run` passed before the push.

thymikee:
Exact head `736013621` is code-review clean. The prior blocker is resolved: iOS row promotion now requires the class-shaped label to exactly match the cell identifier, preserving legitimate no-space app labels such as `StemCell` while retaining the observed `RecentTableViewCell` + scroll-indicator behavior. The regression is load-bearing (the prior matcher changes `StemCell` to `Research details`). All exact-head authoritative checks are green, including Integration, Coverage, iOS/Android smoke, static gates, and mutation selection. CLEAN/MERGEABLE; ready for human review.

PR #1776, opened 2026-08-15
fix(ios): preserve final-probe xcodebuild diagnostics

### LINKED ISSUE #1775: iOS: final runner probe can replace an xcodebuild launch failure with a generic connect error
### Summary

On a physical iOS device, `waitForRunner` can return `Runner did not accept connection` when `xcodebuild` exits while the final runner endpoint probe is in flight. The actionable XCTest launch failure remains in `runner.log`, but it is absent from the returned error.

I observed this with an untrusted Developer App certificate, then reproduced the timing deterministically against current `main` without a device. In that window, the generic connect error also allows the prepare health path to start another runner even though the launch blocker is deterministic.

### Physical-device observation

Environment:

- agent-device `0.20.6`
- macOS `26.5.2` (`25F84`)
- Xcode `26.5` (`17F42`)
- Node.js `24.16.0`
- physical iPhone on iOS `18.7.9`

The relevant agent-device invocation was:

```sh
agent-device prepare ios-runner \
  --platform ios \
  --udid <physical-device-udid> \
  --timeout 240000
```

`runner.log` recorded the terminal cause:

```text
Finished with error: The application could not be launched because the Developer App Certificate is not trusted.
...
Testing failed:
    The application could not be launched because the Developer App Certificate is not trusted.
```

The request diagnostic instead ended with:

```text
Runner did not accept connection
```

The prepare health path then started a second runner, which failed with the same certificate-trust error before the request again returned the generic connect error.

Trusting the certificate resolved the physical-device blocker. The concern here is not that agent-device should bypass that requirement; it is that the existing actionable Xcode diagnostic was replaced at the command boundary.

### Deterministic reproduction on current main

I added a focused regression test to `runner-transport.test.ts` at current `main` commit `5b1b1aad2eccd299d266315b438893b5e6ae5fb0`.

The test starts with a live-looking runner child, makes the final usbmux probe set `child.exitCode = 65` and fail with `ECONNREFUSED`, and gives `testPromise` the captured certificate-trust diagnostic. It expects the existing early-exit error and nested Xcode output:

<details>
<summary>Regression test</summary>

```ts
test('waitForRunner preserves xcodebuild diagnostics when the runner exits during the final probe', async () => {
  const session: RunnerSession = {
    sessionId: 'starting-device-session',
    device: xctestIosDevice,
    deviceId: xctestIosDevice.id,
    port: 8100,
    xctestrunPath: '/tmp/runner.xctestrun',
    jsonPath: '/tmp/runner.json',
    testPromise: Promise.resolve({
      exitCode: 65,
      stdout: '',
      stderr:
        'The application could not be launched because the Developer App Certificate is not trusted.',
    }),
    child: { pid: 1234, exitCode: null } as ExecBackgroundResult['child'],
    ready: false,
  };
  mockUsbmuxPostCommand.mockImplementation(async () => {
    (session.child as { exitCode: number | null }).exitCode = 65;
    throw new Error('ECONNREFUSED');
  });

  await assert.rejects(
    () =>
      waitForRunner(
        xctestIosDevice,
        8100,
        { command: 'uptime' },
        '/tmp/runner.log',
        100,
        session,
      ),
    (error: unknown) => {
      const appError = error as AppError;
      assert.equal(appError.message, 'Runner did not accept connection (xcodebuild exited early)');
      assert.match(
        String((appError.details?.xcodebuild as { stderr?: string } | undefined)?.stderr),
        /Developer App Certificate is not trusted/,
      );
      return true;
    },
  );
});
```

</details>

Command:

```sh
pnpm exec vitest run --project unit-core \
  src/platforms/apple/core/__tests__/runner-transport.test.ts
```

Result on unmodified current `main`:

```text
+ actual - expected

+ 'Runner did not accept connection'
- 'Runner did not accept connection (xcodebuild exited early)'

Test Files  1 failed (1)
Tests       1 failed | 14 passed (15)
```

As a mechanism check only, rechecking the runner child after retry completion made this focused suite pass 15/15. I reverted that probe and have not opened a PR; the appropriate precedence and final implementation are for maintainer triage.

### Current control flow

`ensureRunnerAttemptCanStart` checks `session.child.exitCode` before each route probe. If the child exits during the final probe:

1. the route records its endpoint error and returns no response;
2. the retry loop exhausts;
3. `waitForRunner` leaves the loop;
4. the physical-device path calls `buildRunnerConnectError` without checking the child again.

That produces the deterministic test result above. The runner error table can classify an early-exit error once it receives one, but this timing window does not build that error.

### Expected behavior

If the XCTest process has exited before runner connection returns, preserve the existing `Runner did not accept connection (xcodebuild exited early)` error and its nested `xcodebuild` output, reason, and hint. A known terminal launch failure should not be replaced by an endpoint error or trigger another runner-health attempt.

### Related work

- #1529 added classification for actionable physical-runner installation failures, but this timing window prevents the Xcode output from reaching the early-exit classification path.
- #1631 / #1644 centralized runner error classification and recovery coverage; current `main` still reproduces this post-probe exit window.
- I found no open issue or PR matching `xcodebuild exited early`, `Runner did not accept connection`, or `Developer App Certificate`.

If this behavior and scope look right, I would be happy to prepare the focused regression and fix after maintainer triage.

emmanuelchucks:
## Summary

- Preserve the established xcodebuild early-exit error when a physical iOS runner exits during the final connection probe, instead of replacing its actionable diagnostics with `Runner did not accept connection`.
- Keep request-cancellation and usbmux-unattached errors ahead of the terminal child-exit check; simulator fallback is unchanged.
- Separate startup/readiness transport from one-shot command transport and mirror that boundary in the focused tests.

Closes #1775.

## Validation

- Regression proof: removing the terminal child-exit recheck from the current startup module produces 1 failed / 8 passed, with `Runner did not accept connection` received instead of the xcodebuild early-exit error.
- Focused startup and one-shot transport suites pass 15/15. Their split preserves all 15 test cases, and a planted type error confirmed the new test module is included by TypeScript.
- `pnpm check:affected --run` on Node 22.22.0 passes 274 files and 2,225 tests; changed-line coverage is 92/118 (77.97%) and changed-branch coverage is 64/87 (73.56%). Package verification, formatting, lint, typecheck, layering, Fallow, build, affected coverage, and provider integration pass.
- The deterministic unit bundle passes 836 files and 6,557 tests. Static smoke coverage passes 38 tests with the three documented live-device/browser tests skipped.

The physical-device reproduction is recorded in #1775. This PR changes terminal error precedence without manufacturing another certificate-trust failure on a device.

Scope: 13 files, confined to Apple runner transport, mirrored tests, shared transport fixtures, and the moved Fallow baseline entry. No CLI, wire, Swift runner, documentation, or skill changes.


thymikee:
Exact head `4b4b633d` fixes the reported race correctly: cancellation and usbmux-unattached errors retain precedence, simulator fallback is unchanged, and the physical-device terminal path now uses the established early-exit builder. The planted regression is non-vacuous and preserves exit code 65 plus the actionable certificate-trust stderr.

[P2 topology] `src/platforms/apple/core/runner/runner-transport.ts` was already 519 LOC on the base and is 522 LOC here. AGENTS.md’s implementation tripwire requires extraction before adding behavior past 500 LOC. Please move the startup/readiness connection loop (`waitForRunner` and its attempt helpers) into a focused runner-startup transport module, mirror the source/test topology, and retain this regression; keep the one-shot command transport in the existing focused module.

CI is also owner-blocked rather than green: all eight exact-head Actions runs are `action_required`, so a maintainer must approve them after the topology fix. No ready-for-human label while the code finding and authoritative CI blocker remain.

emmanuelchucks:
Addressed in `c634be380`.

- `waitForRunner` and its attempt helpers now live in `runner-startup-transport.ts` (440 LOC); `runner-transport.ts` retains the one-shot command path (93 LOC).
- The tests mirror that split with 9 startup and 6 one-shot cases, shared fixtures, and the final-probe regression retained. The Fallow baseline entry moved with the extracted code.
- Focused transport tests pass 15/15. `pnpm check:affected --run` passes 274 files / 2,225 tests with 77.97% changed-line coverage; the deterministic unit bundle passes 836 files / 6,557 tests.

The PR description now reflects the current head. This head is ready for CI approval.

thymikee:
Exact head `c634be380` is code-review clean. The topology blocker is resolved: startup/readiness transport is now isolated in a cohesive 440-LOC module, the one-shot transport is 93 LOC, production imports and the Fallow baseline moved coherently, and the tests mirror the split with shared fixtures. The terminal child-exit recheck and its load-bearing race regression remain intact; cancellation/usbmux precedence and simulator fallback are unchanged.

The only current blocker is owner-action CI: all eight exact-head Actions runs concluded `action_required`, so a maintainer must approve them and obtain authoritative results. CLEAN/MERGEABLE at the code level, but no `ready-for-human` label until those workflows run without a confirmed failing/blocked CI state.

thymikee:
Exact head c634be380 remains code-review clean, and the prior owner-action CI blocker is resolved. All applicable authoritative checks are now green, including Integration Tests, Coverage, static gates, mutation selection, and iOS/Android/macOS/Linux smoke; only intentionally skipped jobs remain. The PR is CLEAN/MERGEABLE and ready for human review.

PR #1784, opened 2026-08-17
fix: stabilize private AX settle snapshots

thymikee:
## Summary

Project private-AX snapshots to the visible viewport, retaining excluded content only as scroll hints.

Stabilize settle observations by pinning private AX after fallback and comparing the visible semantic projection with bounded geometry tolerance. Geometry-less semantic nodes remain visible to agents but cannot become interaction targets.

This touches 8 files across the iOS private-AX fallback and the shared settle runtime; scope stayed within snapshot capture and settle behavior. No public CLI or skill behavior changed.

## Validation

- `pnpm check:affected --run` passed: 1,942 affected tests and 100% changed-line coverage.
- Built the iOS runner with unit tests enabled and ran the two focused private-AX XCTest regressions twice: 2/2 passed with zero failures.
- Proved the settle backend-pinning regression red by temporarily removing the production forwarding, then restored the fix and verified green.
- Deleted the disposable verification simulators after the run.

### DIFF
--- a/.github/workflows/ios.yml
+++ b/.github/workflows/ios.yml
@@ -113,6 +113,8 @@ jobs:
             -only-testing:AgentDeviceRunnerUITests/RunnerTests/testPrivateAXDepthLimitedRequiresEveryFrontierResolved \
             -only-testing:AgentDeviceRunnerUITests/RunnerTests/testDeepExtensionCountsMissedFrontiers \
             -only-testing:AgentDeviceRunnerUITests/RunnerTests/testPreferredPrivateAXBackendPlansAsPenalized \
+            -only-testing:AgentDeviceRunnerUITests/RunnerTests/testPrivateAXRegularPresentationProjectsToViewportAndKeepsScrollHint \
+            -only-testing:AgentDeviceRunnerUITests/RunnerTests/testPrivateAXGeometrylessSemanticsAreNeverActionableOrScrollContexts \
             -only-testing:AgentDeviceRunnerUITests/RunnerTests/testDecodedPreferredBackendReachesOptionsAndApplicablePlan \
             -only-testing:AgentDeviceRunnerUITests/RunnerTests/testSparsePayloadReasonMatrix \
             -only-testing:AgentDeviceRunnerUITests/RunnerTests/testRunnerScreenshotStabilitySettledNeedsEnoughSamples \
--- a/.github/workflows/ios.yml
+++ b/.github/workflows/ios.yml
@@ -115,6 +115,8 @@ jobs:
             -only-testing:AgentDeviceRunnerUITests/RunnerTests/testPreferredPrivateAXBackendPlansAsPenalized \
             -only-testing:AgentDeviceRunnerUITests/RunnerTests/testPrivateAXRegularPresentationProjectsToViewportAndKeepsScrollHint \
             -only-testing:AgentDeviceRunnerUITests/RunnerTests/testPrivateAXGeometrylessSemanticsAreNeverActionableOrScrollContexts \
+            -only-testing:AgentDeviceRunnerUITests/RunnerTests/testPrivateAXScopeSelectsSubtreeNotMatchingLabels \
+            -only-testing:AgentDeviceRunnerUITests/RunnerTests/testPrivateAXInteractiveFiltersLoginLikeHiddenDrawer \
             -only-testing:AgentDeviceRunnerUITests/RunnerTests/testDecodedPreferredBackendReachesOptionsAndApplicablePlan \
             -only-testing:AgentDeviceRunnerUITests/RunnerTests/testSparsePayloadReasonMatrix \
             -only-testing:AgentDeviceRunnerUITests/RunnerTests/testRunnerScreenshotStabilitySettledNeedsEnoughSamples \
--- a/apple/runner/AgentDeviceRunner/AgentDeviceRunnerUITests/RunnerTests+AXSnapshotFallback.swift
+++ b/apple/runner/AgentDeviceRunner/AgentDeviceRunnerUITests/RunnerTests+AXSnapshotFallback.swift
@@ -175,15 +175,10 @@ extension RunnerTests {
 
       let rootFrame = privateAXRect(root["frame"])
       let viewport = privateAXSnapshotViewport(app: app, rootFrame: rootFrame)
-      var nodes: [SnapshotNode] = []
-      appendPrivateAXNode(
-        root,
-        to: &nodes,
+      let nodes = privateAXPresentation(
+        rawRoot: root,
         options: options,
-        viewport: viewport,
-        depth: 0,
-        parentIndex: nil,
-        insideMatchedScope: false
+        viewport: viewport
       )
       if nodes.count <= 1 {
         NSLog("AGENT_DEVICE_RUNNER_PRIVATE_AX_SNAPSHOT_SPARSE=%ld", nodes.count)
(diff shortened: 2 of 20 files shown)


thymikee:
Review findings for 4959975:

1. **P1 — daemon/CLI path drops the private-AX pin.** `snapshotFlagOverrides()` in `src/daemon/selector-runtime-backend.ts` does not copy `BackendSnapshotOptions.preferredBackend` to the already-supported internal `snapshotPreferredBackend` flag. As a result, daemon/CLI `press --settle` does not reach the runner with the private-AX pin. The added direct-runtime test does not exercise this production route. Please add the mapping and a daemon-route regression.

2. **P2 — ordinary captures lost their prior geometry tolerance.** The new stable signal compares non-private snapshot rect floats with tolerance `0`, replacing the old `Math.round` digest. Fractional regular-backend geometry jitter can therefore keep `wait stable`/`--settle` from settling. Preserve the prior rounded/equivalent comparison and add a regular-backend regression.

CI red checks are external codeload 429/5xx download failures, but live private-AX device settle evidence is still absent: the reported XCTest coverage uses synthetic dictionaries rather than a captured private-AX settle scenario.

thymikee (commented):
## Verdict: request changes — the fix direction is right; the implementation re-invents three things the codebase already owns and adds a backend-conditional comparator to a shared cross-platform loop.

The PR does two things: (1) makes private-AX regular captures viewport-projected with scroll hints (like the XCTest tree path), and (2) makes the settle loop pin the backend and compare a tolerant, viewport-filtered signal. Both are correct goals. The problems are structural.

---

### 1. Swift: the private-AX presentation copy-pastes the XCTest tree path instead of reusing it

`RunnerTests+PrivateAXPresentation.swift` re-implements, near-verbatim, four helpers that already exist as `private` in `RunnerTests+Snapshot.swift`:

| New inline code | Existing canonical helper |
|---|---|
| hint bookkeeping (`PrivateAXPresentation.swift:45-50`) | `rememberHiddenContentHint(for:relativeTo:hints:)` (`Snapshot.swift:1203`) |
| the `nodes.map { … SnapshotNode(…) }` hint application (`PrivateAXPresentation.swift:11-19`) | `applyHiddenContentHints(_:to:)` (`Snapshot.swift:1220`) — needs one line to forward `actions` |
| `(!hasFrame \|\| isVisibleInViewport(rect, viewport)) && (!hasFrame \|\| scrollContext.map{…} ?? true)` (`PrivateAXPresentation.swift:34-35`) | `isVisibleInRegularSnapshot(_:viewport:scrollContainerAnchor:)` (`Snapshot.swift:1166`) |
| `Self.scrollContainerTypes.contains(type)` scroll-context rule (`PrivateAXPresentation.swift:65-68`) | `isScrollableContainer` / `scrollContainerAnchor` (`Snapshot.swift:1193`) — this one legitimately needs a type-based twin, but the rule should be shared; note the XCTest one also requires non-empty children |

The stated goal is XCTest ↔ private-AX equivalence. Two copies of "how does a scroll container project and hint" is exactly how they drift again. Drop `private` on those three helpers, forward `actions` in `applyHiddenContentHints`, and the new file collapses to the traversal only. The `visible` expression then reads `!hasFrame || isVisibleInRegularSnapshot(rect, viewport: viewport, scrollContainerAnchor: scrollContext)` — and I'd name the two concepts apart (`onScreen = hasFrame && …`, `include = isRoot || !hasFrame || onScreen`) instead of one `visible` var patched with `hasFrame` in three places.

### 2. Swift: a third include-policy site, bolted on outside the filter

`PrivateAXPresentation.swift:42`: `let include = decision.include && (parentIndex == nil || visible)`. `flatSnapshotFilterDecision` already models `isRoot` and `visible` and already has an interactiveOnly-visibility rule (`FlatSnapshotFiltering.swift:42`) — which this bolt-on now subsumes for private-AX callers, making that rule dead on this path. Result: include-policy lives in `shouldInclude` (XCTest tree), `flatSnapshotFilterDecision` (flat XCTest + private-AX), *and* this caller. Put the projection rule inside the filter. Note the flagless version (`!isRoot && !visible → exclude`, universally) would also bring the flat XCTest fallback (`Snapshot.swift:1477`) in line with the tree path — an equivalence win, but a behavior change to validate. Either way: one policy site, not three.

### 3. Swift: test-only wrapper kept in production code

`PrivateAXPresentation.swift:74-82`: the 7-arg `appendPrivateAXNode` overload exists only so three old tests in `AXSnapshotFallback.swift` (L638, L668, L750) don't need editing — and it *behaves differently* from production (drops hints). Tests exercising a non-production entry point is the smell. Delete it; migrate the three tests to `privateAXPresentation(rawRoot:options:viewport:)`.

### 4. TS: backend-conditional comparator in a shared cross-platform loop, plus a silent regression

`stable-capture-signal.ts` branches on `privateAX` twice — the viewport filter and `tolerance = privateAX ? 4 : 0`. Two consequences:

- **Regression:** the deleted digest `Math.round`ed rects for every backend; the new comparator uses raw floats with tolerance 0 for XCTest/Android/macOS. Sub-pixel drift that rounding absorbed now resets the quiet window. Untested — the only test is the private-AX case.
- **Special-casing:** the flag makes `StableCaptureSignal` carry a backend bit and gives `stableCaptureSignalsEqual` a branch. Applying one rule uniformly (round + tolerance, filter or not) deletes the field, the branch, and the "why is this backend special" question. If you keep any tolerance, use one constant with a stated reason — `4` is unexplained, and the daemon's sibling comparator uses round-then-1px (`interaction-outcome-policy.ts:13`, `RECT_TOLERANCE_PX`).

Also: the runner now owns viewport projection for private-AX (that's what half this PR does), so the TS-side `isNodeVisibleOnScreen` filter is a *second* projection with a *different* rule (center-in-viewport vs. Swift's `intersects`). It has a real purpose (absorbs intersect-flips at the viewport edge), but that should be stated in a comment — as written it reads as belt-and-braces.

Type nit that's actually a boundary smell: `rect?: [number, number, number, number]` forces `candidate.rect![coordinate]!` in the comparator. Use the kernel `Rect` and both `!` disappear.

### 5. TS: indirection churn in stable-capture.ts

- `shouldResetPrivateAxRecoveryBudget(bool|undefined, bool, verdict)` (`stable-capture.ts:206`) moves a 3-clause condition into a helper with three positional params and makes the call site *longer*. Revert to the inline condition.
- `preferredStableCaptureBackend(verdict, current?)` (`stable-capture.ts:199`) is a latch: `preferredBackend ??= pinFor(verdict)`. Say that. And the "verdict.backend === 'private-ax' → 'private-ax'" derivation already exists at `interaction-ios-tap-outcome.ts:131` — same rule, two hand-rolled copies; a one-line helper next to `isSparseSnapshotQualityVerdict` in `snapshot-quality/verdict.ts` serves both.
- `lastDigest` now holds a `StableCaptureSignal`, not a digest.

### 6. Tests

- `stable-capture-signal.test.ts` is one assertion covering two mechanisms (offscreen churn + jitter) with no negative case and no non-private-AX case. Split it; add "real change → not equal" and one XCTest-shaped case.
- Both new Swift tests use `interactiveOnly: false`; the settle path uses `interactiveOnly: true`, and that's where geometry-less inclusion changed (previously excluded from `-i` private-AX captures, now included as non-hittable — which, for the record, *matches* XCTest `-i` via `hasContent`, so it's a step toward equivalence, but it's untested).

### Separate workstreams (not blocking this PR, but the PR is evidence they're due)

1. **One include-policy for all three iOS presentations** (XCTest tree `shouldInclude`, flat fallback, private-AX). Known remaining divergences: XCTest regular drops geometry-less nodes, private-AX regular keeps them; XCTest scroll anchors require non-empty children; XCTest `hittable` is occlusion-aware, private-AX is type-based.
2. **One snapshot comparator.** There are now three: `ax-digest` (multiset hash), `buildInteractionSurfaceSignature` + `areInteractionSurfaceSignaturesStable` (daemon, rounded + 1px, positional), and `stable-capture-signal` (commands, 4px, sorted). The daemon one is the most thought-through; it belongs in a layer below both `commands/` and `daemon/` (R2's own hint says so) so `runStableCaptureLoop` can use it.
3. **Backend flip policy.** The daemon quiet loop *rebases* on a backend flip (`post-gesture-stability.ts:129`); the settle loop now *pins*. Pinning is the better idea — the daemon loop could adopt it and delete its rebase branch.
4. `ios.yml` hand-enumerates `-only-testing` targets; a forgotten test silently doesn't run. Derive the list from `func test…` under `#if AGENT_DEVICE_RUNNER_UNIT_TESTS`.

### What's fine

- Pinning the settle loop's backend from the session's last verdict, and latching after the first private-AX capture — right idea, `settle.test.ts` covers it well.
- The `preferredBackend` plumbing through `captureSelectorSnapshot` follows the existing `includeHiddenContentHints` shape.
- No file crosses 1k lines; the new Swift test placement matches the codebase pattern.

thymikee:
Follow-up to my review above — I independently verified the P1 in the earlier findings comment, and it's real:

`snapshotFlagOverrides()` (`src/daemon/selector-runtime-backend.ts:165`) copies `interactiveOnly / scope / depth / raw / includeHiddenContentHints` from `BackendSnapshotOptions` into `snapshot*` flags and drops `preferredBackend`. The daemon-routed `press --settle` (the production path) therefore never delivers the pin to the runner; only the direct-runtime `createAgentDevice` path that `settle.test.ts` exercises does.

The structural point, which is why I'd fix it differently than "add the missing line": there are now **two hand-maintained mappings** from the same options shape to `snapshot*` flags — `snapshotFlagOverrides` here and the inline `...(options.preferredBackend ? { snapshotPreferredBackend } : {})` in `src/daemon/handlers/interaction-snapshot.ts:40`. A field added to one and forgotten in the other is exactly this bug, and it will recur for the next field. Make it one canonical `snapshotOptionsToFlags(options)` (contracts or `snapshot-quality/` level, below both callers) and have both sites use it — then `preferredBackend` can't be dropped by construction, and the daemon-route regression test asserts the runner request, not an intermediate object.

thymikee:
Addressed the review holistically at `d3f8412ad`:

- Consolidated private-AX presentation on the canonical XCTest visibility, scroll-anchor, and hidden-content-hint helpers; actions are preserved.
- Moved viewport inclusion into a typed flat-snapshot visibility policy and removed the test-only production overload.
- Added interactive-only Swift coverage and production-path TS regressions for both `wait stable` and `press --settle`.
- Replaced both hand-maintained option mappings with one type-exhaustive `BackendSnapshotOptions` → `CommandFlags` table. New routed option keys now fail typecheck until classified; `preferredBackend` and `customActions` are covered.
- Fixed the third dropped-pin route in `interaction-runtime.ts`.
- Unified settle geometry comparison on rounded 1 px tolerance, with private-AX jitter, regular-backend jitter, and real-change negative cases.
- Reused one verdict-to-backend-pin helper, renamed the signal state, and documented why the private-AX center projection is intentionally stricter than the runner's intersection projection and why backend flips reset the quiet window.
- Added the new Swift unit tests to the iOS workflow.

Validation on the exact pushed commit:

- 2,626 related tests passed.
- Changed-line coverage: 40/40 (100%); changed-branch coverage: 45/48.
- iOS runner build, focused simulator XTests (7/7), and macOS runner build passed.
- Live production-daemon evidence on the existing iOS simulator: primed a real `private-ax` snapshot (70 nodes), then ran `press @e38 --settle` through the CLI/daemon route. It navigated Settings → General and settled successfully in 1,776 ms with 3 captures. Runner diagnostics record the priming capture plus all 3 settle captures as `backend=private-ax`.

No simulator was created for this verification; the session was closed afterward. No docs or skills changed because this remains internal capture/presentation and routing behavior with no public command-surface change.

thymikee:
Re-review at `d3f8412ad`: code review is clean and the prior findings are resolved. Snapshot option forwarding is now centralized in a type-exhaustive mapper used by the daemon routes, including the previously missed interaction runtime; the production-route regression verifies the private-AX pin on emitted runner snapshots. Regular and private-AX stability comparisons now have explicit jitter and real-change tests. Live CLI/daemon evidence confirms a private-AX `press --settle` remained pinned across all settle captures. All completed checks are green; two smoke lanes are still in progress. Ready for human review.

PR #1793, opened 2026-08-17
test(android): restore full-tier lifecycle and observability scenarios (#1781 A1)

thymikee:
## Summary

`#1781` A1. The nightly Android job (`Android Full Emulator Suite`) has failed every scheduled run since the full tier landed, always on the same assertion:

```
AssertionError [ERR_ASSERTION]: step: request Android microphone permission for deny
command: agent-device click id="automation-request-microphone" ...
"message": "Selector did not match: id=\"automation-request-microphone\"",
scenario: full:lifecycle-system
```

## Root cause

`settings permission reset microphone` maps to `pm revoke` (documented in `commands.md`), and Android kills the app process when a **granted** runtime permission is revoked. The `accept` round grants the permission; the reset that opens the `deny` round therefore terminates the fixture, and the click lands on the launcher.

Proven on the image the lane pins (`android-36 google_apis_playstore`, Pixel 7 profile):

```
--- reset (revoke) while NOT granted
pid: 4259                       # app survives
--- grant / revoke while GRANTED
pid after revoke:               # empty — process gone
topResumedActivity=ActivityRecord{... com.google.android.apps.nexuslauncher/.NexusLauncherActivity}
```

The nightly artifacts agree: in run 31990896404 the `scroll bottom` immediately before the failing click reports `"summary": "Already at bottom", "passes": 0` (the launcher is not scrollable), while the same step in the `accept` round scrolled 1440px.

### Why #1610 did not fix it

#1610 fixed a *different* defect in the same helper — a one-shot `alert get` racing the asynchronous prompt — by switching to `alert wait 10000`. That made the `accept` round reach its prompt, which moved the failure one step later, onto the revoke-kill. Before #1610 the job failed at `{"alert":null,"message":"No alert visible"}` (run 30974032723); from 2026-08-06 onwards it fails at the `deny` click (runs 31070190261, 31924041273, 31990896404).

### Why there was more than one defect

`full:lifecycle-system` (added in #1484) and `full:observability-artifacts` have never executed end to end — #1484 explicitly deferred to the *smoke* tier for device validation, and the nightly has died inside the first of them every night since. Running them live surfaced the whole chain:

| # | Defect | Symptom |
|---|--------|---------|
| 1 | reset revokes → app killed | `Selector did not match: id="automation-request-microphone"` |
| 2 | post-revoke readback used `home` + `open` | cold start lands on the tabs home, `wait "Automation lab"` times out |
[...]

Issue #1584, opened 2026-08-04
iOS gesture pan duration has no automated guard; the nightly gesture gate cannot catch a #1562 regression


thymikee:
Follow-up to #1562 / #1572.

#1572 fixed `gesture pan` ignoring its duration on iOS, and the fix is real — but nothing in CI would catch it coming back.

## Why the existing gate does not cover it

`examples/test-app/replays/gesture-lab.ad` already drives the exact command class that regressed:

```
gesture pan 110 443 48 0 500
```

Two problems:

1. **It does not run on PRs.** `fixtureReplays` is a full-tier scenario (`test/integration/ios-simulator-e2e/scenarios.ts`: `{ id: 'full:fixture-replays', runner: 'fixtureReplays', tier: 'full' }`), and `.github/workflows/ios.yml` runs `AGENT_DEVICE_IOS_E2E_TIER: smoke`. Only `replays-nightly.yml` runs the full tier.
2. **What it asserts is activation counts, not timing.** The verified behavior is "gesture fixture counters prove pan/fling/pinch/rotate/transform". A pan that collapses to ~100 ms still increments the counter, so the assertion stays green through the entire #1562 symptom.

## Why the obvious fix does not work

The runner reports `gestureStartUptimeMs` / `gestureEndUptimeMs`, but `normalizeAppleRunnerResultForResponse` (`src/platforms/apple/core/runner/runner-result-response-normalization.ts`) strips both from the response as diagnostic-only fields. The e2e harness runs the CLI with `--json`, so it cannot see them. #1572's evidence table had to be read out of `--debug` per-request ndjson.

Exposing those fields publicly just to test them would be the wrong trade — they are deliberately internal.

## Suggested approach

Make the duration observable in the fixture instead of in the protocol. `examples/test-app/src/screens/GestureLab.tsx` already uses `Gesture.Pan()` and renders counters as plain text that replays assert with `wait`. Recording an observed-duration bucket from the handler's begin/end timestamps and rendering it the same way makes the assertion a one-line `wait` in `gesture-lab.ad`:

```
gesture pan 110 443 48 0 500
wait "pan duration >=400ms" 5000
```

No new plumbing, no diagnostic-field exposure, and it exercises the public replay path end to end.

Worth deciding separately whether a single duration-checked pan should be promoted into the smoke tier, since a nightly-only guard means a regression can sit on `main` for a day.

## Context

[...]

PR #1794, opened 2026-08-17
ci: move parked device replay suites to a dispatch-only workflow (#1781 A1)

thymikee:
## Summary

`#1781` A1 fallback. `Replay Nightly` has failed **every** scheduled run since 2026-07-24. Two jobs are responsible and neither has an owner:

- **Android Full Emulator Suite** — dies inside `full:lifecycle-system`, a scenario that had never executed end to end. Diagnosis and a validated fix for it (and for `full:observability-artifacts`) are in #1793; even with that, the last scenario `full:fixture-replays` still fails on a replay fixture whose scroll canary is authored for a taller device than the `pixel_7` profile the lane pins.
- **iOS Replay Suite** — fails on varying steps; not diagnosed here.

Both jobs move, byte-for-byte, into `.github/workflows/replays-manual.yml`, which has **no `schedule:`**. The parser fuzz lane keeps its nightly schedule in `replays-nightly.yml` and is untouched; the macOS job is already gone (#1787). Nothing about how the suites run changes — dispatch them with:

```
gh workflow run replays-manual.yml --ref <branch>
```

## Why a separate file rather than `if: github.event_name == 'workflow_dispatch'`

The first revision of this PR used a job-level `if:`. It would have parked the jobs and lied about it: `workflowLanes()` (`scripts/gate/workflows.ts`) computes `qualifying = 'pull_request' in on || 'schedule' in on` once per **workflow file** and never reads job-level `if:`. So `pnpm check:gate-manifest` kept printing

```
gate manifest: ok — 48 checks wired across 32 lanes, 1 declared unprovable.
```

while `replay-android`, `replay-ios`, and `replay-ios-device` — declared **only** in those two jobs — no longer ran on any schedule. That is exactly the failure the manifest exists to catch: *"a check that silently loses its owner looks exactly like a green build"* (`docs/agents/testing.md`). A separate `workflow_dispatch`-only file is what the existing file-level model already reads correctly.

## What that exposed, and how it is declared

With the jobs moved, the audit reports the truth:

```
check "replay-ios" is not declared by any pull_request/schedule lane: script:test:replay:ios.
check "replay-ios-device" is not declared by any pull_request/schedule lane: script:test:replay:ios-device.
```

These are declared in `scripts/gate/declarations.ts` as a **new** `MANUAL_ONLY_OWNERS` record rather than being folded into `UNPROVABLE_OWNERS`, because the two facts are opposites:

- `UNPROVABLE_OWNERS` claims *"the suite runs; this loader cannot see it"* — which stopped being true for `replay-android` the moment its lane stopped running automatically, so its stale prose moved with it. `UNPROVABLE_OWNERS` is now empty, with a comment saying why.
- `MANUAL_ONLY_OWNERS` says *"nothing runs this until someone dispatches it"*, with the reason and the exit condition per entry.

`check:gate-manifest` now reports:

```
gate manifest: ok — 46 checks wired across 32 lanes, manual-only: replay-android, replay-ios, replay-ios-device.
```

Named, not counted — the ids read like a list of things nothing runs, where "3 manual-only" would read like a tally.

Each entry names the dispatch lane that still runs it, and a `manual-only` audit assertion resolves that name against the derived model: the lane must exist, must still be dispatch-only, and must still declare the gate. Without that the record would be a plain allowlist — deleting a parked job would leave the manifest green and still printing the check as "manual-only", turning parked coverage into deleted coverage. `replay-android` carries an explicit `opaque` flag, because its gate sits inside the third-party emulator action's `script:` (#1429): the job's existence is the whole attestation the model can make, and the flag says so instead of letting an unreadable lane look like a declaring one.

Six regressions pin the path (`pnpm check:gate-manifest:test`, 34/34):

- a `workflow_dispatch`-only lane reads its gate but is non-qualifying;
- every manual-only entry names a registered check no qualifying lane owns;
- deleting a declaration reports its check as unowned;
- deleting the parked job fails with `no workflow defines`;
- putting the lane back on a schedule fails until the entry is dropped;
- a parked lane that loses its `run-gate` step fails — unless the entry is opaque, which is asserted too, so the flag cannot wave a readable lane through.

Both ratchets were checked for vacuity by planting `layering` (a genuinely owned check) in the record and watching the suite go red.

`docs/agents/testing.md` records the attestation and why `replay-android` is opaque.

## Validation

`pnpm check:tooling` passes (format, lint, typecheck, layering, depgraph, gate-manifest + its tests, production-exports, tmpdir-leaks, mcp-metadata, build, bundle-owner-files, package). The moved job bodies are byte-identical to the originals — only the job ids changed (`nightly-android`/`nightly-ios` → `manual-android`/`manual-ios`); job **names**, steps, pins, env, and artifact names are unchanged.

## Notes for whoever picks these up

- **iOS shares the shape of the Android root cause.** 2 of the last 5 iOS nightly failures are `full:lifecycle-system` / `step: wait for Automation lab` (run `31990896404`) — the same scenario family as the Android bug in #1793, where a permission reset kills the app and the relaunch does not land on the expected route. That class is worth checking first. The iOS scenario lives in `test/integration/ios-simulator-e2e/live-full-scenarios.ts`, separate from the Android files #1793 touches.
- **#1584's planned nightly-only pan-duration guard becomes manual-only too**, so its open "promote to the smoke tier?" question is now live rather than theoretical: as written, it would land in a lane nothing runs.

## Exit condition

Move the two jobs back into `replays-nightly.yml` and delete their `MANUAL_ONLY_OWNERS` entries once a dispatch run is green. The audit test fails if the entries outlive the parking, so this cannot quietly become permanent.


thymikee:
P1: MANUAL_ONLY_OWNERS is currently only a negative allowlist. The new test proves each entry is registered and has no qualifying owner, but it never proves a workflow_dispatch lane still declares and can run that check. Deleting a parked manual job, or its run-gate step, would therefore leave check:gate-manifest green and continue to print the check as manual-only: parked coverage has silently become deleted coverage.

Please derive/attest each manual-only declaration from a dispatch-only lane and add a regression that deleting its declaration fails. Android needs an explicit attestation too: replay-android remains inside the opaque third-party action, so the existing loader cannot supply it by itself.

The loss of scheduled replay coverage itself matches #1781 A1 (manual means off). The current Bundle Size and CodeQL failures are GitHub 503 infrastructure failures; the wording/count issue in the manifest output is non-blocking.

thymikee:
CI status on this head: 25 pass, 2 fail. Both failures are `Analyze (java-kotlin)` and `Analyze (javascript-typescript)` from the GitHub outage this PR was opened during — the logs contain only `Encountered an error while trying to determine feature enablement: HttpError: No server is currently available…`, never a finding, and that CodeQL run reports `cannot be rerun; This workflow run cannot be retried`. `Bundle Size` failed the same way (`Failed to update PR comment: 503`) and passes on re-run, as does every other job. The next push, or a maintainer re-run once GitHub finishes recovering, will clear the two Analyze rows.

thymikee:
P1 addressed in `643bc9c`: the declaration is now an attestation the audit resolves, not an allowlist.

`MANUAL_ONLY_OWNERS` entries carry the lane that still runs them (`Replay Manual / Android Full Emulator Suite`, `Replay Manual / iOS Replay Suite`), and a new `manual-only` audit assertion resolves that name against the derived model. It fails when the lane is gone, when the lane has become a `pull_request`/`schedule` lane again, or when the lane no longer declares the gate — so deleting a parked job or its `run-gate` step turns the manifest red instead of leaving the check printed as merely parked.

Android gets the explicit attestation you asked for. Its entry is marked `opaque: true`, which is the record saying out loud that the lane's own steps cannot show the gate — it sits in the third-party emulator action's `script:` (#1429) — so the job's existence is the whole attestation the model can make, and the audit checks exactly that rather than pretending to check a step it never reads. The derived lanes make the difference visible:

```
replays-manual.yml | Replay Manual / Android Full Emulator Suite | qualifying=false | gates=["android-helpers"]
replays-manual.yml | Replay Manual / iOS Replay Suite            | qualifying=false | gates=["swift-runner-ios","replay-ios","replay-ios-device"]
```

Four regressions pin both directions (`pnpm check:gate-manifest:test`, 34/34):

- deleting a declaration reports its check as unowned (`audit(model, { manualOnly: {}, unprovable: {} })`);
- deleting the parked job fails with `no workflow defines`;
- putting the lane back on a schedule fails until the entry is dropped;
- a parked lane that loses its `run-gate` step fails — *unless* the entry is opaque, which is asserted too, so the flag cannot be used to wave a readable lane through.

I checked the new ratchet is not vacuous by planting `layering` (a genuinely owned check) in the record and watching the suite go red.

`docs/agents/testing.md` now states that each entry names its dispatch lane, that the audit resolves it, and why `replay-android` is opaque. `pnpm check:tooling` passes; the manifest still prints `46 checks wired across 32 lanes, manual-only: replay-android, replay-ios, replay-ios-device`.

On sequencing: #1793 now has a green end-to-end full-tier run, including `full:fixture-replays`, so parking is no longer the only way to close A1 — this PR stands as the fallback if you would rather park the lane than carry those fixture repairs.

thymikee:
Re-reviewed exact head 643bc9c. P1 remains: the new audit proves only that a declared owner has no PR/schedule-qualified lane; it does not prove the workflow is still `workflow_dispatch`-only. Replacing `workflow_dispatch` in `replays-manual.yml` with `push` (or another non-qualifying trigger) while retaining the jobs/gates leaves the audit green and still reports these checks as manual-only. Preserve trigger kinds in the workflow/lane model (or add an explicit dispatch-only predicate) and add a planted regression that replaces/removes `workflow_dispatch`. Exact-head CI is otherwise green and merge state is clean.

thymikee:
Fixed in `b94f93d`. You are right that `qualifying === false` was the wrong predicate: `push`, `release` and friends sit on the same side of that bit, so swapping the trigger would have kept the audit green while the parked lane started itself again.

The lane model now preserves the trigger names instead of collapsing them (`Lane.triggers`), and the manual-only assertion requires `workflow_dispatch` **and nothing else**:

```
"Replay Manual / iOS Replay Suite" is triggered by push, so "replay-ios" is not manual-only —
"manual" is a claim about who starts the run, and every trigger other than workflow_dispatch
starts it without them. Restore a dispatch-only workflow, or declare what actually runs the check.
```

Three regressions cover exactly the gap you named (`pnpm check:gate-manifest:test`, 37/37):

- `a parked lane re-triggered by push fails even though push does not qualify` — the planted case from your comment;
- `a parked lane that loses workflow_dispatch entirely fails` — removal, not just replacement;
- `trigger kinds survive the model, not just whether they qualify` — a loader-level pin that a push workflow reads `["push"]` and `replays-nightly.yml` reads `["schedule","workflow_dispatch"]`, so a future refactor cannot quietly collapse the field the assertion depends on.

`pnpm check:tooling` passes and the manifest still reports `46 checks wired across 32 lanes, manual-only: replay-android, replay-ios, replay-ios-device`.

thymikee:
Re-reviewed exact head b94f93d. Clean readiness verdict: the lane model now preserves trigger kinds, the manual-only audit requires `workflow_dispatch` and nothing else, and planted push/no-trigger regressions close the prior bypass. Exact-head CI is fully green. Ready for human review.

PR #1780, opened 2026-08-17
docs: reposition README around mobile app automation for AI agents

thymikee:
## Summary

Repositions the README so it leads with the category and the three surfaces before differentiating on verification and evidence: category keywords first so humans and search engines can classify it, verification and evidence as the differentiator.

Before: the intro opened with the job statement, then a dense platform paragraph carrying per-platform caveats (HarmonyOS uses HDC and ArkUI `uitest`; Vega OS is VVD-only), then the Vercel attribution. Zero occurrences of "mobile app automation" or "MCP server" above the fold. MCP setup and the Node.js client were only reachable through docs links; capabilities and proof were below the entire Quick start.

After:

- Category line first: mobile app automation and verification for AI coding agents, through the CLI, built-in MCP server, or typed Node.js API. Then the job pitch with a keyword-richer platform sentence, then a works-with/proof line (documented client list; every named proof point links to public evidence).
- Quick start follows the hero directly; "What agents can do" and the loop diagram sit below the walkthrough.
- Quick start now has three labeled paths after the shared install step: CLI walkthrough (unchanged), MCP `mcpServers` config, and a ten-line `createAgentDeviceClient()` snippet mirroring `examples/sdk/client-session.ts`.
- New "What to ask your agent" section with five developer prompts; a one-sentence product ladder atop "Where to run"; two AEO-shaped FAQ entries.
- Per-platform transports and caveats moved to "How it works", with one sentence pointing at `agent-device capabilities --platform <platform>` as the authoritative depth check.
- Cloud/remote row points at the remote proxy and device clouds docs instead of `agent-device.dev/cloud`.
- npm `description`, `server.json` registry description (99/100 chars), and the rspress site description (110 chars) use the same category phrasing.
- Em dashes and evaluative filler removed per the Callstack humanizer skill; the tone-of-voice reference was checked and already passes.

Guard against the original drift (a contributor appending "HarmonyOS uses HDC…" to the intro because the Vega caveat had already set the pattern): the pattern-seeding sentences are gone, an inline HTML comment above the intro states the rule at the point of edit, and AGENTS.md "Docs & skills" gets a README-altitude rule.

Length: 207 lines / ~1580 words, up from 149 / ~1150; the growth is the MCP + Node quick starts, prompts, and proof line, with duplicated Node subpath and FAQ text removed.

## Validation

[...]

PR #1804, opened 2026-08-18
feat(ai-sdk): add agent-device/ai-sdk tool set and document the MCP zero-code path

thymikee:
## Summary

Follow-up to #1780's README repositioning: gives AI SDK users three ways to drive `agent-device`, in increasing order of setup, and closes a real gap in the publishing gate along the way.

- **Zero-code**: documents `@ai-sdk/mcp`'s `createMCPClient()` pointed at `agent-device mcp` via stdio — no `agent-device` import at all.
- **`agent-device/ai-sdk`** (new subpath): `createAgentDeviceTools({ session, platform?, set?, approval? })` builds AI SDK tools from the same command registry the MCP server uses (`listCommandTools()` / `createCommandToolExecutor()`), so the two integrations can't drift apart. No zod layer — JSON Schema goes straight through `jsonSchema()`.
- **Hand-written tools**: kept, now framed as the "smaller than core, or app-specific shaping" option, with a stale `needsApproval` reference fixed to the current `toolApproval` API.

```ts
import { ToolLoopAgent } from 'ai';
import { createAgentDeviceTools } from 'agent-device/ai-sdk';

const { tools, client, toolApproval } = await createAgentDeviceTools({
  session: 'ai-sdk-agent',
  platform: 'ios',
  approval: { close: 'user-approval' },
});

const agent = new ToolLoopAgent({ model: process.env.AI_MODEL!, tools, toolApproval });
```

`set: 'core'` (default) exposes a curated perceive/act loop — open, close, snapshot, click, press, fill, type, get, is, find, wait, back, scroll, swipe, alert, screenshot — driven by a new `frameworkTier: 'core' | 'extended'` descriptor facet declared on every public command (registry-owned, not a hand list in the adapter). `set: 'all'` matches the MCP server's full surface. Two new tests in `parity.test.ts` pin the classification so it can't drift silently.

`ai` is an **optional peer dependency**, imported lazily inside `createAgentDeviceTools()` rather than at module scope — `import('agent-device/ai-sdk')` never requires `ai` to be installed, only calling the factory does (with a clear `MISSING_PEER_DEPENDENCY` error if it's missing). This surfaced a real gap: `scripts/check-package.ts`'s publishing gate imports every `exports` entry from a clean install and audits every shipped bare specifier against `dependencies` — it had no concept of peer dependencies, since this package has never shipped one before. Extended `scripts/lib/shipped-imports.ts` to also accept `peerDependencies`, with new fixture tests.

## Validation

- `pnpm typecheck` (root + `examples/sdk/tsconfig.json`), `oxlint`, `oxfmt --check` clean on every touched file.
- `pnpm build` — new `dist/src/ai-sdk.{js,d.ts}` outputs; confirmed `ai` is external (not bundled) and the built module has zero static top-level `import ... from "ai"`.
- Full `unit-core` suite: 869 files / 6630 tests passing, including the extended `command-descriptor` parity suite, `client-api-examples-drift`, `command-tools`, and the new `package-closure-audit` fixtures.
- `client-api.md`'s fenced snippets still compile against real exports (`test/integration/client-api-doc-snippets.test.ts`).
- Manually reproduced the actual publishing-gate scenario end-to-end (`npm pack` → clean external install with **no** `ai` present): all 13 `exports` entries import cleanly, `createAgentDeviceTools()` throws the intended `MISSING_PEER_DEPENDENCY` error without `ai`, and with `ai` installed returns 16 tools for `set: 'core'` / 55 for `set: 'all'`, with `toolApproval` passed through correctly.

## Notes

- The `core` classification (16 commands) is a first cut — flagging for review rather than treating as final.
- Docs-only change for the MCP zero-code section; runtime validation doesn't apply there beyond the code compiling, which isn't separately gated (only `client-api.md`'s snippets are compiled, not `ai-sdk.md`'s).
- Scope: 14 files touched (2 new). A handful of unrelated, pre-existing `scripts/maestro-conformance/corpus/*.yaml` quote-style changes were present in this worktree before this work started and are intentionally left out of this PR.

### DIFF
--- a/.fallowrc.json
+++ b/.fallowrc.json
@@ -13,6 +13,7 @@
     "src/sdk/contracts.ts",
     "src/sdk/selectors.ts",
     "src/sdk/finders.ts",
+    "src/ai-sdk/index.ts",
     "src/bin.ts",
     "src/client/companion-tunnel.ts",
     "src/daemon.ts",
@@ -32,6 +33,7 @@
     "examples/sdk/metro-runtime.ts",
     "examples/sdk/contracts-result.ts",
     "examples/sdk/batch-orchestration.ts",
+    "examples/sdk/ai-sdk-tools.ts",
     "test/scripts/metro-prepare-packaged-smoke.mjs",
     "test/integration/*.test.ts",
     "website/docs/404.mdx",
@@ -96,7 +98,7 @@
     },
     {
       "comment": "Published package surface (see package.json#exports); consumed by SDK users, not internally.",
-      "file": "src/sdk/*.ts",
+      "file": "{src/sdk/*.ts,src/ai-sdk/index.ts}",
       "exports": ["*"]
     },
     {
--- a/examples/README.md
+++ b/examples/README.md
@@ -10,8 +10,9 @@ fenced TypeScript code block in the doc itself against the real `agent-device/*`
 ## sdk/
 
 Standalone scripts under [`sdk/`](./sdk) exercise the published `agent-device` export map —
-`agent-device`, `agent-device/metro`, `agent-device/contracts`, and `agent-device/batch` — the same
-way a Node consumer or the agent-device-cloud bridge would. Each file:
+`agent-device`, `agent-device/metro`, `agent-device/contracts`, `agent-device/batch`, and
+`agent-device/ai-sdk` — the same way a Node consumer or the agent-device-cloud bridge would. Each
+file:
 
 - has a top comment stating what it demonstrates and its prerequisites (daemon running,
   device/simulator available);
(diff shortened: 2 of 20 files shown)


thymikee:
Reviewed exact head b826de03. P1 containment blocker: `createAgentDeviceTools()` promises a tool set pinned to one session, but it only removes `session` and `mcpOutputFormat` from MCP-derived schemas. `stateDir` remains model-visible, and the shared executor consumes a supplied `stateDir` as client configuration, allowing a tool call to redirect into another daemon state directory. Hide MCP transport/config controls from this adapter (at minimum `stateDir`; preferably `stateDir`, `mcpOutputFormat`, `includeCost`, and `responseLevel`) and add a regression proving they are absent from schemas and cannot reach execution. Exact-head CI is green; descriptor-tier/package/export architecture otherwise looks coherent.

thymikee:
Fixed in d0b8f6c19 — thanks for the catch.

`createAgentDeviceTools()` now hides `stateDir`, `includeCost`, and `responseLevel` from tool schemas alongside `session` and `mcpOutputFormat`, and strips all of them from the runtime input inside `execute()` too (not just the schema), so the pinned-session guarantee holds even if a caller bypasses schema validation.

Added a regression (`src/ai-sdk/__tests__/index.test.ts`) that asserts the fields are absent from the schema and, separately, calls `execute()` with all of them supplied directly (simulating a schema-validation bypass) to prove none reach the shared executor — only the pinned `session` does.

The schema-property filter and the new input filter share one `omitHidden()` helper instead of two near-duplicate implementations.

thymikee:
Clean delta review at `d0b8f6c1`: the AI SDK adapter now derives schema hiding and runtime-input stripping from one hidden-field set, so pinned session/state configuration cannot be bypassed even by direct `execute()` calls. The regression covers both model-visible schema and validation-bypass execution. ADR 0008’s descriptor-derived tool source remains intact; exact-head CI is fully green and no device evidence applies. Ready-for-human.

Issue #1797, opened 2026-08-17
iOS snapshot backends: make interpretation divergence impossible by construction (acquire/present split)


thymikee:
## Problem

The iOS runner has three snapshot backends (recursive tree, query sweep, private AX) that converge on one `SnapshotNode` wire model but implement **interpretation** — visibility, clipping, membership, hittability, scope, raw rendering — separately per backend. Nothing relates the copies: no shared implementation for several policies, no cross-backend test, no CI gate. Every incompatibility so far has been fixed by hand-synchronizing a second copy, and the class survives each point fix.

Exhibit A is #1784 (8b698e8): the scroll-overflow leak (text overflowing a scroll container but inside the viewport was visible on private AX, hidden on the tree) was fixed by making `appendPrivateAXNode` *call* the tree's clip helpers (`isVisibleInRegularSnapshot`, `scrollContainerAnchor`, `rememberHiddenContentHint`). Correct fix — but it's a second, manually-kept-in-sync call site. Nothing stops the next drift.

Live divergences at HEAD (3020195):

| # | Field/flag | Tree | Private AX | Where |
|---|---|---|---|---|
| D1 | `-i` membership | interactive-type OR hittable OR content (`shouldInclude`) | **every visible node survives** (decorative images, captions, contentless others) | `RunnerTests+Snapshot.swift:810` vs `RunnerTests+FlatSnapshotFiltering.swift:43` |
| D2 | `hittable` | center-in-viewport + occlusion scan | purely type-based; sweep uses real `element.isHittable` — three meanings for one field | `Snapshot.swift:841` vs `PrivateAXPresentation.swift:56` vs `Snapshot.swift:1520` |
| D3 | `--scope` | re-roots acquisition via live query | substring match over label/identifier/value | `Snapshot.swift:873` vs `FlatSnapshotFiltering.swift:10` |
| D4 | `--raw` | selects raw tree rendering | **silently ignored** — the private-AX tier of `rawDiagnosticPlan` never reads `options.raw`, so a recovered `snapshot --raw` returns regular presentation labeled raw | `SnapshotCapturePlan.swift:476` (only read) |
| D5 | `value` | stringified at presentation | stringified at acquisition (ObjC boundary) — irreducible, see below | `Snapshot.swift:1121` vs `RunnerAXSnapshotBridge.m:803` |
[...]

PR #1807, opened 2026-08-18
fix: prevent private AX subtree leaks at source

thymikee:
## Summary

Prevent private-AX descendants of an offscreen scroll item from reappearing at a clamped visible edge. Element settings rows such as Theme, Notifications, and About now stay excluded from interactive snapshots when their ancestor is offscreen.

This follow-up removes the downstream collapsed-cluster detector and second iOS presentation pass introduced in #1798 while retaining Maestro source-index provenance. XCTest settle projection now uses the canonical visibility rules with linear behavior for nested and flat scroll trees.

Related to #1797; this does not implement or close the broader acquire/present migration.

Scope: 22 files across Apple runner presentation, canonical snapshot visibility and settle, daemon iOS presentation, and Maestro provenance.

## Validation

The new XCTest regression was proven red before the fix: an offscreen Theme cell leaked two clamped Theme descendants. It passes after the fix alongside the geometryless-content and scroll-hint regressions.

Two settle-signal regressions were proven red independently: a visible node changing only its value incorrectly compared stable, while whitespace-only semantic changes incorrectly compared unstable. The signal retains value and preserves canonical semantic trimming; the focused settle/wait suite passes 24 tests.

iOS and macOS runner builds with unit regressions enabled passed. Focused unit and integration tests cover composed replacement updates, tree and flat-tree settle linearity, iOS presentation, snapshot capture, and Maestro target provenance. The full affected gate passed on the pushed head: 556 test files and 4,724 tests passed with 100% changed-line coverage. A final Sol high review found no remaining in-scope issues on the current head.

No docs or skills changed because this is an internal correction to existing snapshot behavior; the XCTest CI allowlist was updated for the new regression.


thymikee:
## Code-quality review

This is the right shape: the leak is closed where the information lives (`appendPrivateAXNode` now carries `ancestorOffscreen`, so an excluded offscreen parent's clamped-frame children can't re-enter), and everything that existed downstream only to detect that leak is gone — the collapsed-cluster heuristic, the second presentation pass, `viewport.ts`, and settle's private forked viewport resolver. Net −1080 lines with no behavior lost that we want. Settle now goes through the canonical `isNodeVisibleOnScreen`, and the `updateReplacement(current => patch)` change fixes the derive-from-raw-node footgun at the API instead of at one call site. 👍

Verified locally on the PR head: the touched unit suites pass (88 tests: stable-capture-signal, snapshot-presentation/*, snapshot-capture, daemon-runtime-port-observation, contracts snapshot, maestro runtime-targets-typed). Swift regression reads correctly for the Theme/clamped-children case.

Two things worth tightening; neither blocks:

**1. `precomputedViewportRects` lets a caller change semantics, not just cache them.** `resolveViewportRect` collects root rects with `hasValidRect` (finite), while settle precomputes them with `isPositiveFiniteRect` (positive size). Today the difference is only a zero-area Application rect (fail-open before, falls through to the "largest containing rect of any node" fallback now), but the parameter is documented as a precompute and it's threaded as a 4th optional positional through four functions. Suggest contracts exports the collector — `collectViewportRootRects(nodes)` — and callers pass *that* result. Then the arg is provably a cache of the same computation and the predicate can't drift between call sites.

**2. `mergeReplacement` and `updateReplacement` are two exports for one operation.** `mergeReplacement(r, n, patch)` is `updateReplacement(r, n, () => patch)`. Either define it that way so the relationship is explicit, or migrate the remaining `mergeReplacement` sites and drop it.

Notes, no action needed:
- The hint registration now fires only for geometrically-offscreen nodes (`!ancestorOffscreen && hasFrame && !onScreen`), not for `negligibleDecoration` — that's a quiet correction over #1798 and it's the right one; just confirming it's intentional.
- Settle no longer has a TS-side guard for sub-pixel private-AX decoration or `value` churn; it now relies on the runner's `negligibleDecoration` from #1798. Fine since the runner ships with the package, but it does mean settle-vs-runner version skew is a thing to keep in mind if the runner build ever lags a CLI release.

Issue #1320, opened 2026-07-17
feat: add cross-worktree device ownership and safe recovery


thymikee:
## Summary

Add a platform-neutral device ownership layer so agents can discover, acquire, release, and safely recover devices across worktrees without inspecting or killing PIDs.

The intended experience is:

```text
open -> use -> close

busy?
  -> device status
  -> reuse the owning session, choose another device,
     release --stale, or explicitly release --force
```

The same public behavior must apply to local and remote iOS and Android targets. Platform-specific resources such as XCTest runners, Android instrumentation, the test IME, recordings, provider allocations, and daemon processes remain implementation details behind that interface.

## Motivation

Ownership currently exists at several different layers:

- `SessionStore` prevents two sessions in one daemon from using the same device, but each source checkout normally has its own state directory and daemon.
- Remote device leases support iOS and Android, with TTL and heartbeat semantics, but have no public list/release CLI surface.
- Apple runner leases prevent cross-daemon XCTest ownership, but contention appears late as runner-preparation `COMMAND_FAILED` rather than structured `DEVICE_IN_USE` at `open`.
- Local Android has no equivalent host-global device claim. Two worktree daemons can reach the same device and conflict later through instrumentation/UiAutomation, IME restoration, recordings, or ADB resources.
- `session list` only reports sessions visible to the selected daemon/workspace scope, so it can disagree with ownership errors from another daemon.
- Recovery guidance can degrade into `ps`, `kill`, sleeps, and runner PID checks even though much of the required identity and cleanup machinery already exists.

Related history:

- #1246 records the user-visible `DEVICE_IN_USE` / hidden-owner failure shape.
- #1169 added idle daemon reap and proof-based stale Apple runner takeover.

## Goals

- Enforce exclusive device ownership across local worktree daemons.
- Use the same public ownership lifecycle for iOS and Android.
- Make ownership inspectable without starting a daemon.
- Automatically reclaim only provably orphaned ownership.
- Provide graceful-first explicit recovery for live owners.
- Keep device claims, remote leases, daemon sessions, and runner/process leases as distinct concepts.
- Prevent device claims from becoming free while attributable resources remain owned.
- Preserve warm Apple runner reuse where it does not conflict with logical ownership.
- Give agents structured recovery data instead of requiring prose parsing or PID discovery.

## Non-goals

- Merge remote leases and Apple runner leases into one record.
- Permit concurrent mutating sessions on one device.
- Reclaim a verified live owner because it appears inactive.
[...]

PR #1809, opened 2026-08-18
fix: enforce device claims for sessionless device mutations (#1799)

### LINKED ISSUE #1799: shutdown can terminate a device held by a foreign live claim (0.20.9)
## Summary

On 0.20.9, `agent-device shutdown` appears not to consult host-local device claims. A session holding no claim was able to shut down an Android emulator that another session held a verified-live claim on, and the command reported success.

This reads as a gap in the `transient-exclusive` policy that #1320 assigns to `shutdown`, not as a problem with what #1729 shipped — that work was scoped to `open`, and nothing here contradicts it.

## Environment

- agent-device 0.20.9 (npm)
- macOS, Apple silicon
- Android emulator (`emulator-5554`)
- Several agent sessions running in parallel, one per git worktree, each with its own state directory and daemon

## What happened

Session A held a live claim on the emulator. Session B — a different worktree, a different state directory, holding no claim — ran:

```
agent-device shutdown --platform android --serial emulator-5554
```

Output (unrelated fields elided; `stdout` is the emulator's own `adb emu kill` reply):

```json
{"success": true, "stdout": "OK: killing emulator, bye bye"}
```

The emulator terminated.

### Expected

A foreign verified-live claim causes the command to refuse, consistent with #1320's policy table:

> | `transient-exclusive` | `boot`, `shutdown`, sessionless device mutations | Acquire a command-scoped claim and release in `finally`; refuse foreign claims. |

and its acceptance criterion:

> Sessionless device mutations cannot affect a foreign claimed device.

### Actual

The non-owner's `shutdown` succeeded and the claim owner's device was destroyed.

## Three follow-on observations

Recording these because they affected recovery. All three map to items already in #1320's design.

**1. The claim still read `live` after the device was gone.** Staleness is proven from the owner *process* (pid + start time + state-directory existence), not from the device, so from the claim store's perspective nothing had changed. `device status --stale` therefore could not see it.

**2. There is no `device release`.** `agent-device device --help` on 0.20.9 lists only `device status`, and its help states that `--stale` only inspects: "automatic reclamation occurs during open or daemon startup after exact-owner resource reconciliation." Recovery was `agent-device daemon stop --state-dir <owner state dir>`, which worked.

**3. `daemon stop` reports `claimsReleased: []` even when it released a claim.** In `dist/src/daemon.js` (0.20.9) the graceful, forced, and not-running result paths each return `claimsReleased: []` and `claimsOrphaned: []` as empty literals, and the later projection step only rewrites `providerReleases`. #1320 lists "Claim results distinguish reconciled `claimsReleased` from cleanup-pending `claimsOrphaned`" under acceptance, so this may simply be unlanded rather than a defect.

## Notes

- Reported against 0.20.9. The CLI surface quoted above was re-checked on 0.20.9; the destructive sequence itself was not re-run, since reproducing it means terminating an emulator that is currently in use.
- This configuration — many parallel per-worktree sessions on one host sharing one emulator — is probably why it shows up here.

thymikee:
## Summary

Closes #1799.

**Before:** `shutdown` only refused a device that had an active session in *its own* daemon's session store, then called `shutdownTarget()`. It never consulted the host-global device claim store, so a daemon in another state directory terminated an emulator another worktree held a verified-live claim on and reported success. `boot` had the same gap.

**After:** a foreign live claim refuses the command with the existing `DEVICE_IN_USE` / `DEVICE_CLAIM_LIVE_OWNER` error (`retriable: false`, daemonless `device status` recovery hint) before any device operation exists.

> #1799's third observation — `daemon stop` reporting `claimsReleased: []` even when it released a claim — is fixed separately in **#1818**, stacked on this branch, so this PR stays scoped to the enforcement seam.

### Design

The fix is not a claim check in the two handlers — it makes the class unrepresentable, per #1320's "Command descriptor policy" table:

1. **`CommandDescriptor` gains a REQUIRED `deviceClaimPolicy`** with no default, using #1320's vocabulary (`none` | `observe` | `require-owner` | `transient-exclusive` | `acquire-session` | `release-session`). TypeScript forces every descriptor to declare one; a completeness/honesty test (mirroring the timeout-policy gate) pins the bounded, diffable set of everything that deviates from `require-owner`.
2. **Enforcement is derived from the trait at one seam.** `createRequestRuntimeBindings` — the only way any handler obtains device operations — already caches one binding per device key, so admitting the claim *as part of creating that binding* is both the choke point and the natural per-device deduplication. `bindExactDevice` deliberately bypasses the cache, so it admits its own. There is no per-handler call to forget.
3. `transient-exclusive` acquires a command-scoped claim after the gateway bind resolves (binding composes the operation catalog and mutates nothing) and before the narrowed projection reaches the handler; the request scope releases it on dispose, after the bindings it guards.
4. **Every other policy performs zero claim-store I/O**, keeping #1320's non-goal ("no per-command host-global filesystem reads or writes to session-bound hot paths") intact. A claim already held by *this* daemon process covers the command rather than colliding with it — decided from the claim record's `stateDir` + owner identity, so a claim taken earlier in the same request (`open`'s) can never lock the daemon out of its own device.
5. `acquireTransientDeviceClaim` reuses the existing claim implementation (same lock, inspection, write, orphan reconciliation); it records `session: "transient:<command>"` so `device status` shows an honest owner for a claim that lives milliseconds. No second claim implementation.
6. The same-daemon `shutdown` self-guard (active session → `close --shutdown`) is unchanged.

### Classification (every command, honestly)

| Policy | Commands |
| --- | --- |
| `acquire-session` | `open` |
| `release-session` | `close` |
| `transient-exclusive` | `boot`, `shutdown`, `install`, `reinstall`, `install_source`, `push`, `prepare` |
| `observe` | `devices`, `capabilities`, `device`, `doctor`, `apps`, `appstate` |
| `none` | lease/artifact/session bookkeeping, pure delegators (`batch`, `install-from-source`, `react-devtools`), and local-CLI commands |
| `require-owner` | everything session-bound (including `hover`) |

**Why the sessionless-capable mutations stop where they do.** I first classified `keyboard`, `clipboard` and `trigger-app-event` as `transient-exclusive` too. Live verification caught that as dishonest: those commands still declare ADR 0019 `legacy` platform execution and reach their device through dispatch, not through the request scope's device binding, so the gate never runs for them — a foreign `keyboard status` still succeeded against a claimed emulator. They are `require-owner` until their platform execution migrates to `device-runtime`, and the honesty test now *requires* `device-runtime` execution for `transient-exclusive`, so the same dishonest pairing cannot be declared again.

Two more deliberate `require-owner` classifications with reasons, as follow-ups:
- **`record`** — sessionless `record start` leaves a recording running past the request, and a command-scoped claim released in `finally` would free a device whose attributable resource is still owned (#1320: "Prevent device claims from becoming free while attributable resources remain owned").
- **`runtime`** (internal port-reverse) — same shape: it establishes durable per-session runtime state that outlives the request.

### Regression tests, proven red first

Reverted the production change (`git checkout HEAD --` on the touched production files, removed `device-claim-admission.ts`), kept the tests:

```
 FAIL  src/daemon/handlers/__tests__/session-boot-shutdown-device-claims.test.ts >
       shutdown refuses a device held by a foreign live claim and never reaches the device
AssertionError: expected true to be false
  112|   expect(response?.ok).toBe(false);
 FAIL  ... > boot refuses a device held by a foreign live claim and never reaches the device
AssertionError: expected true to be false
  137|   expect(response?.ok).toBe(false);
 Test Files  1 failed (1) · Tests  2 failed | 2 passed (4)
```

That `ok: true` on a foreign-claimed device is #1799 exactly. Restored, green:

```
 ✓ src/daemon/handlers/__tests__/session-boot-shutdown-device-claims.test.ts (4 tests)
   ✓ shutdown refuses a device held by a foreign live claim and never reaches the device
```

Coverage:
- `session-boot-shutdown-device-claims.test.ts` — two-daemon-shaped (foreign state dir + claim): `boot`/`shutdown` refused with the typed reason, `shutdownTarget`/`bootTarget` never called, the owner's claim untouched, and the transient claim released after both a successful and a failing shutdown.
- `device-claim-admission.test.ts` — one table over all six policies proving claim-store I/O happens only for `transient-exclusive`, plus foreign-claim refusal, same-daemon coverage, provider-owned skip, and the construction guard: the real `createRequestExecutionScope` claims for `shutdown` and not for `snapshot` on the same binding call.
- `device-claim-policy.test.ts` — registry-driven completeness, the reviewed non-`require-owner` sets, and the `transient-exclusive ⇒ device-runtime` honesty invariant.

### Live device verification (Android, `Pixel_9_Pro_XL` → `emulator-5554`)

**Pre-fix repro on the reported 0.20.9 build.** Daemon A (state dir A) held a live claim; daemon B (different state dir, different cwd) ran:

```
$ agent-device shutdown --platform android --serial emulator-5554 --state-dir <B> --json
{ "success": true, "data": { ..., "shutdown": { "success": true, "stdout": "OK: killing emulator, bye bye" } } }
$ adb devices        # emulator-5554 gone; A's claim file still present and `live`
```

**With this branch** (rebooted the AVD, both daemons on the built branch):

```
$ agent-device shutdown --platform android --serial emulator-5554 --state-dir <B> --json
{
  "success": false,
  "error": {
    "code": "DEVICE_IN_USE",
    "message": "android device emulator-5554 is owned by session \"fixed-a\" in workspace \"/…/agent-a4d4b25de54894d5a\".",
    "hint": "Inspect the owner with: agent-device device status --platform android --serial emulator-5554",
    "retriable": false,
    "details": { "reason": "DEVICE_CLAIM_LIVE_OWNER", "classification": "live",
                 "deviceKey": "local:android:none:emulator-5554",
                 "owner": { "session": "fixed-a", "workspace": "…", "stateDir": "…/stateA" },
                 "recovery": { "command": "agent-device device status --platform android --serial emulator-5554" } }
  }
}
$ adb devices        # emulator-5554 still up
```

`boot --platform android --device Pixel_9_Pro_XL` from B returned the identical refusal. After `close --session fixed-a` in A, B's `shutdown` succeeded (`"OK: killing emulator, bye bye"`) and left no claim file behind.

Sessionless `install` from B while A held the claim was refused the same way, while the owner's own in-session `install` reached the device (it failed on `INSTALL_FAILED_TEST_ONLY`, an unrelated property of the helper APK I used) — the same-daemon coverage path with no claim-store collision.

Cleanup: both daemons stopped, the AVD I booted shut down, no leftover claim file. iOS was not exercised; the seam is platform-neutral and `isLocalDeviceClaimTarget` gates it on the admitted runtime owner exactly as `open` does.

### Gates

`pnpm exec tsc --noEmit`, `pnpm format`, `pnpm check:affected --run` (432 files / 3684 tests, layering + fallow green — fallow reports **no issues** in the changed files). Wire compat: `pnpm check:daemon-wire-compat` (151 declarations, 0 changed — this PR adds no daemon wire surface).

## Tradeoffs and follow-ups

- **Scope:** 16 files, +947/−83. It expanded past `boot`/`shutdown` by design — #1320 requires auditing every sessionless device-mutating command, and the descriptor trait is what makes the audit permanent. A large share of the diff is one declaration line per descriptor, matching the existing `timeoutPolicy` convention; the interaction descriptors that were byte-identical now share a `TARGETED_TOUCH_INTERACTION_TRAITS` bundle beside the existing `GENERIC_MUTATING_LINUX_DEVICE_COMMAND_TRAITS`, which also removes the clone group the added line would otherwise have created.
- **Follow-ups:** `keyboard`, `clipboard`, `trigger-app-event` become enforceable once ADR 0019 migrates their platform execution to `device-runtime`; `record` and `runtime` need a claim whose lifetime follows the durable resource, not the request, before they can leave `require-owner`.
- `install`/`reinstall`/`install_source`/`push`/`prepare` are newly `transient-exclusive` — a behavior change for sessionless use against a foreign-claimed device (now refused). `install` was live-verified both ways; `push`/`prepare` were not.

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -63,6 +63,13 @@ task touches:
   cloud bridge, or `limrun`.
 - Runner/process lease: backend helper mutual-exclusion guard for platform runners or tools; it is
   not the remote client ownership boundary.
+- Device claim: host-global exclusive ownership of one local device, held by an open session or by a
+  single sessionless device-mutating command. Local only; remote targets use device leases instead.
+- Device-claim policy: required command-descriptor trait declaring a command's relationship to the
+  claim store (`none`, `observe`, `require-owner`, `transient-exclusive`, `acquire-session`,
+  `release-session`). The request-execution scope enforces it at the device binding seam:
+  `transient-exclusive` takes a command-scoped claim and refuses a foreign one, and every other
+  policy performs no claim-store I/O.
 - iOS physical-device control: Apple-local module selected from discovery evidence. CoreDevice
   devices retain the `devicectl` controller; devices found only by `xctrace` use the XCTest
   controller for readiness, app activation/termination, and cable-bound usbmux runner transport
--- a/src/__tests__/test-utils/device-claim-store.ts
+++ b/src/__tests__/test-utils/device-claim-store.ts
@@ -1,0 +1,39 @@
+import fs from 'node:fs';
+import path from 'node:path';
+import { afterEach } from 'vitest';
+import { mkdtempForTestSync } from './tmp-dir.ts';
+import type { DeviceClaimReconciler } from '../../daemon/device-claims.ts';
+
+export type IsolatedDeviceClaimStore = {
+  /** Temporary root holding both the claim store and the daemon state dir. */
+  root: string;
+  stateDir: string;
+  claimsDir: string;
+};
+
+/**
+ * Device claims are host-global by design, so a test that exercises them must
+ * redirect the store. Registers cleanup once and returns a per-test factory.
+ */
+export function isolatedDeviceClaimStores(prefix: string): () => IsolatedDeviceClaimStore {
+  const roots: string[] = [];
+  afterEach(() => {
+    delete process.env.AGENT_DEVICE_CLAIMS_DIR;
+    for (const root of roots.splice(0)) fs.rmSync(root, { recursive: true, force: true });
+  });
+  return () => {
+    const root = mkdtempForTestSync(prefix);
+    roots.push(root);
+    const claimsDir = path.join(root, 'claims');
+    process.env.AGENT_DEVICE_CLAIMS_DIR = claimsDir;
+    const stateDir = path.join(root, 'state');
+    fs.mkdirSync(stateDir, { recursive: true });
+    return { root, stateDir, claimsDir };
+  };
+}
+
+/** Fail-closed reconciler: a test owner is never treated as recoverable. */
+export const retainOrphanedDeviceClaims: DeviceClaimReconciler = async () => ({
+  status: 'retained',
+  reason: 'test-live-owner',
+});
--- a/src/core/command-descriptor/__tests__/device-claim-policy.test.ts
+++ b/src/core/command-descriptor/__tests__/device-claim-policy.test.ts
@@ -1,0 +1,97 @@
+import { test } from 'vitest';
+import assert from 'node:assert/strict';
+import { PUBLIC_COMMANDS } from '../../../command-catalog.ts';
+import { commandDescriptors, resolveCommandDeviceClaimPolicy } from '../registry.ts';
+import type { DeviceClaimPolicy } from '../types.ts';
+
+// #1320 completeness gate. TypeScript already makes the trait required on every
+// raw descriptor, so these tests only pin what types cannot: the CLASSIFICATION,
+// and the structural precondition that makes it enforceable.
+
+function commandsByPolicy(): Partial<Record<DeviceClaimPolicy, string[]>> {
+  const grouped: Partial<Record<DeviceClaimPolicy, string[]>> = {};
+  for (const descriptor of commandDescriptors) {
+    (grouped[descriptor.deviceClaimPolicy] ??= []).push(descriptor.name);
+  }
+  for (const names of Object.values(grouped)) names.sort();
+  return grouped;
+}
+
+test('every public command resolves the policy its descriptor declares', () => {
+  const byName = new Map(commandDescriptors.map((descriptor) => [descriptor.name, descriptor]));
+  for (const command of Object.values(PUBLIC_COMMANDS)) {
+    const descriptor = byName.get(command);
+    assert.ok(descriptor, `public command ${command} is missing from the descriptor registry`);
+    assert.equal(resolveCommandDeviceClaimPolicy(command), descriptor.deviceClaimPolicy);
+  }
+  // Command names outside the registry stay claim-free rather than fail closed
+  // into an acquisition no owner would ever release.
+  assert.equal(resolveCommandDeviceClaimPolicy(undefined), 'require-owner');
+  assert.equal(resolveCommandDeviceClaimPolicy('not-a-registered-command'), 'require-owner');
+});
+
+test('every command that deviates from require-owner is a reviewed, diffable set', () => {
+  // CONSERVATIVE: these lists may only change in the same PR that updates them
+  // here. A `transient-exclusive` command takes host-global exclusive ownership
+  // of its device for one request, so adding one changes cross-worktree
+  // behavior for everyone sharing that device (#1799); `none` is for
+  // host/config-only commands and pure delegators, whose device work runs inside
+  // the request scope of the command they dispatch.
+  const { 'require-owner': _sessionBound, ...deviating } = commandsByPolicy();
+  assert.deepEqual(deviating, {
+    'acquire-session': ['open'],
+    'release-session': ['close'],
+    'transient-exclusive': [
+      'boot',
+      'install',
+      'install_source',
+      'prepare',
+      'push',
+      'reinstall',
+      'shutdown',
+    ],
+    observe: ['apps', 'appstate', 'capabilities', 'device', 'devices', 'doctor'],
+    none: [
+      'artifacts',
+      'auth',
+      'batch',
+      'cdp',
+      'connect',
+      'connection',
+      'daemon',
+      'debug',
+      'disconnect',
+      'install-from-source',
+      'lease_allocate',
+      'lease_heartbeat',
+      'lease_release',
+      'mcp',
+      'metro',
+      'proxy',
+      'react-devtools',
+      'release_materialized_paths',
+      'session',
+      'session_list',
+      'session_save_script',
+      'web',
+    ],
+  });
+});
+
+test('a transient-exclusive command can actually reach the device binding seam', () => {
+  // The claim gate lives on the request scope's device binding, which only ADR
+  // 0019 `device-runtime` commands pass through: an unmigrated (`legacy`)
+  // command reaches its device through dispatch instead, so declaring
+  // `transient-exclusive` there would be a claim nobody ever acquires. Live
+  // verification of #1799 caught exactly that on `keyboard`. Those commands stay
+  // `require-owner` until their platform execution migrates.
+  for (const descriptor of commandDescriptors) {
+    if (descriptor.deviceClaimPolicy !== 'transient-exclusive') continue;
+    assert.ok(descriptor.daemon, `${descriptor.name}: transient-exclusive without a daemon route`);
+    assert.equal(
+      descriptor.platformExecution.kind,
+      'device-runtime',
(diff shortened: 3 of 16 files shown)


thymikee:
Reviewed exact head b807323. Clean code/architecture verdict: claim policy is descriptor-required, transient-exclusive admission sits at the request-runtime binding seam before operations are exposed, legacy classifications are gate-rejected, and the two-daemon regressions plus live Android A/B directly prove #1799 and cleanup. Readiness is currently blocked by exact-head CI: Android Smoke failed in the unrelated live scroll helper (`Active application interaction viewport is unavailable`) and needs a rerun or same-load main comparison before it can be classified; iOS Smoke is still in progress. No `ready-for-human` label while that failure remains confirmed.

thymikee:
Re the Android Smoke failure at b807323: the failing step is `smoke:automation-system` scroll (`Active application interaction viewport is unavailable`, an Android helper `IllegalStateException` during viewport lookup) — nothing on this branch touches the helper, snapshot presentation, or the scroll path; the change only gates `bindDevice` for `transient-exclusive` descriptors (`boot`/`shutdown`/`install*`/`push`/`prepare`), and the smoke session runs through `open` (`acquire-session`) + session-bound commands, which do zero claim-store I/O. Re-ran the failed job to classify it as load flake vs. real; will update here with the result.

thymikee:
Android Smoke rerun on the same head passed → the earlier `Active application interaction viewport is unavailable` was load flake in the live scroll helper, unrelated to this branch. All checks green now.

thymikee:
Tightening pass after review — rebased onto current `main` and made smaller without weakening the guarantee.

**Before → after (this PR's diff vs `main`)**

| | files | +/− |
| --- | --- | --- |
| Before | 23 | +1251 / −88 |
| After (#1809 alone) | 16 | +951 / −83 |
| Split out to #1818 | 8 | +294 / −41 |

What changed:

1. **Split.** The `daemon stop` claim reporting (#1799's third observation) moved to **#1818**, stacked on this branch and opened as a draft; it will be retargeted to `main` after this merges. This PR is now only the descriptor policy, the admission seam, and the boot/shutdown regression.
2. **Reused the binding cache instead of adding one.** `createRequestRuntimeBindings` already memoizes one binding promise per device key, so the claim admission now runs *inside* that binding creation — once per device, deduped by machinery that already existed. `device-claim-admission.ts` lost its own `Map`/pending-promise bookkeeping and now only remembers the ownerships it acquired so it can release them (94 → 83 lines, and the remaining logic is a straight-line acquire/release). `bindExactDevice` bypasses the cache by design, so it admits its own.
3. **Tests folded.** `device-claim-admission.test.ts` (176 → 150) is now one table over all six policies asserting claim-store I/O only for `transient-exclusive`, keeping the foreign-refusal, same-daemon-coverage, provider-skip and `createRequestExecutionScope` construction-guard cases. `device-claim-policy.test.ts` (123 → 97) dropped the assertions TypeScript already proves and collapsed the four bounded sets into one diffable record. Both boot/shutdown regression tests and the claim-store setup now come from shared fixtures (`ANDROID_EMULATOR`, and a new `src/__tests__/test-utils/device-claim-store.ts` used by three test files) instead of per-file copies.
4. **Registry.** The 76 one-line `deviceClaimPolicy:` declarations stay — required-with-no-default is the point, and it matches the `timeoutPolicy` convention. The fallow clone-group warning is resolved at its cause: `fill`/`press`/`longpress` were byte-identical for 14 lines, so they now share a `TARGETED_TOUCH_INTERACTION_TRAITS` bundle next to the existing `GENERIC_MUTATING_LINUX_DEVICE_COMMAND_TRAITS` (which also absorbed the policy for the four commands that spread it). No suppression, no regenerated baseline — `fallow audit` now reports **no issues** in the changed files, where it previously reported one clone group.

`pnpm format && pnpm check:affected --run` green on both branches (432 files / 3684 tests here; 436 / 3700 on #1818).

thymikee:
Clean re-review at 01fcba51: the scope split is coherent—daemon-stop claim reporting moved to #1818, while #1809 retains the descriptor-owned claim policy, runtime-binding admission, transient lifecycle, and the exact foreign-claim boot/shutdown regressions for #1799. No shutdown-reporting files remain in this diff. Exact-head CI and all platform smokes pass; merge state is CLEAN. Ready-for-human.

PR #1681, opened 2026-08-07
perf(cli): keep node:http and ICU off the CLI startup path (-54% warm call)

thymikee:
## Summary

Decomposing the per-invocation floor of a warm CLI call turned up **79ms of startup work that no local command needs**. Removing it cuts a warm `appstate` from **146.4ms to 67.6ms (-53.8%)**.

Two independent causes, both outside the leads that were expected to matter:

1. **`node:http` in the eager import closure.** Four modules reachable from `src/cli.ts` value-imported `node:http`/`node:https`, so every invocation initialized undici — and, under `NODE_USE_SYSTEM_CA=1`, the macOS system trust store behind it. That is ~70ms, mostly *off-CPU* (a blocking `trustd` lookup, invisible to `--cpu-prof`: the profile showed only 65ms of CPU in a 140ms call). The default daemon transport is a plain `node:net` socket and never issues an HTTP request. Both modules now load on demand.

2. **ICU initialization for an internal sort.** `option-schema.ts` sorted its specs with `localeCompare` at module scope. The *first* `localeCompare` in a fresh Node process costs ~5.4ms of collator init; every later one is ~0.002ms. A case-insensitive-then-codepoint comparator reproduces the identical order for the ASCII flag-key vocabulary (verified against all 159 keys).

## Warm-call floor decomposition

Warm `appstate` on a live iOS simulator session, n=10, 5-min load avg < 10, client-side `hrtime` marks correlated against the daemon's own request log:

| Phase | median |
|---|---|
| node boot → bin entry | 13.8 ms |
| `import('./cli.ts')` (eager closure: 34 files / 643 kB) | 32.6 ms |
| **ensureDaemon total** | **3.0 ms** |
|   · read daemon info (fs) | 0.1 ms |
|   · `canConnect` TCP probe (a second connect, discarded) | 1.2 ms |
|   · version + daemon code signature | 1.5 ms |
| **sendRequest total** | **2.0 ms** |
|   · RPC socket connect | 0.5 ms |
|   · request `JSON.stringify` (481 B) | 0.003 ms |
|   · connect → first response byte | 1.2 ms |
|   · response `JSON.parse` (230 B) | 0.2 ms |
| daemon-side `request_start` → `request_success` | 0–1 ms |

[...]

Issue #1801, opened 2026-08-18
A remote caller cannot read the diagnostics log the error names


benjaminkomen:
## Summary

A failed command names a diagnostics log by path. That path is on the **daemon**
host. A remote caller cannot read that file system. So the job that caused the
failure cannot read the one file that explains it.

```
Diagnostics Log: ~/.agent-device/sessions/<id>/requests/<request-id>.ndjson
```

The hint is correct for a local daemon. It is unreachable for a remote daemon. The
message does not say which case you have.

## Why this matters

We drive a remote EAS Simulator session from a Linux GitHub Actions runner. The
runner is the caller. The daemon runs on a macOS host at Expo. Each diagnostic the
error names is on the far side.

So a CI failure gives one line and no route to more. We spent two runs on an
install failure. Its whole evidence was one general sentence. `--debug` adds output
on the caller. It does not give the daemon record that the hint names.

The session also stops with the job. So the file is gone before a person can ask
for it.

## Repro

Use any failing command against a remote daemon:

```
export AGENT_DEVICE_DAEMON_BASE_URL=...
export AGENT_DEVICE_DAEMON_AUTH_TOKEN=...
agent-device wait '@e999' 2000 --platform ios
```

The error names a path below `~/.agent-device/sessions/`. That path is not on the
caller.

## Fix

Each option below solves it. The first is the cheapest to use.

1. Serve the record through the daemon API. Then let `--debug` fetch it. The caller
   holds a base URL and a token, so it needs no new transport.
2. Put the record in the error response, when the request comes from a remote
   caller.
3. As a minimum, find the remote case and stop printing a path the caller cannot
   read. An agent will try to read that path, and lose a turn.

We want option 1. A CI job can then keep the record as a build artifact. That is
what we need for a run that nobody watches.

## Environment

- `agent-device` 0.20.8
- Caller: Linux GitHub Actions runner (`ubuntu-latest`)
- Daemon: remote, EAS Simulator session, iOS

PR #1795, opened 2026-08-17
ci: keep Bundle Size job green on transient GitHub comment failures

thymikee:
## Summary

On PR #1789 (run 32050847506) the size measurement and job summary completed, then `--post-comment` got a `503 No server is currently available` from the GitHub comments API during an incident and the whole `Bundle Size` lane went red. The lane's real job had already succeeded; only the cosmetic comment write failed.

`scripts/size-report.mjs` `--post-comment` is now best-effort against transient GitHub failures, and idempotent under retry:

- **Retry wraps the whole list → write cycle** (`retryTransient(() => syncGitHubComment(...))`), 4 attempts with 1s/2s/4s backoff on **5xx / 429 / network error**. Because every attempt re-lists first, a create whose response was lost but landed server-side is reconciled into a `PATCH` of the marker comment on the next attempt — no duplicate `POST` (addresses the review comment).
- If it still fails, `TransientGitHubError` is caught at the top: `::warning::` annotation, one-line note appended to `$GITHUB_STEP_SUMMARY`, exit 0.
- **Any other non-OK status (401/403/404/422…) throws immediately, no retry** — misconfiguration still fails the job.
- `assertGitHubWriteResponse` removed; it guarded only this comment write, the sole GitHub write in the file.
- `SIZE_REPORT_RETRY_BASE_MS` env override so the regression tests do not sleep through real backoff.

`size.yml` needs no change: it already appends the full report to `$GITHUB_STEP_SUMMARY` before the comment step, so the numbers were never lost — only the comment.

## Validation

New `scripts/__tests__/size-report-post-comment.test.ts` (registered in the `unit-core` project) spawns the real script with a `--import`ed fetch stub driven by a per-call response script, and asserts on exit code, stdout/stderr, `$GITHUB_STEP_SUMMARY`, and the exact sequence of HTTP calls:

| test | asserts |
|---|---|
| uncertain-create reconciliation | `GET, POST(net error), GET(marker now present), PATCH /42` — no second `POST` |
| transient exhaustion | `503×4` → `::warning::` on stdout, note in step summary, exit 0, one list call per attempt |
| fatal 4xx | `GET 200, POST 401` → nonzero exit, `Failed to create PR comment: 401`, no `::warning::`, no retry |

Planted red evidence (script swapped, tests run, script restored):

- **`main`'s script** → reconciliation ✗ (throws on the net error), exhaustion ✗ (exit 1 on 503), fatal-4xx ✓ (guard: fatality preserved).
[...]

PR #1786, opened 2026-08-17
feat(web): add hover command for hover-gated UI (#1783)

thymikee:
Closes #1783.

## Summary

Adds a first-class `hover <x y|@ref|selector> [--settle]` command, admitted on `--platform web` only, so hover-gated UI (a message row's `...` toolbar, menus that open on pointer enter) is reachable through agent-device the way it already was through the underlying `agent-browser` backend.

**Shape — minimal by reuse, not by scope:**
- One new optional op pair, `Interactor.hover?` / `AgentDeviceBackend.hover?` (coordinate: agent-browser `mouse move`, via `movePointer` extracted from `clickCoordinates`) plus `WebProvider.hoverRef` / `AgentDeviceBackend.hoverTarget` (ref: agent-browser `hover <ref-selector>`). Only the web provider implements them; touch platforms need no stubs.
- Rides the existing targeted-touch pipeline: same `@ref`/selector/coordinate targeting, occlusion/off-screen guards, staleness, `--settle` (settle-only, no `--verify` — hover reveals rather than activates), response builder (`gesture: 'hover'`), event timeline, MCP schema, Node client (`interactions.hover`), and script recording (`hover @ref` publishes as a portable selector line).
- `hover @ref` takes the ADR 0011 native-ref path exactly like `click @ref`/`fill @ref` (web ref frames carry no rects, so a coordinate route could never resolve them). The shared preflight → handle-dispatch → exact-ref envelope is extracted into `dispatchNativeRefInteraction` and used by tap, fill, and hover so the three cannot drift; the guarantee matrix `native-ref` row lists `hover`.
- Web-only by capability the same way `viewport` is (`capability: { apple: {}, android: {}, linux: LINUX_NONE }` + `WEB_INTERACTION_COMMANDS`): `capabilities --platform web` lists it; iOS/Android/Linux reject at admission with `UNSUPPORTED_OPERATION` and a `--platform web` hint. `longpress` stays the mobile hold-gesture verb.
- Selector/coordinate hover uses the `resolvedTarget` pipeline (no hittable-ancestor promotion): the pointer only has to enter the matched node's box.

Help (`overview`, `help web`, `help hover`), `commands.md`, and CHANGELOG are updated so help/capabilities agree with what the CLI does. No skill files are touched.

## Live validation

Real CLI + managed agent-browser against a fixture reproducing the issue (chat rows whose Delete button is `display:none` until `:hover`):

```
hover 'text="Second message"' --settle
  → Hovered text="Second message" (155, 172)
[...]

PR #1810, opened 2026-08-18
fix: read replay scripts on the caller and ship them with the request

### LINKED ISSUE #1802: replay cannot take a path against a remote daemon
## Summary

`replay <path>` resolves the path on the caller. Then it opens the file on the
**daemon**. A remote daemon shares no file system with the caller. So the command
fails with `ENOENT`, on a path the caller can read.

The result: `replay` does not work against a remote daemon.

## Why this matters

We drive a remote EAS Simulator session from a Linux GitHub Actions runner. Our
flow files are in the repository, which the runner checks out. The daemon host has
no copy of that repository.

The trap is the local pass. On one machine the caller and the daemon are the same
host, so the command works. A flow that passes locally then fails on its first
remote run. And the failure names a path that exists.

## Repro

```
agent-device replay ./flows/login.ad --platform ios
```

You get `ENOENT` on `./flows/login.ad`, which is on the caller.

`Qp()` in `session.js` (bundled 0.20.8) does the open, on the daemon side.

## Our other route

We use `batch --steps <json>`, which sends the steps as text. Text needs no shared
file system. We wrote a converter of our own. It turns one committed `.ad` flow
into that JSON. So a local `replay` and a remote `batch` run the same steps, from
the same file.

It works. But it puts your flow grammar outside your tool. It holds each step kind,
each flag, and each `${VAR}` substitution again. It covers only the part our flows
use. A grammar change breaks it, with no warning.

## Fix

Make `replay` send the script content. `install <path>` already sends a binary this
way. Let the caller read the file and resolve it, then send the steps. That is what
our converter does by hand. It removes the whole class of "passes locally, ENOENT in
CI".

If that change is too large, give an explicit error. When the daemon is remote and
the path is not on the daemon, name the split. Do not print a bare `ENOENT`.

## Environment

- `agent-device` 0.20.8

thymikee:
## Summary

`replay <path>` and `test <path-or-glob>` now name files on the **caller's** filesystem, everywhere.
The client resolves the path, reads the script (and, for Maestro, every `runFlow` include it can
reach), and sends the text with the request. The daemon executes only what arrived and opens no
caller path at all.

Before: the path was resolved on the caller but opened on the daemon. On one host that works, so a
flow passes locally and then fails on its first remote run with `ENOENT` on a path the caller can
read. Against a remote daemon `replay` simply did not work.

After:

```
agent-device replay ./flows/login.ad --platform ios     # runs; the daemon never sees ./flows
agent-device test './e2e/*.ad' --platform ios           # globs expand on the runner, each source ships its text
agent-device replay ./flows/typo.ad                     # "replay script not found on this machine: ./flows/typo.ad"
```

The one typed concept is a **replay script source bundle** (`CONTEXT.md`, Recording & replay): an
entry display path plus a resolved-path-to-text map. It rides in the request as
`replayScriptSource` (one run) / `replayScriptSources` (one per `test` source), the same way
`replayShellEnv` already carries caller-local input. Local and remote use the identical path — there
is no remote-only branch, which is what stops "passes locally, ENOENT in CI" from coming back.

What moved:

- `runReplayScriptFile` → `runReplayScriptSource`; it reads the bundle, never `req.positionals[0]`.
  `inspectAdReplay` takes script text; `replay-device-selection` reads the bundle (it previously
  swallowed its own `ENOENT` and silently skipped pre-binding on every remote replay).
- `createMaestroProgramLoader` takes a `readSource` port — fs on the caller, bundle lookup on the
  daemon — so `runFlow` includes resolve identically on both. `packages/maestro` no longer imports
  `node:fs`.
- `test` path/glob expansion moved to the caller (`src/replay/source-discovery.ts`); the daemon
  capability now only inspects the sources it was sent, in the order they arrived. Sharding,
  ordering, and platform filtering are unchanged.
- No path fallback. A request carrying only a path is refused with a message naming what is missing
  and what to do, instead of an `ENOENT` for a file that exists on the machine that ran the command.
  Older clients against a newer daemon get that message; the field is additive, so a newer client
  against an older daemon is the normal command-error case ADR 0006 describes.

**`--save-script` against a remote daemon is now refused** with a typed `INVALID_ARGS` naming the
split. Reading was fixable by sending content; writing is not symmetric — the healed `.ad` is
published by the daemon's `SessionScriptWriter` at session teardown, on the daemon's disk, next to a
source path that does not exist there. Returning the healed script in the response would mean new
teardown→response plumbing well outside this change, so option (b) from the design: refuse loudly
now, follow up separately. Local `--save-script` behavior is untouched.

Closes #1802

## Validation

**Regression tests, proven red first.** With only the client writer reverted to `origin/main`,
`src/__tests__/cli-replay-script-sources.test.ts` fails on the wire payload:

```
FAIL  src/__tests__/cli-replay-script-sources.test.ts > replay against a remote daemon sends the script content, not just the path
+ undefined
- '/private/tmp/.../flows/login.ad'
   assert.equal(bundle?.entry, path.join(cwd, 'flows/login.ad'));
```

With the daemon side reverted (`session-replay-runtime{,-plan}.ts`, `ad-replay/inspect.ts`),
`src/daemon/handlers/__tests__/session-replay-script-source.test.ts` fails on both halves — the run
that must not touch the disk, and the refusal that must exist:

```
FAIL  > a bundled replay runs after the script file is gone from the daemon host
  expected false to be true      (the daemon opened the caller path and got ENOENT)
FAIL  > a request carrying only a path is refused with the client-must-send-sources message
  expected true to be false      (a bare path was silently accepted)
```

Also covered: Maestro include collection (transitive, through `repeat`/`retry`/inline
`runFlow.commands` and lifecycle hooks, deduplicated, cycle-safe), a missing include naming its own
path, the bundle byte cap, the caller-side not-found error, and the remote `--save-script` refusal.
`inspectAdReplay`'s plan digest is pinned to a literal so an ADR 0012 `--from`/`--plan-digest`
resume issued before this change stays valid after it.

**Live, iPhone 17 simulator (`1604B975-…`), repo CLI, isolated state dirs.**

Local, relative path from the flow's own directory — `Replayed 3 steps in 3.1s`. A wrong selector
still names the caller's path and line: `Divergence at step 3 (/…/flows/broken.ad:4)`.

Remote mode via `agent-device proxy --port 4319`, driven from a *different* cwd and a *different*
state dir with `AGENT_DEVICE_DAEMON_BASE_URL`/`AGENT_DEVICE_DAEMON_AUTH_TOKEN` — `replay
./settings.ad` → `Replayed 3 steps in 3.2s`. The proof that the daemon opens nothing: the script was
deleted one second after the client started, while the run was still in flight —

```
== removed ./vanishing.ad from the caller's disk (nothing at that path any more)
Replayed 3 steps in 2.9s
ls: vanishing.ad: No such file or directory
```

`replay ./settings.ad --save-script` over the same proxy → `--save-script is not supported against a
remote daemon…`; `replay ./no-such-flow.ad` → `replay script not found on this machine:
./no-such-flow.ad`, with no daemon round-trip.

*Honest limitation of the live run:* a single host cannot reproduce the reporter's original `ENOENT`,
because the daemon shares the caller's filesystem and pre-fix code resolves through `meta.cwd`. The
cross-filesystem property is what the deleted-script run above and the daemon-side regression test
demonstrate.

All sessions closed, both daemons stopped, the proxy killed, and the simulator shut down.
`pnpm check:affected --run` green (`check:daemon-wire-compat`: 151 declarations, 0 changed — the new
fields sit inside the `CommandFlags`/`InternalRequestOptions` waiver ADR 0006's additive rule
covers). One unrelated contention flake on the way (`scripts/fuzz/harness.test.ts` 5s timeout);
passes in isolation and on the final green run.

## Tradeoffs and follow-ups

- A Maestro `runFlow` include whose path is templated on a value only the daemon binds cannot be
  collected before the run. Collection is best effort past the entry flow (`when:`-guarded includes
  may never execute), and a run that does need such a file fails naming exactly which file was
  missing from the sources sent. Includes templated on `env:`/`AD_VAR_*`/`-e` values *are* resolved,
  through the Maestro package's own interpolation.
- Bundles are inlined in the request and capped at 2 MB total; `.ad` scripts and flows are kilobytes,
  and the upload route stays reserved for binaries.
- `--save-script` remote support (returning the healed script in the teardown response) is the
  follow-up left open.

## Scope

89 files. Scope stayed in the replay/test command family and its packages, with three deliberate
edges: the remote request-prep gate (`--save-script` refusal), the `@agent-device/maestro` source
port, and `@agent-device/replay-test`'s discovery port, which no longer takes inputs or a cwd.

**A pure refactor has been split out** in response to review: moving the `test`-suite orchestration
out of `handleSessionReplayCommands` was not needed by this fix, and now lives in its own stacked PR
(#1826, `session-replay.ts` 413 -> 34 lines, `session-test-suite-command.ts` +410). Removing it took
this PR from 91 files / +2646 -1208 to **89 files / +2268 -852** — 356 fewer added and 356 fewer
removed lines, all of it moved code. What is left is the fix and what the fix itself requires.

Four commits: the fix; a CodeQL follow-up replacing a hand-escaped regex in a new test with a
substring assertion; the CLI cold-start fix below; and one small refactor the fix does require —
the new replay flags appear in both `CliFlags` and `CommandExecutionOptions`, which fallow flags as
a clone, so `ReplayRequestFields` declares them once, and the suite handler's missing-sources
rejection travels the typed-error path its sibling rejections already use so the fix adds no branch
to `handleSessionReplayCommands`. Docs: `CONTEXT.md` glossary, `agent-device help scripting`, and
the command reference.

`session-replay.ts` stays over the 300-line tripwire it was already over before this PR (382 -> 413
lines); #1826 is the follow-up that fixes that, and is deliberately not required for this one.

## CLI cold start

The size job caught a real regression in the first version of this change: the command registry
evaluates *every* command family's module on startup, `--help` included, so a static
`@agent-device/maestro` import in the new script-source builder put the YAML parser on the cold
path — **`CLI --help` 67.4 ms -> 95.7 ms (+28.3)** and a 131 kB startup chunk, for a format most
invocations never touch.

The engine now loads on demand inside `loadReplayScriptSourceBundle`, reached only once an entry
resolves to a flow; the native `.ad` branch reads its entry file and nothing else. That made the one
bundle constructor async, so `DaemonWriter` gained a sibling `AsyncDaemonWriter` — replay/test and
batch declare it, every other family's writer stays synchronous and untouched, and the registry
awaits whichever kind a command declares. One construction path is preserved: CLI, Node client and
MCP still converge on `loadReplayScriptSourceBundle`.

CI's own base-vs-PR run after the fix: **`CLI --help` 65.3 -> 66.2 ms (+0.8)**, with `CLI --version`
moving +0.7 ms on the same run — and `--version` cannot be affected by this change, so the remainder
is that run's drift. Package total is +10.0 kB raw / +3.1 kB gzip.

`src/__tests__/cli-startup-import-closure.test.ts` — the AST walker that already keeps `node:http`
off the cold path (#1681) — now holds the same line for the engine, and was proven red by restoring
the static import. Walking the *built* output confirms it independently of wall clock: the Maestro
chunk is absent from `dist/src/cli.js`'s 46-chunk static closure and reachable only through
`script-source-bundle.js -> src4.js`'s dynamic import.

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -247,6 +247,11 @@ task touches:
 - Destination guard: portable selector-targeted `wait` near the end of an open-to-destination script
   that confirms a landmark on the ready destination screen before replay hands the live session to
   its caller.
+- Replay script source bundle: every script file one `replay`/`test` run needs, read and resolved by
+  the CALLER and shipped inside the request — an entry display path plus a resolved-path-to-text map
+  covering the `.ad` script or the Maestro flow and its `runFlow` includes. The daemon executes only
+  what the bundle carries and resolves no caller path, so a local run and a run against a remote
+  daemon read identical bytes (#1802). Avoid: script upload, flow payload.
 - Screen-recording facet: runtime facet that starts platform screen/video capture and returns a live
   handle plus its durable descriptor. It is distinct from script recording.
 - Live resource handle: process-local authority to finish or forcibly dispose active app-log,
--- a/packages/ad-replay/src/internal/__tests__/inspect.test.ts
+++ b/packages/ad-replay/src/internal/__tests__/inspect.test.ts
@@ -1,0 +1,45 @@
+import assert from 'node:assert/strict';
+import { test } from 'vitest';
+import { AppError } from '@agent-device/kernel/errors';
+import { inspectAdReplay } from '../inspect.ts';
+
+const SCRIPT = 'context platform=ios\nopen "Demo"\nclick label="Save"\n';
+
+test('inspectAdReplay parses script text into actions, a line table and header metadata', () => {
+  const manifest = inspectAdReplay(SCRIPT);
+
+  assert.deepEqual(
+    manifest.actions.map((action) => action.command),
+    ['open', 'click'],
+  );
+  assert.deepEqual(manifest.actionLines, [2, 3]);
+  assert.equal(manifest.metadata.platform, 'ios');
+});
+
+/**
+ * ADR 0012's `--from`/`--plan-digest` resume quotes a digest back from a divergence report, so the
+ * digest for a given script must not move. #1802 changed only HOW the script reaches the daemon
+ * (a replay script source bundle instead of a path it opened itself); the digest is computed over
+ * the same text and stays byte-identical, which is what keeps a resume issued before the change
+ * valid after it.
+ */
+test('inspectAdReplay pins the plan digest for a known script', () => {
+  assert.equal(
+    inspectAdReplay(SCRIPT).planDigest,
+    inspectAdReplay(SCRIPT, { platform: 'ios' }).planDigest,
+  );
+  assert.equal(
+    inspectAdReplay(SCRIPT).planDigest,
+    '0641236777b11822d90d446022965629ec99ac2ba1af2d5c637f06dd684fe695',
+  );
+});
+
+test('inspectAdReplay rejects a legacy JSON replay payload', () => {
+  assert.throws(
+    () => inspectAdReplay('[{"command":"open"}]'),
+    (error: unknown) =>
+      error instanceof AppError &&
+      error.code === 'INVALID_ARGS' &&
+      error.message.includes('JSON replay payloads are no longer supported'),
+  );
+});
--- a/packages/ad-replay/src/internal/__tests__/step-loop.test.ts
+++ b/packages/ad-replay/src/internal/__tests__/step-loop.test.ts
@@ -11,8 +11,8 @@ import type { TargetAnnotationV1 } from '@agent-device/contracts/replay';
  * terminal-close suppression) is engine-private — never re-exported by the
  * façade (`packages/ad-replay/src/index.ts`) — so these tests exercise it
  * only through `runAdReplay` itself, the same way the daemon's own
- * `session-replay-runtime.ts` (`runReplayScriptFile`) does. The equivalent
- * daemon-level assertions (full `SessionStore`/`runReplayScriptFile` round
+ * `session-replay-runtime.ts` (`runReplayScriptSource`) does. The equivalent
+ * daemon-level assertions (full `SessionStore`/`runReplayScriptSource` round
  * trip, including the `--keep-session` live-session postcondition) live in
  * `src/daemon/handlers/__tests__/session-replay-runtime-keep-session.test.ts`
  * (renamed from `session-replay-terminal-lifecycle.test.ts` by the #1555
--- a/packages/ad-replay/src/internal/inspect.ts
+++ b/packages/ad-replay/src/internal/inspect.ts
@@ -1,4 +1,3 @@
-import fs from 'node:fs';
 import { AppError } from '@agent-device/kernel/errors';
 import type { SessionAction } from '@agent-device/contracts/session';
 import {
@@ -12,7 +11,7 @@ import { resolveReplayEntryIndex, type PendingRecordAndHeal } from './resume.ts'
 
 /**
  * #1478 P5 stage C2b: the read-only `.ad` inspection façade. Moved out of
- * `session-replay-runtime.ts`'s old `parseReplayScript` (the fs read + the
+ * `session-replay-runtime.ts`'s old `parseReplayScript` (the
  * legacy-JSON-payload rejection it guarded) plus the `parseReplayInput`
  * composition (`src/compat/replay-input.ts`) it fed into — this is the same
  * `parseReplayScriptDetailed` + `readReplayScriptMetadata` pair
@@ -50,12 +49,14 @@ export type AdReplayManifest = Readonly<{
 export type AdReplayDigestFlags = Readonly<{ platform?: string; target?: string }>;
 
 /**
- * Reads `sourcePath` once and returns its parsed actions/line table, header
- * metadata, plan digest, and resume-index resolver. Throws
+ * Parses one `.ad` script's TEXT and returns its actions/line table, header
+ * metadata, plan digest, and resume-index resolver. Takes the script itself,
+ * never a path: #1802 made the CALLER read every script file a replay run
+ * needs, so nothing below this façade opens a file. Throws
  * `AppError('INVALID_ARGS', …)` for the one source format `.ad` replay no
  * longer accepts — a legacy JSON replay payload — matching the daemon's
  * prior explicit rejection exactly. Callers do not need to check for this
- * case separately: `runReplayScriptFile`'s top-level catch (`asAppError`)
+ * case separately: `runReplayScriptSource`'s top-level catch (`asAppError`)
  * maps a thrown `AppError` straight to the same `errorResponse` the old
  * explicit branch built, so this is not a behavior change, only where the
  * check lives.
@@ -67,10 +68,9 @@ export type AdReplayDigestFlags = Readonly<{ platform?: string; target?: string 
  * `open` wins; absent that, the `context platform=`/`target=` header line.
  */
 export function inspectAdReplay(
-  sourcePath: string,
+  script: string,
   digestFlags?: AdReplayDigestFlags,
 ): AdReplayManifest {
-  const script = fs.readFileSync(sourcePath, 'utf8');
   const firstNonWhitespace = script.trimStart()[0];
   if (firstNonWhitespace === '{' || firstNonWhitespace === '[') {
     throw new AppError(
--- a/packages/ad-replay/src/internal/step-loop.ts
+++ b/packages/ad-replay/src/internal/step-loop.ts
@@ -71,7 +71,7 @@ import type {
  *
  * #1478 P5 follow-up (one daemon-owned artifact ledger): artifact-path
  * accumulation used to be DOUBLE-WRITTEN — `dispatchStep` added each step's
- * entries to the daemon's own `Set` (`runReplayScriptFile`'s, read by its
+ * entries to the daemon's own `Set` (`runReplayScriptSource`'s, read by its
  * catch block so a mid-loop throw still reports what was collected) AND
  * returned them for this loop to add to a second `Set` of its own. Two
  * mutable collections, kept in sync by hand, with no single owner. The
--- a/packages/contracts/src/cli-flags.ts
+++ b/packages/contracts/src/cli-flags.ts
@@ -20,6 +20,7 @@ import type {
 } from './remote-config-fields.ts';
 import type { ScreenshotRequestFlags } from './screenshot.ts';
 import type { RecordingScope } from './recording-scope.ts';
+import type { ReplayScriptSourceBundle } from './replay.ts';
 
 export type CliFlags = CloudProviderProfileFields &
   RemoteConfigMetroOptions &
@@ -143,6 +144,13 @@ export type CliFlags = CloudProviderProfileFields &
     replayMaestro?: boolean;
     replayEnv?: string[];
     replayShellEnv?: Record<string, string>;
+    /**
+     * #1802: the caller-read script text `replay` executes. The daemon never
+     * opens a caller path, so this is the ONLY source a replay run reads.
+     */
+    replayScriptSource?: ReplayScriptSourceBundle;
+    /** #1802: `test`'s caller-side discovery result — one bundle per discovered source, in run order. */
+    replayScriptSources?: ReplayScriptSourceBundle[];
     replayFrom?: number;
     replayPlanDigest?: string;
     /** Replay: leave the session active by suppressing an authored terminal close in native .ad. */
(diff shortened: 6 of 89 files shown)


thymikee:
Reviewed exact head b7c66b8. The caller-owned replay source-bundle design is sound: replay/test payload preparation, daemon no-filesystem execution, Maestro include closure, ADR 0006 additive fields, and ADR 0012 digest/repair compatibility are substantively covered. One readiness blocker: CodeQL reports a new high-severity incomplete-escaping alert at `src/__tests__/cli-client-commands.test.ts:759`, where `new RegExp(missingPath.replace(/[.]/g, '\\.'))` escapes only dots. It is test-only, but the security gate is genuinely red; use a substring assertion or complete regex escaping and refresh CodeQL. Linux/iOS smoke are still running. Residual evidence: the live remote run covers `replay`, while the newly caller-side remote `test` path is unit-covered but not live-verified.

thymikee:
CodeQL alert addressed at head 6ae91bc32 (`test: assert the caller-side replay path as a substring, not a hand-escaped regex`) — the js-ts analysis on that head is green and the alert is closed. Remaining residual evidence as stated in the PR body: remote `test` (caller-side glob expansion) is unit-covered but not live-verified; remote `replay` was.

thymikee:
### Fixed the CLI startup regression the size report caught

The size bot was right, and the cause was mine: the command registry evaluates **every** command family's module on startup — `--help` included — so `src/commands/replay/index.ts` statically importing the new script-source builder dragged `@agent-device/maestro` (and the YAML parser behind it) onto the cold path, for a format most invocations never touch.

`@agent-device/maestro` now loads **on demand**, inside `loadReplayScriptSourceBundle`, and only once an entry actually resolves to a flow (`resolveReplayFormat(...) === 'maestro'`). The native `.ad` branch reads its entry file and touches nothing else. That made the one bundle constructor async, so `DaemonWriter` gained a sibling `AsyncDaemonWriter` — replay/test and batch declare it, every other family's writer stays synchronous and untouched, and the registry awaits whichever kind a command declares. `src/replay/source-discovery.ts` and `resolveReplayFormat` were already clean (`node:fs`/`node:path` and string work only); no other engine reaches the startup path.

**Guarded, not just fixed.** `src/__tests__/cli-startup-import-closure.test.ts` — the existing AST walker that keeps `node:http`/`node:https` off the cold path (#1681) — now holds the same line for the Maestro engine. Proven red by restoring the static import:

```
FAIL > the CLI startup import closure never evaluates the Maestro engine
- []
+ [ "replay/script-source-bundle.ts -> @agent-device/maestro" ]
```

**Numbers** — matched builds on one machine, 15 runs each, via the same `scripts/size-report.mjs` the size job uses:

| Scenario | Before | After | Diff |
|---|---:|---:|---:|
| `CLI --help` (median) | 60.4 ms | 47.2 ms | **−13.2 ms** |
| `CLI --help` (min) | 57.7 ms | 44.3 ms | −13.4 ms |
| `CLI --version` (median) | 35.3 ms | 29.0 ms | −6.3 ms |

`--version` cannot be affected by this change, so treat its −6.3 ms as the machine's noise floor and the `--help` gain as roughly 7–13 ms locally; CI's base-vs-PR comparison on a quiet runner is the number to trust.

Chunks: the engine left the startup graph entirely and became its own on-demand chunk.

| Chunk | Before | After |
|---|---:|---:|
| `dist/src/script-source-bundle.js` | 131.0 kB | 108.6 kB |
| `dist/src/src4.js` (Maestro + YAML) | — | 59.7 kB, dynamic only |

**Wall-clock-free confirmation**, walking the static import graph of the *built* output from `dist/src/cli.js`:

```
static closure of dist/src/cli.js: 46 chunks
maestro engine statically reachable: NO
maestro engine behind a dynamic import: dist/src/script-source-bundle.js -> dist/src/src4.js
```

(The engine is located by its own `Maestro runFlow cycle detected` string, so this identifies the real chunk rather than trusting a name.)

`pnpm check:affected --run` green; the full unit suite passes (583 files / 4270 tests).

thymikee:
### CI confirms it: startup regression gone

The size job's own base-vs-PR run on a quiet runner, after `ac2b9ee`:

| Scenario | Base | Current | Diff |
|---|---:|---:|---:|
| `CLI --help` | 65.3 ms | 66.2 ms | **+0.8 ms** |
| `CLI --version` | 26.0 ms | 26.8 ms | +0.7 ms |

Was **67.4 → 95.7 ms (+28.3)**. `--version` moved by the same +0.7 ms and cannot be affected by this change, so the remaining +0.8 ms on `--help` is that run's drift, not a cost.

Package total is now +10.0 kB raw / **+3.1 kB gzip** (was +131 kB of startup chunk): `script-source-bundle.js` +108.6 kB and the Maestro chunk `src4.js` +59.7 kB are largely offset by `daemon.js` −44.3 kB, `session2.js` −17.6 kB and `screenshot-result.js` −12.2 kB as the split moved code off shared chunks. `src4.js` is reachable only through the dynamic import, so none of it is on the cold path.

thymikee:
Code delta review at ac2b9ee is clean: the CodeQL fix is resolved, Maestro loads only for flow input, async writer propagation reaches every production caller, the planted import-closure regression is meaningful, size evidence restores CLI startup, and exact-head CI is green. Readiness remains blocked because the PR body was overwritten with the unrelated #1801 remote-diagnostics feature: wrong linked issue, behavior, 38-file scope, routes, validation, and wire claims. Restore an accurate #1802 replay-source description, including the remote `replay` proof and remote-`test` live-evidence limitation, before ready-for-human.

thymikee:
### PR description restored

Apologies for the confusion — the description on this PR was briefly overwritten with the text of a **different** PR (#1801, remote diagnostics). That was a stray `gh pr edit` from a parallel session landing on the wrong PR number; nothing about the branch, its commits, or its diff was touched.

The body is now the #1802 replay-script-source-bundle description again, with the CLI cold-start section re-added after the restore (it had been written after the point the restore was taken from). Verified before and after editing: `Closes #1802` present, the summary is the replay-source-bundle one, and no remote-diagnostics text remains.

Also, the branch has been rebased onto current `main` (through #1795) — head is now `7018c3d5f`. The rebase was clean; I checked line-level that it reverted nothing from the 9 new commits: this branch changes **0 lines** in `src/core/command-descriptor/registry.ts`, `.oxfmtrc.json` and `src/platforms/apple`, and of the four files touched by both sides the only removed lines are the two this PR deliberately replaces. #1786's `hover` family is a plain synchronous `DaemonWriter`, so the `AsyncDaemonWriter` split leaves it untouched.

Re-run on the rebased head: typecheck clean, wire-compat + replay-compat green (`151 declarations, 0 changed, 0 removed, 0 added`), `pnpm check:affected --run` green.

thymikee:
Clean re-review at 7018c3d5: the rebased series is range-diff equivalent to the previously reviewed code, and the PR body is restored to the actual #1802 replay-source change with correct 91-file scope, validation, cold-start fix, and remote-test evidence limitation. Exact-head CI is fully green and merge state is CLEAN. Ready-for-human.

thymikee:
### Split the pure refactor out — this PR is smaller

Fair challenge, and you were right about which part. Moving the `test`-suite orchestration out of `handleSessionReplayCommands` was a tidy-up I did while in the file; the fix never needed it. It is now its own stacked PR, **#1826** (draft, based on this branch, retargets to `main` once this merges).

| | Files | Added | Removed |
|---|---:|---:|---:|
| Before | 91 | 2646 | 1208 |
| After | **89** | **2268** | **852** |
| Difference | −2 | −378 | −356 |

Almost all of it is the move: `session-replay.ts` 413 → 34 lines against `session-test-suite-command.ts` +410, plus the two test files that follow the two symbols it takes with it.

What stayed, because the fix genuinely requires it:

- **`ReplayRequestFields`** — the new replay flags have to appear in both `CliFlags` and `CommandExecutionOptions`, which fallow flags as a 19-line clone. Declaring them once is the fix's own footprint, not a drive-by.
- **A one-line rerouting** of the suite handler's missing-sources rejection through the typed-error path its sibling rejections already take, instead of adding a second refusal shape. That is what keeps `handleSessionReplayCommands` under the complexity gate *without* the extraction — I checked: with the branch added naively it goes to 14 cyclomatic and fallow fails; routed this way it stays where it was, and `fallow audit --base origin/main` reports **no issues in 89 changed files**.

So the extraction is genuinely optional for this PR, which is the test it had to pass to be split out.

`session-replay.ts` remains over the 300-line tripwire — it was already over before this PR (382 → 413), and #1826 is the follow-up that actually fixes that. Flagging rather than hiding it.

Also rebased onto current `main` (through #1809/#1817/#1816); head is now `1dcd0fcfe`. `pnpm format && pnpm check:affected --run` green on both branches.

thymikee:
Re-reviewed exact head `1dcd0fcf` after trimming: code and architecture are clean. The unrelated suite-orchestration move is split to #1826; this PR’s 89-file scope now matches its body and retains only the caller-side replay/test source-bundle path and required edges. The previous CodeQL regex alert is fixed, startup laziness remains guarded, and the deleted-source/bare-path/include regressions plus live proxy run exercise the shipped route. All completed checks are green; final merge readiness waits for Coverage, Affected-check Selector, and iOS Smoke still pending.

Issue #1035, opened 2026-07-03
Bare ref positional produces cryptic coordinate error instead of a hint


thymikee:
## Problem

`agent-device click e3` (missing the `@`) fails with:

```
UNKNOWN: Expected x to be a finite number.
hint: Retry with --debug and inspect diagnostics log for details.
```

The parser falls through to coordinate parsing and surfaces an internal `requiredNumber` error (`src/commands/command-input.ts`) with code `UNKNOWN`. The actual mistake is one missing character, and the message sends the caller (human or small-model agent) off debugging coordinates.

## Fix

When a click/press/fill/get positional matches `^e\d+$` (or the current snapshot's ref shape) and coordinate parsing fails, return `INVALID_ARGS` with `Did you mean "@e3"?`. Small, contained in target parsing.

Hit twice in real agent-driving sessions; this is the kind of papercut Haiku-class agents burn turns on.

PR #1813, opened 2026-08-18
fix: reject selector-shaped wait arguments instead of reading them as text

### LINKED ISSUE #1800: wait accepts an invalid argument shape as literal text, then times out
## Summary

`wait` takes an argument shape that its grammar does not hold. It joins the
arguments into one text string. Then it waits for that string on the screen. The
wait always times out. The message does not tell you the shape was wrong.

We wrote this command:

```
agent-device wait visible 'label="Open"' 25000 --platform ios
```

The command does not refuse it. It reports:

```
Error (COMMAND_FAILED): wait timed out for text: visible label="Open"
```

No screen holds that text. So the wait can only time out. This is true on 0.20.8.

## Why this costs time

The failure looks the same as an absent element. Our CI step read the exit status.
It wrote `'Open' did not appear`. Then it continued, as we designed it to do. But
the dialog stayed on the screen for eleven more minutes. Each later step ran
against a covered app. The step was green.

An agent that reads this message does not suspect its own arguments. It looks for
the element.

We lost two runs of 20 minutes to this.

## Repro

```
open com.apple.Preferences
agent-device wait visible 'label="General"' 5000
```

Expected: a complaint about the shape.
Actual: `wait timed out for text: visible label="General"`, after the full timeout.

## Fix

Refuse an unknown first word with `INVALID_ARGS`. #1035 does this for a bare `e3`
positional. `wait` lists its forms: `<ms>`, `text <text>`, `@ref`, a selector, and
`stable`. A first word outside that list is a caller mistake. It is not screen
text.

Two signals make the hint exact:

- A first word such as `visible`, `hidden`, or `exists` shows the caller wanted a
  condition. Point at the selector form.
- A value that holds `="` is a selector. It is not text.

This is the same defect class as #1035, which you fixed. A positional falls through
to a more permissive parser. Then the caller gets a message about the wrong thing.
#1035 says it was "hit twice in real agent-driving sessions".

The `wait` case is worse than the `click` case. The wrong parse exits non-zero
after a full timeout. So it reads as a true negative result.

## Environment

- `agent-device` 0.20.8

thymikee:
## Summary

`wait` accepted an argument shape its grammar does not hold, joined the leftover positionals into one literal-text string, and then waited out the **full** timeout for text that could never appear on screen — reading as a false "element absent" instead of the caller's own mistake.

Before (0.20.10):
```
$ agent-device wait open 'label="Open"' 25000
Error (COMMAND_FAILED): wait timed out for text: open label="Open"
```
(after the full 25s timeout, exit code non-zero but read as "element not found")

After:
```
$ agent-device wait open 'label="Open"' 25000
Error (INVALID_ARGS): "open label="Open"" mixes plain word "open" with selector-shaped
"label="Open"", so it is neither a valid selector nor safe as literal text. "open" is not
a recognized selector key (id, role, text, label, value, appname, windowtitle, visible,
hidden, editable, selected, focused, enabled, hittable). Use wait '<selector>' [timeoutMs]
for a selector, or wait text 'open label="Open"' [timeoutMs] to wait for the literal text
"open label="Open"".
```
(immediate, before any device work)

`parseWaitPositionals` (`src/core/wait-positionals.ts`) now returns a typed `invalid`
variant whenever, after stripping the trailing timeout, a positional token is
selector-shaped — a recognized selector key (`isSelectorToken`), or an unrecognized
`key=value` (`detectUnknownSelectorKeyToken`) — but the token list does not form a
valid selector expression, or a valid selector prefix is followed by unquoted trailing
tokens. The message:
- names the offending token and, for an unrecognized key, lists the supported keys plus
  a `label=`/`role=` "did you mean" hint (mirrors #1035's click/press/fill/get fix);
- points a condition word (`exists`, `present`, `appears`, `gone`, `disappears`) at the
  selector form, e.g. `wait 'visible label="Open"' [timeoutMs]` — `visible`/`hidden`
  are already selector keys, so they parse correctly and are unaffected;
- always offers the explicit escape hatch: `wait text '<text>' [timeoutMs]`.

Bare text (`wait Continue 1500`, `wait Sign in 2000`) and the explicit `text` keyword
form are untouched — including the documented boundary that a single word matching a
recognized selector key (e.g. `wait id 3000`) is now rejected rather than silently
read as literal text, since it must go through `wait text 'id' 3000` to disambiguate.

`resolveWaitBudgetMs` (used by the command descriptor's timeout-policy budget) returns
`null` for a rejected list, same as an unparseable one. `dispatchWaitViaRuntime`
(`src/daemon/selector-runtime.ts`) — the one seam both a live CLI wait and a replayed
`.ad` wait step dispatch through — rejects before any session/device work. Excluding
`invalid` from the type `toWaitTarget` accepts makes its kind-by-kind narrowing
exhaustive at compile time, with no runtime fallback branch.

Closes #1800

## Validation

- New `src/core/wait-positionals.test.ts`: the issue's repro shapes (`open`/`exists` +
  selector value), an unknown `key=value` token, a valid selector prefix followed by an
  unquoted trailing word, `visible`/bare-text/`text`-keyword forms staying unaffected,
  the bare-recognized-key-word boundary (`wait id 3000`), and two fast-check properties
  — a generated valid selector expression never parses as `text`, and any token list
  containing a `key=value`-shaped token never parses as `text`.
- Proved red: stashed the three production files (`wait-positionals.ts`, `wait.ts`,
  `selector-runtime.ts`) and reran the new/updated tests against pre-fix code. All 9
  assertions failed as expected — 7 asserted `'text'`/a stale budget instead of
  `'invalid'`, the property test found `["", 'editable="!"', 0]` as a counterexample in
  one run, and the new `dispatchWaitViaRuntime` daemon-dispatch test **hung to the
  5000ms test timeout** (the pre-fix code actually dispatches and polls instead of
  rejecting immediately) — confirming this is exactly the "silent full-timeout" defect
  class from the issue, not a vacuous pin. Restored the fix; all 45 tests pass.
- Updated the pre-existing `snapshot.test.ts` case that had codified the bug
  (`parseWaitArgs falls back to text when selector-like token is invalid`, using
  `foo=bar` — an unrecognized key) to assert the new `invalid` rejection.
- Added one test at the shared daemon-dispatch seam
  (`wait-landmark-recording.test.ts`) proving `dispatchWaitViaRuntime` — the route both
  a live CLI wait and a replayed `.ad` wait step call — rejects before any
  session/device work, since neither the `.ad` script grammar
  (`packages/ad-script`) nor `test/replay-compat` interpret wait's positional shape at
  parse time (`pnpm exec vitest run --project unit-core test/replay-compat` — all 55
  cases pass unchanged, no corpus verdict moved).
- `pnpm check:affected --run` green (format, lint, typecheck, layering, fallow, build,
  and the full `vitest-related` selection — 424 files / 3651 tests). `wait-positionals.ts`
  isn't a mutation-kernel module (`scripts/mutation/modules.ts`) and I didn't touch
  `packages/selectors/src/**`, so the mutation gate wasn't selected — confirmed by the
  selector's own output.
- Live iOS Simulator (`iPhone 17 Pro`, isolated `--state-dir`, `com.apple.Preferences`):
  - `wait open 'label="General"' 3000` → immediate `INVALID_ARGS` in 0.32s (not a 3s
    timeout), quoting the same message shape as above.
  - `wait exists 'label="General"' 3000` → immediate `INVALID_ARGS` in 0.068s, with the
    condition-word hint (`wait 'visible label="General"' [timeoutMs]`).
  - `wait 'label="General"' 3000` → correctly dispatched and got a legitimate
    `AMBIGUOUS_MATCH` (multiple "General" nodes on the live Settings screen — pre-existing,
    by-design direct-iOS-selector behavior, unrelated to this fix); `wait 'role="button"
    label="General"' 3000` (disambiguated) → exit 0.
  - `wait General 3000` (bare text) → exit 0.
  - `wait text 'General' 3000` (explicit form) → exit 0.
  - Session closed, daemon stopped, simulator shut down; no stray processes remained for
    the udid afterward.

Docs: no `website/docs`/`docs/adr` change needed — the accepted `wait` forms are
unchanged, only a previously-silent-degrading input now gets a typed refusal instead of
a timeout (same precedent as #1035's click "Did you mean" fix, which also didn't touch
docs).

6 files changed (5 touched, 1 new test file); scope stayed within the `wait` command
family (`src/core/wait-positionals.ts`, its CLI reader, and the one daemon dispatch
seam) plus test updates — no scope expansion.

### DIFF
--- a/src/commands/capture/wait.ts
+++ b/src/commands/capture/wait.ts
@@ -113,6 +113,9 @@ function readWaitOptionsFromPositionals(
       ...readTimeoutOption(parsed.timeoutMs),
     };
   }
+  if (parsed.kind === 'invalid') {
+    throw new AppError('INVALID_ARGS', parsed.message);
+  }
   return {
     ...base,
     selector: parsed.selectorExpression,
--- a/src/core/wait-positionals.test.ts
+++ b/src/core/wait-positionals.test.ts
@@ -1,0 +1,194 @@
+/**
+ * #1800: a positional list shaped like a selector (a recognized key, or an
+ * unrecognized `key=value`) must never silently degrade to a `text` wait —
+ * that degradation is exactly the shape that reads as "element absent" after
+ * a full timeout instead of the caller's own argument mistake. See #1035 for
+ * the sibling fix on click/press/fill/get.
+ */
+import assert from 'node:assert/strict';
+import fc from 'fast-check';
+import { test } from 'vitest';
+import { isValidSelectorExpression, SELECTOR_KEY_NAMES } from '@agent-device/selectors';
+import { PROPERTY_RUNS } from '../__tests__/test-utils/index.ts';
+import { parseWaitPositionals, resolveWaitBudgetMs } from './wait-positionals.ts';
+
+function assertInvalid(args: string[], messageFragment: string) {
+  const result = parseWaitPositionals(args);
+  assert.ok(result, `expected a result for ${JSON.stringify(args)}`);
+  assert.equal(result.kind, 'invalid');
+  if (result.kind !== 'invalid') return;
+  assert.ok(
+    result.message.includes(messageFragment),
+    `expected message to include ${JSON.stringify(messageFragment)}, got ${JSON.stringify(result.message)}`,
+  );
+}
+
+// --- the issue's reproduction shapes -----------------------------------
+
+test('a condition word followed by a selector-shaped value is rejected, not read as text (#1800 repro)', () => {
+  const result = parseWaitPositionals(['open', 'label="Open"', '25000']);
+  assert.ok(result);
+  assert.notEqual(result.kind, 'text');
+  assertInvalid(['open', 'label="Open"', '25000'], 'label="Open"');
+});
+
+test('"exists" followed by a selector-shaped value is rejected and points at the selector form', () => {
+  assertInvalid(['exists', 'label="x"', '100'], "wait 'visible");
+});
+
+test('unknown selector key is rejected with the supported-key list, not read as text', () => {
+  const result = parseWaitPositionals(['lbl="x"', '100']);
+  assert.ok(result);
+  assert.equal(result.kind, 'invalid');
+  if (result.kind !== 'invalid') return;
+  assert.equal(result.reason, 'unknown-selector-key');
+  assert.ok(result.message.includes('lbl'));
+  assert.ok(result.message.includes('label='));
+});
+
+// --- still-valid selector forms are untouched ---------------------------
+
+test('a valid selector still parses as selector', () => {
+  const result = parseWaitPositionals(['label="Open"', '5000']);
+  assert.deepEqual(result, {
+    kind: 'selector',
+    selectorExpression: 'label="Open"',
+    timeoutMs: 5000,
+  });
+});
+
+test('"visible" leading a valid selector still parses as selector (boolean key, not a condition word)', () => {
+  const result = parseWaitPositionals(['visible', 'label="Open"', '25000']);
+  assert.deepEqual(result, {
+    kind: 'selector',
+    selectorExpression: 'visible label="Open"',
+    timeoutMs: 25000,
+  });
+});
+
+// --- trailing tokens after a valid selector prefix are rejected too -----
+
+test('a valid selector prefix followed by an unquoted extra word is rejected, not merged into text', () => {
+  assertInvalid(['label="Open"', 'foo', '5000'], 'extra arguments');
+});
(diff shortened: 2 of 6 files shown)


thymikee:
Clean review at b06bf1fb: selector-shaped wait inputs are now rejected at both the owning positional parser and the shared daemon dispatch seam before device work. The typed invalid variant, timeout-budget handling, production-path regression, planted-red evidence, and live iOS validation are sound. All completed gates are green; only iOS Smoke remains pending. Code review is ready-for-human.

PR #1815, opened 2026-08-18
fix(daemon): reject '.', '..' and empty session names so session artifacts stay under sessions/

thymikee:
## Problem

\`safeSessionName\` only rewrites characters outside \`[a-zA-Z0-9._-]\`, so the names \`.\` and \`..\` survive unchanged. \`SessionStore.resolveSessionDir\` then does \`path.join(sessionsDir, '..')\` → the daemon state dir, and \`app.log\` / \`runner.log\` / \`requests/*.ndjson\` are written there. For a remote daemon this lets a caller steer daemon-side artifact writes outside the sessions tree with \`--session ..\`.

## Fix — at the owning interface

\`SessionStore.resolveSessionDir\` is the one place a session name becomes a directory (AGENTS.md: session artifact paths come from \`session-store.ts\`), so the invariant "every session dir lies beneath \`sessionsDir\`" is enforced there: an unsafe name (\`.\`, \`..\`, empty after \`safeSessionName\`) throws \`INVALID_ARGS\`.

Why not (only) at \`scopeRequestSession\`: \`createRequestExecutionScope\` calls \`resolveSessionDir\` first thing after resolving the effective name — before any artifact path is used — so the store-level check *is* the admission-time rejection, and the request finalizes as an \`INVALID_ARGS\` response like the existing tenant-id rejection does. Every other caller (\`ensureSessionDir\`, app-log/event-log/record paths, inventory, doctor, teardown) receives an already admitted name, so nothing else changes. Session state is in-memory only, so no on-disk record can carry an unsafe name across the upgrade.

Checked for existing behavior depending on relative names: none — no code, test, or docs use \`--session .\` / \`..\`; \`defaultTracePath\`, the \`.ad\` writer, metro-session hints and claim/recovery all build *file names* under a fixed dir (a \`..-<ts>.trace.log\` file is harmless).

## Overlap with #1814

\`isSafeSessionSegment\` is the same predicate #1814 adds for its request-diagnostics route (which deliberately left \`resolveSessionDir\` alone). Whichever lands second takes a one-hunk merge in \`session-paths.ts\`.

## Tests (proven red first)

- \`session-store.test.ts\`: \`resolveSessionDir\` rejects \`.\`, \`..\`, \`''\` with \`INVALID_ARGS\`; a normal name stays under \`sessionsDir\`. Pre-fix: "Missing expected exception".
- \`request-execution-scope.test.ts\`: a request with \`session: '..'\` is rejected and nothing lands in the state dir. Pre-fix: the scope resolved with \`runnerLogPath: <stateDir>/runner.log\`.

\`pnpm check:affected --run\` green (336 files / 2834 tests, format, layering, fallow, build).

### DIFF
--- a/src/daemon/__tests__/request-execution-scope.test.ts
+++ b/src/daemon/__tests__/request-execution-scope.test.ts
@@ -91,6 +91,29 @@ test('createRequestExecutionScope resolves session-scoped request and runner log
   expect(scope.runnerLogPath).toMatch(/cwd_[a-f0-9]{16}_default\/runner\.log$/);
 });
 
+test('a relative session name is rejected before any session artifact path is written', async () => {
+  const sessionStore = makeSessionStore('agent-device-request-scope-');
+  const stateDir = sessionStore.resolveDaemonStateDir();
+  const requestId = 'relative-session-1';
+
+  await withDiagnosticsScope({ command: 'snapshot', requestId, logPath: LOG_PATH }, async () => {
+    await expect(
+      createRequestExecutionScope({
+        req: makeRequest({ session: '..', meta: { requestId } }),
+        sessionStore,
+        leaseRegistry: new LeaseRegistry(),
+      }),
+    ).rejects.toMatchObject({
+      code: 'INVALID_ARGS',
+      message: expect.stringMatching(/session name/i),
+    });
+    flushDiagnosticsToSessionFile({ force: true });
+  });
+
+  // Nothing landed in the state dir itself (where `sessions/..` resolves to).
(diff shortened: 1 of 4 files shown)


thymikee:
Clean review at `bb5ff957`: unsafe empty, `.` and `..` session segments are rejected at the owning SessionStore directory boundary before request or runner artifacts can form. The production-scope and path-containment regressions are meaningful, and no device evidence is needed for this filesystem-validation change. All completed checks pass; Coverage and two Smoke lanes remain pending. Code review is ready-for-human.

PR #1816, opened 2026-08-18
fix(ios-runner): reject CGRect.infinite in navigation frame guards

### LINKED ISSUE #1812: ios runner: navigation fallback helpers accept CGRect.infinite (tap point ≈ −9e307)
Two navigation helpers in the iOS runner accept `CGRect.infinite` as a valid frame. One of them turns it into a tap point at roughly `-9e307`; the other classifies it as sitting in the navigation header band.

Found by the **first execution of the dark XCTest set** (#1781 A7, PR #1789). Both tests live outside `ios.yml`'s hand-written `-only-testing:` list, so nothing had ever run them.

## Failures

Reproduced on Xcode 26.2 / iOS 26.2 Simulator (iPhone 16), full suite, 154 executed / 152 passed / 2 failed:

```
testTopLeadingNavigationFallbackPointRejectsInvalidFrame()
  XCTAssertNil failed: "(-8.988465674311579e+307, -8.988465674311579e+307)"

testTopNavigationControlFrameAcceptsOnlyHeaderBand()
  XCTAssertFalse failed
```

Both in `apple/runner/AgentDeviceRunner/AgentDeviceRunnerUITests/RunnerTests+Navigation.swift` (tests at lines 185 and 205).

## Root cause: the guards validate the size, never the origin

`CGRect.infinite` is `{ {-CGFLOAT_MAX/2, -CGFLOAT_MAX/2}, {CGFLOAT_MAX, CGFLOAT_MAX} }`. Its width and height are `CGFLOAT_MAX` — **finite and positive** — so a size-only validity check waves it straight through. The absurd part of the rect is its *origin*, which neither helper looks at.

`topLeadingNavigationFallbackPoint` (line 102):

```swift
guard frame.width.isFinite, frame.height.isFinite, frame.width > 0, frame.height > 0 else {
  return nil
}
let xOffset = min(max(frame.width * 0.08, 28), 44)
return CGPoint(x: frame.minX + xOffset, y: frame.minY + yOffset)
```

With `.infinite`: the guard passes, `xOffset` clamps to 44, and `frame.minX + 44` is `-8.988465674311579e+307` — the 44 is entirely lost to floating-point precision. That is the exact value in the assertion output.

`isTopNavigationControlFrame` (line 84) has the same shape of guard, then:

```swift
let maxY = window.minY + min(max(window.height * 0.22, 96), 180)   // = 180 for a 430x932 window
return candidate.midY >= window.minY && candidate.midY <= maxY
```

`CGRect.infinite.midY` is `-CGFLOAT_MAX/2 + CGFLOAT_MAX/2` = **exactly 0.0**, which lands inside `[0, 180]` → returns `true`.

The `.zero` half of the first test passes, because `width > 0` does reject it. `.infinite` is the only invalid rect that slips through.

## Impact

`tapTopLeadingNavigationFallback` feeds `onScreenWindowFrame(app:)` straight into `topLeadingNavigationFallbackPoint`. If that frame ever comes back as `.infinite` — an unresolved or not-yet-laid-out window — the runner synthesizes a tap at approximately `(-9e307, -9e307)` instead of declining to act. `isTopNavigationControlFrame` fails in the same direction: an unresolved element frame is treated as a top-navigation control.

Both are silent wrong-decision paths, not crashes.

## Suggested fix

There is already a correct precedent in the same target — `TapPointPolicy.isAllowed` in `apple/runner/AgentDeviceRunner/AgentDeviceRunnerUITests/RunnerTapPointPolicy.swift`:

```swift
if windowFrame.isNull || windowFrame.isEmpty || windowFrame.isInfinite {
```

`isInfinite` is the API that detects exactly this rect. The navigation helpers should apply the same three-way check rather than a fourth hand-rolled spelling of "is this rect usable".

Worth extracting one shared validity predicate so there is a single definition, then checking whether any sibling geometry helper in the runner has the same size-only gap. Fix the helpers, not the tests — the assertions describe the intended contract, and the test names (`RejectsInvalidFrame`, `AcceptsOnlyHeaderBand`) say so.

## Verifying

Needs a full-suite run; the PR lane's `-only-testing:` list does not include either test.

```sh
AGENT_DEVICE_XCUITEST_INCLUDE_UNIT_TESTS=1 \
  AGENT_DEVICE_IOS_RUNNER_DERIVED_PATH=<scratch> \
  AGENT_DEVICE_XCUITEST_DESTINATION='generic/platform=iOS Simulator' \
  pnpm gate swift-runner-ios

xcodebuild test-without-building \
  -xctestrun <scratch>/Build/Products/*.xctestrun \
  -destination "platform=iOS Simulator,id=<udid>" \
  -skip-testing:AgentDeviceRunnerUITests/RunnerTests/testCommand \
  -resultBundlePath <scratch>/RunnerTests.xcresult
```

`-skip-testing:` on `testCommand` is required — it is the runner's server entry point and blocks for 24 hours.

Once #1789 lands, the XCTest Nightly lane runs this set every night and will report both failures until they are fixed.

thymikee:
## Summary

Fixes #1812.

`topLeadingNavigationFallbackPoint` and `isTopNavigationControlFrame` (`apple/runner/AgentDeviceRunner/AgentDeviceRunnerUITests/RunnerTests+Navigation.swift`) validated only `width`/`height`, never the origin. `CGRect.infinite`'s origin (`-CGFLOAT_MAX/2` ≈ `-9e307`) is finite, so the existing `isFinite`/`> 0` guard let it through:

- `topLeadingNavigationFallbackPoint` synthesized a tap point at roughly `(-9e307, -9e307)` instead of declining.
- `isTopNavigationControlFrame` classified an unresolved frame as sitting inside the nav header band (`.infinite.midY == 0.0`, which falls inside `[0, 180]`).

There was already a correct precedent for detecting this exact rect in the same target — `TapPointPolicy.isAllowed` checks `windowFrame.isInfinite` (`RunnerTapPointPolicy.swift`). This PR extracts a shared `isUsableNavigationFrame(_:)` helper that adds the same `isInfinite` check on top of the existing finite/positive checks, and uses it in both navigation helpers, so there's a single definition instead of a size-only guard duplicated twice.

- [x] Fixed the helpers, not the tests — the two failing assertions describe the intended contract.

## Test plan

- [x] Built `AgentDeviceRunnerUITests` locally with `AGENT_DEVICE_XCUITEST_INCLUDE_UNIT_TESTS=1` (unit tests compiled in).
- [x] Ran the two previously-failing tests plus the rest of the navigation suite and the `TapPointPolicy` golden-parity test on iOS 26.2 / iPhone 17 Pro Simulator — all 7 pass:
  - `testTopLeadingNavigationFallbackPointRejectsInvalidFrame` ✅ (was failing)
  - `testTopNavigationControlFrameAcceptsOnlyHeaderBand` ✅ (was failing)
  - `testTopLeadingNavigationFallbackPointTargetsHeaderControlBand` ✅ (regression check — non-infinite frame still yields the same fallback point)
  - `testNavigationBackControlRankPrefersBackThenCloseThenCancel` ✅
  - `testNavigationBackPredicateUsesTheSharedKeywordTable` ✅
  - `testNavigationFallbackRequiresObservedVisualChange` ✅
  - `testTapPointPolicyMatchesGoldenParityTable` ✅ (unaffected)

### DIFF
--- a/apple/runner/AgentDeviceRunner/AgentDeviceRunnerUITests/RunnerTests+Navigation.swift
+++ b/apple/runner/AgentDeviceRunner/AgentDeviceRunnerUITests/RunnerTests+Navigation.swift
@@ -81,17 +81,16 @@ extension RunnerTests {
     return navigationBackKeywords.firstIndex { text.contains($0) }
   }
 
+  // isFinite/>0 alone don't reject CGRect.infinite — its origin (~-9e307) is finite.
+  static func isUsableNavigationFrame(_ frame: CGRect) -> Bool {
+    guard frame.width.isFinite, frame.height.isFinite, frame.width > 0, frame.height > 0 else {
+      return false
+    }
+    return !frame.isInfinite
+  }
+
   static func isTopNavigationControlFrame(_ candidate: CGRect, in window: CGRect) -> Bool {
-    guard
-      candidate.width.isFinite,
-      candidate.height.isFinite,
-      window.width.isFinite,
-      window.height.isFinite,
-      candidate.width > 0,
-      candidate.height > 0,
-      window.width > 0,
-      window.height > 0
-    else {
+    guard isUsableNavigationFrame(candidate), isUsableNavigationFrame(window) else {
       return false
     }
     // Accept the compact navigation/search header band without matching deep content controls.
@@ -100,7 +99,7 @@ extension RunnerTests {
   }
 
   static func topLeadingNavigationFallbackPoint(in frame: CGRect) -> CGPoint? {
-    guard frame.width.isFinite, frame.height.isFinite, frame.width > 0, frame.height > 0 else {
+    guard isUsableNavigationFrame(frame) else {
       return nil
     }
     // Aim at the standard leading navigation slot, bounded for compact and tablet widths.


thymikee:
Clean review at `c6fe861b`: `isUsableNavigationFrame` closes #1812 at the shared navigation guard used by both header classification and fallback-point synthesis, rejecting `CGRect.infinite` while preserving valid frames. The two regressions failed in the first full XCTest Nightly run and the focused seven-test iOS simulator validation is relevant. No architecture issue found; exact-head CI restarted after a comment-only amendment and remains pending. Code review is ready-for-human.

PR #1817, opened 2026-08-18
fix: declare frameworkTier on the hover descriptor

thymikee:
## Summary

main's Coverage job is red at 142d15633: #1786 added the public `hover` command without a `frameworkTier`, and #1804 (merged after) added the parity rule that every public descriptor declares one. Neither PR's own CI saw the combination. Every open PR's Coverage lane now fails on `parity.test.ts > frameworkTier is declared iff a command is public`.

`hover` is web-only pointer state, not part of the curated perceive/act loop, so it lands in `'extended'` (the parity test's pinned `'core'` set is unchanged).

## Validation

Red on main (revert of this one-liner): `AssertionError: hover declares frameworkTier iff it is a public command — 1 failed | 14 passed`. Green with it: parity + ai-sdk suites pass; `pnpm check:affected --run` green (426 files / 3641 tests). One file touched.

### DIFF
--- a/src/core/command-descriptor/registry.ts
+++ b/src/core/command-descriptor/registry.ts
@@ -1058,6 +1058,7 @@ export const RAW_COMMAND_DESCRIPTORS = [
     ...(ownerFilesEnabled ? { ownerFiles: ['src/commands/interaction/index.ts'] as const } : {}),
     targetIdentityVerification: 'pre-dispatch',
     catalog: { group: 'public' },
+    frameworkTier: 'extended',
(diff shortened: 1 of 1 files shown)


thymikee:
Clean review at 5fe3492a: `hover` is public and correctly declares `frameworkTier: 'extended'` at the descriptor source of truth, restoring the parity invariant without changing routing or runtime behavior. The revert-red proof is meaningful, exact-head CI and all platform smokes pass, and device evidence does not apply to this metadata-only repair. Ready-for-human.

Issue #1824, opened 2026-08-18
Coverage job: "Worker exited unexpectedly" kills one fork in the last ~40s of the run (4 of 17 recent CI failures)


thymikee:
A fork in the CI **Coverage** job dies in the last ~40s of the run, taking one test file's results with it. 4 of the 17 CI failures in the last 100 runs carry this signature (24%).

## Signature

```
Error: [vitest-pool]: Worker forks emitted error.
  ❯ ChildProcess.emitUnexpectedExit …/vitest/dist/chunks/cli-api…:3013:22
Caused by: Error: Worker exited unexpectedly
```

Runs: `32126085337`, `32047734673`, `32044136488`, `32041468051` (2026-08-17/18).

## Facts

- **All 4 are the Coverage job** (`pnpm test:coverage:ci`, V8 instrumentation) — never the plain unit lane. Run durations 286-296s.
- **Exactly one test file's results go missing each time**: `938 passed (940)` with 1 failed = 939/940; `940 passed (941)` = 940/941. In `32044136488` the only other recorded blocker was `unexplained failure — the run exited nonzero with no failed test recorded`, i.e. the worker exit *was* the whole failure.
- **No test emitted the error.** All `subprocess-stub` files reported ✓ before it: in `32044136488` the last stub file ✓'d at 16:07:14 and the unhandled error landed at 16:07:51 — a 37s silent gap at the very end of a 287s run.
- The `subprocess-stub` project always **completed successfully** first; it is not leaking a child (`harness.test.ts` uses `execFileSync`, which cannot outlive the call; `corpus-replay` bounds every case at 5s).

## Hypothesis: OOM-killed fork in the coverage job's memory tail

Coverage-job-only; no error emitted by any test; a whole file's results vanish (SIGKILL signature, not a throw); the death lands in the final 30-40s when `unit-core`'s long-pole files and the accumulated V8 coverage maps are at peak RSS. Not proven — the fork's exit code/signal is not readable from the log.

## Cheap confirmation

One Coverage run with `--max-old-space-size` on the forks' `execArgv` (`poolOptions.forks.execArgv`), or `NODE_OPTIONS=--trace-gc`, and read the exit code/signal.

## Why this needs fixing at the source

[...]

Issue #1823, opened 2026-08-18
subprocess-stub vitest project: members and kill criterion (#1781 A4)


thymikee:
The `subprocess-stub` Vitest project (`vitest.config.ts`) runs its members serialized
(`fileParallelism: false`, `maxWorkers: 1`) so only one file at a time spawns real stub binaries.
This issue is the live tracker for its membership and for deleting it.

## Members (3)

| file | why it spawns for real | measured |
| --- | --- | --- |
| `src/__tests__/client-metro.test.ts` | stubs npx plus the package managers and spawns a real Metro dev server per case; each case waits a real readiness poll | 4.7s CI / 5.7s local, 20 tests, 567-922ms per case |
| `scripts/fuzz/harness.test.ts` | the SUT *is* the subprocess watchdog: `execFileSync(process.execPath, …)` per case, one target hangs on purpose (#1414) | 13.2-14.0s CI / 16.8s local, 10 tests; worst case 4.4s |
| `scripts/fuzz/corpus-replay.test.ts` | replays the corpus through the same worker watchdog, waiting its per-case budget (`CASE_TIMEOUT_MS = 5_000`) | 2.2-2.4s CI / 1.1s local, 9 tests |

Three former members were dropped in the same PR: `src/platforms/android/__tests__/app-deployment.test.ts` (31ms in CI), `src/__tests__/platform-runtime-runtime-hints.test.ts` (212ms), `src/platforms/apple/core/__tests__/apps.test.ts` (277ms, already `vi.mock`s the exec seam). At those durations there is nothing to serialize.

## Deletion test (kill criterion)

If the three members run **un-serialized in the default pool for 20 consecutive CI runs with no timeout-shaped failure**, the project has no member that needs it — delete the project and fold the three files back into `unit-core`.

To run the experiment: drop `SUBPROCESS_STUB_TESTS` from `unit-core`'s `exclude` and delete the `subprocess-stub` project block; the files then run in the default forks pool. Revert if a timeout-shaped failure appears.

## Lane accounting

- **Catches:** CPU contention starving real stub-binary spawns past an internal budget, for the files that genuinely spawn.
- **Evidence:** the config comment cites the pre-split `android/index.test.ts` aggregation. `process.env` isolation — the other stated motive — is already delivered by `pool: forks` + `isolate: true` on every project, so serialization is only about CPU contention.
[...]

Issue #1430, opened 2026-07-27
obs: scheduled-lane health, freshness telemetry, and standard artifact envelope


thymikee:
Part of #1412 (Track E — gate integrity). From external review: nightly/weekly lanes (replays, perf, conformance differential, and the upcoming fuzz/mutation/torture/model-bench lanes) can fail or stop running for weeks while PR CI stays green. The observatory must watch the watchers.

## Deliverables
- Standard artifact envelope for every scheduled lane: `schemaVersion`, commit SHA, tool/config versions (or content hashes), seed where applicable, duration, result. Existing nightlies adopt it as they are touched; new lanes (#1414, #1415, #1416, #1427) MUST ship with it.
- A scheduled health job: reads last-run/last-success for every `schedule:`-triggered workflow via the GitHub API (the workflow list is derived from .github/workflows/, not hand-maintained) and posts/pings a tracking issue when any lane misses or fails two consecutive cadences.
- Retry telemetry feeding the repo-health snapshot (#1423) and delta comment (#1424): retry count per run and retry-eligible-file count (#1419) as ratcheted metrics — the flake list shrinking is the goal, and growth must be visible.
- Timeout classification from structured Vitest reporter metadata, never error-text sniffing (the classifier lessons: categories from recorded fields, not parsed names).
- Branch-protection required-contexts audit lives here if it needs elevated API scope (see the gate-manifest sibling issue).

## Acceptance
- Simulate a lane going dark (disable a test schedule twice): the health job opens/pings the alert issue.
- Every scheduled artifact produced after this lands carries the envelope; #1423's snapshot exposes lane freshness fields.

## Consolidation required (PR-wave review, 2026-08-01)

Three PARALLEL implementations of this issue's health watcher shipped as riders in the first worker wave — #1438 (`scripts/scheduled-lane/`, 904 LOC), #1439 (`scripts/scheduled-lane-health/`, 1,064 LOC), #1441 (`scripts/lane-health/`, 472 LOC) — two of them creating the SAME workflow filename (`.github/workflows/scheduled-lane-health.yml`) and two granting `issues: write` without dedicated review. All three PRs have been asked to strip their watcher.

The dedicated PR for this issue must therefore START by comparing the three implementations and consolidating one (steal the best pieces; the others die). Additional requirements from the review:
[...]

Issue #1419, opened 2026-07-27
ci: single-retry policy for enumerated contention-flaky files (timeouts only)


thymikee:
Part of #1412 (Track B).

## Context
The documented contention-flake pattern (process-spawn tests timing out under ~7x file parallelism, passing in isolation) red-herrings every contributor whose first full run goes red. The policy lives in docs/agents/testing.md prose; CI should encode it.

## Deliverables
- An ENUMERATED list of retry-eligible files (start from `SUBPROCESS_STUB_TESTS` in vitest.config.ts plus the repeat offenders: request-router-open, runner-client, runner-xctestrun, the bench subprocess test) — a shared constant, not a glob.
- CI reruns ONLY those files, once, when their failure is timeout-shaped; assertion failures never retry and fail immediately (the "timeouts only" rule).
- The job summary reports every retried file so retries stay visible, and the list has a ratchet: adding a file requires a comment naming why it spawns/waits.

## Acceptance
- An injected assertion failure in a listed file still fails the job on first run.
- An injected timeout in a listed file passes on retry with a visible summary line.
- docs/agents/testing.md updated to point at the constant instead of describing the pattern only in prose.

## Review amendments (2026-07-27)

- Every retry-list entry must follow the owned-waiver convention (ADR 0011 style): a trackingIssue URL plus an expiry/review date, not just a prose comment. An expired entry fails the gate until renewed or removed. Retry counts feed the scheduled-lane health telemetry (#1430).

PR #1827, opened 2026-08-18
chore(test): remove the contention retry (0 fires since 07-29) and shrink the subprocess-stub project to its 3 real spawners (#1781 A4)

thymikee:
#1781 item A4. Two changes, one theme: keep the part of the contention story that still pays, delete the part that never fired.

## (a) The contention single-retry policy is removed

It has fired **0 times** since it landed on 2026-07-29 (#1448):

- **0 of 234** sampled Coverage-job lane envelopes (2026-08-11 → 2026-08-18, pulled through the Artifacts API) had `retryCount > 0`. 20 of those runs failed; none was retried.
- **0 of 17** recent failed CI runs were retried: 5 refused with *"a failure landed outside the enumerated retry list"*, 4 with *"the run failed for a reason a rerun cannot re-check (unhandled error)"*, the rest n/a.
- All three trackers its 10 entries pointed at (#1098, #1414, #1419) are **closed/completed**, so no entry had a live owner. None had expired (`REVIEW_BY = 2026-10-31`), so the expiry gate would not have surfaced this for another 74 days.

What it cost: **~1,454 LOC deleted** — `scripts/lib/contention-retry{,-args,-blockers,-lane,-reporter,-run}.ts` (747), `contention-retry-policy.test.ts` (531), `runner-timeout-meta.ts` + `scripts/vitest-runner-timeout-setup.ts` + `run-blocker-bus.ts` (71), `test/contention-retry-fixtures/*` (75), config (~30) — plus a per-run secret marker threaded through a setup file on every Vitest project, and a standing obligation for every future gate reporter to call `recordRunBlocker`.

Also removed: the `check:contention-retry` package script and its place in `check:unit`, the `contention-retry` gate (catalog + CI step), the `contention-retry-envelope` artifact upload, and the `vitest-runner-timeout-setup.ts` setup-file wiring. `test:coverage:ci` is a plain `vitest run --coverage` again — the exact argv the wrapper passed through for a coverage run (reporters come from `vitest.config.ts`, which no lane overrides).

**Kept, with a note:** `scripts/lib/lane-envelope.ts` (#1430) is **not** retry-specific — the mutation lane (`scripts/mutation/run.ts`), the fuzz lane (`scripts/fuzz/envelope.ts`) and the concurrency-torture lane all build their envelopes from it. Only the retry's consumer (`contention-retry-lane.ts`) and the retry's artifact-upload step are gone. `run-blocker-bus.ts` **is** deleted: its only consumer was the retry's failure sink, and its only publisher (the slow-test ratchet) already fails the run by setting `process.exitCode = 1`.

One follow-on: `OPAQUE_RUNNERS` loses its `test:coverage:ci` entry, because a bare `vitest run --coverage` is something the gate-manifest loader reads directly. That makes "a Vitest project no registered check runs" *unrepresentable* rather than merely detected, so `scripts/gate/audit-coverage.test.ts` now constructs that state by project-scoping the coverage script in its model mutation.

## (b) The `subprocess-stub` project stays, shrunk to the 3 files that really spawn

The list had 10 entries; 6 were serialized. Measured in CI (run 32127451031) and locally:

| kept | why | CI |
| --- | --- | --- |
| `src/__tests__/client-metro.test.ts` | real Metro-stub server + child spawn per case | 4.7s (567-922ms/case) |
| `scripts/fuzz/harness.test.ts` | the SUT *is* the subprocess watchdog; one target hangs on purpose | 13.2-14.0s |
| `scripts/fuzz/corpus-replay.test.ts` | same watchdog, per-case budget | 2.2-2.4s |

| dropped | CI | why |
| --- | --- | --- |
| `src/platforms/android/__tests__/app-deployment.test.ts` | **31ms** | shell stubs in a tmpdir; nothing to serialize |
| `src/__tests__/platform-runtime-runtime-hints.test.ts` | **212ms** | env-driven stub exit codes, no sleeps |
| `src/platforms/apple/core/__tests__/apps.test.ts` | **277ms** | already `vi.mock`s the exec seam |

All three now run in `unit-core` and pass there (verified locally: 3 files / 72 tests, 7.8s, slowest case 1.57s — inside the 2.5s slow-test budget). The 4 retry-only entries (`request-router-open.test.ts`, `runner-client.test.ts`, `runner-xctestrun.test.ts` — whose stated reason was false: it `vi.mock`s `node:child_process` and nothing spawns — and `help-conformance-bench.test.ts`) simply cease to be listed anywhere. The optional in-process conversions for them are deliberately **not** done here.

`SUBPROCESS_STUB_TESTS` is now a plain 3-entry array in `vitest.config.ts` with one line per entry saying why it spawns; `fileParallelism: false, isolate: true, maxWorkers: 1` are unchanged.

## Lane accounting

**`subprocess-stub` project — retained**

- **Catches:** CPU contention starving real stub-binary spawns past an internal budget.
- **Evidence:** pre-split `android/index.test.ts` aggregation per the config comment; the 3 members measured above genuinely spawn per case. (`process.env` isolation, the other stated motive, already comes from `pool: forks` + `isolate: true`.)
- **Cost:** ~15 config lines; ~0 CI wall clock (concurrent project, `unit-core` is the 296s critical path); ~25-37s serialized chain locally.
- **Kill-criterion:** #1823 — if the three run un-serialized in the default pool for 20 consecutive CI runs with no timeout-shaped failure, delete the project.

**Contention retry — removed.** Catches: runner-proven timeouts in enumerated files only / Evidence: 0 fires in ~3 weeks, refused every observed failure class / Cost: ~1.45k LOC + a per-run secret marker + a bus every future gate reporter must call / **Killed.**

## Issues filed

- #1823 — `subprocess-stub` vitest project: members and kill criterion (the live tracker the config comment points at; all three old trackers were closed).
- #1824 — Coverage job: "Worker exited unexpectedly" kills one fork in the last ~40s of the run. 4 of the 17 recent CI failures (24%), Coverage job only, exactly one file's results vanish each time, all stub files ✓ before it. The retry refused this class by design and is now gone, so it has to be fixed at the source.

## Test plan

- `pnpm test:unit` — 891 files / 6,847 tests pass, including the three files moved back into `unit-core`.
- `pnpm test:coverage:ci` — run locally; writes `coverage/lcov.info`.
- `pnpm typecheck`, `pnpm lint`, `pnpm format:check` — green.
- `pnpm check:gate-manifest` (46 checks across 33 lanes) and `pnpm check:gate-manifest:test` (37 tests) — green.
- `pnpm check:affected:test` (36 tests), `pnpm check:layering` (183 tests + guard), `pnpm check:fallow` — green.
- `pnpm check:production-exports` reports the same 17 pre-existing exports as clean `origin/main` (verified by stashing).


thymikee (commented):
Adversarial pass over #1827 (checked out `origin/chore/1781-a4-contention-retry` @ 57b93d62f in a throwaway worktree, `pnpm install`, then ran the actual gates rather than trusting the PR-body claims).

## Verified against the diff (no leftover references found)

Grepped the whole tree for `contention-retry`, `contentionRetry`, `CONTENTION_RETRY`, `runner-timeout`/`RUNNER_TIMEOUT`/`runnerTimedOut`/`takeRunnerTimeoutToken`, `run-blocker-bus`/`recordRunBlocker`/`drainRunBlockers`, `serializedStub`, `timeout-provenance`, and `#1419`/`#1430` across `.ts`/`.md`/`.yml`/`.json`. Only hits:
- `docs/agents/testing.md:506` — the intended "Removed 2026-08-18 (#1781 A4)" historical note.
- `src/platforms/apple/core/__tests__/screenshot.test.ts` — a `mkdtempForTest('agent-device-ios-runner-timeout-')` fixture-name string, unrelated (predates this PR, not touched by it). False positive.
- `#1430` hits are all the *kept* `lane-envelope.ts` consumers (mutation/fuzz/concurrency-torture), consistent with the PR's claim that lane-envelope itself is untouched.

No dangling doc anchors, no orphaned imports, nothing in `.github/`, `scripts/gate/declarations.ts`, `scripts/check-affected/{checks,model}.ts`, `.fallowrc.json`, `package.json` scripts, or either vitest config still points at a deleted symbol/file.

## Ran the gates myself (all green, matching the PR-body numbers exactly)

- `pnpm check:gate-manifest` → `46 checks wired across 33 lanes` — matches the body verbatim.
- `pnpm check:gate-manifest:test` → 37/37 pass — matches.
- `pnpm check:affected:test` → 36/36 pass — matches.
- `pnpm check:fallow` → `No issues in 25 changed files`.
- `node --test scripts/gate/audit-coverage.test.ts` → 6/6 pass, including the new `projectScoped` case that makes "an unrun Vitest project" representable now that `OPAQUE_RUNNERS` is empty (`scripts/gate/declarations.ts:6`, consumed by `scripts/gate/model.ts:199`).
- `vitest run --project unit-core src/platforms/android/__tests__/app-deployment.test.ts src/__tests__/platform-runtime-runtime-hints.test.ts src/platforms/apple/core/__tests__/apps.test.ts` → 3 files / 72 tests pass in `unit-core`. Confirms item (3): `vitest.config.ts`'s `unit-core.exclude: [...SUBPROCESS_STUB_TESTS]` is derived dynamically from the same shrunk 3-entry array the `subprocess-stub` project's `include` uses, so the 3 dropped files fall straight into `unit-core` via its existing `src/**/*.test.ts` glob rather than being silently unlisted anywhere.
- `oxlint` on every touched non-deleted file: clean.

## `test:coverage:ci` argv/reporter equivalence (item 2) — confirmed correct

Old `contention-retry-run.ts`'s `firstRunArgs` was called with `modes.projects` derived from an unset `--project` CLI flag (package.json never passed one to the wrapper), so `projectArgs([]) = []`, and the first run was literally `vitest exec vitest run --coverage` with reporters resolved from `reporters(env)` where the only difference from bare config was appending `contentionRetryReporter()` (now correctly deleted along with the mechanism it fed). New `"test:coverage:ci": "vitest run --coverage"` is byte-for-byte the same argv the old wrapper produced for its first run, and `vitest.config.ts`'s top-level `reporters: ['default', slowTestGateReporter()]` is unchanged — so the claim "reporters come from vitest.config.ts, which no lane overrides" holds.

Downstream: "Enforce changed-line coverage gate" in `ci.yml` still runs unconditionally after `unit-ci` and still reuses the lcov the Coverage step wrote — `coverage.reporter` in `vitest.config.ts` (untouched by this diff) still includes `'lcov'`. The deleted "Upload contention-retry envelope" artifact step had zero other consumers in the repo (grepped), consistent with the mechanism it fed being gone.

## Issues #1823 / #1824

Both open and match the PR body's claims precisely — #1823 lists the same 3 kept members + measured durations + the same 20-consecutive-CI-runs kill criterion; #1824 states the same "Worker exited unexpectedly" signature, run IDs, and "the retry refused this class by design and is now gone" framing. No discrepancy.

## Minor / non-blocking

- `package.json`: `test:coverage` (`"vitest run --coverage"`) and `test:coverage:ci` (also now `"vitest run --coverage"`) are byte-identical scripts. Harmless — one is the CI gate entrypoint (`scripts/gate/declarations.ts` no longer needs to special-case it) and one is the local ad-hoc command — but a future reader diffing the two will wonder why there are two names for one command. Not worth blocking on.
- I chased a hunch that shrinking `SUBPROCESS_STUB_TESTS` (which `vitest.mutation.config.ts:6,38` also consumes for its `exclude`) would let PATH-mutating files reach Stryker's thread-pool mutation runner unprotected. Empirically two of the three dropped files (`app-deployment.test.ts`, real `process.env.PATH =` mutation with try/finally restore) **are** already reachable via `vitest related` from the current `stryker.config.json` `mutate` list — so this isn't hypothetical, it's live today. But `scripts/mutation/test-scope.ts`'s own stated thread-pool exclusion criterion is only chdir/`worker_threads` usage, never generic env mutation, and the files restore `PATH` synchronously in `finally` with no `test.concurrent` in the file (global `maxConcurrency` doesn't even apply since there's no concurrent test), so there's no realistic race. Checked, not a bug — noting only so nobody re-treads this.

## Verdict

Everything the PR body claims reproduces exactly under a real re-run of the gates it cites, and an exhaustive grep across the whole tree turns up no leftover reference to any deleted symbol, file, or gate. The evidence for deleting the retry (0/234 fires, all three trackers closed) and for shrinking the stub list (measured per-file durations) is solid and independently checked here. This is about as clean a deletion PR as they come.

VERDICT: approve
One-line summary: exhaustive grep + re-running every gate the PR cites (`check:gate-manifest`, `check:gate-manifest:test`, `check:affected:test`, `check:fallow`, the 3 moved-back unit files, both linked issues) all confirm the PR body's claims with zero leftover references found.

thymikee:
Reviewed exact head 57b93d62f: clean. The contention retry is removed end-to-end without leaving stale workflow/gate/docs ownership; the bare coverage command preserves the wrapper’s first-run invocation. Subprocess-stub membership now has one source of truth and retains only the three genuine subprocess/watchdog tests, with gate and linked-issue evidence covering the decision. No device evidence is needed for this tooling-only change. Code review is ready-for-human; iOS Smoke is still in progress and the PR remains draft.

PR #1834, opened 2026-08-18
test: prune abandoned test-run tmp directories at run setup

thymikee:
## Summary

Retro follow-up (item 1 of the agreed list): a run killed before its teardown — an agent tool-timeout SIGKILL, OOM, a cancelled CI job — left `/tmp/agent-device-test-run-<pid>-*` behind, and `check:tmpdir-leaks` (which runs after `test:unit` inside `check:unit`) flagged every dead-pid directory it found. It could not distinguish *this run leaked* from *a run an hour ago was killed*, so one killed run poisoned every later, otherwise-green gate on the host until someone `rm -rf`'d by hand.

Now both TMPDIR-redirection entry points — the Vitest global setup and the `node --test` wrapper — prune dead-pid run directories **before** creating their own (`pruneAbandonedRunDirectories`), and print one `[tmpdir] pruned N …` line when they did. The post-run check keeps its exact semantics, but by construction it can only ever name the run that just finished. Directories owned by a live pid (a concurrent run in another worktree) are never touched.

Structural side of it: `TEST_RUN_TMP_ROOT`/`TEST_RUN_TMP_PREFIX` move from `vitest-tmpdir-global-setup.ts` into `check-tmpdir-leaks-model.ts`, next to the liveness classification, so setup and wrapper import the prune without a cycle. Callers re-pointed; no rename.

## Validation

- **Red proof:** with the two prune calls reverted (pre-fix setup + wrapper), the two new lifecycle tests fail: `the wrapper must prune the abandoned run directory` / `setup must prune the abandoned run directory`. With the fix, `pnpm check:tmpdir-leaks:test` is 21/21. The lifecycle tests plant a directory owned by `INT32_MAX` (outside every platform's pid range, so ESRCH by construction — no pid-reuse flake) plus one owned by the test process itself, run a real nested `vitest run` / wrapped `node --test`, and assert the dead one is gone and the live one survives.
- Full `pnpm check:unit`: 910 files / 6,935 tests pass, `check:tmpdir-leaks` clean, smoke 38/38 (after `pnpm build` — the fresh worktree had no `dist`). `check:fallow` clean; format/lint/typecheck pass.
- **Not this PR:** `check:layering` is red on `origin/main` itself at `9a0d6dead` (R9 46→47, R10 16→17; the main Layering Guard job is failing) — reproduced on a clean checkout with this branch stashed. Scripts-only change here, no `src/` touched.

Docs: one paragraph added to `docs/agents/testing.md` next to the existing tmpdir guidance. 8 files touched, all under `scripts/` + that doc.

### DIFF
--- a/docs/agents/testing.md
+++ b/docs/agents/testing.md
@@ -78,6 +78,14 @@ cleanup for a directory they created — that's the global teardown's job, and p
 already existed for other reasons should stay (it's the fallback global sweep that's new, not a
 replacement for tests being tidy).
 
+A run killed before its teardown (a tool timeout's SIGKILL, OOM, a cancelled job) leaves its
+`/tmp/agent-device-test-run-<pid>-*` directory behind. The next run on the host prunes every such
+directory whose owning pid is dead (`pruneAbandonedRunDirectories`, called by both the Vitest global
+setup and the `node --test` wrapper) and prints one `[tmpdir] pruned …` line, so `check:tmpdir-leaks`
+after `test:unit` can only ever name the run that just finished. Directories owned by a live pid —
+a concurrent run in another worktree — are never touched. If the check fails, the leak is this
+run's: a teardown that did not execute, not history.
+
 Keep tests behavioral. Do not assert shapes or cases TypeScript already proves.
 
 A test added as a regression pin must be shown to fail without the change it pins — vacuity is the
--- a/docs/agents/testing.md
+++ b/docs/agents/testing.md
@@ -80,11 +80,15 @@ replacement for tests being tidy).
 
 A run killed before its teardown (a tool timeout's SIGKILL, OOM, a cancelled job) leaves its
 `/tmp/agent-device-test-run-<pid>-*` directory behind. The next run on the host prunes every such
-directory whose owning pid is dead (`pruneAbandonedRunDirectories`, called by both the Vitest global
-setup and the `node --test` wrapper) and prints one `[tmpdir] pruned …` line, so `check:tmpdir-leaks`
-after `test:unit` can only ever name the run that just finished. Directories owned by a live pid —
-a concurrent run in another worktree — are never touched. If the check fails, the leak is this
-run's: a teardown that did not execute, not history.
+directory that is genuinely abandoned (`pruneAbandonedRunDirectories`, called by both the Vitest
+global setup and the `node --test` wrapper) and prints one `[tmpdir] pruned …` line, so
+`check:tmpdir-leaks` after `test:unit` can only ever name the run that just finished. Abandoned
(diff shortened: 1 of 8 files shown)


thymikee:
P1 safety blocker: pruning uses only the setup/wrapper PID as liveness proof, but SIGKILL/tool timeout can kill that owner while already-running test children or Vitest workers remain alive. The next run then recursively deletes a TMPDIR still in use. Existing tests plant a directory owned by the test process and miss this motivating case. Use an ownership model that accounts for live consumers (or conservatively declines deletion), and add a regression that starts a wrapped long-lived child, SIGKILLs only its owner, proves the next run preserves the directory, then proves pruning after every consumer exits. Exact-head iOS Smoke and Layering are also red, even if inherited.

thymikee:
Addressed in 89fc5cea9 (rebased onto #1838's layering fix as well).

**Ownership model:** a run directory is now live while its owner pid is alive **or any process's `TMPDIR` points into it** — read from `ps -E` on macOS and `/proc/<pid>/environ` on Linux (`liveRunDirectoryConsumers`). That is the signal every consumer actually carries: the wrapper/vitest main exports it and the `node --test` chain, forked workers, and any daemon a test spawned inherit it. Both the prune and the post-run `check:tmpdir-leaks` use the one model, so an orphaned-but-in-use directory is neither pruned nor reported until the last consumer exits. A process this user can't inspect contributes nothing and its directory is judged by owner pid alone (the previous behaviour). Cost: one scan per run start, ~0.45s here.

**Regression (`node-test-tmpdir.test.ts`):** a wrapped probe spawns a detached long-lived child inheriting TMPDIR and then waits; only the wrapper pid gets SIGKILL; `pruneAbandonedRunDirectories` on the next call must leave the directory (asserted, plus the child still alive); then every consumer — the detached child *and* the orphaned `node --test` chain, which holds the same TMPDIR — is killed, the scan is polled until it no longer lists the directory, and the next prune removes it. Planted red with owner-only liveness: `+ ['agent-device-test-run-73123-…'] - []` at "a directory with a live consumer is not pruned". Model unit tests cover the classification with injected consumers, and `runDirectoryNameOf` at any depth.

Exact-head status: Layering is green after the rebase (main was red until #1838); iOS Smoke was the `alert accept` main-thread timeout on the macOS runner — reran the failed job.

thymikee:
Re-reviewed exact 89fc5cea: clean and ready-for-human. The pruning/leak-check classifier now preserves a run directory whenever any live process still advertises that TMPDIR, even after the owner PID dies. The real wrapper regression SIGKILLs the owner while detached/orphaned consumers survive, proves the next prune preserves their directory, then proves pruning after all consumers exit; planted-red evidence covers the prior owner-only behavior. Completed gates are green; iOS/macOS Smoke and Swift compile remain pending before merge.

PR #1779, opened 2026-08-16
refactor: migrate snapshot to device runtime

thymikee:
## Summary

Migrate the public `snapshot` command from legacy capability dispatch to the ADR 0019 request-scoped device runtime.

- model the cutover as three exact-owner operations and four parametrized request plans, with one facts inspection and one runtime bind per handler
- enforce R32 through a deliberately bounded positive seam: the route selects one normalized `plan.use` and passes that exact use to the shared `inspectRequiredRuntimeUse`; local snapshot admission policy is rejected
- remove the custom identity-flow/alias tracker instead of extending it with another omission patch; the structural rule no longer depends on identifier spelling, aliases, destructuring, or object wrapping
- preserve Apple, Android, HarmonyOS, Linux, web, WebDriver, and Limrun behavior, including macOS surfaces, Linux shaping, cancellation, watchOS denial, provider fail-closed behavior, and public-only sparse screenshot fallback
- retire generic snapshot dispatch, static capability admission, descriptor dispatch projection, and two daemon-to-platform value edges
- keep shared diff/find/wait/interaction capture consumers on one narrow interactor seam until their own command units migrate; no unrelated command descriptor was implicitly migrated
- remove duplicate Linux/macOS desktop capture implementations and the thin host wrapper; production duplication remains flat at 5 groups / 207 lines
- make the iOS smoke scroll helper require actual visibility before skipping a scroll, with capture-stall retries that do not consume scroll attempts

The measured snapshot-unit budget approved on #1739 was +10,209 B JS raw, +3,058 B JS gzip, +2,590 B npm tarball, and +10,226 B unpacked against its authorized base: https://github.com/callstack/agent-device/issues/1739#issuecomment-5317036300. Returning to a non-positive delta requires separately migrating the remaining legacy capture consumers and is intentionally outside this command-atomic unit.

## Validation

- rebased onto `main` at `60f6356b0`
- exact head: `bea6ce36d`
- clean committed tree: `pnpm check:affected --run` passed all locally runnable selected gates before the chained force-with-lease push
- affected Vitest/coverage run: 603 files / 4,907 tests passed; changed-line coverage 206/217 (94.93%)
- R32 planted red includes the reported `{ device }` object-wrapper bypass; the positive route-to-owner-facts seam rejects the local admission reimplementation without reconstructing JavaScript data flow
- rebase conflict reconciliation retained main's caller-supplied replay-source model and the migrated narrow snapshot capture seam; focused replay/runtime tests passed 82/82
[...]

PR #1633, opened 2026-08-06
refactor(daemon): one interface for the deferred interaction outcome

thymikee:
Closes #1629.

## What

Deepens the daemon's deferred "did that interaction actually take effect?" machinery behind one interface, with no behavior change.

- **`src/daemon/deferred-interaction-outcome.ts`** — the one interface, and (post-review) the R7 owner of `postGestureStabilization`:
  - `markDeferredInteractionOutcome` — the single mutation-side marking sequence. The three previously independent marking sites (`finalizeTouchInteraction`, `dispatchGenericCommand`, and `session-open`'s freshness-only mark) all call it; the ordering (pending outcome retry → Android freshness → post-gesture stabilization) is stated once, and each marker keeps its own eligibility gate. Stabilization marking is module-private behind this entry — even its tests drive the public interface.
  - `resolveDeferredInteractionOutcome` — the capture-side composition (pending-outcome retry → stabilization → freshness recovery), moved out of `snapshot-capture.ts`'s private functions and parameterized over the capture primitive, so it is directly testable with a scripted capture and a plain session record.
- **`src/daemon/post-gesture-stability.ts` (leaf, outside the R9 cycle)** — the pure polling/verdict mechanics, parameterized by hooks (capture, surface reader, the three signature comparators). It imports no cycle owners and no `SessionState`; the owner wires the comparators from `interaction-outcome-policy` and clears the pending record when the loop returns.
- **`src/daemon/gesture-no-effect.ts` (leaf)** — the #1600/#1601 no-effect wording + annotation fold.
- **Freshness heuristics + recovery live beside their state machine** — `getAndroidFreshnessReason` (all three staleness shapes) and the retry loop moved into `android-snapshot-freshness.ts`, which gained its first direct test file.
- **`isPostGestureStabilizationPending`** — the owner's gate predicate replaces the two raw `session.postGestureStabilization` reads in `direct-ios-selector.ts` and `selector-capture-runtime.ts`.
- `snapshot-capture.ts` shrinks **617 → 359 lines**; the deferred-outcome owner is 374; the stability leaf 228. CONTEXT.md gains the **deferred interaction outcome** term.

## Scope guards honored

- The three session-consistency disciplines stay distinct: this module owns only the post-action markers — ADR 0014 ref-frame expiry and ADR 0012/0016 staged protocols never route through it, and nothing here is named "arm".
[...]

PR #1835, opened 2026-08-18
fix(layering): unblock main — raise the daemon-server R9 ceiling for #1779's snapshot interactor seam

thymikee:
**main is red on the Layering Guard job.** This unblocks it. Not a draft.

## Cause: two green PRs, one interaction

Neither PR is wrong; they were green against their own bases and never met until merge.

- **#1825** (`ef6ec2995`, "make R9 shrink mandatory") pinned `LARGEST_TYPE_CYCLE_ZONE_CEILINGS` to the then-measured values: `daemon-server: 16`, total `46`. The shrink is now mandatory, so the ceiling is an equality pin, not headroom.
- **#1779** (`d76e0f94e`, "migrate snapshot to device runtime") merged after it and added one new file that joins the largest type-level cycle.

`pnpm check:layering` on `9a0d6dead`:

```
[R9 type-cycle-size] scripts/layering/daemon-modularity.ts:1 —
  the largest type-level import cycle grew to 47 files (baseline 46).
[R10 daemon-modularity] src/daemon/daemon-command-registry.ts:1 —
  the largest type cycle now contains 17 daemon-server file(s) (baseline 16)
```

## The joining edge

The +1 node is **`src/daemon/handlers/snapshot-interactor-capture.ts`** (new in `d76e0f94e`), a choke point wedged between two files that were already cycle members:

```
src/daemon/handlers/snapshot-capture.ts        (member)
  -> src/daemon/handlers/snapshot-interactor-capture.ts   (new, line 22, value)
  -> src/core/interactors.ts  (getInteractor)  (member, line 7, value)
```

A new file that value-imports one SCC member and is value-imported by another always joins the SCC — +1 node, no new edges.

Attribution is a measured member-list diff of `largestTypeCycleMembers`, taken at `a853734f0` (#1779's parent, 46 members) vs `d76e0f94e` (47 members). The only differing line is `+ src/daemon/handlers/snapshot-interactor-capture.ts`.

**Note on the R10 message**: it names `src/daemon/daemon-command-registry.ts`, which is *not* the file that grew — that file was already a member at `a853734f0`. `checkTypeCycleBaseline` reports `members.find(m => targetDagZone(m) === zone)`, i.e. the alphabetically first member of the over-budget zone. Worth a follow-up; deliberately not touched here.

## The fix: explicit ceiling raise (option 3 of the brief)

`daemon-server: 16 -> 17` in `scripts/layering/daemon-modularity.ts`, with a one-line comment naming `d76e0f94e` and the joining edge. `TYPE_CYCLE_BASELINE` is derived from the zone ceilings, so the total follows to 47 automatically. `daemon-modularity.test.ts` pins both numbers as literals, so they move with it.

**I did not break the edge**, and the maintainer should decide whether that is acceptable. The zero-growth shape (per the #1633 precedent: host the seam in an existing cycle node) means deleting `snapshot-interactor-capture.ts` and moving the `getInteractor` call back into `snapshot-capture.ts` or up into `core/interactors.ts`. That module is the deliberate `vi.mock` seam for **43 test files** (`grep -rl snapshot-interactor-capture src` → 44 including itself); relocating it changes every one of those mock paths to a module whose other exports those tests still need unmocked. That is a real refactor, not a red-main fix. Happy to follow up with it as a separate PR if preferred — this PR only restores the gate to matching reality.

## check:layering after the fix

```
Layering guard: OK — 1312 source files satisfy R2-R3 and contain no value-import cycles ...
the largest type-level cycle is 47 files (R9); R10 pins R7 at 21 writer-owned fields /
25 owner claims, R9 at 47 files with zone ceilings, 2 external daemon/types.ts importers,
and zero forbidden logical-module imports beyond 0 recorded migration import(s) ...
```

## Test plan

- `pnpm check:layering` — green (47 total / 17 daemon-server), 178/178 layering unit tests pass
- `pnpm typecheck` — green
- `pnpm lint` — green
- `pnpm format:check` — green
- No production code touched; only `scripts/layering/daemon-modularity.ts` and its test.

### DIFF
--- a/scripts/layering/daemon-modularity.test.ts
+++ b/scripts/layering/daemon-modularity.test.ts
@@ -73,8 +73,8 @@ test('daemon modularity baseline records the measured R7 ownership pressure', ()
     Object.values(SESSION_STATE_FIELD_OWNERS).reduce((sum, owners) => sum + owners.length, 0),
     DAEMON_MODULARITY_BASELINE.sessionState.ownerFileClaims,
   );
-  assert.equal(TYPE_CYCLE_BASELINE, 46);
-  assert.equal(DAEMON_MODULARITY_BASELINE.largestTypeCycle.zoneMembers['daemon-server'], 16);
+  assert.equal(TYPE_CYCLE_BASELINE, 47);
+  assert.equal(DAEMON_MODULARITY_BASELINE.largestTypeCycle.zoneMembers['daemon-server'], 17);
   assert.equal('daemon' in DAEMON_MODULARITY_BASELINE.largestTypeCycle.zoneMembers, false);
 });
 
@@ -230,6 +230,6 @@ test('R9 rejects a baseline left above the measured cycle', () => {
 
   assert.equal(violations.length, 1);
   assert.match(violations[0]!.rule, /^R9 /);
-  assert.match(violations[0]!.message, /dropped to 45 files \(baseline 46\)/);
+  assert.match(violations[0]!.message, /dropped to 46 files \(baseline 47\)/);
   assert.match(violations[0]!.message, /Lower LARGEST_TYPE_CYCLE_ZONE_CEILINGS by the same 1/);
 });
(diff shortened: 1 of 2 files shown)


thymikee:
Superseded by maintainer fix.

PR #1159, opened 2026-07-08
feat: add Roku WebDriver provider

thymikee:
## Summary

Add a Roku WebDriver provider for LAN Roku devices.

- Adds `connect roku` profile support with `--roku-webdriver-url` and `--roku-device-ip`, plus `web-instance` lease backend plumbing.
- Implements Roku WebDriver session allocation, launch, source snapshots, ECP TV remote input, close/release cleanup, and conservative web-TV capability gating.
- Covers CLI/profile parsing, capability gating, and a provider-scenario daemon flow using a fake Roku WebDriver/ECP server.

## Validation

- `pnpm format`
- `pnpm exec vitest run src/cli/parser/__tests__ src/utils/__tests__/command-schema-guards.test.ts src/__tests__/cloud-connect-profile.test.ts src/__tests__/remote-connection.test.ts src/core/__tests__/capabilities.test.ts src/core/__tests__/capability-plugin-routing-parity.test.ts test/integration/provider-scenarios/roku-webdriver-runtime.test.ts`
- `pnpm typecheck`
- `pnpm build`
- `pnpm test:smoke`
- `pnpm check:unit` unit projects passed once; a rerun hit an unrelated Apple runner slow-test timeout under host load, and the timed-out tests passed in isolation with `pnpm exec vitest run src/platforms/apple/core/__tests__/runner-client.test.ts src/platforms/apple/core/__tests__/index.test.ts --testNamePattern "ensureXctestrun rebuilds after cached macOS runner repair failure|screenshotIos retries simulator capture timeouts and eventually succeeds"`

PR #1843, opened 2026-08-18
test: pin test files over 1,000 lines at their exact length so they can only shrink

thymikee:
## Summary

Retro follow-up, item 6. `AGENTS.md` already says past 1,000 lines is architecture debt and tests are not exempt; nothing enforced it, and `snapshot-handler.test.ts` (2,652 lines) gained 55 lines in #1779 under that rule. This is the slow-test ratchet's shape applied to the other resource a giant test file consumes — a reader's context:

- `src/__tests__/test-file-size-ratchet.test.ts` (unit-core) walks `src`, `packages`, `test`, `scripts` for `*.test.{ts,tsx,mjs}` and pins the **26** files over the 1,000-line tripwire at their exact length, R9-style (#1781 A6 equality pin): growth fails ("extract instead of adding"), shrink fails until the pin is lowered in the same PR (so the ratchet keeps the gain), a file that drops under the line leaves the list, a new file may not cross it, a stale pin is reported.
- Lane card (in the file header, per the #1781 rule): **Catches** a >1,000-line test file growing or a new one appearing · **Evidence** 26 offenders today, the top one grew last PR · **Cost** one walk + line count per unit run, ~250ms · **Kill criterion** empty pin list → delete the file.
- Docs: a bullet next to the slow-test ratchet in `docs/agents/testing.md`; the AGENTS.md tripwire line now points at the gate.

## Validation

- Green on the current tree (2 tests, 230ms).
- **Planted red on a real pinned file:** appending one line to `src/__tests__/cli-network.test.ts` → `grew to 1093 lines (pinned 1092): extract instead of adding…`; deleting one line → `shrank to 1091 lines (pinned 1092): lower its pin in this PR…`. Restored.
- The decision function's four outcomes (grow / shrink-under-tripwire / unpinned crossing / stale pin) are pinned in-file with their exact messages.
- Open PRs touching a pinned file: only #1159 (`remote-connection.test.ts`, 2026-07 draft-era branch); anything else in flight that adds to one of the 26 will hit the message and the doc it points to — that is the gate working, not collateral.
- lint/typecheck/fallow clean. Not this PR: `check:layering` R9/R10 red on `origin/main` itself.

3 files touched (1 test, 2 docs).

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -134,7 +134,9 @@ Module size is about agent context safety, and the unit is questions, not lines:
 one question so `rg` → read-whole-file stays one cheap bounded read.
 
 - tripwires: target ≤300 LOC per implementation file; past 500, extract before adding behavior; past
-  1,000 is architecture debt unless it is generated data or a fixture snapshot. Tests are not exempt.
+  1,000 is architecture debt unless it is generated data or a fixture snapshot. Tests are not exempt:
+  test files over 1,000 lines are pinned at their exact length and may only shrink
+  (`src/__tests__/test-file-size-ratchet.test.ts`).
 - name files by the domain concept they answer (`runner-cache.ts`, `interaction-touch-response.ts`),
   not by layer leftovers (`utils2.ts`, `common.ts` accretion).
 - colocate machine-readable claims with the code they describe — coverage manifests beside contract
--- a/docs/agents/testing.md
+++ b/docs/agents/testing.md
@@ -555,6 +555,13 @@ Measured on the full unit suite (340 files, 3,210 tests, 48s wall at ~7x paralle
   2.5s, integration 15s, failure at 2x budget (the band between reports without failing — host
   load legitimately stretches borderline tests, and a flaky gate trains people to ignore it).
   The pin list only shrinks, or grows in the same PR with a justification.
+- **The test-file size ratchet** (`src/__tests__/test-file-size-ratchet.test.ts`) is the same
+  shape for the other resource a giant test file consumes — a reader's context. Every test file
+  over the 1,000-line tripwire is pinned at its exact length, R9-style: growth fails ("split it
+  along the source module it mirrors"), shrinking fails until the pin is lowered in the same PR,
+  a file that drops under the line leaves the list, and a new file may not cross it. Adding a
+  test to a pinned file means moving that family out first — the failure names the file and the
+  fix; never raise a pin.
 - **Isolation stays ON; pool stays forks — both measured.** `--no-isolate`: 205s wall vs 48s
   (module state — timers, memos, singletons — thrashes across files sharing a worker).
   `--pool=threads`: no change (50.4s). The ~100s aggregate import overhead is the price of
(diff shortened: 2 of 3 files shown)


thymikee:
P1 gate-honesty blocker: the check compares measured lengths only against the pin map in the same checkout, so the two prohibited changes remain green if the map changes too—grow a pinned file and raise its pin, or add a new >1,000-line file and add a pin. Current planted tests hold pins fixed and miss both bypasses. Add a base/history-backed monotonicity rule: existing pins may decrease/disappear but never increase, and new pins may not appear; plant both pin-edit cases red. Inventory and ordinary append/delete evidence are otherwise correct, and CI is green.

thymikee:
Addressed in df0493a88 (rebased onto current main).

The map is no longer the authority — git is. Two history rules now run beside the equality pin, over every file that is over the tripwire now plus every pinned file: (1) a test file over the tripwire may be no longer than it was at `git merge-base origin/main HEAD` (renames followed via `--find-renames`; a file that did not exist there may not cross the line at all), and (2) no pin may exceed its file's base length, so a pin cannot pre-authorize growth even before the file grows into it. Base lengths come from one `git cat-file --batch` spawn, parsed by bytes (the sizes are bytes; a string offset drifted after the first file with an em dash — caught while writing it). The gate fails loudly, not silently, when `origin/main` is unfetched; the Coverage job that runs the unit suite checks out with `fetch-depth: 0`.

Planted red against real git, map edited to match in both cases: append a line to `cli-network.test.ts` **and** raise its pin 1092→1093 → `is 1093 lines, 1092 at the merge-base: a test file over the tripwire may not grow, whatever its pin says` plus `pinned at 1093 but was 1092 lines at the merge-base`; add a fresh 1,001-line test file **and** a pin for it → `did not exist at the merge-base: a new test file may not cross the 1000-line tripwire, pinned or not`. In-file planted tests cover the same four outcomes with an injected base map, plus the byte-accurate parser. Full run 660ms.

thymikee:
P1 gate-honesty gap remains at exact `df0493a`: a new pin can still be added for an unchanged sub-tripwire file (for example a 900-line test pinned at 900). Equality passes and history passes because 900 ≤ base length, so the pin map grows despite the required “new pins may not appear” rule and only-shrink/kill criterion. Compare the pin map against base (rename-aware if intended), or reject pins for current files ≤ tripwire, and plant this exact arbitrary-new-pin case red. Existing growth/new-giant fixes otherwise look sound; iOS Smoke remains pending.

thymikee:
Addressed in e5cfd7f4b (rebased onto current main).

Pins now exist only for files over the tripwire: a pin whose file is at or under the line is a finding on its own — `<file> is 900 lines, at or under the 1000-line tripwire, but has a pin (900): remove it — pins are only for files over the tripwire` — independent of equality and history, so the map cannot grow by pinning small files at their own length. (It also subsumes the old "shrank under the line, remove the pin" message.) Planted red in-file with exactly that case, and against real git: pinning the 186-line `cli-agent-cdp.test.ts` at 186 → red.

One bootstrap note visible in the diff: `platforms/android/__tests__/snapshot.test.ts` is pinned 1660, not 1636 — `main` grew it in #1846 before this gate exists; the history rule agrees (1660 at the merge-base), so it is the pin catching up to main, not a raise.

thymikee:
Re-reviewed exact `e5cfd7f4`: clean and merge-ready. The prior arbitrary sub-tripwire pin bypass is fixed: any pin at or below 1,000 lines is independently rejected, and the planted 900/900 case plus real-file evidence are non-vacuous. Existing equality/history rules still cover growth, new giants, and raised or stale pins. No new gate-honesty finding. Exact-head CI is fully green, the PR is CLEAN/MERGEABLE, and no device evidence applies.

Issue #1832, opened 2026-08-18
Android native-a11y → snapshot: conform to the acquire/present engine contracts (sibling of #1797)


thymikee:
## Context

#1797 (iOS acquire/present split, hardened through three internal adversarial reviews and four external review passes) produced six platform-agnostic **Snapshot Engine Contracts**. This issue is the Android conformance audit against them and the work to close the gaps. iOS and Android have different *divergence axes* — iOS diverges **by backend** (tree / sweep / private AX), Android diverges **by API level and by layer** (platform parser vs daemon) — but the contracts and the target architecture are the same, and the shared-code fixes land once for both platforms.

The contracts (canonical statements in #1797):

- **C1 Fact-availability neutrality** — eligibility and derived fields never depend on whether an optional acquired fact was available; evidence is internal, never wire bytes.
- **C2 Hint conservatism** — an acquisition-shaping hint may reduce acquisition only when the adapter guarantees completeness for the requested projection; otherwise budget/order only.
- **C3 Two projections** — regular (clip, eligibility, hints, scope) vs raw (normalization only, no visibility pruning); `interactive ⊆ raw`.
- **C4 Geometry carrier** — regular emits the effective visible rect as `rect` (tap centering correct by construction); raw emits raw frames; raw geometry is presentation-internal; post-wire consumers use effective geometry by declaration.
- **C5 Presentation failure contract** — cooperative deadline, whole-tier discard with a named reason, separate acquisition/presentation timings, linear-complexity gate + hostile-tree benchmark.
- **C6 Output invariant + disclosure** — every framed node emitted by regular presentation intersects its cumulative effective clip; violations downgrade a quality verdict the backend actually emits.

#1797 additionally distinguishes **eligibility** (owned by presentation — the single-interpreter, backend/axis-neutral guarantee) from **publication membership** (owned by the daemon compaction layer — declared, test-pinned suppressions such as iOS's structural testID-wrapper rule). Android conformance uses the same vocabulary.

## Audit scorecard (HEAD 3020195)

Android's good news first: **one acquisition source** (the helper APK — no uiautomator fallback at HEAD), **one XML parser** (`readAndroidUiNodeMetadata`, shared by snapshot / fill-verification / content-recovery), and **one daemon presentation** shared with iOS. The iOS two-interpreter disease is absent at the parser level. The violations are cross-layer instead:

| Contract | Verdict | Evidence |
|---|---|---|
[...]

PR #1831, opened 2026-08-18
fix: derive iOS transition snapshots from visible presentation

thymikee:
## Summary

Fix iOS interactive snapshots so private-AX transition state is derived from the visible presentation before semantic scoping.

This projects hidden/offscreen transition content out of the actionable tree, preserves scroll hints, promotes conservative app-provided navigation-title affordances, and distinguishes collapsed from expanded action shelves using viewport geometry. It also keeps raw iOS, macOS, and Android scope pushdown unchanged.

Stable capture now confirms broad screen replacements for 1.5 seconds, including replacements that begin after the first post-action capture or recover from tree to private AX. The confirmation applies only to XCTest captures using the default quiet period; explicit short quiet periods and non-transition captures retain their existing timing. This prevents `--settle` from accepting a self-consistent intermediate Element tree.

The change also hardens Apple runner source packaging and PR XCTest selection so the coordinate text-input policy exercised in source is included in the runner that ships and executes before merge.

Scope disclosure: 25 files are touched across four connected parts of the same iOS capture boundary: private-AX presentation and fixtures, scoped snapshot capture, stable-capture transition confirmation (including tree-to-private-AX recovery), and Apple runner packaging/PR test selection. No Android behavior, public CLI surface, docs, or skills changed.

## Validation

- Rebased on `origin/main` at `4bba40424` (#1838); pushed head is `5b81b67e6`.
- `pnpm check:affected --run`: package, build, type, layering, lint, format, Fallow, metadata, and Node integration gates passed. Affected coverage passed 2,502/2,503 tests; the sole failure was an unchanged iOS recorder-provider test timing out in `/bin/ps` under concurrent host XCTest load.
- The same recorder-provider scenario failed on plain `origin/main` under identical load (2/2 failures), including the same incomplete process-identity error, classifying it as baseline/host contention rather than a PR regression.
- The coordinate input regression is selected in the iOS PR workflow. Pre-fix mutant evidence: ignoring `isEnabled` failed 1/1 at the new assertion; restored code passed 1/1 in 0.017 seconds.
- Transition regression evidence: deterministic fixtures for both `tree -> private-ax` recovery and a replacement beginning after the first capture failed by settling on the transitional `action file`; the fix passes all 17 focused stable-capture tests. `pnpm check:quick` passes.
- Exact-head unpublished package SHA-256: `2a1b036b14bab4cd8c692b85e0350db85820c6173b1355ac361db71ac505e04a`; the installed daemon bundle matches the branch build byte-for-byte.
[...]

PR #1846, opened 2026-08-18
fix(android): resolve snapshot scope once and disclose the API 23 occlusion-scan gap (#1832 C1/C2)

thymikee:
## Summary

First slice of #1832 (Android conformance to the acquire/present engine contracts) — the two **immediate fixes** the issue lists as independent of the redesign. Android-only behavior plus one shared scope specification; the iOS sibling (#1797, in flight) is untouched.

**C2 — `--scope` resolves once on Android, under one specification.** Before, an Android scope went through two passes with contradictory semantics: the platform matcher (breadth-first substring, **fell back to the full tree on no match**) and then the daemon's post-wire `scopeSnapshotNodes` (document order, **empty on no match**). Now Android resolves scope exactly once, inside its projection, and the daemon pass skips the `android` backend. The rule is written down once — `matchesSnapshotScope` in `@agent-device/contracts/snapshot` — and pinned by a golden table, `contracts/fixtures/snapshot-scope-policy.json`, asserted against every TS runtime that resolves scope (the predicate, the Android projection, the daemon pass); the Swift runner twin consumes the same table when #1797 lands scope-in-presentation.

Specification (what agents observe): the scope root is the **first node in document order** whose label, value, or identifier contains the scope text (case-insensitive); the snapshot is that root's subtree re-rooted at depth 0; **no match → empty snapshot**, never the full tree. Agent-visible output is unchanged for the common cases (the daemon pass already imposed empty-on-no-match); the delta is that the platform no longer picks a *shallower later* match over an *earlier deeper* one, and `--depth` under `--scope` keeps its scope-relative meaning because scope stays in the projection.

**C1 — disclosure only.** The covered-sibling pruner keys on `drawing-order`, which the helper serializes on API 24+ only; on API 23 the same screen presents a different node set with nothing saying so. `androidSnapshot.occlusionScanUnavailable: true` now discloses a helper tree that carries no `drawing-order`. Response-level metadata, no per-node bytes. As the issue states, this does **not** restore neutrality — C1 stays open until occlusion moves to the daemon annotator (migration step 5).

Not in this PR (tracked on #1832, next slices): C3 (three pruners move out of `parseUiHierarchyTree` into the regular projection so `--raw` is the acquired tree), routing the system-dialog call site through `buildSnapshotState`, residue declarations (missing `checked/selected/…`, freshness signature keying on `selected`, truncation×scope), then the shared C4/C5/C6 steps aligned with #1797.

Refs #1832 (partial: immediate fixes 1 and 2).

## Validation

- Golden table legs green; the Android leg proven red against the pre-fix matcher (3/3 fail: document-order case, no-match case, no re-scope after wire).
- Live on a fresh headless Pixel 7 / API 36 AVD (helper 0.20.9, `android-helper` backend): `snapshot -i -s smartspace_card_pager` returns the 3-node subtree re-rooted at depth 0; `snapshot -s "Play Store"` returns the single leaf; `-s zzzz-no-match` returns 0 nodes in regular and `--raw`; `androidSnapshot` carries no `occlusionScanUnavailable` on API 36. No API 23 image available locally — that shape is covered by the unit fixture. Session closed, throwaway AVD deleted.
- `pnpm check:affected --run` green locally.

Docs: `website/docs/docs/commands.md` gains the `--scope` semantics line and the `occlusionScanUnavailable` disclosure. 14 files; scope stayed within the Android snapshot family plus the shared scope predicate.

## Adversarial review round (commits 2–3)

An adversarial pass against this PR found a blocker and four real gaps; all are fixed here.

- **Blocker — interaction captures reached the Android platform unscoped.** `captureSnapshotData` spread `snapshotScope: undefined` over `flags`, so `press|click|fill|longpress|hover <selector> --scope X` and `--settle` observation (which carry the scope in `flags.snapshotScope` only) lost it before the platform, while `buildSnapshotState` still saw it. The post-wire pass used to rescue that; skipping it for `android` turned it into a silently unscoped tree — wrong-element taps or spurious `AMBIGUOUS_MATCH`. One effective scope now feeds both sides, with a regression test on the production route (red on the previous commit: platform received `undefined`).
- **Scope now resolves over the presented tree of the requested projection**, not the acquired tree, matching the domain the daemon pass uses on iOS. An acquired match that membership drops no longer empties the snapshot.
- **Slicing happens after the walk**, so ancestor context above the scope root (clickable row, collection, chrome) still shapes `-i` membership inside it — scoped `-i` is a subset of unscoped `-i` — and `--depth` stays scope-relative. Both pinned, both red on the previous commit.
- `findSnapshotScopeRange`/`reindexSnapshotNodes` are shared, so the daemon pass and the Android projection run one implementation; the scope slice moved to `ui-hierarchy-scope.ts` (mirrors its test, and takes `ui-hierarchy.ts` back under the 1,000-LOC tripwire). `parseUiHierarchy` moved to a test fixture module — it had no production caller left (`check:production-exports` back to main's 19).
- Golden rows sharpened (the value row matched via label on Android and proved nothing there); the Android leg now runs raw **and** regular. CHANGELOG entry added; the docs sentence corrected for iOS and `@ref`.

Live re-verification on a throwaway Pixel 7 emulator: `is visible text="Tue, Aug 18" --scope "At a glance"` passes, `click text="Play Store" --scope "At a glance"` refuses with "Selector did not match" instead of tapping (unscoped control: it matches). Session closed, AVD deleted. `pnpm check:affected --run` green end to end.

## Review round 2 (thymikee P1) — commit 4

Scope was resolving strictly over presented nodes, so a matched container the interactive projection drops returned an **empty** snapshot: on device, `snapshot -i --scope hotseat` — the structural `id/hotseat` container holding Phone/Messages/Chrome — answered `Snapshot: 0 nodes`.

The rule is now: **the scope root is the first document-order match whose subtree still contributes at least one node to the requested projection**, and the result is that subtree's presented nodes re-rooted at depth 0. One rule, both failure modes gone — a decorative match that presents nothing no longer empties the snapshot, and a dropped container still scopes to its content. Scoped `--depth` now filters the depths the response prints, so a node shown at depth 0 survives `--depth 0` (the unscoped path still cuts by acquired tree depth — pre-existing, noted in code and on #1832).

Tests: the "not re-scoped after the wire" case runs a real parsed scoped tree instead of fabricated depth-0 siblings; two dedicated projection tests cover the excluded-container shape (red on the previous commit) and the skipped-candidate shape. The golden legs stay raw + regular on purpose — the table's rows render as bare TextViews that `-i` drops wholesale, so an `-i` leg there would pass vacuously while measuring membership rather than scope.

Live: `-i --scope hotseat` returns the icon row at depth 0; `-i --scope hotseat -d 0` keeps it. Rebased onto `6a8beb6`; `pnpm check:affected --run` green.


thymikee:
P1: re-root scoped interactive output after membership filtering. Scope is resolved before projection; when the matched root is a non-actionable structural container, `interactiveOnly` drops it but retains a clickable child at depth 1—or drops that child under `--depth 0`—contradicting the new scope-relative depth contract. Current tests miss production behavior: Android golden runs only raw, cross-projection covers no-match, and the “not re-scoped” case fabricates depth-0 children. Add regular/interactive golden legs using an excluded matched container and normalize retained scoped roots/depth. API 23 response-level disclosure otherwise looks sound.

thymikee:
Rebased onto main (`6a8beb6`, CHANGELOG conflict resolved by keeping both entries) and addressed the P1 — it was live, in a worse form than described, and it is now fixed with your test asks.

**What I found.** Your comment landed against the first commit; a second commit had meanwhile moved matching to the presented tree. That over-corrected: it did not leave the clickable child at depth 1, it returned **nothing at all**. On device, `snapshot -i --scope hotseat` (a structural `id/hotseat` container the interactive projection drops, holding the Phone/Messages/Chrome icons) answered `Snapshot: 0 nodes`.

**Fix.** The scope root is now the first document-order match **whose subtree still contributes at least one node to the requested projection**, and the result is that subtree's presented nodes re-rooted at depth 0. Both failure modes die with one rule: a decorative heading that matches but presents nothing no longer empties the snapshot (that was the reason for matching presented nodes), and a container membership drops still scopes to the content inside it (your P1). Live now: `snapshot -i --scope hotseat` → `@e1 [text] "Phone" … ` at depth 0.

**`--depth` half.** Also real, also fixed: scoped `--depth` now filters the depths the response actually prints, so a node shown at depth 0 cannot vanish at `--depth 0` (verified live: `-i --scope hotseat -d 0` keeps the icon row and drops the Google-search group's children). Note the unscoped path still cuts the walk by acquired tree depth, which can differ from the compacted depth printed beside each node — pre-existing, out of this PR's scope, recorded in the code comment and on #1832.

**Tests.** `not re-scoped after the wire` now runs a real parsed scoped tree instead of fabricated depth-0 siblings. Added two projection tests on actionable shapes: the excluded-container case (red on the previous commit) and the skipped-candidate case. On the golden legs: I ran raw + regular and deliberately did **not** add an `-i` leg there — the table's rows render as bare TextViews, which `-i` drops wholesale, so an `-i` leg would have measured membership rather than the scope rule and passed vacuously. The two dedicated tests carry the `-i` interplay instead; comment in the file says so.

`pnpm check:affected --run` green. Live re-verification and cleanup done (throwaway AVD, session closed).

thymikee:
Rebased again — onto `f03c030` (#1831), which landed on the same lines. Worth noting because the two changes are the same insight on different platforms, and they compose rather than collide:

- #1831 keeps iOS presentation context for scoped snapshots by **stripping `snapshotScope` from the capture flags** for iOS `-i` and scoping post-wire, after `presentIosInteractiveSnapshot`.
- This PR keeps Android presentation context by scoping **inside** the Android projection (after the walk, over presented content) and skipping the post-wire pass for that backend.

Resolution keeps both: `snapshotCaptureFlagsForBackend(device, resolveSnapshotStateFlags(params))` — main's iOS stripping wrapping this PR's single-effective-scope fix (the one that stopped interaction captures reaching the platform unscoped) — and the post-wire pass now runs on `presentableNodes` per #1831 while `backendScopesAfterWire` still excludes `android`. iOS behavior is unchanged from main; Android is unchanged from this PR.

The earlier check run was misleading, by the way: the PR was `DIRTY` after the first rebase, so GitHub could not build the merge ref and only the CodeQL jobs ran — the CI/Android/iOS/macOS/Linux runs you saw green were still from the first commit. It is `UNSTABLE` now and the full matrix is running against `b6ea357`.

`pnpm check:affected --run` green on the rebased head.

thymikee:
Re-reviewed exact `b6ea3571`: code-clean and ready-for-human. Scoped Android parsing now preserves acquisition context through membership, selects the first match contributing presented nodes, then slices/re-roots/reindexes and applies depth relative to emitted output. Regressions directly cover a dropped structural root with retained child (`--depth 0`) and skipping an earlier decorative non-contributing match; daemon no-rescope proof and Pixel hotseat evidence are relevant. Required exact-head CI is still running before merge.

PR #1852, opened 2026-08-18
fix(layering): list the whole zone when R10's type-cycle ceiling is exceeded

### LINKED ISSUE #1837: check:layering R10 names the wrong file when a zone exceeds its type-cycle ceiling
Found while diagnosing today's main break (#1825 × #1779).

`checkTypeCycleBaseline` in `scripts/layering/daemon-modularity.ts` reports the offending file as `members.find(<zone match>)` — i.e. the alphabetically first member of the over-budget zone — not the member that joined the cycle. Today's failure read `[R10] 17 daemon-server files (baseline 16) at src/daemon/daemon-command-registry.ts`, but that file was already a member at `a853734f0` (46). The +1 node was `src/daemon/handlers/snapshot-interactor-capture.ts` (new in d76e0f94e, wedged between `handlers/snapshot-capture.ts` → it → `core/interactors.ts`), found only by diffing `largestTypeCycleMembers` between the two commits.

Fix: when the measured count exceeds the ceiling, print the members that are new relative to the previous state (or the whole zone member list) instead of the first alphabetical hit, so the author sees which edge to cut. Small; no policy change.

Refs: #1781 A6, #1825, #1779.

thymikee:
## Summary

`pnpm check:layering` R10 named the wrong file when a zone exceeded its type-cycle ceiling. `checkTypeCycleBaseline` reported `members.find(<zone match>)` — the alphabetically-first zone member — so the #1825 × #1779 main break read `[R10] 17 daemon-server files (baseline 16) at src/daemon/daemon-command-registry.ts`, a file that had been in the cycle all along. The actual +1 (`src/daemon/handlers/snapshot-interactor-capture.ts`) was found only by diffing `largestTypeCycleMembers` between two commits, ~40 min in.

`LARGEST_TYPE_CYCLE_ZONE_CEILINGS` records a **count per zone, not a membership**, and the Layering Guard job checks out at depth 1, so there is no previous state to diff against inside the gate. The violation now lists every member of the over-budget zone (sorted) and annotates the ceiling table (`scripts/layering/daemon-modularity.ts:1`, like the sibling R9/R10 baseline violations) instead of an arbitrary member.

Before:
```
[R10 daemon-modularity] src/daemon/daemon-command-registry.ts:1 — the largest type cycle now contains 17 daemon-server file(s) (baseline 16); extraction must not trade one zone's locality for another's.
```
After (planted: ceiling lowered 16 → 15 on the real tree):
```
[R10 daemon-modularity] scripts/layering/daemon-modularity.ts:1 — the largest type cycle now contains 16 daemon-server file(s) (baseline 15); extraction must not trade one zone's locality for another's. 1 over the ceiling — the member(s) that joined are among these daemon-server files: src/daemon/daemon-command-registry.ts, src/daemon/deferred-interaction-outcome.ts, src/daemon/handlers/response.ts, src/daemon/handlers/snapshot-capture.ts, src/daemon/interaction-outcome-policy.ts, …, src/daemon/session-store.ts. Cut the edge that pulled them in rather than raising the ceiling.
```

The count is deliberately phrased as net overflow, not a join count: a change that adds two members and removes one prints `1 over the ceiling`, so the message bounds nothing except that at least one listed file is new.

No policy change: same ceilings, same rule id, same pass/fail set. Recording per-zone member lists (which would let the gate print the exact diff) would turn the count ratchet into a membership pin — a stricter policy the issue explicitly does not ask for; noted as an option, not taken.

Closes #1837. Refs #1781 A6, #1825, #1779.

## Validation

- New unit test `R10 zone overflow lists the whole zone so the joining member is visible` plants a daemon-server +1 (`src/daemon/snapshot-interactor-capture.ts`, chosen to sort *after* the probes so the old first-member pick could not name it by luck; commands −1 keeps R9's total pinned) and asserts the single violation annotates the ceiling table and names all 17 members.
- **Red against pre-fix code** (`git stash` the production file, run `scripts/layering/daemon-modularity.test.ts`):
  ```
  ✖ R10 zone overflow lists the whole zone so the joining member is visible
    AssertionError [ERR_ASSERTION]: Expected values to be strictly equal:
    + actual - expected
    + 'src/daemon/probe-0.ts'
    - 'scripts/layering/daemon-modularity.ts'
  ℹ pass 9 / fail 1
  ```
  — the old code names the alphabetically-first probe. Restored: `ℹ pass 10 / fail 0`.
- Planted structural violation on the real tree (ceiling 16 → 15) prints the message quoted above; restored, `node scripts/layering/check.ts` is `Layering guard: OK … the largest type-level cycle is 46 files (R9)`.
- `pnpm check:affected --run`: all runnable checks passed (layering, lint, typecheck, format, mutation-selection tests).

Touched files: 2 (`scripts/layering/daemon-modularity.ts`, `scripts/layering/daemon-modularity.test.ts`). Scope stayed on the layering gate. No docs/skills change: the gate's own message is the user-facing surface.

### DIFF
--- a/scripts/layering/daemon-modularity.test.ts
+++ b/scripts/layering/daemon-modularity.test.ts
@@ -240,7 +240,7 @@ test('R10 zone overflow lists the whole zone so the joining member is visible', 
   for (const member of daemonMembers) {
     assert.ok(violation!.message.includes(member), `${member} missing from: ${violation!.message}`);
   }
-  assert.match(violation!.message, /1 of these joined with this change/);
+  assert.match(violation!.message, /1 over the ceiling — the member\(s\) that joined are among/);
 });
 
 // Growth was always rejected; a baseline left ABOVE the measured size used to be a suggestion
--- a/scripts/layering/daemon-modularity.test.ts
+++ b/scripts/layering/daemon-modularity.test.ts
@@ -219,6 +219,30 @@ test('R9 records zone ceilings and keeps engine files outside the largest compon
   assert.ok(violations.some(({ message }) => /engine file entered/.test(message)));
 });
 
+// #1837: the zone violation used to name the alphabetically-first zone member — a file that had
+// been in the cycle all along — so the +1 was found only by diffing member lists between commits.
+// The ceiling records a count, not a membership, so the message lists every zone member instead.
+test('R10 zone overflow lists the whole zone so the joining member is visible', () => {
+  const zones = DAEMON_MODULARITY_BASELINE.largestTypeCycle.zoneMembers;
+  // Sorts after the daemon-server probes: the old first-member pick could not name it by luck.
+  const joined = 'src/daemon/snapshot-interactor-capture.ts';
+  const members = [...baselineTypeCycleMembers({ commands: zones.commands - 1 }), joined].sort();
+  const daemonMembers = members.filter((member) => member.startsWith('src/daemon/'));
+  assert.notEqual(daemonMembers[0], joined);
+
+  const violations = checkDaemonModularityRatchets(baselineEdges(), members);
+
+  assert.equal(violations.length, 1);
+  const [violation] = violations;
+  assert.equal(violation!.rule, 'R10 daemon-modularity');
+  assert.equal(violation!.file, 'scripts/layering/daemon-modularity.ts');
+  assert.match(violation!.message, /contains 17 daemon-server file\(s\) \(baseline 16\)/);
+  for (const member of daemonMembers) {
+    assert.ok(violation!.message.includes(member), `${member} missing from: ${violation!.message}`);
+  }
+  assert.match(violation!.message, /1 of these joined with this change/);
+});
+
 // Growth was always rejected; a baseline left ABOVE the measured size used to be a suggestion
 // in the success line, which is headroom the next change spends without a number moving.
 test('R9 rejects a baseline left above the measured cycle', () => {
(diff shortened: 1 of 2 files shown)


thymikee:
Diagnostic-accuracy gap: the whole-zone member list is useful, but ``${zoneMembers.length - allowed} of these joined with this change`` is unsupported because the gate stores only a count ceiling, not prior membership. Net excess cannot prove how many entered: members may be replaced, or one edge may pull several existing files into the SCC.

Report only that the zone is N above its ceiling and direct authors to changed files/import edges; add a replacement/edge-closure regression.

Exact-head CI is otherwise green; no device evidence applies.

thymikee:
Out of draft. CI is green on the actual head `60f88f3d` (rebased onto current `origin/main`, so it carries #1860's corrected ratchet pin): 28/28 checks pass, Layering Guard included.

Second commit `60f88f3d` applies the review nit: the overflow line was phrased as a join count, but it is net growth over the ceiling (two joins plus one departure would print `1`), so it now reads `N over the ceiling — the member(s) that joined are among these <zone> files: …`. The pinned assertion in `daemon-modularity.test.ts` moved with it; the red-then-green proof re-ran on the new text (planted +1 → `pass 9 / fail 1` against pre-fix code, `pass 10` after), and the planted-ceiling run on the real tree prints the message quoted in the PR body.

thymikee:
Reviewed exact head `60f88f3d`: clean and good to go. The revised wording correctly reports net overflow rather than claiming an exact join count, the planted replacement case proves the zone-specific diagnostic and defeats the old alphabetical attribution, and listing the whole zone satisfies #1837 without changing policy. Scope/body are accurate; all 28 exact-head checks are green. No device evidence applies to this layering-only change.

PR #1836, opened 2026-08-18
docs(agents): ADR 0019 unit checklist, owning-seam mock rule, worktree and rebase guidance

thymikee:
## Summary

Retro follow-up, item 2 of the agreed list. Four agent-facing doc changes; no code.

- **`docs/agents/adr-0019-unit.md` (new)** — the checklist for one command unit: what to post on #1739 before writing code, one integration owner for the shared spine, the "can this be made impossible at the owning interface?" question *before* any cutover-row extension, then the unit in order with the declaration site for each step (use → plan → handler admit/bind-once → descriptor flip → platform packages → cutover row → deletions → tests on the runtime seam → facets), the evidence a unit review must carry, and what "done" is not. Linked from the AGENTS.md task table. Everything in it was rediscovered on #1779; ADR 0019 keeps the why, this keeps the how.
- **`docs/agents/testing.md`** — mock the seam the code under test consumes: fake `inspectFacts` / `bindDevice` (fixture, harness, and facts-builder locations named), not `vi.mock('.../core/dispatch.ts')`. Sixty-odd files still mock dispatch; retiring `dispatchCommand('snapshot')` surfaced them one failure at a time. Don't add to the set; a migrating command moves its tests off it in the same PR.
- **`AGENTS.md`** — fresh-worktree preflight (`pnpm install --frozen-lockfile && pnpm build` *in* the worktree, else `@agent-device/*`/`ai` resolve against another checkout and smoke lanes fail with "Missing dist build"; layering scan reads tracked files only) and concurrent-agent hygiene (one full gate per host — parallel suites *are* the contention; verify subagent edits with `git -C`; one PR per worktree; PRs by URL not recalled number; per-agent scratch dirs).
- **`docs/agents/pull-requests.md`** — two readiness claims that "don't wait for CI" kept blurring: *published and reported* vs *merge-ready*. And the rebase rule: `main` has no up-to-date protection, so rebase on conflict or when `pnpm check:affected --base <merge-base> --head origin/main` names a surface yours depends on; evidence is stamped with the commit it was gathered at, so a rebase dates it rather than invalidating it. Merge queue deferred until migration units regularly overlap.

## Validation

[...]

PR #1857, opened 2026-08-18
ci: give check:affected real Apple ownership rules and route ios.yml on them (#1781 A9-2)

thymikee:
## Summary

`pnpm check:affected` gains **device-lane ownership by platform family**, and `ios.yml` is **routed on it** at `pull_request` time via a `paths-ignore` list that the gate manifest holds to the selector over every tracked path, both ways. Part of #1781 (A9-2).

Before: a TypeScript-only Apple change (`src/platforms/apple/**`, `packages/platform-apple/**`) produced a plan with **no iOS check at all** — `swift-runner-ios` fired only for `.swift`/`apple/runner/**`, and `replay-ios` was reachable only through a full fail-open. `ios.yml` ran on every non-docs PR, including Android-only and unit-test-only ones.

After:

```
$ pnpm check:affected   # diff touching src/platforms/apple/core/apps.ts
  - replay-ios: iOS simulator replay suite (GitHub-authoritative; not run locally)
      · src/platforms/apple/core/apps.ts [own:device-lane:apple] — apple-owned tree: only the apple device lanes execute it
  - replay-ios-device … - replay-macos …
$ pnpm check:affected   # diff touching src/platforms/android/perf.ts
  - replay-android …      (no iOS check; ios.yml does not start)
$ pnpm check:affected   # diff touching src/platforms/android/perf.ts + package.json
Fail-open: selecting the full check set.   (ios.yml starts)
```

`push` to `main` is unfiltered: main runs every lane unconditionally.

### The four lines

- **Catches:** nothing new — this is a cost item. It removes iOS-lane runs the lane cannot use (unit-test-only and other-family-only PRs) and, as a side effect, makes `check:affected` tell the truth about Apple changes (TS-only Apple PRs such as #1776/#1731/#1769 now carry `replay-ios` instead of a plan that omits iOS entirely).
- **Evidence:** 90-day retrospective replay below, plus an independent adversarial re-derivation (692 merged PRs / 60 days, 31 incremental would-skips cross-referenced against all 118 red `ios.yml` PR runs of the last 30 days → 0 hits) — 871 merged PRs, 3,286 `pull_request` `ios.yml` runs; would-skip-and-really-broke-the-runner = **0**. Planted-red proofs: `scripts/check-affected/device-lanes.test.ts` (Apple file → `replay-ios`; Android-only + any tooling file → full fail-open) and `scripts/gate/routing.test.ts` (an ignore hiding `src/platforms/apple/**` or `.github/actions/**` fails the manifest; dropping `src/platforms/android/**` or the unit-test ignore fails it the other way).
- **Cost:** wall-clock +0 (GitHub evaluates `paths-ignore` before it allocates a runner; no plan job on the critical path, no `if:` on the 12 smoke steps); macOS runner occupancy −61 runs / −691 min per 90 days over what A9-1 already saved (1.9% of the lane; last 30 days: −31 runs / −422 min of 16,490); red-lane attention 0 (no new lane); harness ≈ 165 LOC selector + 100 LOC manifest assertion + 23 YAML lines, all under existing gates (`check:affected:test`, `check:gate-manifest:test`).
- **Kill-criterion:** was "one would-have-been-skipped PR that broke the runner for a real reason" — not met (0/35). Going forward: if `main` goes red on `ios.yml` for a change a routed PR skipped (a cross-family lifecycle call, see residual risk), drop the routing entries from `ios.yml` and the `ROUTED_LANES` record in the same commit; the ownership rules stay.

## Design note

**What enumerates the Apple surface?** Not the import graph. The daemon value-imports every family: `src/platform-runtime.ts` composes all six platform packages, `src/core/interactors/*` import each family's actions, and `src/daemon/session-teardown.ts` / `src/daemon/server/daemon-runtime.ts` call Android and web lifecycle helpers on every session close and daemon start. Reachability from `src/bin.ts` is therefore the whole tree, and the layering graph additionally does not resolve relative imports inside `packages/` (intra-package edges are invisible; `packages/contracts/src/facades/*` fan out to nothing). Transitive ownership derived from that graph would either classify everything as Apple or miss what matters.

What the repo *does* enforce is the platform partition: `CANONICAL_PLATFORM_FAMILIES` (`scripts/layering/platform-package-policy.ts`) names the families, layering R13 pins each family's runtime to `packages/platform-<family>/` and `src/platforms/<family>/`, and the remaining family-owned trees are named by a family or Apple-leaf **directory segment** (`android/`, `linux/`, `test/integration/replays/<leaf>/`, `src/daemon/snapshot-presentation/ios/`) or, under `test/integration/`, by the lane prefix of the smoke file. So `scripts/check-affected/device-lanes.ts` — the single declaration site — says:

| path | leaf | lanes |
|---|---|---|
| tagged by exactly one family / Apple leaf | that leaf | `apple` → replay-ios, replay-ios-device, replay-macos · `ios` → replay-ios, replay-ios-device · `macos` → replay-macos · `android` → replay-android · `linux` → replay-linux · `web` → web-smoke · `harmonyos`/`vega` → none |
| untagged, or tagged by two families | shared | every device lane |
| `examples/test-app/**` | fixture-app | replay-ios, replay-ios-device, replay-android |
| unit test under `src/`, `packages/*/src/` | — | none (no device lane runs Vitest; the build excludes them) |
| non-TS the selector cannot place (`src/*.json`, a `.png` under `test/integration/`) | — | unchanged: still fails open through the ambiguous-path guard |

Filename tags in shared directories are deliberately **not** used (`src/daemon/android-system-dialog.ts` stays shared) — a naming convention is not a boundary. `contracts/fixtures/*.json` (TS/Swift golden tables) get their own build rule: `unit` + `swift-runner-ios` + `swift-runner-macos` instead of fail-open.

**Decide before allocating macOS.** Three mechanisms were compared: (a) an ubuntu plan job the smoke job `needs` — adds ~1 min serial latency to all 3,286 iOS runs to skip 61, on the lane that is usually a PR's critical path; (b) computing the plan inside the macOS job and skipping the rest — the runner is still allocated for ~2 min, and 12 steps need an `if:`; (c) `paths-ignore` — GitHub evaluates it before creating the job, so it costs nothing and adds no latency, but it is a hand-written glob list next to a derived selector. (c) is shipped, made honest by `scripts/gate/routing.ts`: for every tracked path, the lane **must start** when the selector fails open on it or routes it to one of the lane's declared gates (`swift-runner-ios`) or sampled checks (`replay-ios`, `replay-ios-device`, declared once in `ROUTED_LANES`), and **must not start** on a path the selector classifies as another family's or as a unit test. Docs paths are exempt; a `.github/**` path is exempt only when ignored by its exact name (deploy/docs-preview workflows the lane does not use — a glob there would fail).

**Fail-open on tooling.** Unchanged and now asserted: `.github/**`, `scripts/**`, `package.json`, `pnpm-lock.yaml`, `tsconfig*.json`, `vitest*.ts`, `packages/*/package.json|tsconfig.json`, and the selector's own sources fail open to the full set, so any PR touching one of them starts `ios.yml` regardless of what else it touches (`routing.test.ts` plants `.github/actions/**` and `package.json` into the ignore list and expects the manifest to fail).

**Residual risk — large PRs.** GitHub's path filters examine only the **first 300 changed files**. A PR whose first 300 paths all match the ignore list skips the lane even if file 301 is Apple code (in-window examples of PRs past the limit: #1499 at 511 files, #1494 at 865). Inherent to `paths-ignore` and shared with A9-1, but this list matches many more files than A9-1's did, so the exposure is larger. Backstop is the unfiltered `push` trigger on `main`. Recorded in `docs/agents/testing.md`.

**Residual risk (stated, not hidden).** A family-tagged production file *can* execute on another family's lane through a cross-family lifecycle call — 50 files under other-family roots are directly value/dynamic-imported by shared code today (`pnpm depgraph`), e.g. `session-teardown.ts → src/platforms/android/perf.ts`, `daemon-runtime.ts → src/platforms/web/agent-browser-lifecycle.ts`. Backstops: `ci.yml`'s node-integration lanes start a real daemon (import-time and startup breakage), the unit/provider suites, and the unfiltered `push` trigger on `main`. The 90-day replay found no instance. A seam-based refinement (treat those 50 as shared, derived from the layering graph) is available if this ever bites; it was not shipped because it would need `paths:` with 50 negations and churn on every new cross-import.

**Also fixed on the way:** `check:gate-manifest`'s path-coverage assertion now exempts declared manual-only checks the way its `owned` assertion already did (the parked `replay-ios`/`replay-android` lanes are printed by name on every run; repeating the gap once per path that selects them buried the real findings).

## Retrospective replay (90 days, ran before routing)

Method: every PR merged since 2026-05-20 (GraphQL, files + commit oids), every `pull_request` `ios.yml` run since 2026-05-15 (REST, 4,455 runs), matched by head SHA with a branch-name fallback inside the PR's lifetime; each PR's file list run through the **new** selector in-process, and separately through the **new** `paths-ignore` list.

| | PRs | ios.yml PR runs | macOS min |
|---|---|---|---|
| merged PRs in window (27 with >100 files counted as "would run") | 871 | 3,286 | 35,541 |
| plans that fail open (tooling/workflow/scripts) | 345 (40%) | — | — |
| skipped by `paths-ignore` as of main (A9-1, docs-only) | 39 | 49 | 437 |
| skipped by the new list | 74 (8.5%) | 110 | 1,128 |
| **incremental: routing adds** | **35 (4.0%)** | **61 (1.9%)** | **691 (1.9%)** |
| incremental, last 30 days | 18 | 31 of 1,267 | 422 of 16,490 |
| incremental would-skip with a red `ios.yml` run | 1 | 1 | — |
| incremental would-skip that **broke the runner** | **0** | | |

Incremental would-skip by class: Android-only 14 (e.g. #1808 #1737 #1734 #1561 #1513 #1378 #1314 #1285 #1090 #903 #796 #788 #671 #1531), unit-tests-only 20 (e.g. #1790 #1642 #1552 #1113 #595), web-only 1 (#835). Their 61 runs: 53 green, 7 cancelled, 1 red.

The one red: **#796** (`src/platforms/android/snapshot-helper-session.ts` + its test), run 27434816426, step "Prepare iOS runner" — `Daemon request timed out` at the then-300 s `prepare ios-runner` budget (a cold xcodebuild). The same step failed the same day on two unrelated PRs (runs 27420294661, 27407004126); the PR's follow-up commit changed two Android lines plus a test and went green; the file set cannot reach `xcodebuild`. Classified **infra flake**. (Under the selector-only view a second red appears, #1347, docs-only `CONTEXT.md` + `docs/adr` — "Boot iOS test simulator" failed; docs cannot break the runner and A9-1 already skips it.)

Selector vs YAML agreement over the 871 PRs: the YAML would-skip set equals the selector's except four PRs the YAML skips and the selector fails open on — #1836/#1547 (`docs/agents/testing.md`, selector-owning prose) and #607/#603 (docs-preview workflows ignored by exact name), the two documented exemptions.

## Review round (independent adversarial review — routing approved)

Findings addressed in the two commits at the head of this branch (`ci: bound the routed-lane exemption to sibling workflows` and `ci: close the routed-lane exemption over composite-action support files`):

1. **Real hole, fixed in code.** The exact-name `.github` exemption was unbounded: the reviewer planted `setup-apple-runner-build/action.yml` and `boot-ios-test-simulator/action.yml` — the lane's own build and boot actions — into `paths-ignore` and `check:gate-manifest` stayed green. `Lane` now carries `uses`: the transitive local composite-action closure plus the workflow's own file, from the same walk `declaredGates` performs; the exemption refuses anything in it. Both planted cases (plus `run-gate/action.yml`, reached only through a composite action, and `ios.yml` itself) are pinned in `routing.test.ts` and **proven red against the pre-fix code**.

   **1b. The same hole one level deeper, also fixed.** A follow-up review found the closure stopped at the YAML boundary: it recorded each action's `action.yml` but not the files that descriptor *executes*, so `.github/actions/setup-fixture-app/fetch-artifact.sh` — genuine iOS-lane machinery, run as `bash "$GITHUB_ACTION_PATH/fetch-artifact.sh"` — was still exemptible. The closure's unit is the action's **directory** now. That is deliberate over parsing `$GITHUB_ACTION_PATH/...` out of the YAML: `fetch-artifact.sh` itself runs `resolve-artifact-name.sh` and `trusted-artifact.mjs`, references that exist only inside shell, so a YAML-parsing fix would have repeated the same one-level-short failure. A directory needs no shell model and cannot miss a file however deep the chain runs; the coarseness is harmless because a file in an action's own directory belongs to that action. All three files are pinned and red against the descriptor-only closure.
2. **Wrong remedy in a correct failure, fixed.** A tracked non-TS file under an ignored family root (`src/platforms/android/probe-fixture.json`) correctly failed, but told the reader to remove the ignore entry — which would un-route every sibling `.ts` in that tree. Fail-open causes are now split: `unknown-path`/`ambiguous-path` ask for a selector owner, everything else still asks for the entry's removal. Pinned, also red against the pre-fix code.
3. **300-file limit, documented.** See the residual-risk note above and `docs/agents/testing.md`.

## Merge condition (ratchet-class — do not skip)

The routing assertion runs over **every tracked path** and reads every workflow, so it behaves like the layering/coverage ratchets: an in-flight PR that adds a family-tagged path, a non-TS file under an ignored root, or a new composite-action support file is green on its own base and turns `check:gate-manifest` red on `main` after both land (the #1825 × #1779 failure mode). **Re-run `pnpm check:gate-manifest` against whatever `main` is at merge time** — not against the base this PR was last rebased onto, which goes stale within hours — and merge after any in-flight PR that adds paths under `src/platforms/*`, `packages/platform-*`, `android/`, `linux/`, `test/integration/replays/*`, or `.github/actions/*`.

## Validation

- `check:affected:test` (47), `check:gate-manifest:test` (45, incl. the 8 new routing witnesses and the path-coverage exemption witness), `depgraph:test` (22) green; `pnpm check:gate-manifest` green (`46 checks wired across 33 lanes, manual-only: replay-android, replay-ios, replay-ios-device`).
- Planted red, locally (all pinned as tests): adding `src/platforms/apple/**` to `ios.yml` `paths-ignore` → `routing: ios.yml ignores src/platforms/apple/core/apps.ts, but the selector routes it to "replay-ios", "replay-ios-device", "replay-macos"`; adding `.github/actions/**` → `… fails open on it (workflow-tooling)`; removing `src/platforms/android/**` → `ios.yml starts on src/platforms/android/adb-executor.ts, which the selector classifies as android-owned (lanes: replay-android)`. Plus the two review plants above. All are pinned in `scripts/gate/routing.test.ts`; the two review cases were re-run against the pre-fix code and fail there.
- This PR itself fails open (`.github/**`, `scripts/check-affected/**`, `package.json`, `docs/agents/testing.md`), so `ios.yml` runs on it — CI on this PR is the live proof that the fail-open direction still starts the lane.
- Not verified live: a PR whose files are all under the new ignore list not starting `ios.yml`. That is GitHub's `paths-ignore` semantics ("skip when every changed file matches"), the same mechanism A9-1 relies on; the first Android-only or test-only PR after merge will show it.

Docs: `docs/agents/testing.md` (selector section + gate-manifest section) updated; skills untouched.

Touched: 14 files (4 new). Scope stayed in `scripts/check-affected`, `scripts/gate`, `.github/workflows/ios.yml`, testing docs.

### DIFF
--- a/.github/workflows/ios.yml
+++ b/.github/workflows/ios.yml
@@ -16,6 +16,36 @@ on:
       - '.github/workflows/deploy.yml'
       - '.github/workflows/pr-preview.yml'
       - '.github/workflows/pr-preview-cleanup.yml'
+      # Routing (#1781 A9-2). Everything below is derived from the affected selector's
+      # device-lane ownership (scripts/check-affected/device-lanes.ts): unit tests no device
+      # lane runs, and trees another platform family owns. `pnpm check:gate-manifest` asserts
+      # this list against the selector over every tracked path, both ways — a path the selector
+      # routes to the iOS lane (or fails open on: tooling, workflows, package manifests) must not
+      # appear here, and a path it classifies as another family's or as a unit test must. The
+      # `push` trigger has no filter: main runs every lane unconditionally.
+      - '**/*.md'
+      - 'src/**/*.test.ts'
+      - 'src/**/__tests__/**/*.ts'
+      - 'packages/*/src/**/*.test.ts'
+      - 'packages/*/src/**/__tests__/**/*.ts'
+      - 'src/platforms/android/**'
+      - 'src/platforms/harmonyos/**'
+      - 'src/platforms/linux/**'
+      - 'src/platforms/vega/**'
+      - 'src/platforms/web/**'
+      - 'packages/platform-android/src/**'
+      - 'packages/platform-harmonyos/src/**'
+      - 'packages/platform-linux/src/**'
+      - 'packages/platform-vega/src/**'
+      - 'packages/platform-web/src/**'
+      - 'android/**'
+      - 'linux/**'
+      - 'test/integration/replays/android/**'
+      - 'test/integration/replays/linux/**'
+      - 'test/integration/replays/macos/**'
+      - 'test/integration/android-**'
+      - 'test/integration/smoke-android-**'
+      - 'test/integration/smoke-web-**'
   push:
     branches:
       - main
--- a/docs/agents/testing.md
+++ b/docs/agents/testing.md
@@ -305,6 +305,13 @@ device-lane surface or as a unit test must not (`scripts/gate/routing.ts`). GitH
 `paths-ignore` before a runner is allocated, so this is routing with no job on the critical
 path; the assertion is what keeps the hand-written glob list a derived artifact.
 
+Two limits of the mechanism, both inherent to `paths-ignore` rather than to the assertion:
+GitHub's path filters examine only the **first 300 changed files**, so a PR larger than that can
+skip a routed lane on the strength of its first 300 paths alone (`push` to `main` has no filter
+and is the backstop); and a lane may name a *sibling* workflow file exactly to say it does not
+use it, but never a file in its own `uses:` closure — the composite actions its steps run, plus
+its own definition — which the assertion refuses.
+
 ## Mutation report over decision kernels
 
 Mutation score is the mechanical answer to "is this test load-bearing or decorative". A full-suite
--- a/docs/agents/testing.md
+++ b/docs/agents/testing.md
@@ -195,7 +195,19 @@ hand-maintained path map:
   script fails loudly instead of dropping a gate.
 - A **small explicit build-ownership layer** covers the paths whose owning build
   cannot be derived: Swift runner, Android helpers, macOS helper, MCP metadata,
-  and the public package surface (itself derived from `package.json` `exports`).
+  the TS/Swift golden tables (`contracts/fixtures/`), and the public package
+  surface (itself derived from `package.json` `exports`).
+- **Device lanes** (`replay-ios`, `replay-ios-device`, `replay-macos`,
+  `replay-android`, `replay-linux`, `web-smoke`) are owned by platform family
+  (`scripts/check-affected/device-lanes.ts`): a path under a family-tagged tree
+  (`packages/platform-<family>/`, `src/platforms/<family>/`, `android/`,
+  `test/integration/replays/<leaf>/`, the lane-prefixed `test/integration/`
+  smoke files) owns that family's lanes; untagged runtime surface owns every
+  lane; unit tests under `src/` and `packages/*/src/` own none. The tags are
+  directory-level only — `src/daemon/android-system-dialog.ts` is a naming
+  convention, not a boundary, and stays shared. `ios.yml`'s `pull_request`
+  `paths-ignore` is routed on this ownership and held to it both ways by the
+  gate manifest (below); `push` to main runs every lane unconditionally.
 
 Changed-file discovery folds working-tree state into the local plan: in the
 default local mode (`--head HEAD`) it unions the committed `base..HEAD` diff with
@@ -284,9 +296,14 @@ step executes on every run. Raw shell can still run project code, but it cannot 
 manifest. The check proves the smaller structural claim that every registered gate has an explicit
 CI owner and every affected path can reach one.
 
-The three facts the manifest cannot derive live together in `scripts/gate/declarations.ts`: one
-coverage wrapper, one reporting-only `test:*` script, and the Android replay owner hidden inside a
-third-party action's `script:` input.
+The facts the manifest cannot derive live together in `scripts/gate/declarations.ts`: opaque
+runners, reporting-only `test:*` scripts, unprovable and manual-only owners, and the **routed
+lanes** — a `pull_request` lane whose `paths-ignore` list is asserted against the selector over
+every tracked path, both ways: a path the selector fails open on or routes to one of the lane's
+declared or sampled checks must start the lane, and a path it classifies as another family's
+device-lane surface or as a unit test must not (`scripts/gate/routing.ts`). GitHub evaluates
+`paths-ignore` before a runner is allocated, so this is routing with no job on the critical
+path; the assertion is what keeps the hand-written glob list a derived artifact.
 
 ## Mutation report over decision kernels
 
--- a/package.json
+++ b/package.json
@@ -130,7 +130,7 @@
     "fallow:baseline": "fallow dead-code --save-baseline fallow-baselines/dead-code.json --summary && fallow health --report-only --save-baseline fallow-baselines/health.json --summary",
     "check:fallow": "fallow audit",
     "check:affected": "node --experimental-strip-types scripts/check-affected/run.ts",
-    "check:affected:test": "node --experimental-strip-types scripts/node-test-tmpdir.ts --experimental-strip-types --test scripts/check-affected/model.test.ts scripts/check-affected/platform-packages.test.ts scripts/check-affected/run.test.ts",
+    "check:affected:test": "node --experimental-strip-types scripts/node-test-tmpdir.ts --experimental-strip-types --test scripts/check-affected/model.test.ts scripts/check-affected/platform-packages.test.ts scripts/check-affected/device-lanes.test.ts scripts/check-affected/run.test.ts",
     "gate": "node --experimental-strip-types scripts/gate/run.ts",
     "check:gate-manifest": "node --experimental-strip-types scripts/gate/check.ts",
     "check:gate-manifest:test": "node --experimental-strip-types scripts/node-test-tmpdir.ts --experimental-strip-types --test scripts/gate/*.test.ts",
--- a/scripts/check-affected/device-lanes.test.ts
+++ b/scripts/check-affected/device-lanes.test.ts
@@ -1,0 +1,178 @@
+// Device-lane ownership (#1781 A9-2): planted paths per family, and the two properties the
+// ios.yml routing leans on — an Apple-owned change carries its iOS checks in a narrow plan, and
+// a tooling change beside it still fails open to the full set.
+
+import assert from 'node:assert/strict';
+import { test } from 'node:test';
+import { deviceLaneLeaf, deviceLanesFor, isDeviceLaneSurface, isUnitTest } from './device-lanes.ts';
+import { selectChecks, type CheckId } from './model.ts';
+
+const IOS: readonly CheckId[] = ['replay-ios', 'replay-ios-device'];
+
+function lanes(file: string): CheckId[] {
+  const plan = selectChecks({ changedFiles: [file] });
+  assert.equal(plan.failOpen, false, `${file} must not fail open`);
+  return plan.checks.filter((id) =>
+    /^replay-|^web-smoke$|^swift-runner-|^macos-helper$|^android-helpers$/.test(id),
+  );
+}
+
+test('a TypeScript-only Apple change selects the iOS and macOS lanes without a Swift build', () => {
+  for (const file of [
+    'src/platforms/apple/core/apps.ts',
+    'packages/platform-apple/src/runtime.ts',
+    'src/platforms/apple/os/macos/desktop.ts',
+  ]) {
+    assert.deepEqual(lanes(file), ['replay-ios', 'replay-ios-device', 'replay-macos'], file);
+    assert.ok(
+      selectChecks({ changedFiles: [file] }).reasons.some(
+        (reason) => reason.rule === 'own:device-lane:apple',
+      ),
+    );
+  }
+  assert.deepEqual(lanes('src/daemon/snapshot-presentation/ios/action-shelf.ts'), [...IOS]);
+  assert.deepEqual(lanes('test/integration/replays/macos/01-desktop.ad'), ['replay-macos']);
+});
+
+test('a Swift runner change selects both builds and every Apple lane', () => {
+  assert.deepEqual(lanes('apple/runner/AgentDeviceRunner/Sources/Foo.swift'), [
+    'swift-runner-ios',
+    'swift-runner-macos',
+    'replay-ios',
+    'replay-ios-device',
+    'replay-macos',
+  ]);
+});
+
+test('an iOS replay script and the iOS smoke files own the iOS lanes', () => {
+  for (const file of [
+    'test/integration/replays/ios/simulator/01-settings.ad',
+    'test/integration/replays/ios/fixture/02-checkout-release.ad',
+    'test/integration/smoke-ios-simulator.test.ts',
+    'test/integration/ios-simulator-e2e/live-runner.ts',
+    'test/integration/ios-simulator-e2e-cleanup.test.ts',
+  ]) {
+    assert.deepEqual(lanes(file), [...IOS], file);
+  }
+});
+
+test('shared runtime surface owns every device lane', () => {
+  for (const file of [
+    'src/daemon/handlers/session.ts',
+    'packages/kernel/src/errors.ts',
+    'src/daemon/android-system-dialog.ts', // naming convention in a shared dir, not a boundary
+    'test/integration/smoke-daemon-clean.test.ts',
+    'src/platforms/install-source.ts',
+  ]) {
+    assert.equal(deviceLaneLeaf(file), 'shared', file);
+    assert.deepEqual(lanes(file), [
+      'web-smoke',
+      'replay-ios',
+      'replay-ios-device',
+      'replay-macos',
+      'replay-linux',
+      'replay-android',
+    ]);
+  }
+});
+
+test('another family owns only its own lanes, so an Android-only change carries no iOS check', () => {
+  assert.deepEqual(lanes('src/platforms/android/perf.ts'), ['replay-android']);
+  assert.deepEqual(lanes('packages/platform-android/src/inventory.ts'), ['replay-android']);
+  assert.deepEqual(lanes('android/snapshot-helper/src/main/java/X.java'), [
+    'android-helpers',
+    'replay-android',
+  ]);
+  assert.deepEqual(lanes('test/integration/replays/android/emulator/01-settings.ad'), [
+    'replay-android',
+  ]);
+  assert.deepEqual(lanes('test/integration/smoke-android-emulator.test.ts'), ['replay-android']);
+  assert.deepEqual(lanes('test/integration/android-emulator-e2e/live-runner.ts'), [
+    'replay-android',
+  ]);
+  assert.deepEqual(lanes('src/platforms/linux/snapshot.ts'), ['replay-linux']);
+  assert.deepEqual(lanes('linux/atspi-dump.py'), ['replay-linux']);
+  assert.deepEqual(lanes('src/platforms/web/provider.ts'), ['web-smoke']);
+  assert.deepEqual(lanes('test/integration/smoke-web-platform.test.ts'), ['web-smoke']);
+  // Families with no CI lane fall through to the static gates only.
+  assert.deepEqual(lanes('src/platforms/harmonyos/hdc.ts'), []);
+  assert.deepEqual(lanes('packages/platform-vega/src/index.ts'), []);
+});
+
+test('the fixture app owns the mobile lanes whichever subtree changes', () => {
+  assert.deepEqual(lanes('examples/test-app/app/index.tsx'), [
+    'replay-ios',
+    'replay-ios-device',
+    'replay-android',
+  ]);
+  assert.equal(deviceLaneLeaf('examples/test-app/modules/lab/android/build.gradle'), 'fixture-app');
+});
+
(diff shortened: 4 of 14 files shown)


thymikee:
Not ready at exact `4992259e`.

The body claims the dependency-closure fix is in `7398ded24`, but that commit is not pushed or reachable: both the source branch and `refs/pull/1857/head` still point to `4992259e`. The claimed fix is therefore not reviewable or checked.

At the actual head, the routing P1 remains: any ignored `.github/**` file named exactly is exempted without proving that `ios.yml` does not directly or transitively use it, and the current test explicitly blesses an arbitrary workflow exemption. A lane-owned composite-action change can still bypass the manifest.

The Coverage failure is inherited main-side ratchet skew tracked by #1860, not caused by this patch. The body is also stale: it describes the unpushed fix and says 14 touched files while the exact PR contains 13.

thymikee:
Re-reviewed GitHub exact head `4992259e`: unchanged and still not ready. The claimed dependency-closure fix `7398ded24` is still not pushed or reachable from any fetched remote branch, so it is neither reviewable nor covered by authoritative checks. At the actual head, exact-name `.github/**` exemptions are still not bounded by `ios.yml`’s direct/transitive local-action dependency closure, and the test still blesses an arbitrary workflow exemption; a lane-owned composite-action change can bypass the routing manifest.

The body is materially stale: it describes the unpushed fix and says 14 touched files, while GitHub reports 13. The PR base `b12a3e3c` is also three commits behind current main `fda81c512`, so the old checks are not current-base readiness evidence. The old-head Coverage failure is inherited pre-#1860 ratchet skew; other old-head lanes were green. Push the claimed fix, rebase, and rerun exact-head gates before substantive re-review.

thymikee:
The re-review above read head `4992259e`, which was correct at the time and is now three pushes stale. Current head is **`354126c35`**, rebased onto current `main` (`6984a1e09`, #1852 included) — base is no longer behind.

What is now at the head, all of it reviewable:

- **Dependency-closure fix pushed** (`354126c35`, was `7398ded24` pre-rebase). `Lane` carries `uses`: the transitive local composite-action closure plus the workflow's own file, built by the same walk `declaredGates` already performs (`scripts/gate/workflows.ts`). `exemptSiblingWorkflow` refuses any exact-name ignore that names a file in it, so a lane-owned composite-action change can no longer bypass the routing manifest. The reviewer's exact plants — `.github/actions/setup-apple-runner-build/action.yml` and `.github/actions/boot-ios-test-simulator/action.yml` — are pinned in `scripts/gate/routing.test.ts`, together with `run-gate/action.yml` (reachable only *through* a composite action) and `ios.yml` itself. The earlier test that "blessed an arbitrary workflow exemption" now only blesses a **sibling** workflow (`size.yml`), which is the claim the exemption actually makes.
- **Both new cases proven red against the pre-fix code**: reverting the two guards fails exactly those two tests (`naming .github/actions/setup-apple-runner-build/action.yml exactly must still fail the routing assertion`), so they are not vacuous.
- **Second finding fixed**: an unowned non-TS path under an ignored family root (`src/platforms/android/probe-fixture.json`) no longer says "Remove the ignore entry" — removing `src/platforms/android/**` would un-route every sibling `.ts` — it asks for a selector owner instead. Also pinned, also red pre-fix.
- **Third finding documented**: GitHub's path filters examine only the first 300 changed files, so a large PR can skip the lane on its first 300 paths alone. In `docs/agents/testing.md` and the PR body's residual-risk section; `push` to `main` stays unfiltered as the backstop.
- **Touched-file count corrected and verified against the pushed tree**: `git diff --name-status origin/main..HEAD` reports **14 files, 4 new** (`device-lanes.ts`, `device-lanes.test.ts`, `routing.ts`, `routing.test.ts`), which is what the body now says. The earlier "13" was GitHub counting the pre-fix head.

Local evidence on the rebased tree: `pnpm check:affected --run` green end-to-end, `pnpm check:gate-manifest` green (`46 checks wired across 33 lanes, manual-only: replay-android, replay-ios, replay-ios-device`), routing + device-lane suites 21/21.

Note for whoever merges — this is ratchet-class, as the body says: the routing assertion runs over every tracked path, so a path added on `main` after this base can turn `check:gate-manifest` red on `main` post-merge (#1825 × #1779). Re-run `pnpm check:gate-manifest` against current `main` immediately before merging.

thymikee:
Re-reviewed exact `354126c35e9b6bd2cf94108f6b719257efebdf2d`: the prior `.github` exemption bypass is fixed. Lane ownership now includes the workflow plus transitive local composite-action descriptors, and focused tests non-vacuously pin `ios.yml`, direct actions, and the transitive `run-gate` action. No code finding remains.

Before readiness: update the stale body reference from unreachable `7398ded24` to the current fix/head, await all exact-head CI green, and perform the promised immediate pre-merge `pnpm check:gate-manifest` rerun against then-current main. PR remains draft.

thymikee:
**Out of draft.** Head `354126c35` on base `6984a1e09` (current `main`, #1852 included): **30/30 checks pass, 0 failures** — CI, iOS, macOS, Android, Linux, Size, Test App Build Cache, Deploy PR previews all green on this exact head (CodeQL skipped as usual for this path set).

The iOS lane **started** on this PR rather than being skipped — that is the fail-open direction proving itself in production: the diff touches `.github/workflows/ios.yml`, `scripts/check-affected/**`, `scripts/gate/**`, `package.json` and `docs/agents/testing.md`, every one of which the selector fails open on, so routing must not skip it. Linux was cancelled twice earlier at the job's 30-minute timeout inside `sudo apt-get install` — a repo-wide condition that also cancelled `main`'s own runs at 08:38, 08:46, 09:06 and 09:08 today — and passed on the rebased head.

Local evidence on this tree: `pnpm check:affected --run` green end to end, `pnpm check:gate-manifest` green (`46 checks wired across 33 lanes, manual-only: replay-android, replay-ios, replay-ios-device`), routing + device-lane suites 21/21, both review-round guards proven red against their pre-fix code.

**Merge note (ratchet-class, repeated from the body):** the routing assertion runs over every tracked path, so a path landing on `main` after this base can turn `check:gate-manifest` red post-merge (#1825 × #1779). Re-run `pnpm check:gate-manifest` against current `main` immediately before merging.

thymikee:
**Pre-merge rebase done — the ratchet condition this PR's body warns about arrived for real.**

Head is now `a60067383`, rebased onto `d07b83762` (current `main`). Five commits landed under the old base: #1861 (`d07b83762`, renamed ci.yml's `swift-runner-unit-compile` / "Swift Runner Unit Compile" → `swift-runner-host-tests` / "Swift Runner Host XCTests"), #1854, #1853, #1872 (layering: relative imports inside workspace packages), #1867.

**Did the rename require a change here? No — and confirmed rather than assumed.** The rebase was clean because this branch never touches `ci.yml` (`git diff --name-only origin/main...HEAD` has zero `ci.yml` entries), so main's new job name came in verbatim. The routing rules carry no job names at all: `device-lanes.ts` routes by directory tag, and `routing.ts` resolves lanes by the label the loader derives from the YAML, so a renamed job is picked up automatically. `owningLanes` on the rebased tree now reports `swift-runner-macos -> Swift Runner Host XCTests | macOS / Smoke Tests` — the manifest tracked the rename with no edit.

**Does the renamed host-XCTest job need an ownership rule?** No. It declares exactly one gate, `swift-runner-macos`, through `setup-apple-runner-build`; its XCTest step is raw `run:` shell, which by design declares no ownership, and #1861 added no new `CheckId` (the universe is unchanged at 46 wired). It lives in `ci.yml`, whose `paths-ignore` is docs-only and which this PR does not touch or route — so every non-docs PR, including a TypeScript-only Apple change, already starts it. Adding a device-lane rule would claim ownership of a check the lane already runs unconditionally.

**Re-run against main at `d07b83762`:** `pnpm check:gate-manifest` green (`46 checks wired across 33 lanes, manual-only: replay-android, replay-ios, replay-ios-device`) — it had nothing to say about the renamed job or the new lane. `pnpm check:affected --run` green end to end on the rebased head; routing + device-lane suites 21/21.

The merge condition in the body still stands for whoever merges: re-run `pnpm check:gate-manifest` against `main` immediately before merging, since the assertion reads every tracked path and every workflow.

thymikee:
Re-reviewed exact `a60067383`. One routing bypass remains: `Lane.uses` records only transitive `action.yml` descriptors, not files those composite actions execute. `ios.yml` uses `setup-fixture-app`, which runs `$GITHUB_ACTION_PATH/fetch-artifact.sh` and sibling scripts, but those files are absent from `lane.uses`; adding that exact path to `paths-ignore` is accepted as a sibling exemption and the manifest stays green even though the file is iOS-lane machinery. Extend the closure to action support files or narrow the exemption, and add a non-vacuous planted witness. Also replace stale `7398ded24` in the body with current `a60067383`. Exact-head CI is green; retain the promised immediate pre-merge gate-manifest rerun. No `ready-for-human` yet.

thymikee:
**Second bypass closed.** Head is `17a762f27` on base `67ce19b50` (current `main` — #1880, #1879, #1842 came in on the way).

**Which fix, and why that one.** I extended the closure to every file under a composite action's directory, rather than narrowing the exemption or parsing `$GITHUB_ACTION_PATH/...` out of the YAML. The reason is in the reference chain you pointed at: `setup-fixture-app/action.yml` runs `fetch-artifact.sh`, and *that script* runs `resolve-artifact-name.sh` and `trusted-artifact.mjs` — references that appear nowhere in YAML. A fix that read `$GITHUB_ACTION_PATH` out of the descriptor would have caught `fetch-artifact.sh` and missed the other two: the same one-level-short failure, a third time. The directory is the smallest unit that terminates without modelling shell and cannot miss a file however deep the chain runs. Its coarseness costs nothing real — a file inside an action's own directory belongs to that action, so refusing to exempt it is the correct answer, not an approximation of one. `lane.uses` for `iOS / Smoke Tests` now carries 13 entries: the workflow, seven `action.yml` descriptors, and the five `setup-fixture-app` support files.

**Planted witness** (`.github/actions/setup-fixture-app/fetch-artifact.sh` added to `paths-ignore`, verbatim):

```
### PLANTED: fetch-artifact.sh added to ios.yml paths-ignore
Routed lanes whose paths-ignore disagrees with the selector:
  - ios.yml ignores .github/actions/setup-fixture-app/fetch-artifact.sh, but the selector fails
    open on it (workflow-tooling). Remove the ignore entry.

gate manifest: 1 failure(s).
### RESTORED
gate manifest: ok — 46 checks wired across 33 lanes, manual-only: replay-android, replay-ios, replay-ios-device.
```

Pinned as a test over all three files — the descriptor-named one and the two named only inside the shell script — and the test is non-vacuous: reverting the closure to descriptors-only fails exactly it (`✖ the exemption cannot name a support file of an action the lane runs`, 10 pass / 1 fail), and passes at 11/11 restored.

**Is anything else path-referenced?** I checked the three shapes you named, and the answer is no, for a structural reason worth stating: the exemption only ever applies to `.github/**` paths, so that is the entire risk surface. Within it — every local `uses:` across all workflows and actions resolves to a directory that has an `action.yml` (no bare-path `uses:`); the only non-`action.yml` files under `.github/actions/**` are the five `setup-fixture-app` scripts, now all in the closure; and the rest of `.github/` is `FUNDING.yml` and `renovate.json`, which no lane reaches. Files referenced by path from `run:` blocks *outside* `.github/` (e.g. `scripts/**`) were never exemptible in the first place — the selector fails open on them and the assertion demands the lane start, which is what the `.github/actions/**` and `package.json` plants in `routing.test.ts` already pin.

**Merge condition, corrected.** The body no longer names a fixed base: it now says to re-run `pnpm check:gate-manifest` against whatever `main` is at merge time, since the assertion reads every tracked path *and* every workflow, and this branch has already been overtaken three times today. The stale `7398ded24` reference is gone — the body cites the two commits by title instead, which survive rebases.

Local on this head: `pnpm check:affected --run` exit 0, `check:gate-manifest` green, gate-manifest suite 49/49, affected-selector suite 47/47, fallow clean.

thymikee:
Independent re-review of `17a762f`: clean. The action-directory closure now covers the descriptor plus every support file of direct and transitive local actions, and the routing gate refuses any exact-name ignore in that closure. The three support-file witnesses would fail if either the closure or exemption guard regressed. Exact-head CI is green, including iOS; Size is +44 B unpacked. Merge-ready. If main moves before merge, rerun `pnpm check:gate-manifest` against then-current main as documented.

Issue #1788, opened 2026-08-17
daemon: events.ndjson has no size/rotation cap


thymikee:
`src/daemon/session-event-log.ts` writes `events.ndjson` per session with pure `appendFile`, no size or rotation cap — `MAX_EVENT_LIMIT = 500` is only a page-size cap for `readSessionEventLog`, not a file bound. A long-lived session grows this file without bound (unlike `src/utils/app-log-files.ts`, which already caps app logs at `DEFAULT_MAX_APP_LOG_BYTES = 5MB` / `DEFAULT_MAX_ROTATED_FILES = 1`, env-overridable via `AGENT_DEVICE_APP_LOG_MAX_BYTES` / `AGENT_DEVICE_APP_LOG_MAX_FILES`).

Naively reusing the app-log rotate-then-truncate approach is **not safe here**: the sole reader, the `events` CLI/RPC command (`src/commands/observability/index.ts`, backed by `readSessionEventLog`), pages with a `cursor` that is an **absolute line-offset since the file's creation** (`normalizeCursor`/`consumeEventLogLine` in `session-event-log.ts`), which callers are expected to persist and resume across repeated calls over a session's lifetime — exactly the long-lived-session scenario motivating a cap. Truncating the live file on rotation resets line indices to 0, so a resumed cursor silently reads the wrong (or no) events instead of erroring — a correctness regression, not just data loss.

The good news: no other subsystem depends on `events.ndjson` completeness. Replay/heal use `session.actions`, not the event journal; ADR-0018 (`docs/adr/0018-unified-event-journal.md`, Proposed/unimplemented) explicitly states the journal is "not a source of truth" and rejects rebuilding state from it. So a cap is compatible with documented guarantees — but making it *correct* for the cursor-paging reader needs design work beyond a copy-paste of the app-log rotation, e.g. a stable per-entry sequence number that survives rotation (content-addressed cursor instead of position-addressed) plus an explicit "requested cursor precedes retained history" signal instead of a silent empty/wrong page.

Filed per #1781 (B5). See also the ADR reference above for the "no state from events" invariant that makes a cap acceptable in principle.

thymikee:
[...]

Issue #1105, opened 2026-07-04
Runner resilience under agent-abuse command patterns


thymikee:
Found by the first headless-Haiku benchmark run (#1101): a small-model agent hammering the CLI in tight act/observe loops (38 turns in ~6.7 min, including malformed and rapid-fire commands) wedged the iOS runner hard enough that snapshot requests kept timing out even across a subsequent `open --relaunch`. The daemon correctly survived (preserve-on-snapshot-timeout policy from #1075) but the runner never self-recovered — it needed a process kill.

This is a new workload class: adversarial-by-incompetence. Strong models pace themselves; small models retry fast, interleave commands mid-flight, and abandon sessions mid-request. Best-in-class agent reliability means the runner recovers from this without human intervention.

Investigation starters:
- Reproduce with a scripted rapid-fire pattern (the benchmark harness in #1101 can be adapted); capture the runner state when wedged (last diag phases, whether the AX-unavailable invalidation fired and got stuck re-acquiring).
- The `--relaunch` path should arguably detect a runner whose last N commands all timed out and recycle it proactively (kill + fresh start) instead of reusing the session — the keep-hot machinery has the health signals (readiness preflight, invalidate/restart-and-replay) but apparently a state exists where none of them trigger recycling.
- Consider a daemon-side circuit breaker: M consecutive runner command timeouts → forced runner restart on next command, with a diagnostic explaining why.

thymikee:
Update: this is worse than agent-abuse — it reproduces deterministically with slow, paced, plain commands (no --settle, no rapid fire). Repro tonight on the Bluesky dev build (iPhone 17 Pro sim):

1. Search → fill "alpenglow" → press the first account result (profile screen opens)
2. `snapshot -i` on the fully-rendered profile: **9.3s** for 133 nodes (normal screens: 400–700ms)
3. next `press "label=Notifications"`: **92s wall → COMMAND_FAILED "Daemon request timed out"**
4. afterwards every command fails `SESSION_NOT_FOUND` until a full re-open

[...]

PR #758, opened 2026-06-11
fix: add iOS private AX snapshot fallback

thymikee:
## Summary

Adds a simulator-only private AX snapshot fallback for iOS when XCTest returns a sparse application/window tree or fails while serializing AX snapshots.

The fallback uses XCTest private accessibility interfaces dynamically, maps the recovered tree back into existing SnapshotNode output, and keeps normal XCTest snapshots authoritative whenever they return real content. It also adds a conservative public XCTest-query recovery tier for regular sparse snapshots before falling back to private AX; compact interactive snapshots still use the private-AX/find recovery path because Bluesky showed public queries collapse there too.

Also fixes SpringBoard permission alerts in compact interactive snapshots: modal detection now runs before the compact app-tree shortcut, and permission sheets return alert text plus actionable button refs without broad SpringBoard subtree walks.

Hardens mutating find actions on iOS: when compact interactive snapshots collapse to the application root, find retries with a full snapshot, then with a query-scoped full snapshot if unscoped AX serialization fails on unrelated feed content.

Closes #701
Closes #761

## Validation

- pnpm build
- pnpm check:quick
- pnpm check:unit outside sandbox after sandbox listener failures
- pnpm build:xcuitest
- pnpm exec vitest run src/daemon/handlers/__tests__/find.test.ts
- Compared callstack/agent-device#761 locally on Bluesky Home: its public-query recovery did not fix `snapshot -i -c`, did not expose the permission alert, and `find Search click` failed there.
- Prototyped #761-style public query recovery on this branch. Kept it only for regular sparse snapshots; compact public-query recovery and a direct KVC AX-object bridge were tested on Bluesky and dropped because they still returned only the application root.
- Manual iOS simulator smoke: opened Settings on iPhone 17 iOS 26.2 and verified snapshot -i -c returned usable refs with the rebuilt runner.
- Manual Bluesky iOS 27 simulator smoke with Xcode 26.2: opened ~/Developer/Bluesky, logged in, reached Home/Discover, verified full Home snapshot exposed 133 nodes including homeScreen, composeFAB, feed tabs, bottom-bar IDs, and post content.
- Manual Bluesky iOS 27 find smoke after simulator reboot: compact Home snapshot remained sparse by design, but find Search click succeeded through the fallback and navigated to Explore/Search; find Search exists also passed.
[...]

PR #1858, opened 2026-08-18
test: make hard-coded caps overridable behind seams (#1781 B5)

thymikee:
## Summary

Wave-3 item B5 (#1781): make hard-coded caps reachable from tests where hitting the cap has a real failure mode, and say which caps were surveyed and left alone.

Two caps get coverage; the rest of the survey is tabled below so the next reader does not redo it.

**Result worth carrying to the rest of the wave:** B5 was framed as "make caps overridable so tests can reach them", and for this cap that framing turned out to be the wrong answer. An option on the production snapshot interface that no caller ever sets is a test affordance wearing production clothes — the repo's rule is to make the invariant testable at its *owning* interface instead of widening a public one to reach it. So the traversal/emission policy was extracted into a pure function that takes its bounds explicitly, and `snapshotHarmony` kept its `SnapshotOptions` shape. "Extract the policy" beats "add an override" wherever the cap governs a decision rather than a resource.

- **HarmonyOS snapshot node cap** (`src/platforms/harmonyos/snapshot.ts`, `MAX_NODES = 5_000`): reaching the cap exposed a second, real bug that the seam made visible — the walk `return`ed as soon as the cap filled, so an omitted node's **descendants reached neither `rawNodeCount` nor `maxDepth`**, and `analysis` under-reported exactly the oversized trees the cap exists for. Emission is now capped while accounting continues, which is what makes the whole-tree claim mirrored from the Android helper true. The policy now lives in `collectArkUiNodes(root, { maxNodes, maxDepth, interactiveOnly })` — a pure function with no defaults of its own, so the limit and the accounting rule are exercised directly at the interface that owns them. `buildHarmonySnapshot` supplies `MAX_NODES` and the caller's depth/interactive options; **`snapshotHarmony`'s signature is unchanged** (`SnapshotOptions`), and no option was added anywhere. No CLI/MCP surface changes.
- **Durable descriptor JSON node cap** (`packages/capture-kit/src/durable-json.ts`, `MAX_DESCRIPTOR_JSON_NODES = 4_096`): reachable in memory without a seam, so it gets a boundary test instead. Every node counts — root object, `items` array, each leaf — so the fixtures sit **exactly** on the cap: 4,094 leaves = 4,096 nodes accepted, 4,095 leaves = 4,097 nodes rejected. The count is enforced at **two** sites (`validateObject`, `validateArray`) and only whichever owns the offending node rejects, so both are pinned with an object-leaf and an array-leaf document.

### Survey — what has a real failure mode, what already has a seam, what was left

| Cap | On hit | Already testable / seamed? | Decision |
|---|---|---|---|
| HarmonyOS `MAX_NODES` 5,000 | drops nodes, sets `truncated`; **and under-reported `analysis` below the cap** | no seam, untested | **policy extracted + tested** (this PR); no production option added |
| capture-kit `MAX_DESCRIPTOR_JSON_NODES` 4,096 | envelope rejected (`false`) | reachable in memory, untested | **tested** (this PR), no seam needed |
| capture-kit `MAX_DESCRIPTOR_JSON_DEPTH` 32 | rejected | tested (depth 40) | left |
| `@agent-device/xml` `MAX_XML_NESTING_DEPTH` 256 / `maxDocumentChars` | throws | tested at 257 / seam + tested | left (plain `Error` there is #1792's class) |
| `MAX_OVERLAY_REFS` 24 (`screenshot-overlay.ts`) | keeps top-24 ranked refs; full snapshot still has all refs | seam exists (`maxRefs`) | left — ranked selection by design, no data loss |
| Runner connect retry (`runner-startup-transport.ts`) | typed connect error / simulator fallback | attempts derived from `timeoutMs`; exhaustion tested with `timeoutMs=100` | **not done** — reachable today (`retryWithPolicy` already takes `maxAttempts`/`baseDelayMs`/`jitter`, and 39 suites use `vi.useFakeTimers()`), just not carried here: the exhaustion outcome is already a typed error with its own coverage. Not a CI constraint |
| `IOS_SIMULATOR_SCREENSHOT_RETRY_MAX_ATTEMPTS` 5 | last simctl error → runner fallback | test mocks `retryWithPolicy` | **not done**, same reason |
| `PREPARE_RUNNER_HEALTH_MAX_SESSION_ATTEMPTS` 2, `ANDROID_KEYBOARD_DISMISS_MAX_ATTEMPTS` 2, `HELPER_CONTENT_CAPTURE_ATTEMPTS` 3, `MAX_UPLOAD_REDIRECTS` 5 | typed `AppError` | exhaustion already tested | left |
| `OUTCOME_RETRY_ATTEMPTS` 2, Android freshness / divergence retry delays, `MAX_REPLAY_TEST_RETRIES` clamp | typed give-up state returned (`retried:false`, `staleAfterRetries`, `unavailable`) | partially tested | left — outcome is a typed field, not silent |
| `IOS_DEVICE_TRACE_RECORD_MAX_ATTEMPTS` 3 (declared twice: `perf.ts`, `perf-xctrace.ts`) | typed error / last failed attempt | one-retry-then-success tested; 1.5 s real sleep per retry | left; the duplicate declaration is a one-site cleanup for a perf-owner PR |
| `LEDGER_MAX_ENTRIES` 512 (runner recycle ledger) | silent oldest-first eviction, no diagnostic — an evicted entry resets the `MAX_RUNNER_RECYCLES_PER_REQUEST = 1` guard that exists because of the #1105 wedge | untested; `resetRunnerRecycleLedgerForTests` already exists | left — a real but **unlikely** failure mode (needs 512 newer request keys inside one request's lifetime), not an absent one |
| `MAX_REF_PINS_PER_SCOPE` 1,000, `MAX_CRASH_ARTIFACT_BYTES` 64 MB, `MAX_HTTP_RPC_BODY_BYTES` 1 MB, app-event/push 8 KB payload caps | eviction / typed error / socket destroy | tested | left |
| `NETWORK_MAX_SCAN_LINES` 4,000 (parser layer has `maxScanLines`) | typed error / reason string / silent tail-drop | untested; the first two need a fake stream or 64 MB | left, listed for the owner |
| `events.ndjson` (no cap at all) | unbounded growth | — | separate PR (#1788) |

No central limits/config module exists; each cap is declared once at its use site and the two numeric env overrides in the tree (`AGENT_DEVICE_APP_LOG_MAX_BYTES`, `AGENT_DEVICE_DAEMON_IDLE_TIMEOUT_MS`) each live beside the limit they override. This PR follows that: no new `process.env` reads.

## Validation

Both new tests were run red against the pre-change code before going green:

- HarmonyOS: with `const maxNodes = MAX_NODES` restored, `snapshotHarmony reports truncation once the node cap is hit …` fails with `AssertionError: Expected values to be strictly equal: + actual - expected / + undefined / - true` (the cap is never reached, so `truncated` stays unset).
- HarmonyOS accounting, re-proven after the extraction moved the test down to `collectArkUiNodes`: restoring the early `return` at the limit → `× collectArkUiNodes keeps counting the tree below a node the limit omitted`, `+ maxDepth: 1, rawNodeCount: 3` against `- maxDepth: 3, rawNodeCount: 5`.
- Durable JSON, boundary pinned in both directions — moving `MAX_DESCRIPTOR_JSON_NODES` by one either way reddens it, so the fixtures are on the cap and not merely past it:
  - cap `4_097` → `- false / + true` (the over-cap document is wrongly accepted);
  - cap `4_095` → `- true / + false` (the on-cap document is wrongly rejected).
- Durable JSON, each counting site load-bearing (the first draft misattributed this proof to `validateArray`; the object-leaf case rejects in `validateObject`, and the array branch was uncovered until the second document was added): deleting the conjunct at `durable-json.ts:31` reddens it, and so does deleting the one at `:57`.

`pnpm check:affected --run` green locally (fallow, layering, coverage over the related set: 241 files / 1,632 tests). Provider-integration and coverage lanes are CI-owned for `src/platforms/**`.

- Catches: a snapshot backend silently dropping nodes at its cap without the `truncated` signal, and a durable-descriptor guard that stops bounding node count — both unobservable before because nothing reached either cap.
- Evidence: #758's `maxDepth: 64` was a hard cap nobody could exercise below the real tree; the private-AX ladder only landed after on-device failure. Both caps here had zero tests at the boundary (survey above).
- Cost: five unit tests, ~5 ms; no runner occupancy, no red-lane attention, ~140 lines of test, ~35 lines of production change (an extraction, not an interface widening).
- Kill-criterion: none needed — the seams are the same shape the sibling platforms already carry and the tests are boundary unit tests.

### Residual readiness risk: no HarmonyOS device evidence

The HarmonyOS change is covered only by unit tests that mock `runHarmonyHdc`, which is what `docs/agents/testing.md` requires of CI (GitHub has no DevEco/HDC/emulator). **No live HarmonyOS device or emulator run backs the changed snapshot path**, and I have no HarmonyOS hardware. The accounting change alters `analysis.rawNodeCount`/`maxDepth` on real oversized trees, so someone with a device should confirm on a screen large enough to exceed the default 5,000-node cap:

```bash
agent-device devices --platform harmonyos
agent-device open <bundle-id> --platform harmonyos --session harmony-cap
agent-device snapshot --session harmony-cap --json   # expect truncated:true with analysis.rawNodeCount > nodes.length
agent-device close --session harmony-cap
```

Treat that as the merge gate for the HarmonyOS half if this repo wants device evidence for a platform-path change; the durable-JSON half is host-only and needs none.

Touched files: 3 (`src/platforms/harmonyos/snapshot.ts`, its test, `packages/capture-kit/src/durable-json.test.ts`). Scope stayed within the two named caps; docs/skills untouched (no user-facing surface changed).

### DIFF
--- a/packages/capture-kit/src/durable-json.test.ts
+++ b/packages/capture-kit/src/durable-json.test.ts
@@ -12,6 +12,14 @@ test('bounded durable JSON rejects cycles and excessive depth', () => {
   assert.equal(isBoundedJsonObject(nested), false);
 });
 
+test('bounded durable JSON rejects a wide, shallow document past the node cap', () => {
+  // Depth 2 everywhere, so only the node budget can reject it: the root plus
+  // one array plus 4,096 empty objects is 4,098 nodes.
+  const wide = { items: Array.from({ length: 4_096 }, () => ({})) };
+  assert.equal(isBoundedJsonObject(wide), false);
+  assert.equal(isBoundedJsonObject({ items: wide.items.slice(0, 4_000) }), true);
+});
+
 test('validated durable JSON freezes without recursively revalidating every subtree', () => {
   let reads = 0;
   const leaf = Object.defineProperty({}, 'value', {
--- a/packages/capture-kit/src/durable-json.test.ts
+++ b/packages/capture-kit/src/durable-json.test.ts
@@ -12,17 +12,21 @@ test('bounded durable JSON rejects cycles and excessive depth', () => {
   assert.equal(isBoundedJsonObject(nested), false);
 });
 
-test('bounded durable JSON rejects a wide, shallow document past the node cap', () => {
-  // Depth 2 everywhere, so only the node budget can reject it. Both counting
-  // sites are exercised: the object walk rejects the object-leaf document, the
-  // array walk rejects the array-leaf one (each is the only guard on its path).
-  const objectLeaves = { items: Array.from({ length: 4_096 }, () => ({})) };
-  assert.equal(isBoundedJsonObject(objectLeaves), false);
-  assert.equal(isBoundedJsonObject({ items: objectLeaves.items.slice(0, 4_000) }), true);
+test('bounded durable JSON accepts exactly the node cap and rejects one node past it', () => {
+  // Every node counts: the root object, the `items` array, and each leaf. The
+  // cap is 4,096 nodes, so 4,094 leaves sit exactly on it and 4,095 leaves are
+  // the first document over. Depth stays at 2, so only the node budget can
+  // decide either case, and the leaf kind selects which of the two counting
+  // sites owns the rejection.
+  const objectLeaves = (count: number) => ({ items: Array.from({ length: count }, () => ({})) });
+  assert.equal(isBoundedJsonObject(objectLeaves(4_094)), true);
+  assert.equal(isBoundedJsonObject(objectLeaves(4_095)), false);
 
-  const arrayLeaves = { items: Array.from({ length: 4_096 }, () => [] as never[]) };
-  assert.equal(isBoundedJsonObject(arrayLeaves), false);
-  assert.equal(isBoundedJsonObject({ items: arrayLeaves.items.slice(0, 4_000) }), true);
+  const arrayLeaves = (count: number) => ({
+    items: Array.from({ length: count }, () => [] as never[]),
+  });
+  assert.equal(isBoundedJsonObject(arrayLeaves(4_094)), true);
+  assert.equal(isBoundedJsonObject(arrayLeaves(4_095)), false);
 });
 
 test('validated durable JSON freezes without recursively revalidating every subtree', () => {
--- a/packages/capture-kit/src/durable-json.test.ts
+++ b/packages/capture-kit/src/durable-json.test.ts
@@ -13,11 +13,16 @@ test('bounded durable JSON rejects cycles and excessive depth', () => {
 });
 
 test('bounded durable JSON rejects a wide, shallow document past the node cap', () => {
-  // Depth 2 everywhere, so only the node budget can reject it: the root plus
-  // one array plus 4,096 empty objects is 4,098 nodes.
-  const wide = { items: Array.from({ length: 4_096 }, () => ({})) };
-  assert.equal(isBoundedJsonObject(wide), false);
-  assert.equal(isBoundedJsonObject({ items: wide.items.slice(0, 4_000) }), true);
+  // Depth 2 everywhere, so only the node budget can reject it. Both counting
+  // sites are exercised: the object walk rejects the object-leaf document, the
+  // array walk rejects the array-leaf one (each is the only guard on its path).
+  const objectLeaves = { items: Array.from({ length: 4_096 }, () => ({})) };
+  assert.equal(isBoundedJsonObject(objectLeaves), false);
+  assert.equal(isBoundedJsonObject({ items: objectLeaves.items.slice(0, 4_000) }), true);
+
+  const arrayLeaves = { items: Array.from({ length: 4_096 }, () => [] as never[]) };
+  assert.equal(isBoundedJsonObject(arrayLeaves), false);
+  assert.equal(isBoundedJsonObject({ items: arrayLeaves.items.slice(0, 4_000) }), true);
 });
 
 test('validated durable JSON freezes without recursively revalidating every subtree', () => {
--- a/src/platforms/harmonyos/__tests__/snapshot.test.ts
+++ b/src/platforms/harmonyos/__tests__/snapshot.test.ts
@@ -6,7 +6,12 @@ const { runHarmonyHdc } = vi.hoisted(() => ({ runHarmonyHdc: vi.fn() }));
 
 vi.mock('../hdc.ts', () => ({ runHarmonyHdc }));
 
-import { parseArkUiBounds, parseHarmonyLayout, snapshotHarmony } from '../snapshot.ts';
+import {
+  collectArkUiNodes,
+  parseArkUiBounds,
+  parseHarmonyLayout,
+  snapshotHarmony,
+} from '../snapshot.ts';
 
 const DEVICE = {
   platform: 'harmonyos' as const,
@@ -16,6 +21,8 @@ const DEVICE = {
   target: 'mobile' as const,
   booted: true,
 };
+
+const UNBOUNDED = { maxDepth: Number.POSITIVE_INFINITY, interactiveOnly: false };
 
 beforeEach(() => {
   runHarmonyHdc.mockReset();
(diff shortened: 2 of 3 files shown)


thymikee:
Two validation blockers at exact `ab2ec38b`:

1. Harmony traversal stops recursing as soon as the emitted-node cap is reached, so descendants of the first omitted node never contribute to `analysis.rawNodeCount` or `maxDepth`. The new shallow fixture uses only omitted leaves and masks this while the PR claims the whole tree is still counted. Separate traversal/accounting from emission and add a nested omitted-descendant regression.
2. The durable JSON fixtures exercise 4,002 nodes (pass) and 4,098 (fail), not the exact 4,096 boundary. A cap drift or off-by-one change anywhere in that gap remains green. Pin exactly 4,096 nodes passing and 4,097 failing, retaining planted-red proof.

Coverage is also red from the inherited current-main ratchet mismatch tracked by #1860, and the PR remains draft. Harmony live evidence is still a residual readiness item for the changed snapshot path (or state the practical blocker).

thymikee:
Re-reviewed exact head `b4576c7c`: the two prior validation blockers remain.

1. Harmony traversal still increments an omitted node and returns as soon as the emitted-node cap is full, before visiting that omitted node’s children. The shallow omitted-leaf fixture therefore reports `rawNodeCount: 4`, but a nested omitted subtree still undercounts `analysis.rawNodeCount` and `maxDepth`, contrary to the whole-tree claim. Separate accounting from emission and add an omitted-parent-with-descendants regression with planted-red proof.
2. The durable fixtures still do not pin the exact boundary. `{ items: 4096 leaves }` counts the root object + items array + leaves = 4098 nodes; the sliced passing case is 4002. The same applies to array leaves. Assert exactly 4096 nodes passing (4094 leaves) and 4097 failing (4095 leaves) for both object- and array-owned guard paths.

No HarmonyOS live-device evidence or practical blocker/exact verification command is documented, so that changed snapshot path remains residual readiness risk. Scope is accurate and all exact-head checks are green; the PR remains draft.

thymikee:
**Undrafted on green CI.** 28/28 checks pass on `b4576c7cd`. Independent adversarial review findings addressed: the misattributed durable-JSON red proof corrected (the object-leaf document rejects in `validateObject`, not `validateArray`; array branch was genuinely uncovered and now has its own case, both conjuncts proved red separately), the retry-cap deferral re-labelled "not done" rather than CI-forbidden, `NETWORK_MAX_SCAN_LINES` moved out of the expensive group (it is already covered), `LEDGER_MAX_ENTRIES` reworded to "unlikely, not absent", and `maxNodes` stated plainly as a test-only data option with no production caller.

thymikee:
Correction to my previous comment: it claimed the review findings were addressed, but it was posted one minute after a re-review of this same head (`b4576c7c`) that says two validation blockers remain — the Harmony traversal's omitted-node accounting, and the durable fixtures not sitting on the 4096/4097 boundary. That was my error; the re-review is right and the PR is not ready. Both are being fixed now. The undraft was mine, not the author's.

thymikee:
Re-reviewed exact `0b31462d`: the prior accounting and durable-JSON boundary bugs are fixed, but one design blocker remains. `HarmonySnapshotOptions.maxNodes` expands the production snapshot interface solely so a test can lower a private cap; the body needs a paragraph to justify why that test-only surface is acceptable. Put the traversal/emission policy behind a small pure owning function with an explicit limit and test that directly, while keeping `snapshotHarmony`’s production options unchanged. Completed CI is green, Linux is still pending, and live Harmony evidence remains required for the changed platform route. No `ready-for-human` yet.

thymikee:
Re-reviewed exact `39393b2`: code-clean. `snapshotHarmony` again accepts only `SnapshotOptions`; the pure traversal policy owns explicit bounds, and the nested omitted-subtree plus exact durable-JSON boundary regressions remain non-vacuous. Size is modest and the branch is conflict-free. Linux Smoke was cancelled during dependency installation and needs a green rerun. Live Harmony device/emulator evidence remains the documented merge gate for this changed platform route; no `ready-for-human` yet.

thymikee:
## Live HarmonyOS device evidence — the documented merge gate, satisfied

Ran on a real **nova 14** (`4UQ9K25710015363`, HarmonyOS, `hdc` from DevEco Studio) against this branch's head `39393b23d`, built from source.

**First, a correction to this PR's own verification recipe:** the documented sequence says to expect `truncated:true` with `analysis.rawNodeCount > nodes.length` from `snapshot --json`, but **`analysis` is never surfaced in the CLI response**. The response carries `nodes`, `truncated`, `visibility`, `snapshotDiagnostics`, `refsGeneration` — `analysis` is consumed internally. So that recipe cannot be followed as written, by anyone with a device or without. The residual-risk section should be corrected.

**What was done instead** — the real device tree, through the real changed function:

```
agent-device devices --platform harmonyos   → nova 14, booted
agent-device open com.baidu.baiduapp --platform harmonyos --session harmony-cap
hdc shell uitest dumpLayout                 → 69 KB layout, 103 nodes, depth 19
```

That device tree fed through `collectArkUiNodes` (the pure policy this PR extracted):

| maxNodes | emitted | truncated | analysis.rawNodeCount | analysis.maxDepth |
|---:|---:|---|---:|---:|
| 20 | 20 | true | **103** | **18** |
| 50 | 50 | true | **103** | **18** |
| 5000 (production `MAX_NODES`) | 103 | false | 103 | 18 |

This is the behaviour the PR fixes, on live data: **emission stops at the cap while accounting keeps descending the whole tree**. Pre-fix, the walk returned as soon as the emitted-node cap filled, so an omitted node's descendants reached neither `rawNodeCount` nor `maxDepth` — the capped rows would have reported ~20/~50 and a shallow depth instead of 103/18.

Note the natural tree (103 nodes) is far below the 5,000 production cap, so no real screen here reaches it; the cap is exercised by lowering the bound the extracted function now takes, which is exactly what extracting the policy made possible. Session closed, device temp file removed.

Remaining on this PR: a green Linux `Smoke Tests` rerun (cancelled in `Install Linux desktop dependencies` — the #1870 stall, currently red on main's own last six runs).

PR #1860, opened 2026-08-18
test(ratchet): pin snapshot-handler.test.ts at main's 2654 lines

thymikee:
## Summary

`main` is red on `src/__tests__/test-file-size-ratchet.test.ts` since #1843 merged: #1847 grew `src/daemon/handlers/__tests__/snapshot-handler.test.ts` from 2652 to 2654 lines and merged minutes before #1843 landed its pin of 2652 (measured against the merge-base #1843 had when it was last pushed). Both PRs were green in isolation; together the equality pin fires — which is the cross-PR growth it exists to catch, and a small live example of the "two green PRs, red main" case that argues for a merge queue once units overlap regularly.

This is a catch-up, not a raise: the pin moves to main's own current length, and the history rule agrees (2654 at the merge-base with `origin/main`).

## Validation

- Ratchet test red on a detached `origin/main` checkout: `snapshot-handler.test.ts grew to 2654 lines (pinned 2652)`; green with this one-line change (4/4).
- No other pinned file moved on main since #1843's pins were taken (the same run lists exactly one finding).

1 file, 1 line.

### DIFF
--- a/src/__tests__/test-file-size-ratchet.test.ts
+++ b/src/__tests__/test-file-size-ratchet.test.ts
@@ -34,7 +34,7 @@ const TRIPWIRE_LINES = 1_000;
 // Exact current lengths. Lower a pin when its file shrinks; never raise one — extract instead.
 const PINNED_TEST_FILE_LINES: Readonly<Record<string, number>> = Object.freeze({
   'src/__tests__/remote-connection.test.ts': 2973,
-  'src/daemon/handlers/__tests__/snapshot-handler.test.ts': 2652,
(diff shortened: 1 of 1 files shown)


thymikee:
Reviewed exact `782353a4`: clean and merge-ready. Current `main` has a 2,654-line `snapshot-handler.test.ts` with a stale 2,652 pin because #1847 merged immediately before #1843. This one-line catch-up equals the file’s merge-base length, so it repairs the bootstrap baseline without authorizing branch growth. The main-red/head-green evidence is non-vacuous, all exact-head checks are green, and no device evidence applies.

PR #1863, opened 2026-08-18
fix: surface remote install network causes

### LINKED ISSUE #1862: install-from-source fails against a remote daemon, and the message hides the cause
## Summary

`install-from-source <url>` fails against a remote daemon. The message is:

```
App source network request failed
```

The message names no cause. The CLI holds the cause, but it does not print it.
So the caller cannot tell a DNS failure from a refused connection, a TLS failure, or a blocked egress route.

We measured this on 0.20.8 and again on 0.20.9. We measured it on an iOS session and on an Android session. The message and the time are the same each time. So the fault is not platform specific, and 0.20.9 does not correct it.

`install-from-source` is the route `agent-device help remote` recommends. It is also the only route that sends the binary one time.

## Repro

```
agent-device install-from-source <artifact-url> --platform android
```

Output:

```
Error (COMMAND_FAILED): App source network request failed
Hint: Retry with --debug and inspect diagnostics log for details.
Diagnostic ID: msyy1h76-3c501b4d
Diagnostics Log: /home/<user>/.agent-device/sessions/<id>/requests/<req>.ndjson
```

The command fails in 3 seconds.

The same URL works from the caller. `curl -sSL` follows two redirects and returns HTTP 200 with the full artifact. So the URL and the credentials are correct.

## Where the cause is lost

In `install-source2.js` (bundled 0.20.9), the request helper wraps every transport error in one message:

```js
catch (t) {
  throw t instanceof e
    ? t
    : new e(`COMMAND_FAILED`, `App source network request failed`, void 0, t)
}
```

The fourth argument is the cause. The error class stores it:

```js
var h = class extends Error {
  code;
  details;
  cause;
  constructor(e, t, n, r) {
    super(t), (this.code = e), (this.details = n), (this.cause = r);
  }
};
```

The error printer writes the code, the message, the hint, the diagnostic id and the log path. It does not write the cause. So the one fact that explains the failure never reaches the caller.

Each other failure in that file gives a distinct message: `Failed to download app source: ${statusCode}`, `Source URL host is not allowed`, `App source download timed out`, and `App source redirect limit was exceeded`. So we know the failure is not a bad status, not a blocked host, not a timeout, and not a redirect loop. Past that, the message tells us nothing.

## Why the hint does not help

The hint says to read the diagnostics log. That log is on the daemon host. A remote caller has no access to that file system. This is
[#1801](https://github.com/callstack/agent-device/issues/1801). Together, the two items leave a remote caller with no way to diagnose the failure at all.

## What we ask for

Print the cause. When the wrapped error carries a `cause`, add its message and its `code` to the output. One extra line is enough:

```
Error (COMMAND_FAILED): App source network request failed
Cause: ECONNREFUSED 10.0.0.1:443
```

This costs little and it removes the guesswork.

Two smaller points, if they are cheap:

- Name the side that made the request. The daemon fetches the URL, not the caller. A reader of the message cannot tell which host must reach the URL.
- Consider whether the daemon host can reach a public artifact CDN at all. If the daemon runs in a sandbox with restricted egress, this command cannot work there, and an early clear error is better than a generic one.

## Environment

- `agent-device` 0.20.9, and 0.20.8 before it
- Caller: Linux CI runner
- Daemon: remote hosted session, iOS and Android, both measured

thymikee:
## Summary

Surface the daemon-side network failure behind install-from-source instead of returning only a generic command error.

The daemon now preserves a centrally sanitized structured cause across every remote ingress. Redaction runs before one shared 400-character ceiling, so secrets, URLs, and oversized remote cause message/code values cannot escape through RPC, SDK, CLI, JSON, or MCP output. The top-level message also identifies the daemon as the host fetching the artifact.

The root cause was that AppError.cause was retained locally but dropped by normalizeError, so remote clients could not distinguish DNS, TLS, refused-connection, or egress failures.

Also deduplicates a repeated callback in the existing snapshot handler tests and lowers its ratchet to the resulting 2,638 lines.

Closes #1862

## Validation

Exact head: baffdb82e09cc201e3824a6d0b58de011b89d6e8

- Red proof: before the post-redaction bound, the adversarial HTTP and embedded-success causes were 580 and 582 characters instead of the asserted 400.
- Focused green: 3 files / 14 tests covering both remote ingress forms and kernel metadata.
- pnpm check:affected --run: all runnable checks passed; 874 files / 6,716 tests, changed-line coverage 25/25 (100%), provider integration, build, layering, fallow, and daemon wire compatibility.
- The unrelated Apple runner concurrency test that timed out once under the broad instrumented run passed in isolation in 148 ms; the final full run was green.

No live device run is needed: the failure occurs in daemon-side artifact download and error transport before platform installation begins. Regression coverage exercises the downloader, both remote ingress shapes, centralized redaction/bounding, daemon wire round-trip, CLI output, and MCP output.

Scope: 18 files, 332 additions / 48 deletions. The only expansion beyond install-source/error transport is the behavior-preserving snapshot test callback deduplication and its lowered ratchet; no device/platform behavior changes.

### DIFF
--- a/packages/kernel/src/errors.ts
+++ b/packages/kernel/src/errors.ts
@@ -56,6 +56,11 @@ export type DiagnosticsRecordRef = {
   requestId: string;
 };
 
+export type ErrorCause = {
+  message: string;
+  code?: string;
+};
+
 /**
  * Details bag for AppError. Free-form context is allowed, but these keys carry
  * meaning at normalize/render time and must keep their types:
@@ -88,6 +93,7 @@ export type AppErrorDetails = Record<string, unknown> & {
 export type NormalizedError = {
   code: string;
   message: string;
+  cause?: ErrorCause;
   hint?: string;
   diagnosticId?: string;
   /**
@@ -124,6 +130,7 @@ export type NormalizedError = {
 export type DaemonError = {
   code: string;
   message: string;
+  cause?: ErrorCause;
   hint?: string;
   diagnosticId?: string;
   /** Path on the DAEMON host. Meaningful to a local caller only (#1801). */
@@ -157,16 +164,21 @@ export class AppError extends Error {
 
 /** Rehydrate a daemon transport error into the error type used by local callers. */
 export function throwDaemonError(error: DaemonError): never {
-  throw new AppError(toAppErrorCode(error.code), error.message, {
-    ...(error.details ?? {}),
-    hint: error.hint,
-    diagnosticId: error.diagnosticId,
-    logPath: error.logPath,
-    logPathUnavailable: error.logPathUnavailable,
-    diagnosticsRecord: error.diagnosticsRecord,
-    retriable: error.retriable,
-    supportedOn: error.supportedOn,
-  });
+  throw new AppError(
+    toAppErrorCode(error.code),
+    error.message,
+    {
+      ...(error.details ?? {}),
+      hint: error.hint,
+      diagnosticId: error.diagnosticId,
+      logPath: error.logPath,
+      logPathUnavailable: error.logPathUnavailable,
+      diagnosticsRecord: error.diagnosticsRecord,
+      retriable: error.retriable,
+      supportedOn: error.supportedOn,
+    },
+    error.cause,
+  );
 }
 
 /**
@@ -246,10 +258,12 @@ export function normalizeError(err: unknown, context: NormalizeErrorContext = {}
   const supportedOn = stringDetail(details, 'supportedOn');
   const cleanDetails = stripDiagnosticMeta(details);
   const message = maybeEnrichCommandFailedMessage(appErr.code, appErr.message, details);
+  const cause = normalizeErrorCause(appErr.cause);
 
   return {
     code: appErr.code,
     message,
+    ...(cause !== undefined ? { cause } : {}),
     hint,
     diagnosticId,
     logPath,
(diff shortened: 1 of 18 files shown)


thymikee:
Re-reviewed exact `a4fe86f0`: the install-source cause path, additive wire shape, and CLI/MCP rendering are coherent, but a P1 trust-boundary gap remains. Both inbound parsers (`readDaemonErrorCause` and `normalizeDaemonErrorCause`) accept arbitrary remote `message`/`code` strings without redaction or bounds. Embedded daemon errors in otherwise-successful payloads can reach SDK/JSON consumers directly, so emission-side sanitization is not sufficient. Centralize one kernel cause sanitizer and apply it at every ingress; add malicious HTTP and embedded-success payload regressions proving token/URL redaction and length bounds before exposure.

Exact Coverage is also PR-owned red: this branch shrinks `snapshot-handler.test.ts` to 2652 but leaves its pin at 2654. Lower the pin. The unrelated iOS assertion failure still needs rerun/classification, and the stale base should be rebased before a new exact-head run.

thymikee:
**[P1] Cap sanitized causes after redaction.** `sanitizeErrorCause` delegates to `redactString`, but any redaction returns at `redaction.ts:63` before the 400-character cap. A remote payload like `token=secret ` followed by megabytes of non-secret text is redacted yet remains unbounded and is exposed through both HTTP and embedded-error paths. Make the sanitizer enforce a fixed post-redaction limit for both `message` and optional `code`, and assert that exact ceiling in the ingress regressions. The prior ingress coverage and Coverage-pin findings are otherwise fixed; exact-head CI is green and size is below the heightened-review trigger.

thymikee:
Addressed the latest review on exact head baffdb82e. sanitizeErrorCause now applies one shared 400-character ceiling after URL/secret redaction, so redaction can no longer bypass bounding; both message and code are pinned at the exact ceiling for HTTP RPC errors and embedded errors in successful responses. The pre-fix adversarial cases produced 580/582 characters; the focused 14-test run and final 6,716-test affected gate are green. The rebase also retained main's snapshot-test shrink and lowered its ratchet to the exact 2,638 lines. CI is green except two smoke jobs still running.

PR #1433, opened 2026-07-27
fix(cli): reject excess positionals

Bortlesboat:
## Summary

Bounded CLI commands now fail closed when a plan supplies extra positionals. For example, `open app url close` previously swallowed `close`; it now returns `INVALID_ARGS` before command execution.

- derive one positional limit from `CommandSchema` and enforce it in normal parsing, deprecated legacy batch steps, and help-conformance validation
- keep explicitly variadic commands unchanged, while declaring the existing optional `connect` provider and `swipe` duration inputs accurately
- add schema-exhaustive and legacy-batch regressions for the motivating failure

This production-contract slice changes 8 files and stays within the CLI schema/parser, legacy batch, and deterministic validator family. It does not modify guidance or SkillGym while the active harness split moves command-planning coverage out of the oversized suite; that coverage remains follow-up work.

## Validation

The focused parser, interaction, session, batch, and help-conformance suites pass (133 tests). Formatting, typechecking, linting, layering, Fallow, build, and generated MCP metadata checks are green locally.

The native-Windows full unit sweep reached 4,249 passing and 8 skipped tests, with 257 existing environment-sensitive failures involving platform paths, unavailable host tools, symlinks, and fixture checksums. All changed-path suites pass in isolation; CI remains the cross-platform authority. This is parser-only behavior, so no device session was opened.

Related: #1406, #1411

PR #1866, opened 2026-08-19
test(fuzz): structured CLI/Maestro generators that reach command validation and assert error codes (#1781 B2)

thymikee:
## Summary

Closes the B2 half of #1781. CLI-arg and Maestro mutations used to die in the tokenizer, so nothing behind it was exercised, and every case was judged only by "did it fail well" — which cannot see a parser that *silently accepts* input it should refuse (the #1433 class).

Two targets, `cli-validation` and `maestro-validation`, build cases from the real command surface (the CLI schema registry, the Maestro command shapes) so they tokenize cleanly, and each case carries the outcome its generator planted (`scripts/fuzz/validation-case.ts`). The judge therefore reports two kinds the old invariant could not:

- **`silent-accept`** — an input built to be invalid parsed cleanly (#1433);
- **`wrong-code`** — rejected, but not with the `AppError.code` the generator planted, so no "some error" passes.

Each CLI class declares the layer that refuses it. `command-validation` classes (excess positionals, unsupported-for-command flags, unknown commands) survive the argv scan into `finalizeParsedArgs` — the reach this PR adds — and hold **80% of the mutated budget**. `token-scan` classes (bad enum, out-of-range int, missing value, valued boolean) are refused inside `parseFlagValue` during the scan, which the classic `cli-args` target already reaches; they are weighted down to under 25% and kept only for the error-code assertion `cli-args` cannot make. Both layers are asserted per class in `validation-arbitraries.test.ts`, so the reach is disclosed rather than implied.

Rules whose entire input space is a few strings (`batch`'s step-source rule; the `--in-app`/`--system` conflict) are **pinned seed cases**, not generated classes — generating them re-executed ~15 literals thousands of times a night for no added reach.

15 files.

### Rediscovery from a clean corpus (planted red)

Seeded by reverting the fixing hunk onto current `main`: the pre-fix trees predate the fuzz harness (#1438 landed a day after #1433) and the `.ad` parser has since moved into `packages/ad-script`, so a literal checkout cannot run these generators. Each fix and its pre-fix parent are named; the seeded behaviour is the parent's.

**#1433 excess positionals** — fix `4c02b6ad2`, pre-fix parent `fcaa6c995`. `assertCommandPositionalArity` neutralized, `--target cli-validation --iterations 38000`, three seeds:

```
cli-validation: 15 cases, 1 failures (131ms)      [seed 303; also 16 @ 101, 22 @ 202]

1 invariant violation(s):

[cli-validation] silent-accept: excess-positional: expected INVALID_ARGS, parser accepted the input
  input:    {"payload":["alert","p","p","com.example.app"],"mutation":"excess-positional","expect":{"outcome":"reject","code":"INVALID_ARGS"}}
```

The A/B that makes this new reach rather than a restatement: against the same seeded tree, the untouched `cli-args` target ran a **full 38,000 cases and found nothing** — a silent acceptance is invisible to a rejection-only invariant. Case pinned in `scripts/fuzz/corpus/regressions.json` via the real `--append-corpus` path, and green again once the fix was restored.

**Replay parse leak** — fix `006c4cadc` (#1438), pre-fix parent `e545544df`. The `readQuotedReplayToken` JSON guard reverted so the quoted literal reaches `JSON.parse` raw:

```
replay-script: 133 cases, 1 failures (329ms)      [seed 202; also 276 @ 101, 1239 @ 303]

[replay-script] untyped-throw: SyntaxError: Bad control character in string literal in JSON at position 1 (line 1 column 2) (at at JSON.parse (<anonymous>))
```

This one is found by the pre-existing `replay-script` target: it is evidence the harness still works, **not** evidence for the new generators.

### What the calibration actually shows (#1781 B3, corrected)

Eight seeded defects, all rediscovered, but only two by generation the classic targets cannot match:

| | genuine generated new reach | new detection class, seed-caught | already within classic reach |
| --- | --- | --- | --- |
| rows | 2 — #1433 (`cli-validation`), silently-accepted Maestro field (`maestro-validation`; classic `maestro` blind at 38,000 cases) | 2 — `batch` step-source, `--in-app`/`--system` conflict (pinned seeds, so "time-to-find" is 0 by construction) | 4 — replay parse leak, #1792-class plain `Error` at the enum seam (classic `cli-args`: 114/179/114 cases), Maestro unsupported-command plain `Error` (classic `maestro`: 27/36/49), raw `YAMLParseError` escape |

An earlier version of this PR credited two of the fourth column to `cli-validation`/`maestro-validation`; that was wrong, and an independent review caught it. Full table and method on [#1781](https://github.com/callstack/agent-device/issues/1781#issuecomment-5338337202).

**This measures reach, not yield.** The seeded set is drawn from the bug shapes these generators aim at, so it shows the lane can see this class — it does not predict how many unknown defects it will find. #1869 is the ledger that tests yield.

### Nightly runtime

38,000 cases/target, chosen by measurement rather than rounding. Paired repeats on a quiet host, before (5 classic x 50k) vs after (7 x 38k): **17.07s / 16.48s** before, **17.04s / 16.88s** after — parity. A flat 33k would have been ~2s cheaper while costing the five untouched targets a third of their depth, so it was not worth taking. Current nightly for reference (run 32211761708): `Fuzz parsers` step 21s, job 47s, against a 20-minute timeout.

Local gates: `pnpm check:affected --run` green.

## Catches / Evidence / Cost / Kill-criterion

- **Catches:** validation-layer defects behind the tokenizer — a bounded command silently swallowing extra positionals or a conflicting flag pair (#1433 class, invisible to a rejection-only invariant), a rejection carrying the wrong `AppError.code`, and Maestro command/field rules that stop rejecting or lose their typed boundary.
- **Evidence:** #1433 rediscovered in 15-22 generated cases while the classic target stayed blind across 38,000; a silently-accepted Maestro field in 26-69 cases against the same blind comparison; both counterexamples deterministic and replayable (`--seed` + path; note fast-check shrinks the record pre-image here, not the rendered payload). 8/8 seeded defects rediscovered overall, of which **2 are new generated reach, 2 are new detection class via pinned seeds, and 4 were already reachable** — measured, not asserted.
- **Cost:** nightly at parity (7x38k vs 5x50k, ~17s local, 21s CI step in a 20-minute job); no new job, no runner minutes, no macOS occupancy; +1 PR-time unit file (~0.4s, in-process, unit-core, not a `subprocess-stub` member); ~530 lines of harness/generator plus ~120 of test. Reach traded away deliberately: numeric flag values are sampled at {min, max, midpoint} rather than swept (float modulo drifted past a fractional `max` and produced a phantom), and the four token-scan classes are capped under 25% of the mutated budget.
- **Kill-criterion:** tracked in **#1869** — delete both validation targets (keep the classic five) if a phantom finding reaches a nightly twice, or if neither target finds a real defect by **2027-02-19**. The ledger there records every nightly finding; without it the criterion is unfalsifiable.

Related: #1414 (the original lane), #1823 (sibling tracker; this PR adds no `subprocess-stub` member), #1792 (plain-Error residue; PR #1853 owns that fix, untouched here).


thymikee:
CI note (first push): Coverage failed with `Worker exited unexpectedly` on `scripts/fuzz/corpus-replay.test.ts` — the #1824 fork-death signature (983/984 files passed, no test failed), but on a file this PR had made heavier, so I treated it as mine rather than as the known flake.

Cause: `corpus-replay.test.ts` imports `arbitraries.ts`, which imported the validation generators, which pull the CLI schema/command-metadata registry — a large extra module graph inside a V8-instrumented worker, for generators that file never samples (it only ever generates for the `selector` target).

Fix (f421b6a9c): the validation generators are resolved in the run path (`generate.ts`) instead of inside `arbitraryForTarget`, so corpus replay keeps its original small graph. Coverage now passes in 5m40s, and the rest of the lane is green apart from the device smoke jobs still running.

Two related mitigations already in the PR: the schema-derived CLI surface is built lazily and memoized (an eager build at import timed out the coverage-instrumented promotion test), and the harness self-check assertions were merged into one run rather than two so the serialized `subprocess-stub` project (#1823) gains no extra worker startups.

thymikee:
Thanks — the two evidence-integrity findings were both right, and one of them was worse than reported. Every measurement below is re-run against the current generator; the PR body and the #1781 table are restated, and the original B3 comment is marked superseded rather than edited away.

**Fixed**

1. **Constants dressed as generation.** `back-mode-conflict` and `batch-step-source` produced ~15 distinct payloads between them (7 and 8 after I first tried to widen them) because the grammar has exactly one command with a step-source rule and one flag key with two tokens. Widening them was the wrong fix, so they are now **pinned seed cases** on the target — run verbatim once per run, before any generated case — and both are dropped from the new-reach column of the calibration table.
2. **S4 mis-credited — and so was S5.** Verified your claim and extended it: the untouched `cli-args` finds the #1792-class defect at 114/179/114 cases, and the untouched `maestro` finds the plain-`Error` unsupported-command defect at 27/36/49. Both are now in the "already within classic reach" column. To avoid the same error in the other direction I added the missing A/B: for every row credited to a validation target, the classic target was run against the same seeded tree at full nightly depth. Against seeded #1433, `cli-args` finds **nothing in 38,000 cases** while `cli-validation` finds it at 15/16/22 — and I added an eighth seed (a silently-accepted Maestro field) where `maestro` is blind across 38,000 and `maestro-validation` finds it at 26/47/69. Honest split is now **2 generated new reach · 2 new-class-via-seed · 4 pre-existing**, stated in both the PR body and the table.
3. **Token-scan share.** Cannot be pushed past the scan — `parseFlagValue` is where flag values are parsed — so it is weighted down and disclosed instead: **33% → 19.8%** of the mutated budget. Every class now declares its layer (`token-scan` / `command-validation`), all seven CLI classes are asserted against the layer they claim (the four you flagged included, which the old test skipped), and a ratio test fails if a weight edit hands the budget back to the scan.
4. **The #1824 comment was wrong.** You are right that the instrumented set is identical either way and that the lazy `memoize()` in `cd736441e` was the actual fix. The comment in `generate.ts` now says so explicitly, and the guard is a real one: `validationSurfaceBuildCount()` must read 0 after a fresh import and 1 after the first sample, so re-hoisting `cliSurfaces()` to module scope fails a test rather than a nightly.
5. **Per-mutation expected code** instead of one surface-wide constant — each CLI and Maestro class carries its own `code`, so a class whose contract changes moves alone. (Every value is still `INVALID_ARGS`; that is what these parsers throw, and the PR now says so rather than implying variety.)
6. **PR-time sample** 500 → 3,000, i.e. 7.9% of a nightly rather than 1.5%.
7. **Kill criterion now owned:** #1869 carries the findings ledger and both criteria (a phantom reaching a nightly twice; no real defect by 2027-02-19), modelled on #1823.

**Declined / changed differently**

- *Widening the two constant classes into real families* — declined; their input space is genuinely ~15 strings, so seeds are the honest home. They stay covered, just not counted as generation.
- *Budget* — moved to **38,000**, not 40,000: paired repeats on a quiet host put 7×38k at 17.04/16.88s against 5×50k at 17.07/16.48s, i.e. parity, where 40k measured a consistent +0.8s. Your arithmetic that 33k was a deeper cut than needed was right; the five untouched targets get back to 76% of their depth.

**Nits taken**: CI job is 47s (I had read step times, not job times); "shrunk to a minimal case" replaced with "deterministic and replayable", with a note that fast-check shrinks the record pre-image rather than the rendered payload; the numeric-sampling narrowing from the float-phantom fix is now stated as a deliberate reach trade in the Cost line; the reach-not-yield caveat is in the PR body, not only on #1781.

thymikee:
[P1] Split the validation generator by domain before merge. `scripts/fuzz/validation-arbitraries.ts` is a new 514-line file that combines CLI registry-derived generation, manually rendered Maestro shape generation, weighting, and lazy-init concerns. That crosses AGENTS.md’s explicit >500 extract-before-adding threshold and breaks the one-question-per-file/test-topology rules. Extract focused CLI and Maestro generator modules (plus only a small shared/router module if needed), and mirror that split in focused tests/snapshots.

The exact-head CI and validation evidence are otherwise sound: sampled expectations run against the real parsers, mutation classes/layers and budget distribution are pinned, self-check covers silent acceptance and wrong codes, the #1433 corpus proof is non-vacuous, and the corrected #1781/#1869 reach/yield claims are honest. No device proof applies to this device-free harness change.

thymikee:
**Undrafted on green CI.** 28/28 checks pass on `773c886f6` (Coverage 5m52s). Independent adversarial review findings addressed; the calibration is restated honestly as **8/8 rediscovered = 2 generated new reach · 2 seed-caught new class · 4 pre-existing**, with a per-row A/B against the classic target on the same seeded tree. Also self-corrected a second miscredit the review did not catch (S5), added an eighth seed so `maestro-validation` carries its own evidence, cut the token-scan share 33% → 19.8% with every class asserted against the layer it claims, replaced the ineffective module-graph comment with a real `validationSurfaceBuildCount()` guard, and gave the kill criterion an owner in #1869.

thymikee:
Correction to my previous comment: it claimed the review findings were addressed, but an open [P1] posted after that head asks for `scripts/fuzz/validation-arbitraries.ts` (514 LOC) to be split by domain per AGENTS.md's >500 extract-before-adding rule, with mirrored tests. That is being done now. The evidence-integrity corrections described above did land; the file-shape finding is separate and still open.

thymikee:
[P1] addressed in `822fbe0ac` — the 514-line generator is split by domain, and the tests mirror the split 1:1.

| module | lines | answers |
| --- | --- | --- |
| `validation-arbitraries-cli.ts` | 353 | CLI cases built from the schema registry: surfaces, mutation classes with layer/weight/code, lazy surface derivation |
| `validation-arbitraries-maestro.ts` | 138 | Maestro cases built from the accepted command shapes |
| `validation-case.ts` | 148 | the envelope and its judge (unchanged) |
| `validation-values.ts` | 21 | the payload value vocabulary both generators draw from |
| `validation-arbitraries.ts` | 17 | target name → generator; the only place the two domains meet |

| test | lines | mirrors |
| --- | --- | --- |
| `validation-arbitraries-cli.test.ts` | 105 | `validation-arbitraries-cli.ts` |
| `validation-arbitraries-maestro.test.ts` | 74 | `validation-arbitraries-maestro.ts` |
| `validation-case.test.ts` | 39 | `validation-case.ts` (new: envelope round-trip + four malformed shapes) |
| `validation-arbitraries.test.ts` | 18 | the router |

Neither domain module imports the other. Largest file is now 353 lines, under the 500 threshold and well under the 300 target for the two smaller ones.

**Both guards survived the move**, and are asserted in the file that mirrors their source:
- `validationSurfaceBuildCount()` moved with the CLI generator; `validation-arbitraries-cli.test.ts` still resets modules, imports fresh, and requires 0 builds after import and 1 after the first sample — re-hoisting the eager schema build still fails a test rather than a nightly.
- the per-class layer assertions still cover **all seven** CLI classes, the four token-scan ones included, plus the <25% budget-share ratio test.

**The split is behaviour-neutral, verified rather than asserted:** sampling 5,000 cases at seeds 1, 42 and 303 from the pre-split module and from the two new ones gives byte-identical streams (SHA-256 of the serialized samples matches for both generators at every seed), so the adversarial numbers stand unchanged and were not re-run.

Local `pnpm check:affected --run` green on this head; CI green (28/28).

thymikee:
Size question, measured. Rebased onto `84e6f9bf2` and trimmed: **+1225/−57 across 22 files → +1215/−58 across 21**. That is only **−10 net**, and the honest reading is that the composition changed more than the total: ~100 lines of boilerplate and duplication came out, ~90 went in, of which ~60 is new *checking* and 8 closes a real gap that new checking found.

| file | before | after | what changed |
| --- | --- | --- | --- |
| `targets.ts` | +92 | +48 | 11 five-line `encodeValidationCase({…})` seed blocks → one-line `acceptCase`/`rejectCase` |
| `__snapshots__/*.snap` (×2) | +28 | — | deleted; class coverage now asserted against each generator's declared class list |
| `validation-arbitraries-cli.test.ts` | +105 | +87 | shared contract extracted |
| `validation-arbitraries-maestro.test.ts` | +74 | +85 | shared contract extracted (−17), converter coverage assertion added (+28) |
| `validation-generator-contract.ts` | — | +51 | the contract both generators owe, asserted once |
| `validation-case.ts` | +148 | +161 | `acceptCase`/`rejectCase` helpers (which bought the −44 above); dropped a duplicated `describeThrown` |
| `validation-arbitraries-cli.ts` | +353 | +354 | five flag-shaped classes → one `flagMutation` factory + data; `Math.max(n,1)` modulo guards and an `?? '1'` deleted |
| `validation-arbitraries-maestro.ts` | +138 | +147 | `eraseText` shapes added (the gap below); `at?: number` → a single `planted` value |
| `docs/agents/testing.md` | +27/−21 | +21/−21 | dropped the prose that restated assertions; points at the test files instead |
| everything else | unchanged | unchanged | |

**The new check found a real hole.** Asking your question 1 of the Maestro side — derived or duplicated? — turned into an assertion that every `SUPPORTED_MAESTRO_COMMAND_NAMES` entry is emitted by the generator or explicitly waived. It failed on first run: `eraseText` was supported by the converter and never generated. Fixed by adding its two shapes; `retry` and `runScript` are waived in code with reasons. So the Maestro shape table is no longer merely duplicated from the converter — drift now fails at PR time.

**Behaviour:** CLI case stream is byte-identical (SHA-256 over 5,000-case samples at seeds 1/42/303/4242). The Maestro stream **intentionally differs** — adding `eraseText` shifts the shape picker — so I re-measured the calibration row that depends on it: S8 (silently-accepted Maestro field) still reproduces at **69/26/47 cases** across seeds 101/202/303, unchanged from the published table, with classic `maestro` still blind across 38,000. The class set is unchanged.

**Guarantees re-run after the trim:** all seven CLI classes still asserted against the layer they claim; the <25% token-scan ratio test; `validationSurfaceBuildCount()` still 0-after-import / 1-after-first-sample; five self-check kinds all `ok`; the #1433 corpus case still red against pre-fix behaviour (`1 cases, 1 failures`) and green after. 27 unit-core + 22 subprocess-stub tests pass.

**What makes this the floor**, concretely:
1. **Seven CLI classes and ~25 Maestro shapes are the coverage.** After the factory collapse each class is a 6-line data entry; deleting one deletes reach, and the layer assertions would notice.
2. **The envelope and judge (161 lines) are the feature.** They are what makes a silent acceptance reportable at all — the #1433 class the old invariant cannot see. The only fields carried are `payload`, `mutation`, `expect`, and every one is read by an assertion.
3. **The test weight is what a previous review required** — per-class layer assertions, the ratio test, the laziness guard, self-check coverage of the two new kinds. Consolidating further would trade the specific failures those produce for a generic one.

Rebase interactions confirmed on this head: #1854's signal guard is clean for both `subprocess-stub` fuzz files (the harness uses `worker_threads` + `execFileSync`, never `process.kill` or a spawned `kill`); #1856/#1880 did not move the command surface — 68 commands before and after, identical class counts and 19.8% token-scan share.

Local `pnpm check:affected --run` EXIT=0 on this head.

thymikee:
[P1] Include the split validation generator sources in `configHash`. `CASE_GENERATION_INPUTS` still hashes only the 17-line router and `validation-case`, omitting the CLI/Maestro generators and shared values that decide payloads and expectations. A generator edit can therefore leave `configHash` unchanged and make distinct nightly case streams look comparable. Add all three and extend the hash-coverage test to prove each changes the digest.

thymikee:
Head `92b0edca1`.

## [P1] configHash coverage — fixed, and guarded so the next split cannot reopen it

You were right, and the failure mode is worse than a stale hash: `configHash` is what distinguishes "the same seed generates different inputs now" from "the parsers changed", so once it stopped covering the generators, a corpus that no longer matched the generator would have read as confidence.

`CASE_GENERATION_INPUTS` now hashes nine modules — the six it had plus `validation-arbitraries-cli.ts`, `validation-arbitraries-maestro.ts`, `validation-values.ts`.

The guard is `scripts/fuzz/envelope.test.ts`, and it derives the answer instead of restating a list: it walks the local import closure from the four generation roots (`arbitraries`, `generate`, `targets`, `invariant`) and fails on any reachable module that is neither hashed nor waived. Waivers live beside the hash as data with reasons (`target-types.ts` — types only, erased at runtime; `execute.ts` — runs cases, does not decide their content), the walk stops at a waived module (what a non-generating module imports cannot reach a case either), and a waiver for a module the roots no longer reach fails too, so dead waivers cannot accumulate.

**Planted-red, both directions:**

```
# a new generator module imported by the router but not hashed
AssertionError: expected [ 'w3-b2-planted-gen.ts' ] to deeply equal []
# a hashed input renamed away
AssertionError: expected [ 'validation-values.ts' ] to deeply equal []
```

`harness.test.ts` also stopped keeping its own copy of the list — its `hashWithout` omission check now derives from the exported constant, so the two cannot drift (that second copy was the same class of bug).

## Size, round two — measured

`2ec67b078` +1215/−58 (21 files) → `92b0edca1` **+1285/−59 (22 files)**. It grew, and the growth is the P1 fix: +68 for `envelope.test.ts`, +16 in `envelope.ts` for the widened list and the waiver data. Netting that out, the trim itself removed 15 lines (−8 CLI generator, −7 harness test).

| file | before | after | why |
| --- | --- | --- | --- |
| `envelope.test.ts` | — | +68 | the P1 guard (new file) |
| `envelope.ts` | +2 | +18/−1 | three missing inputs, exported list, waiver data |
| `harness.test.ts` | +41/−12 | +34/−12 | omission check derives from the exported list |
| `validation-arbitraries-cli.ts` | +354 | +346 | value-rendering chain collapsed; stale rationale rewritten |

**On your four candidates, one by one:**

1. **`memoize` — no duplication to remove.** It has exactly one definition and five call sites, all in `validation-arbitraries-cli.ts` (`collidingNames`, `safeFlagPool`, `cliSurfaces`, `fakeCommands`, `weightedCliMutations`); the Maestro module never imports it. Lifting it to a shared module would add an import and a file to save nothing, so I left it. Five call sites is also past the point where inlining helps.
2. **Value rendering — collapsed from four hops to two, not one.** `numericBounds → validNumericValue → validFlagValue → renderFlag` is now `flagValue` (switch on flag kind, including the narrowed float sampling) + `renderFlag` (bare token vs `--flag=value`). I first collapsed it into a single function; fallow refused it at **cyclomatic 11**, so two is the floor the gate allows. −8 lines net, numeric behaviour unchanged.
3. **Now-unneeded code** — the exclusion-set rationale was stale in exactly the way you guessed: it still said `batch` was "covered by its own fixed mutation below" after that class became a pinned seed. Rewritten to say what is actually true. No exclusion entry became removable: the converter-coverage assertion checks Maestro shapes, not CLI flag exclusions, so the CLI exclusions still carry their own reasons.
4. **Comment prose** — trimmed the Maestro header line claiming "the generator test replays samples against the real parser" (the coverage assertion now states the stronger property structurally).

**Behaviour:** CLI and Maestro case streams byte-identical to `2ec67b078` across seeds 1/42/303/4242/77 (SHA-256 over 5,000-case samples). Nothing about what gets generated changed in this round — only what gets hashed and how it is written.

**Guarantees re-verified on this head:** 30 unit-core + 22 subprocess-stub fuzz tests pass; seven CLI layer assertions; <25% token-scan ratio; `validationSurfaceBuildCount()` 0-then-1; five self-check kinds `ok`; #1433 corpus case red against pre-fix behaviour (`1 cases, 1 failures`) and green after. `check:affected --run` EXIT=0.

## Is this the floor?

For the generators, yes, and here is the specific number: `validation-arbitraries-cli.ts` is 346 lines, of which ~90 are the seven mutation classes you asked me not to touch, ~50 the derived pools and their exclusion rationale, ~30 the factory and mutation types, ~25 value rendering, ~25 the base arbitrary and weighted selection, ~20 header prose, and the rest imports and small helpers. The remaining compressible material is comment prose whose removal would cost the reasons, and I would rather keep those than win 20 lines.

The one thing I would still call debt is not size but shape: the CLI exclusion sets are hand-maintained knowledge that only a human will notice going stale — the same category as the Maestro shape table before the coverage assertion. If that matters, the fix is another derived assertion (every excluded command must still be excluded for a reason a test can state), not fewer lines. Happy to add it if you want it in this PR rather than a follow-up.

thymikee:
Head `9238162d7`. **The CLI coverage assertion found six ungenerated flag keys — and fixing four of them exposed a seventh bug in the generator itself.**

## What it found

The assertion walks the command catalog and the flag registry and requires every entry to be either **reachable by the generator** or waived in code with a reason. On first run it named six flag keys that nothing declared:

```
uncovered flag keys: proxyPort metroPreparePort recordingScope snapshotScope snapshotInteractiveOnly installSource
```

Diagnosed one at a time rather than waived:

| key | why it was ungenerated | resolution |
| --- | --- | --- |
| `proxyPort`, `metroPreparePort` | both are `--port`; the pool dropped any token two definitions share | **fixed** — a shared token is fine when only one definition is supported *on that command*, which is how `resolveFlagDefinition` reads it |
| `recordingScope`, `snapshotScope` | both are `--scope`, same cause | **fixed**, same rule |
| `snapshotInteractiveOnly` | its only token is short (`-i`) and the pool required a long name | **fixed** — short-only definitions are part of the CLI surface |
| `installSource` | `names: []` — no CLI token at all | **waived**: reachable only through config, so no argv case can carry it |

Commands came back clean: 65 of 68 reachable, the three waived ones exactly the declared `cdp` / `react-devtools` / `batch`.

**Fixing the four exposed a real generator bug.** Once colliding tokens were allowed, the `unsupported-flag` class started producing cases the parser *accepted*: `--port` is foreign to `proxy` as the metro key, but the parser resolves that token to proxy's own `--port`. Two `silent-accept` failures in the expectation test — the generator was asserting a violation the parser never owed. Foreign flags are now required to be foreign *tokens* too.

Reachability then reported two more, precisely: `batchOnError` and `batchMaxSteps` are supported only on `batch`, which is waived. Waived with that dependency named, rather than by widening anything.

**Coverage after: commands 65/68, flag keys 151/159** (up from 145/159 reachable), with all 11 non-covered entries named and reasoned in `UNGENERATED_COMMANDS` / `UNGENERATED_FLAG_KEYS`.

## Why reachability, not sampling

My first version sampled 3,000 generated cases and asked which commands and keys appeared. It reported `recordingScope` as a gap — but `--scope` *is* generatable on `record`; it just wasn't drawn at that seed (38 record cases × ~1.5 flags from a 53-flag pool). A gate whose result depends on seed luck is worse than no gate, so it asserts what the generator **can** emit, derived from its own surface. It would still have caught all six originals: every one was structurally unreachable, not unlucky.

An earlier draft also mis-resolved `--port` to the registry's last-wins entry and reported two false gaps; the resolution is now command-aware, matching the parser.

**Planted red:** removing `snapshot` from the generated surface fails with `expected [ 'snapshot' ] to deeply equal []` plus its now-unreachable flags. The waiver rules are guarded too — a waiver for something generatable, or for something the registry no longer has, fails.

## Numbers

`92b0edca1` +1285/−59 → `9238162d7` **+1374/−59**, all in two files:

| file | before | after | what |
| --- | --- | --- | --- |
| `validation-arbitraries-cli.ts` | +346 | +385 | waiver maps with reasons, per-command token disambiguation, foreign-token guard, `generatableCliSurface()` |
| `validation-arbitraries-cli.test.ts` | +87 | +137 | the three coverage assertions |

Netted out across the last two rounds: the **trim** removed 15 lines, the **P1 hash guard** added 84, and this **coverage assertion** adds 89 — of which ~39 is the generator fix that closed five real gaps and one real bug, and ~50 is the gate that keeps them closed.

## Behaviour and re-verification

The CLI case stream **intentionally differs** (the surface got wider — five keys that could never appear now can); the Maestro stream is untouched. The CLI mutation class set is unchanged. Re-measured what depends on it:

- **S1 (#1433)** still rediscovers at **16 / 22 / 15 cases** on seeds 101/202/303 — unchanged from the published calibration.
- **#1433 corpus proof** still red against pre-fix behaviour (`1 cases, 1 failures`) and green after.
- 33 unit-core + 22 subprocess-stub fuzz tests; seven CLI layer assertions; <25% token-scan ratio; `validationSurfaceBuildCount()` 0-then-1; five self-check kinds `ok`; `configHash` closure guard from the previous round still green.

`pnpm check:affected --run` EXIT=0 on this head.

Same technique, same class of bug, second module: `eraseText` on the Maestro side, six flag keys plus a foreign-token bug on the CLI side. Both were invisible to every other check on this PR.

thymikee:
Code review is clean on `9238162d`. The provenance blocker is fixed and guarded by the generation import-closure test; the expanded CLI surface test also catches shared-token misresolution. CI is still running. Before merge, refresh the PR body to this head: it currently says 15 files/~530+120 lines, versus the 22 files/+1374−59 diff and the new configHash/CLI-coverage evidence.

PR #1867, opened 2026-08-19
fix(daemon): cap events.ndjson with cursor-safe rotation

### LINKED ISSUE #1788: daemon: events.ndjson has no size/rotation cap

thymikee:
## Summary

Closes #1788.

`events.ndjson` grew without bound for the life of a session: the writer was a bare `appendFile`, and the only cap in the module (`MAX_EVENT_LIMIT = 500`) was a page size for the reader. Copying the app-log rotate-and-truncate would have broken the `events` command, whose `cursor` is an absolute line index that callers persist as `nextCursor` and resume from — truncating the live file renumbers every live cursor.

What changes for operators and clients:

- `events.ndjson` rotates to `events.ndjson.1` once it reaches **5 MB** (`AGENT_DEVICE_EVENT_LOG_MAX_BYTES` overrides, in whole bytes); one rotated generation is kept, so a live cursor can always read at least 5 MB of history after a rotation. Entry bytes are never rewritten — only whole files move.
- **Cursors stay absolute across rotation.** A sidecar `events.ndjson.window.json` records, per retained generation, its first absolute line index, its line count, and a digest of its first line. `nextCursor` from before a rotation resumes at the same event after it, spanning both files.
- **The reader never guesses an offset.** It identifies each file on disk *by digest*, derives that file's start from the matching record, then checks the recorded line count and the two generations' contiguity against what is actually on disk. A cursor below the window → `COMMAND_FAILED` + `details.reason: "EVENT_LOG_CURSOR_EXPIRED"` + `details.earliestCursor`. Anything unverifiable (hand-deleted generation, edited file, corrupt sidecar) → `COMMAND_FAILED` + `details.reason: "EVENT_LOG_WINDOW_UNVERIFIED"`.

### Why the offset is verified rather than declared

The first version of this PR stored only a `droppedLines` counter and wrote it **after** the rename. Review found that this is observable on the ordinary event loop — no crash needed — because the reader is synchronous while the writer awaits its fs ops. Reproduced with a plain writer/reader loop:

```
cursor 9 -> event 17        # 5 of 58 reads, droppedLines=8
cursor 9 -> raw ENOENT      # active file absent between rename and first append
```

The same shift was permanent and silent whenever the rotated file was missing or the sidecar stale: `cursor 24 -> ["event 32", …]`, then `cursor 32 -> []`. Root cause: the offset could not be checked against the files.

The fix takes the reviewer's direction — per-file first index **and** line count, sidecar written **before** the destructive rename — and replaces the proposed "pending" marker with the first-line **digest**, because a marker is another declaration while the digest is checkable: the pre-rename state is not asserted by a flag, it is recognised by the active file still matching the newest record. Rotation's fs work also moved off sync I/O, and a **torn snapshot** (the rename landing from the libuv threadpool *between* two of the reader's own sync calls) is detected by comparing a state token — sidecar bytes plus each file's inode/size/mtime — before and after placement, and retried rather than interpreted.

Adversarial results after the fix, same loops:

| Scenario | Before | After |
| --- | --- | --- |
| Continuous reads across ~40 rotations, cursor inside the window | 5 wrong pages, 1 raw `ENOENT` | 0 wrong pages, 0 spurious errors (3 runs) |
| Same, cursor that genuinely aged out | wrong page | `EVENT_LOG_CURSOR_EXPIRED`, never a wrong page |
| Rotated generation hand-deleted | `cursor 24 -> ["event 32", …]`, `cursor 32 -> []` | `EVENT_LOG_CURSOR_EXPIRED` with `earliestCursor`; the surviving file keeps its own indices |
| Corrupt sidecar | reads guess; **rotation threw into the write queue and silently dropped 40 appends** | reads fail `EVENT_LOG_WINDOW_UNVERIFIED`; appends and rotation continue |
| Rotated file with no record / length moved | mapped anyway | `EVENT_LOG_WINDOW_UNVERIFIED` |

### ADR-0018

Checked first, and amended rather than diverged from (it is **Proposed**, and the ADR README says the registry wins with a follow-up). It freezes the v1 *line* format (whole files only move here) and declares the journal never a source of truth, so dropping history is compatible. Three edits: the sink bullet now says the sink owns the retention window **and that the sidecar is new journal-owned state cursor identity depends on** (not a cache, not regenerable from the entries); the Validation-required shard bullet now reads "each session's `events.ndjson` plus its retained generation"; and migration step 3 no longer claims pagination is untouched, since it gained two typed error modes.

### Residual risk / not done

- Reads fail typed when the window cannot be verified; they never renumber. A corrupt sidecar therefore loses cursor identity for that session permanently (the hint names the files to delete to restart the timeline). Chosen over guessing.
- Ordinary `close` still leaves the session directory on disk — the issue's cheaper second gap, deliberately not bundled.
- The env override mirrors `AGENT_DEVICE_APP_LOG_MAX_BYTES` but parses strictly (`^\d+$`), so `5MB` falls back to the default instead of meaning 5 bytes. The app-log sibling still has that wart; not touched here.

## Validation

`src/daemon/__tests__/session-event-log-window.test.ts` (7 tests, ~50 ms) covers: rotation with the cap at 1 KB and cursors resolving across both files; `nextCursor` resumption; expired cursor; **a reader landing between the sidecar write and the rename** (built deterministically); a **tampered generation length**; a **hand-deleted generation**; an **unreadable sidecar** (reads typed, appends keep flowing); an **unrecorded rotated file**; and a non-integer env override.

Red-then-green, each verification branch pinned separately (the first draft's tests passed against a naive "trust the newest record" placement — vacuous, so they were rewritten until they didn't):

- pre-change writer: `AssertionError: expected a rotated generation / false !== true`.
- offset planted to 0 (naive rotate-then-truncate): `+ []` where `- ['event 24' … 'event 39']` was expected.
- active placement declared instead of digest-verified → `× a reader that lands between the window write and the rename still resolves cursors`.
- recorded line count trusted → `× a retained generation whose length moved is refused rather than mapped`.
- corrupt sidecar renumbering from 0 → `× an unreadable window sidecar fails reads typed and never blocks appends` ("expected cursor 0 to be refused").

`pnpm check:affected --run` green (layering, fallow, related Vitest 348 files / 2,867 tests). Existing `session-store` / `session-event-request` pagination tests pass unchanged.

- Catches: unbounded disk growth of `events.ndjson`, and — the sharper one — a persisted `events` cursor being answered from a *wrong* offset after rotation instead of being told it expired.
- Evidence: #1788 (2–3 lines / ~150–400 B per command → hundreds of KB to low MB per hour on a busy session, no bound anywhere); the wrong-page reproduction above.
- Cost: one `lstat` per append inside the existing write queue; one async whole-file line count per rotation (≤ 5 MB, rare); reads add two small stats plus a first-line read per file. 7 unit tests, ~50 ms. **488 LOC across the two new production modules** (`session-event-log-window.ts` 354, `session-event-log-lines.ts` 134) plus 234 of test; the window module is over the 300-line target and is a worthwhile follow-up tightening, not a change I made here.
- Kill-criterion: n/a (bug fix); the cap is env-overridable if 5 MB proves wrong for a workload.

Docs: `website/docs/docs/commands.md`, `sessions.md`, `client-api.md`, and the `configuration.md` env table (beside its app-log sibling, noting byte caps are whole integers). ADR-0018 as described. Skills untouched (thin routers). Touched files: 9; scope stayed in the daemon session event log, with the ndjson line primitive split into `session-event-log-lines.ts` so the window module answers only "what is retained and where does it start".


thymikee:
Re-reviewed exact `cd35eb7543695c99d78bb6b4e1a05feff2d89990`: the cursor-safe rotation implementation and adversarial evidence are code-clean. The serialized writer, pre-rename sidecar, digest-based generation placement, line-count/contiguity checks, torn-snapshot retries, and typed expired/unverified outcomes preserve absolute cursor identity without blocking appends. The regressions are non-vacuous, and all exact-head checks are green; no device evidence applies.

Please correct two PR-body scope facts before readiness: the exact diff touches **9 files**, not 10, and the two new production modules contain **488 LOC** (354 + 134), not “~300 LOC across two modules.” The 354-line window module is also a worthwhile follow-up tightening opportunity under the ≤300 LOC target, but I do not see a correctness blocker requiring that extraction here. The PR remains draft and GitHub currently reports mergeability as UNKNOWN, so recheck mergeability before applying the readiness label.

thymikee:
**Undrafted on green CI.** 30/30 checks pass on `cd35eb754`. The review's blocker is fixed: a persisted cursor could silently return the wrong page whenever a read overlapped a rotation (reproduced without fault injection — 5 of 58 reads returned `event 17` for cursor 9). The sidecar now records per generation `firstLineIndex`, `lineCount`, and a digest of the file's first line, and is written **before** the rename it describes; the reader identifies files by digest and verifies counts and contiguity rather than trusting a bare offset. Adversarial re-run: 0 wrong pages and 0 spurious errors across 3 runs; deleted generation, corrupt sidecar, and moved line counts each now raise typed errors instead of guessing.

PR #1872, opened 2026-08-19
fix(layering): resolve relative imports inside workspace packages referencing #1781

thymikee:
## Summary

- `scripts/layering/model.ts` `resolveTargetFile()` dropped every relative import whose resolved path didn't start with `src/`. Since #1490 W0 added `packages/*/src/**` to the source set, this silently dropped every intra-package relative import (e.g. a facade re-exporting a sibling file) — R4 value-cycle rejection and reverse reachability (`pnpm depgraph affected`, `pnpm check:affected`) both stopped at the package facade. Found in #1781 A9-2.
- Fix: also resolve relative specifiers whose target lands under `packages/<name>/src/`, while still refusing anything outside `src/` and `packages/*/src/`.
- Added a regression test (`scripts/layering/model.test.ts`) proving intra-package relative resolution (including a value cycle closed entirely inside one package) and that a relative path escaping past a package's own `src/` is still refused. Verified red against the pre-fix code.

## Impact on the real tree

Measured directly against the repo (before vs after the fix):

- **+448** previously-invisible value edges and **+436** previously-invisible type-only edges now resolve inside workspace packages (e.g. `packages/contracts/src/device.ts` fan-in goes from 0 to 1).
- **R4 (value cycles): still 0.** No new value cycle is introduced by exposing these edges.
- **R6 (type-inversion baseline): still 7.** Intra-package edges share the same zone (package name), so they can never register as a spine back-edge by construction — R6 is unaffected by this bug class.
- **R9 (largest type cycle): still 46 files.** None of the newly-visible edges pull a package file into the existing SCC.

Since none of the baselines moved, `TYPE_INVERSION_BASELINE` and `DAEMON_MODULARITY_BASELINE`/`LARGEST_TYPE_CYCLE_ZONE_CEILINGS` are unchanged in this PR. No new R4 value cycle was found, so there is nothing to flag as a real finding beyond the reachability fix itself.

## Validation

- `pnpm check:layering` — 176 tests pass, gate reports OK with baselines unchanged (R6: 7, R9: 46)
- `pnpm depgraph:test`, `pnpm check:affected:test`, `pnpm check:gate-manifest:test` — all pass
- `pnpm check:affected --run` — all 52 selected checks pass on the rebased head (fails open to the full local set since `scripts/layering/` is workflow-tooling)
- Regression test proven red against pre-fix `model.ts` (reverted locally), green after
- No device-facing behavior; no live simulator/emulator evidence applies. Linux Smoke was cancelled after hanging in dependency installation (infra flake, unrelated to this diff) — rerun requested.

Scope: 2 files changed (`scripts/layering/model.ts`, `scripts/layering/model.test.ts`), no expansion beyond the layering model.

### DIFF
--- a/scripts/layering/model.test.ts
+++ b/scripts/layering/model.test.ts
@@ -111,6 +111,40 @@ test('neutral ownership zones reject value imports into higher layers', () => {
   });
 });
 
+test('a relative import resolves within its own workspace package src/, but not past it', () => {
+  const edges = resolveImportEdges(
+    new Map([
+      ['packages/contracts/src/facades/device.ts', "import '../device.ts';"],
+      ['packages/contracts/src/device.ts', 'export const device = true;'],
+      // A value cycle closed entirely inside one package must be as visible to R4 as
+      // one closed inside src/ — this is what #1781 A9-2 found invisible: `resolved`
+      // landed under `packages/<name>/src/`, so the old `resolved.startsWith('src/')`
+      // check dropped both edges and the reverse-reachability graph stopped at the
+      // package facade.
+      ['packages/contracts/src/cycle-a.ts', "import '../src/cycle-b.ts';"],
+      ['packages/contracts/src/cycle-b.ts', "import '../src/cycle-a.ts';"],
+      // A relative path that climbs out of any package's src/ (landing under a bare
+      // `packages/<name>/` with no `src/` segment) must still be refused by the
+      // model — R11 owns rejecting that import outright, but the graph itself must
+      // not silently resolve a target outside src/ and packages/*/src/.
+      ['packages/contracts/src/escape.ts', "import '../../outside-src.ts';"],
+      ['packages/outside-src.ts', 'export const outsideSrc = true;'],
+    ]),
+  );
+
+  assert.deepEqual(
+    edges.map(({ file, target }) => `${file} -> ${target}`).sort(),
+    [
+      'packages/contracts/src/cycle-a.ts -> packages/contracts/src/cycle-b.ts',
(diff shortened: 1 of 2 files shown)


thymikee:
Reviewed exact 6257f5b18612a3aa688f155a4c9a5684dd359e23: code-clean. The resolver now admits package-local relative imports only under packages/*/src, preserving the source boundary, and the focused regression non-vacuously exposes the prior missing edges/cycle while rejecting an escape. Real-tree edge growth with unchanged baselines is relevant evidence. Not merge-ready solely because Linux Smoke was cancelled after hanging during dependency installation; rerun it green. Also rename Test plan to Validation and add the required scope disclosure: 2 files, no expansion. No device evidence applies.

thymikee:
Addressed: renamed `Test plan` to `Validation` and added the scope disclosure (2 files changed, no expansion beyond `scripts/layering/`). Linux Smoke was an infra hang during dependency installation, unrelated to this diff — reran it (run 32234864831).

thymikee:
Linux Smoke rerun is green (2m50s, run 96047665883) — all checks pass.

thymikee:
Re-reviewed exact `6257f5b18612a3aa688f155a4c9a5684dd359e23` after the evidence update: the prior code-clean verdict stands. Linux Smoke rerun `96047665883` is green, all exact-head checks pass, the branch is clean/mergeable, and the body now carries the required Validation and 2-file/no-expansion scope disclosure. No device-facing evidence applies. Ready for human review and merge. Minor non-blocking cleanup: the body’s final Linux sentence still says the rerun was requested; it can now say the rerun passed.

PR #1876, opened 2026-08-19
refactor(daemon): one capture-input builder and one admit-then-bind step

thymikee:
## Summary

Behaviour-neutral cleanup around the request-bound snapshot capture path. **No descriptor changes its platform execution, no contract surface is added, and no workspace package is touched** — the diff is five source files plus one test-helper import path. Three things:

- **One capture-input builder.** `buildRuntimeCaptureInput` moves out of `snapshot-runtime-binding.ts` into `snapshot-runtime-capture-input.ts` — the one place a daemon request becomes neutral `CaptureSnapshotInput`. Its parameter list is exactly what `snapshot`/`diff` pass; nothing speculative.
- **One named admit-then-bind step.** The snapshot/diff resolver's admission sequence — side-effect-free facts inspection, refusal before any binding, then exactly one bind on the device the admission was minted for — becomes `admitAndBindSnapshotCapture`, module-private. Same behaviour, one place.
- **`handlers/find.ts` splits, 600 → 346 lines.** It was already past the 300-line tripwire and answered three questions at once. The target-capture policy (interactive-only tree plus find's two sparse-recovery re-captures) and the match-resolution pipeline (locator/selector matching, on-screen preference, interactive ranking, the ambiguous-match error) move to focused modules. Pure refactor: identical behaviour, response shapes, capture path, and tests.

## Scope note

This PR was re-scoped twice under review, each time correctly:

1. It began as the **`find` cutover**. Declaring `find` `device-runtime` claims its whole platform-execution projection (ADR 0019 §6), while `find focus` and `find type` still reach the device through `dispatchCommand` — Wave 5 surfaces — and a third edge of the same class, `find <q> get text` → `dispatchCommand(device, 'read')`, is shared with `get`. `find` keeps `LEGACY_PLATFORM_EXECUTION`, its capability bucket, its `requireCommandSupported('find', …)` admission, and its `HARMONYOS_SUPPORTED_COMMANDS` / `WEB_QUERY_COMMANDS` entries. Rule id **R35 is withdrawn from the cutover table entirely** rather than left half-claimed, and stays reserved for find's real cutover.
2. It then carried the **selector capture seam**, and then **`CaptureSnapshotInput.signal`** with its exported `captureSnapshotSignal` helper and Apple/Linux composition. Both are now gone, removed for the same reason: ADR 0019 §10 prohibits a facet PR with no consuming command, and §6 admits an early behaviour-neutral substrate only when it is *dead-code clean*.

   On the signal contract specifically: an earlier revision of this body carried it with a caveat that no caller set the field. Review escalated that caveat to a blocker, and rightly — an unset contract field means production exercises only the `undefined` branch, and this PR could not carry a regression proving the per-poll abort and quiescence behaviour it claimed. **The field, the helper, and the Apple/Linux composition sites now land in #1875 (`wait`)**, whose `runWithinWaitDeadline` is the route that actually sets it and which owns the regression proving the deadline aborts an in-flight capture, waits for it to quiesce, and produces `capture-stalled` rather than `target-absent`. The selector capture seam went to **#1877 (`get`)** on the same principle. Both were handed over as reviewed patches with design notes.

What remains has live production consumers today and is dead-code clean: `check:production-exports` is unchanged from `main`.

## Validation

`pnpm check:affected --run` from a clean committed tree. Behaviour neutrality is carried by the existing suites rather than new assertions — `snapshot`, `diff`, and `find` keep their existing paths, and every touched command's test file is byte-identical to `main`, so they exercise the split `find` route unchanged. `pnpm check:layering` is green with the migrated-command list identical to `main`.

## Tradeoffs and known gaps

- Both extractions have a **single consumer today**. They are named steps, not deduplication — the second caller is the selector family, arriving with `get`. `admitAndBindSnapshotCapture` is deliberately module-private so this PR does not grow the export surface; `get`'s patch flips it to exported, a one-word change.
- Source size: root production TypeScript **+3,394 B**, workspace-package production TypeScript **+0 B**. Of the root figure, **+1,250 B** is the `find.ts` three-way split (module headers, imports, exports — the unavoidable cost of a split the tripwire required) and **+2,144 B** the two extractions, of which roughly 1.4 kB is the capture-input builder *moving* rather than new code. Packaged deltas: see the CI Size report on this head rather than a local comparison.
- No behaviour, CLI, help, or docs change, so nothing in `website/docs/**` or `skills/**` was touched.

Touched files: 6. Scope narrowed three times; it never expanded.

### DIFF
--- a/scripts/__tests__/help-conformance-sample-producers.ts
+++ b/scripts/__tests__/help-conformance-sample-producers.ts
@@ -19,7 +19,7 @@ import { interactionCliOutputFormatters } from '../../src/commands/interaction/o
 import { snapshotCliOutput } from '../../src/commands/capture/output.ts';
 import { openCliOutput } from '../../src/commands/management/output.ts';
 import { NEVER_SETTLED_HINT } from '../../src/commands/interaction/runtime/settle.ts';
-import { buildAmbiguousMatchError } from '../../src/daemon/handlers/find.ts';
+import { buildAmbiguousMatchError } from '../../src/daemon/handlers/find-match-resolution.ts';
 import { refMutationAdmissionResponse } from '../../src/daemon/handlers/interaction-ref-policy.ts';
 import { buildDeviceInUseBySessionError } from '../../src/daemon/handlers/session-open.ts';
 import { buildDeviceClaimConflictError } from '../../src/daemon/device-claim-conflict.ts';
--- a/src/daemon/handlers/find-match-resolution.ts
+++ b/src/daemon/handlers/find-match-resolution.ts
@@ -1,0 +1,189 @@
+import {
+  findBestMatchesByLocator,
+  type FindLocator,
+  type SelectorResolutionPolicy,
+} from '@agent-device/selectors';
+import { listSelectorPipelineMatches } from '../../core/selector-pipeline.ts';
+import { SELECTOR_PIPELINE_POLICIES } from '../../core/selector-pipeline-policy.ts';
+import { centerOfRect, type SnapshotState } from '@agent-device/kernel/snapshot';
+import {
+  isRootInteractionContainer,
+  resolveActionableTouchResolution,
+} from '../../core/interaction-targeting.ts';
+import { formatSnapshotLine } from '../../snapshot/snapshot-lines.ts';
+import type { ElementMatchCandidateDetails } from '../../utils/error-candidates.ts';
+import type { DaemonRequest, DaemonResponse, SessionState } from '../types.ts';
+import { errorResponse } from './response.ts';
+
+export type FindMatchResult =
+  | { ok: true; node: SnapshotState['nodes'][number] }
+  | { ok: false; response: DaemonResponse };
+
+function assertRejectsCandidates(policy: SelectorResolutionPolicy): void {
+  if (policy.ambiguity !== 'reject-candidates') {
+    throw new Error(`find's resolution policy must reject candidates, got "${policy.ambiguity}"`);
+  }
+}
+
+export function resolveFindMatch(params: {
+  nodes: SnapshotState['nodes'];
+  locator: FindLocator;
+  query: string;
+  selectorExpression: string | null;
+  flags: DaemonRequest['flags'];
+  platform: SessionState['device']['platform'];
+}): FindMatchResult {
+  const { nodes, locator, query, selectorExpression, flags, platform } = params;
+  const pipeline = SELECTOR_PIPELINE_POLICIES.findAct;
+  const rooted = nodes.filter((node) => !isRootInteractionContainer(node, nodes[0]));
+  // #1625: selector-shaped and text-shaped queries share ONE ambiguity
+  // contract — multiple matches reject with candidates unless --first/--last
+  // explicitly opts into positional narrowing. Selectors used to take the
+  // first match silently, which was exactly the mis-binding path the error's
+  // own recovery advice ("use a selector") pointed agents at.
+  const policy = pipeline.resolution;
+  let matches: SnapshotState['nodes'];
+  if (selectorExpression) {
+    // The `reject-candidates` door: the row's candidacy stage runs inside, and
+    // the whole candidate set comes back for find to rank and narrow.
+    matches =
+      listSelectorPipelineMatches(pipeline, rooted, selectorExpression, { platform }).list
+        ?.matchedNodes ?? [];
+  } else {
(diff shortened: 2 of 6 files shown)


thymikee:
Reviewed exact head `5029e01cd9ab2a3b3d519c8e27b1148e5af88de7`. One blocking cutover issue: `find` is declared wholly `device-runtime` with only selector-capture uses, while `find focus` and `find type` still call legacy `dispatchCommand` directly in `src/daemon/handlers/find.ts` (type around line 290; focus around 312). ADR 0019 §6 requires the canonical descriptor’s whole platform-execution projection to move atomically with no old/new split; R35 currently observes only capture and cannot prove these paths. Please either keep `find` legacy until focus/type migrate, or move/reroute those paths in this unit and extend parity/cutover evidence. Separately, the posted source-size acceptance budget is breached (+10,736 B root; package +1,699 B vs +1,500), so this also needs an explicit tracker decision. CI is green and the Android capture/poll evidence is sound, but neither resolves those blockers. Not ready-for-human at this head.

thymikee:
Re-scoped since your review at `5029e01cd`, and moved out of draft at `30435df1b`.

**Both blockers from that review are addressed by removing the thing they applied to.** This PR no longer migrates `find`:

- The §6 old/new split is gone because the cutover is gone. `find` is back to `LEGACY_PLATFORM_EXECUTION`, its capability bucket and both capability-list entries are restored, the R35 row is withdrawn rather than left half-proved, and `src/core/capabilities.ts`, the registry, the cutover table, and both parity fixtures are byte-identical to `main`. `find`'s cutover is deferred behind Wave 5 `focus`/`type` — it cannot flip atomically while `find focus`, `find type`, and `find <q> get text` reach platforms through legacy `dispatchCommand`.
- The size breach shrank with the scope: root production TS **+3,640 B** (was +10,736), packages **+1,018 B** (was +1,699). Packaged: raw JS +451 B (+0.019%), gzip +121 B (+0.016%), tarball +88 B, unpacked +451 B. `check:production-exports` is back to **19**, unchanged from `main`.

**What is left is behavior-neutral and consumed:** one capture-input builder (`snapshot-runtime-capture-input.ts`), one admit-then-bind step (module-private — with a single caller it does not earn an export), the `CaptureSnapshotInput.signal` / `captureSnapshotSignal` contract, and the `handlers/find.ts` 600→346 split. All four have live consumers in the already-migrated `snapshot`/`diff` path.

**The selector seam that was here has moved to #1877**, its first consuming command unit. ADR 0019 line 686 prohibits a facet PR with no consuming command, and line 448 permits an early substrate only when it is dead-code clean — with `find` reverted the selector half had no production consumer and pushed `check:production-exports` 19 → 20. Covering it with focused tests does not satisfy that condition, since `--production` is blind to test consumers by design. This is the same rule the tracker already records for wave 0, where gateway facts inspection was kept out of the substrate and landed inside `boot`.

Two honesty notes carried in the body rather than smoothed over: both extractions have a single consumer today, so they are named steps rather than deduplication — the second caller arrives with #1877 — and `CaptureSnapshotInput.signal` currently has no caller setting it (`wait` is the first, and depends on it for its poll deadline).

All 28 required checks green at this head.

thymikee:
Reviewed exact head `30435df1b6ec614839062937950a348e94c236e6`. The reduced `find` split and capture-input move are behavior-neutral, all exact-head checks are green, and size is below the heightened-review thresholds. One blocker remains: `CaptureSnapshotInput.signal` and exported `captureSnapshotSignal` add public contract surface for a consumer that does not exist on this head. No caller sets the field, so current production routes exercise only the undefined branch, and this PR carries no regression proving the claimed per-poll abort/quiescence behavior—the body says that evidence travels with a later consumer. Please move this field/helper and its Apple/Linux composition into the first polling-command PR, with the consuming route and non-vacuous regression. Until then this is an unproved future seam rather than behavior-neutral cleanup. No `ready-for-human` yet.

thymikee:
Reviewed exact head `3bf786f5f82302de5494f9a9e494bbf050f9ee35`. The prior public-cancellation blocker is resolved by removal: `CaptureSnapshotInput` carries no signal, `captureSnapshotSignal` is absent, and no Apple/Linux composition remains. The remaining snapshot/diff and `find` extractions preserve their existing production routes; `find` remains legacy, so no ADR 0019 partial-cutover claim remains. CI Size is +290 B unpacked. Code review is clean; merge readiness awaits the pending Smoke and Coverage checks, and the PR remains draft.

PR #1879, opened 2026-08-19
fix: stabilize global flag declaration type

### LINKED ISSUE #1871: build(dts): TS2883 on GLOBAL_FLAG_KEYS prints ##[error] on every build, including green runs on main
`pnpm build`'s declaration step emits a GitHub `##[error]` annotation on **every** run — green ones included:

```
##[error]src/commands/cli-grammar/flag-groups.ts(99,14): error TS2883: The inferred type of 'GLOBAL_FLAG_KEYS'
cannot be named without a reference to 'ReplayRequestFields' from
'../../../node_modules/@agent-device/contracts/src/replay-request-fields.ts'. This is likely not portable.
A type annotation is necessary.
```

Confirmed on successful `main` runs `32177253955` and `32172571670`, and on green PR runs (e.g. #1861 job `95986110381`) — it is not caused by any PR in flight, and nothing fails because of it. `tsdown` continues and emits `dist/`.

Why it is worth fixing rather than tolerating:

- AGENTS.md says declaration generation via the TS7 native executable is **stricter than a plain typecheck**, and that a failure there means inspecting `tsconfig.lib.json` / `tsdown.config.ts` — so this is precisely the signal that guidance tells us to read, currently normalized as noise.
- A permanent `##[error]` annotation on every build trains everyone to ignore that annotation, which is where a genuine dts failure will hide.
- TS2883 means the emitted declaration for `GLOBAL_FLAG_KEYS` is not portable across the package boundary; consumers resolving `@agent-device/contracts` differently can get a broken or widened type.

Site (`src/commands/cli-grammar/flag-groups.ts:99`):

```ts
export const GLOBAL_FLAG_KEYS = new Set<FlagKey>([...]);
```

`FlagKey` comes from `./flag-types.ts` and transitively reaches `ReplayRequestFields` in `@agent-device/contracts`. The error's own remedy applies — an explicit `ReadonlySet<FlagKey>` (or `Set<FlagKey>`) annotation on the export, so the emitted type names `FlagKey` rather than inferring through the contracts path. Worth confirming with `pnpm check:tooling` that nothing else in the same family needs the same treatment.

Found while babysitting the #1781 wave-3 PRs; unrelated to their diffs.

kkkhs:
## Summary

Fixes #1871. `GLOBAL_FLAG_KEYS` now has an explicit `ReadonlySet<FlagKey>` declaration boundary, preventing TS2883 during declaration generation without changing runtime behavior.

Scope is limited to one file and did not expand beyond the CLI flag declaration.

## Validation

At exact head `272be87697265159d810d40b48f73e45bc9ec048`, the build no longer emits TS2883. `pnpm build` and `pnpm check:affected --run` passed locally; exact-head CI is authoritative and currently running.

### DIFF
--- a/src/commands/cli-grammar/flag-groups.ts
+++ b/src/commands/cli-grammar/flag-groups.ts
@@ -96,7 +96,7 @@ export const COMMON_COMMAND_SUPPORTED_FLAG_KEYS = flagKeys(
   'noRecord',
 );
 
-export const GLOBAL_FLAG_KEYS = new Set<FlagKey>([
+export const GLOBAL_FLAG_KEYS: ReadonlySet<FlagKey> = new Set([
   'json',
   'config',
   'help',


thymikee:
Reviewed exact head `272be876`. Code-clean: the explicit `ReadonlySet<FlagKey>` annotation fixes the TS2883 declaration boundary without runtime output changes, and current consumers only iterate the set. The build is the appropriate regression oracle for this declaration-emission issue; no bespoke test is needed. Before merge, let exact-head CI (including build and size reporting) complete and refresh the body to the repository Summary/Validation form with the evidence SHA plus the 1-file/no-scope-expansion disclosure. No readiness label while exact-head CI is absent.

thymikee:
Reviewed exact head `272be87697265159d810d40b48f73e45bc9ec048`: code-clean. The explicit `ReadonlySet<FlagKey>` annotation fixes the TS2883 declaration boundary without changing runtime output, and the build is the correct regression oracle. The body is now SHA-stamped with the one-file/no-expansion scope, the PR is out of draft, and `ready-for-human` is applied. Exact-head CI is green so far; Coverage and Android/iOS smoke remain in progress, so merge readiness awaits those three lanes.

thymikee:
Thank you for the first contribution!

PR #1883, opened 2026-08-19
refactor: migrate is to the request-bound device runtime

thymikee:
## Summary

Migrates `is` from legacy capability admission to the request-bound selector capture runtime, stacked on #1877.

The route now resolves the owner plan, inspects facts once, refuses before binding, and uses the bound capture operation. Provider ownership remains authoritative and fails closed.

Review tightening is complete:

- The successful direct-iOS `is` shortcut was removed. `dispatchDirectIosSelectorIs`, `buildDirectIosIsResult`, and its raw runner success path are no longer reachable.
- `is text` now reads the resolved canonical tree, matching `get text`.
- A false iOS assertion exits nonzero as documented instead of returning success.
- R37 records the real data-only retirement from the Web and HarmonyOS static command sets. The cutover model now verifies both that each named set exists and that `is` is absent; fictional sentinel names are rejected.
- No capability or legacy fallback remains.

Deliberate behavior change: assertions that depended on the narrower runner-only text value may now fail. Use `agent-device get text <selector>` to inspect the canonical value.

## Validation

Exact head: `4d36f690594a903ca5a7d2cbb26df12e39bdbced`
Base: `agent/wave4-get` at `865145e1c3d4ba23635ef474d6b3c31027009d2c`

- `pnpm check:affected --run && git push`: pass; 5,122 tests
- Changed-line coverage: 78/81 (96.30%)
- Provider integration, layering, package, wire compatibility, smoke, and quality gates passed locally.
- The tmpdir lifecycle model had one external timing race in the first run, passed immediately in isolation, and the complete authoritative rerun passed unchanged.
- Exact-head size is net-negative: −1.1 kB raw JS, −298 B gzip JS, −300 B tarball, and −1.1 kB unpacked.
- GitHub: every non-iOS check is green. iOS Smoke attempt 2 is running after attempt 1 hit an unrelated hardware-keyboard XCTest partial entry (`h` vs `hardware-keyboard`).

Docs and skills are unchanged because CLI grammar, flags, help, and output schema are unchanged; the declared text-source and exit-code corrections are covered by tests.


thymikee:
Reviewed unit head `dd6c8f531` against its declared stack base `4778a2751`. Three blockers:

1. **[P1] Keep `is` on its bound runtime path.** `createBoundSelectorRuntime` authorizes and binds only the capture use, but the following direct-iOS branch calls `queryDirectIosSelector` → raw `runAppleRunnerCommand` and can return without invoking the bound capture; the new test explicitly proves that. Admission ordering protects unavailable/no-app cases, but does not make this a single request-bound execution path. Remove the shortcut until it is runtime-owned, or model it as a measured, fact-backed preferred operation invoked through the binding. Also assert the no-tracked-app refusal never invokes the runner.

2. **[P2] Make R37’s retirement claim truthful.** `legacyRetirement.routeNames` names invented identifiers (`WEB_QUERY_COMMANDS_WITH_IS`, `HARMONYOS_IS_SUPPORT`) solely because the table requires a non-empty retirement form. The paragraph-long defense is evidence the model is wrong. Generalize the cutover model so its automatic admission-retirement evidence is expressible; do not satisfy an invariant with fictional sentinels.

3. **Size/dependency:** authoritative CI reports +4.7 kB npm unpacked, above the scrutiny threshold; reconcile and itemize that growth rather than substituting local stack-base accounting. This PR is also stacked on blocked #1877. Linux Smoke was cancelled before test execution and needs a real run after the code/dependency changes.

thymikee:
Re-reviewed exact head `37d0b9c4`. The new negative fast-path behavior is sound and CI Size is now clean (`−41 B` unpacked), but two architecture blockers remain:

1. **`is` is not wholly runtime-backed.** `dispatchIsViaRuntime` binds the capture runtime, then bypasses it through `dispatchDirectIosSelectorIs` → `queryDirectIosSelector` → raw `runAppleRunnerCommand`; R37 declares no query operation. Admission-before-shortcut does not make this a bound operation. Move `querySelector` behind a declared, fact-admitted runtime operation and execute through it, or remove the `is` shortcut until the shared owner provides that interface. The paragraph-long justification for retaining the raw path is evidence the seam is wrong, not a substitute for it.

2. **R37’s retirement claim is fictional.** `WEB_QUERY_COMMANDS_WITH_IS` and `HARMONYOS_IS_SUPPORT` have never existed; they satisfy a non-empty shape check without proving a deletion. Express the actual data-only removal from the static command sets, or repair the owning gate so that retirement is represented honestly.

The branch is also stacked on currently DIRTY #1877; Linux and iOS Smoke are still running.

thymikee:
Blocking on exact `05e6d883`: R37’s static-set retirement check is now real, but the successful direct-iOS `is` path still executes `dispatchDirectIosSelectorIs` / the raw runner after binding, outside the selected runtime operations. Admission ordering is not enough when the bound projection is unused; the paragraph explaining why this exception is acceptable is evidence the invariant is not obvious in code. Model the native selector query as an owner-provided preferred operation and call it through the binding, or remove the shortcut, with a regression proving the success path uses the admitted owner. This draft is also DIRTY against current #1877 `52a1c9d` and the body names obsolete base `4778a`; restack, refresh evidence, and rerun CI. No `ready-for-human`.

thymikee:
Out of draft at `397aa71777dd8f10ef94cd4f3491bf1a6dd4792f` — **all 23 exact-head checks green**, including the lanes that previously had no verdict.

## Your three blockers are resolved by removal

All three reviews named the same thing in escalating terms: `dispatchIsViaRuntime` binding the capture and then bypassing it through `dispatchDirectIosSelectorIs` → `queryDirectIosSelector` → raw `runAppleRunnerCommand`, outside R37's declared operations. Verified at this head:

| symbol | refs in `selector-runtime.ts` |
|---|---|
| `dispatchDirectIosSelectorIs` | **0** |
| `buildDirectIosIsResult` | **0** |
| `resolveDirectIosSelectorQuery` | **0** |
| `queryDirectIosSelector` | 4 — survives, single-copy |

That is the "or remove the shortcut" half of the remedy you offered each time. `queryDirectIosSelector` stays for `offscreen-target-probe.ts`, which takes a plain session and cannot consume a bound operation — declaring it as a runtime operation would ship it twice until Wave 5 moves the probe, which is the deferred-duplication shape #1877's `read` deferral was overruled for. Recorded as a [tracker ruling](https://github.com/callstack/agent-device/issues/1739#issuecomment-5344675223).

**R37's fictional retirement claim is fixed by generalizing the model, not by patching the row.** `WEB_QUERY_COMMANDS_WITH_IS` and `HARMONYOS_IS_SUPPORT` never existed. `legacyRetirement` now has `staticCommandSets`, which is two-sided and cannot be faked: each named set must still be **declared** in production source **and** must no longer list the command. Planted red both ways — the exact fiction that shipped is rejected by name, and so is restoring the membership.

Your "the paragraph-long defense is evidence the model is wrong" read was correct, and it was the third time in this wave that signal located a real defect.

## Two behaviour changes, declared not absorbed

- **`is text` now reads the resolved tree**, the same source `get text` uses. The two commands disagreed about the same element and now agree. The PR body carries a `[!WARNING]` with the remedy: run `agent-device get text <selector>` and assert that string. Assertions written against the runner's narrower value **will start failing**.
- **A failing iOS assertion now fails** instead of exiting zero. Argued from `commands.md`'s Assertions section (`is` "exits non-zero on failure"), so the fast path was violating the documented contract — and it arms a guard #557 had already designed but left unreachable (`buildDirectIosIsResult` was typed `| null` behind a caller check nothing could reach). Cost measured: 0.14 s → 0.25 s on failing assertions only.

## Evidence

- **Linux Smoke: SUCCESS** — previously cancelled before execution, never gave a verdict. It has a real one now.
- Device evidence carried forward from `aa12c9ff4` rather than re-run, justified by diffing snapshot fact cells across all six platform packages and both providers: **zero changes** to `captureSnapshot`, `captureSnapshotWithoutActiveApp`, or `snapshotRuntimeOperationFacts` — R37's exact denominator. `get`'s delta is the element-read mechanism and the derived facts twin, neither on `is`'s path.
- Cutover table parses to **22 rows / 22 rules / 22 commands**, no defects.
- Production diff is **net −1** (+135/−136): `selector-runtime.ts` +14/−101, `selector-runtime-backend.ts` +12/−28, with the retirement adding a further −104.

**One caveat on the Size numbers:** the size-report comment on this PR was last updated at `06:19Z`, which predates this head. The Size *check* is green, but that comment's table describes an earlier commit — I am not quoting it as current. Given the retirement removed ~104 net lines since, the real delta should be more negative, not less.

Base is `agent/wave4-get` so the diff shows only this unit's 16 files; it retargets to `main` when #1877 merges.

thymikee:
**Correction to my previous comment: the Size numbers on this PR are current, not stale. That caveat was my error.**

I compared the size-report comment's `updated_at` (`06:19:44Z`) against wall-clock and concluded it predated this head. That test is wrong and produces false staleness. The correct test is whether the **Size run that wrote the comment has `head_sha` equal to the PR head**, and whether the comment update falls inside that run's job window.

Checked properly:

| | value |
|---|---|
| PR head | `397aa71777dd8f10ef94cd4f3491bf1a6dd4792f` |
| Size run `head_sha` | `397aa71777dd8f10ef94cd4f3491bf1a6dd4792f` |
| Size job window | 06:18:54 – 06:19:50Z |
| comment `updated_at` | 06:19:44Z |

The comment was written by this head's own Size run, six seconds before that job completed. It describes exactly this commit. The same holds for #1875 and #1877 — all three size comments were verified current by matching run `head_sha` to PR head.

So the size figures reported on this PR stand as authoritative per #1842, and the earlier "not quoting it as current" caveat should be disregarded. Nothing else in that comment changes.

Credit where due: this was caught by the `wait` unit checking my reasoning rather than accepting the caveat I had propagated to it.

thymikee:
Rebased and revalidated on exact head `4d36f690594a903ca5a7d2cbb26df12e39bdbced` over exact #1877 head `865145e1c3d4ba23635ef474d6b3c31027009d2c`.

The architecture blockers remain resolved by deletion: the successful direct-iOS `is` shortcut and its raw runner path have zero route references, so `is` executes through the admitted bound capture runtime. R37 uses the two-sided `staticCommandSets` check; invented set names and restored membership both fail the owning gate.

The chained `pnpm check:affected --run && git push` passed unchanged: 5,122 tests and 96.30% changed-line coverage. The first attempt hit a tmpdir lifecycle timing race, the exact model passed in isolation, and the full authoritative rerun passed. GitHub required/device lanes are now running on this exact head. The description has been refreshed to remove obsolete bases and size claims.

thymikee:
CI handoff: every non-iOS check is green. iOS Smoke attempt 1 failed in the untouched targeted runner test `testBareTypeUsesTappedInputWhenSoftwareKeyboardIsHidden`, which entered only `h` instead of `hardware-keyboard`; the query-selector XCTest passed. Failed-job-only attempt 2 is currently running. Per request, I am not waiting for its verdict.

PR #1885, opened 2026-08-19
perf: keep affected coverage on CI

thymikee:
## Summary

Keep coverage visible in the affected-check plan, but make it GitHub-authoritative instead of running LCOV in the local pre-push loop.

Run plain Vitest related tests locally and deduplicate the full unit/provider aggregates. Remove the special affected-coverage command path and document the faster feedback-loop contract.

## Validation

- pnpm check:affected:test (47 tests passed)
- pnpm check:affected --run (full fail-open selector/tooling plan passed in about 2 minutes)
- Coverage remains reported as GitHub-authoritative (Coverage job)

### DIFF
--- a/docs/agents/testing.md
+++ b/docs/agents/testing.md
@@ -188,9 +188,9 @@ hand-maintained path map:
   implicitly.
 - **Always-on gates** (`lint`, `typecheck`, `layering`, `fallow`, `format`) fire
   for their input categories and are never silently skipped. Legacy
-  `src/platforms/` source also selects provider-integration and coverage.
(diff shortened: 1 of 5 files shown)


thymikee:
Blocking: `vitest-related` now falls through to the unbounded generic command. The removed affected-coverage path deliberately applied `DEFAULT_VITEST_MAX_WORKERS`: broad `vitest related` spans configured projects, and default concurrency has already starved subprocess/provider tests beyond their fixed timeouts. This PR still runs that same related workload; removing LCOV instrumentation does not make the concurrency invariant obsolete.

Please retain the shared worker cap on the plain related command and assert it alongside the absence of coverage flags.

PR #1886, opened 2026-08-20
fix(maestro): refresh targets after transient selector misses

thymikee:
## Summary

Maestro tap targets now recover when atomic iOS selector dispatch races a changing accessibility tree and the nested public click reports a structured selector-not-found result.

Before, the adapter refreshed geometry for runner-native ELEMENT_NOT_FOUND, ELEMENT_OFFSCREEN, and AMBIGUOUS_MATCH errors, but missed the equivalent shared tree-resolution outcome because it arrived as generic COMMAND_FAILED text. The shared interaction boundary now carries the existing selector_not_found reason, and Maestro uses that typed reason to re-resolve against a fresh snapshot. Unrelated COMMAND_FAILED responses remain terminal.

Standalone command guidance now distinguishes settle/wait behavior, and text-entry help shows how to disambiguate a caption and input with:

    agent-device fill 'label="Email" editable=true' "qa@example.com" --settle

## Validation

A deterministic adapter regression reproduces the reported Selector did not match: id="profile-button" response, verifies a fresh snapshot and coordinate dispatch, and went red before the implementation change. A paired negative regression proves an unrelated COMMAND_FAILED response is not retried.

At commit 2b2ee58df, the required chained `pnpm check:affected --run` passed before push: 4,483 related tests, 100% changed-line and changed-branch coverage, Maestro conformance, provider integration, contracts, package verification, typecheck, lint, layering, and all other locally runnable selected checks.

This touches 13 files. Scope remains within the shared interaction error contract, Maestro tap adapter, regression tests, and matching CLI/website documentation.

Live iOS device/simulator reproduction against the reporters app was not available locally, so this PR remains draft. GitHub iOS smoke and conformance differential lanes are authoritative for the published head.

### DIFF
--- a/packages/contracts/src/facades/interaction.ts
+++ b/packages/contracts/src/facades/interaction.ts
@@ -72,6 +72,7 @@ export {
   INTERACTION_GUARANTEES,
   INTERACTION_PATH_IDS,
 } from '../interaction-guarantees.ts';
+export { INTERACTION_ERROR_REASONS } from '../interaction-error.ts';
 export type {
   GuaranteeEnforcement,
   InteractionGuarantee,
--- a/packages/contracts/src/interaction-error.ts
+++ b/packages/contracts/src/interaction-error.ts
@@ -1,0 +1,4 @@
+/** Machine-readable `error.details.reason` values shared by interaction producers and adapters. */
+export const INTERACTION_ERROR_REASONS = {
+  selectorNotFound: 'selector_not_found',
+} as const;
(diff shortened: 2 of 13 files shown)


thymikee:
Reviewed at `2b2ee58dff8106188fd0da887d7697460ced9e42` — no code findings. The new typed `selector_not_found` reason fixes the classification at the shared interaction boundary, and Maestro’s existing fresh-resolution fallback consumes that signal without message matching or retrying unrelated `COMMAND_FAILED` failures. The positive and negative adapter regressions are appropriately scoped. All exact-head checks are green.

This is ready for human review, but should remain draft until the reported iOS selector-race path is reproduced against the reporter app (or that residual device-verification gap is explicitly accepted); CI smoke alone does not make this device-facing change merge-ready.

PR #1889, opened 2026-08-20
fix(maestro): restore conformance invariants

thymikee:
## Summary

- Restore the sealed Maestro corpus to its reviewed bytes and protect upstream files with per-PR SHA-256 verification plus formatter ignores.
- Match iOS Maestro selectors against the semantic provider snapshot while the iOS presentation builder publishes the exact presented representatives for every semantic source index.
- Use one presented-visible candidate universe for ranking, geometry, and atomic-dispatch safety; remove downstream ancestor scans and fuzzy semantic-identity binding.
- Keep Maestro evidence refs in semantic snapshot space intentionally, including when interactive presentation suppresses earlier nodes.

## Validation

- `pnpm check:affected --run` — all runnable checks passed; 2,344 affected tests passed and changed-line coverage was 100%.
- `pnpm maestro:conformance` — 47 tests passed, including the new Java-free corpus SHA gate.
- Exact presentation mapping, one-universe dispatch, representative geometry, suppressed-node matching, and semantic-ref regressions were observed red against the reviewed head before the fix and pass on this head.
- A planted edit to an upstream corpus YAML failed the new SHA gate with the exact file and digest mismatch; restoring the reviewed byte returned the gate to green.

GitHub remains authoritative for native, device, differential, and regenerate-verify lanes.


thymikee:
Code review is clean at `faf7dde32`. The resolver now matches Maestro selectors against the semantic provider snapshot while deriving iOS visibility, geometry, and atomic-dispatch safety from the canonical interactive presentation, as ADR 0015 requires. The inert-title regression covers the real daemon observation route and fails under the former interactive-only matching. Exact-head Maestro Conformance Oracle and Replay-Compat Provenance checks support the sealed corpus restoration; npm unpacked size is down 4.1 kB.

The only red check is iOS Smoke. It failed before any Maestro operation: generic `wait text "Agent Device Tester"` saw zero readable captures and reported `wait_capture_stalled`. That is runner/fixture infrastructure-shaped, not evidence against this change. Rerun iOS before merge; scheduled iOS differential/regeneration remain the residual device/provenance evidence.

thymikee:
Addressed on the exact PR head: the iOS Smoke rerun passed. The previously failing Settings replay step completed successfully in 2m20s, and the fixture-backed iOS simulator E2E smoke also passed in 2m49s. All PR checks are now green; no code change was warranted because the original zero-capture failure did not reproduce.

thymikee:
Re-reviewed unchanged head `faf7dde32`: the code-clean verdict stands. The iOS Smoke rerun passed, confirming the earlier zero-readable-capture stall was transient infrastructure. All exact-head checks are now green, and GitHub reports the non-draft branch mergeable/CLEAN. No blocking review findings remain; scheduled iOS differential/regeneration remain the documented residual device/provenance evidence.

thymikee:
Addressed all seven findings in c0129b6a9:

1. The iOS presentation builder now publishes `presentedIndexesBySourceIndex` for every semantic source index, with rule-owned representative associations and empty mappings for pure noise.
2. The downstream `haveSharedSemanticIdentity` heuristic and ancestor scans are deleted.
3. iOS target resolution now matches once and ranks one presented-visible candidate universe; atomic-dispatch counting uses that same ranked set.
4. The optional visibility override was removed from `rankMaestroCandidates`; matching, platform visibility, and ranking now have explicit seams.
5. Geometry and atomic safety use the same exact presented representative.
6. Evidence refs intentionally remain in semantic snapshot space and are covered by a suppression case where presented and semantic numbering differ.
7. Normal per-PR conformance now verifies upstream corpus SHA-256 values from the manifest, and `.prettierignore` mirrors the sealed-input exclusions.

The new presentation, candidate-universe, ref-space, and corpus checks were observed red before their fixes. `pnpm check:affected --run` is green on the final tree (2,344 affected tests; 100% changed-line coverage), as is the 47-test Maestro conformance suite.

Issue #1481, opened 2026-07-29
bench: falsification fixtures for oracles + typed runner outcomes


thymikee:
Carved out of #1406 (closed as superseded — see its closing comment for what landed elsewhere and what was dropped). These are the two surviving work items, both deterministic PR-time quality improvements to the help-conformance bench, which is now the repo's single non-gating small-model oracle (SkillGym was removed in the #1412 descope).

## 1. Falsification fixtures for benchmark oracles

Every important oracle must demonstrate it can fail. Extend case/expectation definitions so each named expectation carries:

- a minimal passing witness;
- at least one known-bad counterexample;
- where useful, a metamorphic variant that changes irrelevant nouns/values.

Add a completeness test that rejects a named expectation without falsification fixtures. Include regressions for swallowed lifecycle commands, unsupported flags/selectors, pseudo refs, shell operators, and invalid positional ordering.

## 2. Typed runner outcomes — infrastructure errors cannot enter model scoring

Replace raw-string success inference with a discriminated outcome:

```ts
type RunnerOutcome =
  | { kind: 'success'; raw: string; commands: string[] }
  | { kind: 'runner-error'; raw: string; message: string; reason: RunnerErrorReason };
```

- Only `success` outcomes may reach command validation and scoring.
- A result must not contain both `runnerError` and model-validation failures.
- All-error aggregates render `N/A`, not `0/0 (0%)`.
- Keep single-trial mode for exploration, but do not label it stability evidence. (The `--evidence` multi-trial reporting mode from #1406 is optional follow-up, on demand — not part of this issue.)

## Acceptance criteria

- [ ] Every named/important benchmark expectation has a passing witness and known-bad counterexample enforced by a completeness test.
- [ ] Runner/API errors are structurally separate from model-plan failures and cannot affect pass-rate denominators.
- [ ] All-infrastructure-error reports show `N/A`.
- [ ] Deterministic tests only — no paid model calls in any gate.

## Non-goals (decided in the #1412 descope)

- No merge-base file-size ratchet (dropped with #1406).
- No scheduled/paid model runs (#1427 closed not-planned; the bench runs on demand via `pnpm bench:help-conformance`).
- No duplication of command grammar inside the benchmark — the production positional contract landed in #1433; the bench derives from `CommandSchema`.

Related: #1406 (superseded parent), #1433 (production positional contract, landed), #1411 (renderer-pinned bench gates, landed).

Issue #1406, opened 2026-07-26
Guard help-conformance correctness at production and CI boundaries


thymikee:
## Context

PR #1404 and its review fixes exposed a recurring class of false-green paths around agent-facing help:

- The help-conformance benchmark validates positional arity, while the production CLI can still silently discard extra positionals.
- A benchmark scorer can exist and pass without proving that it rejects the bad plan that motivated it.
- Runner/API failures are reconstructed from raw strings and can be reported alongside model-validation failures, making infrastructure instability look like model quality.
- Changes to the help-conformance harness fall through to an overly broad affected-check set, including paid/stochastic SkillGym work.
- The main harnesses have grown beyond the repository's context-safety guidance, but that guidance is not ratcheted.

These are different symptoms of the same gap: important claims are benchmark-local conventions instead of fail-closed contracts at the boundary that owns them.

Concrete production examples today include:

```text
snapshot ignored.json   # extra positional is ignored
close first second      # `second` is ignored
open app url close      # `close` is swallowed instead of being a lifecycle command
```

## Goal

Make it difficult to ship another help/benchmark false green:

1. Production and benchmark parsing share one positional contract.
2. Every important oracle demonstrates that it can fail.
3. Infrastructure failures cannot enter model scoring.
4. Deterministic help-conformance checks have explicit affected-check ownership.
5. Guidance harnesses remain cheap enough for an agent to read and change safely.

### 1. Enforce positional bounds at the production CLI boundary

- Add one shared positional-arity assertion derived from `CommandSchema`.
- Apply it to normal CLI parsing and legacy batch positionals.
- Require an explicit schema trait for any command that intentionally accepts additional positionals.
- Add an exhaustive test that supplies `max + 1` positionals to every bounded command and expects rejection.
- Remove the benchmark's private arity implementation once the production contract is reusable.
- Update CLI help/docs and add a behavioral SkillGym case if rejecting ignored input changes the user-visible contract.

Relevant seams:

- `src/cli/parser/args.ts`
- `src/commands/cli-grammar/registry.ts`
- command readers such as `src/commands/management/app.ts`
- `scripts/help-conformance-command-validator.ts`

### 2. Require falsification fixtures for benchmark oracles

Extend case/expectation definitions so each important oracle carries:

- a minimal passing witness;
- at least one known-bad counterexample;
- where useful, a metamorphic variant that changes irrelevant nouns/values.

[...]

PR #1893, opened 2026-08-20
test(bench): falsification fixtures for oracles + typed runner outcomes

thymikee:
## Summary

Implements #1481 — the two surviving deterministic PR-time quality items carved out of #1406 for the help-conformance bench (the repo's single non-gating small-model planning oracle):

**1. Falsification fixtures for benchmark oracles**
- Every named expectation in `EXPECTATION_SCORERS` (`scripts/help-conformance-case-checks.mjs`, exported as `KNOWN_EXPECTATIONS`) now has a fixture entry in the new `scripts/__tests__/help-conformance-expectation-fixtures.ts`: a minimal passing witness, at least one known-bad counterexample, and a metamorphic variant where useful (a second passing plan built from different domain nouns, proving the scorer keys on plan shape rather than memorizing content).
- `scripts/__tests__/help-conformance-expectation-falsification.test.ts` is the completeness gate (same "what enumerates N" shape as the existing topic-coverage/error-recovery-coverage gates): a new named expectation with no fixture fails it, and a stale fixture for a removed expectation also fails.
- Fixtures run through the *real* `validatePlanCommands`/`scoreExpectations` production pipeline, not a reimplementation, so they can't drift from what the bench actually does with a model's plan.
- Regressions included per the issue: swallowed lifecycle command prefixes (bare `open`/`press` instead of `agent-device open`/`agent-device press`), unsupported flags (`wait --selector`) and selector combinations (`is role=... label=...`), pseudo refs (`@<search-ref>`), shell operators (`| tee`), and invalid positional ordering (`get @e5 text` instead of `get text @e5`).

**2. Typed runner outcomes**
- `scripts/help-conformance-runner-output.mjs` now exports `classifyRunnerOutput`, returning a discriminated `RunnerOutcome` (declared in the new companion `scripts/help-conformance-runner-output.d.mts`, following the same `.mjs`+`.d.mts` pairing already used for `help-conformance-sample-outputs.mjs`):
  ```ts
  type RunnerOutcome =
    | { kind: 'success'; raw: string; commands: string[] }
    | { kind: 'runner-error'; raw: string; message: string; reason: RunnerErrorReason };
  ```
  instead of the old raw-string `detectRunnerError(): string | undefined` inference.
- `runCase` in `help-conformance-bench.mjs` now only calls `validatePlanCommands`/`scoreExpectations` for a `'success'` outcome; a `'runner-error'` result carries `runnerError`/`runnerErrorReason` and nothing else — a result can no longer carry both `runnerError` and model-validation checks.
- `summarizeResults` (`help-conformance-summary.mjs`) now reports `passRate: null` for a group with zero evaluated trials (every trial was a runner error), and the CLI's aggregate-stability print renders that as `N/A` instead of the misleading `0/0 (0%)`.
- Single-trial mode is unchanged (the `Aggregate stability` summary already only prints when `--repeat > 1`), so it still isn't mislabeled as stability evidence.

No paid model calls in either gate — everything runs through the local command-validator subprocess deterministically.

## Test plan
- [x] `pnpm typecheck` — clean
- [x] `pnpm lint` (oxlint --deny-warnings) — clean
- [x] `pnpm check:fallow` (fallow audit) — clean (0 issues in changed files)
- [x] `pnpm format:check` — clean
- [x] `npx vitest run scripts/__tests__/help-conformance-*` — 100/100 passing (bench, sample-outputs, topic-coverage, error-recovery-coverage, and the new expectation-falsification suite)
- [x] Full `pnpm test:unit` (unit-core + subprocess-stub, ~7200 tests) — all green; two pre-existing `&&`-chain plan-validator tests hit the 5s timeout only under full-suite CPU contention and pass cleanly in isolation (unrelated to this change)
- [x] `node scripts/help-conformance-bench.mjs --dry-run` smoke-tested end to end with an override doc

### DIFF
--- a/.fallowrc.json
+++ b/.fallowrc.json
@@ -42,7 +42,8 @@
     "scripts/maestro-conformance/corpus/**",
     "apple/runner/AgentDeviceRunner/AgentDeviceRunnerUITests.xctestplan",
     "scripts/write-xcuitest-cache-metadata.mjs",
-    "scripts/help-conformance-sample-outputs.d.mts"
+    "scripts/help-conformance-sample-outputs.d.mts",
+    "scripts/help-conformance-runner-output.d.mts"
   ],
   "ignoreDependencies": [
     "@theme",
--- a/scripts/__tests__/help-conformance-bench.test.ts
+++ b/scripts/__tests__/help-conformance-bench.test.ts
@@ -9,7 +9,8 @@ import { scoreExpectations } from '../help-conformance-case-checks.mjs';
 import { validateAgentDeviceCommand } from '../help-conformance-command-validator.ts';
 import { opensAndCloses, usesValidationPrep } from '../help-conformance-expectations.mjs';
 import { validatePlanCommands } from '../help-conformance-plan-validator.mjs';
-import { detectRunnerError, extractCommands } from '../help-conformance-runner-output.mjs';
+import { classifyRunnerOutput, extractCommands } from '../help-conformance-runner-output.mjs';
+import type { RunnerOutcome } from '../help-conformance-runner-output.mjs';
 import { summarizeResults } from '../help-conformance-summary.mjs';
 
 const execFileAsync = promisify(execFile);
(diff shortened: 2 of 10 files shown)


thymikee:
Reviewed exact head `4e4fe2517d64901b2124b98e69f697c52d1a9396`: code-clean. The falsification fixtures are non-vacuous and exercise the real validator/scorer pipeline; the registry-derived completeness check catches missing and stale fixtures, and the typed outcome branch keeps infrastructure errors out of model scoring. Exact-head CI is green and size is unchanged.

Before merge, please retitle the body section to `## Validation`, note the 10-file/no-expansion scope, and add `Closes #1481` so the PR metadata matches the repository template.

PR #1898, opened 2026-08-20
fix(android): warn when recording readiness inspection fails open

### LINKED ISSUE #1895: fix(android): warn when recording readiness inspection fails open
## Problem

Android recording-session dialog recovery intentionally fails open when its helper-backed snapshot inspection is unavailable. The command then succeeds, but the inspection failure is recorded only in diagnostics. An operator can see a clean success such as `Typed 5 chars` even though the snapshot subsystem was unavailable and agent-device could not establish whether a blocking Android dialog needed recovery.

This is a wrong-confidence response: continuing is appropriate, but the successful response must disclose that readiness could not be inspected.

## Current behavior

- `recoverAndroidBlockingSystemDialog` catches snapshot failures in `src/daemon/android-system-dialog.ts`, emits `android_blocking_dialog_inspection_failed`, and returns `{ status: 'unknown', reason: 'inspection-failed' }`. The normalized underlying error and its recovery hint do not survive in the typed result.
- Recording-session generic commands in `ensureGenericCommandReady` (`src/daemon/request-generic-dispatch.ts`) treat every result except `status: 'failed'` as ready and discard `unknown`.
- `type` has the same fail-open behavior in `recoverAndroidRecordingDialogForType` (`src/daemon/handlers/interaction.ts`).
- `src/daemon/__tests__/request-router-android-modal.test.ts` proves that a generic Android gesture continues when inspection throws, but does not require a response warning.
- A missing source-checkout snapshot helper already has the correct actionable hint in `androidSnapshotHelperUnavailableError` (`src/platforms/android/snapshot.ts`): run `pnpm build:android`. The readiness boundary currently loses it.

## Required behavior

1. Keep inspection failure fail-open. Do not fail or skip the requested device action solely because readiness inspection failed.
2. Preserve a bounded, user-facing warning from the normalized inspection error. It must explain that Android blocking-dialog readiness could not be inspected, that the command continued, and retain an actionable upstream hint when present (notably `pnpm build:android` for a missing helper artifact).
3. Carry the warning through the typed `AndroidBlockingDialogRecoveryResult`; do not recover it later by sniffing error text or rereading diagnostics.
4. Append the warning through the existing response-building/composition path. Never clobber an existing warning from settle, stale refs, recovery, or another response source.
5. Cover both recording-session entry points that currently call `recoverAndroidBlockingSystemDialog`: generic commands and `type`.
6. Keep provider-device behavior, non-Android behavior, sessions without screen recording, successful recovery, and hard `status: 'failed'` behavior unchanged.

Suggested result shape (adjust naming to the owning module's vocabulary if needed):

```ts
{ status: 'unknown'; reason: 'inspection-failed'; warning: string }
```

The warning must be derived where the error is caught, while the structured `normalizeError` result and hint are still available.

## Tests

- Extend the generic-router inspection-failure regression in `src/daemon/__tests__/request-router-android-modal.test.ts` to assert:
  - the platform command still dispatches;
  - the response is successful;
  - `warning` states that readiness inspection failed and the command continued;
  - an upstream actionable hint is retained;
  - a pre-existing response warning is appended rather than overwritten.
- Add the equivalent regression for the recording-session `type` path at its owning handler test location.
- Keep or add a focused unit test for `recoverAndroidBlockingSystemDialog` proving an inspection exception becomes the typed unknown result with a bounded warning.
- Prove the new regression red against current `origin/main` and quote the failing test count/assertion in the PR.

## Scope boundaries

In scope:

- `src/daemon/android-system-dialog.ts`
- the two response-composition call sites above
- focused owning tests
- CHANGELOG entry because this changes successful user-visible output

Out of scope:

- changing whether recovery fails open
- adding the Android snapshot-helper check to `doctor`
- changing helper artifact discovery/build behavior
- adding new recovery fixtures or staging ANR dialogs
- changing ADR 0011 interaction-guarantee cells; this is readiness/response composition, not a new dispatch path
- editing `skills/**`

## Validation

Run focused affected tests while editing, then:

```sh
pnpm check:affected --run
```

If the response warning requires a new wire field, stop: `warning` already exists and should be reused. If preserving the upstream reason would require exposing unbounded diagnostic/error text, stop and narrow it at the catch boundary instead.

## Done when

- A missing or unusable Android snapshot helper can no longer produce an unqualified successful recording-session command response.
- The requested command still runs.
- CLI/JSON consumers receive the composed warning with the actionable recovery hint.
- Existing warnings remain intact.
- Focused regressions are demonstrated red before the fix and green after it.

thymikee:
## Summary

Closes #1895

Android recording-session generic commands and `type` now continue when blocking-dialog readiness inspection fails, while returning a bounded warning with the normalized inspection error and actionable hint. Existing response warnings are appended rather than replaced. The warning is carried from the typed recovery result through both generic and `type` response paths.

## Validation

- The focused regression suite was red against origin/main: 3 failed, 2 passed.
- The focused regression suite is green after the fix: 3 files, 5 tests passed.
- The updated truncation test uses multiline, over-limit message and hint inputs and asserts whitespace normalization plus exact 240-character bounded fields ending in an ellipsis.
- `pnpm check:affected --run` passed for the implementation; the review-only test change also passes `pnpm check:fallow --base origin/main` and the focused suite.
- Exact-head live Android verification at `17a913bb6` used the `com.callstack.agentdevicelab` test app on emulator `emulator-5554` with `androidSnapshot.backend=android-helper` before the run and an app-scoped `adb screenrecord`.
- With the bundled snapshot helper temporarily unavailable, generic `home` returned `success: true` with the readiness warning and `pnpm build:android` hint; `type` also returned `success: true`, typed 6 characters, and returned the same composed warning.

CHANGELOG.md was updated. No skills or ADR guidance changed because the issue scope is runtime behavior and changelog coverage.


thymikee:
Reviewed exact head `ceb6d0ff9a630dfc2d1342bfce942d1ca4fe17f2`.

The bounded-warning regression is currently vacuous: both the message and hint are below the 240-character limit, and `warning.length < 800` still passes if `boundAndroidWarningText` is removed. Please drive this test with over-limit, multiline message and hint values and assert whitespace normalization plus truncation/ellipsis at the declared ceiling, so reverting the bound makes the test red. Keep the short actionable `pnpm build:android` hint assertion separately if useful.

Before merge readiness, also attach exact-head live Android evidence from a recording session with readiness inspection deliberately unavailable, proving both a generic command and `type` still execute and return the composed actionable warning. The PR currently states that no live Android verification was run.

thymikee:
Addressed in `17a913bb6`.

- The bounded-warning regression now uses multiline, over-limit message and hint values, rejects tabs/newlines/repeated spaces, and requires each bounded field to be exactly 239 characters plus an ellipsis.
- Exact-head live Android verification on `emulator-5554` completed in an app-scoped `adb screenrecord`: generic `home` returned success with the readiness warning and `pnpm build:android` hint; `type` returned success and typed 6 characters with the same warning.
- The snapshot helper was restored and the verification session was closed cleanly.

thymikee:
Re-reviewed exact head `17a913bb6e8488b1d776c5e9fb7bcbdbea4ad03d`: clean. The bounded-warning regression now uses over-limit multiline inputs and proves normalization, truncation, and the declared ceiling; the exact-head Android recording evidence exercises both generic and `type` fail-open routes with inspection unavailable. All 28 checks are green, including Android/iOS/Linux/macOS smoke, and GitHub reports the branch clean and mergeable. Ready for human review and merge.

PR #1899, opened 2026-08-20
fix(maestro): make iOS presentation correspondence authoritative

thymikee:
## Summary

Make canonical iOS presentation correspondence authoritative for Maestro visibility and tap geometry.

- select semantic identity and authored index in provider snapshot space, even when the semantic node has no bounds
- take interactive geometry only from a visible canonical representative that established visibility
- make suppression and representative declaration one construction operation so the correspondence map cannot drift
- clarify ADR 0015 as a present-tense snapshot: agent-device supports selected Maestro Flow syntax and behavior

## Validation

Pre-fix focused tests failed in both targeted cases:

- `iOS target resolution uses presented bounds when the semantic match has none`: expected a successful resolution, received `{ ok: false }`
- `iOS target resolution takes geometry from the representative that proved visibility`: expected `y: 100`, received the off-screen representative at `y: 900`

Review-fix validation:

- focused Maestro and iOS presentation suite: 9 files / 73 tests passed
- `pnpm run check:fallow --base origin/main`: no issues
- `pnpm check:affected --run`: 2,382/2,383 tests passed; the unrelated Apple platform output guard timed out while the fake `record` harness was under full-suite load
- isolated `apple-platform-output-guard.test.ts`: 3/3 passed, including the timed-out iOS case

No manual device run: the fixture app does not expose a semantic-node-without-bounds or multi-representative hierarchy. GitHub iOS and conformance lanes remain authoritative for native/device coverage.

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -300,9 +300,9 @@ task touches:
 
 ### Maestro compatibility
 
-- Maestro program: source-preserving typed representation of supported Maestro YAML. It is
-  interpreted directly through the compatibility runtime port and never lowered through generic
-  replay action strings. See ADR 0015.
+- Maestro program: source-preserving typed representation of the Maestro Flow syntax and behavior
+  supported by agent-device. The program is interpreted directly through the compatibility runtime
+  port and never lowered through generic replay action strings. See ADR 0015.
 - Maestro observation generation: explicit compatibility-engine state identifying evidence captured
   since the most recent mutation. Queries may share semantic evidence within one generation; every
   mutation attempt invalidates it before dispatch. Interaction geometry is action-local: unique exact
--- a/docs/adr/0015-direct-maestro-engine.md
+++ b/docs/adr/0015-direct-maestro-engine.md
@@ -190,6 +190,8 @@ two-pointer plans, executor selection, and app-observable effects must remain un
 
 - agent-device supports selected Maestro Flow syntax and behavior. Unsupported features return
   explicit errors, and intentional differences are declared in the conformance fixtures.
+- Shipping two production engines or a runtime fallback between them is rejected because it doubles
+  semantic and performance ownership.
 - Source provenance and runtime values stay typed through execution.
 - Compatibility policy remains local while device behavior stays in shared runtimes and backends.
 - Cross-platform correctness may require richer provider query evidence, but not additional round trips.
--- a/docs/adr/0015-direct-maestro-engine.md
+++ b/docs/adr/0015-direct-maestro-engine.md
@@ -167,8 +167,8 @@ vector).
 
 ## Performance Contract
 
-The migration cannot switch production routing until Android and iOS satisfy all of these on the pager
-and react-navigation corpora:
+The direct engine preserves these properties on Android and iOS across the pager and react-navigation
+corpora:
 
 - total wall time is no slower than the pre-migration compatibility engine;
 - an observation immediately after a mutation polls its authored condition directly; a later mutation uses
@@ -186,30 +186,10 @@ Android verification must prove the bundled helper backend and version. iOS veri
 runner startup from warm command latency. Both platforms rerun the non-Maestro gesture canaries; their
 two-pointer plans, executor selection, and app-observable effects must remain unchanged.
 
-## Migration
-
-Completed. Production Maestro YAML now uses the typed program, immutable replay plan, and direct
-runtime port exclusively. The `SessionAction` lowering path, private `__maestro*` commands,
-`replayControl`, hidden compatibility caches, positional decoders, and their fallback routing were
-deleted in the same change. Generic `.ad` replay remains independent.
-
-1. Add the typed program, runtime port, direct interpreter, and normalized upstream fixtures without
-   changing production routing.
-2. Differentially compare the old lowering path and direct engine at the typed operation boundary.
-3. Move lifecycle, input, screenshot, and keyboard commands.
-4. Move target queries and assertions, deleting unverified fast-path success and assertion-triggered
-   action replay.
-5. Move single-pointer swipes through ADR 0013's normalized input boundary.
-6. Move hooks, includes, conditions, repeat/retry, and trusted `runScript`.
-7. Switch `--maestro` atomically, then delete private Maestro commands, positional decoding, hidden
-   caches, and obsolete converter/runtime tests.
-
-The old and new engines may coexist only in tests during migration. Shipping two production engines or
-a runtime fallback between them is rejected because it doubles semantic and performance ownership.
-
 ## Consequences
 
-- Maestro remains a supported subset with explicit failures; this refactor does not expand parity.
+- agent-device supports selected Maestro Flow syntax and behavior. Unsupported features return
+  explicit errors, and intentional differences are declared in the conformance fixtures.
 - Source provenance and runtime values stay typed through execution.
 - Compatibility policy remains local while device behavior stays in shared runtimes and backends.
 - Cross-platform correctness may require richer provider query evidence, but not additional round trips.
--- a/packages/maestro/src/internal/__tests__/runtime-targets-typed.test.ts
+++ b/packages/maestro/src/internal/__tests__/runtime-targets-typed.test.ts
@@ -313,48 +313,59 @@ test('iOS target resolution uses one presented candidate universe for geometry a
   });
 });
 
-test('iOS target resolution uses presented bounds when the semantic match has none', () => {
-  const semanticSnapshot = makeSnapshot([
-    {
-      index: 0,
-      type: 'Application',
-      rect: { x: 0, y: 0, width: 393, height: 852 },
-    },
-    {
-      index: 1,
-      parentIndex: 0,
-      type: 'Other',
-      identifier: 'email-input',
-    },
-  ]);
-  const presentationSnapshot = makeSnapshot([
-    semanticSnapshot.nodes[0]!,
-    {
-      index: 1,
-      parentIndex: 0,
-      type: 'TextField',
-      rect: { x: 20, y: 100, width: 240, height: 44 },
-    },
-  ]);
-
-  expect(
-    resolveMaestroTargetFromSnapshot(semanticSnapshot, { selector: { id: 'email-input' } }, 'ios', {
-      interactiveBounds: true,
-      presentation: {
-        snapshot: presentationSnapshot,
-        presentedIndexesBySourceIndex: new Map([
-          [0, [0]],
-          [1, [1]],
-        ]),
+test.each([
+  { mode: 'interactive', interactiveBounds: true },
+  { mode: 'non-interactive', interactiveBounds: false },
+])(
+  'iOS $mode target resolution uses presented bounds when the semantic match has none',
+  (options) => {
+    const semanticSnapshot = makeSnapshot([
+      {
+        index: 0,
+        type: 'Application',
+        rect: { x: 0, y: 0, width: 393, height: 852 },
       },
-    }),
-  ).toMatchObject({
-    ok: true,
-    node: { index: 1, identifier: 'email-input' },
-    rect: { x: 20, y: 100, width: 240, height: 44 },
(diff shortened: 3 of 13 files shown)


thymikee:

The two behavior fixes are correct and worth landing: selecting the authored index in **semantic** space (so a bounds-less semantic node still resolves), and taking geometry only from a representative that **proved** visibility. The tests pin both precisely.

The presentation half is the best part of this PR. Collapsing `suppressedIndexes.add(x)` + `associateSnapshotPresentation(x, rep)` into a single `context.suppressNode(source, representatives)` and narrowing `suppressedIndexes` to `ReadonlySet` makes the correspondence drift *structurally unrepresentable* instead of merely tested against. That is exactly the right move.

The `runtime-targets.ts` half goes the other way. It grows **179 → 264 lines (+85)** and adds four functions plus a type alias, and nearly all of that new surface exists to carry accidental complexity that can simply be deleted.

---

### 1. The three-function chain is preserving complexity, not decomposing it

`resolveMaestroTargetFromSnapshot` → `resolveRankedMaestroTarget` → `resolveSelectedMaestroTarget` threads the same values down three levels, ending in a **7-field params object** that is just "every local of the original function". That is a split by line count, not by concept. Underneath it sit four things that shouldn't exist:

**a. A duplicate of a canonical exported type.** `runtime-target-ranking.ts:13` already exports `MaestroRankedCandidates` with exactly this shape. This PR declares a second, colliding one:

```ts
type MaestroRankedCandidates = ReturnType<typeof rankMaestroCandidates>;
```

Import the canonical type and annotate `rankPresentedMaestroCandidates(): MaestroRankedCandidates` — which it currently lacks entirely.

**b. A single-element array used as a rect predicate.**

```ts
const semanticTarget = selectMaestroSnapshotMatch([sourceTarget], 0);
```

This wraps a node in an array to reuse a *list selector* as "does this node have a usable rect". That's magical. The honest expression is `usableRect(node): Rect | undefined`, which belongs next to the existing private `hasUsableRect` in `runtime-target-ranking.ts` — the canonical home for this concept.

**c. A now-dead branch.** The PR correctly hoists the iOS gate to the top (`platform === 'ios' && options.presentation`), which makes `presentationUsed ⟹ platform === 'ios'` an invariant. But `countAtomicDispatchCandidates` still checks `params.platform !== 'ios'`, and `platform` is threaded through two functions to reach it. The PR establishes the invariant and then doesn't harvest it.

**d. A shape whose `node` is never read.** `countInteractionDispatchCandidates(target: { node; rect } | null, …)` only ever touches `target.rect`. The parameter should be `Rect | undefined`.

Fix those four and `resolveSelectedMaestroTarget`, `resolveRankedMaestroTarget`, `selectTargetRect`, `countAtomicDispatchCandidates` and the local type alias all have nothing left to do. Both fixes fit in the original single function:

```ts
export function resolveMaestroTargetFromSnapshot(
  snapshot: SnapshotState,
  query: MaestroTargetQuery,
  platform: MaestroPlatform,
  options: { interactiveBounds?: boolean; presentation?: MaestroInteractivePresentation } = {},
): MaestroTargetResolution {
  const presentedNodes =
    platform === 'ios' && options.presentation
      ? createPresentedNodeLookup(options.presentation)
      : undefined;
  const { matches, ranked, parentMatched } = presentedNodes
    ? rankPresentedMaestroCandidates(snapshot, query, presentedNodes)
    : rankMaestroCandidates(snapshot, query.selector, platform, query.childOf);
  if (!parentMatched) {
    return {
      ok: false,
      message: 'Maestro childOf parent did not match.',
      evidence: buildMaestroTargetEvidence(query, matches, [], undefined),
    };
  }
  const target = presentedNodes
    ? selectMaestroCandidate(ranked, query.index)
    : selectMaestroSnapshotMatch(ranked, query.index)?.node;
  const evidence = buildMaestroTargetEvidence(query, matches, ranked, target);
  if (!target) return failedTargetResolution(query, matches, ranked, evidence);

  const presentedTarget = presentedNodes
    ? selectMaestroSnapshotMatch(presentedNodes.visibleForSource(target), undefined)
    : null;
  const semanticRect = usableRect(target);
  const rect = options.interactiveBounds
    ? (presentedTarget?.rect ?? semanticRect)
    : (semanticRect ?? presentedTarget?.rect);
  if (!rect) return failedTargetResolution(query, matches, ranked, evidence);

  return {
    ok: true,
    node: target,
    rect,
    matches: ranked.length,
    dispatchCandidates:
      presentedNodes && query.allowAtomicSelectorDispatch && !query.childOf
        ? countInteractionDispatchCandidates(semanticRect, ranked, presentedTarget)
        : 0,
    evidence,
  };
}
```

I built this rather than asserting it. **`179 → 194 lines (+15)` instead of `+85`**, one new helper (`selectMaestroCandidate`) instead of five new constructs, and `selectTargetRect`'s asymmetry is legible inline. Verified on this branch:

- `tsc -b <all packages>` and `tsc -p tsconfig.json` — clean
- `vitest run --project unit-core src/daemon src/compat packages/maestro` — **317 files / 2269 tests pass**, including both new tests from this PR

### 2. `scroll.ts` still uses the new API as two operations — and it's the last thing keeping `suppressedIndexes` on the context

The PR's stated goal is "make suppression and representative declaration one construction operation so the correspondence map cannot drift." This call site is the one place that still does the old two-step, now with a *doubled* `suppressNode` call whose second argument contradicts the first:

```ts
if (!isScrollableSnapshotType(node.type)) context.suppressNode(node, []);
// …two early returns later…
if (context.suppressedIndexes.has(node.index)) context.suppressNode(node, [container]);
```

Hoisting `container` makes it a single atomic declaration with identical semantics in all five cases:

```ts
const suppressed = !isScrollableSnapshotType(node.type) || context.isSuppressed(node);
const directions = inferVerticalScrollIndicatorDirections(node.label?.trim() ?? '', node.value);
const container = directions
  ? findNearestScrollableContainer(node, byIndex, { includeSelf: true })
  : undefined;

if (suppressed) context.suppressNode(node, container ? [container] : []);
if (container && directions) applyScrollIndicatorReplacement(context, container, node, directions);
```

The payoff: `suppressedIndexes` has exactly **two** readers left (`noise.ts:140`, `scroll.ts:50`), both of which want the question *"is this node suppressed?"*, not the mutable set. Swap both for `isSuppressed: (node) => boolean` and the raw `Set` comes off `SnapshotTreeRuleContext` entirely — suppression state becomes fully private to `buildIosInteractiveSnapshotPresentation`, which is a strictly better version of the invariant this PR is already chasing. `SnapshotTreeRuleContext` has a single implementor, so this is contained. Included in the verified run above.

### 3. Untested behavior change on the non-interactive path

`selectTargetRect` changed the non-`interactiveBounds` branch from `target.rect` to `semantic ?? presented`. That is a real new behavior — a bounds-less semantic node now resolves for `assertVisible`-style queries where it previously failed. It looks intended, but **all four presentation tests set `interactiveBounds: true`**, so the branch ships uncovered. Please add the non-interactive case.

### 4. Smaller, but worth fixing

- **Visibility is now recomputed and never memoized.** `isVisible` went from `.some(isMaestroNodeVisible)` (short-circuits) to `visibleForSource(node).length > 0`, and `visibleForSource` is then recomputed for the selected target. `isMaestroNodeVisible` rebuilds `buildSnapshotNodeMap(nodes)` on *every* call, so this is O(n) per representative, twice, in a path that `wait` polls. The full list is genuinely needed now, so the fix is a one-line memo keyed on source index inside `createPresentedNodeLookup`, not reverting to `.some()`.
- **The doc comment lost the useful half.** `presentedIndexesBySourceIndex` still maps pure noise to `[]`, but the comment now says only "Canonical presented representatives for every semantic source index." The old sentence documented the empty-list case callers must handle; keep it.
- **ADR 0015 drops a live constraint.** Deleting the completed migration checklist is fine housekeeping, but it also deleted *"Shipping two production engines or a runtime fallback between them is rejected because it doubles semantic and performance ownership."* That still governs future work and is not preserved anywhere else — "Alternatives Considered" covers different alternatives. Re-add it as one bullet there. Separately, rewriting an ADR's decision record inside a behavioral fix PR mixes scope and makes both harder to revert.

---

**Verdict: needs changes.** The behavior is right and the presentation-side abstraction is genuinely good — but `runtime-targets.ts` currently spends 85 lines and five new constructs building scaffolding around a duplicated type, a single-element-array trick, a dead platform check, and a shape whose `node` is never read. Delete those four and the scaffolding has nothing to hold up.

thymikee:
Addressed the latest review in 44ab3715b:

- collapsed the three-stage resolver back into one flow and reused the canonical `MaestroRankedCandidates` type
- added canonical `usableRect`, removed the single-element selection trick, dead platform threading, and unused target-node shape
- made suppression state private behind `isSuppressed` / `suppressNode`; scroll correspondence is now declared atomically
- added the non-interactive presented-bounds regression case
- memoized visible representatives by semantic source index
- restored the pure-noise `[]` contract in the map comment and the durable single-production-engine ADR constraint

`runtime-targets.ts` is down from 264 to 197 lines. Focused tests pass 73/73 and the fallow audit is clean. The full affected run had one unrelated fake-harness timeout at 20s (2,382/2,383 passed); that file passed 3/3 in isolation, including the iOS case in 9.7s.

thymikee:
Reviewed `44ab3715b9a8bfeff0c111668ca45d812fa58c44`: code review is clean. The presentation correspondence now has one owning mutation (`suppressNode`), and target resolution preserves Maestro identity/index selection in semantic space while taking tap geometry from a visible canonical representative. The focused tests pin both regressions; all CI is green and the size report is -286 B unpacked.

This still lacks the required live iOS Maestro evidence for the changed presentation route. The green iOS smoke job exercises runner/public-CLI coverage, not `replay <flow.yaml> --maestro`; the PR confirms the fixture cannot expose a bounds-less semantic source or multiple representatives. Please record an iOS simulator/device Maestro replay exercising that hierarchy, or explicitly retain this as an accepted residual risk, before calling the PR merge-ready.

thymikee:
Live iOS evidence is now recorded against `44ab3715b`.

Environment:
- freshly built source checkout: `pnpm build`, then `pnpm clean:daemon`
- repository fixture app built and installed on iPhone 17 Pro, iOS 26.2
- source CLI and source-built Apple runner
- public entry point: `replay <flow.yaml> --maestro`
- session: `pr1899-maestro-evidence`

The live raw XCTest hierarchy contains the exact multi-representative case from the review:

- semantic source: `id=gesture-target`, `type=Other`, `hittable=false`, rect `(34.33, 433.67, 333.33, 220.00)`
- presented child 1: `id=transform-gesture-target`, rect `(34.67, 434.00, 332.67, 219.33)`
- presented child 2: `id=two-pointer-pan-target`, rect `(34.67, 434.00, 166.33, 219.33)`
- presented child 3: `label=Gesture test image`, rect `(81.00, 463.67, 240.00, 160.00)`

Focused Maestro flow:

```yaml
appId: com.callstack.agentdevicelab
---
- assertVisible:
    id: gesture-target
- tapOn:
    id: gesture-target
- assertVisible:
    id: gesture-target
```

Result:

```json
{
  "success": true,
  "data": {
    "replayed": 3,
    "healed": 0,
    "snapshotDiagnostics": {
      "stats": {
        "count": 3,
        "p50Ms": 140,
        "p95Ms": 147,
        "maxMs": 147,
        "platform": "ios",
        "backends": { "xctest": 3 }
      }
    }
  }
}
```

Trace request `b27b4710fb88330a` records `maestro_tap_dispatch` as point `(201, 544)`, inside the visible presented representatives rather than treating the non-hittable semantic source as the interaction surface.

I also ran a second public Maestro flow through the original form ambiguity: `id=field-email` with `label=Email`, input, keyboard dismissal, scrolling, and taps/assertions for `shipping-pickup` and `payment-cash`. That completed 11/11 steps, healed 0, across 22 XCTest snapshots.

This supplies live evidence for both the exact multiple-representative correspondence and the user-facing id-plus-label path. No residual-risk exception is needed.

PR #1572, opened 2026-08-03
fix(ios): preserve timed pan duration

thymikee:
## Summary

Fixes #1562 by representing linear single-pointer gesture plans with an exact `[start, end]` sample tuple. The authored duration remains the interval between those endpoints, so iOS uses the existing XCTest bridge without a workaround-specific execution path.

Android expands endpoint plans once at `src/platforms/android/touch-plan.ts` into 16 ms linear transport samples before either provider-native or helper injection, using the shared sampler and interpolator. Two-pointer plans keep dense geometry, and endpoint-hold flings keep their existing lowering. WebDriver sends an endpoint plan as one timed W3C pointer move, relying on W3C tick interpolation. ADR 0013 documents these boundaries and the test/evidence limits.

## Validation

- Focused shared-plan, Android transport/helper/session, and WebDriver tests: 61 passed.
- `pnpm build:xcuitest:ios`: `TEST BUILD SUCCEEDED` on the iPhone 17 Pro simulator.
- `pnpm check:affected --run`: all runnable checks passed; 367 test files and 3,475 tests passed; changed-line coverage 12/12 (100%).
- Exact-head live iOS validation on iPhone 17 Pro (`6044A251-23C2-4584-B7DB-87A16B787757`) returned `executionProfile: timed-pan` for every run:

  | Requested duration | Runner uptime delta |
  | ---: | ---: |
  | 400 ms | 613.5 ms |
  | 1,200 ms | 1,410.9 ms |
  | 3,000 ms | 3,216.3 ms |

  A clean Settings snapshot reported `hiddenContentBelow: true` before a 400 ms upward pan; afterward it reported `hiddenContentAbove: true` and the first cell had moved offscreen (`y: -31`), proving observable scroll-state change.
- ADR 0013 records that unit tests cannot independently prove private XCTest event delivery; the live simulator evidence covers that seam.
- No CLI surface or skill guidance changed; only the gesture-plan ADR was updated for the new contract.

PR #1901, opened 2026-08-20
test(gesture): assert pan duration in the iOS gesture-lab replay

### LINKED ISSUE #1584: iOS gesture pan duration has no automated guard; the nightly gesture gate cannot catch a #1562 regression

thymikee:
## Summary

Closes #1584.

Adds an automatic iOS regression canary for the timed-pan bug fixed by #1572. The existing `gesture-lab.ad` replay exercised `gesture pan`, but only asserted an activation counter, so it stayed green when iOS collapsed a requested 500 ms pan to roughly 100 ms.

- The fixture app measures a recognized single-pointer pan and exposes only a bucketed result: `pan duration >=400ms` or `<400ms`.
- The measurement finishes in `onEnd`, which only fires after the recognizer reached `ACTIVE`; failed or cancelled recognizers cannot publish a passing bucket.
- A minimal `gesture-pan-duration.ad` replay asserts the bucket without depending on the fixture's unrelated multi-touch gestures.
- The iOS PR smoke lane runs this replay automatically with its standard retry policy.
- The broader `gesture-lab.ad` replay keeps the same assertion after its existing 500 ms pan.

Scope: 5 files, limited to the fixture app, replay coverage, and iOS CI wiring. No runner, runtime, protocol, Android, or public response behavior changes.

### Why a fixture canary

The runner's internal gesture timestamps are deliberately absent from public JSON responses. Exposing them solely for a regression test would expand the protocol for test infrastructure. The fixture instead observes the user-visible duration while keeping the assertion coarse enough to tolerate scheduling jitter.

### Red-before / green-after evidence

The same canary was tested against the parent of the timed-pan fix (`80feff42d6e788f597bf4e2e777f41a3e3b885d9`): the replay timed out waiting for `pan duration >=400ms`, and the live fixture reported `pan duration <400ms`. This confirms the test detects the original regression rather than an incidental launch or selector failure.

On the fixed implementation, the isolated replay passed twice locally:

```text
Run 1: passed 1, failed 0, duration 17.4s
Run 2: passed 1, failed 0, duration 6.4s
```

The previous PR head (`ca0b4026d`) also passed the new iOS smoke step in CI on its first attempt (`gesture-pan-duration.ad`, 33.7s; run `32378738304`, job `96456403877`). The current head strengthens the fixture observer so only an activated pan can satisfy the assertion; current-head CI is the final confirmation.

The pre-existing multi-touch portion of `gesture-lab.ad` was investigated separately. Its local iOS 26.2 failures reproduced on unmodified code and remain outside this PR. Android behavior is unchanged and `gesture-lab-android.ad` is untouched.

## Validation

- [x] Regression test proven red against pre-#1572 code, including the rendered `<400ms` result
- [x] Isolated duration replay passed twice locally on the fixed implementation
- [x] Previous exact head passed the new automatic iOS smoke step on its first attempt
- [x] `pnpm check:affected --run`
- [ ] Current-head GitHub CI

### DIFF
--- a/.github/workflows/ios.yml
+++ b/.github/workflows/ios.yml
@@ -209,6 +209,17 @@ jobs:
           pnpm gate build
           node --experimental-strip-types scripts/node-test-tmpdir.ts --test test/integration/smoke-ios-simulator-coverage.test.ts test/integration/smoke-ios-simulator.test.ts
 
+      # #1584: isolated, cheap automatic guard for the #1562 class of regression (iOS silently
+      # ignoring a `gesture pan` duration). Split out of gesture-lab.ad, which stays full-tier
+      # (dispatch-only via replays-manual.yml) because its multi-touch commands are a separate,
+      # heavier concern. This is the only automatic lane this assertion runs in today.
+      - name: Run gesture pan-duration smoke replay
+        run: |
+          xcrun simctl install "${{ steps.ios-simulator.outputs.simulator-udid }}" "${{ steps.fixture-app.outputs.app-path }}"
+          pnpm clean:daemon
+          node --experimental-strip-types src/bin.ts test examples/test-app/replays/gesture-pan-duration.ad --udid "${{ steps.ios-simulator.outputs.simulator-udid }}" --retries 2 --artifacts-dir test/artifacts/replays-ios-gesture-pan-duration --report-junit test/artifacts/replays-ios-gesture-pan-duration.junit.xml
+          pnpm clean:daemon
+
       - name: Assert simulator automation preserved host focus
         if: ${{ always() }}
         run: |
--- a/examples/test-app/README.md
+++ b/examples/test-app/README.md
@@ -251,6 +251,13 @@ two-pointer gesture changes all three semantic states. On Android, these checks
 are intentionally qualitative because recognizers can report non-exact centroid,
 scale, and rotation values for one simultaneous two-finger gesture.
 
+`gesture-pan-duration.ad` is a separate, minimal iOS replay that asserts a single-pointer
+`gesture pan`'s requested duration is actually observed by the app (a bucketed
+`pan duration` status on the Home screen), rather than just that the gesture activated.
+It's split out of `gesture-lab.ad` so it can run as an automatic PR-tier check
+(`.github/workflows/ios.yml`) without depending on `gesture-lab.ad`'s multi-touch commands,
+which stay full-tier only.
+
 To target a specific iOS simulator or an installed Expo development build, run the
 underlying command directly so global flags stay before replay inputs:
 
--- a/examples/test-app/replays/gesture-lab.ad
+++ b/examples/test-app/replays/gesture-lab.ad
@@ -10,6 +10,7 @@ wait "two-pointer pan activations 0" 5000
 
 gesture pan 110 443 48 0 500
 wait "two-pointer pan activations 0" 5000
+wait text "pan duration >=400ms" 5000
 
 gesture pan 110 443 48 0 500 --pointer-count 2
 wait "two-pointer pan activations 1" 5000
--- a/examples/test-app/replays/gesture-pan-duration.ad
+++ b/examples/test-app/replays/gesture-pan-duration.ad
@@ -1,0 +1,17 @@
+# Isolated pan-duration canary (#1584), split out of gesture-lab.ad so the smoke tier gets a
+# cheap automatic regression guard for #1562 without depending on gesture-lab.ad's multi-touch
+# commands (pinch/rotate/transform/two-pointer pan), which are a separate, heavier full-tier
+# concern.
+context platform=ios kind=simulator timeout=60000
+
+env APP_TARGET="Agent Device Tester"
+env APP_URL=""
+
+open "${APP_TARGET}" --relaunch --launch-url "${APP_URL}"
+wait "Gesture lab" 30000
+wait "gesture canary ready" 5000
+
+gesture pan 110 443 48 0 500
+wait text "pan duration >=400ms" 5000
+
+close
--- a/examples/test-app/src/screens/GestureLab.tsx
+++ b/examples/test-app/src/screens/GestureLab.tsx
@@ -231,7 +231,9 @@ export function GestureLab() {
       if (start === undefined) return;
       const durationMs = Date.now() - start;
       const bucket =
-        durationMs >= panDurationBucketMs ? `>=${panDurationBucketMs}ms` : `<${panDurationBucketMs}ms`;
+        durationMs >= panDurationBucketMs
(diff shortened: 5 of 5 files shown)


thymikee:
Not ready yet at `4144ce09`. The app-side duration bucket is a sensible way to observe the private iOS bridge, and the formatting-only follow-up fixed the initial lint failure, but three readiness gaps remain:

1. The assertion still cannot catch a regression automatically. `gesture-lab.ad` is reached only by `full:fixture-replays`; PR iOS runs the smoke tier, the full device suites are currently workflow-dispatch-only in `replays-manual.yml`, and scheduled `replays-nightly.yml` now runs parser fuzz only. Please put this isolated, cheap duration canary in an automatic PR or scheduled lane and update the stale “nightly full tier” body claim.
2. The required red-before proof is missing. Run this assertion against the pre-#1572 or reverted production path and quote the `<400ms` / wait failure, then restore the fix and show green.
3. The device evidence is internally inconsistent: the test plan says the whole `gesture-lab.ad` passed twice without flake, while the body says the same replay multi-touch commands (two-pointer pan, pinch, rotate, transform) fail on that simulator. Please correct the claim and stamp the actual validation to the exact head.

Until the assertion is owned by an automatic lane and its regression/device evidence is accurate, this does not close #1584.

thymikee:
Addressed all three. Rebased onto main; pushed as `7b9f1a2fe` (assertion) → `ca0b4026d` (CI wiring). PR body rewritten with the corrected claims and full evidence inline; summary here:

**1. Not automatic — now fixed.** You're right that it's currently not automatic anywhere, not just nightly-only as I'd claimed: `replays-nightly.yml` moved its device replays to dispatch-only `replays-manual.yml` back on 2026-07-24 (#1781 A1), so `full:fixture-replays` (and `gesture-lab.ad` with it) doesn't run on any schedule right now. Split the duration check into its own isolated replay, `examples/test-app/replays/gesture-pan-duration.ad` (single-pointer pan + duration wait only, no multi-touch), and wired it as a new smoke-tier step in `ios.yml` so it runs on every PR without pulling in `gesture-lab.ad`'s multi-touch commands. Real evidence from this PR's own CI run: `✓ gesture-pan-duration.ad 33.7s` / `Test summary: 1 passed (1) in 33.7s` ([job](https://github.com/callstack/agent-device/actions/runs/32378738304/job/96456403877)).

**2. Red-before proof.** Built and ran live against `80feff42d` (the actual parent of #1572), with the same duration-observation code ported onto that snapshot. It fails as expected — `REPLAY_DIVERGENCE` / `wait timed out for text: pan duration >=400ms` — and I didn't stop at "it timed out": read the actual rendered text off the still-live app afterward, which shows `pan duration <400ms`, confirming the failure mode is specifically the duration collapsing, not incidental breakage (the earlier "Gesture lab" / "gesture canary ready" waits both passed). Then rebuilt and ran the same assertion against this PR's exact HEAD at the time (`ca0b4026d`) twice — both pass (`durationMs: 17445` and `6423`). Full commands/output in the PR body.

**3. Evidence inconsistency.** Corrected — these were two different claims I'd conflated. The duration assertion itself has never been flaky across any run (local 2x, red/green proof, now real CI). The multi-touch flakiness (pinch/rotate/transform/two-pointer pan) is a separate, pre-existing finding on my local machine's iOS 26.2 Simulator, confirmed via matched-pair testing against unmodified code, and it does **not** apply to anything this PR touches or runs automatically — I was not precise about that boundary before and the PR body now says so explicitly, including that I have not verified whether it reproduces in CI's environment.

Let me know if you'd like the full multi-touch flakiness written up as its own issue — it's out of scope here but real.

thymikee:
Re-reviewed exact head `ca0b4026`. The automatic iOS canary now exercises the intended replay path on every PR, and the red-before proof confirms the actual `<400ms` regression rather than an incidental timeout. Exact-head CI, including the isolated iOS smoke replay, is green; no code findings. Please rename `## Test plan` to `## Validation` and add the touched-file/no-expansion scope note as final body hygiene. Substantively merge-ready.

thymikee:
Re-reviewed exact head `cb9d3321`: code-clean. Moving the duration publication from `onFinalize` to `onEnd` makes the bucket contingent on an ACTIVE single-pointer pan, so failed or cancelled recognition cannot satisfy the canary. The isolated replay still exercises the public iOS route, and the red-before proof targets the original `<400ms` regression. All completed exact-head checks are green; the current iOS Smoke run, including this canary, is still in progress. Ready for human review; formal merge-readiness awaits that lane.

Issue #1900, opened 2026-08-20
test: close web platform command-coverage gaps from #1426


thymikee:
Part of #1426.

The web platform manifest records these commands as known gaps because HEAD has no command-specific web evidence that meets the manifest contract, even where generic runtime facts or handlers exist:

- artifacts
- boot
- shutdown
- logs
- events
- replay
- test
- install
- reinstall
- push
- batch
- diff
- trace
- press
- install-from-source

Follow-up work should add or identify executable web command evidence before changing the manifest rows. Keep the existing live web smoke scope unchanged unless a separate platform-evidence decision explicitly expands it.

Issue #1426, opened 2026-07-27
test: platform command-coverage manifests — web, macOS, Linux, tvOS (+ visionOS/watchOS decision)


thymikee:
Part of #1412 (Track D), carried forward in #1781. **Rescoped 2026-08-20:** the iOS and Android halves are done; this issue now tracks the remaining platform manifests.

## Done

- iOS: registry-derived manifest classifying all public commands (live / command-contract / workflow-live / capability-denial / known-gap) with a static gate (#1408; `test/integration/ios-simulator-e2e/coverage-manifest.ts` + `smoke-ios-simulator-coverage.test.ts`).
- Android: #1482 (`test/integration/android-emulator-e2e/coverage-manifest.ts`, `satisfies Record<PublicCommand, …>`, plus `smoke-android-emulator-coverage.test.ts` and the coverage-report test).

## Remaining platforms

One manifest per platform, one PR per platform, in this order (strongest existing evidence first — inventory verified against HEAD 2026-08-20):

- [x] **web** — _(merged: #1902; known-gap rows tracked in #1900)_ best first target. `web-smoke` CI job runs a real live smoke (`test/integration/smoke-web-platform.test.ts`) covering ~12 commands against a managed browser; `WEB_QUERY_COMMANDS`/`WEB_INTERACTION_COMMANDS` in `src/core/capabilities.ts` already enumerate the supported surface, so denial rows are mechanical; dedicated unit suites under `src/platforms/web/` and `packages/platform-web/`.
- [x] **macOS** — _(merged: #1922; known-gap rows tracked in #1916)_ `smoke-macos` CI job replays `test/integration/replays/macos/01-system-settings.ad` (~8 commands live); provider-scenario tests (`macos-desktop.test.ts`, `macos-recording.test.ts`) cover ~15 more; shared Apple runtime unit tests (deployment/network/logs/recording/inventory in `packages/platform-apple`) are citable command-contract evidence, same as iOS's manifest already does.
- [x] **Linux** — _(merged: #1921; known-gap rows tracked in #1915)_ `smoke-linux` CI job replays `01-desktop-smoke.ad` (~5 commands live); `linux-desktop.test.ts` provider-scenario covers ~15; dedicated unit tests under `src/platforms/linux/__tests__/`. Thinner than macOS; expect more known-gap rows.
[...]

PR #1902, opened 2026-08-20
test: add web platform command coverage manifest

thymikee:
## Summary

- Add an exhaustive `satisfies Record<PublicCommand, ...>` web manifest for all 54 public commands.
- Credit 12 commands to the existing live web smoke, 12 to existing web-specific contract evidence, 15 to mechanical capability denials, and 15 to explicit known gaps.
- Persist the manifest rollup and executed command steps in the existing smoke artifact.
- Add the grouped follow-up issue #1900 for web gaps and route `test/integration/web-e2e/**` away from the iOS workflow.
- This PR is web-only and adds no live scenarios; macOS, Linux, and tvOS are not included. Part of #1426.

## Validation

- `pnpm install --frozen-lockfile && pnpm build`
- `pnpm format`
- `pnpm check:quick`
- Focused web coverage gate: 7 tests passed.
- Planted-red proof: removing only the `artifacts` row failed catalog completeness and the expected 54-command count; restoring it passed.
- `AGENT_DEVICE_WEB_E2E=1 pnpm test:smoke:web`: 1 live test passed, covering 12 public commands. Artifact: `test/artifacts/web/live-web-platform-e2e-smoke/2026-08-20T11-48-29.572Z/`.
- `pnpm check:affected --run`: all runnable checks passed; GitHub-authoritative web/native lanes remain for CI.

Touched files are limited to the web manifest/report, existing web smoke integration, and the iOS workflow routing metadata needed by the affected-check gate.

### DIFF
--- a/.github/workflows/ios.yml
+++ b/.github/workflows/ios.yml
@@ -46,6 +46,7 @@ on:
       - 'test/integration/android-**'
       - 'test/integration/smoke-android-**'
       - 'test/integration/smoke-web-**'
+      - 'test/integration/web-e2e/**'
   push:
     branches:
       - main
--- a/test/integration/smoke-web-platform-coverage.test.ts
+++ b/test/integration/smoke-web-platform-coverage.test.ts
@@ -1,0 +1,120 @@
+import assert from 'node:assert/strict';
+import fs from 'node:fs';
+import path from 'node:path';
+import test from 'node:test';
+
+import { WEB_DESKTOP_DEVICE, mkdtempForTest } from '../../src/__tests__/test-utils/index.ts';
+import { PUBLIC_COMMANDS } from '../../src/command-catalog.ts';
+import { isCommandSupportedOnDevice } from '../../src/core/capabilities.ts';
+import {
+  WEB_COVERAGE_GAP_ISSUE,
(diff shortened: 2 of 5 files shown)


thymikee:
P2: Please make `coverage-report.json` unconditional. `runWebSmoke` awaits `cleanupWebSmoke()` before `writeCoverageReport()`; when `close` fails, `runStep` records the failed step and cleanup rethrows, so the report that should carry the manifest rollup and executed steps is absent on exactly that failure path. Write the report in a nested `finally` while preserving the cleanup error, and add a regression that forces `close` to fail and asserts the report remains.

thymikee:
Addressed in commit `a25526802`.

- The existing web smoke now writes `coverage-report.json` from a nested `finally`, so the report survives a failed `close` while the cleanup error still propagates.
- If cleanup and report writing both fail, the resulting `AggregateError` preserves both failures.
- Added regressions for failed-close report persistence and cleanup/report failure precedence.

Validation: 9 focused coverage tests passed, `pnpm check:quick` passed, the live web smoke passed, and `pnpm check:affected --run` passed all runnable checks. No review thread was present to resolve.

PR #1903, opened 2026-08-20
test: cover host runner tv remote and recording paths

### LINKED ISSUE #1897: test(ios): runner coverage gaps from the A7 pass — TvRemote has zero tests, ScreenRecorder.append monotonicity untested
Two coverage gaps surfaced by the #1781 A7 confirmation pass (the independent re-derivation of #1861's classification — full table in https://github.com/callstack/agent-device/issues/1781#issuecomment-5353990769). Both are gaps in what exists, not misclassifications; neither blocks anything.

## 1. `RunnerTests+TvRemote.swift` has zero test methods

The file exists in `apple/runner/AgentDeviceRunner/AgentDeviceRunnerUITests/` but contains no `func test…` — the tv-remote command path has no runner-side test at all, in any lane. The only tvOS evidence anywhere is the transcript-fake `test/integration/provider-scenarios/tvos-remote.test.ts`, which exercises the daemon side, not the Swift dispatch. A decision-level test (key-name → event mapping, invalid-key rejection) would be host-runnable under the #1861 classification — no simulator needed.

## 2. `ScreenRecorder.append(image:)` frame-timestamp monotonicity is untested

The sole recording test in the suite (`testRecordStopIsIdempotentAfterNativeRecorderAlreadyStopped`, `RunnerTests+RecordingTests.swift`) keeps `activeRecording` nil throughout, so the frame-append path — including the monotonicity guard on frame timestamps — never executes under test. A regression here produces unplayable/glitched MP4s, the same failure family as the Android moov bug (duration-0 validator, patched in place). A host-runnable test feeding fabricated timestamps through `append` covers it.

## Lane accounting (per the #1781 rule)

- Catches: silent tv-remote dispatch regressions; non-monotonic frame timestamps producing corrupt recordings.
- Evidence: gap found by the A7 classification pass; the recording family has a real prior on Android.
- Cost: two host-runnable tests in the existing Swift Runner Host XCTests lane (~0 added CI time).
- Kill criterion: if tv-remote is dropped as a command, delete with it; the recorder test lives as long as `ScreenRecorder` does.

thymikee:
## Summary

- Add a host-runnable tvOS remote-button mapping regression covering supported names, case normalization, and invalid keys.
- Add a host-runnable ScreenRecorder regression for monotonic timestamps through a shared integer-candidate allocator, without AVAssetWriter output or readiness timing in the test.
- Keep the existing host-XCTest lane and scope the change to three Swift files.

Closes #1897

## Validation

- `pnpm check:xctest-selection`
- Signed macOS host XCTest: 2 passed, 0 failed.
- Mutation proof: removing the monotonic clamp made the focused recording test fail deterministically (`0` != `101`; 1 test, 1 failure), then the clamp was restored.
- `pnpm check:affected --run` passed; GitHub-authoritative native lanes remain for CI.

### DIFF
--- a/apple/runner/AgentDeviceRunner/AgentDeviceRunnerUITests/RunnerTests+RecordingTests.swift
+++ b/apple/runner/AgentDeviceRunner/AgentDeviceRunnerUITests/RunnerTests+RecordingTests.swift
@@ -1,4 +1,31 @@
+import CoreGraphics
+import Foundation
 import XCTest
+#if canImport(AppKit)
+import AppKit
+#elseif canImport(UIKit)
+import UIKit
+#endif
+
+private func makeRecordingTestImage() -> RunnerImage {
+  let context = CGContext(
+    data: nil,
+    width: 2,
+    height: 2,
+    bitsPerComponent: 8,
+    bytesPerRow: 0,
+    space: CGColorSpaceCreateDeviceRGB(),
+    bitmapInfo: CGImageAlphaInfo.premultipliedLast.rawValue
+  )!
+  context.setFillColor(CGColor(red: 0.2, green: 0.4, blue: 0.6, alpha: 1))
+  context.fill(CGRect(x: 0, y: 0, width: 2, height: 2))
+  let cgImage = context.makeImage()!
+#if canImport(AppKit)
+  return NSImage(cgImage: cgImage, size: NSSize(width: 2, height: 2))
+#else
+  return UIImage(cgImage: cgImage)
+#endif
+}
 
 extension RunnerTests {
 #if AGENT_DEVICE_RUNNER_UNIT_TESTS
@@ -14,6 +41,29 @@ extension RunnerTests {
       XCTAssertEqual(response.data?.message, "recording already stopped")
       XCTAssertNil(activeRecording)
     }
+  }
+
+  func testScreenRecorderClampsNonMonotonicFrameTimestamps() throws {
+    let outputPath = URL(fileURLWithPath: NSTemporaryDirectory())
+      .appendingPathComponent("agent-device-screen-recorder-\(UUID().uuidString).mp4")
+      .path
+    defer { try? FileManager.default.removeItem(atPath: outputPath) }
+
+    let recorder = ScreenRecorder(outputPath: outputPath, fps: 30)
+    let image = makeRecordingTestImage()
+    try recorder.startForTesting(image: image)
+    defer { try? recorder.stop() }
+
+    XCTAssertEqual(
+      recorder.appendForTesting(image: image, timestampValue: 100),
+      100
+    )
+    // Non-vacuity: without the monotonicity guard, this append records 0 or fails instead of 101.
+    XCTAssertEqual(
+      recorder.appendForTesting(image: image, timestampValue: 0),
+      101
+    )
+    XCTAssertNoThrow(try recorder.stop())
   }
 #endif
 }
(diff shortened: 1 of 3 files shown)


thymikee:
Blocking: the new host XCTest is red on this head: `testScreenRecorderClampsNonMonotonicFrameTimestamps` expected `101` but returned `100`. That result is not a timestamp-clamp failure; `appendForTesting` returns `lastTimestampValue` when `append` exits at `input.isReadyForMoreMediaData`, and the immediate second append did exactly that. The regression therefore has not reached the guard it claims to pin. Make the test deterministic by proving the second frame was accepted/reached the append path (or by introducing a focused timestamp-allocation seam), then rerun the host XCTests and include the actual red/revert proof.

thymikee:
Addressed in 02c70b346.

The original test was genuinely exposed to AVAssetWriter backpressure: the immediate second append could be rejected before timestamp allocation, so the assertion observed 100 without exercising the monotonicity guard. I extracted the timestamp calculation into a focused seam and made the test helper explicitly record each allocated timestamp as an accepted frame, removing the writer-readiness race while keeping the production path unchanged.

Validation:
- Signed macOS host XCTest: 2 passed, 0 failed.
- Mutation proof: temporarily removing the clamp made the focused recording test fail deterministically (0 != 101; 1 test, 1 failure), then the clamp was restored.
- pnpm check:affected --run passed.

thymikee:
The writer-readiness race is gone, but the replacement is still not deterministic. `allocateTimestampForTesting(100)` converts 100 to `Double`, subtracts it from the host uptime, then reconstructs and floors it; cancellation can produce 99. For example, the exact Swift calculation at `nowUptime = 100000` is `99.99999999985448`, so this required host test can fail based only on runner uptime. Please put the monotonic clamp behind an integer-candidate seam and test 100 → 0 directly, leaving uptime-to-candidate conversion outside that seam. Also update the PR body: this head no longer uses real AVAssetWriter output or exercises `append`.

thymikee:
Addressed in 86a01e548.

The timestamp logic is now split into two seams:
- `timestampCandidateValue(for:)` performs the production-only uptime-to-integer conversion.
- `monotonicTimestampValue(for:)` owns the integer candidate clamp and is shared by production and the test helper.

The host test now drives the allocator directly with `100` then `0`; the helper records each returned value as an accepted frame, so there is no floating-point uptime reconstruction or AVAssetWriter readiness involved. The PR body has also been updated to describe this accurately.

Final validation:
- Signed macOS host XCTest: 2 passed, 0 failed.
- Mutation proof: removing only the clamp produced `0` != `101` (1 test, 1 failure), then the implementation was restored and the pair passed again.
- `pnpm check:xctest-selection` passed.

apple/runner/AgentDeviceRunner/AgentDeviceRunnerUITests/RunnerTests+RecordingTests.swift
```diff
@@ -15,5 +15,13 @@ extension RunnerTests {
       XCTAssertNil(activeRecording)
     }
   }
+
+  func testScreenRecorderClampsNonMonotonicFrameTimestamps() {
+    let recorder = ScreenRecorder(outputPath: "", fps: 30)
+
+    XCTAssertEqual(recorder.allocateTimestampForTesting(100), 100)
+    // The helper records each returned value as an accepted frame.
+    XCTAssertEqual(recorder.allocateTimestampForTesting(0), 101)
```
thymikee, line 24: shouldn't this be wrapped in AGENT_DEVICE_RUNNER_UNIT_TESTS?

Issue #1868, opened 2026-08-19
web: graceful daemon shutdown has no web-session close step — Chrome fleet survives until the idle window


thymikee:
`teardownSessionResources` (`src/daemon/session-teardown.ts:124-149`) finalizes recording, app-log, audio, and perf captures on daemon shutdown, but has no step for an open web session — unlike `session close`, which does dispatch a platform close (`src/daemon/handlers/session-close.ts:395`).

SIGTERM of a daemon holding an open web session therefore leaves the full agent-browser Chrome fleet (~15 processes) alive until agent-browser's own idle timer fires — `DEFAULT_AGENT_BROWSER_IDLE_TIMEOUT_MS = 5 * 60_000` (`src/platforms/web/agent-browser-lifecycle.ts:19`).

A fresh daemon on the same state dir does not reap it either: `cleanupManagedAgentBrowserOrphans` runs only at daemon/provider startup and skips fleets with `recent-browser-activity` (`agent-browser-lifecycle.ts:124-127`).

Not a regression — this is within #1109's accepted "within the idle window" behaviour — but it is the only capture-like resource with no shutdown step, and it blocks wiring #1859's daemon leak oracle to a web lane.

Wanted: a best-effort web close step in `teardownSessionResources`, mirroring the recording step added in #1325, plus a lane that asserts zero owned processes immediately after shutdown.

Found while building the B1 daemon leak oracle (#1781 wave 3, PR #1859); code paths independently verified in review.

PR #1904, opened 2026-08-20
test: add filesystem boundary fault matrix

### LINKED ISSUE #1431: test: deterministic boundary fault-injection matrix for daemon HTTP and subprocess seams

thymikee:
## Summary

- Add a shared synchronous atomic publisher that owns same-directory temp creation, replace or link-exclusive publication, cleanup, and primary-error preservation.
- Migrate the six issue-owned publish paths plus the adjacent process-lock, provider-lease, and remote-connection writers; durable capture keeps its wx, fsync, directory-sync, and symlink checks while sharing temp cleanup.
- Replace the duplicated filesystem fault drivers with one shared driver that tracks durable file descriptors explicitly and recursively checks for leftover temp files.
- Make filesystem owners and acceptance rows keyed typed records, so omitting a required row fails typechecking instead of comparing a registration list to itself.
- Assert process-death errors through the canonical AppError helper and add a structural ownership gate for simple publishers.
- The final diff touches 15 files. The scope expanded beyond the initial issue-owned modules only to migrate adjacent tmp→rename writers and centralize the shared test/ownership infrastructure; this does not add new fault axes or command-class coverage.

Part of #1431

This PR intentionally does not close #1431. The implemented scope is the filesystem errno matrix over the issue-owned tmp→rename modules plus the two acceptance-mandated rows. Remaining issue-owned work includes the socket and exec fault axes, lifecycle/process-tree cleanup assertions over those axes, the full fault-axis × command-class satisfies-cross, and the web-session SIGTERM candidate tracked in #1868.

## Validation

- pnpm install --frozen
- pnpm build
- pnpm check:quick
- Focused boundary and adjacent suites: 51 and 82 tests passed.
- Non-vacuous red proof: removing shared temp cleanup caused the session-script write/EIO row to fail with the exact leftover temp path; restoring it passes.
- pnpm check:affected --run passed all runnable checks: 448 related test files / 3,543 tests passed; changed-line coverage 35/36 (97.22%), changed-branch coverage 14/16 (87.50%).
- The restarted Android smoke run is currently in progress after the earlier retriable wait_capture_stalled failure: https://github.com/callstack/agent-device/actions/runs/32384140719
- Native, device, and web lanes remain GitHub-authoritative.

No CLI, docs, or skills changes were needed; this is internal fault handling and test ownership only.

### DIFF
--- a/src/__tests__/test-utils/filesystem-boundary-faults.ts
+++ b/src/__tests__/test-utils/filesystem-boundary-faults.ts
@@ -1,0 +1,133 @@
+import assert from 'node:assert/strict';
+import fs from 'node:fs';
+import path from 'node:path';
+import { test, vi } from 'vitest';
+import { isAtomicPublishTemporaryPath } from '../../utils/atomic-file.ts';
+import { mkdtempForTestSync } from './tmp-dir.ts';
+
+export type FilesystemErrno = 'EIO' | 'ENOSPC' | 'EMFILE';
+export type FilesystemFaultPoint = 'write' | 'rename';
+
+export type FilesystemBoundaryFixture = {
+  targetPath: string;
+  run: () => unknown | Promise<unknown>;
+  expected: 'throw' | 'return';
+  verifyReturn?: (value: unknown, errno: FilesystemErrno) => void;
+};
+
+export type FilesystemBoundaryOwner = (root: string) => FilesystemBoundaryFixture;
+
+export const FILESYSTEM_ERRNOS: readonly FilesystemErrno[] = ['EIO', 'ENOSPC', 'EMFILE'];
+export const FILESYSTEM_FAULT_POINTS: readonly FilesystemFaultPoint[] = ['write', 'rename'];
+
+/** Registers every fault row for a typed owner table. Missing required owners fail typechecking. */
+export function registerFilesystemBoundaryMatrix(
+  owners: Readonly<Record<string, FilesystemBoundaryOwner>>,
+  prefix: string,
+): void {
+  for (const [ownerName, createFixture] of Object.entries(owners)) {
+    for (const faultPoint of FILESYSTEM_FAULT_POINTS) {
+      for (const errno of FILESYSTEM_ERRNOS) {
+        test(`${prefix} ${ownerName} ${faultPoint} ${errno} leaves no unpublished temporary file`, async () => {
+          const root = mkdtempForTestSync(
+            `agent-device-${prefix}-${ownerName.replaceAll('/', '-')}-boundary-`,
+          );
+          const fixture = createFixture(root);
+          const fault = installFilesystemFault({
+            faultPoint,
+            errno,
+            targetPath: fixture.targetPath,
+          });
+
+          let value: unknown;
+          let thrown: unknown;
+          try {
+            value = await fixture.run();
+          } catch (error) {
+            thrown = error;
+          }
+
+          if (fixture.expected === 'throw') {
+            assert.ok(thrown, `${ownerName} should surface ${errno}`);
+            assert.equal((thrown as NodeJS.ErrnoException).code, errno);
+          } else {
+            assert.equal(thrown, undefined);
+            fixture.verifyReturn?.(value, errno);
+          }
+          assert.equal(fault.wasInjected(), true, `${ownerName} ${faultPoint} fault was injected`);
+          assert.equal(fs.existsSync(fixture.targetPath), false);
+          assert.deepEqual(listTemporaryFiles(root), []);
+        });
+      }
+    }
+  }
+}
+
+export function installFilesystemFault(options: {
+  faultPoint: FilesystemFaultPoint;
+  errno: FilesystemErrno;
+  targetPath: string;
+}): { wasInjected: () => boolean } {
+  const failure = createErrno(options.errno);
+  let injected = false;
+  const temporaryDescriptors = new Set<number>();
+  const isTargetOrTemporary = (value: unknown): boolean =>
+    value === options.targetPath || isAtomicPublishTemporaryPath(value, options.targetPath);
+
+  if (options.faultPoint === 'write') {
+    const originalOpenSync = fs.openSync;
+    vi.spyOn(fs, 'openSync').mockImplementation((...args) => {
+      const descriptor = Reflect.apply(originalOpenSync, fs, args);
+      if (isAtomicPublishTemporaryPath(args[0], options.targetPath)) {
+        temporaryDescriptors.add(descriptor);
+      }
+      return descriptor;
+    });
+
+    const originalWriteFileSync = fs.writeFileSync;
+    vi.spyOn(fs, 'writeFileSync').mockImplementation((...args) => {
+      const destination = args[0];
+      const writesPublishedTemporary =
+        isAtomicPublishTemporaryPath(destination, options.targetPath) ||
+        (typeof destination === 'number' && temporaryDescriptors.has(destination));
+      if (writesPublishedTemporary) {
+        injected = true;
+        Reflect.apply(originalWriteFileSync, fs, args);
+        throw failure;
+      }
+      return Reflect.apply(originalWriteFileSync, fs, args);
+    });
+  } else {
+    const originalRenameSync = fs.renameSync;
+    vi.spyOn(fs, 'renameSync').mockImplementation((...args) => {
+      if (isTargetOrTemporary(args[0]) || isTargetOrTemporary(args[1])) {
+        injected = true;
+        throw failure;
+      }
+      return Reflect.apply(originalRenameSync, fs, args);
+    });
+  }
+
+  return { wasInjected: () => injected };
+}
+
+function createErrno(errno: FilesystemErrno): NodeJS.ErrnoException {
+  const error = new Error(`injected filesystem ${errno}`) as NodeJS.ErrnoException;
+  error.code = errno;
+  return error;
+}
+
+function listTemporaryFiles(root: string): string[] {
+  const temporaryFiles: string[] = [];
+
+  function visit(directory: string): void {
+    for (const entry of fs.readdirSync(directory, { withFileTypes: true })) {
+      const entryPath = path.join(directory, entry.name);
+      if (entry.isDirectory()) visit(entryPath);
+      else if (entry.name.endsWith('.tmp')) temporaryFiles.push(entryPath);
+    }
+  }
+
+  visit(root);
+  return temporaryFiles;
+}
--- a/src/__tests__/test-utils/filesystem-boundary-faults.ts
+++ b/src/__tests__/test-utils/filesystem-boundary-faults.ts
@@ -17,8 +17,8 @@ export type FilesystemBoundaryFixture = {
 
 export type FilesystemBoundaryOwner = (root: string) => FilesystemBoundaryFixture;
 
-export const FILESYSTEM_ERRNOS: readonly FilesystemErrno[] = ['EIO', 'ENOSPC', 'EMFILE'];
-export const FILESYSTEM_FAULT_POINTS: readonly FilesystemFaultPoint[] = ['write', 'rename'];
+const FILESYSTEM_ERRNOS: readonly FilesystemErrno[] = ['EIO', 'ENOSPC', 'EMFILE'];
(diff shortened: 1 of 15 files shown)


thymikee:
This cannot close #1431 yet. The PR explicitly defers the issue’s required complete fault-axis × command-class matrix, and the current looped test registrations do not make a missing required row a compile/test failure. The new HTTP open/close test also only observes one request reaching a disposable endpoint; it does not assert the issue’s lifecycle cleanup or CLI error-contract invariants.

Please either implement the declared matrix completeness ownership and required invariants before retaining Fixes #1431, or change this to Part of #1431 and leave the issue open with the remaining acceptance work explicitly tracked.

thymikee:
Addressed in 6e3134bd2:

- Changed the relationship from `Fixes #1431` to `Part of #1431`; the issue remains open.
- Added app-log durable-store rows and post-write fault injection so write cleanup is exercised after a temp file exists.
- Added scoped completeness assertions for every filesystem/errno row and both acceptance command pairs.
- Strengthened the open/close transport-death seam with request-id, retry-hint, error-code, and no-replay assertions; lifecycle teardown/process-tree cleanup remains explicitly deferred to the full matrix.
- Added session-script temp cleanup in `finally`, preserving the primary filesystem error.

Final `pnpm check:affected --run` passed all runnable checks; native/device/web lanes remain GitHub-authoritative.

thymikee:

**Do not approve.** The production cleanup is the right behavior, but the implementation copies a protocol that still has no owner, then builds a second copy of the test driver to compensate. The code-judo move is visible and unused.

Issue #1431 asked for an owned fault matrix whose completeness is a compile/test error. This PR patches three `writeFile`+`rename` sites by hand, then proves them with ~400 lines of cartesian tests whose “completeness” checks mostly compare a list to itself.

### 1. Structural: the publish protocol is still unowned

The durable store already has the correct protocol: write a sibling temp, publish, always unlink the temp, never let cleanup clobber the primary errno (`src/daemon/durable-capture-resource-store.ts`).

This PR pastes a weaker subset of that into `writeClaim` and `writeRunnerLease`, and nests the same `try { rmSync } catch {}` around `publishHealedScriptAtomically`. `writeRunnerLease` is the same six lines as `writeClaim`. That is spaghetti growth: a new invariant bolted onto every call site instead of one primitive.

The judo: extract one sync helper, e.g. `publishFileSync({ destination, contents, mode })`, that owns temp creation, replace-publish, temp unlink, and primary-error preservation. Point claims, runner-lease, shutdown-report, and the other write-then-rename owners at it (`process-lock.ts`, `provider-lease-expiry.ts`, `remote-connection-state.ts` still have the unfixed sequence). Keep the durable store special — it needs `wx`, fsync, dirsync, and symlink refusal. Session-script can take `{ publish: 'replace' | 'link-exclusive' }` or at least share the temp-unlink helper.

Then the error class becomes impossible at the owning interface, which is the bar in `AGENTS.md`. The 36 errno×owner tests collapse to: helper write-fault, helper rename-fault, helper success leaves no temp, plus a construction/lint gate that tmp-rename owners call the helper.

Without that, the next owner will copy this `finally` a fourth time, or ship another leak. The test matrix is compensating for a missing module.

`session-script-writer.ts` is already 549 lines and `runner-lease.ts` already 519. AGENTS.md says extract before adding behavior past 500. This PR added nesting to both instead of extracting the protocol they were touching.

### 2. Structural: two fault drivers, and completeness that cannot go red

`src/daemon/__tests__/filesystem-boundary-faults.test.ts` (295 lines, at the 300 tripwire) and `runner-lease-filesystem-boundary-faults.test.ts` (116 lines) duplicate `FilesystemErrno`, `FaultPoint`, `installFilesystemFault`, `createErrno`, path matching, and a completeness test. The runner-lease completeness test is tautological: it pushes `runner-lease:${faultPoint}:${errno}` in the same loop that builds the expected list from `['runner-lease']`.

Worse, the acceptance file:

```ts
const PROCESS_DEATH_COMMANDS = ['open', 'close'] as const;
const REQUIRED_PROCESS_DEATH_COMMANDS = ['open', 'close'] as const;
assert.deepEqual([...PROCESS_DEATH_COMMANDS].sort(), [...REQUIRED_PROCESS_DEATH_COMMANDS].sort());
```

That cannot fail unless someone edits one copy and not the other. It is not the guarantee-matrix idiom. ADR 0011 completeness is a typed record: omitting a key is a type error. Here, forgetting both the fixture and the required list silently drops a row.

The daemon version is only slightly better: `REQUIRED_DAEMON_MATRIX_MODULES` vs `registerFilesystemMatrix(...)` side effects. Adding a module to one list and not the other fails; adding to neither does not. `FILESYSTEM_ERRNOS` and `FAULT_POINTS` drive both sides of the comparison, so a missing errno cannot fail the completeness test.

Judo: one typed owner table in `src/__tests__/test-utils/` (testing.md already tells you to put new helpers there):

```ts
const DAEMON_FS_BOUNDARY_OWNERS = {
  'device-claims': createDeviceClaimFixture,
  'session-script-writer': createSessionScriptFixture,
  // ...
} as const satisfies Record<DaemonFsBoundaryOwner, (root: string) => BoundaryFixture>;
```

`test.each(Object.entries(DAEMON_FS_BOUNDARY_OWNERS))`. Delete the push-array completeness tests. Runner-lease is one more owner in the same table, not a second harness. `listTemporaryFiles` should be the recursive one; the runner-lease copy is shallow for no reason.

The lost-response `REQUIRED_LOST_RESPONSE_ACCEPTANCE_COMMANDS` dual list has the same smell. `test.each(LOST_RESPONSE_MUTATION_ROWS)` plus `satisfies readonly { acceptanceCommand: 'press' | 'fill' }[]` is enough. The parameterization itself is the good part of that change — keep it, drop the ceremony.

### 3. Spaghetti: fd matching because the publish path is not a single function

Durable-store write faults identify the target with:

```ts
matchesPublishPath: (value) =>
  typeof value === 'number' || matchesTargetOrTemporary(value, targetPath),
```

That is a special-case branch in an already heuristic matcher (`String(value)`, two `startsWith` prefixes for two temp-naming conventions). Any `writeFileSync(fd, …)` during that test is treated as the publish write.

This exists because the test has to reverse-engineer three temp layouts (`file.json.pid.time.tmp`, `.file.pid-uuid.tmp`, fd write) instead of injecting at one helper. If the helper stays unextracted, spy `openSync` for the tmp path, record that fd, and match it. Do not match `typeof value === 'number'`.

`expected: 'throw' | 'return'` plus optional `verifyReturn` is the same problem in the fixture type. Shutdown-report swallows the errno and has no `verifyReturn`, so six matrix rows assert “no tmp, dest missing” and nothing about error identity. That is a different contract stuffed into the same cell shape. Either classify it as best-effort (and do not pretend errno survival is covered) or stop swallowing inside a shared `publishFileSync`.

### 4. Boundary / canonical helpers

The process-death row hand-rolls `assert.rejects` and reads `error.details?.hint`. `src/__tests__/test-utils/app-error.ts` exists specifically so tests assert the hint `normalizeError` returns, not `details.hint`. ADR 0010: the wire contract is `code`, `message`, `hint`, `diagnosticId`, `logPath`. Use `assertRejectsAppError`.

`verifyReturn` casts to `{ written?: boolean; error?: { code?: string; message?: string } }` instead of `SessionScriptWriteResult`. The writer already has that type.

Empty `catch {}` around `rmSync` preserves the primary errno, which is correct, but ADR 0010/0019 wanted cleanup failure as secondary diagnostic evidence, not discarded. That policy belongs in the one helper, not in three silent catches.

### What is fine

- Moving `writeFileSync` into the session-script `try` so a post-write fault still hits `finally` is a real bugfix. The claimed red proof (tmp left behind without it) matches the code.
- Spying `fs.writeFileSync` / `renameSync` instead of adding a test-only DI parameter follows `docs/agents/testing.md`.
- Lost-response `test.each` for tap/type plus “exactly one mutation request” is the right invariant at `runAppleRunnerCommand`. Recovery already keys off `isReadOnlyRunnerCommand`; these two rows are the issue’s required names, not a new policy.
- Test file topology (daemon vs runner-lease) is right. The driver should be shared, not the suites merged into one blob.
- Not closing #1431 is correct. Socket/exec axes and lifecycle teardown are still open.

### Approval bar

Not met. Visible judo that would delete the duplicated `finally`, the second harness, the fd special case, and the vacuous completeness tests:

1. Own sync atomic publish in one helper; migrate the simple write-then-rename sites.
2. One typed owner table + one fault driver in test-utils; drop the twin completeness arrays.
3. Assert process-death through `assertRejectsAppError`.

Until then this is a working local patch that makes the publish invariant harder to keep, not easier. Behavior can stay; the structure should not.

thymikee:
Addressed in 4cdffd74b and e1ef66f15:\n\n- Added src/utils/atomic-file.ts as the owner of same-directory temp creation, replace/link-exclusive publication, cleanup, and primary-error preservation.\n- Migrated claims, runner leases, shutdown reports, session scripts, process locks, provider-lease persistence, and remote connection state. Durable capture remains specialized for wx, fsync, directory sync, and symlink refusal while sharing the temp-path cleanup owner.\n- Replaced both local fault drivers with one shared test utility. Durable write injection records the temp file descriptor rather than treating every numeric write target as a match; temp cleanup checks are recursive.\n- Replaced self-comparing registration arrays with keyed typed owner/row tables. Missing filesystem owners or open/close and press/fill acceptance rows are now type errors.\n- Added a structural ownership gate for the simple publishers.\n- Switched process-death assertions to assertRejectsAppError, while retaining request-id and no-replay assertions.\n\nNon-vacuous proof: temporarily removing the shared cleanup made the session-script write/EIO row fail with the exact leftover temp path. Restoring cleanup made the focused 51-test run green.\n\nFinal pnpm check:affected --run passed all runnable checks and the gated push completed at e1ef66f15.

thymikee:
Re-reviewed exact head `e1ef66f`. Code is clean: the PR now correctly identifies itself as `Part of #1431`, the shared atomic-file owner removes duplicated temp/publish cleanup, and durable capture retains its stronger `wx`/fsync/directory-sync/symlink protocol. The sole Android failure is a retriable `wait_capture_stalled` before any readable snapshot, unrelated to this filesystem-only diff; rerun that lane before merge. Please also add the 15-file count and explicit scope-expansion statement to the body.

thymikee:
Updated the PR description:

- Added the final 15-file count.
- Added an explicit scope-expansion statement: the adjacent tmp→rename writers and shared test/ownership infrastructure are included to centralize cleanup and typed coverage, without expanding beyond the filesystem errno matrix and two acceptance rows.
- Kept the relationship as Part of #1431 and listed the deferred socket/exec, lifecycle/process-tree, full satisfies-cross, and web SIGTERM work.
- Recorded that the restarted Android smoke run is currently executing: https://github.com/callstack/agent-device/actions/runs/32384140719

PR #1905, opened 2026-08-20
fix: calibrate iOS scroll execution

thymikee:
## Summary

Calibrates default iOS scrolling for dense settings-style screens and makes edge verification follow the viewport-centered gesture instead of an unrelated nested scroll container.

The iOS amount and shared mobile duration now come from one cross-runtime parity contract, explicit distance/timing remains unchanged, Swift runner defaults are command-tested, and recording telemetry reflects each platform transport. Explicit point-target edge semantics are preserved. Scope is 17 files within the scroll command family.

## Validation

- Live iOS Simulator verification on a nested settings screen: the default scroll repeatedly reached the intended intermediate section, and target-less `scroll top --settle` correctly recognized the viewport was already at the edge without scrolling the nested control.
- `pnpm check:affected --run`: 4,562 tests passed; changed-line coverage 100%; formatting, lint, typecheck, layering, fallow, build, and XCTest selection passed.
- Exact iOS Simulator and macOS XCUITest runner builds passed.
- Regression tests were proven red against the prior default and prior broad-container selection, then green after the fix.
- Adversarial review completed with no remaining actionable findings.

Docs and skills are unchanged because command syntax and workflow guidance did not change.

### DIFF
--- a/apple/runner/AgentDeviceRunner/AgentDeviceRunnerUITests/RunnerTests+CommandExecution.swift
+++ b/apple/runner/AgentDeviceRunner/AgentDeviceRunnerUITests/RunnerTests+CommandExecution.swift
@@ -1905,13 +1905,14 @@ extension RunnerTests {
       guard let x = command.x, let y = command.y, let x2 = command.x2, let y2 = command.y2 else {
         return Response(ok: false, error: ErrorPayload(message: "drag requires x, y, x2, and y2"))
       }
+      let defaults = runnerDragCommandDefaults(command)
       return executeDragGesture(
         activeApp: activeApp,
         x: x,
         y: y,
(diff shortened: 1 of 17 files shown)


thymikee:
✅ Ready for human — reviewed at `3624fbaf9df3a264e13552a788f90e8253f7bfaf`. The calibrated iOS defaults reach the fused runner command without changing explicit distance/timing or macOS/tvOS behavior, and edge verification now follows the viewport-centered gesture rather than an off-center nested scroller. The regression and exact-head simulator evidence are non-vacuous, size remains below the scrutiny threshold, the PR is mergeable, and all checks—including iOS smoke and Swift runner XCTests—are green.

PR #1906, opened 2026-08-20
fix: make iOS scroll release controlled

thymikee:
## Summary

Make ordinary iOS viewport scrolling release predictably. The default path now uses the existing private XCTest synthesis route with a dedicated controlled-scroll profile: a 400 ms monotonic cubic deceleration over a 65% viewport finger path. `scroll top` and `scroll bottom` retain the fast inertial profile; Android and macOS behavior is unchanged.

Keep one drag execution path and one sampled-pointer builder instead of a scroll-only executor. The shared TypeScript/Swift fixture remains the source of truth for defaults and release policy, while the iOS-only cadence and curve stay beside their owning runner implementation. Continuous drag retains its pre-existing uncapped ~16 ms cadence (62 frames at 1 s, 625 at 10 s); only controlled scroll uses the 30-frame cap. Help and docs clarify that `amount` describes finger travel, while app pan thresholds, headers, bounds, and physics determine the final content offset.

This supersedes the first revision's public `XCUICoordinate` drag and destination hold, which still produced inertia in Element and diverged from the private-synthesis guarantee. The final implementation reuses the existing private gesture bridge and pointer builder. The exact +5.2 kB npm-unpacked growth covers the shared release-policy contract and fixture, runner controlled-scroll profile and cadence tests, dispatch plumbing, and command documentation; a distinct profile is needed because the smaller existing symmetric smoothstep path still produced post-release movement in Element.

Scope: 25 files across the scroll command family, runner implementation/tests, shared contracts, and command docs.

## Validation

- Required affected gate passed at `74ea556f2`: 570 files / 4,714 tests, with 100% changed-line and changed-branch coverage. One documented runner-cache contention test first timed out at 5 s, then passed in isolation in 215 ms and in the clean full rerun in 1.99 s.
- iOS XCTest build passed; macOS host XCTest build passed after exercising the non-iOS branch.
- Regression proof: the pre-fix continuous-drag test failed with 30 frames instead of 62 at 1 s and 625 at 10 s; the split-policy focused suite then passed (2 tests, 0 failures).
- Exact-head live iOS Settings run: the requested 400 ms controlled gesture measured 449.48 ms from `gestureEndUptimeMs - gestureStartUptimeMs`, traversed the planned 568 pt finger path, activated the scroll recognizer, and advanced the content from the top screen to the `Privacy & Security` section. Its row remained at exactly y=537.67 in a delayed read, with no post-release drift.
- Live Element Settings verification: seven controlled scrolls across both directions showed no visible post-release inertia. Four consecutive downward samples settled in 1.13–1.32 s and advanced progressively through the screen.
- Fast edge traversal remains mapped to the inertial profile; ordinary scroll remains on private XCTest synthesis with no coordinate fallback.


thymikee:
Reviewed exact head `1df68df7`: two blockers. First, Swift Host XCTests and macOS Smoke fail to compile because the non-iOS branch of `executeScrollDragGesture` references case-local `scrollPolicyKind` at `RunnerTests+CommandExecution.swift:2330`; use the declared `.scroll` policy (or pass it explicitly) and rerun both lanes. Second, this replaces default iOS scroll’s private synthesized schedule with public `XCUICoordinate` drag. ADR 0011 keeps that path private-synthesis/no-coordinate-fallback, and ADR 0013 requires live evidence for both recognizer activation and post-release movement before changing its schedule. The body proves only post-release AX stability. Preserve the declared schedule, or explicitly amend the ADR with matched recognizer + post-release evidence and non-vacuous route coverage. No `ready-for-human` yet.

thymikee:
Size also reports +3.3 kB npm unpacked, so refresh the body with an itemized growth rationale and why the smaller existing private-synthesis path is not sufficient.

thymikee:
Re-reviewed `70cd581d1`. The intended controlled-scroll design is now on the right route: ordinary iOS scroll stays on private XCTest synthesis, edge traversal remains inertial, the duplicate public-coordinate executor is gone, and the prior macOS compile failure is fixed. One hidden regression remains: `RunnerContinuousDragPointerPath` now inherits the new 30-frame cap. Explicit synthesized drag previously emitted ~16 ms smoothstep samples (62 at 1 s, 625 at 10 s); it now emits 30 for both, reaching 333 ms cadence at 10 s. That changes an unrelated ADR-0013 schedule without evidence. Preserve the old uncapped continuous-drag count and make any capped count controlled-scroll-only, with 1 s/10 s continuity assertions. Also record requested 400 ms versus `gestureEndUptimeMs - gestureStartUptimeMs` alongside the live recognizer/content result, as ADR 0013 requires. Finally, refresh the size rationale to the exact +5.2 kB unpacked report. No `ready-for-human` yet.

thymikee:
Addressed at `74ea556f2`.

- Split cadence ownership: continuous drag preserves the original uncapped 16 ms schedule (62 frames at 1 s, 625 at 10 s); only controlled scroll keeps the 30-frame cap.
- Added 1 s/10 s continuity assertions. They failed red at 30/30 before the fix, then the focused suite passed 2/2.
- Recorded exact-head live timing: requested 400 ms, measured 449.48 ms runner-side. The 568 pt Settings gesture activated scrolling, reached the `Privacy & Security` section, and its y=537.67 position was unchanged in the delayed read.
- Refreshed the PR body with the exact +5.2 kB unpacked rationale and 25-file scope.
- `pnpm check:affected --run` passed: 570 files / 4,714 tests, 100% changed-line and changed-branch coverage. (The documented runner-cache contention test first timed out, then passed in isolation at 215 ms and in the clean full rerun at 1.99 s.)

Current-head CI is now running; I have not added `ready-for-human`.

thymikee:
Re-reviewed exact head `74ea556f2`: clean on the prior blocker. Continuous drag again keeps its uncapped ~16 ms cadence (62 frames at 1 s, 625 at 10 s), while the 30-frame cap is controlled-scroll-only. Ordinary iOS scroll remains on the ADR 0011 private-synthesis/no-coordinate-fallback route; the 400 ms → 449.48 ms runner-uptime and no-drift evidence now satisfies ADR 0013.\n\nOne body correction: Size CI reports **+5.6 kB** npm unpacked, not +5.2 kB. The rationale is otherwise sufficient. iOS Smoke is still pending, so this is published and reported rather than merge-ready.

PR #1710, opened 2026-08-10
test: cap Vitest workers for parallel worktrees

thymikee:
## Summary

Cap default Vitest concurrency at two workers so three concurrent Codex worktrees leave CPU headroom for test subprocesses, simulators, and the OS on a 12-core development host.

Use the same shared cap for affected coverage runs. The subprocess-stub project remains serialized at one worker.

Scope: 5 files; test-runner configuration and its regression coverage only.

## Validation

The full unit bundle passed under three concurrent Codex workers: 680 files and 5,913 tests in 221.81s. Node integration passed 55 tests with 8 live-device cases skipped, and provider integration passed 156 tests. Affected coverage, integration progress, replay compatibility, formatting, lint, typecheck, layering, fallow, build, and package verification passed.

Red-before proof: removing only the new top-level cap makes the config regression fail with undefined !== 2; restoring it passes.

The package smoke initially timed out in the host Apple inventory probe after broad validation but passed in isolation. The final affected gate passed with xcrun unavailable, matching the package check stated device-free/offline contract.

No command behavior, docs, or skills changed.

PR #1908, opened 2026-08-20
perf(check-affected): stop running coverage locally, CI stays authoritative

thymikee:
## Summary

- Keep coverage out of the default local `check:affected --run` loop; the plan reports the authoritative GitHub `Coverage` job, while the dedicated coverage scripts remain available for explicit diagnosis.
- Run affected tests once through capped, uninstrumented `vitest related`, preserving the shared `--maxWorkers=2` contention bound and deduplicating the full unit/provider aggregates.
- Update the affected-gate model, regression tests, and testing guidance. Scope is 6 files with no expansion beyond the local affected-check workflow.

## Validation

- Exact head `3a1cc99413b751d565fd0071e379c711e3a58d10`
- `pnpm check:affected:test` — 47/47 pass
- Planted red: removing the shared Vitest worker cap fails three affected-gate assertions; restoring it returns 47/47 green
- `pnpm check:gate-manifest` — 46 checks wired across 33 lanes; coverage remains owned by GitHub CI
- `pnpm check:affected --json` — coverage reports `localRunnable: false` and the `Coverage` CI job
- `pnpm check:affected --run` — full fail-open local gate passes; coverage is skipped, related tests run once with `--maxWorkers=2`, and unit/provider aggregates are deduplicated

### DIFF
--- a/docs/agents/testing.md
+++ b/docs/agents/testing.md
@@ -224,11 +224,10 @@ docs-only short-circuit its path would otherwise take. If the matrix moves
 again, move that entry with it.
 The plan documents the rule and changed path behind every selected check.
 
-`coverage` (the changed-line coverage gate) is GitHub-authoritative and never runs locally:
-instrumenting Vitest is real overhead on top of an already-run suite, and the CI `coverage` job
-enforces this gate on every PR, so a local rerun only spends time without adding local-only
-signal. `--run` reports it as skipped; `vitest-related`, `unit`, and `provider-integration` run
-locally on their own, uninstrumented, instead of being folded into a coverage pass.
+Coverage is never instrumented by `check:affected --run`. The plan still reports the coverage
+obligation and its authoritative GitHub job, while the local path runs one capped, uninstrumented
+`vitest related` command and deduplicates full unit/provider aggregates. Run the dedicated coverage
+scripts explicitly when diagnosing a red CI result; do not make every pre-push loop pay for LCOV.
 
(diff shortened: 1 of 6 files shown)


thymikee:
Reviewed exact head `9b273e352a62e87f56e9431770de33e14c85b83c`. One blocker: the plain `vitest-related` path is now unbounded. Removing local coverage should remove instrumentation, but it does not remove the concurrency invariant from #1710: `vitest related` still spans configured projects on broad diffs, and uncapped workers have already starved otherwise-green subprocess/provider tests past their fixed timeouts in parallel worktrees. This head deletes `DEFAULT_VITEST_MAX_WORKERS` from the command and deletes the assertion that pinned it; the replacement test even describes the remaining run as high-parallelism. Keep `--maxWorkers=${DEFAULT_VITEST_MAX_WORKERS}` on the `vitest-related` command in `resolveCommand` and assert the cap alongside the absence of coverage flags. Coverage can remain GitHub-authoritative. Also align the body with the required `## Validation` heading and note the 4-file/no-expansion scope. CI is still running; no `ready-for-human` yet.

thymikee:
Addressed at exact head `3a1cc9941`. Coverage is now CI-authoritative by default, local affected tests run once without instrumentation under the shared `--maxWorkers=2` cap, and the full unit/provider aggregates are deduplicated. The planted cap removal reds three assertions; `pnpm check:affected --run` passes before push. CI has been triggered and is intentionally not awaited for this local-loop change.

Issue #1894, opened 2026-08-20
feat(maestro): support relational selectors with enforced conformance


thymikee:
## Problem

`agent-device replay|test --maestro` supports `index` and `childOf`, but rejects the rest of Maestro's relational selector family:

- `above`
- `below`
- `leftOf`
- `rightOf`
- `containsChild`
- `containsDescendants`

This prevents common disambiguation patterns such as focusing the input below an `Email` caption:

```yaml
- tapOn:
    below:
      text: Email
- inputText: qa@example.com
```

The request followed #1886, but the concerns are orthogonal. #1886 repaired a transient live-selector miss after target resolution; relational selectors express which target the author intended before dispatch.

Historical umbrella #558 listed relationship selectors as deferred work for the retired lowering engine. ADR 0015 now owns the production typed Maestro engine, so this issue supersedes only that unchecked compatibility item rather than reopening the old architecture.

Official syntax: https://docs.maestro.dev/reference/selectors/relational-selectors

Pinned semantic reference: Maestro v2.5.1 / `a4c7c95f` (`scripts/maestro-conformance/pinned-upstream.json`).

## Confirmed root causes

1. `MaestroSelector` is flat. `childOf` and `index` are carried separately on command, query, observation, and replay-plan types, while the other recursive fields are absent. Parsing, matching, diagnostics, and conformance therefore cannot preserve the official selector model through one owning interface. Upstream places both `childOf` and `index` inside recursive `ElementSelector`, so nested anchors can use them too.
2. `CanonicalSelector` and `upstreamSelector()` project only the existing flat fields plus side-car `childOf`. Accepting a relation without extending the recursive canonical projection would let both engines erase it and report false parity.
3. `matchesMaestroTypedSelector(node, selector)` accepts no snapshot. It cannot evaluate topology or geometry, yet `selectMaestroScrollableViewport` calls it with the authored `scrollUntilVisible` selector. Leaving that path unchanged would silently ignore relations while choosing a scroll container.
4. `label` has split ownership. Upstream v2.5.1 stores it on the YAML selector-shaped input but removes it from `ElementSelector` and passes it as command-description metadata. Agent-device currently treats it as a matching constraint only for selector-form `tapOn`, as metadata for point-form `tapOn`, and rejects it on other selector-bearing commands. Conformance and export currently hide the mismatch.

### One recursive selector owns all seven relations

[...]

PR #1909, opened 2026-08-20
fix(maestro): align label metadata ownership

thymikee:
## Summary

Part of #1894. This is the first PR in a three-PR stack; followed by #1910 and #1911.

- Treat Maestro `label` consistently as outer command metadata instead of an element selector.
- Keep native `.ad` replay export round-trippable by projecting `label=` to Maestro `text:` with an explicit broader-match warning.
- Extend the canonical conformance projection, supported-capability declaration, authored corpus, help, and replay docs so this breaking ownership decision is visible and mechanically checked.
- Extract parser and conformance responsibilities before adding behavior to files already over the repository size tripwire.

This PR changes 38 files across the Maestro parser/export/conformance and replay integration surfaces. That cross-module scope is intentional: landing only parser acceptance or only export behavior would create the silent/invalid round trip described in #1894.

## Validation

The cumulative stack head `3d7208a07` passed `pnpm check:affected --run`: 4,273 affected tests, changed-line coverage at 90.60%, package/build/layering/fallow checks, and Maestro conformance at 47/47. The label-focused parser, export, replay, support-matrix, and conformance tests are included in that run.

GitHub CI on this PR head remains authoritative for independent-layer readiness. No live device run was performed for this parser/export ownership slice; it does not change a platform interaction path.

### DIFF
--- a/packages/maestro/package.json
+++ b/packages/maestro/package.json
@@ -8,6 +8,7 @@
   "dependencies": {
     "@agent-device/contracts": "workspace:*",
     "@agent-device/kernel": "workspace:*",
+    "@agent-device/selectors": "workspace:*",
     "yaml": "^2.9.0"
   },
   "exports": {
--- a/packages/maestro/src/internal/__tests__/conformance-selector-projection.test.ts
+++ b/packages/maestro/src/internal/__tests__/conformance-selector-projection.test.ts
@@ -1,0 +1,51 @@
+import { expect, test } from 'vitest';
+import { canonicalizeAgentCommands, canonicalizeUpstreamFlow } from '../conformance-normalize.ts';
+import {
+  canonicalizeAgentSelector,
+  canonicalizeUpstreamSelector,
+} from '../conformance-selector-projection.ts';
+import { parseMaestroProgram } from '../program-ir-parser.ts';
+
+test('canonicalizes command labels separately from selector identity', () => {
+  const program = parseMaestroProgram(
+    ['---', '- tapOn:', '    text: Save', '    label: save action'].join('\n'),
+  );
+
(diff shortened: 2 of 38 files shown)


thymikee:
Reviewed exact head `b076272e`. One maintainability issue before this slice is done: `readMaestroCommandLabel` is introduced as the owner for outer-command label parsing, but only some callers use it. `inputText`, `runFlow`, `swipe`, and `tapOn` still spell the same `label` read inline. That leaves us with both a new abstraction and the duplication it was meant to replace. Please route every outer-command label read through the helper and remove the redundant imports and branches, or delete the helper and keep the direct reads; for this ownership PR, one shared owner is the smaller design. Also, all three bodies refer to stack `#1912`, but no issue or PR #1912 currently resolves. Please replace that with valid stack metadata.

thymikee:
Addressed at `a4f713bff`: all outer command labels now route through `readMaestroCommandLabel`, including `tapOn`, `swipe`, `inputText`, and `runFlow`; the focused label/parser tests pass, and a labeled `runFlow` regression was added. I also replaced the invalid `#1912` metadata with direct dependencies on #1910/#1911.

thymikee:
Reviewed exact head `a4f713bff`. Code-clean: `label` now has one outer-command metadata owner across supported parser paths, selector identity no longer accepts it, and native export’s warning-backed `label=` → `text:` projection self-parses while compound selectors remain hard errors. The extraction removes duplicated parsing rather than adding ceremony; Size reports −7.0 kB npm unpacked. All completed exact-head checks are green; iOS Smoke is still running. Please refresh the body’s Validation SHA/CI evidence from the cumulative stack head to this PR head.

Issue #1907, opened 2026-08-20
fix(maestro): preserve exact clickableFirst selector ordering


thymikee:
## Problem

Pinned Maestro v2.5.1 applies `Filters.clickableFirst()` after every selector filter that has no authored `index`. Agent-device's normalized `RawSnapshotNode` does not preserve the provider's exact `clickable` fact, so the direct Maestro engine cannot mechanically reproduce that final ordering for every platform.

Using `hittable` is not equivalent: on Android it deliberately combines touch-clickable and focusable/focused targets. Type-based interactive-control heuristics are also not exact. Either substitution would violate the compatibility rule against silently approximating accepted Maestro syntax.

This limitation predates relational selector support and affects ordinary and relational selectors alike. #1894 therefore proves relational membership, `Filters.relativeTo` pair ordering, set composition, stable selection when upstream clickable evidence is equal/absent, and authored `index`; it does not pretend that the missing provider fact is available.

Pinned references:

- `Orchestra.buildFilter`: `a4c7c95f`, lines 1520-1531
- `Filters.clickableFirst`: `a4c7c95f`, lines 241-245
- `TreeNode.clickable`: `a4c7c95f`, `maestro-client/src/main/java/maestro/TreeNode.kt`

## Required design work

- Identify a provider-internal way to retain exact clickability for Maestro resolution without adding bytes to public snapshot tree output.
- Keep one hierarchy capture per resolution attempt and avoid a second production matcher or provider query.
- Define platform behavior where the provider does not expose an exact equivalent; declare an explicit divergence rather than infer from `hittable` or control type.
- Apply clickable-first only when `index` is absent; authored index continues to use pinned y/x ordering.

## Acceptance criteria

- Generated pinned Layer-2 vectors cover `clickable=true`, `false`, and absent evidence, including stable order within equal groups.
- The live resolver selects the same no-index candidate as pinned `Filters.clickableFirst()` for supported providers.
- An authored `index` bypasses clickable-first and retains pinned `Filters.INDEX_COMPARATOR` behavior.
- The solution does not add per-node bytes to public snapshot output or perform another hierarchy capture.
- Any provider that cannot supply exact evidence has a documented, machine-enforced divergence rather than a heuristic fallback.

PR #1910, opened 2026-08-20
feat(maestro): add recursive tree selectors

thymikee:
## Summary

Part of #1894. Depends on #1909 and is followed by #1911.

- Replace command-side `index`/`childOf` ownership with one recursive selector IR.
- Add `childOf`, `containsChild`, and `containsDescendants` end to end: parsing, canonical projection, snapshot-aware matching, target resolution, scrolling, replay, and support metadata.
- Make the flat matcher accept only leaf selectors so recursive relations cannot be silently ignored at a call site.
- Add generated pinned-JVM tree vectors and authored corpus coverage, including final selection/index behavior rather than parse-only acceptance.

This PR changes 43 files across the recursive-selector ownership seam. The breadth is deliberate: each relation lands parse-to-runtime with its conformance oracle, and `scrollUntilVisible` uses the same snapshot-aware resolver.

## Validation

The cumulative stack head `3d7208a07` passed `pnpm check:affected --run`: 4,273 affected tests, changed-line coverage at 90.60%, package/build/layering/fallow checks, and Maestro conformance at 47/47. The recursive-selector parser, matcher, scroll, replay, and Layer 2 tree-vector tests are included in that run.

Live device evidence was recorded at exact PR head `0307db29d` on iPhone 17 Pro simulator `F7D6F9A4-4FCC-4DD7-AC0B-3280C9319CB9`. A healthy XCTest snapshot confirmed the Agent Device Tester Form surface. A four-step Maestro flow used `id: field-name` with `childOf: {text: Checkout form}` through `scrollUntilVisible`, `tapOn`, `inputText`, and relational `assertVisible`; fresh readback returned `Recursive Tree` from the targeted field. Session `pr1910-tree-live` was closed with the same platform, UDID, and isolated state directory.

The regression was also observed red on the pre-change base: the authored relation was ignored and both candidates survived where one was expected. GitHub CI is green on this exact PR head. Exact `clickableFirst` evidence is intentionally not approximated from `hittable`; the broader snapshot evidence gap is tracked in #1907.

PR #1911, opened 2026-08-20
feat(maestro): add positional selectors

### LINKED ISSUE #1894: feat(maestro): support relational selectors with enforced conformance

thymikee:
## Summary

Closes #1894. Depends on #1910 and completes the three-PR stack begun by #1909.

- Add recursive `above`, `below`, `leftOf`, and `rightOf` selectors end to end.
- Match pinned Maestro v2.5.1 exactly: strict top-left comparisons, finite zero/negative-size geometry, integer-truncated centers, Float32 distance, pair intermediate ordering, and stable final set composition/index selection.
- Route positional selectors through snapshot-aware `scrollUntilVisible` targeting and exclude them from atomic iOS selector dispatch.
- Add the Email caption/input regression plus generated positional oracle vectors and authored corpus coverage for zero-size, missing geometry, multiple anchors, odd centers, composition, and nested tree/position selectors.
- Resolve each recursive selector and relation once per immutable snapshot, with cached ancestor and positional match sets.

This PR changes the positional-selector, resolver, and conformance seams. It intentionally keeps parse acceptance and runtime support together so no accepted relation can be silently ignored.

## Validation

Stack head `3d7208a07` passed `pnpm check:affected --run`: 4,273 affected tests, changed-line coverage at 90.60%, package/build/layering/fallow checks, and Maestro conformance at 47/47. All 30 exact-head GitHub checks are green.

Live device evidence was recorded at exact head `3d7208a07` on iPhone 17 Pro simulator `F7D6F9A4-4FCC-4DD7-AC0B-3280C9319CB9`. A healthy XCTest snapshot confirmed the Agent Device Tester Form surface. Maestro replay completed 4/4 steps with no healing: `scrollUntilVisible` and `tapOn` both targeted `{below: {text: Email}}`, followed by input and relational `assertVisible`. Fresh readback returned exactly `qa-position@example.com` from `id="field-email"` with `hittable: true`. Session `pr1911-email-live` was closed with the same platform, UDID, and isolated state directory.

Mechanical red proofs were captured before restoring the implementation:

- clean base retained two candidates where `above` should retain one;
- the old positive-area geometry guard excluded finite zero-size relation candidates;
- fractional centers inverted the odd-dimension pinned Layer 2 ordering;
- removing array-offset parent fallback broke non-contiguous containment;
- the prior per-candidate resolver repeated nested relation computation instead of computing each distinct selector once per snapshot.

Exact `clickableFirst` evidence remains the separate #1907 follow-up; this stack does not infer it from `hittable`.


thymikee:
Blocking stack-level design issue at exact head `d194a6c97`: the selector resolver is set-oriented at its API but node-oriented internally. `selectMaestroSnapshotMatchesWithoutOwnIndex` filters N candidates; for every candidate it rebuilds the node map and re-resolves every `childOf`, containment, and positional selector. A positional relation also scans all nodes by anchors, so ordinary positional selection becomes cubic-shaped and nested relations compound it. The current tiny fixtures cannot expose that cost.

Please make one snapshot-scoped resolver own this work: build the node map once, resolve each nested selector once, compute relation match sets once, then intersect those sets with the leaf matches while preserving snapshot order and index semantics. Keep visibility and iOS ranking as the consumer, and move relation evaluation out of the now-308-line `runtime-target-ranking.ts`. That is a deeper but smaller abstraction: it removes repeated searches and the one-node/full-set impedance mismatch instead of adding caches around the current shape. Add a bounded large-snapshot or call-count regression so the resolver cannot return to per-candidate recursive resolution.

thymikee:
Addressed at `3d7208a07`: recursive selector resolution is now snapshot-scoped. The node map is built once; nested selector, ancestor, and positional match sets are cached; `childOf` intersects cached ancestor sets; and positional pair order remains available to conformance. A 128-candidate call-count regression proves one computation per root/childOf/below selector and one positional computation, plus a non-contiguous parent-link regression. The cumulative affected gate passes 4,273 tests at 90.60% changed-line coverage and Maestro conformance 47/47.

thymikee:
Reviewed exact head `3d7208a07`. Code is clean: positional selectors are recursive end to end, resolution is snapshot-scoped, pinned conformance semantics and iOS atomic-dispatch exclusion are covered, and all exact-head CI is green. One readiness item remains: record a live Android or iOS run of the Email-caption → input-below tap/scroll path and close the session; the current runtime-port fixture evidence does not replace device-facing proof. With that evidence, this stack tip is merge-ready.

thymikee:
Live device evidence recorded at exact head `3d7208a07` on iPhone 17 Pro simulator `F7D6F9A4-4FCC-4DD7-AC0B-3280C9319CB9`, using the rebuilt source CLI after `pnpm clean:daemon`. A healthy XCTest snapshot confirmed the Agent Device Tester Form surface. Maestro replay then completed 4/4 steps with no healing: `scrollUntilVisible` and `tapOn` both targeted `{ below: { text: Email } }`, followed by `inputText: qa-position@example.com` and relational `assertVisible`. A fresh `get text id="field-email"` returned exactly `qa-position@example.com` on the `TextField` with `hittable: true`. Session `pr1911-email-live` was closed successfully with the same platform, UDID, session, and isolated state dir. This closes the requested device-facing readiness item.

thymikee:
Re-reviewed exact head `3d7208a07`: the live Email → `below` flow now covers scroll, tap, input, relational assertion, exact readback, and session cleanup; all 30 current checks are green. The snapshot-scoped resolver is the final shared tree/position owner and deletes the intermediate per-candidate work rather than adding a parallel matcher. No findings; merge-ready as the tip of `#1909 → #1910 → #1911`.

PR #1913, opened 2026-08-20
docs: capture gesture verification lessons

thymikee:
## Summary

Capture two durable lessons from the iOS gesture verification work:

- callback-based canaries must observe semantic success, not a terminal callback that also runs after failure or cancellation;
- a replay is automatic evidence only when an automatic PR or scheduled lane actually executes it on the exact head.

ADR 0013 now also distinguishes an intended zero-velocity curve from observed controlled release: destination holds and analytic endpoint slope are insufficient evidence without measuring post-lift app behavior.

Scope: 2 documentation files. No runtime, protocol, fixture, or workflow behavior changes. `CONTEXT.md` remains untouched because these are verification rules, not domain vocabulary.

## Validation

- `pnpm check:affected --run` (`agent-guidance` passed)
- Runtime/device validation does not apply to this documentation-only change


thymikee:
Reviewed 058f4cb. Clean: the guidance accurately requires semantic-success canaries, app-observed post-lift gesture evidence, and an automatic lane that actually ran on this head. Docs-only; relevant checks are green. Ready for human.

PR #1917, opened 2026-08-20
fix: preserve Maestro clickable-first ordering

### LINKED ISSUE #1907: fix(maestro): preserve exact clickableFirst selector ordering

thymikee:
## Summary

Preserve pinned Maestro v2.5.1 clickableFirst ordering for Android selector resolution without adding clickability to the public snapshot wire shape.

- Retain exact Android helper clickability in an internal snapshot sidecar from the single hierarchy capture.
- Apply stable true, false, absent ordering only when no selector index is authored.
- Preserve the existing y/x comparator for authored indices.
- Classify iOS XCTest/private AX as an explicit unsupported divergence instead of inferring clickability.
- Add generated Layer 2 vectors, runtime parity tests, transport/public-shape tests, and one-capture evidence.

Closes #1907

## Validation

- pnpm check:affected --run passed; all runnable checks passed.
- pnpm maestro:conformance passed 47/47.
- Related Vitest passed 4,620 tests across 563 files.
- Red proofs observed for reversed ordering, public enumerable clickability, and a second hierarchy capture; all were restored green.
- Native/device, coverage, differential, fuzz, mutation, concurrency, and replay lanes remain GitHub-authoritative.

Scope: 21 files across contracts, Android capture propagation, Maestro resolution, and generated conformance coverage.

### DIFF
--- a/examples/test-app/README.md
+++ b/examples/test-app/README.md
@@ -88,7 +88,9 @@ The `/automation` route is intentionally JavaScript-only and can be opened from
 **Settings → Open automation lab** or with the
 `agent-device-test-app:///automation` scheme. Its stable `automation-*` ids expose durable
 outcomes for long press, native alert actions, app-event name/payload, app state, appearance,
-window orientation, and microphone permission recovery. CI repacks JavaScript-only changes into the
+window orientation, and microphone permission recovery; the
+`maestro-clickable-first-target` duplicate pair exercises Android Maestro clickable-first
+ordering. CI repacks JavaScript-only changes into the
 cached Release app without starting Metro; native configuration changes intentionally produce one
 new fingerprinted build that all simulator consumers share.
 
--- a/examples/test-app/src/screens/AutomationLabScreen.tsx
+++ b/examples/test-app/src/screens/AutomationLabScreen.tsx
@@ -161,26 +161,28 @@ export function AutomationLabScreen(props: {
         />
       </SectionCard>
 
-      <SectionCard
-        subtitle="The first duplicate is inert; the second duplicate is clickable."
-        title="Maestro clickable ordering"
-      >
-        <View style={styles.maestroTarget} testID="maestro-clickable-first-target">
-          <Text style={styles.maestroTargetLabel}>Inert duplicate target</Text>
-        </View>
-        <Pressable
-          accessibilityLabel="Clickable duplicate target"
-          accessibilityRole="button"
-          onPress={() => setMaestroSelection('clickable')}
-          style={({ pressed }) => [styles.maestroTarget, pressed ? styles.pressed : null]}
-          testID="maestro-clickable-first-target"
+      {Platform.OS === 'android' ? (
+        <SectionCard
+          subtitle="The first duplicate is inert; the second duplicate is clickable."
+          title="Maestro clickable ordering"
         >
-          <Text style={styles.maestroTargetLabel}>Clickable duplicate target</Text>
-        </Pressable>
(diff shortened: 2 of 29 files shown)


thymikee:
The Android/Maestro implementation is code-clean at `57ffb12`. Before this is merge-ready, please add one Android device/emulator-facing scenario that captures two matching nodes through the helper and resolves a no-index Maestro selector, proving the clickable candidate is selected first. Current tests separately prove helper evidence and resolver ordering, but not the full shipped path.

thymikee:
[P1] Keep this Android-only fixture out of the shared iOS automation tree. #1907 needs Android helper clickability, while this unguarded section changes the iOS fixture immediately before its input/alert sequence; the current iOS smoke is red waiting for `Alert result: opened`. Gate it with `Platform.OS === "android"` (or place it beyond unaffected shared checks), then rerun iOS and Android. The Android run failed on an emulator system overlay before `smoke:maestro-clickable-first` executed, so it still needs a successful live result.

thymikee:
Reviewed exact head `61d35eb`. Clean and merge-ready. The Android-only fixture preserves iOS coverage; the Android helper’s exact clickability now survives the interactor and daemon snapshot reshapes in a non-serialized sidecar; and the successful Android emulator smoke (`smoke:maestro-clickable-first`) proves a no-index Maestro tap selects the clickable duplicate. iOS stays an explicit, tested divergence rather than inferring from `hittable` or type. All checks are green; npm unpacked size is +1.5 kB.

Issue #1915, opened 2026-08-20
test: close Linux platform command-coverage gaps from #1426


thymikee:
Part of #1426.

The Linux platform manifest records these commands as known gaps because HEAD has no command-specific Linux evidence that meets the manifest contract:

- artifacts
- capabilities
- doctor
- shutdown
- logs
- events
- replay
- test
- install
- reinstall
- push
- batch
- diff
- record
- trace
- find
- swipe
- install-from-source

Follow-up work should add or identify executable Linux command evidence before changing the manifest rows. Keep the existing Linux replay smoke scope unchanged unless a separate platform-evidence decision explicitly expands it.

PR #1919, opened 2026-08-20
test: add tvOS platform command coverage manifest

thymikee:
## Summary

- Add an exhaustive hand-rolled tvOS manifest satisfying `Record<PublicCommand, ...>` for all 54 public commands.
- Classify 0 live, 13 command-contract, 3 capability-denial, and 38 known-gap rows; no new tvOS live scenario is added and no row claims live evidence.
- Add the static completeness/report gate to the existing Node integration lane; every known gap points to grouped follow-up issue #1914.
- Cite the existing tvOS provider scenario, shared Apple deployment/inventory contracts, the executable tvOS-inclusive capability oracle, and the typed multi-touch policy. Part of #1426.

## Validation

Evidence gathered at commit `8611dcfde`:

- Fresh worktree install/build, `pnpm check:quick`, and the full `pnpm check:affected --run` completed successfully; all runnable checks passed.
- The new tvOS static gate passed all 7 tests, and the existing provider-backed tvOS remote scenario passed.
- Planted-red proof: omitting only the `artifacts` row failed catalog completeness and changed the report to 53 total / 38 gaps; restoring it returned all 7 gate tests to green.
- Follow-up fix: host-dependent `audio` is now a command-contract row backed by the existing exhaustive tvOS-inclusive capability parity test. The gate restores full oracle parity and projects the declared host-dependent admission; the pinned report is 3 denials / 13 contracts / 38 gaps.
- GitHub-authoritative native/device/coverage lanes remain for CI. The follow-up has been pushed and those checks are rerunning; this PR is published and reported, not merge-ready on live evidence.

## Follow-up

The coverage summary builder remains local to each platform manifest to keep this tvOS change platform-scoped and avoid introducing a shared row type. The macOS coverage-manifest PR should factor the pure summary builder into `test/integration/support/` before adding the fourth copy; this PR's count assertion keeps tvOS command additions explicit.

Touched files: 2, tvOS-only. Docs/help/skills were unchanged because this is a test-coverage manifest and no command behavior changed.

PR #1921, opened 2026-08-20
test: add Linux platform command coverage manifest

thymikee:
## Summary

Add an exhaustive Linux desktop command-coverage manifest over all 54 public commands. It records 5 existing replay-live commands, 20 command-contract citations from the existing Linux provider/runtime/unit evidence, 11 descriptor capability denials, and 18 explicit known gaps grouped in #1915. The static catalog, denial, evidence, and report gate runs in the existing Node integration PR lane. The pure classification summary helper is shared by the Android, web, and Linux manifests while each platform keeps its hand-rolled row union; Linux also reuses the canonical `LINUX_DEVICE` fixture. No live scenarios were added and no macOS manifest files were changed. Part of #1426.

## Validation

- `pnpm install --frozen-lockfile && pnpm build`
- `pnpm format`
- `pnpm check:quick`
- Android, web, and Linux coverage gates: 25 tests passed.
- Linux unit evidence: 6 files / 36 tests passed.
- Linux provider scenario: 1 test passed.
- Planted-red proof: removing only `install-from-source` failed the catalog completeness invariant and changed expected counts from 54 total / 18 gaps to 53 total / 17 gaps; restoring it passed.
- `pnpm check:gate-manifest` passed.
- `pnpm check:affected --run`: all runnable checks passed; Linux replay and other device/native/coverage lanes remain GitHub-authoritative.

Review follow-up is committed at `5db8b0e78`: the redundant Linux workflow rerun was removed because the existing Node integration PR lane already executes the gate. The open tvOS PR should adopt the shared helper before its separate platform branch merges.

### DIFF
--- a/.github/workflows/ios.yml
+++ b/.github/workflows/ios.yml
@@ -45,6 +45,8 @@ on:
       - 'test/integration/replays/macos/**'
       - 'test/integration/android-**'
       - 'test/integration/smoke-android-**'
+      - 'test/integration/smoke-linux-**'
+      - 'test/integration/linux-e2e/**'
       - 'test/integration/smoke-web-**'
       - 'test/integration/web-e2e/**'
   push:
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -79,9 +79,6 @@ jobs:
       - name: Setup toolchain
         uses: ./.github/actions/setup-node-pnpm
 
-      - name: Run Linux desktop catalog coverage contract
-        run: node --experimental-strip-types scripts/node-test-tmpdir.ts --test test/integration/smoke-linux-coverage.test.ts
-
       - name: Start Xvfb and D-Bus
         run: |
           # Start virtual framebuffer (1280x1024, 24-bit color)
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -79,6 +79,9 @@ jobs:
       - name: Setup toolchain
         uses: ./.github/actions/setup-node-pnpm
 
+      - name: Run Linux desktop catalog coverage contract
+        run: node --experimental-strip-types scripts/node-test-tmpdir.ts --test test/integration/smoke-linux-coverage.test.ts
+
       - name: Start Xvfb and D-Bus
         run: |
           # Start virtual framebuffer (1280x1024, 24-bit color)
--- a/test/integration/android-emulator-e2e/coverage-manifest.ts
+++ b/test/integration/android-emulator-e2e/coverage-manifest.ts
@@ -2,6 +2,7 @@ import { PUBLIC_COMMANDS } from '../../../src/command-catalog.ts';
 import { ANDROID_AUDIO_CONTRACT_EVIDENCE } from '../../../src/daemon/handlers/__tests__/session-audio.coverage.ts';
 import { ANDROID_INSTALL_SOURCE_CONTRACT_EVIDENCE } from '../../../src/platforms/__tests__/install-source.coverage.ts';
 import { ANDROID_LIFECYCLE_CONTRACT_EVIDENCE } from '../provider-scenarios/android-lifecycle.coverage.ts';
+import { buildCoverageClassificationSummary } from '../support/coverage-classification.ts';
 import {
   defineAndroidContractEvidence,
   type AndroidContractEvidence,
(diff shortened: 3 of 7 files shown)


thymikee:
## Review — approve with one required coordination item and one cut

Read the manifest, gate, and both workflow diffs against the #1426 ground rules; verified #1915 exists and the ios.yml routing question independently.

**What's right — this is the strongest gate of the three so far:**
- **Live claims are parsed, not pattern-matched** — `parseReplayScriptDetailed` over the actual `.ad` replay, then set-membership per claimed command. A comment or a renamed command can't fake evidence. Better than the web/tvOS substring approach; the macOS manifest should copy this. ✓
- **Denial rows carry owner + declaration citations** (`CapabilityDeclarationEvidence`) checked against source, *and* the denial set is mechanically diffed against `isCommandSupportedOnDevice`. Both directions guarded. ✓
- Planted-red proof is in the PR body (54/18 → 53/17 and red) — exactly what the ground rule asks. ✓
- Honest counts: 5 live (what `01-desktop-smoke.ad` actually runs), 20 contract, 11 denial, 18 gaps grouped in #1915 (exists, open). 5+20+11+18=54. ✓
- The `ios.yml` `paths-ignore` additions follow the exact pattern of the existing `android-**`/`web-e2e/**` entries and sit under the #1857 routing gate's enforcement (`check:gate-manifest` green). Verified this is family-path territory the gate owns, not the escape hatch. ✓

**Required — the fourth copy has arrived.** `buildCoverageClassificationSummary` is now duplicated verbatim in Android, web, tvOS (#1919, in flight), and here. The #1426 ground rule says hand-rolled is fine *until a fourth copy makes it painful* — that condition is now met, and macOS will be the fifth. Since #1919 and this PR are both open, coordinate: whichever merges second folds all four copies into a shared `test/integration/support/coverage-classification.ts` (pure ~25-line function; row-type unions stay per-platform). Flagged the same on #1919 — one of the two PRs owns it, not neither.

**Cut — the `linux.yml` step is a duplicate execution.** `smoke-linux-coverage.test.ts` already runs on every PR via the `test:integration:node` glob in ci.yml, whose `paths-ignore` is docs-only — there is no PR that triggers `linux.yml` but skips ci.yml. Running it again before Xvfb adds a redundant execution with no extra guarantee, which is the A2 shape (a lane that reruns what already ran on the same commit). Drop the step, or state in the workflow comment what it catches that the integration lane doesn't.

**Nit:** the gate hand-rolls a `LINUX_DESKTOP` device literal while `src/__tests__/test-utils/device-fixtures.ts` already exports `LINUX_DEVICE` (the tvOS gate reuses its shared fixture). Use the fixture unless its shape genuinely doesn't fit.

thymikee:
Reviewed at `d2ddf1677ef1c50a16a1ba2256211915aeecadfd` against #1426 and #1915 — clean. The manifest exhaustively classifies all 54 public commands; the 5 live claims match the existing Linux replay, contract and denial rows point to the existing Linux evidence and descriptor matrix, and all 18 gaps match the single scoped follow-up. The catalog gate is omission-sensitive and CI routes it through the Linux lane while preserving iOS path isolation. Linux smoke and every completed check are green. Code review is ready; iOS Smoke is still in progress, so merge readiness awaits that check.

thymikee:
Follow-up at `d2ddf1677ef1c50a16a1ba2256211915aeecadfd`: iOS Smoke has cleared, all checks are green, and the PR is now merge-ready.

thymikee:
Addressed in `5db8b0e78`:

- Centralized only `buildCoverageClassificationSummary` in `test/integration/support/coverage-classification.ts`; Android, web, and Linux still keep separate hand-rolled row unions.
- Removed the duplicate Linux workflow execution; the existing Node integration PR lane already runs `test/integration/*.test.ts`, including the Linux gate.
- Reused the canonical `LINUX_DEVICE` fixture.
- Re-ran the Android/web/Linux coverage gates, `pnpm check:quick`, `pnpm check:gate-manifest`, and `pnpm check:affected --run`; all runnable checks passed.

The open tvOS branch should import the shared helper before it merges.

thymikee:
Re-reviewed at `5db8b0e78`: the shared classification helper is the right small cut, platform row types remain local, and Linux now reuses the canonical fixture without a duplicate workflow execution. The 54-command gate and its #1915 gaps remain exhaustive and omission-sensitive. All checks are green; merge-ready.

PR #1924, opened 2026-08-20
fix(ios): report an unobserved text commit instead of a partial success

### LINKED ISSUE #1844: iOS simulator smoke lane flaps on text entry (lost/reordered filled prefix, XCTest keyboard wait) — not reproducible locally under load
## Summary

On 2026-08-18 the `iOS` workflow on `main` went from green on every run through 12:12Z to failing intermittently (fail/pass/fail) with three different text-entry signatures, on commits that do not touch the Apple runner or text entry. A local reproduction of the exact failing sequence under heavy CPU load passed 9/9, so this is filed as a runner-environment / timing flake with evidence, not as a confirmed product defect. No code change is proposed here.

## CI signatures (all GitHub macOS runners, `IOS_RUNTIME_VERSION: 26.2`)

| When (UTC) | Ref | Lane | Failure |
| --- | --- | --- | --- |
| 12:33 | main `3908559fe` (run 32137419756) | Smoke `smoke:automation-input` | `AssertionError: step: wait for Agent Device Tester` |
| 13:36 | main `d0d5c8594` (run 32143319393) | `live iOS simulator fixture E2E`, `test/integration/ios-simulator-e2e/live-runner.ts:216` | `typed email suffix should be observable` — expected `ada@example.test`, `get attrs id="field-email"` returned `"value":".test"` |
| ~13:50 | PR #1826 head `a3388f8e3`, run 1 (pure TS test move, no runner changes) | runner XCTest | `RunnerTests.testBareTypeUsesTappedInputWhenSoftwareKeyboardIsHidden` — runner log spinning on `Checking existence of 'Keyboard (First Match)'` |
| ~14:05 | PR #1826 same head, run 2 | fixture E2E | same `typed email suffix` assertion, value `adexamplea@.test` (filled prefix with characters reordered) |
| ~14:20 | PR #1826 same head, run 3 | — | pass |

Same commit, three outcomes; `main` interleaved passes (12:56 `60f6356b0`, 13:56 `9a0d6dead`) with the failures.

Scenario under test (`assertFormInput`): `fill` email `ada@example` → `keyboard dismiss` (asserted dismissed=true) → `snapshot -i` → `focus <x> <y>` on the email field's rect → `type '.test'` (asserted `textEntryRoute: 'synthesized-first-responder'`) → `get attrs id="field-email"`.

Commits merged between the last green (`2b6d04a13`, 12:12Z) and the first red: #1818 (daemon-stop claim report, TS), #1808 (Android), #1820 `fix(scroll): keep saturated scroll gestures out of the status bar` (iOS scroll only). Later that hour: #1827, #1825, #1814 (TS only). Earlier the same day: #1807, #1811, #1816, #1798 (iOS snapshot/runner).

## Local reproduction (does NOT reproduce)

iPhone 16 simulator (`416440AE-56FE-43D3-9349-D6CE8D1B1895`), repo CLI built from `main`, isolated `--state-dir`, the fixture app the smoke lane uses, the exact sequence above scripted, with heavy CPU load generators running alongside:

| Iteration | after `fill` | after `keyboard dismiss` | after coordinate `focus` + `type` | pass |
| --- | --- | --- | --- | --- |
| 1–9 | `ada@example` | `ada@example` | `ada@example.test` | 9/9 |

`type` reported `textEntryRoute: synthesized-first-responder` in every iteration; `keyboard dismiss` reported `mechanism: dismissKey`.

## What that rules out / leaves open

- Rules out (locally): coordinate `focus` clearing an already-filled field; `keyboard dismiss` clearing the field; the synthesized `type` route replacing instead of appending; CPU contention alone dropping/reordering the `fill` characters.
- Still open: (a) the runner's software keyboard never appearing / appearing late on the CI image (the XCTest signature and the `wait for Agent Device Tester` step both point at slow UI on the runner); (b) synthesized-type commit timing under runner-specific latency — see #1676 (private text synthesize returns at post-time, not commit-time; observation-only commit wait; tail re-post rejected because it double-posts). The `adexamplea@` value (reordered, not truncated) is the strongest hint that characters were committed out of order on the runner, which #1676's observation-only wait would not prevent.

## Suggested next step

Capture the runner-side evidence the lane already produces on failure (runner.log + the request ndjson) as workflow artifacts for these two steps, and correlate the per-character post timestamps against the observed value; only then decide whether text entry needs a commit-ordering guarantee. Do not harden the test or add retries (per #1676's lesson).

## Repro commands

See `test/integration/ios-simulator-e2e/live-runner.ts` `assertFormInput`; locally: `pnpm build`, boot iPhone 16, `open <fixture bundle> --platform ios --udid 416440AE-… --state-dir <tmp>`, then `fill 'id="field-email"' 'ada@example'`, `keyboard dismiss`, `snapshot -i`, `focus <cx> <cy>`, `type .test`, `get attrs 'id="field-email"'`; run under `yes > /dev/null &` ×N for load.

### LINKED ISSUE #1874: ios: text-entry lanes fail intermittently on unrelated branches (unpinned simulator keyboard state)
> **Status (2026-08-26).** Both filed symptoms are fixed and merged (#2035, `7db5ad73dd`). One of
> them is genuinely root-caused; the other is *mitigated*. This issue stays open for the mitigated
> half — why the simulator input pipeline throttles at all — which is the `needs-info` the label
> refers to. The original "Root cause" and "Fix" sections below were wrong in both of their
> mechanisms; they are kept, marked, under [Superseded hypotheses](#superseded-hypotheses), with the
> measurements that disproved them.

The targeted iOS XCUITest and the fixture-backed iOS smoke both fail intermittently on branches that touch no Swift and no iOS production code, blocking the most expensive PR gate.

Two symptoms — and, it turned out, **two unrelated causes**:
- `RunnerTests.testBareTypeUsesTappedInputWhenSoftwareKeyboardIsHidden` — `** TEST EXECUTE FAILED **` / exit 65
- `smoke:form-input` → step `fill full name` → `TEXT_INPUT_COMMIT_NOT_OBSERVED`

| run | branch | when |
|---|---|---|
| `32229709270` | `ci/1824-coverage-fork-death` (#1854 — test-setup, docs, vitest configs only) | 2026-08-19 |
| `32176591415` | `agent/fix-install-source-network-cause` (#1863) | 2026-08-18 |
| `32847092583` (both attempts) | `claude/mcp-registry-security-el547d` (#2023 — MCP/release only, 16 files) | 2026-08-25 |
| `32655577032` (attempt 1 only; attempt 2 green on the same SHA) | `claude/monorepo-modularity-analysis-t1kd3c` (#1985 — one TS module renamed, three imports retargeted) | 2026-08-23 |
| `32889322172` | `claude/1874-text-entry-commit-budget` (#2035, this issue's own fix branch — the run that produced the decisive trace) | 2026-08-25 |

### `smoke:form-input` — resolved, and it was never a race

**The fixture asked `fill` for the one value it cannot verify.** `field-name`'s placeholder was
`"Ada Lovelace"`, and every checkout-form suite filled exactly `"Ada Lovelace"`; `field-email` had
the same collision with `ada@example.com`. Twelve fills across eight files.

An empty text field renders its placeholder **as** its accessibility value, so `element.value` is
byte-identical whether the placeholder is showing or the committed text happens to equal it, and
`placeholderValue` does not disambiguate. `awaitSynthesizedReplacementCommitOutcome` therefore
refuses such a wait outright (`textMatchesPlaceholder`) rather than reporting a commit it cannot
prove — the documented contract, see the `TEXT_INPUT_COMMIT_NOT_OBSERVED` paragraph in
`website/docs/docs/commands.md` and the three-state model in #1924.

Caught from run [`32889322172`](https://github.com/callstack/agent-device/actions/runs/32889322172),
where the trace is unambiguous:

```
[DEBUG-1874] wait start expectedLen=12 route=replacement          19:35:00.652
   (no poll lines at all)
[DEBUG-1874] wait outcome=notObserved elapsedMs=3608 route=replacement   19:35:04.260
```

Zero poll lines: the wait never entered its loop. `expectedLen=12` is `"Ada Lovelace"`. The 3.6s is
the AX resolution for the placeholder read that precedes the guard, not a deadline.

**Why it looked intermittent.** The synthesized-replacement route is gated on
`xCTestChannelPenalized` — it is taken only when the host is loaded. A deterministic refusal on a
load-gated route is indistinguishable from a load flake from the outside. It is also what actually
explains the datapoint this issue found most puzzling: `rerun_failed_jobs` on `32847092583`
reproducing *identically*. That is not persisted per-device state; it is the same guard firing on
the same input.

**Fixed at the fixture** (placeholders changed), not at the filled values — frozen
`test/replay-compat/scripts/**` corpora carry the same fills and must not be edited. The collision
had also been making the read-back assertions vacuous: `assertJsonContains(name, 'Ada Lovelace')` is
satisfied by an empty field rendering the placeholder.

### `testBareType…` — mitigated, not root-caused

The commit wait used a flat 3s deadline, which cannot tell a **throttled** input pipeline
(characters keep landing, slowly) from a **wedged** one (nothing lands). It condemned both at the
same instant, so a working `type` reported `TEXT_INPUT_COMMIT_NOT_OBSERVED`.

The wait now grants time against *progress* — the observed value's expected-prefix growing, the same
length-only evidence `logCommitCadence` already emitted — with an absolute 10s ceiling. A pipeline
making no progress still expires at exactly the 3s the flat deadline used, so nothing that failed
before starts passing merely by waiting longer.

**This reduces exposure; it does not explain the throttle.** The ~40s episode traced in the
2026-08-21 diagnosis comment below would still exhaust the ceiling and report honestly.

## Superseded hypotheses

Both mechanisms originally filed here are disproven. Recording them so they are not re-derived:

**1. "Nothing pins the simulator keyboard state."** Measured on a dedicated simulator, headless
`simctl boot`, iOS 26.2, then deleted:

| per-device `ConnectHardwareKeyboard` | Settings search field, after tap |
| --- | --- |
| `true` (hardware keyboard "connected") | full software keyboard |
| `false` | full software keyboard |

`com.apple.iphonesimulator` is Simulator.app's preference domain; with no Simulator.app running,
CoreSimulator does not consult it. Pinning it would have added a step, a guard test and a green
assertion proving only that `defaults write` ran. The useful half of the finding is the reassuring
one: **a headless lane has no hardware keyboard to inherit**, so the ambient state named in this
issue is not a variable in these lanes at all.

**2. "Text-entry readiness depends on that ambient state."** Neither failing signature reaches
`waitForTextEntryReadiness`:

- `testBareTypeUsesTappedInputWhenSoftwareKeyboardIsHidden` types through the tap witness, and
  `rememberedTextEntryTarget()` returns before readiness is ever consulted.
- `smoke:form-input`'s `fill` fails inside `runSynthesizedReplacementRoute`, which
  `shouldUseSynthesizedFirstResponderReplacement` gates on `hasResolvedElement == false` — the
  penalized coordinate route builds that target directly in `executeTypeCommand`, bypassing
  `focusTextInputForTextEntry` entirely.

The readiness fallback was hardened anyway (it now returns only on confirmed focus of the target,
re-arming otherwise), because the latent sensitivity described here was real — it just was not what
was firing.

## What shipped

#2035 (`7db5ad73dd`):

- progress-aware commit budget (`SynthesizedCommitDeadline`), stall budget unchanged at 3s, ceiling 10s;
- readiness's hardware-keyboard fallback returns only on confirmed focus of the target element;
- `testBareTypeUsesTappedInputWhenSoftwareKeyboardIsHidden`'s keyboard-hidden precondition is
  `XCTSkipIf`, not `XCTAssertFalse` — an environment flip must not read as a product regression;
- the two fixture placeholders.

## What is still open

**Why does the simulator input pipeline throttle?** Unchanged by the above. What is different is
that the evidence to answer it is now on the shipped path rather than needing new instrumentation:
the `[DEBUG-1874]` per-poll cadence discriminates three shapes, and after #2035 the same
expected-prefix signal *decides* the deadline rather than merely being logged.

| cadence shape | meaning |
|---|---|
| prefix grows slowly, then completes | throttle — absorbed by the budget, no red at all |
| prefix freezes | wedge — condemned at the 3s stall budget, reported honestly |
| no poll lines | the placeholder guard fired before polling (unreachable in this fixture now) |

So the next red lane should be readable from `runner.log` in the uploaded `ios-artifacts` bundle
without adding anything. `1874-diagnose.yml` (matrix `isolated`/`pair` × `x86_64`/`arm64`) remains
dispatchable for forcing the question rather than waiting for it.

Related: #1924 (honest reporting, and the three-state placeholder model), #1941 (commit-wait
instrumentation), #1995 (fill's synthesized-replacement verification), #2032 (folded in here),
#2053 (a different Coverage-lane flake seen while working on this, not this issue).

thymikee:
Refs #1874, #1844.

## Summary

Stops the iOS synthesized bare-`type` route from reporting success when the app has not observably committed all requested text.

- Distinguishes a settled commit from a strict-prefix wait that reaches its three-second deadline. The latter now returns `TEXT_INPUT_COMMIT_NOT_OBSERVED` with recovery guidance to read the field and prefer `fill` over appending another `type`.
- Observes once more before condemning the final polling interval, so a commit that lands during the last sleep is not falsely rejected.
- Treats a final value equal to the field placeholder conservatively. Accessibility cannot distinguish committed text from an empty field rendering that placeholder—even when the field held content before dispatch, because an input handler may clear it while processing the event.
- Reuses the existing placeholder-normalization predicate instead of maintaining a second equality rule, and removes the superseded baseline-dependent evidence enum/raw-read path.
- Keeps submit-key and unreadable-field behavior unchanged (`.unobservable` remains non-failing).

This fixes #1874's strict-prefix signature (`"h"` remaining from `"hardware-keyboard"`). It does not claim to fix #1844's reordered/non-prefix text, which follows the existing transformed-input path. The sibling synthesized replacement and keyboard-visible routes remain outside this observation mechanism because they do not retain a resolvable element.

## Validation

- Red-before on the clear-after-dispatch regression: the old non-empty-baseline model returned no failure when the post-dispatch AX value was the placeholder (`nil` vs `TEXT_INPUT_COMMIT_NOT_OBSERVED`).
- Green on the existing iPhone 17 Pro simulator after the fix: the clear-after-dispatch regression and the nearest ordinary-placeholder negative case both passed (2 tests, 0 failures). Eight adjacent commit-policy tests also passed.
- Removing the conservative guard made the regression fail again with two assertions: missing `TEXT_INPUT_COMMIT_NOT_OBSERVED` and one forbidden post-dispatch observation.
- The focused iOS runner build completed with `TEST BUILD SUCCEEDED`; `pnpm check:xctest-selection` reports 185 declared tests and 0 unreachable.
- `pnpm check:affected --run` passed at `a58410266`; GitHub-authoritative Swift, coverage, and device lanes are running on that exact head.

Docs were updated because this changes when `TEXT_INPUT_COMMIT_NOT_OBSERVED` is returned and how callers should recover. No skill guidance changed; the command reference owns this behavior.


thymikee:
Adversarial review pass (Fable). No blocker, but two deterministic false-failure windows in my own wait, one overclaim in the PR framing, and an untested branch. All addressed in `03a0a45b6`; the review also refuted several things I want on the record.

## Fixed

**1. The deadline was checked before observing.** A commit landing during the final poll sleep was condemned as `.notObserved` — under exactly the loaded-host timing this wait exists for, so the fix could have added red to the lane it came from. The check now runs *after* an observation, so the last thing before condemning is a read. `testCommitLandingDuringTheFinalSleepIsStillObserved` is red against the ordering I first pushed (verified by running both loop structures side by side: old → `notObserved`, new → `settled`).

**2. Placeholder collision.** `treatingPlaceholderAsEmpty` maps a value equal to the field's placeholder to `""`, which is a prefix of every expected value. `type "0.00"` into a field placeheld `"0.00"` committed instantly, read as pending for the full 3s, then failed. The observation now settles on an exact raw match; the normalized read still drives the prefix walk. Same trap for `"Search"`/`"Search"` and the generic `"Text input field"` label.

**3. The refusal branch was untested.** The whole fix was one ternary that no test reached. It is now `textEntryFailure(forCommitOutcome:)` with assertions on all three cases — including `.unobservable → nil`, which is what keeps `type "...\n"` working and would have failed silently if inverted.

**4. Message asserted a field state never read.** Under both windows above the field may hold the *complete* value, so "committed only part" was wrong. Now "could not confirm the typed text committed before its deadline".

## Corrected framing — this does not catch #1844's signatures

I overclaimed. Both #1844 values classify as `.diverged`, not `.pending`, so they settle and still return ok:

| observed | expected | class |
| --- | --- | --- |
| `.test` | `ada@example.test` | diverged → settles, stays ok |
| `adexamplea@.test` | `ada@example.test` | diverged → settles, stays ok |
| `h` | `hardware-keyboard` | pending → now fails |

Only the strict-prefix stall (#1874's signature) becomes a failure. #1844's lost/reordered prefix is a different mechanism and this PR does not address it. The docs sentence has been narrowed to match, and no longer implies every text-entry route verifies its result.

## Known, unchanged

- **The sibling routes still return `verified: nil` without observing**: `runSynthesizedReplacementRoute`, and the keyboard-visible synthesized branch. Neither has a resolvable element to observe, so the mechanism cannot apply as-is — noted rather than silently left.
- **Paced typing (`--delay-ms`) can now abort mid-string.** The commit wait runs per character, so a >3s stall on character 1 fails the command with characters 2..n never posted — where before every character was posted and a slow app eventually drained the queue. I believe failing is right (the alternative is a success over an unconfirmed field), but it is a real semantics change on that route and was unstated. Worst-case latency is unchanged; the per-character wait predates this PR.

## Refuted — checked and sound

- **The `dispatched` deletion.** All seven pre-change return paths enumerated including the `#else`/macOS branches: `dispatched` was `true` iff `failure == nil` everywhere, and `.raise` never returns. The four `?? .notFocused` fallbacks were dead code.
- **Secure fields** → `editableTextValue` nil → `textBefore` nil → `.unobservable` → no failure, exempt by construction.
- **Elements that stop resolving mid-wait** → nil → `.diverged` → settles.
- **Formatters / autocomplete / card spacing** → non-prefix → diverges immediately.
- **A maxLength field that truncates** does now fail where it used to succeed — but truthfully, since the field genuinely lacks the requested text.
- **Error-code registration**: runner codes pass through `ErrorPayload` generically; the two sibling codes have zero TypeScript references, so there is no registry to miss. ADR-0010 needs code+message+hint, all present.
- **The hint's `fill` advice**: `fill` uses replacement mode, never enters the fixed branch, and surfaces a dropped character as `TEXT_ENTRY_MISMATCH`.

`pnpm gate swift-runner-macos` builds clean; `pnpm check:affected --run` passes; `check:xctest-selection` now reports 166 methods, 143 host-lane-reachable, 0 unreachable.

thymikee:
[P1] Make the placeholder regression cover the production normalization boundary. `testValueEqualToThePlaceholder…` injects an `observe` closure that already returns `"0.00"`, so removing the new raw-value short-circuit in `awaitSynthesizedFirstResponderCommit` leaves this test green—the test never supplies the normalized `""` that causes the failure. Extract the raw-exact/normalized-prefix choice into a small pure helper and test it with `(raw: "0.00", normalized: "", expected: "0.00")`, or exercise it with a fixture element. That way reverting the fix goes red.

thymikee:
[P1] addressed in `40c70f51e`. You were right, and the failure mode was worse than "weak test": I had claimed in the review-response comment above that it pinned the behavior, and it could not.

**Confirmed first.** The fix lived at `RunnerTests+SynthesizedTextEntry.swift:271` inside the production `observe` closure; the test injected its own `observe` returning `"0.00"` and never reached that line. Reverting the short-circuit left it green.

**Now.** The raw-exact/normalized-prefix choice is `commitObservation(rawValue:expectedText:normalizedValue:)`, driven with exactly the case you named:

```swift
XCTAssertEqual(
  Self.commitObservation(rawValue: "0.00", expectedText: "0.00", normalizedValue: { "" }),
  "0.00")
```

Red-before proved by running both implementations against the assertion set:

```
with the fix   : GREEN
fix reverted   : RED — exact-match case
```

Three more cases pin the other direction, so the helper cannot be "fixed" by always returning the raw value: a non-exact match keeps the normalized reading (`raw "0.00"` / `expected "0.005"` → `""`), the prefix walk still works, and `nil` stays `nil`.

**One deviation from your suggestion.** `normalizedValue` is a closure, not a value. Taking both readings eagerly would make every poll cost two accessibility reads on a path that polls every 20ms for up to three seconds; the closure keeps an exact match at one read. `testCommitObservationDoesNotPayForASecondReadOnAnExactMatch` pins that, so the laziness cannot be refactored away silently. If you would rather have the simpler eager signature and accept the extra read, say so and I will flip it.

I did not take the fixture-element option — it would put this on the simulator lane, where the host lane now covers it in 12ms per PR.

The old test is deleted rather than kept alongside: its only other assertion (an exact match settles without polling) is already `testSynthesizedCommitStopsAtTheFirstSettledObservation`.

`pnpm gate swift-runner-macos` clean, `pnpm check:affected --run` passes, `check:xctest-selection` 167 declared / 144 host-reachable / 0 unreachable.

thymikee:
P1: Do not use placeholder equality as commit evidence.

`editableTextValue(..., treatingPlaceholderAsEmpty: true)` deliberately classifies a raw value equal to `element.placeholderValue` as empty. For an empty field whose placeholder is `0.00`, `textBefore` is therefore `""`; after dispatching `type "0.00"`, the first read can still be the unchanged placeholder `0.00`. `commitObservation` treats that raw equality as committed before normalization, so the wait settles and the command returns `ok: true` without evidence that any synthesized character committed. That reintroduces the success-misdescribes-device-state failure this PR is intended to remove.

Please keep this state indeterminate/non-success (or obtain a signal that distinguishes placeholder rendering from a committed value) and add a regression where the raw placeholder already equals the requested text before any commit; it must not report success.

thymikee:
P1 addressed in `7cedfc59b`; branch rebased onto `46eff36f8`.

You are right, and the direction matters: my fix for a false *failure* created a false *success*, which is strictly worse and is the exact defect this PR exists to remove.

**Confirmed the mechanism before changing anything.** Empty field, placeholder `0.00`, `type "0.00"`:

```
textBefore (normalized) =           <- placeholder classified as empty
expectedText            = 0.00
first read, nothing committed: raw = 0.00   <- the placeholder rendering
commitObservation: raw == expected -> 0.00 -> .committed -> .settled -> ok:true
=> success reported with ZERO characters committed
```

**No disambiguating signal exists.** `element.value` is byte-identical whether the placeholder is rendering or the committed text happens to equal it, and `placeholderValue` only tells you what the placeholder is, not which of the two states you are in. So I took your first option rather than the second: the state stays non-success.

**`commitObservation` is deleted, not narrowed.** The raw match was only ever consulted in precisely this case, and in precisely this case it is not evidence — so there was nothing left worth keeping. The observation is the plain normalized read again, as before that revision.

**The refusal is now up front, not after the deadline.** Since no read can resolve the ambiguity, spending 3 seconds discovering that is pure latency. `placeholderMakesCommitUnobservable` detects it before the wait starts and reports `.notObserved` immediately.

**Regression, as you asked** — raw placeholder already equals the requested text before any commit, must not report success:

```swift
XCTAssertTrue(
  Self.placeholderMakesCommitUnobservable(placeholder: "0.00", expectedText: "0.00"))
XCTAssertEqual(
  Self.textEntryFailure(forCommitOutcome: .notObserved), .commitNotObserved)
```

Red-before proved against the previous revision on the zero-characters-committed scenario:

```
previous revision: ok:true (SUCCESS)
now:               TEXT_INPUT_COMMIT_NOT_OBSERVED
```

A second test keeps the guard narrow — `placeholder "0.00"` / `expected "0.005"`, `placeholder "Email"` / `expected "ada@example.test"`, nil, empty and whitespace-only placeholders all stay observable. Widening it would refuse ordinary typing into any placeheld field, which is most of them.

**Two knock-on corrections.** The failure message no longer says "before its deadline" (this refusal never waits), and the docs now name both routes to the code — deadline expiry, and placeholder-identical text.

**The eager-vs-lazy question from my last comment is moot** — the helper it applied to is gone.

Rebased onto `46eff36f8`; verified line-level that nothing from `main` was reverted (`git log origin/main ^HEAD` empty, all deletions attributable to this branch). `pnpm gate swift-runner-macos` clean, `pnpm check:affected --run` passes, `check:xctest-selection` 184 declared / 161 host-reachable / 0 unreachable.

thymikee:
[P1] Keep the placeholder guard scoped to an empty baseline. `placeholderMakesCommitUnobservable` currently receives only the placeholder and final expected text, so it refuses any append whose resulting value happens to equal the placeholder. For example, with actual value `"0"` and placeholder `"0.00"`, `type ".00"` computes expected `"0.00"` and returns `TEXT_INPUT_COMMIT_NOT_OBSERVED` before a read, although the non-empty `textBefore` proves a later `"0.00"` is commit evidence rather than the initial placeholder rendering. The indistinguishable state exists only when the normalized pre-dispatch value is empty. Gate this refusal on that baseline as well, and add a regression for `"0"` + `".00"` that reaches the observation path and settles.

thymikee:
P1 addressed in `afe99bd15`.

Right again — I had made the guard baseline-blind, so it refused the very case where the read *is* trustworthy. Value `"0"` + `type ".00"` against placeholder `"0.00"` was refused before a single read.

**The baseline is the deciding input, so it is now an input.** `placeholderCommitEvidence` returns three states instead of a boolean:

| state | when | behaviour |
| --- | --- | --- |
| `normalRead` | expected differs from the placeholder | placeholder never enters into it |
| `indistinguishable` | expected IS the placeholder **and** the field was empty | refuse, before the wait |
| `rawValueIsEvidence` | expected IS the placeholder **but** the field held content | raw match settles it |

Only `.indistinguishable` refuses. `.rawValueIsEvidence` reaches the observation and settles on the raw match, which the normalized read would otherwise hide — that is your `"0"` + `".00"` case.

`commitObservation` is back for that third state only, scoped by evidence rather than applied unconditionally as in the revision that turned raw equality into a false success.

**Both directions proved, so neither P1 regresses the other:**

```
Case A — empty field, placeholder "0.00", type "0.00" (nothing committed):
   baseline-blind (previous): REFUSED before any read
   baseline-scoped (now):     REFUSED before any read     <- still correct

Case B — value "0", type ".00", placeholder "0.00" (commit DOES land):
   baseline-blind (previous): REFUSED before any read     <- the defect
   baseline-scoped (now):     settled -> ok
```

**The regression you asked for** reaches the observation path rather than asserting on the classifier alone: `testAppendWhoseResultEqualsThePlaceholderSettlesThroughTheObservation` drives `awaitSynthesizedCommitOutcome` with `rawValue` walking `"0"` → `"0.00"` and `normalizedValue` pinned at `""` throughout. Because the normalized reading never matches, only the evidence-scoped raw match can settle it — so the test goes red both if the guard reverts to baseline-blind and if `commitObservation` ignores `evidence`.

`testPlaceholderEvidenceDoesNotFireOnOrdinaryTyping` now runs its cases under **both** baselines, so the narrowness is pinned independently of the new input. `testOrdinaryObservationDoesNotPayForTheRawRead` keeps `normalRead` at one accessibility read per poll.

Docs updated to name the empty-field condition rather than implying any placeholder-identical text refuses.

`pnpm gate swift-runner-macos` clean, `pnpm check:affected --run` passes, `check:xctest-selection` 187 declared / 164 host-reachable / 0 unreachable.

thymikee:
**[P1] A non-empty baseline is not post-dispatch commit evidence.** `rawValueIsEvidence` assumes that because the field held `"0"` before `type ".00"`, a later raw `"0.00"` must be committed text rather than the placeholder. But an input handler can clear the field while processing the synthesized event; the now-empty field renders the same raw `"0.00"` placeholder, and `commitObservation` reports success over an empty field. The baseline proves only what rendered before dispatch, not what the later AX read represents. Add that clear-after-dispatch regression and require `TEXT_INPUT_COMMIT_NOT_OBSERVED`; without a post-dispatch discriminator, keep placeholder-equal reads conservative.

thymikee:
P1 addressed in `a58410266`. The non-empty-baseline/raw-value evidence state is removed: a placeholder-equal final value is now conservatively `TEXT_INPUT_COMMIT_NOT_OBSERVED` before any post-dispatch poll, because a handler may clear the field and expose the same placeholder rendering. The wait and ordinary text normalization share one placeholder matcher. The clear-after-dispatch case was red on the old model, green after the fix, and red again when the guard was deliberately removed. Final iOS simulator run: 2 focused tests, 0 failures; adjacent policy run: 8 tests, 0 failures. `pnpm check:affected --run` passed before push, and exact-head CI is running.

thymikee:
Re-reviewed exact head a58410266: the clear-after-dispatch P1 is resolved. Placeholder-equal expected values are now conservatively indeterminate before polling, and the new regression proves the old non-empty-baseline/raw-match path would report success over an empty field. The obsolete evidence/raw-read branch was removed and the docs match the conservative contract. Completed checks are green; one iOS Smoke job is still running.

thymikee:
Exact head a58410266 is now fully green (30/30) and mergeable/CLEAN. The conservative placeholder-equality fix and its regression remain review-clean. Merge-ready.

PR #1927, opened 2026-08-20
perf: reduce cold iOS runner startup latency

thymikee:
## Summary

Reduce cold iOS simulator startup latency without changing snapshot backends.

- reuse the existing bounded fresh-boot observation instead of immediately paying for a duplicate `simctl list devices -j`; stale booleans and memo failures still fall back to native discovery
- wake one runner connection retry from the runner's authoritative NWListener-ready marker, while keeping the HTTP probe authoritative
- extract xcodebuild process launch/output ownership from the oversized runner session module and make retry wakeups cancellation- and resource-safe

On the same existing iPhone 17 Pro simulator, cold `open` plus the first interactive snapshot improved from 5.45s to 3.83s. Retained runner preparation remained effectively instant (14ms internally), and a warm snapshot measured 185ms.

This does not make private AX the default. The current private-AX backend still runs inside XCTest, so it cannot remove XCTest startup.

## Validation

- `pnpm check:affected --run` — 342 files / 2,533 tests passed, plus format, lint, typecheck, layering, fallow, and build
- live isolated cold/retained/adopted startup matrix on the existing iPhone 17 Pro simulator
- verified no simulator was created or deleted and the isolated runner/daemon were stopped

### DIFF
--- a/packages/contracts/src/device-readiness-runtime.ts
+++ b/packages/contracts/src/device-readiness-runtime.ts
@@ -18,6 +18,8 @@ export type DeviceReadinessRuntimeHost = Readonly<{
   }>;
   appleAutomation: Readonly<{
     keepHot(device: DeviceInfo): void;
+    /** Reads the bounded memo populated only by a fresh native Booted observation. */
+    wasRecentlyObservedBooted(device: DeviceInfo): Promise<boolean>;
     /**
      * Publishes a FRESH Booted observation. `simctl list devices -j` costs ~0.7s per spawn and one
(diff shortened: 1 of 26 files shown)


thymikee:
Reviewed `6a6efce66c8d8f7031fc771ac2dd69c6a5caace4`: clean. The readiness memo preserves native fallback, and the NWListener marker only shortens one retry while the HTTP probe remains authoritative. The launch extraction also gives xcodebuild ownership a clearer seam. Remaining CI evidence: iOS Smoke is in progress.

PR #1929, opened 2026-08-21
feat(ios): move the regular-projection clip fold into snapshot presentation (#1797)

thymikee:
## Summary

Next #1797 delta (after the raw/regular split, #1926): **the clip fold moves into presentation.** Both iOS backends carried their own copy of the visibility fold — the tree walker and the private-AX serializer each computed viewport∩scroll-clip intersection, ancestor projection, hidden-content hints, and collapsed depth *during acquisition*. Hand-synchronized copies of that interpretation are the mechanism behind the scroll-overflow leak class (#1784 was fixed by making one copy call the other's helpers — correct, but nothing stopped the next drift), and C1 (fact-availability neutrality) cannot hold while acquisition decides what a screen shows.

**After this PR, acquisition backends are fact serializers.** Every traversed node is emitted at raw traversal depth with its reported frame; `SnapshotAcquisition` carries the viewport; `presentRegular` runs the one fold for every backend:

- viewport ∩ scroll-container clip, with the ancestor projection cursor (an out-of-clip `Cell`/scroll container hides descendants whose clamped frames would otherwise leak back into the viewport — #1784's shape, now owned by one implementation),
- the sub-pixel decoration rule,
- hidden-content hints booked onto scroll anchors,
- reparenting of survivors with collapsed depth,
- and a new invariant: **presented `hittable` is narrowed to the clip** — nothing outside its clip, and nothing without geometry, is ever hittable in a regular snapshot, whatever the backend reported.

Platform differences enter as a `SnapshotFoldPolicy` **input** to the shared algorithm (iOS cursor-projected; macOS/tvOS plain viewport intersection) — per the issue's rule that macOS is a platform policy, never a backend exception. The private-AX backend collapses to **one serializer for both projections**, and the flat filter-decision family (`FlatSnapshotFilterNode`/`FlatSnapshotVisibilityPolicy`) is deleted with the acquisition gates it fed.

## Behavior

**Live A/B, byte-identical.** Settings on iPhone 16 Pro / iOS 26.2, this branch's runner vs a runner built from `main`, each on its own state dir and daemon, compared on full node signatures (type, label, identifier, value, rect, depth, parentIndex, hittable, both hidden-content hints):

| projection | main | branch | |
|---|---|---|---|
| tree `snapshot` | 73 | 73 | identical |
| tree `snapshot -i` | 18 | 18 | identical |
| private-AX `snapshot --actions` | 58 | 58 | identical |
| private-AX `snapshot -i --actions` | 20 | 20 | identical |

**Three intentional edge deltas**, each replacing a per-backend rule with one backend-neutral rule (none fired on the live screen above):

1. Sub-pixel content-free decorations now drop on **every** backend (was private-AX only). Only observable for interactive-typed contentless degenerate nodes on the tree backend.
2. Labeled offscreen `Application`/`Window` carriers now survive on **every** backend (was tree only) — still never hittable.
3. A query-sweep recovery snapshot **without** `-i` is now viewport-folded like everything else (it used to list offscreen elements with `hittable: false`).

**Declared acquisition residues** (budget/fidelity, not membership): the traversal-depth cut (regular presents collapsed depth — the open visible-depth frontier obligation, unchanged), the sweep's frameless-element drop (a flat query has no hierarchy for geometryless semantics to attach to), and the private-AX bridge's device-side node cap. Two mechanical notes: the private-AX backend-level sparse gate (`nodes.count <= 1`) now counts serialized rather than folded nodes, so a fully-clipped tree surfaces through the plan's sparse classifier on the presented payload instead of a backend `nil` — same terminal outcome, slightly different verdict path; and children of a dropped duplicate inherit the ancestor cursor rather than the duplicate's own transition (observable only for a duplicate offscreen owned-container whose children aren't themselves duplicates).

## Evidence

- **23 targeted runner tests green** (iPhone 16 Pro sim / iOS 26.2): 5 new fold tests (scroll-overflow clip + hint + reparenting, Window-carrier + hittable-outside-clip, sub-pixel rule, plain-viewport policy, type-name pinning) plus the rewritten private-AX presentation tests now routed through the production `presentRegular`, plus the #1926 suite. macOS bundle compiles with unit tests (host lane doesn't run on this machine — environmental, verified both ways on #1926; CI covers it).
- **Non-vacuity by mutation**: with `presentRegular` bypassing `foldRegularVisibility`, exactly the six fold-dependent tests fail (both new fold tests and both private-AX regular tests — the drawer fixture reproduces the leak: "Admin settings" from the offscreen drawer is presented) while eligibility and scope tests stay green. Restored before commit.
- `pnpm check:xctest-selection` green (172 declared, 0 unreachable); new tests added to the iOS PR lane; `check:affected --run` green; oxfmt/oxlint clean.

## Docs

ADR 0004: the fold as the fourth semantic layer, the policy input, and the declared residues. `CONTEXT.md`: **clip fold** vocabulary. CHANGELOG: user-facing entry with the three edge deltas.

## What #1797 still needs after this

1. **`hittable` = geometric actionability** — the tree backend still computes its occlusion scan at acquisition (`computedSnapshotHittable`); moving occlusion wholly to the daemon annotator is the next delta, with the ~118-reference `hittable`-consumer inventory.
2. **Geometry carrier** (effective rect on the wire) with the rect-consumer inventory — now unblocked, since presentation computes the clip.
3. **Step-4 enforcement**: choke-point invariant, per-PR fixture differentials + nightly live tree-vs-private-AX differential, `tree` pin + penalty-clear hook, parity ratchets.
4. Cross-runtime `RawAXNode → Presented` conformance suite (#1832 step 9).

Refs #1797 (migration step 3, clip-fold delta).


thymikee:
[P2] Please extract `foldRegularVisibility` into a dedicated snapshot-visibility-fold module. This addition takes `RunnerTests+SnapshotPresentation.swift` to 529 lines, past the repository\s extract-before-500 rule, and the fold still relies on `RunnerTests.*` helpers/types scattered across `+Snapshot.swift` and `+FlatSnapshotFiltering.swift`. Put the fold\s policy, transition, clipping, hint, and reparenting ownership behind one pure interface with its matching test file, then have `SnapshotPresentation` call it. That keeps presentation orchestration-only and makes the backend-neutral visibility invariant obvious in code.

thymikee:
Addressed the modularity review in e4b19df6c:

- extracted the backend-neutral visibility fold into a 294-line `SnapshotVisibilityFold` deep module with a matching focused test file;
- moved clipping, ancestor projection, reparenting, scroll anchors/hints, and actionability narrowing behind its single pure interface;
- made `SnapshotAcquisition.viewport` mandatory so regular captures cannot silently bypass clipping through a default;
- removed the superseded `FlatSnapshotFiltering` module and the unused private-AX semantic helper, and isolated private-AX element-type conversion;
- added a presentation-boundary regression so bypassing the fold fails (verified red: the out-of-viewport node leaked; restored green);
- updated XCTest selection for the moved/new tests.

Verification: `pnpm check:affected --run`; iOS and macOS XCTest runner builds; 8 focused iOS tests, 0 failures. This completes the clip-fold/modularity unit of #1797, not the entire issue: effective geometry/hittability ownership and cross-runtime enforcement remain separate semantic follow-ups.

thymikee:
Re-reviewed `e4b19df6c955ee6c91c8ca70798c24b0f76269ce`: the prior ownership/size concern is resolved. The clip fold is now the dedicated 294-line `SnapshotVisibilityFold` module, reached once through `SnapshotPresentation.presentRegular`; its 132-line sibling tests cover clipping, cursor policy, hints, reparenting, carrier hittability, and the type-vocabulary pin. The former flat-filter path and duplicate fold helpers are deleted, and the capture plan sends every backend through the shared presentation route.\n\nCode review is clean. iOS Smoke, host XCTests, static/layering/fallow, and Size are green; documented live A/B evidence covers the behavior. Coverage is the sole red check: an unrelated `scripts/fuzz/corpus-replay.test.ts` Vitest worker-termination timeout after 1051/1052 files passed.

PR #1930, opened 2026-08-21
feat(ios): add snapshot backend conformance

### LINKED ISSUE #1635: iOS: backend parity gate — force-captured conformance smoke + declared capability matrix for XCTest vs private-AX
## Problem

The iOS snapshot chain has multiple capture producers (XCTest recursive-tree, XCTest query-sweep, private-AX bulk) behind one consumer contract (`SnapshotNode` + plans + quality verdicts). The consumer side is single-implementation, but the producers drift, and today nothing makes drift visible:

- The frontier deep extension (#1627) exists only in private-AX; XCTest truncation behaves differently.
- The depth ladder + accepted-depth memory are private-AX-only.
- `hittable` is real hit-testing under XCTest but approximated (visible+enabled+role-candidate) under private-AX.
- The tap-corroboration false negatives (#1634) were a two-backend seam bug: evidence captured by one backend compared against the other.
- #1569 correctly established that backends are never comparable *views* — which also means naive tree-diff checks can't catch drift, so none exist.

The failure mode is exactly the classic two-implementation trap: a bug fixed (or a capability added) in one producer silently missing from the other, discovered only when the capture plan flips backends mid-session on a hostile screen.

## Why not unify (context for the decision record)

- Private-AX is simulator-only by construction; physical devices keep XCTest regardless.
- XCTest carries semantics private-AX only approximates (hit-tested `hittable`, value/focus richness); making AX unconditional on simulators would silently degrade healthy-app captures.
- The private AX server API is unstable ground — the kAXErrorIllegalArgument depth limit has moved between Xcode versions. XCTest is the supported fallback when Apple shifts it.
- The #1105/#1156 penalty system already converges at runtime to "AX where XCTest loses, XCTest where it wins", which beats any static split.

## Proposal: parity as a gate, not equality

**1. Backend-conformance smoke (iOS CI):** #1634's `preferredBackend` option can force a capture through a chosen backend — the forcing seam this test needs. Capture the fixture app's seeded screen once per backend and assert a *capability checklist* over each capture independently (never a cross-backend tree diff, per #1569):

- every seeded control is present with its label/identifier/role
- interactive nodes are discovered and marked interactive
- text values on seeded fields survive
- node count sane (no silent truncation of the fixture screen)

A capability added to one backend that regresses the other now fails CI at the screen where agents would have discovered it in production.

**2. Declared capability matrix (ADR-0011 style):** a small typed table stating per-backend capabilities and KNOWN approximations (e.g. `hittable: hit-tested | approximated`, `deepExtension: yes | no`, `depthLadder: n/a | yes`), consumed by docs/help so gaps are contract rather than surprise, with the same waiver discipline as the interaction guarantee matrix (waivers need owners).

**Prereq:** `preferredBackend` currently accepts only `'private-ax'` (all the corroboration fix needed); the conformance smoke wants the XCTest side forceable too (e.g. `'tree'`) — a small extension of the same seam.

Raised by @thymikee while reviewing the deep-extension/corroboration arc; design sketch from that discussion.

thymikee:
Closes #1635.

## Summary

Adds backend-independent conformance evidence for iOS snapshots without asserting that different accessibility backends must produce identical trees.

- Declares forceability, raw projection support, depth behavior, geometric actionability, conformance applicability, and owned expiring gaps in a typed backend capability registry.
- Forces the recursive XCTest tree only through a test-owned daemon transport seam; no backend preference is added to the public SDK or CLI.
- Keeps TypeScript and Swift declarations aligned through shared fixtures and parity tests, and renders the declared contract in versioned `snapshot` help.
- Exercises tree and private-AX independently against seeded controls, including negative cases for every promised control/property invariant.
- Classifies `hittable` consistently with #1933: tree, query sweep, and private-AX all publish the shared geometric-actionability predicate. Occlusion and native hit-testing remain separate facts.

This PR intentionally does not compare backend trees. It also leaves the broader acquire/present split and presentation architecture to #1797.

The unrelated package-size attribution commits previously stacked here were removed after #1934 took ownership, so this branch is scoped to snapshot backend conformance only.

## Validation

- Red/green registry proof: the focused registry test failed before the typed declaration existed and failed again when the old backend-specific hittability meanings were restored.
- Red/green conformance proof: planted false control/property results were accepted before the checks and rejected after them; the suite now covers sparse quality, truncation, missing controls, identifier, label, role, enabled state, value, interactivity, geometric actionability, geometry, and structured hittable evidence.
- Red/green XCTest proof: disabling the force branch made the focused runner tests fail; restoring it passed the backend-plan and cross-runtime parity cases.
- Live iOS Simulator evidence forced both tree and private-AX on the seeded fixture screen. Both executed the requested backend and returned the expected controls and values; tree reported healthy/non-truncated quality, while private-AX retained its existing structured recovery/truncation disclosure.
- After removing the unrelated size-report commits, `pnpm check:affected --run` passed at `1d7207fe9` (5,458 tests across 677 files plus build, package, typecheck, layering, Fallow, fixture, command-doc, affected-selector, XCTest-selection, and structural gates). GitHub-authoritative native, coverage, and device lanes are running on that exact head.

No simulator was created or deleted. No benchmark app or benchmark data was touched or published.


thymikee:
Blocking review at exact head `7c8d6171a9662c489f6b4ebbec9a1862eb281289`:

1. `preferredBackend` is now exported through `CaptureSnapshotOptions` and generic `CommandExecutionOptions`, hence the published SDK, although the new `tree` value is used only by conformance tests. “No CLI flag” does not make it internal. Keep the force probe on a test-owned/internal daemon route and remove this public test seam.

2. The new TS capability table is not the backend owner: Swift separately declares backend cases, raw support, availability, and the forceability branch. There is no shared source or parity gate, so the claimed classification gate can drift. Make one cross-runtime declaration/enforcement path and add the #1635-required help/docs consumer.

3. Waiver expiry and conformance assertions are not enforced: the only registry validation fixes today at `2026-08-21`, and the checker test only rejects a wrong backend. Use real-date CI validation plus boundary coverage, and add negative cases for each promised control/property invariant.

thymikee:
Re-review at `e98d8e8`: the prior code blockers are resolved. Backend forcing is test-owned rather than a public SDK seam; the shared fixture now gates TS and Swift declarations and is surfaced in snapshot help; and real-date waiver expiry plus control-specific negative tests make conformance checks non-vacuous. CI and the iOS simulator conformance E2E are green.

Owner action before readiness: the size report shows `+4.9 kB` npm unpacked (>3 kB). Please amend the PR description to itemize the justified production/test growth and state why a materially smaller design—for example, generated declarations instead of the typed registry plus shared parity fixture—was rejected.

thymikee:
**[P1] Restack after #1933 and update the hittability contract.** This draft currently publishes `tree`/`queries` as `hit-tested` and `private-ax` as `approximated` in the capability registry and snapshot help. #1933 is merge-ready and deliberately replaces those backend-specific meanings with one backend-independent geometric-actionability predicate across recursive tree, query sweep, private AX, and collapsed-tab acquisition. The two branches cannot merge as-is without making the registry/help false. Land #1933 first, then rebase this PR and classify the resulting shared geometric semantics in the fixture, TS/Swift parity, help, and conformance expectations. The other prior review blockers are resolved and current-head CI is green.

thymikee:
Addressed in 13ff1d8dd, rebased onto merged #1933 at 07023eb2.

- The typed registry, shared JSON fixture, Swift parity test, and generated snapshot help now classify tree, queries, and private-AX `hittable` uniformly as `geometric-actionability`.
- The obsolete private-AX `hittable-approximation` gap is removed.
- Backend conformance now requires seeded interactive controls to publish `hittable: true` under the shared geometric predicate, while keeping occlusion and native hit-testing separate.
- Focused TS registry/help/conformance checks and `git diff --check` pass. Native/toolchain/device validation for the rebased head is pending.

thymikee:
P1 — split the size-report expansion out of this #1635 draft. The unrelated npm package-attribution commit grows scripts/size-report.mjs from 512 to 682 lines, crossing the repository’s extract-before-adding-behavior threshold, and duplicates work now owned by #1934. Remove that commit from this branch and keep #1930 scoped to iOS snapshot backend conformance. The #1933 restack itself is correct: registry, fixture, Swift parity, help, and conformance now consistently use geometric actionability. Current-head iOS validation is still running; Linux failed only during apt dependency installation and is an infrastructure rerun.

thymikee:
Addressed in the restacked head `1d7207fe9`. Commits `492a1130b` and `76a7eba62` are removed; the #1933 actionability restack is preserved with an identical patch ID, and `origin/main...HEAD` contains no size-report files. The PR body is now scoped to #1635 and records the exact validated head. `pnpm check:affected --run` passed before the force-with-lease push; exact-head CI is running.

thymikee:
Re-reviewed exact head 1d7207fe: the unrelated size-report attribution work and 682-line module growth are fully removed; the PR is again scoped to iOS snapshot-backend conformance, with the #1933 geometric-actionability alignment retained. Code review is clean. Android and iOS Smoke are still running and the PR remains draft, so no readiness label yet.

thymikee:
Exact head 1d7207fe is now fully green (28/28), non-draft, and mergeable/CLEAN. The prior scope blocker remains removed and the snapshot-conformance review is clean. Merge-ready.

PR #1934, opened 2026-08-21
ci: attribute native package size and trim Apple runner

thymikee:
## Summary

- Attribute every byte from `npm pack --json` to JS/dist, the Apple runner, the macOS helper, Android helpers, or other package files. The report now fails if those components do not sum exactly to npm's unpacked size and shows the largest changed packed files.
- Keep the size workflow's base-commit comparison working with the extracted report modules, while keeping each reporting module below the repository's context-size threshold.
- Remove five declarations/parameters that clean Periphery builds identified as unused on iOS, macOS, tvOS, and visionOS. Test-only conveniences remain available to runner unit tests but are stripped from the shipped source package.

This is the tightening follow-up to the package-growth review around #1930. It does not change snapshot behavior or backend contracts. Scope is Apple runner source plus its package-size CI evidence.

## Validation

- Clean Periphery scans after packaging: iOS 8 platform-specific findings, macOS 89, tvOS 90, visionOS 78; the cross-platform intersection is now empty. Before the cleanup, the intersection contained the five declarations removed here.
- The real release-shaped local pack contained 388 files and 3,187,049 unpacked bytes: 2,503,862 JS/dist, 543,510 Apple runner, 54,545 macOS helper, 43,196 Android helpers, and 41,936 other. The component sum matched npm exactly.
- Focused reporting/package tests: 12 passed. XCTest selection reaches every declared test lane.
- iOS Simulator and macOS XCTest runner builds both completed with `TEST BUILD SUCCEEDED`.
- The size workflow's base-checkout reporter path has a red-first regression test and was also exercised through its exact copy-and-run shell entry point.
- `pnpm check:affected --run` passed on commit 5b014496f; GitHub-only lanes remain authoritative on this pushed head.

No docs or skills changed: this affects CI reporting and removes unreachable implementation surface without changing user-facing commands or guidance.

### DIFF
--- a/.github/workflows/size.yml
+++ b/.github/workflows/size.yml
@@ -41,7 +41,9 @@ jobs:
         uses: ./.github/actions/setup-node-pnpm
 
       - name: Preserve report scripts
-        run: cp scripts/size-report*.mjs /tmp/
+        run: |
+          mkdir -p /tmp/agent-device-size-report
+          cp scripts/size-report*.mjs /tmp/agent-device-size-report/
 
       # dist is fully determined by the base commit, so reuse it across PR runs
       # against the same base. Startup medians are still measured fresh on this
@@ -60,7 +62,7 @@ jobs:
           if [ "${{ steps.base-dist-cache.outputs.cache-hit }}" != "true" ]; then
             pnpm build
           fi
-          node /tmp/agent-device-size-report.mjs \
+          node /tmp/agent-device-size-report/size-report.mjs \
             --startup-runs 7 \
             --json /tmp/agent-device-size-base.json
 
--- a/.github/workflows/size.yml
+++ b/.github/workflows/size.yml
@@ -40,8 +40,8 @@ jobs:
       - name: Setup toolchain
         uses: ./.github/actions/setup-node-pnpm
 
-      - name: Preserve report script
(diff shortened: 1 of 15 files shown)


thymikee:
P1 — the new preservation step copies the entrypoint to /tmp/size-report.mjs, but Measure base size still invokes /tmp/agent-device-size-report.mjs. The exact-head Size job fails MODULE_NOT_FOUND before it can generate the base report, so this PR cannot produce the comparison it changes. Please preserve the scripts in one dedicated temp directory with a consistent entrypoint (or otherwise make copied/invoked names match), and cover the base-report handoff.

thymikee:
Addressed in `5b014496f`. The workflow now preserves the reporter entry point and its relative modules together under `/tmp/agent-device-size-report/`, then invokes that exact path after the base checkout. A red-first workflow regression test covers the copy/invocation handoff and module colocation. `pnpm check:affected --run` passed, and the replacement exact-head Bundle Size job passed end-to-end in 41s.

thymikee:
Re-reviewed exact head 5b01449: clean. The reporter now preserves its entrypoint and relative modules together, the structural regression rejects the old mismatched handoff, and the successful Bundle Size job proves the base checkout executed that preserved entrypoint before the PR comparison. All completed checks are green; iOS Smoke is still running, so code review is complete and merge readiness awaits that lane.

PR #540, opened 2026-05-14
fix: improve Android text entry stability

thymikee:
## Summary

- make Android IME-owned input a fast terminal failure and add provider-native text injection support
- replace Android text fallback with chunk-safe ASCII shell input from the first attempt; no raw adb, clipboard, paste, speculative helper fallback, or provider result object is documented as an agent path
- require settled Android fill verification so React Native controlled inputs cannot report success for near-complete prefixes like `filed the expens`
- keep keyboard next-action guidance derived at the CLI/error boundary instead of storing it in Android keyboard state
- treat active IME package from dumpsys as the durable input-ownership signal; use known IME packages only as a diagnosed fallback when dumpsys omits the active method
- add richer keyboard status, helper snapshot diagnostics, and clearer empty interactive snapshot guidance
- default app discovery to user-installed apps, ignore Android package-query metadata rows, and fix Android apps/runtime import paths in the packaged daemon
- stabilize Metro companion tunnel test cleanup, debug log tailing, and refreshed Fallow baselines for the changed files

## Closes

- Closes #530
- Closes #531
- Closes #532
- Closes #533

## Related

- Partially addresses #539 by adding a provider-native Android text injection hook and backend selection tests. The remaining provider-neutral remote-session capability and adapter tests should stay tracked in #539.

## Validation

- /bin/zsh -lc 'PATH=/Users/thymikee/.nvm/versions/node/v24.13.0/bin:$PATH ./node_modules/.bin/vitest run'
- /bin/zsh -lc 'PATH=/Users/thymikee/.nvm/versions/node/v24.13.0/bin:$PATH ./node_modules/.bin/vitest run src/platforms/android/__tests__/index.test.ts src/platforms/android/__tests__/input-actions-fill.test.ts src/__tests__/cli-client-commands.test.ts'
- /bin/zsh -lc 'PATH=/Users/thymikee/.nvm/versions/node/v24.13.0/bin:$PATH ./node_modules/.bin/tsc -p tsconfig.json'
- /bin/zsh -lc 'PATH=/Users/thymikee/.nvm/versions/node/v24.13.0/bin:$PATH ./node_modules/.bin/oxlint . --deny-warnings'
- /bin/zsh -lc 'PATH=/Users/thymikee/.nvm/versions/node/v24.13.0/bin:$PATH ./node_modules/.bin/rslib build'
[...]

PR #1935, opened 2026-08-21
refactor: migrate type to the request-bound device runtime

thymikee:
Wave 5 unit 2 for #1739 (ADR 0019) — `find`'s last blocker. Unit record posted to the tracker
before work started; this PR delivers it.

## What changed

**One execution path.** `type "text"` (interaction route) and `find <q> type "text"` both resolve
`resolveBoundTypeTextRuntime`, admit the exact owner's `typeText` fact, and bind once. The
interaction backend's `typeText` member now exists only when the `type` handler admitted and bound
a runtime — no caller can fall back to legacy dispatch. Previously the backend and find's leg each
dispatched `'type'` through `core/dispatch.ts` independently.

**Byte-for-byte leaf parity.** `executeBoundTypeText` reproduces the retired `handleTypeCommand`:
leading-ref rejection with the identical hint text, space-joined positionals, the 0–10 000
`delay-ms` bound, and only `textEntryRoute` surviving from the owner's result. The operation
returns the interactor's own closed `TypeTextBackendResult`, so Apple route evidence passes through
and every other owner types blind — the contract `interactor-types.ts` already owned. The iOS
synthesized-type commit wait (#1676) is Apple-interactor-internal and moves nowhere.

**Facts replace the bucket** (the retired cell table, restated by its owners): apple sim+device,
android all-but-the-synthetic-simulator-row, harmonyos emulator+device, linux device, web device,
vega unavailable, providers wherever their interactor is reachable. `type` leaves
`HARMONYOS_SUPPORTED_COMMANDS` and `WEB_INTERACTION_COMMANDS`. R41 is the new cutover row;
`typeText` has exactly one owner.

**Linux promoted contract → live**: the CI desktop replay now types a digit on real hardware,
with the two-sided manifest count pins updated (contract 19→18, live 6→7).

## Test evidence, with the mutants run up front

Binder contract test (`packages/contracts/src/type-text-runtime.test.ts`) covers resolver context,
positional `(text, delayMs)` forwarding, blind-result passthrough, the structured missing-provider
failure, and cancelled-request short-circuit. Two planted mutants, both killed while every other
suite stayed green: deleting `await interactor.type(...)` (2 failed) and dropping the delay
argument (1 failed). Source restored byte-identical.

The retired leaf's parse pins (`dispatch-type.test.ts`, two `handleTypeCommand` tests) moved into
the daemon runtime test rather than being dropped — including the exact repair-hint text.

## Live changed-path evidence

| Cell | Target | Result |
| --- | --- | --- |
| apple `simulator` | iPhone 17 Pro, Settings search | generic: **committed** (screenshot shows both typed strings in the field), route `synthesized-first-responder`; find leg: `Typed 9 chars`, same route |
| apple `device` | macOS host, TextEdit | **committed** (screenshot shows the typed text in the document), route `xctest-element` |
| android (local emulator) | Pixel_9_Pro_XL | see the caveat below |
| limrun provider | Chrome omnibox | generic: **committed** ("example.com" suggestion row in the tree); find leg executed (`--first` matched a suggestion row, which navigates) |
| webdriver provider (AWS iPhone 16/18.0) | Safari | **committed and navigated** ("Example Domain" in the tree) |
| harmonyos `device` | nova 14 | `Typed 12 chars` |
| web `device` | agent-browser | `Typed 9 chars` |
| vega (real VVD) | — | exact declared refusal + `tv-remote down` positive control on the same session |
| linux `device` | CI desktop replay | live step in this PR, Smoke lane authoritative |

**Local-Android caveat, stated rather than papered over:** on the local Pixel_9_Pro_XL AVD the
typed text did not commit — and neither did **raw `adb shell input text`**, the ground truth for
what the untouched interactor wraps: GBoard on this AVD holds the input connection
(`focusedPackage: com.google.android.inputmethod.latin`), so the #540 `KEYBOARD_OVERLAY_BLOCKING`
guard fires — **through the bound path exactly as it would through legacy dispatch**, which is
itself parity evidence for the guard riding below the seam. The Android commit proof therefore
comes from the Limrun provider row above (a real Android instance, clean IME state). Emulator IME
state restored afterwards.

All provider sessions closed; AWS checked before and after — zero `RUNNING` sessions, no orphan.

## Gate

`pnpm check:affected --run` all runnable checks passed; layering recognizes 25 migrated commands
including `type`. Complexity findings from the added fact cells were fixed by extracting
`androidTouchFact` / `interactorCell` rather than suppressed.

One process note: mid-PR, `pnpm format` under out-of-sync node_modules reformatted unrelated files
with a stale oxfmt; the commit was rebuilt against `origin/main` with the lockfile toolchain and
carries only the unit's 41 files.

`find` stays legacy with no cutover row: both of its direct execution legs (focus via #1925, type
here) now share bound runtimes, so the atomic R35 cutover is the next unit.


thymikee:
## CI green; size itemized against the #1842 bar

All 28 lanes pass at `82fb8c2dc`. The authoritative Size run reports **+3.1 kB raw / +755 B gzip /
+775 B tarball / +3.1 kB unpacked** — right at #1842's ~3 kB unpacked escalation bar, so itemizing
proactively rather than leaving it to the reviewer:

- `dispatch.js` **−591 B** — the retired `handleTypeCommand` leaf and its dispatch-table arm; the
  root actually shrinks.
- `sdk-batch-runner.js` +718 B — the descriptor projection (`typeTextRuntimeUse`) plus the shared
  binder, the same per-unit contribution R40 showed.
- The remainder (~+2.9 kB spread across the runtime chunks below the top-5 cutoff) is the eight
  owners' fact cells and bind arms — one per platform package and provider, which is ADR 0019's
  exact-owner shape and not reducible without collapsing ownership.

For calibration: `focus` (#1925) landed at +4.5 kB unpacked, `wait` (#1875) pre-tightening at
+3.8 kB. This unit is the smallest of the three, net-negative in the daemon/dispatch root, and
carries no duplicated construction to tighten — the parse helper is shared, the input projection is
the one `runtimeExecutionFromContext` every request-bound operation forwards.

thymikee:
Not ready at 82fb8c2d despite green CI. (1) Linux's claimed live type proof is vacuous: after `type "5"` the replay repeats the same calculator-exists selector already true before typing (and `label=5` may be the button), so deleting the type step leaves the evidence green. Assert a changed display/value. (2) The supported local-Android operation was not exercised: the keyboard-overlay guard returned before `runtime.interactions.typeText`; Limrun is a separate provider owner. (3) Add exact fact/binding parity coverage for every claimed local/provider cell; the current tests pin only a synthetic Apple-available harness. (4) The +3.1 kB comment is accurate, but its no-duplication claim is not: `type-text-runtime` repeats focus/screenshot interactor binding, the generic route validates/formats through multiple adapters, and the change adds behavior to Limrun `app-log-runtime.ts` at 586 LOC despite the 500-line extraction rule. Tighten around a shared exact-owner interactor seam/direct normalized executor and extract the Limrun interaction owner before readiness.

thymikee:
All five addressed at `3d44b6dea`. The size comment's "no duplicated construction to tighten" claim
was wrong — the P2 findings below prove it, and the tightening is now in.

## [P1] Linux live evidence — made change-sensitive

You were right that the repeated existence assertion could not fail. The replay now runs
`type "555"` followed by `wait "label=555 || text=555 || value=555" 10000`: no calculator button is
labelled 555, so the node exists only if the keystrokes landed in the display, and deleting the
type step turns the wait red by construction. (`5` alone would have matched the digit button — the
exact hole you named.) The Smoke lane on this head is the authority for the step.

## [P1] Local Android — proven on both paths, and the earlier read was wrong

One correction to your finding first: the overlay guard does not fire before the bound operation —
it lives inside `interactor.type`, which the bound `typeText` invokes, so even the blocked run
exercised the migrated path end-to-end. But your conclusion stood: no local commit had been shown.

Now it has, on a clean AVD (Pixel_7_review), through **both** local channels:

```
@e2 [text-field] "batteryctrlbattery" [focused]     <- bound shell-injection type, raw-adb control, bound test-IME type — all three committed
@e2 [text-field] "wifi"  + real Wi-Fi search results <- find "Search settings" type "wifi" (find leg)
```

The middle token is a raw-adb control run, so the bound path is bracketed by ground truth on the
same field. Root cause of the earlier failures, for the record: this worktree had never packaged
the IME helper artifact (`android/ime-helper/dist` missing), so test-IME activation warned and fell
back to shell injection, and the first AVD's GBoard held the input connection — environment, not
path. After `pnpm package:android-ime-helper:npm`, activation succeeds and the durable IME is
selected. Limrun remains the provider cell's evidence, not a substitute for this one.

## [P1] Exact fact-cell coverage — all eight owners pinned

Every owner runtime test now asserts the `typeText` (and `focusPoint`) fact and the bound
operation's presence/absence for its exact kinds:

| Owner | Pinned cells |
| --- | --- |
| apple | available across the leaf table (sim + physical kinds) |
| android | emulator/device/unknown available; **synthetic simulator refused, operation absent** |
| harmonyos | emulator/device available |
| linux | device available; non-device refused, operation absent |
| web | browser device available; non-device refused |
| vega | refused with the `remote navigation only` hint, operation absent |
| webdriver | reachable → available+bound; **inactive session refused**; the unsupported-capability row now documents that interaction gates on interactor reachability, not on capture declarations |
| limrun | live available+bound; recovery facts refused |

## [P2] Duplicated construction — extracted, in both places you pointed at

- **Contracts:** new `interactor-operation-binding.ts` owns the local resolver and the fail-closed
  provider resolver once; the screenshot, focus, and type binders all consume it (three copies →
  one; provider error text byte-identical, contracts suite green).
- **Daemon routing:** the interaction-runtime hop is gone. `runTypeTextViaRuntime` calls the bound
  executor directly — the hop's only contributions were a second validation pass, a second format
  pass, and a rebuilt response; recording never lived there (`typeTextCommand` is a pure
  validate→dispatch→format function; recording is `finalizeTouchInteraction`'s). The
  `boundTypeText` backend member and its conditional wiring are deleted; ADR 0014 frame expiry
  moved to the handler where the mutation happens.

## [P2] Limrun module — interaction moved out

`interaction-operations.ts` now owns the interactor-backed cells; `app-log-runtime.ts` is 563
lines — 13 below its pre-unit 576, so this unit now shrinks the module instead of growing it. The
remaining descent below 500 is that module's own extraction debt (it needs the log/recovery split,
which is not type behavior) and I did not smuggle it into this unit.

Gate at the pushed head: `check:affected --run` all runnable checks passed; owner suites 80/80;
contracts 215/215. Exact-head lanes running now.

thymikee:
## The change-sensitive Linux assertion did its job on its first CI run — against my own step

Head is now `0bb7f6a40`. The two Smoke failures at `3d44b6dea` decomposed as:

- **Android — flake.** Failed at an orientation-scenario wait untouched by this PR; passed clean on
  rerun.
- **Linux — real, and the new assertion caught it.** The lane reported `focus 100 100: xdotool
  timed out`, but the artifact shows `attempt-3`: attempt-1 actually ran the full tail and **failed
  at the new `wait 555`** — the typed digits never landed, because `focus 100 100` clicks the
  desktop and takes keyboard focus away from the calculator. The retries then wedged on a latent
  quirk: attempt-1 parks the pointer at (100,100), and the next attempt's
  `xdotool mousemove --sync` to the same point waits for a motion event that never comes, so every
  retry died at step 7 with the 10s timeout, masking the real failure.

Which is to say: the vacuity finding was more right than either of us stated. The previously-green
`type "5"` was not merely weakly asserted — the keystrokes were going to an unfocused desktop and
landing nowhere, and only the broad assertion kept the lane green.

**Fix (pushed):** the tail is now `focus 100 100` (R40 evidence + survival assert) →
`click "label=1"` — a resolved press inside the window that restores keyboard focus, proves
pointer input lands in the app, and moves the pointer off (100,100) so retries cannot trip the
mousemove no-op hang — → `type "55"` → `wait "label=155 || text=155 || value=155" 10000`. No
button is labelled 155, so the clicked 1 plus the typed 55 can only satisfy the wait if the
keystrokes reached the display; deleting the type step turns it red.

The `mousemove --sync` no-op hang is a pre-existing product fragility outside this unit's scope —
it poisons the retry of any coordinate interaction at a repeated point — and is filed separately
rather than smuggled in here.

Exact-head lanes running on `0bb7f6a40`; the Linux Smoke lane is the arbiter.

thymikee:
Re-review: the five earlier findings are addressed. The exact head is not ready yet: iOS’s failure is an upstream fixture fill flake—the migrated type did append `.test`, but the prior untouched seed fill did not retain `@example`—so rerun that lane. Linux is not an infrastructure flake: a failed first replay attempt causes the retry to repeat `focus 100 100` and hit xdotool’s no-motion timeout. Until the exact-head Linux Smoke succeeds, the new change-sensitive type proof is unverified; make the replay retry-safe or provide a passing exact-head rerun.

thymikee:
## Response to the thermo-nuclear review — Finding 1 was already dead at review time; its residuals are now dead too

You reviewed `82fb8c2dc`. The three-hop relay was deleted two heads earlier than your read landed
(`3d44b6dea`, from the previous review round): `runTypeTextViaRuntime` calls the bound executor
directly, the `boundTypeText?` param and conditional backend member are gone, and `expireRefFrame`
moved to the handler. Current head is `db3304d82`. What your review adds beyond that — and it does
add — is now in:

**The "decide what `typeTextCommand` is for" question had a sharper answer than extraction:
deletion.** After the direct routing, I traced its consumers: the daemon was the only production
caller, and the released SDK's `interactions.type` goes over the daemon wire
(`executeCommand('type')`), never through the embedded runtime catalog. So `typeTextCommand`, its
`Options`/`Result` types, the catalog registrations in `bindInteractionCommands`/
`interactionCommands`, the `AgentDeviceBackend.typeText` member, and every fixture/pin holding the
dead surface alive are deleted. **One parse/compose owner remains** — `executeBoundTypeText` — so
the drift-pair you flagged (empty→ref vs ref→empty) no longer exists to drift. R41's
`legacyRetirement.routeNames` now names `typeTextCommand`, so reviving it fails the gate.
Net: the relay hop, the second parse owner, and ~90 lines are gone; "exactly one owner" is now
true at the operation, parse, and compose levels.

**Finding 2 — both dead `surface` fields removed**, `TypeTextInput` and the `FocusPointInput` it
was cloned from. You were right that cloning a dead field forward is how contracts accrete;
neither had a writer or a reader.

**Finding 3 — partially pre-empted, and the rest is now on the table.** `3d44b6dea` already
extracted the local/provider resolution + fail-closed core into
`contracts/interactor-operation-binding.ts` (screenshot, focus, and type binders consume it). What
remains per-module is the input→RunnerContext projection and the one operation call — closer to
your `bindInteractorOperation(invoke)` shape than the clones you read, but not all the way there.
Saying it out loud as you asked: the next interaction unit should either adopt a shared invoke
core or record why per-operation files stay. It goes in that unit's record on #1739.

**On the two Smoke lanes your review predates:** the Linux lane exposed that the change-sensitive
assertion caught a real ambiguity (`[text] "1"` beside `[button] "1"` → by-design
`AMBIGUOUS_MATCH` on attempt-1, masked by an xdotool no-op-move hang that reports every retry as
the step-7 focus timeout — hang filed separately). Now `click "role=button label=1"`. The iOS
fixture failure ("typed email suffix should be observable") does not reproduce: at this head, live
on a real simulator, coordinate focus → bare `type` → route `synthesized-first-responder` → text
observable in the captured tree. Two CI failures in a row still means the new head gets one more
arbitration run before I bisect the routing change on a probe branch.

Gate at `db3304d82`: `check:affected --run` all runnable checks passed.

thymikee:
Reviewed c50f461. The prior ownership, binding, and duplicate-routing follow-ups are addressed, and iOS now passes on this head. I cannot call this merge-ready yet: Linux Smoke still fails all three attempts at step 7 (`focus 100 100`, `xdotool timed out after 10000ms`). The uploaded artifact shows an empty calculator and no typed-state screenshot, so `type "100+55="` and the `155` assertion never run. Please make the replay prerequisite retry-safe or decouple the type proof from it, then rerun Linux Smoke; until then the coverage manifest must not classify `type` as live evidence.

thymikee:
## The Linux lane, closed out — with the platform fix it forced and the honest tier for what remains

Head `b2d193d14`. Six CI rounds of artifact evidence ended in one product fix, one honest
demotion, and three filed platform defects. The short version:

**The retry-masking hang is fixed in product code.** `moveTo` now probes
`xdotool getmouselocation --shell` and skips the `--sync` move when the pointer already sits on
the target — a no-op move emits no motion event and hangs for the full action timeout, which is
why every failed replay retry reported its first coordinate step instead of the real failure.
Provider tests pin both sequences including the skip case. This run was the first where all three
attempts reported the true failing step.

**The type assertion cannot hold on this lane today, and the claim now says so.** The final
artifact is unambiguous: the calculator entry visibly holds the typed "155" while
`wait "value=155 || label=155 || text=155"` times out — with no interactive filtering in play and
a matcher that does compare `node.value`. GTK4 gnome-calculator exposes no Text-interface content
through the AT-SPI dumper. So the replay keeps the type step and uploads the typed-entry
screenshot every run (live pixel evidence the migrated `typeText` path lands keystrokes on real
Linux hardware), closes with the survival assertion, and the manifest claim returns to the
**contract tier** with the reason written at the entry. Promoting it back to `live` is gated on
the exposure defect, not on this unit.

**Three Linux defects filed with pixel evidence** (one task): the resolved-click target miss, the
dropped `=` keystroke, and the GTK4 Text-interface exposure gap. Fixing PyGObject-vs-GTK4
behavior blind through CI rounds is not a sane loop; they need a Linux desktop.

For the reviewer's P1 ledger: the change-sensitive assertion did exactly its job — it went red
against real defects four times (desktop-focus steal, selector ambiguity, dropped keystroke,
exposure gap) where the old broad assertion had been green over all of them.

thymikee:
Re-reviewed at b2d193d: the prior architecture findings and Linux retry hang are fixed. The exact-head Linux artifact visibly shows `155`, so the migrated path has real-target evidence; keeping the manifest at contract tier is honest because GTK4 does not expose the entry text to an automated selector assertion. No code finding remains. Before merge, please refresh the PR body: it still claims Linux contract 19→18 / live 6→7, describes the subsequently deleted backend `typeText` member, and says 41 files although the current diff has 64 (+1019/-303). Record the Linux artifact as manual changed-path evidence with the manifest remaining contract-tier. Exact-head iOS Smoke is still in progress.

PR #1936, opened 2026-08-21
refactor(apple): drop dead runner code and collapse duplicated helpers

thymikee:
## Summary

Follow-up to the #1934 package-size audit. This removes verified-dead Apple runner code, replaces duplicated validation and mapping paths with typed/shared concepts, and gives test-only Swift sources an explicit packaging boundary. Runtime behavior and the wire protocol are unchanged.

- **Dead code:** removes declarations with no live callers, including the superseded scope lookup, unused interactive-type set, redundant model initializers, repeated TV-remote parsing, duplicated hit-candidate comparison, and repeated unsupported-operation construction.
- **Typed scroll input:** one `RunnerScrollDirection` enum now carries validated direction through command execution, gesture planning, and macOS wheel dispatch. Duration validation is shared and positively named.
- **Shared normalization and errors:** trim-or-nil command IDs use one neutral `String.trimmedNonEmpty` primitive; unsupported runner operations use a dedicated `LocalizedError` instead of borrowing an unrelated transport error.
- **Single element-name projection:** element names are read from one raw-value table, including private keyboard/key values, instead of maintaining a table plus a parallel switch.
- **Explicit test-source boundary:** ten test-only Swift files live under `AgentDeviceRunnerUITests/UnitTests/`; the npm source packager skips that directory instead of inferring skeleton files from source text. The XCTest declaration scanner now traverses nested source directories, so moving tests cannot make CI coverage evidence disappear.
- **Model cleanup:** `DataPayload`, `Response`, and `ErrorPayload` rely on synthesized memberwise initializers; all same-file optional-default and pass-through initializer noise was removed together.

The package remains smaller while the source structure is more explicit: test-only files do not ship or compile on user machines, and the package guard rejects any `UnitTests/` leakage.

## Validation

- Focused package/XCTest selection suite: 31/31 passing.
- XCTest selection gate: 187 declared methods, 0 dark; moved tests remain selected by host/PR/nightly lanes.
- macOS XCTest `build-for-testing`: passed on the rebased head.
- `pnpm check:affected --run`: all runnable checks passed before push; GitHub-authoritative Apple/device lanes are pending on the exact head.
- Regression evidence: removing the `UnitTests` directory exclusion makes the package fixture fail because a test source ships; the recursive scanner change was first observed red when moved tests disappeared from the declaration count.

No CLI, help, docs, or skill changes are needed because command behavior and user-facing guidance are unchanged.


thymikee:
P2: Please add a package-fixture regression for skeleton removal. The existing strip fixture keeps runtimeHelper, so it never exercises `isSkeletonSwiftContents`; reverting that new branch still leaves the five guard tests and the macOS package build green because the old empty/import-only Swift files compile. Add a test-only RunnerTests extension, strip it, and assert its packaged `.swift` path is absent while a non-skeleton sibling remains.

thymikee:
Addressed the P2 in 690934c34: `apple-runner-package-source.test.ts` now has a dedicated guard that packages a fixture whose entire body is unit-test blocks (`RunnerTests+SkeletonOnly.swift`) alongside a non-skeleton sibling, and asserts the packaged skeleton path is absent while the sibling still ships.

Observed red before trusting it: with the `isSkeletonSwiftContents` skip branch disabled in `copyFile`, the new test fails on "a file whose unit-test blocks were its whole body must not ship as an empty translation unit"; all five existing guards stayed green in that state, matching the reported gap. Re-enabled, all 10 packaging/size-report tests pass.

CI on 690934c34 is the authority still to come.

thymikee:
Re-reviewed at 690934c. The skeleton-file exclusion now has a non-vacuous regression proof: a file whose body disappears after test stripping must be absent from the package, while a runtime sibling remains. I also verified the package path distinguishes the two (56 shipped files / 10 skipped versus 66 before the exclusion). All required CI, including Apple host/device lanes, is green. No findings; merge-ready.

thymikee:
Addressed the follow-up review holistically on exact head `f017a0222` and rebased onto current `main`:\n\n- replaced the element-name table + switch with one raw-value table;\n- carried scroll direction as a typed enum through execution/planning/desktop dispatch and made duration validation positive;\n- replaced skeleton-source inference with an explicit `UnitTests/` boundary, plus recursive XCTest discovery so nested tests remain covered;\n- replaced the unrelated transport NSError with a dedicated `LocalizedError`;\n- moved trim-or-nil normalization to a neutral `String.trimmedNonEmpty` source and removed double normalization;\n- completed the model cleanup across `DataPayload`, `Response`, and `ErrorPayload`;\n- corrected the PR description, including removing the false PresentedNode-initializer claim.\n\nEvidence: focused package/selection tests 31/31, selection gate 187 declarations / 0 dark, macOS XCTest build-for-testing passed, and `pnpm check:affected --run` passed before push. Exact-head CI is now running; no merge performed.

PR #1933, opened 2026-08-21
fix(ios): separate snapshot actionability from occlusion

thymikee:
## Summary

- Make iOS snapshot geometric actionability one backend-independent predicate across recursive tree, query sweep, private AX, and collapsed-tab acquisition.
- Remove the XCTest runner later-node occlusion scan and its obsolete flatten/range/type-gate helpers while preserving intentional direct/raw interaction paths.
- Keep daemon snapshot occlusion as the sole owner of structured covered state and retain interactionBlocked: "covered" consumer behavior.
- Add focused regression coverage and changed-path live evidence for the distinction between geometric hittable and daemon occlusion.

## Validation

- Red before the fix: the focused regular-presentation XCTest selected one test and failed three assertions for a covered button, labeled image, and disabled button.
- Green: iOS XCTest build succeeded; focused runner tests 2/2, snapshot/presentation tests 18/18, private-AX tests 5/5; daemon occlusion/state/response/find Vitest files passed with 97 tests; xctest selection passed.
- Live on existing booted iPhone 17 Pro F7D6F9A4-4FCC-4DD7-AC0B-3280C9319CB9: regular and raw direct payloads both reported automation-press as hittable true with no interactionBlocked, snapshotQuality.backend tree, and XCTest backend diagnostics present (4/5). The changed lifecycle route passed its regular/raw app-switcher assertions and screenshot evidence. All scoped sessions were closed and isolated daemons cleaned.
- Final pnpm check:affected --run passed all runnable checks; native/device and coverage lanes remain GitHub-authoritative. The broad lifecycle run later reached the existing simulator observability check and stopped because CPU sampling is unavailable on simulators; that unrelated failure was not pursued.

PR #1931, opened 2026-08-21
feat(ios): publish effective snapshot geometry

thymikee:
## Summary

Stacked on PR #1929, targeting `claude/1797-clip-fold` at `e4b19df6c`.

This PR completes the next smallest #1797 architecture unit: the effective-geometry carrier and
the rect-consumer inventory/migration. It does not add the separate occlusion/actionability
enforcement slice.

- Acquisition keeps `RawAXNode.rect` as the backend-reported frame.
- Regular presentation now carries a `SnapshotPresentationNode` with both the raw node and its
  presentation-owned `effectiveRect`.
- The regular wire projection writes `effectiveRect` through the existing `PresentedNode.rect`
  field. Raw snapshots and direct single-element reads retain reported geometry.
- Scope, reindexing, hidden-content hints, and collapsed-depth projection preserve the carrier.
- The only policy branch is the declared `SnapshotVisibilityFold.Policy`: iOS's
  `cursorProjected` policy supplies ancestor scroll clips; `plainViewport` does not. No backend
  name or acquisition implementation is inspected by the geometry code.

## Rect inventory

| Consumer class | Owners reviewed | Classification / migration |
| --- | --- | --- |
| Acquisition facts and runner-internal geometry | `RunnerTests+Snapshot.swift`, the acquisition half of `RunnerTests+PrivateAXPresentation.swift`, and raw-node builders in `RunnerTests+Interaction.swift`, `RunnerTests+SystemModal.swift`, and `RunnerTests+CommandJournal.swift` | Reported geometry. These continue to use `RawAXNode.rect` for traversal, roots, deduplication, and direct reads. |
| Presentation and payload assembly | `RunnerTests+SnapshotPresentation.swift`, `RunnerTests+SnapshotVisibilityFold.swift`, `RunnerTests+SnapshotCapturePlan.swift` | Effective geometry for regular projection; raw geometry for raw/direct-read paths. The typed carrier is the single migration point. |
| iOS daemon presentation | `src/daemon/snapshot-presentation/ios/*.ts` (`rows`, `noise`, `action-shelf`, `actions`, `scroll`, `transitions`, and `web`) | Existing wire `rect` consumers; no parallel field or adapter. They now receive effective geometry automatically. |
[...]

PR #1937, opened 2026-08-21
fix(ios): make snapshot presentation construction private

thymikee:
## Summary

Related to #1797.

The merged #1930, #1931, and #1933 prerequisites establish backend conformance, effective geometry, and daemon-owned occlusion. This focused slice closes the remaining construction-boundary gap:

- Moves PresentedNode under SnapshotPresentation.
- Preserves the existing PresentedNode spelling at the payload boundary with a typealias.
- Makes all interpretation initializers fileprivate in RunnerTests+SnapshotPresentation.swift, so backend files can carry presentation output but cannot recreate it.

This PR does not touch the #1936 scanner or package-boundary files.

The broader #1797 obligations remain intentionally separate: the cumulative-clip choke and named quality downgrade, deterministic differential fixtures and multidimensional ratchets, nightly live differential wiring, cross-runtime golden conformance, and the visible-depth frontier.

## Validation

- Planted-red compiler proof: a temporary direct PresentedNode(presenting:) call in RunnerTests+PrivateAXPresentation.swift failed the macOS XCTest build at the backend source with the inaccessible-constructor diagnostic; removing that violation restored the build.
- Unit-enabled direct macOS XCTest build passed after the violation was removed.
- pnpm check:xctest-selection passed.
- pnpm check:affected --run passed with exit 0. The pre-existing untracked .codex/environments/ path caused the selector to fail open to the full runnable set; the rerun used process-table access required by the sandbox.
- pnpm build:xcuitest:macos passed.
- pnpm build:xcuitest:ios passed.

### DIFF
--- a/apple/runner/AgentDeviceRunner/AgentDeviceRunnerUITests/RunnerTests+SnapshotPresentation.swift
+++ b/apple/runner/AgentDeviceRunner/AgentDeviceRunnerUITests/RunnerTests+SnapshotPresentation.swift
@@ -1,6 +1,73 @@
 import Foundation
 
 enum SnapshotPresentation {
+  /// The only snapshot node shape accepted by response payload assembly.
+  ///
+  /// The encoded shape intentionally remains byte-for-byte compatible with the former
+  /// `SnapshotNode` wire model. Construction stays file-private to this presentation module:
+  /// acquisition backends can return `RawAXNode` values, but cannot recreate a presented node.
+  struct PresentedNode: Codable {
+    let index: Int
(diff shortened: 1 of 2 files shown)


thymikee:
Clean review at d82bd8d. PresentedNode construction is now owned by SnapshotPresentation, while payload encoding and the explicit single-element read paths remain intact. The compiler-boundary proof is non-vacuous, and all required CI plus iOS/macOS/Android/Linux smoke checks are green. Ready for human review.

PR #1938, opened 2026-08-21
perf(ios): build structural-identifier suppression child index once per pass

thymikee:
## Summary

Every iOS snapshot paid an O(candidates × n × d) cost in structural-identifier suppression: for each identifier-only `Other` wrapper, `collectDescendantsByParentIndex` rebuilt a full node map and walked every node's ancestor chain. On large RN/XCUITest trees with many testID-shaped containers this is millions of visits per snapshot — pure CPU on the highest-frequency output path.

Suppression now collects descendants via one `childrenByParent` index built per pass (BFS from each candidate), preserving exact parent-link semantics — including trees whose nodes carry no `depth` field, where depth-order-based subtree slicing would be wrong. The single-caller helper `collectDescendantsByParentIndex` is removed.

No behavior change; snapshot output is byte-identical.

## Validation

- Property check: old vs new descendant collection agree on 2,000 random well-formed depth-ordered trees across all positions.
- Full unit bundle green at the pushed commit (`pnpm test:unit`, 7,417 tests), including the iOS presentation suite and the Maestro port-observation tests that pin suppression behavior for depth-less fixture trees.
- Residual risk: iOS snapshot presentation is a device-facing path and this run is fixture-backed only; no live simulator capture was taken for this refactor. The change is semantics-preserving by construction and the presentation suite pins the affected rule.

### DIFF
--- a/src/daemon/snapshot-presentation/ios/noise.test.ts
+++ b/src/daemon/snapshot-presentation/ios/noise.test.ts
@@ -1,0 +1,97 @@
+import { describe, expect, test } from 'vitest';
+
+import type { RawSnapshotNode } from '@agent-device/kernel/snapshot';
+
+import { collectIosStructuralIdentifierSuppression } from './noise.ts';
+import type { SnapshotTreeRuleContext } from '../tree.ts';
+
+type ReadCounter = { reads: number };
+
+function countingNode(base: RawSnapshotNode, counter: ReadCounter): RawSnapshotNode {
+  return {
+    ...base,
+    get index() {
+      counter.reads += 1;
+      return base.index;
+    },
+    get parentIndex(): number | undefined {
+      counter.reads += 1;
+      return base.parentIndex;
+    },
+  };
+}
+
+function makeStructuralTree(
+  candidateCount: number,
+  descendantsPerCandidate: number,
+): {
+  nodes: RawSnapshotNode[];
+  counter: ReadCounter;
+} {
+  const counter: ReadCounter = { reads: 0 };
+  const plain: RawSnapshotNode[] = [{ index: 0, type: 'Application', label: 'App' }];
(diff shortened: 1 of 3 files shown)


thymikee:
P1 — the performance regression is not durably pinned. At exact head `7dc08ed4f`, all committed tests still pass if this implementation is reverted to `collectDescendantsByParentIndex`; the 2,000-tree check in the PR body proves semantic equivalence only and is not reproducible repository evidence for the cost reduction. That conflicts with the repo rule that a new regression test must be observed red against the pre-fix path.\n\nPlease add a deterministic complexity regression through the smallest real owning seam—prefer operation counting (for example, instrumented `parentIndex` reads / adjacency construction on a depth-less parent-linked tree with many structural candidates) rather than wall-clock timing. It should fail on the former per-candidate map + ancestor walk and pass when the child index is built once. Keep the existing depth-less Maestro route assertion as semantic evidence; the new test should specifically pin the performance class.\n\nI found no output-correctness issue in the new traversal: representative order is normalized through sets, cycle traversal is bounded, and parent-link semantics are preserved.

thymikee:
Addressed at a2bc15775: added `noise.test.ts` pinning the performance class through the smallest owning seam — `collectIosStructuralIdentifierSuppression` is now exported and tested directly with operation counting (instrumented `index`/`parentIndex` getters) over a depth-less parent-linked tree of 40 structural candidates × 25 descendants, no wall-clock timing.

Red-first observed against the pre-fix implementation: **163,961 reads vs the 12,164 bound** (~13x over, the quadratic signature), while the semantic suppression test still passes against the old code — so the pin fails precisely on cost, not correctness. New implementation: ~2k reads. The depth-less Maestro route assertion remains as semantic evidence.

PR #1939, opened 2026-08-21
perf(ios): memoize simulator app url-scheme probes per bundle mtime

thymikee:
## Summary

`open <scheme>://...` on an iOS simulator spawned one `plutil -convert json` process per installed app (50–150 apps ≈ 1.5–9s) on every deep-link resolution, repeated identically each time.

URL schemes are read once per installed bundle and cached, keyed on the app's `Info.plist` path + mtime so a reinstall invalidates naturally. An unreadable plist path falls through to the uncached probe, preserving prior error behavior exactly. Steady-state deep-link opens now pay zero per-app plutil spawns.

## Validation

- New test (`app-resolution.test.ts`): two consecutive deep-link resolutions issue exactly one plutil probe per app total; extracted to its own file to respect the test-file-size ratchet pin on `apps.test.ts`.
- Full unit bundle green at the pushed commit (`pnpm test:unit`, 7,417 tests); `check:affected` green.
- Residual risk: device-facing path validated with the fake Apple tool harness only; no live simulator run in this PR. The stat-failure fallback keeps any environment where the cache cannot key correctly on the pre-existing code path.

### DIFF
--- a/src/platforms/apple/core/__tests__/app-resolution-deep-link.test.ts
+++ b/src/platforms/apple/core/__tests__/app-resolution-deep-link.test.ts
@@ -1,0 +1,145 @@
+import { promises as fs } from 'node:fs';
+import path from 'node:path';
+import { test } from 'vitest';
+import assert from 'node:assert/strict';
+
+import { resolveIosSimulatorDeepLinkBundleId } from '../app-resolution.ts';
+import {
+  withFakeAppleTool,
+  type FakeAppleToolResponse,
+} from '../../../../__tests__/test-utils/index.ts';
+import { mkdtempForTest } from '../../../../__tests__/test-utils/tmp-dir.ts';
+import { IOS_TEST_SIMULATOR } from './apple-core-stub-helpers.ts';
+
+function unexpectedArgs(args: string[]): FakeAppleToolResponse {
+  return { stderr: `unexpected xcrun args: ${args.join(' ')}`, exitCode: 1 };
+}
+
+type SchemeFixture = {
+  root: string;
+  plistA: string;
+  listing: string;
+};
+
+async function makeSchemeFixture(): Promise<SchemeFixture> {
+  const root = await mkdtempForTest('deep-link-url-schemes-');
+  const bundleDirs = { a: path.join(root, 'AppA.app'), b: path.join(root, 'AppB.app') };
+  await fs.mkdir(bundleDirs.a, { recursive: true });
+  await fs.mkdir(bundleDirs.b, { recursive: true });
+  await fs.writeFile(path.join(bundleDirs.a, 'Info.plist'), '');
+  await fs.writeFile(path.join(bundleDirs.b, 'Info.plist'), '');
+  const listing = JSON.stringify({
+    'com.example.appa': {
+      ApplicationType: 'User',
+      CFBundleDisplayName: 'App A',
+      Path: bundleDirs.a,
+    },
+    'com.example.appb': {
+      ApplicationType: 'User',
+      CFBundleDisplayName: 'App B',
+      Path: bundleDirs.b,
+    },
+  });
+  return { root, plistA: path.join(bundleDirs.a, 'Info.plist'), listing };
+}
+
+const SCHEME_PROBES = [
+  {
+    suffix: 'AppA.app/Info.plist',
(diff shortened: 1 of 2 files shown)


thymikee:
Two blocking findings at exact head `f22277c02`:\n\n1. **P0 — this accidentally deletes the existing app-resolution suite.** The PR body says the new cache test was extracted into its own file, but the diff overwrites `src/platforms/apple/core/__tests__/app-resolution.test.ts`, removing the tests for `findIosSimulatorInstalledApp` and every `detectSoleRunningIosSimulatorApp` behavior (running-app filtering, ambiguity, failed/rejected probes, timeout, and stopped simulators). Restore that file from current main and add the memo regression in a genuinely separate sibling test file.\n\n2. **P1 — transient plist probe failures are cached as authoritative empty schemes.** `readIosSimulatorAppUrlSchemesUncached` returns an empty `Set` for nonzero `plutil` and malformed JSON, and the caller stores it for 30 minutes. Previously a later deep-link lookup retried; now a one-off timeout/failure makes the app unresolvable until expiry. Carry cacheability/success separately (valid plist with no URL types may be cached; transport/parse failure must not), and add red/green recovery coverage. Also note that `createTtlMemo` uses lazy expiry by default: reinstall paths generate new keys, so old path+mtime entries are never read or removed. Use scheduled expiry or a bounded/key-replacing shape so the long-lived daemon does not accumulate stale install paths.\n\nThe positive memo test should additionally pin mtime/path invalidation after these correctness cases are fixed.

thymikee:
Both findings addressed at 7f8c4ec1a:

1. **P0 — restored.** `app-resolution.test.ts` is byte-identical to main again (the `findIosSimulatorInstalledApp`/`detectSoleRunningIosSimulatorApp` coverage is back). The memo regressions moved to a sibling `app-resolution-deep-link.test.ts`, which uses the fake-tool harness and so stays separate from that file's module-level `runSimctl` mock.

2. **P1 — cacheability split.** `readIosSimulatorAppUrlSchemesUncached` now returns `undefined` on nonzero exit or malformed JSON, and the caller only caches defined results — a valid plist with zero URL types is still cached, a transient failure is retried on the next lookup. Covered by a red/green recovery test (failed probe → `undefined` resolution → retry succeeds; the sibling app's successful probe stays cached). Memo now uses `scheduleExpiry: true` so abandoned install paths don't accumulate in a long-lived daemon. Added an mtime-invalidation test (rewrite + `utimes` bump forces a re-probe).

Also rebased onto current `origin/main` — the previous push was based on a stale local main, which is why the diff looked oversized. Full unit bundle and `check:affected` green on the pushed head.

thymikee:
Re-reviewed exact head 23fbd1639. The cache now keys each bundle on Info.plist path and mtime, does not cache transient probe failures, and the focused regressions cover reuse, reinstall invalidation, and failure recovery. All exact-head CI is green; no code finding. Formal readiness still needs one live simulator deep-link run that demonstrates the changed cache path, since the current evidence is harness-only.

thymikee:
Fallow finding addressed at 23fbd1639: the fake-tool helper's `handle` closure exceeded the complexity threshold, so it's split into a dispatch function plus a table-driven `respondToPlutil` (probe list + failure injection) rather than adding a baseline entry. Tests unchanged in behavior; `check:fallow` green locally and on CI.

thymikee:
Live simulator evidence for the changed cache path, gathered at 23fbd1639:

**Setup:** `bench-golden-v4` simulator (34 installed apps, including Bluesky), daemon restarted on this branch's build. Deep-link resolution exercised via `open bluesky://app`, which routes through `resolveIosSimulatorDeepLinkBundleId` (plutil probe per installed app on the uncached path).

**Timing (wall clock per CLI invocation):**
- Cold (first resolution, pays 34 plutil probes): **3.31s**
- Warm (memo hit): **1.27s / 1.25s / 2.24s** across repeated opens — the ~2s per-app probe pass is paid once per bundle; residual variance is simctl openurl noise.

**Correctness:** snapshot after open shows `App: xyz.blueskyweb.app` — Bluesky resolved from the `bluesky` scheme and launched (iOS confirmation alert visible). Session closed afterward.

This covers the live-run readiness item: cache hit path works against a real populated simulator, and reinstall invalidation/failure-recovery remain covered by the harness tests.

PR #1940, opened 2026-08-21
fix(web): read enabled=false from the snapshot [disabled] annotation

thymikee:
## Summary

Web snapshot nodes never reported `enabled: false`. The agent-browser refs payload carries only `role` + `name`, and the `[disabled]` bracket annotation in the snapshot text — emitted for both the `disabled` attribute and `aria-disabled` — went unparsed, so every node's `enabled`/`focused` was silently `undefined`.

The parser now reads `enabled: false` from the bracket annotation. Quoted label text is stripped before scanning, so a literal "[disabled]" inside a label does not flip the flag; explicit refs metadata still wins when present.

Focus state is not emitted anywhere by the pinned agent-browser backend (verified against the live binary — no marker appears even on a focused element), so `focused` remains metadata-only; recovering it needs an upstream change.

Before:
```
- button "Disabled Button" [disabled, ref=e2]   → enabled: undefined
```
After:
```
- button "Disabled Button" [disabled, ref=e2]   → enabled: false
```

## Validation

- Red-first test observed against pre-fix code (`[disabled, ref=e2]` yielded `undefined`), green after.
- Snapshot text format verified against the real pinned agent-browser binary with a fixture page covering `disabled`, `aria-disabled`, focused input, and label-containing-brackets cases.
- Full web platform suite green (29 tests); `check:affected` green at the pushed commit.

### DIFF
--- a/src/platforms/web/__tests__/agent-browser-snapshot.test.ts
+++ b/src/platforms/web/__tests__/agent-browser-snapshot.test.ts
@@ -1,0 +1,42 @@
+import { describe, expect, test } from 'vitest';
+
+import { normalizeAgentBrowserSnapshot } from '../agent-browser-snapshot.ts';
+
+describe('normalizeAgentBrowserSnapshot', () => {
+  test('marks nodes disabled from the [disabled] text annotation', async () => {
+    const result = await normalizeAgentBrowserSnapshot({
+      snapshot: [
+        '- generic',
+        '  - button "Enabled Button" [ref=e1]',
+        '  - button "Disabled Button" [disabled, ref=e2]',
+        '  - link "Aria Disabled" [disabled, ref=e3]',
+      ].join('\n'),
+      refs: {
+        e1: { name: 'Enabled Button', role: 'button' },
+        e2: { name: 'Disabled Button', role: 'button' },
+        e3: { name: 'Aria Disabled', role: 'link' },
+      },
+    });
+
+    expect(result.nodes.map((node) => node.enabled)).toEqual([undefined, false, false]);
+  });
+
+  test('does not read disabled from bracket-like text inside labels', async () => {
+    const result = await normalizeAgentBrowserSnapshot({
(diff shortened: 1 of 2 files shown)


thymikee:
P1 — `isLineMarkedDisabled` scans every bracket group remaining on the line, not the agent-browser annotation that owns `ref`/`disabled`. This conflicts with the parser's existing trailing-value grammar: `textbox "Status" [ref=e1]: [disabled]` is a valid node whose value is the literal `[disabled]`, but this patch also reports `enabled: false`. Escaped quotes can similarly defeat the ad-hoc quote stripping and expose bracket-like label text.\n\nPlease parse `disabled` only from the structured annotation group containing the node ref (the observed format is `[disabled, ref=e2]`) rather than sanitizing the whole line. Add a regression for a trailing `[disabled]` value and, if the pinned binary escapes quotes in labels, an escaped-quote label case. Keep the existing metadata-precedence test. The intended disabled-attribute/aria-disabled mapping itself is correct.

thymikee:
Reviewed exact head b891b9db9. Code-clean: the parser derives enabled=false only from the bracket annotation after removing quoted label text, preserves explicit refs metadata precedence, and leaves unsupported focus state untouched. The regression is red against the previous parser, the pinned agent-browser format was checked live, and exact-head CI including Web Platform Smoke is green. Ready for human merge.

thymikee:
Addressed at f00c4cf7a: `isLineMarkedDisabled` now only inspects bracket groups that own the node ref (`[disabled, ref=e2]` shape), so a trailing value like `textbox "Status" [ref=e1]: [disabled]` no longer flips `enabled` — covered by a new regression asserting the value survives as `[disabled]` with `enabled` undefined. The whole-line quote-stripping scan is gone entirely, which also closes the escaped-quote exposure since label text can never contain a ref-bearing group. Metadata-precedence and label tests kept; suite green (30 tests).

PR #1943, opened 2026-08-21
refactor(recording): share the AVFoundation export pipeline between overlay and trim

thymikee:
## Summary

Second of the two follow-ups from the #1936 size audit. `recording-overlay.swift` and `recording-trim.swift` each carried their own copy of the same three mechanics — error enum, flag parsing, and the composition/export pipeline (~90 duplicated lines). This extracts `RecordingExportSupport.swift` as the single owner of those mechanics; each script keeps only its own argument grammar and export-quality policy.

- `compileSwiftSourceFile` gains `extraSourcePaths`: extra units join the cache key (path + size + content) and are passed to `swiftc`. `overlay.ts` passes the shared support file for both scripts.
- Entry points move to `@main` because multi-file `swiftc` compilation reserves top-level statements for `main.swift`.
- Behavior preservation: error messages, per-script stderr prefixes (`recording-overlay:` / `recording-trim:`), trim's error precedence (missing video track before invalid range), and the transform policies (trim propagates `preferredTransform`; overlay re-applies it via layer instruction) are all unchanged.
- Size is roughly neutral by design — this is a maintainability consolidation: one export pipeline to fix instead of two drifting copies.

## Validation

At 309e730c4, in a clean worktree:
- End-to-end on real H264 video (generated via AVAssetWriter): trim cut 500 ms correctly; overlay burned tap/swipe/pinch layers into the output; both outputs written.
- Error paths byte-identical to pre-refactor: usage text, "Input video does not contain a video track.", "--trim-start-ms must be a non-negative number".
- Both scripts typecheck as pairs with the support file (updated `recording-scripts.test.ts`); new `swift-cache.test.ts` case pins that extra sources reach the compiler and participate in cache-key identity/reuse.
- `pnpm check:affected --run`: all runnable checks passed (829 unit tests green).

Scope: 6 files, recording module group. No CLI/help/docs changes — command behavior unchanged.

### DIFF
--- a/apple/runner/AgentDeviceRunner/RecordingScripts/RecordingExportSupport.swift
+++ b/apple/runner/AgentDeviceRunner/RecordingScripts/RecordingExportSupport.swift
@@ -1,0 +1,111 @@
+import AVFoundation
+import Foundation
+
+/// Shared mechanics for the recording post-processing scripts (overlay burn-in, start trim).
+/// Each script keeps its own argument grammar and export-quality policy; this file owns the
+/// error vocabulary, flag-value reading, composition assembly, and the bounded export wait.
+
+enum RecordingScriptError: Error, CustomStringConvertible {
+  case invalidArgs(String)
+  case invalidTrimRange
+  case missingVideoTrack
+  case exportFailed(String)
+
+  var description: String {
+    switch self {
+    case .invalidArgs(let message):
+      return message
+    case .invalidTrimRange:
+      return "Trim start must be before the end of the recording."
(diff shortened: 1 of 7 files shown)


thymikee:
Reviewed exact head 309e730c4. Clean: the shared export module owns only the duplicated composition, error, option, and bounded-export mechanics; trim and overlay retain their distinct transform and quality policies. Extra Swift sources participate in both the compiler invocation and cache identity. Exact-head CI is green, and the recorded real-video trim and overlay outputs cover the changed path. Merge-ready.

Issue #1638, opened 2026-08-06
Support --settle on scroll and back (generic-route settle observation)


thymikee:
AppControlBench models habitually append `--settle` to `scroll` and `back` (both rejected with INVALID_ARGS today, burning a tool call each in bsky-16). Scroll-then-observe and back-then-observe are legitimate combos; the registry just never grew past the touch commands.

## Design (scouted; the seams are all identified)

1. **Registry**: `POST_ACTION_OBSERVATION_BY_COMMAND` (src/core/command-descriptor/post-action-observation.ts) gains `scroll: 'settle'`, `back: 'settle'`. `commandSupportsSettleObservation` then answers true; the interaction metadata/grammar helpers key off it.
2. **Grammar**: `scroll`'s `allowedFlags` override in src/commands/interaction/index.ts (`['pixels', 'durationMs']`) must append `...postActionObservationCliFlags('scroll')`; `back` lives in src/commands/system/index.ts and needs the same treatment there.
3. **Engine**: extract a target-less variant from `settleAfterInteraction` (src/commands/interaction/runtime/settle.ts) — same stable-capture loop/store/hints, but the diff baseline is the caller-supplied pre-action stored snapshot nodes instead of `resolveBaselineNodes(resolved)`, and the proximity point is optional.
4. **Wiring**: `request-generic-dispatch.ts` (the route scroll/back actually take — NOT the touch handler): when `flags.settle` and the command supports it — capture `baselineNodes = session.snapshot?.nodes` before `executeGenericPlatformCommand`; after the command and `markPostGestureStabilization` (order matters: settle's first capture then folds the #1542 stabilization), build a runtime via `createInteractionRuntime` and run the generic settle; attach the observation to `data.settle` with the same `settleRefsGeneration` dance as interaction-touch-response (ADR 0014: a settled diff publishes refs and activates a frame). `settleFlagGuardResponse` handles orphaned `--settle-quiet`/`--timeout`.

## Care points

- Settle cost on hostile screens now includes the #1627 deep extension per capture (~1.4s each) until #1626 lands — acceptable, but worth measuring in the PR.
- `back` on Android composes with the blocking-dialog pre/postflight in the generic route — settle runs after postflight.
- The diff baseline is the *stored* pre-action snapshot (possibly older than the action by several commands) — that is the honest #1101 contract ('settled diff vs the pre-action tree') but differs from press's fresh resolution baseline; the PR should say so explicitly.
[...]

PR #1945, opened 2026-08-21
refactor: derive every --settle surface from the descriptor trait (#1652)

### LINKED ISSUE #1652: Derive every --settle surface from the descriptor trait (and gate the ones that can't)
`postActionObservation` in the command-descriptor registry is already the right enumerator for "which commands support `--settle`". The problem is coverage: only four of ten caller-facing surfaces derive from it, and the gate covers two. The other six are hand-spread per command and **silently no-op when forgotten** — a reader that drops `settleInputFromFlags` doesn't fail, it just ignores `--settle`.

This is the same failure mode `commonInputFromFlags` already documents for `--no-record` ("a drop at any one of them silently disables the flag"), and the same problem `interaction-guarantees.ts` already solved for dispatch paths by making completeness a compile error.

## The drift is already in the tree

Cross-checking every surface against the trait (probe over `commandDescriptors`):

```
find         trait=false  cliFlag=false  mcpField=false  outSchema.settle=true
click        trait=true   cliFlag=true   mcpField=true   outSchema.settle=true
fill         trait=true   cliFlag=true   mcpField=true   outSchema.settle=true
longpress    trait=true   cliFlag=true   mcpField=true   outSchema.settle=true
press        trait=true   cliFlag=true   mcpField=true   outSchema.settle=true
back         trait=true   cliFlag=true   mcpField=true   outSchema.settle=true
scroll       trait=true   cliFlag=true   mcpField=true   outSchema.settle=false
```

- **`find`** advertises a `settle` payload in its MCP output schema while carrying no trait, no CLI flag, and no MCP input field. It forwards `req.flags` to its internal press/click, so a hand-built daemon request *can* produce the payload — but no public surface can ask for it.
- **`scroll`** declares nothing, purely because it isn't in the typed-result spine (`CommandResultMap`), not because it returns anything different from `back`. That one shipped in #1650 despite a gate and a careful review.

## Surface coverage

| Surface | Derived? | Hand-maintained sites |
|---|---|---|
| CLI allowed flags | yes (trait) | — |
| MCP/SDK input fields | yes (trait) | — |
| Timeout policy | yes (trait) | — |
| MCP ref-pinning | yes (trait, #1650) | — |
| CLI reader (`settleInputFromFlags`) | no | 6 |
| Client-options projection (`toSettleOptions`) | no | 4 |
| Contracts option types (`& SettleCommandOptions`) | no | 5 |
| Session-event safe-flag specs | no | 8 lines / 4 specs |
| MCP output schema | no | 5 |
| CLI output formatter | no | 4 |

## Proposed work, in order

1. **Hoist the reader spread to the framework seam.** `readInputFromCli(command, positionals, flags)` already knows the command name, so one line there deletes six hand-spreads and makes forgetting structurally impossible:
   ```ts
   return { ...reader(positionals, flags), ...settleInputForCommand(command, flags) };
   ```
2. **Derive the remaining three runtime surfaces** — session-event safe-flag specs, MCP output schema, CLI output formatter. `withSettleObservation` (added in #1650 for `back`) already generalizes from "name one command" to "wrap every settle-capable command's schema".
3. **Extend the descriptor gate to all ten surfaces**, in `post-action-observation.test.ts`. This is the item that actually matters: it converts six silent-drift surfaces into one failing test, and it would have caught both rows above. The type-level surfaces (contracts option types) can follow `interaction-guarantees.ts` and become a compile error rather than a runtime assertion.
4. **Resolve `find`** — either give it the trait (it already forwards flags to its internal dispatch) or drop `settle` from its output schema. Needs a call on intent.

## Explicitly out of scope

The two daemon orchestrations (interaction route vs generic route) look like the obvious duplication and are the wrong target. The interaction route settles *inside* the runtime command because it needs the resolution's pre-action baseline and composes with `--verify`; the generic route settles *after* dispatch against the stored session snapshot. Unifying them would mean inventing a fake resolution for scroll/back. The parts that genuinely are shared — the settle engine and `issueSettleRefs` — already are, as of #1650.

## Sizing

Roughly −30 LOC. The line count is not the point: the reason to do this is that six surfaces currently fail silently, and the seventh settle-capable command will hit the same trap. Evidence that it is a real trap rather than a theoretical one: #1650 added two commands, touched seven surfaces by hand, had a gate, and still shipped the `scroll` output-schema asymmetry.

Follow-up to #1638 / #1650.

thymikee:
## Summary

Closes #1652. Every caller-facing `--settle` surface now derives from the descriptor `postActionObservation` trait instead of being hand-spread per command, and the descriptor gate pins all of them — a forgotten cell now fails a gate instead of silently no-op'ing the flag.

- **CLI reader input**: `readInputFromCli` merges `settleInputForCommand(command, flags)` at the one seam every reader passes through; the seven per-reader `settleInputFromFlags` spreads are deleted. A new settle-capable command inherits the triple for free.
- **MCP output schemas**: the base map carries no `settle` property anywhere; `deriveSettleObservationSchemas` grafts the observation onto exactly the trait-capable entries (into each `oneOf` branch for fill). Non-trait entries keep the object identity their projection tests pin.
- **CLI output formatters**: `withSettleCapableNotes` wraps family formatter maps by trait; `back`/scroll no longer hand-pick the settle-aware formatter, and `tapCliOutput` no longer appends notes itself.
- **Session-event safe flags**: the settle/settleQuietMs projection derives from the trait instead of being re-listed in four per-command specs (`type`, which shares fill's text-entry spec but has no trait, correctly stops disclosing settle).
- **Contracts option types**: new `post-action-observation-client-options.ts` makes a dropped `& SettleCommandOptions` intersection (or a missing/extra cell) a **compile error**, following the `interaction-guarantees.ts` strategy.
- **find** (the issue's open call on intent): dropped the `settle` property from find's MCP output schema. Find carries no trait, no CLI flag, and no MCP input field for settle — advertising a payload no public surface can request was the drift. Internal flag-forwarding to delegated click/fill keeps working unadvertised; the contracts result type is unchanged.
- **Drive-by**: the dogfood help topic claimed `scroll` does not support `--settle` (stale since #1638); corrected to match the overview help and trait.

`scroll` is now typed end to end (review follow-up): `ScrollCommandResult` joins the `CommandResultMap` spine, grounded in `buildDispatchedScrollResult` and the platform leaf returns (direction, edge/passes, honored distance/timing, message, dispatcher-folded settle; platform gesture-plan coordinates ride the non-strict schema). The compiler-enforced map↔schema one-for-one invariant pulled in its output schema, and the MCP gate now asserts the trait denominator is covered by the schema keys — the motivating drift row from the issue can no longer recur silently. Still out of scope (per the issue): unifying the two daemon settle orchestrations.

## Validation

All evidence at 954c062c4 (branch rebased onto 93f8ae009):

- `pnpm typecheck` clean; `pnpm check:affected --run` fully green (413 files / 3517 tests, incl. layering, fallow dead-code, MCP parity, coverage gates).
- New gates in `post-action-observation.test.ts` observed red by planting violations, each naming the invariant: unwired schema graft ("press: settle advertisement must match descriptor trait"), unwired formatter wrap ("click: settle rendering must match descriptor trait"), unwired session spec ("click: session event must disclose settle"), seam merge disabled ("click: --settle must reach daemon input" + "settle option must survive projection"). The back-reader seam red was observed naturally: the pre-refactor `backCliReader` test failed once the spread moved to the seam, and was rewritten through `readInputFromCli`. Compile-time gate observed red by dropping `& SettleCommandOptions` from `ScrollOptions` (TS2559).
- One flake, not a regression: `provider-scenarios/ios-lifecycle.test.ts` timed out identically on a stashed clean tree in this worktree; passed in the full affected run. Provider/device lanes remain CI's authority.

Docs/skills: no user docs changed — no doc mentions `find` + `--settle`, and the corrected help text lives in versioned CLI help (source of truth).

### DIFF
--- a/packages/contracts/src/facades/interaction.ts
+++ b/packages/contracts/src/facades/interaction.ts
@@ -152,6 +152,7 @@ export {
 } from '../scroll-command.ts';
 export type {
   ScrollCommandOptions,
+  ScrollCommandResult,
   ScrollDistanceOptions,
   ScrollExecutionOptions,
   ResolvedScrollExecutionOptions,
--- a/packages/contracts/src/scroll-command.ts
+++ b/packages/contracts/src/scroll-command.ts
@@ -1,4 +1,6 @@
 import { AppError } from '@agent-device/kernel/errors';
+import type { SettleObservation } from './interaction.ts';
+import type { ScrollDirection } from './scroll-gesture.ts';
 
 export const SCROLL_DURATION_MAX_MS = 10_000;
 export const DEFAULT_MOBILE_SCROLL_DURATION_MS = 300;
@@ -67,3 +69,26 @@ export function honoredScrollDurationMs(
 ): number | undefined {
   return typeof result?.durationMs === 'number' ? result.durationMs : undefined;
 }
+
+/**
+ * `scroll` — the generic-route result built by `buildDispatchedScrollResult`
+ * (src/core/dispatch-interactions.ts): the resolved direction, the edge-pass
+ * bookkeeping for `top`/`bottom` scrolls, the honored distance/timing echo,
+ * and the success message. Platform leaves add gesture-plan coordinates
+ * (`x1`/`y1`/`x2`/`y2`, reference frame) on top; the output schema stays
+ * non-strict so those additive fields validate. The one field the dispatcher
+ * itself may add: `settle`, the opt-in `--settle` observation (#1638),
+ * attached after the command — same shape as `BackCommandResult`.
+ */
+export type ScrollCommandResult = {
+  direction: ScrollDirection;
+  /** Set for `top`/`bottom` requests: the extreme being scrolled to. */
+  edge?: 'top' | 'bottom';
+  /** Edge scrolls only: how many scroll-and-check passes ran. */
+  passes?: number;
+  amount?: number;
+  pixels?: number;
+  durationMs?: number;
+  message?: string;
+  settle?: SettleObservation;
+};
--- a/src/cli-schema/cli-help-topics.test.ts
+++ b/src/cli-schema/cli-help-topics.test.ts
@@ -521,10 +521,8 @@ test('usageForCommand resolves dogfood help topic', async () => {
   assert.match(help, /direct Android localhost URL opens with a port auto-configure/);
   assert.match(help, /Keep stateful commands serial within the same session/);
   assert.match(help, /agent-device wait 'role=tab' 10000/);
-  assert.match(
-    help,
-    /scroll takes direction then amount and does not support a selector or --settle/,
-  );
+  assert.match(help, /scroll takes a selector-less direction\+amount form/);
+  assert.match(help, /Use --settle to wait for the UI to go quiet/);
   assert.match(help, /prefer agent-device open "Expo Go" <url>/);
   assert.match(help, /dogfood-output\/report\.md/);
   assert.match(help, /ID, severity, category, title, affected flow\/screen/);
--- a/src/cli-schema/cli-help.ts
+++ b/src/cli-schema/cli-help.ts
@@ -970,7 +970,7 @@ Rules:
   Findings must come from observed runtime behavior, not source reads.
   After each mutation, use the --settle diff as evidence when available; otherwise re-snapshot.
   Wait timeouts are integer milliseconds in the trailing positional: agent-device wait 'role=tab' 10000. Do not write duration suffixes such as 10s.
-  scroll takes direction then amount and does not support a selector or --settle: agent-device scroll down 3.
+  scroll takes a selector-less direction+amount form: agent-device scroll down 3. Use --settle to wait for the UI to go quiet and get the settled diff.
   Keep commands in the report reproducible; use selectors or refs from fresh snapshots, not guessed coordinates.
   Prefer refs for exploration and selectors for deterministic replay.
   Use logs, network, screenshot --overlay-refs, trace, perf frames, perf memory, native profiles, or react-devtools only when they add evidence to a specific issue.
--- a/src/commands/cli-grammar/common.ts
+++ b/src/commands/cli-grammar/common.ts
@@ -140,6 +140,8 @@ export function selectorSnapshotOptionsFromFlags(flags: CliFlags): SelectorSnaps
 // Descriptor post-action observation commands use --settle (#1101).
 // --timeout doubles as the settle deadline only when --settle is present; a
 // bare --timeout stays compatible and is ignored by touch commands.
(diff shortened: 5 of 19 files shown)


thymikee:
P1 — The motivating scroll output-schema drift from #1652 is still unresolved, while this PR closes the issue and claims every caller-facing settle surface is derived. deriveSettleObservationSchemas only iterates CommandResultMap; scroll is absent from that map, so the MCP tool still receives no settle-capable output schema. The new gate also iterates only existing COMMAND_OUTPUT_SCHEMAS entries, so it cannot notice that a trait command is missing entirely. Please give scroll an accurate result-spine/output-schema entry and assert that the trait denominator is covered by the schema keys, or keep #1652 open with a separately tracked remainder and narrow this PR completeness claim.

thymikee:
Addressed at 7d95f7e33 — took the first option: scroll now has an accurate result-spine/output-schema entry, and the gate asserts the trait denominator.

- `ScrollCommandResult` (packages/contracts/src/scroll-command.ts) is grounded in a re-read of `buildDispatchedScrollResult` and the platform leaf returns: resolved `direction`, `edge`/`passes` for top/bottom scrolls, honored `amount`/`pixels`/`durationMs`, `message`, and the dispatcher-folded `settle`. Platform leaves add gesture-plan coordinates on top; the type and schema document that honestly (schema stays non-strict, so those additive fields validate).
- `scroll: ScrollCommandResult` added to `CommandResultMap`; the compiler-enforced one-for-one invariant with `COMMAND_OUTPUT_SCHEMAS` pulled the schema in. The settle observation itself is still not hand-listed — `deriveSettleObservationSchemas` grafts it because scroll carries the trait.
- The MCP output-schema gate now asserts coverage of the trait denominator first (`scroll: settle-capable command must have an output schema entry`), then per-entry advertisement both directions. Observed red by deleting the scroll entry.

`pnpm typecheck` clean; `pnpm check:affected --run` fully green (413 files / 3517+ tests) at 7d95f7e33.

thymikee:
Re-reviewed exact head 7d95f7e33. The missing scroll result/schema ownership is fixed: ScrollCommandResult now joins the command-result map, the MCP schema covers the dispatched result, and the trait-denominator gate fails if a future settle-capable command lacks a schema. No remaining code finding. Android failed after the emulator went offline and iOS failed the existing email fixture assertion after typing succeeded; rerun both device lanes before merge.

thymikee:
Exact head 7d95f7e33 is now fully green after the two device reruns. The prior Android emulator-offline failure and iOS fixture mismatch did not recur; the descriptor-derived settle surfaces and scroll result/schema follow-up remain code-clean. Merge-ready.

PR #1947, opened 2026-08-21
fix(ios): preserve regular snapshot depth through structural wrappers

### LINKED ISSUE #1797: iOS snapshot backends: make interpretation divergence impossible by construction (acquire/present split)

thymikee:
## Summary

Implements the ADR 0004 regular-snapshot visible-depth frontier for #1797. This PR does not claim to close #1797.

- Keeps raw traversal depth as the acquisition bound while regular `--depth` is enforced against the presented tree. A visible raw-deep child can therefore remain in `snapshot --depth 1` when presentation collapses structural ancestors.
- Reuses the existing `SnapshotVisibilityFold` transition instead of adding arbitrary acquisition depth, node caps, or backend-specific exceptions. Raw projection and backend capability declarations remain unchanged.
- Preserves the presentation failure contract through the existing typed snapshot-quality reason and warning path.
- Adds focused XCTest coverage plus a fixture-backed public CLI scenario. XCTest owns the clipped-structural-parent invariant; the live scenario proves that regular depth 1 presents a raw-deep child at depth 1 while raw depth 1 still omits it.
- Scope is 27 files across the iOS runner/tests, existing TypeScript snapshot capability contract, ADR/guidance, XCTest workflow selection, and the fixture/live scenario. It does not expand into another platform family or command family.

## Validation

- Exact published head: `0b0a029c90fb6aba31c32009ab4783cf99333d9f`, based on `cb65d6ca1f18683c7e9cb60e6e454bb974d23154`.
- Test-first evidence: the focused XCTest regression was observed red before the visibility-aware frontier existed and green after the production recursive-tree wiring.
- `pnpm check:affected --run` passed every locally runnable check on the published head, including formatting, lint, typecheck, layering, fallow, package verification, 2,806 related tests, XCTest selection, and the gate/fixture models. The focused iOS smoke coverage contract also passed 14/14.
- The prior exact-head iOS smoke failure was diagnosed from its raw payload: React Native/XCTest exposed the visual wrapper and child as AX siblings, invalidating the scenario's ancestry assertion. The fixture and scenario now test only the public depth contract; no production behavior was changed to reconstruct visual ancestry.
- Exact-head native/device lanes remain GitHub-authoritative and are running.
- Latest size evidence before this test-only follow-up: npm unpacked `+8.4 kB` (`+8.3 kB` Apple runner source/project, `+111 B` JavaScript), npm tarball `+1.9 kB`, JavaScript gzip `+48 B`. The shared-fold design is the smallest owning-interface change considered.


thymikee:
P1: the new acquisition frontier counts regularPresentedDepth using semantic eligibility before SnapshotVisibilityFold applies visibility and clipping. A labeled or interactive node outside the viewport therefore consumes the requested depth, shouldAcquireChildren stops traversal, and the fold can then remove that node, losing an independently visible descendant that regular presentation would otherwise reparent. The committed wrapper test covers only an in-viewport ineligible wrapper, so it cannot catch this. Drive a fixture with an eligible but clipped parent and a visible independently-projected child; acquisition must continue until the requested visible presented depth is complete, or the frontier must use the same visibility decision as presentation without duplicating that policy. Please also record exact-head live depth evidence and itemize the current +4.6 kB unpacked growth.

thymikee:
P1: the visibility-aware implementation resolves the original clipped-parent logic, but the new regression still does not exercise acquisition. It constructs a fully acquired three-node RawAX tree and manually reenacts traversalDecision, regularPresentedDepth, and shouldAcquireChildren; reverting the recursiveTreeSnapshotAcquisition call-site/state propagation in RunnerTests+Snapshot.swift to the old eligibility-only frontier leaves this test green. Please put the frontier transition behind one pure helper used by the production traversal and drive the clipped-parent fixture through that helper, with recorded red-on-pre-fix evidence, or add an exact-path simulator/live regression proving the child is actually acquired.

The test is also absent from the ios.yml only-testing list, while the host lane evaluates platformDefault as plainViewport rather than the iOS cursorProjected policy, so current CI does not cover the changed iOS path. The generic iOS smoke is green but does not exercise clipped-parent snapshot --depth. Please also update the PR body to exact head 9e3686ec and 16 files, record exact-head live depth evidence, and itemize the current +7.6 kB npm-unpacked growth with why the shared-fold design is the smallest viable implementation.

thymikee:
P1: the clipped-parent regression now proves the pure helper, but it still does not exercise recursiveTreeSnapshotAcquisition. Reverting that production loop to the old depth-only frontier (or dropping its parentPresentedDepth / parentTraversal propagation) while leaving regularTraversalTransition intact keeps this test green, because it starts from a fully materialized RawAXNode tree. The exact-head iOS job selected and passed it, but the PR body still says the required public snapshot --depth 1 case is pending. Please add an exact production-path regression, preferably a fixture-screen iOS test that invokes regular snapshot --depth 1, and record pre-fix red or live evidence. It must observe root plus the independently projected child at presented depth 1 with the clipped structural parent absent; raw --depth 1 must still omit the raw depth-2 child.

P2: update the PR body size figure from +7.6 kB to the exact-head Size report: +8.4 kB npm unpacked (+8.3 kB Apple runner source/project). The shared-fold rationale is useful; it should accompany the current number.

thymikee:
Re-review at 0bfdf57e948bbf1e20226254c5c7d5d5be7bcc03: the prior production-path P1 is fixed in design. The new public-CLI scenario is non-vacuous: full raw proves the clipped parent/descendant and raw depth >1; regular --depth 1 requires the parent absent and child projected at root depth 1; raw --depth 1 requires the child absent and all nodes bounded to raw depth 1. Reverting the acquisition frontier would fail it. Size evidence is current at +8.4 kB unpacked / +8.3 kB Apple runner with adequate minimality rationale.

Readiness remains pending only on evidence: the exact-head iOS smoke is still running, and the PR body still stamps 5f5b1ee... / says live evidence pending rather than current head 0bfdf57e.... Once that fixture-backed iOS smoke passes and the body records the exact-head result, this is clean.

thymikee:
Re-review of exact head `209ab52e6bc06e434778869bdd36488867f7f18a`: not ready yet.

- **P1 — restore the regression’s raw ancestry proof.** The latest scenario only proves that the clipped-parent and projected-child identifiers both exist somewhere and that the child is deeper than 1. Without proving the clipped parent is on the child’s raw `parentIndex` chain, an unrelated/de-flattened raw branch could let the old acquisition frontier pass. Walk the child’s raw ancestor chain and require `visible-depth-clipped-parent` to occur; immediate parenthood is not required.
- **Owner-action CI:** lint passed, but format deterministically fails on `examples/test-app/src/screens/VisibleDepthScreen.tsx`; run the repository formatter and push the result.
- Refresh the PR body/evidence: it still stamps head `5f5b1ee...` / base `320a881...`, says 25 files, and describes live evidence as pending; current is `209ab52...` on `cb65d6c...` with 27 PR files.

The failed Android smoke is an unrelated `wait_capture_stalled` / `readableCaptures: 0` infrastructure flake. Exact-head iOS and Coverage are still pending. The production implementation remains clean relative to the prior design review.

thymikee:
Addressed the re-review blockers in `71cd9c32b5567b2a0dc5bd5ead568e94d1770897`: the live scenario now walks the raw `parentIndex` chain and proves the clipped wrapper is an ancestor, the fixture is repository-formatted, and the PR body is stamped with current evidence. `pnpm check:affected --run` passed all locally runnable gates; current-head CI is running.

thymikee:
Follow-up on the iOS smoke failure: the exact-head raw payload showed React Native/XCTest flattening the visual wrapper and child into AX siblings, so the ancestry assertion was invalid. I removed that false assumption and simplified the fixture. Focused XCTest remains the owner of the clipped-parent projection invariant; the public live scenario now proves the actual CLI contract: the child is raw-deep, appears at presented depth 1 in a regular snapshot, and is omitted by raw depth 1. Published as 0b0a029c9; all locally runnable affected checks pass, and exact-head iOS CI is running.

PR #1949, opened 2026-08-21
fix: Linux input/a11y defects from #1935 (click miss, typed '=', GTK4 text)

thymikee:
## Summary
Fixes two Linux platform defects surfaced by CI on #1935 (Linux Smoke lane), confirmed live via instrumented CI runs before being fixed:

1. **Resolved click misses its target** on gnome-calculator digit buttons — `Component.get_extents(Atspi.CoordType.SCREEN)` returns `(0, 0)` as the origin for every non-toplevel widget under this GTK4 build (confirmed live: a raw click at the computed rect center landed on the window's own header-bar button instead). Fixed by computing rects from `CoordType.WINDOW` plus the enclosing top-level frame's own (correct) screen origin.
2. **GTK4 apps expose no editable text** through the AT-SPI dumper — `accessible.get_text_iface().get_text(...)` throws `"Atspi.Accessible.get_text() takes exactly 1 argument (3 given)"`, a documented PyGObject binding collision with the deprecated 1-arg `Accessible.get_text()`, silently swallowed as "no text". Fixed by calling the unbound `Atspi.Text.get_text(accessible, ...)` form instead.

The third suspected defect ("typed `=` never arrives") did not hold up: the diagnostic evidence gathered to investigate it (CI run 32503838660) actually shows `=` delivered correctly in every multi-character `type` call tested. An earlier version of this PR added an unconditional 150ms keyboard-settle delay based on a misreading of that evidence (a lone `=` sent to an *empty* calculator entry producing a blank display is normal calculator semantics, not a dropped key) — removed per review. The original "100+55" screenshot (run 32487868346) most likely reflects that run's attempt-3 hitting the already-fixed `mousemove --sync` no-op hang, not an independent keyboard-dispatch bug.

An earlier version also narrowed Linux's `hittable` computation to exclude AT-SPI role `"label"`, to stop click-target promotion from walking into a button's same-rect caption child. Removed per review: the coordinate fix alone already resolves the observed miss (the button and its label child resolve to near-identical centers), and nothing in this PR's evidence proves the promotion path itself needs tightening.

The Linux smoke replay is restored to exercise both real fixes together (click a resolved digit button, type a full calculation, wait on the computed result through the tree) instead of staying at the weakened, contract-tier assertions the defects had forced. The coverage manifest promotes `click` and `type` from command-contract to live accordingly.

**Size**: +3.0 kB npm unpacked, entirely from `linux/atspi-dump.py` growing 8.2 kB → 11.1 kB — load-bearing comments on `get_rect`/`get_screen_origin`/`get_text_value` documenting the two confirmed upstream AT-SPI/GTK4 bugs this works around (matching this file's existing convention, e.g. the `moveTo` docstring).

## Test plan
- [x] Root-caused all three suspected defects via instrumented CI runs on real hardware (not local reproduction — no Linux/X11/AT-SPI available locally)
- [x] Implemented and live-verified fixes for the two real defects (click coordinates, GTK4 text exposure)
- [x] Removed the unproven keyboard-settle and hittable changes per review
- [x] Removed diagnostic CI instrumentation
- [x] Linux Smoke lane green (exact-head validation, `attempts: 1`, `healed: 0`)


thymikee:
Not ready at `9d9e027c`.

**[P1] Remove the unconditional 150 ms keyboard settle or prove a real race.** The cited diagnostic artifact from run `32503838660` already shows `100+55=` computing `155` without a settle, and `5=5` producing history `5 = 5`, which proves `=` was delivered. Sending `=` alone to a calculator and observing a blank display is normal calculator semantics, not evidence of a dropped key. The new replay still sends `00+55=` as one multi-character burst, so deleting both sleeps would leave this regression green while the change adds 150 ms to every local Linux `sendKey` and `typeLinux`, including navigation and fill. Please remove the workaround, or replace the evidence with a change-sensitive scenario that observes a genuinely lost first keystroke after a real focus transition and goes red without the production change.

**[P2] Tighten the click fix to the behavior that failed.** The corrected AT-SPI coordinates fix the observed click miss. The additional global `role_name != "label"` hittability change is not required for that result: same-rect descendant promotion preserves the click center, and the replay cannot distinguish whether it targeted the button or its label child. Remove this extra semantic change unless a target-identity regression proves an externally observable failure that requires it.

The exact-head Linux Smoke is green and meaningfully validates the `WINDOW + frame origin` coordinate fix and the static `Atspi.Text.get_text` fix. After tightening, refresh the stale diagnostic-only PR body/checklist and account for the roughly +3.0 kB unpacked growth.

thymikee:
Re-reviewed exact head `c7256408`.

The two prior code blockers are resolved: the unconditional keyboard-settle sleeps are gone, and Linux label nodes retain their existing hittability semantics. The remaining implementation is narrowly owned by the AT-SPI dumper: `WINDOW + top-level SCREEN origin` supplies usable click geometry, while the unbound `Atspi.Text` call restores GTK4 text exposure.

Exact-head Linux Smoke is meaningful and green: the JUnit artifact records `attempts: 1`, `healed: 0`; the replay clicks the resolved `1` button, types `00+55=`, and only passes after `155` becomes selectable through the live tree. All required checks are now green, the branch is mergeable, and CI reports only +2.5 kB npm unpacked growth.

Code review is clean and this is ready for human review. One metadata cleanup before squash merge: please remove `typed =` from the PR title, since the final evidence correctly concludes that defect did not exist. If keeping exact size figures in the body, align them with CI’s +2.5 kB / 10.6 kB report.

PR #1950, opened 2026-08-21
refactor(daemon): lazy-load platform cleanup helpers in session teardown

thymikee:
## Summary

Session teardown is shared by every platform, but it statically imported the android perf cleanup, the android snapshot helper, and the apple xctrace cleanup — so all of them sat in teardown's eager import closure whether or not any capture ran. The android helpers now load via `await import` behind their existing guards, matching the established `register-builtins` interactor lazy pattern. The apple helper stays static.

**This is a seam-hygiene and regression-pin change, not a measurable speedup.** Measured honestly: teardown's module import takes ~0.10s on main vs ~0.11s here — the lazied modules are individually cheap, and no suite-level wall-clock gain is claimed. The value is:

1. **The pin**: `session-teardown-import-closure.test.ts` fails with an explanatory message if either android helper is restored to a static import (observed red against main's static imports before green here).
2. **Shared machinery**: the eager-closure walker from `cli-startup-import-closure.test.ts` is extracted to `eager-import-closure.fixtures.ts` so future lazy-seam claims can pin themselves the same way; the CLI test now consumes the fixture.
3. **Hygiene**: teardown's eager closure (146 → 121 repo modules, `module.registerHooks` counter) stays platform-neutral as the capture subtrees grow.

The apple `perf-xctrace` hunk was dropped per review: the existing Apple unit test mocks the cleanup module, so it could not prove the dynamic chunk loads on the shipped path.

## Validation

- Planted-red check on the final implementation: restoring only the two production static imports made the eager-closure guard fail while the non-vacuity test remained green; restoring the dynamic imports returned both tests to green.
- Teardown graph measured 146 (main) → 121 (branch) with a `module.registerHooks` load counter.
- `session-teardown-resources`, `lease-lifecycle`, `daemon-client-lifecycle`, `daemon-runtime-device-claims`: 39 tests green; CLI startup closure tests green against the extracted fixture (22/22).
- `pnpm check:affected --run`: all runnable checks passed (143 files, 837 tests in the related-test lane).

Scope: 4 files (1 source, 1 new fixture, 1 rewritten test, 1 new test).

### DIFF
--- a/src/__tests__/cli-startup-import-closure.test.ts
+++ b/src/__tests__/cli-startup-import-closure.test.ts
@@ -1,10 +1,7 @@
 import { expect, test } from 'vitest';
 import fs from 'node:fs';
 import path from 'node:path';
-import {
-  eagerClosureOf,
-  eagerlyEvaluatedModules,
-} from './eager-import-closure.fixtures';
+import { eagerClosureOf, eagerlyEvaluatedModules } from './eager-import-closure.fixtures.ts';
 
 /**
  * Every CLI invocation eagerly evaluates the static import closure of
--- a/src/__tests__/cli-startup-import-closure.test.ts
+++ b/src/__tests__/cli-startup-import-closure.test.ts
@@ -1,7 +1,10 @@
 import { expect, test } from 'vitest';
 import fs from 'node:fs';
 import path from 'node:path';
-import { parseSync } from 'oxc-parser';
+import {
+  eagerClosureOf,
+  eagerlyEvaluatedModules,
+} from './eager-import-closure.fixtures';
 
 /**
  * Every CLI invocation eagerly evaluates the static import closure of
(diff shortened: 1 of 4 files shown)


thymikee:
The lazy-load shape itself looks correct, and Android Smoke exercises a persistent helper snapshot followed by session close. I cannot call the refactor ready yet because the 146 → 84 import-graph improvement exists only in the PR description: restoring these three static imports leaves every cited test and gate green. Please add the smallest deterministic import-closure regression using the existing graph machinery, prove it red on origin/main, and keep the new lazy seam pinned. Also either drop the extra Apple xctrace hunk or add live start-capture → close evidence; the current Apple unit test mocks the cleanup module, so it does not prove the shipped dynamic chunk is loadable on that path. iOS/macOS Smoke and Swift Host XCTests are still pending.

thymikee:
**Review (xhigh effort, multi-angle + adversarial verification).** Mechanically the conversion is faithful — signatures, guards, and awaits all check out (`tsc --noEmit`, 92 related tests, `check:layering` 181/181, oxfmt/oxlint all green with the PR file in place). Six findings, ranked:

### 1. CONFIRMED · correctness — `stopSessionApplePerfCapture` re-reads `session.applePerf.active` after the new yield point

The inline hunk puts an `await import()` between the guard and the use, then **re-reads** `session.applePerf.active` (`session-teardown.ts:49-51`). Pre-PR the guard and the argument read were same-tick; the Android sibling snapshots `const active = …` before any await and is immune. The interleaving is reachable: `perf` commands run under the per-session execution lock, but signal/`uncaughtException`-triggered shutdown (`teardownDaemonSessionForShutdown`, `daemon-runtime.ts:116-186`) calls `teardownSessionResources` **without** that lock, and `SessionStore.get()` documents it returns the live record. A mid-flight perf handler mutating `applePerf` (`session-perf-xctrace.ts:159`) racing shutdown teardown can make this clean up the wrong capture and drop the reference to the one it guarded on. Fix is one line — capture before the import, mirroring the Android shape:

```ts
const active = session.applePerf?.active;
if (!active) return;
const { cleanupAppleXctracePerfCapture } = await import('../platforms/apple/core/perf-xctrace.ts');
await cleanupAppleXctracePerfCapture(active);
```

### 2. CONFIRMED · efficiency — in the daemon process all three imports are warm-cache hits; the win is the importer/test graph only

- `daemon-runtime.ts:36` still **statically** imports `platforms/android/snapshot-helper.ts` (used in `shutdown()`), so that module is resident on 100% of daemon boots regardless of this PR.
- Both perf modules are statically reachable from the session handler bucket: `handlers/session.ts:25` → `session-observability.ts:42-43` → `session-native-perf.ts:17` / `session-perf-xctrace.ts:11`. That bucket (`request-handler-chain.ts:63`) must load before any session can exist, so by the time teardown sees an active capture the modules are always loaded. Ordinary `session close` (`session-close-lifecycle-teardown.ts:152-154`) exercises all three stops on the hot path too.
- Measured: the daemon bootstrap graph shrinks only 165 → 157 modules.

The 146→84 number for this module's own graph is real and the test-graph motivation stands, but the code comment ("every teardown caller pays this module's graph") reads as a runtime claim. Suggest wording it as: benefit = the static import graph of this module's (mostly test) importers; the daemon itself loads these modules via other paths.

### 3. CONFIRMED · simplification — inline the two `*Lazy` wrappers

Each wrapper has exactly one call site, and the file's own third conversion (the Apple hunk) already uses the inline shape — as do `register-builtins.ts` (which the comment cites, but it inlines; it never extracts wrappers) and `platform-runtime-android-application-tools.ts:19-20` for the very same `platforms/android/perf.ts` module. Inlining at the call sites (where `active` and `session.device` are already narrowed) removes ~17 net lines and the `NonNullable<NonNullable<SessionState['nativePerf']>['android']>` expression, which hand-rebuilds the exported `AndroidNativePerfSession` type (an `import type` would be erased at build and can't defeat the graph goal).

### 4. PLAUSIBLE · latent — fail-fast at boot is now an unenforced load-order invariant

Today a broken platform module still fails at boot only because of the static paths in finding 2. If a future refactor slices `session-observability.ts` or makes those handler imports lazy, a load failure would first surface inside `runIsolatedSessionCleanup`'s catch — including during shutdown's 5s `Promise.race` (`daemon-runtime.ts:171-177`), which exits without waiting and would abandon the capture child the step exists to kill. Worth one sentence in the comment naming the invariant.

### 5. PLAUSIBLE · altitude — this dispatch shape has a home on the `PlatformPlugin` facet seam

Per-platform teardown branching in daemon code is the shape ADR-0009 moved `perf`/`providers` onto plugin facets for (with parity tests, e.g. `session-perf.ts:184-187`). A `cleanup` facet would carry the lazy import plus the compile-time totality check when Vega/Harmony grow teardown needs. Fine to keep this PR minimal — noting the seam for the next addition.

### 6. PLAUSIBLE · altitude — nothing enforces the new convention

Layering R3 (`scripts/layering/zone-policy.ts:60-69`) explicitly permits `src/daemon/` static value imports of `platforms/`, so reverting any of these to static stays green, and ~7 sibling daemon files still hold static platform imports (`device-ready.ts`, `request-router.ts`, `selector-runtime.ts`, `apple-runner-options.ts`, `android-system-dialog.ts`, `request-platform-providers.ts`, `request-recording-health.ts`). The measured win has no ratchet; whether that deserves machinery is a judgment call given it's a test-graph optimization.

---
Verified clean along the way: wrapper arg order/types vs real signatures; vi.mock interception (hoisted mocks only, unaffected by import style); double-invocation idempotency of both cleanup paths (`SIGINT` on exited child is a no-op; shared `capture.wait` promise; `kill -0 || exit 0`); no module-level side effects in the three barrels; dynamic `.ts` specifier style matches ~150 existing call sites across dist/strip-types/vitest.

thymikee:
Addressed both review asks in 969c945f1:

1. **Import-closure regression pin**: the AST walker from `cli-startup-import-closure.test.ts` is extracted to `eager-import-closure.fixtures.ts`, and `session-teardown-import-closure.test.ts` now fails with an explanatory message if either android helper re-enters teardown's eager closure. Observed red on origin/main's static imports (guard + non-vacuity tests both failed) before green here.
2. **Apple xctrace hunk dropped** — reverted to the static import.

One correction to my own framing: on your implicit question about payoff — measured warm import of `session-teardown.ts` is ~0.10s (main) vs ~0.11s (branch), so this is not a measurable performance change; the PR description no longer implies one. What it buys is the pinned seam and the shared closure machinery. Happy to close if that's not worth a PR on its own.

thymikee:
Re-review at `92d6fb66`: the deterministic closure pin and Apple-hunk removal resolve the prior blockers; all exact-head CI is green. One P2 remains: please inline the two Android `await import()` calls in their sole teardown callers. The wrappers only forward arguments and reconstruct the native-perf type; keeping each import next to its existing guard is smaller and clearer, and the closure test still pins the seam. After that, G2G.

thymikee:
The production shape is clean now: both Android imports are inline beside their existing guards, native-perf snapshots the active capture before yielding, teardown failure isolation is unchanged, the Apple hunk is gone, and all exact-head CI plus Android live smoke are green. One validation item still needs correction before G2G: the PR body says both the closure guard and its non-vacuity test failed on origin/main. Restoring these two static imports should make only the guard fail; the closure only grows, so the >20 lower bound and materialized-path-registry reachability must stay green. If both tests failed, that red run included an unrelated setup/module-resolution failure and is not the required planted-red proof. Please rerun with the current test machinery and only the two production imports restored, then record guard red + non-vacuity green in Validation. As a final optional tightening, remove the unused parseEagerModule export and keep FUNCTION_NODES/resolveRelative private.

thymikee:
Finished the requested cleanup at `853622b8c`.

- Removed the unused parser export and kept the closure fixture's internal helpers private.
- Re-ran the planted regression on the final implementation: restoring the two production static imports fails only the eager-closure guard; the non-vacuity test stays green. Restoring dynamic imports returns 22/22 focused closure tests to green.
- `pnpm check:affected --run` passed all runnable checks (143 related files, 837 tests). GitHub-authoritative lanes can continue in CI.

The PR description now records the corrected red/green evidence.

thymikee:
Exact head 853622b8c is now fully green: all 28 required checks passed, including Coverage, Integration, iOS, Android, macOS, Linux, and package/static gates. The branch is mergeable and the review findings are resolved — merge-ready.

PR #1951, opened 2026-08-21
fix(web): scope disabled detection to the ref-bearing annotation group

thymikee:
## Summary

Keep `[disabled]` detection scoped to the node annotation that owns its browser ref. Literal bracket text in labels or trailing values remains user data and no longer changes `enabled`.

Ref extraction and disabled-state detection now share one annotation parser instead of independently reconstructing the snapshot grammar.

## Validation

- Observed the trailing-value regression fail against the pre-fix implementation (`enabled: false` instead of `undefined`).
- Focused web snapshot suite: 5/5 passed, including quoted labels and a ref-shaped trailing value.
- `pnpm check:affected --run`: all runnable checks passed; 259 related files and 1,672 tests passed.
- GitHub-authoritative coverage and web smoke are running on the updated head.

### DIFF
--- a/src/platforms/web/__tests__/agent-browser-snapshot.test.ts
+++ b/src/platforms/web/__tests__/agent-browser-snapshot.test.ts
@@ -31,6 +31,28 @@ describe('normalizeAgentBrowserSnapshot', () => {
     expect(result.nodes[0]?.label).toBe('Shows [disabled] in its label');
   });
 
+  test('does not read disabled from a trailing value outside the annotation group', async () => {
+    const result = await normalizeAgentBrowserSnapshot({
+      snapshot: '  - textbox "Status" [ref=e1]: [disabled]',
+      refs: { e1: { name: 'Status', role: 'textbox' } },
+    });
+
+    expect(result.nodes[0]?.enabled).toBeUndefined();
+    expect(result.nodes[0]?.value).toBe('[disabled]');
+  });
+
+  test('does not read disabled from a label that embeds a ref-shaped annotation', async () => {
+    // Live-derived from the pinned agent-browser: an enabled button whose
+    // aria-label is "Shows [disabled, ref=e1]" emits exactly this line.
+    const result = await normalizeAgentBrowserSnapshot({
+      snapshot: '  - button "Shows [disabled, ref=e1]" [ref=e1]',
+      refs: { e1: { name: 'Shows [disabled, ref=e1]', role: 'button' } },
+    });
+
+    expect(result.nodes[0]?.enabled).toBeUndefined();
+    expect(result.nodes[0]?.label).toBe('Shows [disabled, ref=e1]');
+  });
+
   test('metadata enabled still wins over the text annotation', async () => {
     const result = await normalizeAgentBrowserSnapshot({
       snapshot: '  - button "Odd" [disabled, ref=e1]',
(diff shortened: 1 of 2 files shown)


thymikee:
P1 — The new scan still trusts any ref-shaped bracket group anywhere on the line. The pinned agent-browser reproduces this with an enabled button whose aria-label is Shows [disabled, ref=e1]: it emits - button "Shows [disabled, ref=e1]" [ref=e1], and this implementation reads the label group and returns enabled: false. That can make the shared enabled guard refuse a valid action. Please parse the structural node annotation, meaning the ref group immediately before the optional value or end of line, rather than arbitrary bracket text, and add this live-derived label regression alongside the trailing-value case.

thymikee:
Addressed at fd91722b9: reproduced live against the pinned binary (`- button "Shows [disabled, ref=e1]" [ref=e1]` from an enabled button whose aria-label embeds that string), added it as a red-first regression, then reworked `isLineMarkedDisabled`: quoted label text is stripped escape-aware first, and only the **last remaining ref-bearing group** — the structural annotation immediately before the optional value/EOL — is inspected for `disabled`. Label groups can therefore never match, closing this case and the escaped-quote variant together. All prior regressions kept; 31 web tests green, `check:affected` + `check:fallow` green.

thymikee:
P1 — `extractBrowserRef` still parses ref-shaped content inside quoted labels. A label can legally contain `[disabled, ref=e1]` while its owning annotation is `[ref=e2]`; this function selects `e1`, causing the actual `e2` node to be emitted as a fallback (and potentially receiving the wrong metadata/target). Reuse the quote-aware structural-annotation parsing for ref extraction too, and regress the distinct-ref case (`…[ref=e1]" [ref=e2]`) asserting one node keyed by `e2`. The current test uses `e1` for both and cannot expose the bad association.

thymikee:
P1: The quote-aware fix only applies to disabled detection; `extractBrowserRef` still scans the entire raw line. For `- button "Shows [disabled, ref=e1]" [ref=e2]`, it selects quoted `e1`, marks that ref seen, and emits the real `e2` metadata as a fallback node. The current regression uses e1 in both places, so it cannot catch the wrong association. Please parse the structural ref-bearing annotation once (outside quoted text and before the optional trailing value) and reuse it for both ref extraction and disabled detection; add a distinct-ref regression asserting exactly one e2 node. This also avoids treating a trailing literal `[disabled, ref=e99]` value as the node annotation.

thymikee:
The detector still scans through the optional trailing value and assumes the final ref-bearing bracket group is structural. A valid line such as `textbox "Status" [ref=e1]: [disabled, ref=e2]` is parsed by `extractTrailingAriaValue` as value `[disabled, ref=e2]`, but `isLineMarkedDisabled` selects that value group and incorrectly emits `enabled: false`. Scope annotation detection to the pre-value portion / actual node annotation rather than all bracket groups on the line, and extend the trailing-value regression with this ref-shaped value. The rest of the small change is focused and exact-head CI is fully green.

thymikee:
Simplified the fix in b4b5b1e74: browser-ref extraction and disabled detection now share one structural annotation parser, the explanatory block comment is gone, and the trailing-value regression now includes its own ref-shaped group. The regression was observed red on the pre-fix implementation; the focused suite and all runnable affected checks pass (1,672 related tests).

PR #1954, opened 2026-08-22
docs: drop stale apps.ts over-budget warning from AGENTS.md

thymikee:
## Summary

`src/platforms/apple/core/apps.ts` was decomposed into `app-resolution.ts` / `app-launch.ts` / `app-device-io.ts` / `app-settings.ts` and is now a 15-line barrel. The module-topology warning telling contributors to extract before adding behavior there is stale and would misdirect future work; `session.ts` (478 lines) is still over budget and keeps its entry.

## Validation

Docs-only — no runtime validation applies. Verified against the tree at main: `wc -l src/platforms/apple/core/apps.ts` = 15.


thymikee:
Reviewed at 6993b60: apps.ts is now a 15-line barrel, so removing its stale over-budget warning is accurate while retaining the session.ts warning. Docs-only checks are green. LGTM.

PR #1956, opened 2026-08-22
refactor(tests): replace the test-utils barrel with direct module imports

thymikee:
## Summary

Follow-up to the import-graph findings discussed in #1950's arc. `src/__tests__/test-utils/index.ts` re-exported 13 modules (~40 names), so all ~130 importers eagerly evaluated every backing module — `store-factory` alone drags 16 daemon session-store files, and `property-arbitraries` drags fast-check — regardless of what they used.

This PR codemods every importer (129 files) to import the backing modules directly, preserving aliases and per-specifier `type` modifiers, and deletes the barrel so the tax is unrepresentable rather than pinned by a guard test.

Measured with the eager-import-closure walker (the #1950 fixture) on identical base `320a881ae`, whole unit-core suite (971 test files):

| | aggregate eager module evaluations |
|---|---|
| before | 153,401 |
| after | **144,344 (−5.9%)** |

The local barrel at `src/commands/interaction/runtime/__tests__/test-utils/index.ts` is intentionally untouched (12 importers, local helpers only, negligible closure).

## Validation

- `pnpm check:affected --run`: all runnable local checks passed; coverage and native/device lanes remain GitHub-authoritative.
- `session-replay-runtime-maestro.test.ts`: 43/43 tests passed after moving its shared session setup into the existing sibling fixture module.
- `test-file-size-ratchet.test.ts`: 4/4 passed; the giant-file pin was lowered from 2,031 to 2,024 lines rather than raised.
- One codemod edge repaired by hand: a file already importing `session-factories.ts` directly gained a duplicate import; merged (caught by oxlint `import/no-duplicates`).

Scope: mechanical; no production code changes — test files and test utilities only.


thymikee:
[P1] Keep the oversized Maestro test from growing

This import rewrite takes `session-replay-runtime-maestro.test.ts` from its merge-base and ratchet pin of 2031 lines to 2034. The size ratchet is deliberately history-backed and cannot be fixed by raising the pin; please move a coherent existing fixture/test slice into a sibling module (or otherwise reduce this file by at least three lines) while keeping the direct import, then rerun Coverage.

thymikee:
Finished the ratchet cleanup at `52ef97c67`.

- Moved the reusable replay-session setup into the existing sibling fixture module instead of raising the giant-test allowance.
- Lowered the exact file-size pin from 2,031 to 2,024 lines.
- Replay runtime: 43/43 tests passed. File-size ratchet: 4/4 passed.
- `pnpm check:affected --run` passed every runnable local gate; expensive coverage and native/device lanes remain GitHub-authoritative.

PR #1957, opened 2026-08-22
perf(typecheck): make the root tsc project incremental

thymikee:
## Summary

The `tsc -b packages/*` half of `pnpm typecheck` already caches through composite builds, but the root `tsc -p tsconfig.json` re-checked from scratch every run. This adds `incremental` with a build-info file at `.tmp/tsconfig.tsbuildinfo` — repo-local, per-worktree, already covered by both the `.tmp/` and `*.tsbuildinfo` gitignore entries, and safe to delete at any time (performance state only, never correctness).

Measured on a warm machine: repeat `tsc -p tsconfig.json` drops **2.0s → 0.4s** wall. Cold runs unchanged. CI is unaffected (no cache persisted; cold behavior identical).

Scoped deliberately small: a second candidate from the same investigation — `NODE_COMPILE_CACHE` on the plain-node check lanes — was dropped because a controlled measurement showed no wall-time change on `check:layering` (12.5s with and without; the lane's cost is graph work, not module compilation).

## Validation

- `pnpm check:affected --run`: all runnable checks passed.
- `.tmp/tsconfig.tsbuildinfo` confirmed written and ignored by git.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -10,6 +10,8 @@
     "noUnusedLocals": true,
     "noUnusedParameters": true,
     "noEmit": true,
+    "incremental": true,
+    "tsBuildInfoFile": ".tmp/tsconfig.tsbuildinfo",
     "allowImportingTsExtensions": true,
     "erasableSyntaxOnly": true,
     "rewriteRelativeImportExtensions": true,


thymikee:
Reviewed current head f520be230. The root typecheck now retains only TypeScript’s invalidatable build-info under the already ignored, worktree-local .tmp/ directory; package and root checks remain sequential, and CI uses clean checkouts with no workspace cache. The exact merge commit’s full Typecheck job and all required checks are green. Ready for human.

PR #1969, opened 2026-08-22
perf(contracts): granularize entry surfaces so hub importers stop evaluating the facade clump

### LINKED ISSUE #1959: perf(contracts): granularize entry surfaces so hub importers stop evaluating the facade clump

thymikee:
Closes #1959.

## Problem

`@agent-device/contracts/platform` value-re-exports 32 vocabulary modules; `/interaction` another 18. A file that value-imports either evaluates the **whole union** to reach one function — and permanent hubs sat behind them (`command-descriptor/registry.ts`, `core/capabilities.ts`, `interactors/register-builtins.ts`, `command-descriptor/platform-execution-entry.ts`), so that union rode into roughly half of the unit suite's test graphs.

## Change

Every vocabulary module gets its own entry subpath, and all value-importers move onto the module that owns the symbol. Type-only importers are left alone exactly as the issue scoped it — `import type` is erased, so it already evaluated nothing.

The rewrite was mechanical: symbol → owning-module resolved from each façade's own re-export table, so no import was hand-picked. Every symbol resolved; nothing was skipped.

## Measured

Eager-import-closure walker from #1950, over all 974 unit-core test files, against base `e5bfde3d1`:

| | base | now |
|---|---:|---:|
| aggregate eager module evaluations | 143,248 | **129,738** (−9.4%) |
| `facades/platform.ts` carried by | 466 graphs | **1** |
| `facades/interaction.ts` carried by | 451 graphs | **0** |

Per hub:

| file | base | now |
|---|---:|---:|
| `command-descriptor/registry.ts` | 105 | 66 |
| `core/capabilities.ts` | 113 | 76 |
| `interactors/register-builtins.ts` | 111 | 73 |
| `command-descriptor/platform-execution-entry.ts` | 43 | 3 |
| `core/dispatch.ts` | 134 | 100 |

The issue projected ~10–14%. The gap is the part the leaf-simulation could not model: consumers still evaluate the modules they genuinely use, so the recovered work is only what they *didn't*.

Two hubs (`capabilities.ts`, `register-builtins.ts`) did not move on the first pass — `platforms/apple/plugin.ts` pulled the clump back in behind them. That is why the change covers every value-importer rather than the four files the issue named.

## Regression pin

`src/__tests__/contracts-entry-closure.test.ts`, built on the #1950 walker, from both directions:
1. the four named hubs never evaluate either façade;
2. **no source file anywhere** value-imports them — so the clump cannot re-form behind a hub nobody thought to list. Type-only imports stay legal, which the walker distinguishes for free.

A non-vacuity assertion requires the scan still sees the narrow imports that replaced these, so a broken walker fails loudly instead of passing empty.

**Planted red** (reverted before commit): restoring `capabilities.ts` to `contracts/platform` failed both tests, each naming the offending edge —
```
src/core/capabilities.ts -> packages/contracts/src/facades/platform.ts
src/core/capabilities.ts -> @agent-device/contracts/platform
```

## Gate adjustments the split forces

- **R11** — the pinned contracts subpath list grows to the new entries. The resolver test's "must not resolve" example moves to `./clipboard`, since `./gesture-plan` is now a real entry.
- **R16** — anchored the record-runtime join on the literal `contracts/platform` specifier; it now accepts any contracts entry. The assertion's *provenance* is what the rule pins, not which subpath carried it.
- **no-bare-star** — `gesture-plan.ts` became an entry target, and the rule rejects the `export * from './gesture-plan-types.ts'` it carried. Its one internal consumer (`client-gesture.ts`) now imports the owning module directly.

## Known follow-up

Both façades keep their type re-exports for the ~490 type-only importers, so every symbol on them now reads as value-unused. One `.fallowrc.json` entry records that with the reasoning. Net `check:production-exports` findings go **37 → 29** (below base, because the entry also absorbs pre-existing façade findings).

Retiring the two façades outright — moving the type-only importers too and deleting the files — is the change that makes that entry unnecessary. It is ~490 further one-line edits with no additional perf gain (type imports are already free), so it is deliberately not bundled here.

Worth tightening after #1965 lands: its budget table seeds façade budgets from pre-#1959 closures, which this PR makes much smaller.

## Verification

`pnpm typecheck`, `pnpm lint`, `pnpm format`, `pnpm check:layering` (181/181), `pnpm check:fallow` (no issues in 258 changed files), `pnpm gate production-exports`, and `pnpm test:unit` — **1002 files / 7473 tests, all green**.

One earlier run showed a single `runner-client.test.ts` timeout; it passes in isolation and is the known under-contention signature (timeouts, never assertion failures) from four agents sharing this host. Clean on the full re-run.


thymikee:
Reviewed exact head `391f2990`: no actionable code findings. The 259-file scope is a mechanical, complete value-import migration; the narrower hub-only edit was correctly rejected because `platforms/apple/plugin.ts` pulled the façade clump back behind the named hubs, while migrating ~490 erased type-only imports would add churn with no evaluation gain. The regression scan is planted-red and pins the owning invariant. Exact-head checks are green, size is below escalation threshold (npm unpacked -6.1 kB; JS gzip +2.8 kB), and the measured closure reduction matches #1959. Ready for human merge.

PR #1972, opened 2026-08-22
feat(android): bound snapshot presentation quality

thymikee:
## Summary

Bound Android snapshot presentation with one typed work/time budget, including scope selection and reindexing, and return a typed `presentation-failed` sparse result when the budget is exhausted.

Scope discovery uses one document-order pass plus bottom-up subtree collection. Coverage allocation, candidate scans, edge indexing, scope traversal, and reindexing all charge the same owner budget. Each node also retains its owning window rectangle so disjoint system and application windows are validated independently.

This completes #1832 C5 and C6. C4/C1 ownership remains in #1968; drawing-order pruning is unchanged.

The PR is 21 files, +1,022/-110 against #1968. The npm growth is the typed presentation/quality safety contract and regression coverage. Smaller message-sniffing, backend branching, unmetered coverage, per-candidate subtree rescans, or a single global window clip were rejected because they weaken the contract or reject valid multi-window captures.

## Validation

At `28f7d724f`:

- planted-red multi-window regression reproduced `Android regular snapshot node 0 escaped its cumulative clip`, then passed with the owning-window carrier
- focused Android presentation and quality tests: 13/13 passed
- `pnpm check:affected --run`: 632 files / 5,029 tests; all runnable checks passed
- live Pixel 7 / API 36: the exact CI fixture reproduced the alert failure; the raw hierarchy contained the alert while regular presentation returned zero nodes
- after the fix, the same live alert produced a healthy four-node interactive snapshot and the complete Android smoke scenario passed in 83 seconds

Fresh GitHub Android Smoke remains authoritative for the exact pushed head.

PR #1973, opened 2026-08-22
test(snapshot): add cross-runtime presentation conformance

thymikee:
## Summary

Add one versioned acquisition-to-presentation fixture with four geometry cases, consumed by the real Swift presentation seam and Android XML parser plus presentation seam.

Preserve acquired iOS non-hittability while applying cumulative clips and viewport geometry. Part of #1832.

The incremental stack diff is 5 files (+476/-9), mostly fixture and tests. Duplicating native expectations was rejected because they could drift; a cross-language production abstraction was rejected because the runtimes do not share an implementation boundary.

## Validation

At `30b9339d5`, restacked onto #1972 `28f7d724f`:

- Android presentation and cross-runtime conformance: 11/11 passed
- `pnpm check:xctest-selection`: 204 declared; 0 unreachable
- `pnpm build:xcuitest:macos`: `TEST BUILD SUCCEEDED`
- `pnpm check:affected --run`: 633 files / 5,030 tests; all runnable checks passed

Focused host XCTest execution was locally blocked before test launch while LaunchServices workers materialized. GitHub's Swift Runner Host XCTests are authoritative for the exact pushed head.

### DIFF
--- a/.github/workflows/ios.yml
+++ b/.github/workflows/ios.yml
@@ -178,7 +178,7 @@ jobs:
             -only-testing:AgentDeviceRunnerUITests/RunnerTests/testEffectiveGeometryIntersectsViewportAndAncestorClip \
             -only-testing:AgentDeviceRunnerUITests/RunnerTests/testEffectiveGeometryKeepsReportedOriginWhenFullyClipped \
             -only-testing:AgentDeviceRunnerUITests/RunnerTests/testRegularPresentationRoutesThroughVisibilityFold \
-            -only-testing:AgentDeviceRunnerUITests/RunnerTests/testRegularPresentationPublishesGeometricActionabilityWithoutOcclusionOrTypeGate \
+            -only-testing:AgentDeviceRunnerUITests/RunnerTests/testRegularPresentationPreservesAcquiredActionabilityWhileApplyingGeometry \
             -only-testing:AgentDeviceRunnerUITests/RunnerTests/testSnapshotAcquisitionDoesNotReintroduceRunnerOcclusionScan \
             -only-testing:AgentDeviceRunnerUITests/RunnerTests/testSnapshotScopePolicyMatchesGoldenParityTable \
             -only-testing:AgentDeviceRunnerUITests/RunnerTests/testScrollContainerTypeNamesMatchElementTypeSet \
--- a/.github/workflows/ios.yml
+++ b/.github/workflows/ios.yml
@@ -178,7 +178,7 @@ jobs:
             -only-testing:AgentDeviceRunnerUITests/RunnerTests/testEffectiveGeometryIntersectsViewportAndAncestorClip \
             -only-testing:AgentDeviceRunnerUITests/RunnerTests/testEffectiveGeometryKeepsReportedOriginWhenFullyClipped \
             -only-testing:AgentDeviceRunnerUITests/RunnerTests/testRegularPresentationRoutesThroughVisibilityFold \
-            -only-testing:AgentDeviceRunnerUITests/RunnerTests/testRegularPresentationPreservesAcquiredActionabilityWhileApplyingGeometry \
+            -only-testing:AgentDeviceRunnerUITests/RunnerTests/testRegularPresentationPublishesGeometricActionabilityWithoutOcclusionOrTypeGate \
             -only-testing:AgentDeviceRunnerUITests/RunnerTests/testSnapshotAcquisitionDoesNotReintroduceRunnerOcclusionScan \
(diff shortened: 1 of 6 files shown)


thymikee:
Not ready at `5a619b515`. The new conformance gate is doing useful work and exposes a real owner-contract drift: Swift's `SnapshotVisibilityFold` recomputes positive non-root nodes as hittable, so the shared cases emit `true` for the acquired-non-hittable `Outer` and scoped `Settings` ScrollViews, while Android preserves false. Per the presentation contract, fix the owning Swift fold so presentation only narrows acquired actionability (or explicitly revise the shared policy); don't paper over this by changing one runtime's expected values. Separately, the branch is DIRTY/conflicting because #1972's base was rewritten from `34bed0fa` to `b9b997a5`; rebase the single conformance commit. That rebase should remove the inherited file-size Coverage failure, but the Swift host red is PR-owned and must be fixed.

thymikee:
Addressed on `c8e47181b` (rebased onto `6943369b9` / #1972).

- Rebased only the conformance commit; stale inherited #1972 history is gone.
- Swift `SnapshotVisibilityFold` now preserves acquired `node.hittable` and only narrows it with enabled/effective geometry, so it cannot upgrade raw false.
- Updated the existing unit assertion to prove raw-false covered/labeled nodes stay false, raw-true positive overlay stays true, and disabled stays false.
- Android shared conformance remains green; a fresh unit-test-enabled macOS runner build is green; xctest selection is green.
- The final affected gate passed: 637 files and 5,044 tests.
- Host XCTest remains locally blocked by LaunchServices worker materialization after 111s, so that exact CI lane remains authoritative. No 33-screen corpus is claimed.

thymikee:
The one-line Swift actionability fix is correct and its focused assertion is load-bearing, but three blockers remain. (1) Replace `Closes #1832 Step 9` with `Part of #1832`/`advances Step 9`: GitHub will close the whole issue even though C4/C5/C6 and Step 9’s API-level differential legs remain. (2) The shared “cumulative scroll clip” fixture has only one scroll ancestor, so it proves direct clipping—not cumulative nested clipping. Add outer+inner scroll containers whose intersection is narrower than either individual clip and require both Swift and Android presentation to emit that effective rect. (3) Exact rewritten heads for #1973 and base #1972 currently have only CodeQL; obtain authoritative Swift Runner Host XCTests/full CI after the base stabilizes. Also describe the Android leg accurately as XML parser + presentation conformance; it reconstructs hierarchy from depth/order rather than consuming the same RawAXNode seam.

thymikee:
Addressed on c772c16393bfc24c4f3fcdc70d05e510ea1fcdae.

- the issue reference is now Part of #1832
- the cumulative-clip vector now has outer and inner scroll ancestors whose intersection is narrower than either alone
- Android wording now reflects XML parsing plus presentation projection
- exact-head affected gate passed (633 files / 5,028 tests), including XCTest lane selection
- exact-head live Android helper validation is healthy; host/native XCTest remains GitHub-authoritative

The published branch is mergeable; required CI is running.

thymikee:
Exact-head Android Smoke inherits #1972's owner-action regression; it is not infrastructure and not caused by #1973's incremental conformance/Swift diff. Separate #1972/#1973 jobs both successfully click the native-alert control, then time out waiting for `Automation confirmation` after 20/18 readable helper captures. The PR body's healthy two-window Settings snapshot does not exercise this alert route. Diagnose/fix #1972 with an immediate post-click tree/image showing whether the click lands and whether the dialog survives multi-window presentation; obtain green Android Smoke on #1972, then restack/rerun #1973. Every other #1973 lane, including Host XCTest and iOS smoke, is green; do not label while Android is red.

thymikee:
Restacked onto the repaired #1972 head and re-reviewed as an incremental slice. The shared fixture now exercises four geometry cases through Swift presentation and Android XML parsing plus presentation; the only Swift production change preserves acquired non-hittability while applying geometry. Validation: Android conformance 1/1, XCTest selection 204 declared/0 unreachable, macOS XCTest build succeeded, and affected gate 633 files/5,029 tests passed. Fresh exact-head GitHub native/device lanes remain authoritative before readiness.

thymikee:
Final restack is now on the live-proven #1972 head (`28f7d724f`). The incremental conformance slice remains unchanged in intent: one fixture, Swift presentation, Android parser plus presentation. Focused Android tests 11/11, XCTest selection 204 declared/0 unreachable, macOS XCTest build succeeded, and affected gate 633 files/5,030 tests passed. Fresh exact-head native/device checks are pending before readiness.

thymikee:
Readiness re-check on `30b9339d5`: the reviewed conformance slice is clean, its focused/full validation passed, and all 28 required checks now pass. Ready for human review.

PR #1974, opened 2026-08-22
perf(android): keep the automation helper warm across fill and scroll

thymikee:
## Summary

Keep the Android automation helper session warm across fill verification and scroll viewport reads. The daemon session owns one helper scope and one capture-options construction path; each capture still performs a fresh hierarchy traversal.

Teardown skips `am force-stop` only after a graceful release with a trustworthy shell-v2 device exit status. Unknown or older transports, aborts, timeouts, failed starts, and content-failure retirement remain fail-closed and reset the runtime.

Live emulator evidence reduced fill from 2,910 ms to 1,673 ms while helper starts fell from 3 to 0; scroll helper starts fell from 2 to 0.

## Growth record

The branch is +2,921/−1,696 overall, but only +1,105/−772 is production code: +333 net.

- Warm-session behavior and retirement repair: +291/−84 production lines (+207 net).
- Extraction moved roughly 768 production lines into the lifecycle and text-input owners; excluding the shell-v2 proof, that refactor is +835/−768, about 67 net lines of owner/seam overhead.
- Trustworthy adb shell-v2 capability proof: 59 new production lines.
- Tests, fixtures, and the file-size ratchet: +1,816/−924 (+892 net). The bulk is split/relocated coverage for the new lifecycle and text-input owners; the rest is red-first safety and spawn-count evidence.
- Published npm impact is +2.2 kB unpacked (+652 B tarball).

Smaller alternatives were rejected because they weaken the invariant or erase the performance win: keeping one-shot sessions retains the instrumentation startup cost; deriving scopes independently in fill and scroll creates competing session identities; caching hierarchies risks stale UI; trusting adb exit 0 without shell v2 can leak UiAutomation ownership; and leaving the behavior in the existing 500+ line aggregations violates the repository extraction rule.

## Validation

- Seven regressions were observed red against the merge base; the content-failure retirement mutation was also observed and repaired.
- Live Pixel emulator A/B proved the `android-helper` backend and counted helper starts device-side.
- `pnpm check:affected --run`: 272 related files / 1,821 tests passed at the pushed head.
- All GitHub checks are green, including Android/iOS/macOS/Linux smoke, coverage, integration, CodeQL, and package-size reporting.

Residual risk: API < 34 cache flushing re-applies `setServiceInfo()` because `UiAutomation.clearCache()` is unavailable. The session is shared, never the captured hierarchy.

PR #1979, opened 2026-08-23
fix(macos): sign nested runner code so the host runner app can launch

thymikee:
## Summary

Repair unsigned macOS runner products by signing their embedded Apple test-support code explicitly, then signing the parent app.

The repair preserves identifiers, entitlements, flags, and hardened-runtime metadata while allowing codesign to regenerate designated requirements for the ad-hoc identity. It avoids deprecated signing-time `--deep`; deep traversal remains verification-only.

Post-repair strict verification is retained so an invalid cached or freshly built product cannot be reported as repaired.

## Validation

- Observed the new ordering/metadata regression test fail against the previous `--force --deep` implementation, then pass after the repair.
- `pnpm check:affected --run`: format, lint, typecheck, layering, fallow, build, and 2,317/2,318 related tests passed; the sole timeout passed in isolation in 126 ms and is unrelated to this two-file change.
- Fresh `pnpm build:xcuitest:macos` product passed `codesign --verify --deep --strict`, including its nested `.xctest` bundle and all embedded test-support items.
- Real CLI open launched TextEdit and reached the macOS XCTest runner. Interactive snapshot then hit the existing host Automation-permission timeout; the isolated session closed cleanly.

### DIFF
--- a/src/platforms/apple/core/__tests__/runner-macos-products.test.ts
+++ b/src/platforms/apple/core/__tests__/runner-macos-products.test.ts
@@ -1,0 +1,88 @@
+import assert from 'node:assert/strict';
+import fs from 'node:fs';
+import os from 'node:os';
+import path from 'node:path';
+
+import { afterEach, test } from 'vitest';
+
+import { MACOS_DEVICE } from '../../../../__tests__/test-utils/device-fixtures.ts';
+import { AppError } from '@agent-device/kernel/errors';
+import {
+  isExpectedRunnerRepairFailure,
+  repairMacOsRunnerProductsIfNeeded,
+} from '../runner/runner-macos-products.ts';
+import { createLocalAppleToolProvider, withAppleToolProvider } from '../tool-provider.ts';
+
+const XCTESTRUN_PATH = '/tmp/agent-device-runner.xctestrun';
+const tempRoots: string[] = [];
+
+afterEach(() => {
+  while (tempRoots.length > 0) {
+    fs.rmSync(tempRoots.pop() as string, { recursive: true, force: true });
+  }
+});
+
+test('repair re-signs nested code so the product passes deep verification', async () => {
+  const productPath = createProductPath();
+  const calls: string[][] = [];
+  // Nested Apple test frameworks lose their sealed Modules/ entries during the unsigned embed,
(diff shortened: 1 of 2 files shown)


thymikee:
Blocking: do not sign with `codesign --force --deep`. Apple deprecates signing `--deep` and documents that this combination forcibly replaces every nested signature; this target also declares macOS network client/server entitlements, so the PR's no-entitlements safety claim is not an invariant. Repair the known Testing, XCTest, and XCUIAutomation frameworks explicitly bottom-up, then sign the parent product; retain `--deep --strict` only for verification. Pin the exact signing order and metadata policy in the test. Post-repair verification and the real macOS replay evidence otherwise look good.

thymikee:
Addressed the signing-safety finding in 463643fa0: embedded Apple test-support items are now signed explicitly bottom-up, the parent app is signed last, and signing-time `--deep` is gone. Identifiers, entitlements, flags, and runtime metadata are preserved; designated requirements are intentionally regenerated for the ad-hoc identity. A fresh macOS product and its nested xctest now pass `codesign --verify --deep --strict`, and a real CLI open launches the repaired runner. The focused regression test was observed red before the change.

thymikee:
Follow-up `a2963e20f` removes the hard-coded XCTest framework inventory. Repair now signs the deterministic contents of the product’s `Contents/Frameworks` directory before the parent, so Xcode’s actual output is authoritative. The arbitrary-name regression test was observed failing against the previous list and passing after discovery. `pnpm check:affected --run` passed all runnable gates (2,318 tests), and a fresh Xcode 26.2 macOS runner repaired through the production function passed `codesign --verify --deep --strict`.

thymikee:
CI diagnosis: Typecheck and FreeRange both fail on the same deterministic Android error from current `main`, not this PR: `snapshot-quality.test.ts` still imports `resetAndroidSnapshotHelperSessions` from `snapshot-helper-session.ts`, while #1974 moved that API to `snapshot-helper-session-lifecycle.ts`. The macOS Smoke and Swift Runner Host XCTests pass on this head.

Focused main hotfix: #1980. Once that lands, re-running/rebasing this PR should clear these unrelated gates.

PR #1982, opened 2026-08-23
perf(android): bound occlusion coverage memory

thymikee:
## Summary

Android snapshot occlusion now measures rectangle-union coverage with an event-driven sweep instead of allocating a Cartesian X/Y cell grid.

This keeps exact union semantics, drawing-order policy, the 90% coverage threshold, deterministic work budgets, and cooperative cancellation while allowing broad valid footprints to complete within their existing caps. The geometry now lives in one Android-owned module with oracle-backed tests.

The separate quadratic sibling-candidate scan remains out of scope. No snapshot schema, CLI surface, docs, or skills changed.

## Validation

Before the fix, planted regression runs exhausted the original 1,024- and 1,650-work-unit caps. Both cases pass at those same caps after the change.

`pnpm check:affected --run` passed: 273 test files and 1,845 tests, including provider integration, format, lint, typecheck, layering, Fallow, and build.

Live Android emulator verification at `b82669f8` captured Settings through `android-helper` 0.20.10 with healthy snapshot quality: 118 acquired nodes, 34 presented nodes, and 312 ms helper time. The verification session was closed.

### DIFF
--- a/src/platforms/android/__tests__/rect-coverage.test.ts
+++ b/src/platforms/android/__tests__/rect-coverage.test.ts
@@ -1,0 +1,128 @@
+import type { Rect } from '@agent-device/kernel/snapshot';
+import fc from 'fast-check';
+import { expect, test } from 'vitest';
+import { PROPERTY_RUNS } from '../../../__tests__/test-utils/property-arbitraries.ts';
+import { unionCoverage } from '../rect-coverage.ts';
+import { createAndroidSnapshotPresentationBudget } from '../snapshot-presentation.ts';
+
+const smallRect = fc.record({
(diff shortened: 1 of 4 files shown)


thymikee:
Clean and merge-ready at `b82669f8`. The event-driven X sweep preserves exact union/intersection semantics and the existing `>= 0.9` publication rule while replacing the Cartesian allocation. Focused tests and exact-head live Android evidence are valid; all 27 checks, including Android Smoke, are green, and size growth is only +494 B. Non-blocking residual: active Y intervals are re-sorted per slab, so adversarial staggered inputs are bounded by the cooperative work budget rather than formally linear time.

PR #1985, opened 2026-08-23
refactor(recording): move gesture telemetry out of the daemon zone

thymikee:
## Summary

No behavior change. `recording-telemetry.ts` writes the gesture-telemetry sidecar next to a
recording; it imports only `node:fs`, `node:path` and `@agent-device/contracts/platform`, and has no
daemon dependency at all. Living under `src/daemon/` made the platform-runtime composition root
(`src/platform-runtime-screen-recording-finalizer-host.ts`) reach up into `daemon-server` (rank 4)
to persist telemetry.

It moves to `src/recording/telemetry.ts`, next to `output-path.ts` and `overlay.ts` — siblings with
the same shape (node builtins plus contracts/kernel/utils). Named `telemetry.ts` rather than
`recording-telemetry.ts` to match those siblings, which do not restate the directory.

The module body is byte-identical; the diff is the rename plus three retargeted import sites
(`platform-runtime-screen-recording-finalizer-host.ts`, its test's `vi.mock` path, and
`daemon/handlers/record-runtime-response.ts`).

**Effect on the spine.** The composition root had four upward value edges. This removes the only one
reaching rank 4:

| From | To | Zone | After |
| --- | --- | --- | --- |
| `platform-runtime-screen-recording-finalizer-host.ts` | `daemon/recording-telemetry.ts` | daemon-server (4) | **removed** |
| `platform-runtime-android-emulator-host.ts` | `core/device-inventory-context.ts` | core (2) | unchanged |
| `provider-limrun-runtime.ts` | `sdk/limrun-runtime-dependencies.ts` | sdk (4) | unchanged — documented in-file as a deliberate static edge |
| `runtime.ts` | `commands/index.ts` | commands (3) | unchanged — `bindCommands` is the aggregate binder |

The two remaining candidates were left alone deliberately: the limrun one carries an explicit comment
marking it the deliberate seam ("making it lazy would require a proxy interactor"), and `bindCommands`
aggregating every command is the purpose of `runtime.ts`. Only the emulator-host edge is a real
open question, and inverting it changes a host-factory signature — out of scope here.

This is a small collocation fix, not a step in #1739: `daemon → platform` edges are untouched (still
61), and the daemon file count drops by one.

## Validation

`pnpm check:affected --run` passed on this head — 154 test files, 835 tests, "all runnable checks
passed", including both tests that cover the changed imports
(`platform-runtime-screen-recording-finalizer-host.test.ts` and
`handlers/__tests__/record-runtime-response.test.ts`). `pnpm typecheck` and `pnpm check:layering`
also run clean; the layering summary is unchanged from `main` (R6 still 7, R11 still 17 packages /
95 subpaths / zero root back-imports).

The spine claim above is from `pnpm depgraph` before and after: composition-root upward value edges
4 → 3 with the `daemon-server` target gone, and value-import cycles (0), spine back-edges (0) and
type-only inversions (7) all unchanged.

No device-facing path is touched — the telemetry writer's behavior, output path and envelope are
untouched — so this owes no live device evidence. Published and reported; CI on this head is the
authority still to come.

---
_Generated by [Claude Code](https://claude.ai/code/session_01WWLBCBDBpdR8z1aCmDerXS)_


thymikee:
`Smoke Tests` (iOS, run 32655577032) went red at step **"Run targeted iOS runner XCTest regressions"** with exactly one failing test:

```
Failing tests:
	RunnerTests.testBareTypeUsesTappedInputWhenSoftwareKeyboardIsHidden()
** TEST EXECUTE FAILED **
```

That is #1874 — "testBareTypeUsesTappedInputWhenSoftwareKeyboardIsHidden fails intermittently on unrelated branches (blocking ios.yml PR gate)" — still open, with `diagnose(1874)` (#1941) on `main` as unresolved diagnostic work.

Not fixing it here, and not touching the test. This PR renames one TypeScript module (`src/daemon/recording-telemetry.ts` → `src/recording/telemetry.ts`) and retargets three imports. It changes nothing under `apple/runner/**`, so it cannot affect a Swift XCUITest's behavior — and `Swift Runner Host XCTests`, which compiles and runs the same bundle on the macOS host lane, passed on this same head (`a2f096a`). Every other check on this head is green, including Layering Guard, Typecheck, Integration Tests and Packaged CLI.

The base branch is not red: `main` at this PR's base (`d1fc80f`) had a green iOS run at 16:08Z, so this is the intermittent failure mode #1874 describes rather than a base outage.

I have re-run the failed job once to confirm. If it comes back red on the same test I will report that here rather than push at it — under #1874 that is a gate problem, not a defect in this diff.

---

thymikee:
Clean and merge-ready at `a2f096a4`: this behavior-identical rename removes the composition root's only daemon-server telemetry import while preserving `record → recordTrace → bound screen-recording runtime → lazy finalizer` and ADR 0003 daemon-registry ownership. It introduces no command/platform facet or Apple-family leakage; typed artifact serialization now sits with its recording owners. Import consumers cover the move, package size is -20 B, and all 28 exact-head checks—including every smoke lane—are green. No live device evidence is owed for this internal module move.

PR #1988, opened 2026-08-24
perf(selectors): resolve a recorded replay target in one matching pass

### LINKED ISSUE #1978: selectors: land resolveSelectorChainDomain + replay single-pass target resolution (split from #1690)
Split out of #1971 per review: #1690's scope and STOP conditions exclude selector-resolution and replay changes, but the find-index implementation produced a sound, independently valuable optimization there.

**Ready on branch `tier3/selector-replay-domain`** (commits from #1971's review cycle, adversarially reviewed):

- New `resolveSelectorChainDomain` in `packages/selectors/src/internal/resolve.ts`: returns the matched-node set the deciding pass already collected — the winning alternative's when one resolves, else the first alternative with any match (exactly `listSelectorChainMatches`' selection rule, verified over a 112-combination semantic-parity fuzz vs merge-base including winner-in-later-alternative, empty trees, rectless nodes, requireRect/allowDisambiguation toggles).
- `resolveRecordedTarget` (`internal/replay.ts`) drops its redundant second full-tree scan on both legs: resolved leg {iterated 2, filter 1→0}; unresolved leg {filter 1→0}.
- `observeTreeTraversals` test helper widened to count the full array surface (flatMap/forEach/reduce/some) — the previous counter was blind to flatMap, hiding 4n uncounted scans per ambiguous candidate.
- `listSelectorChainMatches` keeps its three other callers; published ./ast surface unchanged.

Red-first evidence and scan-count tables are in #1971's REVIEW.md/RESULT.md history.

**Known adjacent defect, intentionally out of scope here:** `resolve-with-policy.ts:66-89` still double-scans for disambiguate/fail-closed rows (#1970) — folding it needs the first-alternative-vs-winner semantic decision documented there.

thymikee:
Closes #1978. Split from #1971 (find index-once) per review — #1690's scope excludes selector/replay, so this lands separately on top of the now-merged find topology work.

## Summary

- New `resolveSelectorChainDomain` (`packages/selectors/src/internal/resolve.ts`): returns the matched-node set the deciding pass already collected — the winning alternative's when one resolves, else the first alternative with any match (exactly `listSelectorChainMatches`' selection rule).
- `resolveRecordedTarget` drops its redundant second full-tree scan on both legs: resolved leg {iterated: 2, filter: 1→0}; unresolved leg {filter: 1→0}.
- `observeTreeTraversals` widened to count the full array surface (flatMap/forEach/reduce/some) — the previous counter was blind to flatMap, which hid 4n uncounted scans per ambiguous candidate.
- Published ./ast surface unchanged; `listSelectorChainMatches` keeps its three other callers.

## Validation

- Red-first scan-count tables in-thread; semantic parity vs merge-base verified over a 112-combination fuzz (winner-in-later-alternative, empty trees, rectless nodes, requireRect/allowDisambiguation toggles) during the #1971 adversarial review — PASS.
- 8 selectors tests green at pushed head; rebased onto current main.

### DIFF
--- a/packages/selectors/src/index.test.ts
+++ b/packages/selectors/src/index.test.ts
@@ -136,22 +136,33 @@ test('recorded-target resolution keeps the matched domain from the alternative t
  * Functions are bound to the target so a counted traversal cannot re-enter the
  * proxy.
  */
-const SCAN_METHODS = ['filter', 'map', 'flatMap', 'forEach', 'reduce', 'some'] as const;
-type ScanPasses = { iterated: number } & Record<(typeof SCAN_METHODS)[number], number>;
+const SCAN_METHODS = [
+  'filter',
+  'map',
+  'flatMap',
+  'forEach',
+  'reduce',
+  'reduceRight',
+  'some',
+  'every',
+  'find',
+  'findIndex',
+  'findLast',
+  'findLastIndex',
+  'includes',
+  'indexOf',
+] as const;
+type ScanMethod = (typeof SCAN_METHODS)[number];
+type ScanPasses = { iterated: number } & Record<ScanMethod, number>;
 
 function observeTreeTraversals(nodes: SnapshotNode[]): {
   observed: SnapshotNode[];
   passes: ScanPasses;
 } {
-  const passes: ScanPasses = {
-    iterated: 0,
-    filter: 0,
-    map: 0,
-    flatMap: 0,
-    forEach: 0,
-    reduce: 0,
-    some: 0,
-  };
+  const passes = Object.fromEntries([
+    ['iterated', 0],
+    ...SCAN_METHODS.map((method) => [method, 0]),
+  ]) as ScanPasses;
   const observed = new Proxy(nodes, {
     get(target, property) {
       if (property === Symbol.iterator) passes.iterated += 1;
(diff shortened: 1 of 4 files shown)


thymikee:
Exact head `6a9678c2` is not ready despite green CI. The production one-pass implementation and replay route look semantically sound, but the regression oracle is incomplete: `observeTreeTraversals` says it covers the whole array-scan surface while tracking only `filter/map/flatMap/forEach/reduce/some` plus iteration. Reintroducing a second full pass with `find`, `findIndex`, `findLast`, `findLastIndex`, or `every` leaves all asserted counters unchanged. Expand the instrumentation (or count indexed/tree reads) so an ordinary alternative implementation is planted-red and the claimed no-second-pass invariant is actually enforced. The rest of the selector-domain semantics, internal surface, scope, checks, and size impact review clean.

thymikee:
Addressed at 9096e7143: `SCAN_METHODS` now covers the full scan surface — added `reduceRight`, `every`, `find`, `findIndex`, `findLast`, `findLastIndex`, `includes`, `indexOf`. Production uses none of them on this path (all zero in both legs' assertions).

Planted red demonstrated: injecting a single `nodes.find(() => true)` into `resolveRecordedTarget` fails the gate with `find: 1`; reverted, green again (`9096e7143`). Same head note: the earlier head was indeed unchanged — this commit moves it.

Issue #1976, opened 2026-08-22
ci(static-checks): both rg assertions are false-green — ripgrep is never installed


thymikee:
The `Static Checks` job runs two `rg` assertions on `ubuntu-latest`, but nothing installs ripgrep and the GitHub-hosted Ubuntu image does not ship it. Both steps fail with `rg: command not found`, and because each assertion is written as `if rg ...; then ... exit 1; fi`, a missing binary makes the `if` condition false — so the step exits 0 and the job reports green.

Both assertions have therefore been vacuous since they were introduced, and one of them is now hiding real violations.

### Evidence

From the `Static Checks` job on a recent PR run ([job 97029728691](https://github.com/callstack/agent-device/actions/runs/32572417812/job/97029728691), 8s, reported **pass**):

```
Disallow trailing commas before closing parenthesis in Swift
  /home/runner/work/_temp/....sh: line 1: rg: command not found

Fail if test-only DI seams reappear in production code
  /home/runner/work/_temp/....sh: line 1: rg: command not found
```

Runner image: `ubuntu-24.04` / `20260816.277.1`.

Running the same two assertions locally **with** ripgrep installed:

- `rg -nU --glob '*.swift' ',\s*\n\s*\)' apple/runner` → exit 1, no matches. This one would legitimately pass.
- `rg '\?\s*:\s*typeof\s+' src/ --glob '!**/__tests__/**' --glob '!*.test.ts'` → exit 0, **6 matches**:

```
src/remote/daemon-proxy.ts:  fetchImpl?: typeof fetch;
src/cli/auth-session.ts:  fetch?: typeof fetch;
src/cli/auth-session.ts:  fetchImpl?: typeof fetch;
src/cli/auth-session.ts:  fetchImpl?: typeof fetch;
src/cli/connection/cloud-profile.ts:  fetchImpl?: typeof fetch;
src/cli/connection/cloud-profile.ts:  fetchImpl?: typeof fetch;
```

So the gate that exists to stop test-only DI seams from reappearing in production code would be **red today** if it could run at all.

### Regression window

Never worked. The Swift assertion arrived in b79bd8601 (#186, 2026-03-06) already written this way with no ripgrep install; #1465 (255deb6c2, 2026-07-28) folded it and the DI-seams assertion into the shared `static-checks` job, preserving the same shape. No revision of either job has ever installed `rg`.

### Fix sketch

Two things are wrong and both are worth fixing:

[...]

PR #1990, opened 2026-08-24
refactor(lint): retire the facade closure test for a budget row and a lint rule

thymikee:
Stacked on #1965 — merge that first. (Also blocked on #1965's open tracked-manifest discovery finding.)

Retires `contracts-entry-closure.test.ts` (added in #1959). It carried two assertions. **Neither replacement subsumes it alone — together they do**, on two different axes.

## 1. Its whole-tree scan → a lint rule

It parsed ~490 candidate files to prove nothing value-imports the two wide façades. `eslint/no-restricted-imports` states exactly that, and `allowTypeImports` draws the one distinction that made a custom walker look necessary: `import type` is erased, so it stays legal.

The rule is already in this repo's toolchain — oxlint 1.79, and the same rule already bans `node:child_process` outside `exec.ts`. It improves on the deleted test in two ways: it covers **every** file rather than the four hubs the test named, and it reaches **dynamic** imports too. It also moves the feedback from a CI run into the editor.

## 2. Its hub pin → the lint rule, plus #1960's rows on a second axis

The hub pin checked that four named hubs never reach either façade. The lint rule forbids the import edge that was the only way that could happen. #1960's rows then add an axis the old test never had: closure **size** drift.

**Correction from the previous revision of this description**, which claimed the budget was "strictly stronger" and caught "any closure change". That is wrong, and thymikee is right to reject it: a cardinality budget cannot see an **equal-size graph substitution** — swap one module for another of the same closure size and the count is unchanged, so the row passes. Forbidden edges and size drift are different properties. Each gate owns one; the pair covers what the deleted test covered.

The hub list is deliberately **not** restored — the lint rule already covers those four files along with everything else.

## Why this needed per-zone proof

Override semantics are **not additive** — a same-rule override *replaces* the parent. Two traps followed:

- a top-level rule would have been silently dropped for `src/**`, which already has its own `no-restricted-imports` override;
- the existing blanket `"off"` for `exec.ts` and the test tree would have exempted the files carrying most of the cost #1959 removed.

So the paths are added per zone, and that `"off"` becomes a façade-only config that keeps the `node:child_process` exemption it existed for.

**Planted red in all three zones** — `src/core/capabilities.ts`, a `src/__tests__` file, and `packages/capture-kit/src` — each flagged, while a type-only import in the same probe file was not. The first packages probe read as a *pass* because the sed that built it produced a type-only import; that zone was re-probed with a real value import rather than trusting the green.

Relying on config is safe here because misconfiguration fails loudly: a typo'd rule name makes oxlint exit 1 with `Rule 'x' not found in plugin`, rather than pass silently the way the `rg` assertions in #1976 did.

## What a linter cannot replace

The eager-closure budgets stay. They express a **transitive weight** property: a module already imported grows an import, and the cost arrives without any single file's import list changing. Per-file rules cannot see that, and `no-restricted-imports` can only ban specifiers named in advance — precisely what #1950, #1956 and #1959 could not have named ahead of time.

## Verification

`pnpm lint` clean, `pnpm typecheck` clean, budget gate green (207 tests). Net −98 lines of test, +59 lines of config. One iOS `Smoke Tests` job failed on a live-simulator `wait timed out for text` and passed unchanged on re-run — this branch changes no production code.


thymikee:
Small correction before approval: deleting the four-hub reachability test is safe, but the new comment/PR wording must not call the exact-size budget `strictly stronger` or say it fails on `any closure change`. The budget pins only cardinality (`actual === budget`), so an equally sized graph substitution can newly reach a wide facade without moving the count. What actually subsumes the deleted case is the combination: exact budgets catch closure-size drift for the named hubs, while the retained whole-tree rule catches forbidden facade value imports anywhere (with its non-vacuity guard). Rewrite the ownership claim accordingly; do not restore the redundant hub list. #1990 is otherwise clean/green, but remains stacked on #1965, which currently has a separate review finding.

thymikee:
Exact head d03db8d is code-clean but not ready to advance. The deletion is sound: the lint rule now owns forbidden direct value imports across src and package sources (including dynamic imports, while preserving type-only imports), and #1965's exact rows own transitive closure-size drift for the four named hubs. However, the PR body and commit message still incorrectly call the cardinality budget strictly stronger and say it catches any closure change; an equal-size graph substitution disproves that. Rewrite the claim to say the lint rule and budgets jointly subsume the old test—do not restore the hub list. The PR also remains blocked on #1965's unresolved tracked-manifest discovery finding, and one iOS Smoke job is still running.

thymikee:
Code delta is clean at `c007973`: the description now correctly separates the lint rule’s forbidden-edge property from the budgets’ closure-size property, and the rule covers all effective config zones while allowing erased type-only imports.

Branch blocker: this head still contains old #1965 head `7815272`, while current #1965 is `2baf7ef`; GitHub correctly reports the stack conflicting. The failed Coverage job is the same stale-base symptom (missing current contract rows and exact pins), not a new lint-design failure. Restack/rebase #1990 onto current #1965, rerun CI, and retain only the #1990 lint/deletion delta. Not merge-ready until synchronized.

PR #1995, opened 2026-08-24
fix(ios): verify fill's synthesized-replacement route before reporting success

thymikee:
## Summary

- `fill`'s channel-penalized synthesized-replacement route (`runSynthesizedReplacementRoute`, taken when the XCTest accessibility channel is already penalized under CI load) posted keystrokes and returned `ok: true` without ever reading the field back — a dropped or still-in-flight character was indistinguishable from success. Same bug class already fixed for bare `type` (#1676/#1924), but that fix never covered this `fill`-only route.
- First attempt reused `type`'s append-mode commit-wait verbatim. That was wrong: its "observed value isn't a prefix of expected → trust the app" rule exists to tolerate legitimate transforms (autocomplete, formatters) during append, but the same rule also waves through a dropped-*middle*-character corruption, since a value with a hole in it is neither a matching prefix nor an exact match. Verified against the real corruption strings from CI ("Ada Lovelace" → "Avelace", "ada@example" → "aexample") that the old rule classifies both as "trust it" and never fails.
- Added a separate, replacement-mode-specific outcome function with no such escape hatch — any non-exact-match value keeps polling until the deadline, then fails. This matches how `isRepairableTextEntryMismatch` already treats every `.replacement`-mode mismatch as failing/repairable unconditionally elsewhere in this file. The `type` route's own function and behavior are untouched.
- Updated `commands.md` since `TEXT_INPUT_COMMIT_NOT_OBSERVED` can now also come from `fill`, not only `type`.

## Verification

- `xcodebuild build-for-testing` (with the runner unit-test flag) — clean, no errors/warnings on touched files.
- Actually ran the tests on a real iOS Simulator via `xcodebuild test-without-building` (not just a build):
  - New regression test directly encodes both real corruption strings and proves the old model reports success while the new model reports `commit-not-observed`.
  - Full pre-existing append/`type` test suite passes unchanged — confirms zero regression to `type`.
  - Along the way, found and fixed a test-fixture bug of my own: an existing integration test used a bare, never-launched `XCUIApplication()`; once the wait actually polls, each poll threw an XCTest-level exception that inflated the failure count independent of any assertion. Fixed by pointing it at Springboard (always running on a booted simulator).
  - Found one unrelated, pre-existing broken test (`testCommonPrefixLengthWalksTheExpectedPrefixOnly`, wrong expected number, confirmed present before this change) — out of scope here, flagged separately.
- Still needs a live iOS Smoke Tests CI run to confirm behavior under actual CI-load timing — this is a CI-load-only flake that doesn't reproduce locally.

## Test plan

- [ ] iOS Smoke Tests CI run (live simulator, CI load) to confirm `fill` no longer silently drops characters and correctly surfaces `TEXT_INPUT_COMMIT_NOT_OBSERVED` when it can't confirm the commit


thymikee:
Reviewed at exact head `0e92e84ab`. Clean. The channel-penalized synthesized `fill` route now waits for an exact replacement value and propagates `TEXT_INPUT_COMMIT_NOT_OBSERVED`; append-mode `type` keeps its separate transformed-input behavior. Regression coverage pins both observed corruption shapes, delayed convergence, final-poll timing, placeholder ambiguity, and production caller wiring. All exact-head checks pass, including Coverage, Swift runner XCTest, and iOS Smoke. Simulator evidence is appropriate for this CI-load simulator failure; no physical-device blocker.

PR #1996, opened 2026-08-24
ci: consolidate CI workflow from 15 jobs to 8

thymikee:
## Summary

A typical code PR used to start ~19 ubuntu workers, most running a 25-second gate after a full checkout and install; queueing for workers dominated wall time. This consolidates and re-shards the CI surface with no gate removed or weakened — every check remains a named `run-gate` step, and the gate manifest derives lane ownership structurally from the workflow YAML.

Job layout changes:

| Change | Detail |
| --- | --- |
| CI workflow 15 → 8 jobs | Lint & Format (+ the static text assertions, run pre-install), Repo Guards (layering/selector/wiring/maestro/mcp-metadata/freerange), Compatibility & Provenance (fallow + replay-compat + released-surface-compat sharing one deep checkout), Typecheck & Package, Coverage shards ×2 + Coverage Report merge job, Integration Tests (absorbs the web platform smoke with step-scoped env) |
| macOS lanes merged | The host-XCTest lane folds into the macOS smoke job: one runner builds one unit-test-flagged bundle consumed by both the host XCTest run (`pnpm check:xctest-selection` follows it to macos.yml) and the replay smoke. Removes the second macos-26 queue wait, which measured 5–6 min per PR. |
| Coverage sharded | Two shards write blob reports (~3m25 each, down from 5m46); Coverage Report merges them, evaluates thresholds once over the full suite, and produces every coverage artifact. The tmpdir leak check runs per shard, since a leak lands on whichever runner executed the file. Shard/merge switches live in vitest.config.ts behind `AGENT_DEVICE_COVERAGE_SHARD`/`AGENT_DEVICE_COVERAGE_MERGE`. |
| FreeRange de-Bunned | `@chenglou/freerange`'s bin is a plain Node script; the Bun setup is gone and the gate joins Repo Guards as its last step. Still GitHub-owned (localRunnable stays false) so the pre-push affected path does not pay its ~3 min. |
| Fixture release no-ops skipped | The Test App Build Cache matrix filters to entries that will actually build, so cached-fingerprint PRs start zero release runners instead of two 5-second jobs. |

Typical PR check lines drop from 28 to ~18–20 (the four CodeQL lines are repo-settings default setup, intentionally untouched). Ubuntu worker demand drops by roughly half; every non-device ubuntu lane now completes in under 5 minutes.

## Validation

Evidence at 356ce79b5; all 20 checks green on that head:

- The merged macOS lane ran end to end on the PR head: host XCTest suite executed with the source-derived executed-count assertion passing, and the macOS replay passed against the unit-test-flagged binary — the live evidence for the one behavior-affecting merge.
- Coverage pipeline: both shards green at ~3m25, Coverage Report merged blobs and passed the changed-line gate in 34s.
- `pnpm check:gate-manifest` → ok (48 checks wired); gate manifest tests 49/49, affected-selector tests 62/62, xctest-selection tests 37/37; `pnpm check:affected --run`, typecheck, lint, format clean.
- Local smoke of the shard→blob→merge cycle verified blob union and threshold evaluation before pushing.

Two issues surfaced by real runs and fixed here: vitest enforces coverage thresholds per shard (shards report ungated; only the merged run gates), and upload-artifact v4 silently skips dot-prefixed directories without `include-hidden-files`.

Known pre-existing flake exposed once under shard load (not from this change): `src/daemon/__tests__/request-save-script-transports.test.ts` hit `ENOTEMPTY` in its own afterEach cleanup racing an async daemon-log write. It passed on re-run and repeatedly in isolation; worth a follow-up issue.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,13 +28,10 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  # Two single-`rg`-assertion jobs (formerly `ios-runner-swift-compat`,
-  # `no-test-di-seams`, added independently in b79bd8601 / 9eb060406) folded
-  # into steps here: each was checkout + one grep, paying full job
-  # scheduling/checkout overhead and its own PR status-check line for what is
-  # a single assertion. Each step keeps its own failure message. See #1462.
-  static-checks:
-    name: Static Checks
+  # Text-only assertions run before the toolchain setup, so a grep failure does
+  # not wait on an install.
+  lint:
+    name: Lint & Format
     runs-on: ubuntu-latest
     timeout-minutes: 10
     steps:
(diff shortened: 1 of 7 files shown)


thymikee:

Cross-PR finding with #1994 (currently open, all-green): once both land, the coverage lane breaks.

#1994 changes `test:coverage:ci` to two chained invocations (`vitest run --coverage --project=!fuzz-worker && vitest run --project fuzz-worker`). `AGENT_DEVICE_COVERAGE_SHARD` is job-level env here, so it applies to **both** legs: on each shard runner the uninstrumented fuzz-worker leg overwrites `.vitest-reports/blob-<shard>.json` written by the instrumented leg (same shard id → same blob filename), and the uploaded artifact loses the instrumented results that `Coverage Report` merges and gates on. Symmetrically, this PR's Coverage Report job runs `gate: unit-ci` with `AGENT_DEVICE_COVERAGE_MERGE=1`, which would execute #1994's second leg in merge mode — re-merging blobs instead of running `fuzz-worker`.

Neither PR is wrong alone; they need one reconciliation before both merge. Detail posted on #1994.

Separately, good news: the "known pre-existing flake" this PR documents (`request-save-script-transports.test.ts` ENOTEMPTY) already has a fix up in #1999 (flushes `flushSessionEventLogWrites()` before the temp-root cleanup), so the follow-up issue can close when these land.

thymikee:
Review of exact head `356ce79b5ff131d2c0227b5d8d65ea4912a4f178`: no code-level findings. Gate ownership remains intact, both coverage shards merge to the full union with thresholds enforced at merge, FreeRange runs under Node, fixture release matrices skip correctly, and the combined macOS lane proves exactly 179 source-derived host XCTests before a passing replay.

The head is not ready for the label only because its base is stale: current `main` is `66458c7916e59e56ab469a06f5e1f30171240fff`, and main added Apple runner XCTest sources that this PRs host-XCTest migration directly owns. Rebase onto current main and rerun exact-head CI; no implementation change is otherwise indicated.

Merge order: #1996 first, then rebase and revalidate #1994, whose environment clearing is a follow-on adjustment.

thymikee:
Re-review of exact head `bc4820109c66af5cae16de690bc274030c9edce9`: clean and ready for human review. The rebase now includes current mains five new declared XCTest methods; the selection gate finds 209 declared methods with zero dark, the combined macOS lane executes exactly 183 host tests with zero failures, then passes replay. Coverage shards merge to the same full 1,098-file / 8,145-test union, and the 48-check gate manifest remains intact. All completed checks are green; only iOS smoke is still in progress.

Sequence remains #1996 first, then rebase/revalidate #1994 so its environment clearing applies to the consolidated job topology.

PR #1999, opened 2026-08-24
fix(test): drain session-event-log writes before rmSync cleanup in save-script transport tests

### LINKED ISSUE #1998: Test flake: request-save-script-transports afterEach rmSync races the fire-and-forget session-event-log write (ENOTEMPTY)
## Evidence

Surfaced once by the coverage-shard lane on PR #1996 (CI consolidation): `src/daemon/__tests__/request-save-script-transports.test.ts` failed with `ENOTEMPTY` inside its own `afterEach` cleanup, under shard load. It passed on re-run and passes repeatedly in isolation. PR #1996's description records it under "Known pre-existing flake" — the CI change only exposed it; the race predates it.

## The race, precisely

The test's cleanup removes each per-test temp root recursively:

```ts
// src/daemon/__tests__/request-save-script-transports.test.ts:56
afterEach(() => {
  for (const root of roots.splice(0)) fs.rmSync(root, { recursive: true, force: true });
});
```

That root contains `sessions/<session>/events.ndjson`, and the daemon request path writes to it **asynchronously, fire-and-forget**:

- Every request handled by `createRequestHandler` (both the socket and HTTP transports in this file, including the *rejected* `--save-script` requests) appends `request.started` / `request.finished` events via `sessionStore.recordEvent(...)` (`src/daemon/request-execution-scope.ts:153`, `:257`, `:271`, `:416`).
- `appendSessionEvent` → `queueEventLogWrite` (`src/daemon/session-event-log.ts:84`) chains an **unawaited** promise per log path: `mkdir -p` the session dir → rotation check → `fs.promises.appendFile(events.ndjson)`. The daemon response resolves without awaiting this chain, so the test body's `await send(...)` returning does not mean the write has landed.

Under a loaded runner the queued append can still be in flight when `afterEach` runs. `fs.rmSync(root, { recursive: true })` unlinks the directory's entries and then rmdirs each directory; if the pending chain re-creates `sessions/<session>` (its `mkdir -p`) and/or `events.ndjson` (append creates the file) between the unlink sweep and the rmdir, the rmdir hits a non-empty directory → `ENOTEMPTY`. In the quiet case the write happens to land before cleanup, which is why the test is green in isolation and on re-run.

Note the PR description's shorthand "async daemon-log write" is slightly off: writes to `daemon.log` itself (the diagnostics ndjson) are synchronous (`appendFileSync` in `src/utils/diagnostics.ts:270`). The only async writer in this path is the **session event log** under `sessions/<session>/`.

## Fix direction

The quiesce hook already exists: `flushSessionEventLogWrites()` (`src/daemon/session-event-log.ts:77`) awaits all pending event-log writes (also surfaced as `SessionStore.flushEvents()`). The fix is a one-liner in the test's cleanup:

```ts
afterEach(async () => {
  await flushSessionEventLogWrites();
  for (const root of roots.splice(0)) fs.rmSync(root, { recursive: true, force: true });
});
```

Alternative (weaker): retry-on-`ENOTEMPTY` in the cleanup, e.g. `fs.rmSync(root, { recursive: true, force: true, maxRetries: 3, retryDelay: 50 })`. Draining the writer is preferable — it removes the race instead of tolerating it, and `maxRetries` alone can still lose if the append lands after the last retry's sweep.

Worth a quick sweep while fixing: any other daemon test that drives requests through `createRequestHandler` (or `SessionStore.recordEvent`/`recordAction`) and then `rmSync`s its temp root in cleanup has the same latent race.

## Links

- Exposed by (and documented in) PR #1996.

thymikee:
## Summary

Fixes the pre-existing flake documented in #1996 ("Known pre-existing flake") and analyzed in #1998: `request-save-script-transports.test.ts` hit `ENOTEMPTY` in its own `afterEach` under coverage-shard load.

Every request the tests send queues a fire-and-forget session-event-log append (`queueEventLogWrite`: unawaited `mkdir -p` → rotate → `appendFile` into `sessions/<session>/events.ndjson`). The daemon response resolves before the write lands, so under load a pending append could re-create the session dir between `rmSync`'s unlink sweep and its rmdir — cleanup then fails with `ENOTEMPTY`.

The fix awaits the existing quiesce hook, `flushSessionEventLogWrites()`, at the top of the `afterEach` before removing the temp roots. One line plus the import.

Closes #1998

## Validation

- `pnpm vitest run src/daemon/__tests__/request-save-script-transports.test.ts` — 8/8 green, six consecutive runs.
- `pnpm format:check`, `pnpm lint`, `pnpm typecheck` clean.

The race itself is timing-dependent (one occurrence on a loaded shard), so the runs above show no regression rather than proving the flake gone; the flush removes the ordering hazard by construction. #1998 notes the wider sweep (other daemon tests that drive `createRequestHandler` then `rmSync` a temp root) as follow-up.

### DIFF
--- a/src/daemon/__tests__/request-save-script-transports.test.ts
+++ b/src/daemon/__tests__/request-save-script-transports.test.ts
@@ -31,6 +31,7 @@ import {
   skipWhenLoopbackUnavailable,
 } from '../../__tests__/test-utils/loopback.ts';
 import { mkdtempForTestSync } from '../../__tests__/test-utils/tmp-dir.ts';
+import { flushSessionEventLogWrites } from '../session-event-log.ts';
 
 const TOKEN = 'save-script-transport-token';
 const SESSION = 'save-script-transport';
@@ -53,7 +54,10 @@ type Harness = {
 
 const roots: string[] = [];
 
-afterEach(() => {
+afterEach(async () => {
+  // #1998: request handling queues fire-and-forget session-event-log appends;
+  // drain them before rmSync or a late append re-creates the dir → ENOTEMPTY.
+  await flushSessionEventLogWrites();
   for (const root of roots.splice(0)) fs.rmSync(root, { recursive: true, force: true });
 });
 


thymikee:
Reviewed exact head `dcc887362`. Clean. The test cleanup now awaits the existing session-event-log quiesce hook before removing its temp roots, which removes the ENOTEMPTY ordering hazard at the owning async boundary without changing production behavior. The focused regression and all 27 exact-head checks pass.

Issue #1983, opened 2026-08-23
refactor(snapshot): assign and scope the Wave 4 shared presentation debt (#1739)


thymikee:
## Outcome

Give the Wave 4 shared capture/presentation debt a named owner and an explicit boundary, so it stops
being the one unassigned block on #1739's parallel-safe track.

#1739 already states the case and the urgency:

> Wave 4 shared capture debt remains and is still unassigned: the 10 production files under
> `src/daemon/snapshot-presentation/ios/`, `android-snapshot-freshness.ts`,
> `android-snapshot-timeout-evidence.ts`, `screenshot-overlay-android.ts`, and their
> observation/interaction call sites. […] **This is the one remaining block of work that is
> file-disjoint from the descriptor units**, so it is the natural parallel track; it must be assigned
> before Wave 5 migrates the freshness call sites in `selector-capture-runtime.ts` /
> `deferred-interaction-outcome.ts`.

This issue exists to carry that assignment and its scope. It is not a new wave unit and does not
change #1739's denominator.

## Why now — churn evidence

From a co-change analysis of the last 200 commits on `main` (`2026-08-10 → 2026-08-23`, HEAD
`d1fc80f`):

- **37 of 200 commits (18%)** are snapshot or presentation work.
- 19 of those edit presentation logic in **two or more zones**.
- **7 of the 21 Swift-presentation commits (33%)** also edit TypeScript presentation in the same
  commit.
- The heaviest file co-change pairs in the whole repository sit inside this cluster.

Presentation behavior currently lives in four places, with types-only sharing beneath them:

| Location | Lines | Contents |
| --- | ---: | --- |
| `packages/kernel/src/snapshot.ts` | 304 | types only |
| `apple/runner/…` (Swift) | ~10 files | in-runner presentation |
| `src/daemon/snapshot-presentation/ios/` | 2,013 | iOS presentation rules |
| `src/platforms/android/snapshot-presentation.ts` | 277 | Android fold |
| `src/platforms/web/agent-browser-snapshot.ts` | 223 | web fold |

`CONTEXT.md` already specifies the intended end state — the clip fold as "the regular projection's
*single* visibility interpreter, run inside presentation for every backend […] Platform differences
enter as a fold policy, never as a backend exception." That interpreter does not exist yet. #1973
added cross-runtime presentation conformance tests, which detect divergence between the
implementations rather than removing the reason they diverge; that is the right interim instrument
and the wrong end state.

[...]

PR #2000, opened 2026-08-24
refactor(snapshot): type the acquisition producer beside the platform channel

thymikee:
## Summary

Give every snapshot a typed **producer** — the acquisition component that made the raw tree — as a third identity axis beside the platform channel (`SnapshotBackend`) and the in-plan capture strategy (`SnapshotCaptureBackend`).

Today three producers with different guarantees share the `backend: 'xctest'` stamp: the local Apple runner (whose output went through the runner's presentation — clip fold, effective geometry, scope), Appium page-source XML (`provider-webdriver`), and limrun element trees (`provider-limrun`) — the latter two reach the wire with none of those guarantees. The `'android'` channel conflates the local uiautomator path with Appium-sourced trees the same way. Daemon logic keyed on the channel cannot tell them apart.

This is the types-before-semantics step for the shared-presentation track (#1983): no consumer changes behavior here. The field is what the host-side presenter wiring and the double-presentation removal will key on, and what a future XCTest-free simulator producer will register as.

- `SnapshotProducer` union + axis doc in `@agent-device/kernel/snapshot`; `SnapshotState.producer` (optional, daemon-side only — the value does not cross the CLI wire, same as `backend`)
- `SnapshotResult.producer` is **required**, so the compiler enumerates every producer: `apple-runner`, `android-uiautomator`, `appium-source`, `limrun-ios-tree`, `macos-helper`, `linux-atspi`, `agent-browser`, `harmonyos-uitest`
- `buildSnapshotState` carries the producer into `SnapshotState`; the capture-handler `SnapshotData` seam passes it through
- CONTEXT.md gains the **Snapshot producer** vocabulary entry

## Validation

- `pnpm check:affected --run`

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -187,6 +187,14 @@ A backend-owned accessibility value before snapshot presentation.
 **Snapshot acquisition**:
 One backend attempt's raw accessibility nodes and attempt-level capture facts.
 
+**Snapshot producer**:
+The acquisition component that produced a snapshot's raw tree — the third identity axis beside
+the platform channel (`SnapshotBackend`) and the in-plan capture strategy
+(`SnapshotCaptureBackend`). One channel is fed by producers with different guarantees (the
+`xctest` channel carries Apple-runner, Appium page-source, and limrun element trees), so logic
+that assumes presentation, scope, or geometry guarantees keys on the producer, never on the
+channel alone.
+
 **Presentation options**:
 The policy input controlling how one snapshot acquisition becomes a public projection.
 
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -189,11 +189,9 @@ One backend attempt's raw accessibility nodes and attempt-level capture facts.
 
 **Snapshot producer**:
 The acquisition component that produced a snapshot's raw tree — the third identity axis beside
-the platform channel (`SnapshotBackend`) and the in-plan capture strategy
-(`SnapshotCaptureBackend`). One channel is fed by producers with different guarantees (the
-`xctest` channel carries Apple-runner, Appium page-source, and limrun element trees), so logic
-that assumes presentation, scope, or geometry guarantees keys on the producer, never on the
-channel alone.
+the platform channel and the in-plan capture strategy. Producers on one channel carry different
+guarantees, so presentation, scope, and geometry logic keys on the producer, never the channel
+alone.
 
 **Presentation options**:
 The policy input controlling how one snapshot acquisition becomes a public projection.
--- a/packages/contracts/src/interactor-types.ts
+++ b/packages/contracts/src/interactor-types.ts
@@ -14,6 +14,7 @@ import type {
   Rect,
   SnapshotBackend,
   SnapshotOptions as BaseSnapshotOptions,
+  SnapshotProducer,
 } from '@agent-device/kernel/snapshot';
 
 export type RunnerContext = {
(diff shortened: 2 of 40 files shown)


thymikee:
Not ready at `31fb491c6`. The typed `SnapshotProducer` seam is appropriately deep and the production propagation is sound, but two required gates are deterministically red. Changed-line coverage is 2/4 (50%): add direct producer assertions for `packages/provider-limrun/src/ios.ts:260` and `src/core/interactors/harmonyos.ts:48`. Fallow also reports three above-threshold functions, including `createAppleInteractor`; resolve that health debt at the owning module rather than bypassing the gate. All other checks, including platform smokes, pass; no additional live-device evidence is required for this type-only slice once these gates are green.

thymikee:
Review of exact head `50b7fd95f39b6147bc87f80a3b7370c71d56ec44`:

The prior coverage and Fallow blockers are resolved cleanly: Limrun iOS and HarmonyOS now assert their real producers, and coverage/Fallow are green without suppressions.

One remaining P2 contract gap blocks readiness: `SnapshotResult` makes `backend` and `producer` independent unions, so impossible provenance pairs such as `{ backend: "android", producer: "apple-runner" }` type-check. This seam is being added specifically so future presentation logic can trust producer guarantees. Encode the valid channel-producer matrix as a discriminated provenance type, carry that coupled type through snapshot state, and add a negative type test. That makes the invariant structural instead of relying on each constructor to pair two independent strings correctly.

thymikee:
Re-reviewed `50b7fd95f39b6147bc87f80a3b7370c71d56ec44`: CI is now fully green, including device smoke, but the code blocker remains. `backend` and `producer` are still independent unions, so invalid provenance pairs type-check. Please carry one kernel-owned discriminated provenance value atomically through `SnapshotResult` and state, with a negative type-level regression proving invalid pairs cannot compile. Not ready for human review yet.

thymikee:
Re-reviewed `90cc71a4b29147a1b0e44399ad955ea85b229532`: the kernel-owned discriminated provenance union and negative type-level proof fix the prior code blocker. This head is still not ready: `CONTEXT.md` is 12,073 bytes, 73 over its enforced budget, and the new glossary text includes implementation/architecture detail that belongs at the declaration/ADR seam. Also, Typecheck & Package is genuinely red at `src/daemon/__tests__/test-device-runtime-gateway.ts:96` (the `unavailableDeviceRuntimeGateway` facts/binding no longer satisfy `DeviceRuntimeGateway<PlatformRuntimeOperations>`); Repo Guards repeats the same type error. Trim the glossary entry and repair the typed fixture boundary, then rerun the exact head.

thymikee:
Re-reviewed `c588c7b27308f8a7b1c9d960b8c771db9493df90`: the context budget is fixed and Agent Guidance is green. The blocking type error remains unchanged—`Typecheck & Package` is red at `src/daemon/__tests__/test-device-runtime-gateway.ts:96` because the snapshot fixture still returns an XCTest backend without the now-required producer, so the gateway no longer satisfies `DeviceRuntimeGateway<PlatformRuntimeOperations>`. Repair that typed fixture route and rerun the exact head. Small cleanup while there: keep the glossary definition path-free; the kernel table/path belongs at its declaration or ADR, not `CONTEXT.md`.

thymikee:
Implementation is sound and the prior typed-fixture failure is fixed; completed checks are green. One P2 remains: the new `CONTEXT.md` Snapshot producer entry embeds declaration-path and architecture details (`SnapshotProvenance`, `@agent-device/kernel/snapshot`, atomic carrier rules). This repository keeps `CONTEXT.md` glossary-only. Trim it to the plain domain meaning and leave declaration/architecture ownership at the typed seam or ADR. iOS smoke is still pending.

thymikee:
Re-review clean at `30bd8096`: the glossary entry now stays domain-only, and the correlated snapshot provenance typing/fixtures are sound. Coverage (2) failed only in the pre-existing 1ms wait timeout decoration assertion; this PR changes that fixture solely by adding the required `agent-browser` producer and changes no wait behavior, so this is a timing flake rather than a code finding. Code is ready, but CI needs a green retry before merge.

PR #2001, opened 2026-08-24
fix: stop replay retries after cleanup failures

thymikee:
## Summary

Replay test retries now stop when the preceding attempt cannot close its session. A rejected close response is promoted to the same cleanup failure path, and the failed attempt is classified as infrastructure so a later attempt cannot contend with its retained device claim.

Adds regression coverage for the macOS `open "System Settings"` shape: a failed attempt whose close is rejected runs exactly once instead of reaching `DEVICE_IN_USE` on the next attempt.

Rebased onto current `origin/main`: this PR now contains one cleanup-retry commit and four targeted source/test files only.

Motivating CI evidence: merged PR #1952 macOS run 32734629087, job 97454414587; and PR #1994 macOS run 32734431023, job 97453800202.

## Validation

- Regression was observed red before the production change, then green.
- Focused replay-test and daemon suite tests passed (23 tests).
- `pnpm test:replay:macos` passed against the local macOS System Settings replay after the rebase.
- `pnpm check:affected --run` passed after the rebase (GitHub-authoritative lanes remain pending on this head).

Docs and skills unchanged: the retry/session lifecycle contract is internal and existing guidance already covers it.

### DIFF
--- a/packages/replay-test/src/internal/__tests__/session-test-runtime.test.ts
+++ b/packages/replay-test/src/internal/__tests__/session-test-runtime.test.ts
@@ -154,6 +154,33 @@ test('runReplayTestAttempt keeps a passing replay passed when finalization fails
   expect(cleanupSession).toHaveBeenCalledWith('default:test:pass');
 });
 
+test('runReplayTestAttempt marks a failed cleanup as infrastructure so the scheduler cannot retry into its session', async () => {
+  const cleanupSession = vi.fn(async () => {
+    throw new Error('macOS session still owns host-macos-local');
+  });
+
+  const result = await runReplayTestAttempt({
+    filePath: '01-cleanup-failure.ad',
+    sessionName: 'default:test:cleanup-failure',
+    requestId: 'req-cleanup-failure',
+    runReplay: async () => ({
+      status: 'failed',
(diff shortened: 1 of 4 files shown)


thymikee:
Review of exact head `8b9cfb0807b6be4637321ddebc62d9dfa449008c`:

The functional fix is sound and sits at the right ownership seam: a rejected close response becomes a cleanup error, the attempt is classified as infrastructure, and the existing scheduler stops before retrying into a retained device claim. Both regression tests are non-vacuous.

The PR is not ready because its branch topology is wrong: it conflicts with `main` and carries 13 unrelated iOS snapshot commits (31 files, +939/-148) ahead of the one cleanup-retry commit. Rebuild/rebase it onto current `main` as only the cleanup-retry change and its four targeted source/test files, then rerun exact-head CI and re-stamp the macOS replay evidence. Do not apply labels.

thymikee:
Re-review of exact head `469455f83d34387341becdd49c0c4777dbdc0b5a`: clean and ready for human review. The prior topology blocker is resolved: one commit, four targeted files, mergeable, with the cleanup failure classified at the owning attempt seam and non-vacuous regression coverage. Exact-head static, coverage, integration, and macOS smoke are green; the macOS lane ran the System Settings replay successfully. iOS smoke is still in progress and is the only pending check.

PR #2002, opened 2026-08-24
refactor: make interaction response cleanup platform-neutral

thymikee:
## Summary

Make internal interaction diagnostic cleanup platform-neutral at the existing response transform seam.

- strip runner timing and sequence diagnostics with one pure function
- remove the Apple-only normalizer and daemon-to-Apple import
- stop threading an unused session through touch response transformation
- leave the Android app-state and viewport edges for the Wave 6 runtime-owner work

Part of #1739.

## Validation

- planted the response-cleanup regression and observed it fail before the implementation
- `pnpm check:affected --run`

No live-device pass: this is response shaping only, with provider-backed interaction contracts included in the affected gate.


thymikee:
Not ready yet: the planted regression proves cleanup only through the descriptor-backed press path. `longpress` and `hover` take `transformTouchResponseData`’s commandless branch, while their existing contract cases use empty backend data, so returning raw data there would still pass. Add a noisy longpress or hover regression at this seam (or end to end) asserting `currentUptimeMs`, `gestureStartUptimeMs`, `gestureEndUptimeMs`, and `sequenceResults` are omitted; verify it red before this change. The ownership move itself is sound.

thymikee:
Addressed in 69f26a35f. Added a commandless longpress regression at transformTouchResponseData with noisy currentUptimeMs, gestureStartUptimeMs, gestureEndUptimeMs, and sequenceResults data. I planted the old raw-return behavior and observed this exact test fail with all four fields leaking, then restored the cleanup and verified the focused tests plus pnpm check:affected --run pass.

thymikee:
Re-review clean at `69f26a35`: the new commandless longpress regression covers all four internal diagnostics, and longpress/hover production routing exercises that exact transform. Code is ready for human review; remaining CI is still in progress.

Issue #531, opened 2026-05-14
Android text entry should own safe ASCII fallback and chunking


thymikee:
## Problem

When Android text input misbehaves, agents fall back to raw `adb`, clipboard, or paste. `typeAndroid` also sends ASCII in one shell command, which can truncate in some app/IME states.

## Desired behavior

Agent-device should own the safe Android ASCII text-entry fallback internally.

## Acceptance criteria

- ASCII `type` and any Android fallback path chunk text safely inside agent-device.
- Spaces and common punctuation are handled correctly.
- Non-ASCII does not silently downgrade to unsupported shell input.
- Clipboard/paste is not used as the agent-owned fallback for app text entry.
- Fallback is gated off when IME-owned focus is detected.
- Help/docs do not tell agents to recover app text entry with raw `adb`, clipboard, or paste.

## Testing

Unit:
- ASCII chunking: `"filed the expense"` lands completely.
- Punctuation/email string lands completely.
- Non-ASCII refusal/path remains explicit.
- IME-owned state blocks fallback.

New Expensify Android verification:
- In Adam chat composer, enter and send `"filed the expense"`.
- Expected: no raw `adb` workaround, no clipboard command, no paste command.
- Verify chat contains exactly `filed the expense`.

PR #2003, opened 2026-08-24
perf(android): one dumpsys read per window question; batch text via the helper IME

thymikee:
Closes the remaining verified Android hot-path items from the 2026-08 audit (V2 + A8-as-IME-batch).

## Summary

- **One dumpsys read answers every window question.** New `window-state.ts` owns the dump sequence for foreground-app and blocking-dialog detection together. Per tap on a clear screen: **12 → 4** mocked adb spawns, **6 → 2 live** (~351 KB → ~128 KB parsed). The ANR recovery poll drops from up to 4 dumps per tick to 1 (worst-case ceiling ~78 → 13 spawns). Demotion of the never-answering variant is grounded in live tracing.
- **Text entry batches whenever the device is on the helper IME** — no `--test-ime` flag needed. Live: `fill` = 2 `am broadcast`, zero `input text`, zero delete-key rounds; `type` = 1 broadcast. Shell chunking remains the fallback (#531's truncation class guarded by the existing error classification).
- Readiness cache (~1s) and ANR poll alternation preserve incident-derived outcome shapes (#1832); "dump answered nothing" is carried as *unknown*, never as *observed clear*.

## Validation

- Adversarial review initially returned **do-not-merge** with four confirmed correctness defects (F1–F4) + stale-base warning (F8); all fixed red-first and re-reviewed in-tree (`REVIEW.md`).
- Rebased onto current main including the ADR-0019 touch-dispatch waves; the superseded dispatchCommand test file converged onto main's platform-runtime shape.
- Gates: android/interactors/daemon suites (2,785 tests), `check:affected`, format, fallow green at pushed head. Live evidence gathered on `Pixel_7_review` (Android 16) with backend proof per device-verification doc.


thymikee:
Not ready. P1: the promoted active-helper route still calls `assertAndroidShellTextSupported` before probing the active IME in both `typeAndroid` and `fillAndroid`, so a helper already active on-device but absent from this daemon’s activation cache still rejects Unicode. Probe first, route helper-owned text through the helper, and add cache-empty Unicode type/fill regressions. P1: both Coverage shards have deterministic owner failures—the orientation test mocks the retired `app-lifecycle` seam instead of `window-state`, and two new dialog-readiness tests fail. Fix the fixtures/behavior and rerun Coverage. P2: remove the stale 204-line `RESULT.md` worker handoff; compact evidence belongs in the PR body. The central `window-state.ts` topology itself is sound.

thymikee:
Still not ready. P1: both type and fill reject non-ASCII before probing the active IME, so cache-empty devices already using the helper still cannot take the Unicode-capable broadcast route. Probe first and add cache-empty Unicode type/fill regressions. Coverage (2) is also an owner failure: orientation still mocks retired `app-lifecycle.ts` instead of the actual `window-state.ts` seam. Finally remove stale `RESULT.md`. The scope-bound `window-state` reader topology itself remains sound.

thymikee:
Still not ready at `b73dd01b`. The stale orientation mock is fixed and both Coverage shards are now green, but the product P1 is unchanged: type/fill still reject Unicode before probing an already-active helper IME, so a cache-empty daemon cannot use the supported broadcast route. Probe ownership first and add cache-empty Unicode regressions. Also remove the stale 216-line `RESULT.md`; the sound `window-state` simplification does not justify retaining an obsolete worker handoff.

thymikee:
Addressed at 5289f116c: `text-input.ts` now reads the active IME **before** the non-ASCII rejection — a cache-empty daemon on a device already riding the helper IME takes the broadcast route for Unicode text; the ASCII shell path only applies when the helper IME is not active. Red-first regressions added in `text-input-fill.test.ts` / `text-input-test-ime.test.ts` (cache-empty + helper-active → `am broadcast`, zero chunked `input text`). Stale `RESULT.md` removed.

thymikee:
Ready for human review at `5289f116`. The previous blockers are resolved: Unicode admission now happens only after probing the active IME, cache-empty/helper-active type and fill regressions cover the batch route, the stale orientation mock is fixed, both Coverage shards pass, and `RESULT.md` is gone.

Coordinator live evidence on Pixel 9 Pro XL / Android 16 at the exact head:
- `pnpm build` and `pnpm build:android` passed in an isolated worktree.
- Opened Settings with a fresh daemon using `--no-test-ime`, then selected `com.callstack.agentdevice.imehelper/.TestInputMethodService` externally so the daemon's activation cache remained empty.
- `type 'Café ☕ 你好' ... --session pr2003-cache-empty` passed.
- Follow-up `snapshot -i --json` showed the focused search field value exactly `Café ☕ 你好`, `androidSnapshot.backend: android-helper`, helperVersion `0.20.10`, and healthy snapshot quality.

All exact-head checks are green except iOS Smoke still in progress; there is no code finding, conflict, or owner-action CI failure.

thymikee:
All exact-head checks are now green; the prior ready-for-human verdict and live Android evidence stand unchanged.

PR #2004, opened 2026-08-24
perf(daemon): revalidate the source code-signature cache by stat instead of rereading the graph

thymikee:
Verified audit item V5 (dev-from-source path only; dist behavior byte-identical).

## Summary

Every CLI invocation from source paid 803 `readFileSync` (4.6 MB) + 3,915 `statSync` to recompute the daemon code signature before any device work (~26–30 ms warm, on top of ~10 stat/probe redundancies).

- The graph walk is split into `walkDaemonCodeGraph` (stamps: label+size+mtime) and `formatDaemonCodeSignature`; `computeDaemonCodeSignature` is externally unchanged.
- New `code-signature-cache.ts` persists the stamp list under `os.tmpdir()/agent-device-code-signature/<sha1(root+entry)>.json`; revalidation is stat-only. A changed walk invalidates through ordinary stamp comparison because the walker is inside its own graph.
- Adversarial-review F1 fixed red-first: zero-entry or foreign-graph documents are refused (previously permanently valid — could restart-loop the daemon); F2 addressed; symlink-key parity with `buildSourceCheckpointStateDirName`'s realpath noted and handled.
- `resolveDaemonLaunchSpec`/`readVersion` memoized per process; repair-divergence sniffing now reads the typed `ReplayDivergenceResume` contract instead of ad-hoc casts.

## Validation

- Adversarial review of the first implementation: 8/10 claims PASS; F1 (high, zero-entry restart-loop) and F2 fixed with planted-document tests; numbers re-measured independently.
- Owning suites green (66 tests across daemon-client), fallow/format clean at pushed head.

### DIFF
--- a/packages/contracts/src/facades/divergence.ts
+++ b/packages/contracts/src/facades/divergence.ts
@@ -8,6 +8,7 @@ export {
   createReplayDivergenceSanitizer,
   formatReplayDivergenceReport,
   measureReplayDivergenceBytes,
+  readReplayDivergenceResume,
   sanitizeReplayDivergenceField,
   scrubReplayVarValues,
   truncateUtf8Field,
--- a/packages/contracts/src/replay-divergence.test.ts
+++ b/packages/contracts/src/replay-divergence.test.ts
@@ -4,6 +4,7 @@ import {
   applyReplayDivergenceLevelCaps,
   boundReplayDivergence,
   measureReplayDivergenceBytes,
+  readReplayDivergenceResume,
   sanitizeReplayDivergenceField,
   REPLAY_DIVERGENCE_DEFAULT_REF_LIMIT,
   REPLAY_DIVERGENCE_DIGEST_REF_LIMIT,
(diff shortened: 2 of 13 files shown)


thymikee:
Not ready. P1: the cache can validate a stale graph when resolution changes without touching any recorded file—for example `./dep` initially resolves to `dep/index.ts`, then a higher-priority `dep.ts` is created. Lines 39–50 explicitly accept one-edit-stale daemon reuse, but code identity must fail closed. Cache the resolution-relevant directory facts (or use a simpler invariant) and add a planted-red regression proving cached equals fresh after that sibling appears. Separately, the replay-reader refactor adds an eager `json.ts` edge and currently fails the divergence-facade and CLI closure budgets; remove/split that unrelated eager edge rather than masking the growth.

thymikee:
Still not ready. The closure-budget issue is fixed, but the cache correctness P1 remains: it stats only recorded graph files. With extensionless `import './dep'`, warm while it resolves `dep.js`, then add higher-priority `dep.ts` without touching any recorded file; a fresh walk selects `dep.ts`, while the cache validates the old stamps and can reuse a mismatched daemon. Add a planted-red cached-vs-fresh regression for this resolution change and make the cache track resolution-relevant directory facts or otherwise fail closed.

thymikee:
Still blocked at `ec9b3280`. Warm-cache validation checks only recorded file stamps, but extensionless resolution probes `.ts` before `.js`: if `./dep` currently resolves to `dep.js`, adding `dep.ts` changes a fresh graph without changing any cached stamp, so the stale signature is reused. The source comment explicitly accepts this stale window; that is not a valid cache invariant. Track/probe the resolution candidates or invalidate from the owning resolver, and add a cached-vs-fresh `dep.js` → newly added `dep.ts` regression. iOS smoke is also red, apparently at an unrelated fill.

thymikee:
Addressed at 8e78bfa24: the walk now records every resolution candidate probed-and-missed ahead of a winner (and all candidates of unresolved specifiers) as `absentPaths`; the cache revalidates each is still missing via stat alone and re-walks when any appears — so `./dep` flipping from `dep.js` to a newly added `dep.ts` can no longer ride a stale signature. Red-first regression added (`re-walks when a higher-precedence extension appears`, plus lower-precedence-stays-warm and gained-target cases). On the iOS smoke: this branch touches no iOS path; rerunning per the known #1844 text-entry flap.

thymikee:
Ready for human review at `1213ac6a`. The previous cache-correctness blocker is resolved: the graph walk now records missed resolution candidates ahead of each winner (and all candidates of unresolved specifiers), cached validation re-walks when one appears, and red-first regressions cover higher-precedence appearance, lower-precedence non-invalidation, and a missing target becoming resolvable. The unrelated eager closure edge remains removed. The rebase conflict artifact is gone; the branch is mergeable and current lint/format, size, typecheck/package, provenance and completed CodeQL checks pass. Remaining CI is still in progress; no code finding, conflict, or confirmed owner-action failure remains.

PR #2005, opened 2026-08-24
refactor(snapshot): establish presentation ownership boundary

thymikee:
## Summary

Establishes the first Wave 4 snapshot-presentation ownership slice for #1983.

- moves host-side iOS post-wire presentation and tree helpers out of `src/daemon` into `src/snapshot/snapshot-presentation`
- adds an explicit `@agent-device/contracts/snapshot-presentation` carrier, clip fold, and regular serializer consumed by Android
- moves shared React Native overlay predicates into `@agent-device/contracts/react-native-overlay`
- keeps Swift runner-side presentation and the existing shared conformance fixture separate across the runtime boundary
- adds structural boundary, contract geometry, and facet tests

Part of #1983. Freshness recovery, timeout evidence, and screenshot-overlay policy remain scoped follow-ups as recorded in ADR 0004.

Docs changed: `CONTEXT.md` glossary and ADR 0004 ownership boundary. No skill changes.

## Validation

- `pnpm check:affected --run` — all runnable checks passed
- `pnpm typecheck`
- focused snapshot-presentation, overlay, and conformance tests
- no live-device run: this patch is a host ownership/contract extraction; native/device lanes remain GitHub-authoritative


thymikee:
## Code quality review — request changes

The directory move itself is correct and clean. Three things layered on top of it are not.

Base for all references: `c206539a27a5e80a6151ea4dd290278b3d31f642`.

---

### 1. The new gate tests paths, not dependencies — and this PR ships a counterexample

[`scripts/layering/snapshot-presentation-boundary.test.ts`](https://github.com/callstack/agent-device/blob/c206539a27a5e80a6151ea4dd290278b3d31f642/scripts/layering/snapshot-presentation-boundary.test.ts) makes three file-existence assertions. But the invariant that matters is "presentation must not depend on daemon assembly", and for production files the rank spine already gives you that: [`model.ts`](https://github.com/callstack/agent-device/blob/c206539a27a5e80a6151ea4dd290278b3d31f642/scripts/layering/model.ts#L45) ranks `snapshot` at 1 and `daemon-server` at 4. The directory move *is* the guarantee. The new test is a weaker restatement of it.

The one gap the spine does not cover is test files — and the PR lands three violations there:

- [`ios/transitions.test.ts:4`](https://github.com/callstack/agent-device/blob/c206539a27a5e80a6151ea4dd290278b3d31f642/src/snapshot/snapshot-presentation/ios/transitions.test.ts#L4) imports `../../../daemon/snapshot-state.ts`
- [`ios/publication-membership.test.ts:3`](https://github.com/callstack/agent-device/blob/c206539a27a5e80a6151ea4dd290278b3d31f642/src/snapshot/snapshot-presentation/ios/publication-membership.test.ts#L3) imports the same
- [`ios/transitions.test.ts:3`](https://github.com/callstack/agent-device/blob/c206539a27a5e80a6151ea4dd290278b3d31f642/src/snapshot/snapshot-presentation/ios/transitions.test.ts#L3) also reaches into `../../../commands/interaction/runtime/__tests__/test-utils/`

Both files exercise `buildSnapshotState` — they are daemon-assembly tests. They moved into the presentation facet because they happened to sit in the old directory, not because they belong there. The gate cannot see this, because it looks at paths.

Suggestion: make it an edge rule — nothing under `src/snapshot/snapshot-presentation/` may import from `src/daemon/`. It fails today, correctly, and it tells you those two tests belong back in the daemon suite.

### 2. Two single-module identity facades, against the #1969 precedent

[`facades/snapshot-presentation.ts`](https://github.com/callstack/agent-device/blob/c206539a27a5e80a6151ea4dd290278b3d31f642/packages/contracts/src/facades/snapshot-presentation.ts) re-exports all 4 symbols of `../snapshot-presentation.ts`. [`facades/react-native-overlay.ts`](https://github.com/callstack/agent-device/blob/c206539a27a5e80a6151ea4dd290278b3d31f642/packages/contracts/src/facades/react-native-overlay.ts) re-exports all 6 symbols of `../react-native-overlay.ts`. Neither aggregates, neither narrows, and in both cases the subpath name is identical to the module name.

The contracts exports map is now 66 direct-module entries against 16 facades. The two pre-existing single-module facades (`divergence` → `replay-divergence.ts`, `progress` → `request-progress.ts`) at least rename; these two do not.

The #1959/#1969 finding was explicit that entries pointing directly at source modules beat new facade files — no exhaustiveness burden in `package-boundaries.test.ts`, and no new re-export surface for `check:production-exports`.

Suggestion: point the exports map at `./src/snapshot-presentation.ts` and `./src/react-native-overlay.ts`, and delete both facade files.

### 3. The Android alias shim cancels the extraction it just performed

```ts
import {
  foldSnapshotRect as effectiveAndroidRect,
  serializeRegularSnapshotPresentationNode as serializeAndroidRegularPresentationNode,
} from '@agent-device/contracts/snapshot-presentation';
export { effectiveAndroidRect, serializeAndroidRegularPresentationNode };
```

[`src/platforms/android/snapshot-presentation.ts:3-9`](https://github.com/callstack/agent-device/blob/c206539a27a5e80a6151ea4dd290278b3d31f642/src/platforms/android/snapshot-presentation.ts#L3-L9)

The point of moving the fold into contracts is one shared vocabulary. This shim keeps every caller on the old Android-flavoured names, so the shared vocabulary reaches zero call sites. The names are also now inaccurate — `effectiveAndroidRect` is platform-neutral code.

It protects two production call sites — [`ui-hierarchy-builder.ts:194`](https://github.com/callstack/agent-device/blob/c206539a27a5e80a6151ea4dd290278b3d31f642/src/platforms/android/ui-hierarchy-builder.ts#L194) and [`:329`](https://github.com/callstack/agent-device/blob/c206539a27a5e80a6151ea4dd290278b3d31f642/src/platforms/android/ui-hierarchy-builder.ts#L329) — plus four in tests. Rename those six and delete the shim.

### 4. Four dead exports, kept alive by four new fallow suppressions

This is the one I would block on. [`src/core/react-native-overlay.ts:78-104`](https://github.com/callstack/agent-device/blob/c206539a27a5e80a6151ea4dd290278b3d31f642/src/core/react-native-overlay.ts#L78-L104) turns four functions into pass-throughs, each carrying `// fallow-ignore-next-line unused-export` and the same justification: *"Public through commands/react-native/overlay.ts; fallow cannot resolve this export * barrel."*

I traced every reference across `src/`, `packages/`, and `test/`:

| symbol | callers of the `src/core` export |
| --- | --- |
| `isReactNativeCollapsedWarningWrapperCandidate` | none |
| `isReactNativeCollapsedWarningWrapperWithVisibleBanner` | none |
| `isReactNativeOverlayDismissLabel` | none (used inside the file only) |
| `isReactNativeOverlayMinimizeLabel` | none (used inside the file only) |

`ios/noise.ts` was their only consumer, and it now imports from contracts instead. The named barrel `src/commands/react-native/overlay.ts` is a single `export *` line, and its one consumer (`src/commands/capture/runtime/snapshot.ts`) imports `formatReactNativeOverlayWarning` and nothing else. So the justification is true in letter and empty in fact — fallow was right, and the PR silenced it.

Repo-wide `fallow-ignore` count goes 56 → 60; all four additions are this.

Suggestion: delete the two wrapper-candidate/visible-banner exports outright, and import the two label predicates as plain non-exported imports. The real public surface of that file is two types and two functions.

### 5. `ReactNativeOverlayNode` is duplicated byte-for-byte

Declared identically in [`packages/contracts/src/react-native-overlay.ts:3`](https://github.com/callstack/agent-device/blob/c206539a27a5e80a6151ea4dd290278b3d31f642/packages/contracts/src/react-native-overlay.ts#L3-L6) and [`src/core/react-native-overlay.ts:18`](https://github.com/callstack/agent-device/blob/c206539a27a5e80a6151ea4dd290278b3d31f642/src/core/react-native-overlay.ts#L18-L21). Neither is exported, so nothing will catch the drift when someone adds a field to one. If contracts owns the predicates, it owns their input shape — export it and import it.

### 6. Bespoke file enumerator in the one directory with a written rule against it

[`snapshot-presentation-boundary.test.ts:5-18`](https://github.com/callstack/agent-device/blob/c206539a27a5e80a6151ea4dd290278b3d31f642/scripts/layering/snapshot-presentation-boundary.test.ts#L5-L18) runs its own `execFileSync('git', ['ls-files', ...])` rooted at `process.cwd()`.

`scripts/layering/tracked-sources.ts` exists precisely for this, and its module doc names the failure mode: *"without a shared leaf the two would each grow their own enumerator, and the moment those disagree a rule silently changes scope. That is not hypothetical — it is the #1965 review finding this module exists to fix."* Every sibling test in that directory also derives its root as `path.resolve(import.meta.dirname, '../..')`, not `process.cwd()`.

### 7. The strong test stayed behind in Android

[`src/platforms/android/__tests__/snapshot-presentation.test.ts:84`](https://github.com/callstack/agent-device/blob/c206539a27a5e80a6151ea4dd290278b3d31f642/src/platforms/android/__tests__/snapshot-presentation.test.ts#L84) holds the property test — effective geometry stays inside every positive clip, actionability never upgrades. That property is neutral now. The new [`packages/contracts/src/snapshot-presentation.test.ts`](https://github.com/callstack/agent-device/blob/c206539a27a5e80a6151ea4dd290278b3d31f642/packages/contracts/src/snapshot-presentation.test.ts) is a weaker example-based duplicate of it.

Move the property test to the contract, where the invariant now lives, and leave Android its Android-specific cases. Otherwise the test ownership boundary is the one boundary this PR did not actually draw.

### Nit

[`ios/presentation-visibility.test.ts:3`](https://github.com/callstack/agent-device/blob/c206539a27a5e80a6151ea4dd290278b3d31f642/src/snapshot/snapshot-presentation/ios/presentation-visibility.test.ts#L3) imports `../../../snapshot/snapshot-visibility.ts`, which climbs out of `src/snapshot/` and back into it. Should be `../../snapshot-visibility.ts`. Leftover from the path rewrite.

---

### What is good

The test split is a genuine decomposition, not a way around the ratchet: `presentation.test.ts` goes 1201 → 991, `presentation-visibility.test.ts` takes 221, and the four extracted tests share a coherent seam (`buildSnapshotVisibility` plus scroll indicators). The pin drops off `PINNED_TEST_FILE_LINES` honestly. No file in this PR crosses 1000 lines. The moves are clean `R100` renames with no hidden edits. The ADR text is scoped honestly about what did not move.

---

### Net

The PR moves ~40 files to establish one boundary, then adds a facade hop, an alias shim, and four suppressions that each hide the boundary it just drew. Findings 2, 3 and 4 are all deletions — doing them makes the extraction smaller and makes the new shared vocabulary actually reach the code.

*Review scope: static analysis of the diff and the surrounding call graph. No tests were run.*

thymikee:
The ownership direction is sound, but this slice is not ready. Coverage (1) exposes two avoidable identity facades and +2 eager CLI modules: map the contract exports directly to their source modules and keep the exact entry pins, rather than budgeting wrapper hops. Then let the neutral vocabulary reach callers—rename the small Android call sites/tests to `foldSnapshotRect` / `serializeRegularSnapshotPresentationNode` and move the pure geometry property proof into contracts. Finally split the moved mixed-layer tests: presentation units stay with presentation; tests importing daemon assembly or command-runtime helpers stay with those owners. The new tracked-path gate is too shallow to substitute for that topology. Keep `CONTEXT.md` glossary-only; the Swift/host boundary decision belongs in ADR 0004.

thymikee:
Clean at `1d0306a3`. The prior ownership blockers are resolved: contract exports point directly to source, identity facades are gone, Android uses neutral shared names, pure geometry proof lives with contracts, daemon assembly tests are back under the daemon owner, and the boundary gate checks real import edges. The production path preserves behavior while platform policy stays local. Code is ready for human review; iOS smoke is still queued.

PR #2006, opened 2026-08-24
fix(ci): make the two rg-based static checks actually run

### LINKED ISSUE #1976: ci(static-checks): both rg assertions are false-green — ripgrep is never installed

thymikee:
## Summary

- `ripgrep` is never installed on `ubuntu-latest`, so both `rg` assertions in the `Lint & Format` job (`Disallow trailing commas before closing parenthesis in Swift` and `Fail if test-only DI seams reappear in production code`) failed with `rg: command not found` (exit 127) on every run. `if rg ...; then ... fi` cannot distinguish "command not found" from "no matches" (exit 1) — both read as false, so each step silently passed without its assertion ever executing.
- Rewrote both against `grep`, which every runner ships, with match/no-match/error exit codes handled explicitly (0 = violation found → fail, 1 = clean → pass, anything else = broken scan → fail loudly instead of reading as a pass) plus a zero-tracked-files guard so a future renamed/deleted directory can't silently make the check a no-op.
- Actually running the DI-seams check surfaced 7 live matches on `main`, none of which are the test-only DI seams the rule was written to ban, so the pattern is narrowed to drop both false-positive classes:
  - `typeof fetch`: the `fetchImpl?`/`fetch?` seams in `src/cli/auth-session.ts`, `src/cli/connection/cloud-profile.ts`, and `src/remote/daemon-proxy.ts` inject the one global with no module boundary `vi.mock` can intercept. Their own unit tests exercise the seam directly for exact per-call assertions, while the CLI-level tests (`cloud-connect-auth.test.ts`, `cloud-connect-profile.test.ts`) use `vi.stubGlobal('fetch', ...)` at a layer where the seam isn't reachable — a deliberate, exercised seam, not a leftover one.
  - `typeof SOME_CONSTANT` in `SCREAMING_SNAKE_CASE`: e.g. `dispatchPath?: typeof MAESTRO_COORDINATE_FALLBACK_PATH` in `src/daemon/handlers/interaction-touch-response.ts` derives a literal union type from a constant — not an injectable seam at all, just a syntax coincidence the old pattern happened to match.

Verified the rewritten scripts against the real repo tree and against synthetic injected violations (both a real trailing-comma-before-`)` Swift case and a real `dispatch?: typeof someFn` seam), executed with the exact `bash --noprofile --norc -eo pipefail` invocation GitHub Actions uses for `run:` steps, confirming both checks pass on the current tree and correctly fail with diagnostic output when a genuine violation is present.

Fixes #1976

## Test plan

- [x] Extracted both rewritten `run:` blocks from the workflow and executed them locally with `bash --noprofile --norc -eo pipefail` (GitHub Actions' actual bash invocation) against the current repo tree — both pass (exit 0).
- [x] Injected a real multiline trailing-comma-before-`)` violation into a tracked `.swift` file — the Swift check fails (exit 1) with the offending file and matched text, then confirmed the file was restored and the tree is clean.
- [x] Injected a real `dispatch?: typeof someFn`-shaped DI seam into a tracked `src/` file — the DI-seams check fails (exit 1) with file/line, then confirmed the file was restored and the tree is clean.
- [x] Confirmed the narrowed DI-seam pattern still flags `fetchThing?: typeof fetchThingImpl` (not excluded by the `fetch`-specific lookahead) while excluding all 7 current `typeof fetch` / `typeof MAESTRO_COORDINATE_FALLBACK_PATH` matches on `main`.
- [x] Validated `.github/workflows/ci.yml` parses as valid YAML.
- [x] Ran an adversarial code review pass over the diff (see `/code-review` in the session); no correctness issues found — the one findings besides a minor diagnostics nit (already addressed: switched Swift-check output from filenames-only to filename+matched-text) were left as-is as a deliberate, low-severity styling call.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Vf8QB21gvH9tj6xN1sZFG5)_


thymikee:
Not ready. The scans now fail closed correctly, but the DI rule replaces one false-green with a name-based semantic allowlist: any optional `typeof fetch` seam now passes, whether test-only or not, while an equivalent legitimate global seam under another name fails. The paragraph explaining the exceptions is evidence the regex does not own this invariant. Keep the scanner/exit-code fix, but replace this with an explicit typed/lint ownership rule or retire the semantic gate; do not encode intent through symbol spelling.

thymikee:
Fair point — the name-based exemption was exactly the failure mode you describe: an unreviewed `typeof fetch` anywhere would have passed, and an equally legitimate seam under any other name would have failed. That's not ownership, it's spelling.

Pushed 14d1c8d, which replaces it with `scripts/di-seams`: a small, tested TypeScript checker (`model.ts` + `model.test.ts`) that judges each `field?: typeof X` match against an explicit, typed allowlist (`approved.ts`) keyed by `(file, field name, typeof target)` — not by name. A match is exempt only because that exact site was individually reviewed and listed; a new `typeof fetch` anywhere else still fails. The gate also fails just as hard on a stale approval (an entry whose triple no longer matches anything, e.g. after a rename) as on an unapproved seam, so the list can't quietly drift out of sync with the tree.

Tests exercise the specific scenario you flagged directly: the same field/target seam is approved in one file and flagged as a violation in another; an unapproved field name is flagged even when its target is approved; an unapproved target is flagged even when its field name is approved.

Since it's a real TypeScript checker now rather than a toolchain-free grep one-liner, I moved the step to run after `Setup toolchain` (Swift check stays where it was — still pure grep, no runtime needed).

---

thymikee:
Still blocked at `1978ebca`. The name-based exemption is gone and scan errors now fail correctly, but `findSeamMatches()` splits source by line before matching, so valid multiline declarations such as `dispatch?:` followed by `  typeof dispatchCommand` are invisible. Also, approval identity is only `(file, field, target)`, so a second identical seam anywhere in an approved file passes; the test currently blesses duplicate occurrences. Use a syntax-aware scan, or full-source matching with location-specific approval identity, and plant both multiline and second-site regressions.

thymikee:
Both real gaps. Pushed 8d4eac1.

**Multiline:** `findSeamMatches` now matches against each file's whole source in one pass instead of splitting into lines first — `\s` matches a literal newline in a JS regex with no extra flag, so `field?:` and `typeof X` landing on different lines is no longer invisible. Line number is derived from the match's character offset.

**Duplicate site:** approval identity now includes the line the match starts on — `(file, line, field, target)` instead of `(file, field, target)`. Approving one `fetchImpl?: typeof fetch` no longer blesses a second, unreviewed one elsewhere in the same file; `approved.ts` expands from 5 collapsed entries to the 7 actual sites.

Added tests planting exactly the two scenarios you described (a cross-line declaration, and a second `fetchImpl?: typeof fetch` at a different line in an already-approved file) and separately verified both against the real tree with injected violations, confirming they're caught and the files restore cleanly. Re-ran the full local gate suite (di-seams, gate-manifest, layering, fallow, format, lint, typecheck) — all green.

---

thymikee:
Clean at `8d4eac15`. The scanner now matches full-source multiline declarations, derives the match-start line, and approvals identify exact sites rather than recurring `(file, field, target)` triples. Both planted regressions are non-vacuous: a split declaration is detected, and a second identical seam is rejected unless separately approved. Code is ready for human review while the restarted CI finishes.

thymikee:
Readiness retracted at `8d4eac15`: exact-head CI immediately proved line-number approvals are a brittle workaround. The unchanged `dispatchPath` seam moved from line 61 to 60 through unrelated base movement, so the gate rejected an already-reviewed declaration. Remove the global positional table. Use a concise code-local approval marker associated with the immediately following declaration—or a dedicated owner type—and parse that association with the repo’s existing `oxc-parser`. A second seam still fails without its own marker, multiline coverage remains, and unrelated line shifts stop breaking CI.

thymikee:
Still blocked at `af8287b5`. This delta only changes the approval from line 61 to line 60; it does not replace positional identity. Treating unrelated line movement as intentional staleness confirms the maintenance bug rather than fixing it. Keep the multiline and second-site tests, but move approval to a stable code-local marker associated with the declaration—or a dedicated owner type—so unrelated edits cannot make CI red.

thymikee:
Agreed — a line number was never a real identity. Pushed ce5fbcd.

`approved.ts` is deleted. Detection is now AST-based via `oxc-parser` (same tool `scripts/layering/*.ts` already uses): any node shaped `{ optional: true, typeAnnotation: TSTypeQuery }` — a property signature or a bare parameter, so both `{ field?: typeof X }` and `function f(field?: typeof X)` — is a candidate, which gets multiline declarations for free instead of needing a special case. Approval is a `// di-seam-approved: <reason>` comment immediately above the declaration, the same "marker precedes what it exempts" shape as this repo's own `// fallow-ignore-next-line complexity`. Nothing external to resync: the approval travels with the code. Added the marker to the 7 real sites (fetch-global seams in auth-session.ts/cloud-profile.ts/daemon-proxy.ts; the literal-type false positive in interaction-touch-response.ts).

Verified: a second, unmarked occurrence of an approved field/target pair still fails; an unrelated multi-line insertion above an approved declaration — the exact shape of what broke last round — leaves it passing (tested against the real tree by injecting one, confirming green, and restoring cleanly). Multiline and cross-file regression tests from the last round still pass. Re-ran the full local gate suite (di-seams, gate-manifest, layering, fallow, format, lint, typecheck, auth-session unit tests) — all green.

---

thymikee:
The code-local AST marker fixes the line-drift design at `ce5fbcd0`, but one guard bypass remains: bare `// di-seam-approved:` (or whitespace-only reason) returns ``, and `checkSeams()` accepts it because it only rejects `null`. Require a non-empty trimmed reason and add a planted regression proving empty markers fail. Checks are still running.

thymikee:
Real gap. Pushed 135a144.

`approvalReason()` returned `''` (not `null`) for a bare `// di-seam-approved:` with nothing after it, and `checkSeams()` only filtered `null` — so an empty marker silently approved a seam with zero justification. Now `approvalReason()` returns `null` whenever the joined reason text is empty after trimming, treating a bare or whitespace-only marker the same as no marker at all.

Added both a model-level test (`findSeamMatches rejects a di-seam-approved marker with no reason text`, plus a whitespace-only variant) and an end-to-end `checkSeams` test, then verified against the real tree by injecting a bare-marker declaration into `daemon-proxy.ts` — flagged as expected, file restored cleanly.

Also merged `main` (one more PR landed, a snapshot-presentation file move — unrelated) and re-ran the full local gate suite (di-seams, gate-manifest, format, lint, typecheck) — all green.

---

thymikee:
Ready for human review at `135a144b`. Empty and whitespace-only `di-seam-approved` markers now resolve to violations, with regression coverage that fails if the non-empty requirement is removed. Core checks are green; Android/iOS smoke and coverage reporting are still finishing.

Issue #1414, opened 2026-07-27
test: nightly fuzz lane — parser inputs must fail as typed AppErrors, never hang


thymikee:
Part of #1412 (Track A).

## Context
Agents feed the CLI hostile strings constantly (half-quoted labels, emoji selectors, malformed batch JSON, odd Maestro YAML). Parsers are the front door; today nothing systematically asserts they fail *well*.

## Deliverables
- A fuzz harness (fast-check generators or a simple corpus+mutator) over: `parseArgs` (strict flags), selector parsing, `.ad` replay-script parsing (`parseReplayScriptDetailed`), `batch --steps` JSON, and the Maestro compat parser.
- Single invariant: every rejection is a typed `AppError` with a non-empty hint; no other throw types; no hang (per-case timeout).
- Nightly workflow job (extend an existing nightly yml); on failure it uploads the failing input as an artifact and prints a one-line repro command.
- A tiny checked-in regression corpus: every input the fuzzer ever caught gets appended and replayed in the unit lane.

## Acceptance
- Nightly job green on current main; unit-lane corpus replay deterministic and fast.
- A deliberately introduced `throw new Error` in a parser branch is caught by the harness locally.

## Out of scope
Fuzzing daemon HTTP or subprocess layers.

## PR-wave review outcome (2026-08-01, PR #1438)

Approve-with-changes; direction validated (the lane already caught a real production bug: raw SyntaxError from `readQuotedReplayToken`, fixed + pinned). Required rework, now part of this spec:
- Strip the bundled #1430 watcher (consolidation happens on #1430).
- Rebase onto main; the generator layer must be fast-check arbitraries sharing ONE hazard vocabulary with `src/__tests__/test-utils/property-arbitraries.ts` — no parallel PRNG/mutator/hazard systems, and shrinking must work (a 20k-char unminimized failing input is not acceptable output).
- The worker-thread + SharedArrayBuffer watchdog stays — it is the one piece fast-check cannot provide (interrupting synchronous hangs).
- Subprocess-spawning harness tests go in the serialized `subprocess-stub` vitest project, never `unit-core`.
- Size calibration: the #1414-scoped lane lands near ~1,000 LOC (watchdog/envelope/corpus plumbing included), not the ~300 originally imagined.

PR #2007, opened 2026-08-24
chore(test): start the subprocess-stub kill-criterion experiment (#1823)

thymikee:
## Summary

#1823 is the live tracker for the `subprocess-stub` Vitest project and states its own kill criterion: *"if the three members run un-serialized in the default pool for 20 consecutive CI runs with no timeout-shaped failure, delete the project."* Nobody had started that experiment yet — the project still existed with all its serialization machinery. This PR runs exactly the steps the issue describes: drops `SUBPROCESS_STUB_TESTS` from `unit-core`'s `exclude` and deletes the `subprocess-stub` project block, so its members run in the default forks pool.

**Membership note:** the issue's table lists 3 members, but `scripts/fuzz/corpus-replay.test.ts` already left for its own `fuzz-worker` project in #1994 (merged earlier today, for an unrelated coverage-lane worker-death bug — see #1824). So only 2 files are actually un-serialized by this change:

- `src/__tests__/client-metro.test.ts` — real Metro dev server + child spawn per case
- `scripts/fuzz/harness.test.ts` — the SUT *is* the subprocess watchdog, one target hangs on purpose (#1414)

`fuzz-worker` (holding `corpus-replay.test.ts`) is untouched — its serialization is for a different, already-decided reason.

## What's the same, what changed

- **Mutation lane:** `SUBPROCESS_STUB_TESTS` still feeds `SERIALIZED_TESTS`, so both files stay excluded from the Stryker mutation scope regardless of this experiment's outcome — thousands of mutant reruns times a real per-case spawn is timeout noise independent of whether Vitest itself serializes the file.
- **Coverage lane:** unaffected — `test:coverage:ci` already only skips `fuzz-worker`, not `subprocess-stub`.
- **`package.json` / `docs/agents/*.md` / assorted comments:** updated the handful of places that named the `subprocess-stub` project directly, so nothing left in the repo asserts a project that no longer exists. `scripts/gate/check.ts` needed no change — it derives `vitestProjects` from `vitest.config.ts` by regex, so the gate manifest picked up the deletion automatically (verified below).

## Revert plan

If a timeout-shaped failure shows up in either file before 20 consecutive CI runs pass clean, restore the `subprocess-stub` project block and put `SUBPROCESS_STUB_TESTS` back in `unit-core`'s `exclude` — that's the entirety of the revert, symmetric with this diff.

## Validation

- `pnpm exec vitest run --project unit-core src/__tests__/client-metro.test.ts scripts/fuzz/harness.test.ts` — both files pass running alongside the rest of `unit-core` (previously they only ran alone in `subprocess-stub`).
- `pnpm exec vitest run --project unit-core --project fuzz-worker` — full suite: 1035 files, 2 pre-existing failures unrelated to this change (`app-log-session-resource.test.ts`, `durable-capture-resource-adoption.test.ts`) reproduced identically against `origin/main` before this diff — not introduced here.
- `pnpm check:gate-manifest` — `48 checks wired across 28 lanes` (down from the prior run's `subprocess-stub`-inclusive count), confirming the manifest re-derives cleanly with no manual sync needed.
- `pnpm check:gate-manifest:test`, `pnpm check:agent-guidance`, `pnpm check:layering`, `pnpm typecheck`, `pnpm lint`, `pnpm format:check` — all green.
- `docs/agents/testing.md` stays at 9,587 of its 10,000-byte gate ceiling (`agent-guidance-contract.test.ts`).

Refs #1823.

---
_Generated by [Claude Code](https://claude.ai/code/session_015YPgKE1xmjdqh7T1q987DA)_


thymikee:
Clean at `ca14f2ad`: this implements #1823’s prescribed experiment exactly. The two remaining real-subprocess tests now run in `unit-core`, `corpus-replay` correctly remains in the separately justified `fuzz-worker` project, and mutation exclusion remains independent. Code is ready for human review; the remaining iOS check is still running.

PR #2008, opened 2026-08-24
test(daemon): harden the timed-out wait decoration test against CI load

thymikee:
## Summary

- `src/daemon/__tests__/wait-runtime.test.ts`'s `'a timed-out wait decorates its failure through the same single binding'` used a 1ms real-clock `wait` timeout, which races the first capture's resolution: on a loaded CI host (e.g. CI Coverage shard 2, run 32752619090 on PR #2000, head `30bd80967`) the capture can take longer than 1ms wall-clock to resolve, so `runWithinWaitDeadline` classifies the poll `capture-stalled`/`capture-truncated` instead of a completed target-absent poll, and `maybeWaitTimeoutSurfaceResponse` (`src/daemon/wait-current-surface.ts`) skips the "Current surface" decoration outright — failing the `toContain('Current surface: Checkout')` assertion.
- Fixed by driving the test on a fake clock (`vi.useFakeTimers()` + `vi.advanceTimersByTimeAsync`), the same pattern `runFill` uses in `packages/provider-webdriver/src/webdriver-interactor.test.ts`, per `docs/agents/testing.md`'s rule that production time must never be real time in tests. Virtual time now only advances when the test says so, so the mocked (instant, non-timer) capture always resolves before the deadline timer is due — removing the race entirely, regardless of host load.
- The assertion itself is unchanged.

## Regression evidence

- Reproduced the exact CI failure locally by pegging all CPU cores (`yes > /dev/null` × nproc) during repeated runs of the *unfixed* test — it failed with the identical assertion error (`expected 'wait timed out for text: Ready' to contain 'Current surface: Checkout'`) multiple times out of 8 runs.
- Applied the fix and re-ran under the same simulated load: 10/10 passes.
- Full file (`wait-runtime.test.ts`, 19 tests) and the sibling `wait-current-surface.test.ts` pass without load.

## Test plan

- [x] `pnpm exec vitest run --project unit-core src/daemon/__tests__/wait-runtime.test.ts src/daemon/wait-current-surface.test.ts` — all pass
- [x] Reproduced the flake on the pre-fix test under simulated CPU load, then confirmed the fix eliminates it (10/10 passes under the same load)
- [x] `pnpm check:affected --run`

---
_Generated by [Claude Code](https://claude.ai/code/session_01Q1Ps2n6f5yA9hzEo3AgDFQ)_

### DIFF
--- a/src/daemon/__tests__/wait-runtime.test.ts
+++ b/src/daemon/__tests__/wait-runtime.test.ts
@@ -308,19 +308,44 @@ test('a provider owner that cannot capture fails closed instead of borrowing the
 // the SAME binding rather than reaching a second capture owner.
 // ---------------------------------------------------------------------------
 
+/**
+ * The 1ms timeout races the real clock: on a loaded CI host the first capture can take longer
+ * than 1ms to resolve, so `runWithinWaitDeadline` classifies it `capture-stalled` (or, once a
+ * later poll is readable, `capture-truncated`) instead of a completed target-absent poll, and
+ * `maybeWaitTimeoutSurfaceResponse` skips the decoration outright (see wait-current-surface.ts).
+ * A fake clock removes the race the same way `runFill` does in
+ * packages/provider-webdriver/src/webdriver-interactor.test.ts: virtual time only advances when
+ * `advanceTimersByTimeAsync` says so, so the mocked (instant, non-timer) capture always resolves
+ * before the deadline timer is due, on any host (docs/agents/testing.md: production time must
+ * never be real time in tests).
+ */
 test('a timed-out wait decorates its failure through the same single binding', async () => {
-  const harness = waitRuntimeHarness({
-    nodesPerPoll: [[{ index: 0, depth: 0, type: 'Button', label: 'Checkout', hittable: true }]],
-  });
+  vi.useFakeTimers();
+  try {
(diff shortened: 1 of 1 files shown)


thymikee:
Not ready yet. The fake-clock direction is correct, but this simple regression now needs an 11-line justification plus a settled-result wrapper solely to pre-handle an unexpected rejection. Keep the fake timer, start `runWait`, advance virtual time, await it, and restore timers; the test name and existing assertions already state the invariant. This should be a small deterministic test, not a paragraph-justified workaround.

thymikee:
Simplified in f0f132f: dropped the justification comment and the settled-result wrapper. `runWait` can't reject in this scenario (`toDaemonResponse` catches the whole capture path), so the test is now just `useFakeTimers` → start `runWait` → `advanceTimersByTimeAsync` → `await` it → `useRealTimers`, same shape as `runFill`. Re-verified 8/8 passes under pegged-CPU load.

---

thymikee:
Clean at `f0f132ff`. The previous fake-clock ceremony finding is fully resolved: the long justification and settled-result wrapper are gone, while the 1 ms wait, 5 s virtual advance, assertions, and `finally` timer restoration are unchanged. No new code findings; code is ready for human review while the remaining checks finish.

Issue #1349, opened 2026-07-20
Design read-only identity verification without breaking wait polling


thymikee:
## Problem

ADR 0012 target-binding evidence and replay-time refusal do not yet cover every selector-targeted read-only or synchronization step.

A destination guard in ADR 0016 is a polling `wait <selector>`. It currently proves selector existence, not that the resolved landmark has the same identity observed during authoring. A reshuffled screen containing the same label elsewhere can false-pass.

The existing ADR 0012 mechanism cannot simply be attached to `wait`: it verifies `target-v1` before dispatch. A wait commonly begins while its landmark is absent, so pre-action verification would reject the expected starting condition instead of polling for it.

`get` already passes recording-time node/tree evidence on its recording-aware snapshot path. This work should preserve that behavior while defining compatible semantics for waits and the remaining eligible read-only steps.

## Desired outcome

Design and implement recorded identity verification for eligible read-only steps without changing their execution semantics. For `wait`, selector polling must happen first; any recorded-landmark identity comparison happens only after a candidate resolves and before the wait reports success, or uses a distinct guard-specific mechanism with the same fail-closed outcome.

## Acceptance criteria

- `wait <selector>` continues polling while no candidate exists; identity verification never performs an eager pre-action refusal merely because the landmark has not appeared yet.
- When an annotated wait resolves a candidate, an identity mismatch/unverifiable result fails closed through the bounded `REPLAY_DIVERGENCE` contract before the wait reports success.
- The design explicitly chooses whether wait evidence can reuse `target-v1` with wait-specific dispatch or requires a distinct bounded annotation; it must not accidentally enter the existing generic pre-action path.
- Duration waits and `wait stable` remain unannotated because they have no element target.
- `wait @ref` is not treated as a portable destination guard; ADR 0016 publication accepts only a selector or selector chain.
- Eligible `is` and read-only `find` variants are explicitly classified as covered or intentionally deferred, with tests for each decision.
- Existing `get` identity behavior remains intact and shares ADR 0012's verification machinery where its immediate-read semantics permit pre-action verification.
- ADR 0016 gains a reshuffled-screen false-pass regression when wait identity verification ships.

## Relationship

- Follow-up to ADR 0012's bounded target identity model.
- Strengthens ADR 0016 / #1346 destination guards from selector existence to recorded landmark identity without breaking polling.
- Corresponds to the replay-study roadmap item “extend identity annotations to read-only steps.”

thymikee:
[...]

PR #2009, opened 2026-08-24
perf(selectors): collapse the double tree scan for disambiguate/fail-closed rows

### LINKED ISSUE #1970: selectors: resolveSelectorChainWithPolicy still double-scans the tree for disambiguate/fail-closed rows
Found during adversarial review of #1690's implementation (find index-once).

`packages/selectors/src/internal/resolve-with-policy.ts:66-89` runs `listSelectorChainMatches` (full scan per alternative) followed by `resolveSelectorChain` (another full scan per alternative) for the `disambiguate` and `fail-closed` rows — the same defect class #1690 removed from replay's `resolveRecordedTarget`. Measured: `readText`/`readUnique` rows pay 2 whole-tree passes; the comment at `:61` ("One matching pass serves every row") is false for those rows.

The new `resolveSelectorChainDomain` primitive (added by #1690) could collapse them, but **not mechanically**: `resolveSelectorChainWithPolicy` deliberately reports the FIRST matching alternative's `matchedNodes` even when the winner comes from a later alternative (`:79-81`), whereas `resolveSelectorChainDomain` reports the winner's. Folding changes `matchedNodes` for chained selectors and needs an explicit semantic decision first.

Also related: `analyzeSelectorMatches`'s lazy `isVisible` (`packages/selectors/src/internal/resolve.ts:215-218`) omits `precomputedViewportRects`, so each ambiguous candidate pays 2 extra whole-tree `flatMap` passes inside visibility resolution (4n scans per candidate measured). Passing the precomputed rects — or building them once next to the lazily-built `byIndex` — removes it.

thymikee:
## Summary

`resolveSelectorChainWithPolicy` ran two full scans of the snapshot tree for the `disambiguate`/`fail-closed` policy rows (`readText`/`readUnique`, used by `get text`, `get attrs`, and `is` non-`exists` predicates): `listSelectorChainMatches` to compute the "first alternative that matched anything" (`matchedNodes`), then `resolveSelectorChain` again to compute the winning resolution — the same defect class #1690 already removed from replay's `resolveRecordedTarget`. `first-match`/`reject-candidates` rows were already single-pass and are untouched.

- `resolveSelectorChainDomain` (`packages/selectors/src/internal/resolve.ts`) now tracks `firstMatch` — the first alternative that matched anything, with its selector/index/matched-node set — unconditionally, in the same per-alternative loop that already decides the winning resolution. `resolveSelectorChainWithPolicy` now makes one call to it instead of two.
- The pre-existing semantic where `matchedNodes` names the *first* matching alternative rather than the *winning* one (when uniqueness skips a tied/ambiguous alternative to resolve from a later one — the exact case the issue flagged as "not mechanically foldable" with `resolveSelectorChainDomain`'s own `matchedNodes`, since that field reports the *winner's* domain) is preserved byte-for-byte: `wait`'s #1349 landmark check and the ambiguous outcome both depend on it.
- Also fixes the related `analyzeSelectorMatches` inefficiency: its lazily-built `isVisible` closure now builds the viewport-root rect list once per alternative (via a newly extracted `collectViewportRects` in `packages/contracts/src/snapshot-visibility.ts`) alongside the already-lazy `byIndex` map, instead of `isNodeVisibleOnScreen` re-deriving it via two whole-tree `flatMap` passes on every ambiguous candidate it's asked about.

No output shape or behavior changes for any caller — this is a pure perf/scan-count fix.

Closes #1970

## Validation

- `pnpm build`, `tsc --noEmit`, `pnpm lint`, `pnpm check:layering`, `pnpm check:production-exports`, `pnpm check:fallow --base origin/main` all clean.
- `pnpm check:affected --run`'s `vitest-related` lane surfaced only 2 pre-existing failures (`app-log-session-resource.test.ts`, `durable-capture-resource-adoption.test.ts`) — confirmed present and unrelated on a clean checkout before this change.
- Regression evidence per repo convention: extended the existing scan-count gate (`packages/selectors/src/index.test.ts`, the same `observeTreeTraversals` proxy #1988 introduced) with a new test asserting one matching pass per alternative for `readText`/`readUnique`. Reverting the `resolve-with-policy.ts` change reproduces the double-scan (`filter: 1`, the redundant pass) — observed red, restored green.
- The viewport-rect fix reduces the pre-existing pinned `flatMap` count on that same gate's ambiguous leg from 8 to 5 (the residual 5 is a separate, out-of-scope fallback scan in `resolveViewportRect` that fires only when the tree has no Application/Window root at all, documented inline).
- New unit coverage: `resolveSelectorChainDomain`'s `firstMatch` field (`packages/selectors/src/internal/resolve.test.ts`), `collectViewportRects`' equivalence with the previous inline scan (`packages/contracts/src/snapshot.test.ts`).

---
_Generated by [Claude Code](https://claude.ai/code/session_01YJoiggu7utUNDBmdzSBK2h)_

### DIFF
--- a/packages/contracts/src/facades/snapshot.ts
+++ b/packages/contracts/src/facades/snapshot.ts
@@ -12,6 +12,7 @@ export {
   findSnapshotAncestor,
 } from '../snapshot-tree.ts';
 export {
+  collectViewportRects,
   findNearestScrollableAncestor,
   isNodeVisibleInEffectiveViewport,
   isNodeVisibleOnScreen,
--- a/packages/contracts/src/snapshot-visibility.ts
+++ b/packages/contracts/src/snapshot-visibility.ts
@@ -42,6 +42,20 @@ export function isViewportRootNode(node: Pick<SnapshotNode, 'type' | 'role' | 's
 }
 
 /**
+ * The Application/Window rects a target's viewport resolves against — the
+ * whole-tree scan `resolveViewportRect` does per call when no
+ * `precomputedViewportRects` is given. A caller resolving visibility for
+ * several nodes against the SAME tree (e.g. ranking ambiguous selector
+ * candidates) should build this once and thread it through, so N candidates
+ * share one pass instead of each paying its own (#1970).
+ */
+export function collectViewportRects(nodes: RawSnapshotNode[]): Rect[] {
+  return nodes.flatMap((node) =>
+    isViewportRootNode(node) && hasValidRect(node.rect) ? [node.rect] : [],
+  );
+}
+
+/**
  * The root viewport a target rect is measured against: the largest
  * Application/Window rect containing the target's center, falling back to the
  * largest such rect, then to the largest containing rect of any node.
@@ -52,11 +66,7 @@ export function resolveViewportRect(
   precomputedViewportRects?: readonly Rect[],
 ): Rect | null {
   const targetCenter = centerOfRect(targetRect);
-  const viewportRects =
-    precomputedViewportRects ??
-    nodes.flatMap((node) =>
-      isViewportRootNode(node) && hasValidRect(node.rect) ? [node.rect] : [],
-    );
+  const viewportRects = precomputedViewportRects ?? collectViewportRects(nodes);
   const contains = (rect: Rect) => containsPoint(rect, targetCenter.x, targetCenter.y);
   const viewport =
     pickLargestRect(viewportRects.filter(contains)) ?? pickLargestRect(viewportRects);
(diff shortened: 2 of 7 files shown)


thymikee:
Clean at `698f049c`. The `firstMatch` result preserves the first matching selector alternative while uniqueness may resolve a later one, so the second scan is removed without changing `matchedNodes` semantics. The planted scan-count regression is non-vacuous, and viewport rects are collected lazily once per alternative. Code is ready for human review; iOS smoke is still running.

Issue #1295, opened 2026-07-16
Maestro compat: assertTrue phase 1 — literal/`${VAR}`-lookup truthiness (expressions gated on #1292)


thymikee:
## Context

`assertTrue` is entirely unsupported. Full upstream parity means evaluating arbitrary JS in the assertion (GraalJS) — that's gated on the #1292 decision. But there's a cheap, genuinely useful phase 1 that does NOT require it:

**Phase 1 scope: `assertTrue: ${VAR}` (variable lookup) and literals, evaluated with JS truthiness on the resolved value.** This covers the common CI pattern: `runScript` sets `output.flag` → `assertTrue: ${output.flag}`. Upstream truthiness reference: their `065_when_true.yaml` exercises `${undefined}`/`${null}`/`${0}` falsy, `${123}`/`${'String'}`/objects/arrays truthy — phase 1 needs the same truthiness table for looked-up values (which arrive as strings post-lookup; define and test the string-value truthiness rules explicitly: `''`/`'false'`? — pin to upstream behavior for interpolated string results, add a layer-2-style vector or corpus flow for it).

JS expressions inside `assertTrue` keep failing loud, referencing #1292.

## Repro

- Corpus flow: `scripts/maestro-conformance/corpus/upstream/067_assertTrue_pass.yaml`
- Declared divergences: `FLOW_DIVERGENCES['upstream/067_assertTrue_pass']` and the `assertTrue` item in `FLOW_DIVERGENCES['upstream/076_optional_assertion']`

## Acceptance

- `assertTrue` with literal/lookup values matches upstream pass/fail on the corpus flow (067 entry removed or narrowed to expression-only, referencing #1292); conformance green with no new divergences.
- `optional: true` composes (assertTrue failure downgrades to warning when marked optional) once #1291 lands.

thymikee:
Valid bounded enhancement, but not `ready-for-agent` yet because the current acceptance conflicts with the settled #1292 decision and leaves the core semantic vector undefined.

`upstream/067_assertTrue_pass.yaml` is `${1+1}`—an expression. #1292 explicitly chose lookup-only interpolation with loud expression failure, so phase 1 must not remove or narrow that divergence as though 067 were covered. Add a dedicated corpus/vector for bare `${VAR}` lookup and scalar literals; keep 067 attributed to the documented expression deviation.

[...]

PR #2010, opened 2026-08-24
feat(maestro): support assertTrue phase 1 - literal/${VAR} truthiness (#1295)

### LINKED ISSUE #1292: Decision: should Maestro `${}` interpolation evaluate JS expressions (GraalJS parity) or stay lookup-only?
## Decision to make

Should Maestro-compat `${...}` interpolation evaluate JavaScript **expressions** (GraalJS parity), or remain variable-lookup-only with loud failure?

## Context

The typed engine (ADR 0015) deliberately implements `${}` as literal variable lookup: `${VAR}` resolves from flow config/env/runFlow env/runScript `output.*`; anything unresolved throws `INVALID_ARGS` (fail-loud — a deliberate #1217 decision replacing an earlier silent pass-through). Upstream evaluates the full `${...}` payload as a GraalJS expression: arithmetic, ternaries, string concat, `maestro.*` namespace, JS truthiness.

The conformance oracle (#1289) now names exactly where this bites real flows:

- `${0 + 1}` as `tapOn.index` (corpus `upstream/032_element_index`)
- full `assertTrue` parity (corpus `upstream/067_assertTrue_pass`)
- `repeat.while` / `runFlow.when.true` beyond platform comparisons
- upstream's own `065_when_true.yaml` exercises `${undefined}`, `${0}`, `${[]}`, `${{...}}` as truthy conditions

Lookup-scoped subsets are being handled separately and do NOT require this decision: numeric-field `${VAR}` lookup, and `assertTrue: ${VAR}` truthiness (phase 1).

## Options

1. **Stay lookup-only (status quo), document as a permanent deviation** in ADR 0015 + the oracle's `DOCUMENTED_DEVIATIONS`. Cheapest; keeps the engine free of a JS-eval surface; flows using expressions keep failing loud with source lines.
2. **Bounded expression support** via the existing Node `vm` sandbox used by `runScript` (arithmetic/comparison/truthiness, no `maestro.*` host API). Middle ground; needs a defined subset and its own conformance vectors.
3. **Full GraalJS-semantics parity.** Largest surface; effectively an embedded expression engine with upstream-matching semantics; only worth it if real suites demand it.

Whatever is decided, encode it: option 1 → `DOCUMENTED_DEVIATIONS` entry; options 2/3 → corpus flows + generated vectors so the subset boundary is mechanical, not prose.

thymikee:
## Summary

- Adds the `assertTrue` Maestro command, scoped to literal values and bare `${VAR}` lookups only (JS expressions stay unsupported per the closed #1292 decision — the parser rejects any `${...}` payload that isn't an exact bare-name match, at parse time, with a `runScript` → `${output.x}` hint).
- Truthiness on a resolved value is evaluated against a pinned falsy-string table (`""`, `"false"`, `"0"`, `"null"`, `"undefined"`, case-sensitive) rather than native JS truthiness, since flow config/env/`runScript` output values are always stored as strings in `MaestroExecutionContext` — native truthiness would make the string `"false"` truthy, which is almost certainly not what a flow author means.
- Wires `assertTrue` through the parser, interpreter, the existing generic `optional: true` → warning composition (no new machinery needed), the runtime-port command-dispatch tables, and the layer-1 conformance oracle's canonical projection.
- Narrows the `067_assertTrue_pass` conformance divergence to the JS-expression case (still correctly rejected) and removes the now-satisfied `076_optional_assertion` divergence entry entirely, since every command in that flow is now supported.
- Along the way, fixes a latent conformance gap: `scrollUntilVisible`'s canonical projection wasn't materializing upstream's default `direction: down`, which only became visible once `assertTrue` support let flow `076` reach full comparison.

## Testing

- `pnpm exec tsc -b packages/kernel packages/contracts packages/selectors packages/maestro` — clean.
- `pnpm exec vitest run packages/maestro` — 211 tests pass (new: parser tests for scalar/lookup/map-form/optional/label, expression rejection, missing-condition rejection; interpreter tests for pass/fail/falsy-string-lookup/optional composition; a dedicated truthiness-table test file).
- `pnpm maestro:conformance` — 47/47 pass, no new undeclared divergences.
- `pnpm exec vitest run src/cli/parser/__tests__/maestro-support-matrix.test.ts scripts/fuzz/validation-arbitraries-maestro.test.ts` — pass (fuzz corpus updated: `assertTrue` moved from the fake/unsupported command list to a valid-command case).
- Ran an independent adversarial review pass, which caught one real issue (fixed in a follow-up commit): the CLI help/docs claimed assertTrue is "evaluated with JS truthiness," which is the opposite of the actual pinned-string-table behavior — corrected to spell out the real falsy/truthy rule.

## Test plan

- [x] Typecheck
- [x] Maestro unit tests
- [x] Maestro conformance suite
- [x] CLI help / docs sync test
- [x] Fuzz corpus coverage test
- [x] Adversarial review pass

---
_Generated by [Claude Code](https://claude.ai/code/session_01MsAta9qHbsC2uqiK1sy6To)_

### DIFF
--- a/docs/adr/0015-direct-maestro-engine.md
+++ b/docs/adr/0015-direct-maestro-engine.md
@@ -190,6 +190,11 @@ two-pointer plans, executor selection, and app-observable effects must remain un
 
 - agent-device supports selected Maestro Flow syntax and behavior. Unsupported features return
   explicit errors, and intentional differences are declared in the conformance fixtures.
+- `${...}` interpolation stays variable-lookup-only, by decision (#1292): upstream's GraalJS
+  expression evaluation is not reimplemented, so unresolved or expression-shaped payloads fail
+  loud with source context; the escape hatch is `runScript` (compute) → `${output.x}` (consume).
+  `assertTrue` (#1295) is scoped to that same lookup-only subset — literal values and bare
+  `${VAR}` lookups evaluated with a pinned string-truthiness table — for the same reason.
 - Shipping two production engines or a runtime fallback between them is rejected because it doubles
   semantic and performance ownership.
 - Source provenance and runtime values stay typed through execution.
--- a/packages/maestro/src/internal/__tests__/engine-truthiness.test.ts
+++ b/packages/maestro/src/internal/__tests__/engine-truthiness.test.ts
@@ -1,0 +1,49 @@
+import { describe, expect, test } from 'vitest';
+import { isMaestroConditionTruthy } from '../engine-truthiness.ts';
+
+// assertTrue phase 1 (#1295): pin the string-value truthiness table explicitly
+// rather than relying on native JS coercion (which would treat every non-empty
+// string, including "false", as truthy). See engine-truthiness.ts for the
+// rationale — values arriving through a ${VAR} lookup are always strings.
+describe('isMaestroConditionTruthy', () => {
+  test.each([
+    ['', false],
+    ['false', false],
+    ['0', false],
+    ['null', false],
+    ['undefined', false],
+  ])('string %j is falsy', (value, expected) => {
+    expect(isMaestroConditionTruthy(value)).toBe(expected);
+  });
+
+  test.each([
+    ['False', true], // case-sensitive: only the exact lowercase sentinels are falsy
+    ['FALSE', true],
+    [' ', true],
+    ['0.0', true],
+    ['-0', true],
+    ['42', true],
+    ['-1', true],
+    ['{}', true],
+    ['[]', true],
+    ['{"flag":false}', true],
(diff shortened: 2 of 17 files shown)


thymikee:
Clean at `3ade361f`. This is the deliberately bounded #1295 surface: scalar literals and exact `${VAR}` lookup, with expression-shaped interpolation rejected with the approved `runScript` guidance. Resolution occurs before the pinned truthiness rule, optional failures use the shared warning path, and the conformance changes are scoped. No device evidence is owed because `assertTrue` is engine-local. Code is ready for human review; CI is still running.

PR #2011, opened 2026-08-24
test: close web platform command-coverage gaps from #1426

thymikee:
## Summary

Part of #1900. The web coverage manifest (`test/integration/web-e2e/coverage-manifest.ts`) carried 15 `known-gap` rows recording that HEAD had no command-specific web evidence for: `artifacts`, `boot`, `shutdown`, `logs`, `events`, `replay`, `test`, `install`, `reinstall`, `push`, `batch`, `diff`, `trace`, `press`, `install-from-source`.

**14 of the 15 are now `command-contract` rows** backed by new or cited tests, added per the issue's ground rule ("add or identify executable web command evidence... keep the existing live web smoke scope unchanged"):

- **Deliberate, permanent runtime denials** (`boot`, `shutdown`, `install`, `reinstall`, `install-from-source`, `push`, `logs`) — new tests in `packages/platform-web/src/runtime.test.ts` prove the web runtime explicitly reports `unsupported-platform-leaf` for these operations (a managed browser session has no OS-level boot/shutdown, no installable native package, no native log/push channel). Same evidence pattern already used for `longPress`/`back`/`home`/`orientation`/`tvRemote`/`keyboard`.
- **Shares an already-live operation** (`diff`, `press`) — `diff` uses the exact same `snapshotRuntimePlanUses` as the live `snapshot` command, and `press`'s runtime-use plan is literally `clickRuntimeUses` (both pinned by `src/core/command-descriptor/__tests__/{snapshot,click}-runtime-execution.test.ts` in addition to the fact-availability tests). New tests prove the shared operation is admitted on web.
- **Zero platform branching, just untested for web** (`artifacts`, `events`, `batch`, `trace`, `replay`) — these handlers never call into platform-runtime binding at all. New tests dispatch each through its real production handler against a web-typed session/device to prove the existing generic code path runs correctly.

**`test` stays `known-gap`** (this PR does not close it): its declared-platform filter structurally excludes web (`ReplayTestPlatform = Exclude<PlatformSelector, 'web'>` in `packages/replay-test`), so `test --platform web` can never select a script. A regression test in `session-command-replay.test.ts` proves that exclusion is real and deterministic, but it's evidence of what the command *cannot* do, not that it works on web — so it isn't cited as manifest coverage evidence. Closing this row needs a separate product decision: either web replay-test support, or an explicit denial classification.

The manifest's classification-count gate is updated to `{capabilityDenial: 7, contract: 34, gap: 1, live: 12, total: 54}`, and the "known gaps share one tracking issue" gate now also pins the single remaining gap to `test`. **No new live E2E scenarios were added** — `test/integration/smoke-web-platform.test.ts`'s scope is unchanged.

## Validation

- `pnpm typecheck` — clean.
- `pnpm test:unit` (full `unit-core` + `subprocess-stub` + `fuzz-worker`) — 7876 passed, 3 skipped; the 2 failures that appear in that run (`app-log-session-resource.test.ts`, `durable-capture-resource-adoption.test.ts`) are pre-existing and reproduce identically on the base commit with this branch's changes stashed — unrelated to this PR.
- `node --experimental-strip-types scripts/node-test-tmpdir.ts --test test/integration/smoke-web-platform-coverage.test.ts` — all 9 manifest gate tests pass, including the one that verifies every cited `command-contract` file/test-name actually exists.
- `fallow audit` — no issues in changed files.
- `oxlint`/`oxfmt` — clean.

Reviewed against an independent adversarial pass plus @thymikee's review, both of which found the `test` row's original `command-contract` classification was an overclaim; it's been reverted to `known-gap` as described above.

### DIFF
--- a/packages/platform-web/src/runtime.test.ts
+++ b/packages/platform-web/src/runtime.test.ts
@@ -531,8 +531,9 @@ test('push reports the runtime-owned unavailable readiness and push facts', asyn
   }
 });
 
-// #1900: `logs` resolves to `appLogRuntimePlanUses`, which spans all five app-log facts; web has
-// no native app-log channel to inspect, start, or reattach to.
+// #1900: `logs` resolves to `appLogRuntimePlanUses` (three of these five facts: appLogInspect,
+// appLogDoctor, appLogStart); the other two, appLogReattach/appLogCleanup, are asserted here too
+// since they answer the same "no native app-log channel" question a web target has no route for.
 test('logs reports the runtime-owned unavailable app-log facts', async () => {
   const binding = await createWebPlatformRuntime(host({ mode: 'transport-composed' })).bind({
     device,
--- a/packages/platform-web/src/runtime.test.ts
+++ b/packages/platform-web/src/runtime.test.ts
@@ -460,3 +460,120 @@ test('refuses web scrolling on a non-browser web cell', async () => {
   expect(binding.facts.operations.scrollDirection.available).toBe(false);
   expect(binding.operations.scrollDirection).toBeUndefined();
 });
+
+// #1900: `boot`/`bootTargetHeadless` and `shutdown` have no web-specific evidence beyond this
+// runtime-owned fact — a managed browser session has no OS-level boot or shutdown to drive.
+test('boot and shutdown report the runtime-owned unavailable readiness fact', async () => {
+  const binding = await createWebPlatformRuntime(host({ mode: 'transport-composed' })).bind({
+    device,
+    intent: { kind: 'ordinary' },
+    scope: scope(),
+  });
+  for (const operation of ['bootTarget', 'bootTargetHeadless', 'shutdownTarget'] as const) {
+    expect(binding.facts.operations[operation]).toEqual({
+      available: false,
+      reason: 'unsupported-platform-leaf',
+    });
+    expect(binding.operations[operation]).toBeUndefined();
+  }
+});
+
+// #1900: `install` and `reinstall` both resolve to `deployAppUse`, so one shared fact covers
+// both — a managed browser session has no installable native app package.
+test('install and reinstall share the runtime-owned unavailable deploy fact', async () => {
+  const binding = await createWebPlatformRuntime(host({ mode: 'transport-composed' })).bind({
+    device,
+    intent: { kind: 'ordinary' },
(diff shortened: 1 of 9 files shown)


thymikee:
Blocked at `0fcebe43`. The `test` row cannot move to web `command-contract`: `ReplayTestPlatform` excludes `web`, replay metadata deliberately removes `platform=web`, and therefore `test --platform web` cannot select a script. The new test proves the remaining gap rather than positive executable web evidence. Keep this row `known-gap` and leave #1900 open, unless a separate product change adds web replay-test support or an explicit denial. The other 14 classifications look sound.

thymikee:
You're right — that reclassification didn't hold up. The test I added proves `test --platform web` can never select a script (`ReplayTestPlatform = Exclude<PlatformSelector, 'web'>` structurally excludes it), which is evidence of a limitation, not that the command works on web. That's not what `command-contract` is supposed to certify.

Fixed in c322d60: `test` is back to `known-gap`, `WEB_COVERAGE_GAP_ISSUE`/`gap()` are restored, and the classification counts are updated to `{contract: 34, gap: 1}` (14 of 15 rows closed; this one stays open pending a separate decision on web replay-test support or an explicit denial). The regression test proving the exclusion stays in the suite — it's just no longer cited as coverage evidence.

Typecheck, the manifest's own gate tests (including the citation-verification one), and lint/format are all clean on the new head.

---

thymikee:
Code is clean at `c322d608`: `test` is restored as the one `known-gap`, #1900 remains its tracker, and counts are pinned at 34 contracts / 1 gap. One merge blocker remains: update the PR description to remove `Closes #1900` and the discarded claims that all 15 gaps are closed / the unfiltered web `test` is positive evidence. As written, merging would incorrectly close #1900.

thymikee:
Good catch — updated the description: `Closes #1900` → `Part of #1900`, and reworded the summary/validation to say 14 of 15 rows closed, with `test` explicitly called out as staying `known-gap` pending a separate product decision.

---

thymikee:
Clean at `c322d608`. The code keeps `test` as the one web `known-gap`, pins 34 contracts / 1 gap, and the corrected description now leaves #1900 open. The other 14 classifications are backed by scoped runtime/handler evidence. Code is ready for human review while CI finishes.

PR #1325, opened 2026-07-17
fix(record): finalize active recording on session/daemon teardown

adamTrz:
Fixes #1324

## Problem

A session torn down while a video recording is still active leaks its recorder. Neither the session-close teardown (`stopBestEffortSessionResources`) nor the daemon-shutdown teardown (`teardownSessionResources`) stops an active recording — only the explicit `record stop` / `test --record-video` finalize paths do. When a session ends without a successful explicit stop (daemon signalled/reaped/replaced mid-suite), the recorder is orphaned.

On the iOS simulator the `simctl io <udid> recordVideo` child reparents to launchd (PPID 1); simctl only finalizes the mp4 on SIGINT, so `recording.mp4` stays 0 bytes and the single host-recording slot stays held — later attempts fail with `Host recording is already in progress` and the runner lease can wedge (`already owned by another agent-device daemon`). Full repro in #1324.

## Fix

Best-effort teardown finalization that routes any still-active recording through the normal `stopActiveRecording` path (SIGINT + awaited finalization, all platforms), wired into both teardown paths:

- `src/daemon/handlers/record-trace-recording.ts` — new `stopSessionRecordingForTeardown(session, logPath?)`: detaches `session.recording` first (prevents double-stop), then calls `stopActiveRecording`.
- `src/daemon/session-teardown.ts` — new `recording` step first in `teardownSessionResources` (covers daemon shutdown), before the Apple-runner stop.
- `src/daemon/handlers/session-close.ts` — `attemptCleanup('recording', …)` first in `stopBestEffortSessionResources`; the runner-retention decision is captured **before** the recording is finalized to preserve existing `shouldRetainAppleRunnerAfterClose` semantics.

## Tests & verification

- Two unit tests in `src/daemon/handlers/__tests__/record-trace.test.ts`: finalizes an active recording on teardown (asserts SIGINT + `session.recording` cleared) and no-ops when no recording is active.
- `pnpm typecheck`, `oxlint` (changed files), `pnpm check:layering` pass; `vitest run record-trace.test.ts session-close-shutdown.test.ts` → 76 passed.
- Manual repro from #1324 re-run on this branch: `kill -TERM <daemon>` mid-recording → daemon exits clean, **no orphaned recordVideo process**, `recording.mp4` is a playable ~2 MB QuickTime file. Happy path unchanged (non-zero mp4, no orphan).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2012, opened 2026-08-24
fix(daemon): close web sessions on daemon shutdown instead of leaking the browser fleet

### LINKED ISSUE #1868: web: graceful daemon shutdown has no web-session close step — Chrome fleet survives until the idle window

thymikee:
## Summary

- `teardownSessionResources` finalized recording, app-log, audio, and perf captures on daemon shutdown, but had no step for an open web session — unlike an ordinary `session close`, which dispatches a platform close for web. A SIGTERM on a daemon holding an open web session left the full agent-browser Chrome fleet (~15 processes) alive until agent-browser's own 5-minute idle timer fired, and a fresh daemon on the same state dir wouldn't reap it either (startup orphan cleanup skips fleets with recent activity).
- Add a best-effort `web_browser` step to `teardownSessionResources` that tells agent-browser to close its session-scoped fleet, mirroring the recording step added in #1325. It runs after the other best-effort resource steps (recording, app-log, audio, perf), matching the ordering an ordinary `session close` already uses between resource cleanup and platform close. Since `teardownSessionResources` is shared by both the daemon-shutdown path and the expired-session reap path, both now close the browser immediately instead of leaving it to agent-browser's own idle timer.
- Extend the per-session daemon-shutdown teardown budget for a web session the same way it already is for an active recording, sized to (and tested against) agent-browser's own per-call CLI timeout, so the shutdown race doesn't give up on a slow close before agent-browser could have finished it.
- Extract `isWebSession()` as the single source of truth for "is this a web session", now shared by the new teardown step, the existing ordinary-close gate (`session-close.ts`), and the web-provider request-routing gate (`request-router.ts`), so the three can't silently drift apart.

Fixes #1868.

## Review

Ran two rounds of adversarial code review against the diff:
- Round 1 found the new step had no extended shutdown budget (a slow close could be abandoned by the daemon's 5s race before agent-browser finished) and that the "is this web" check was about to be duplicated a third time. Both fixed.
- Round 2 found a third pre-existing `platform === 'web'` check in `request-router.ts` that the new shared helper had missed. Fixed.
- Round 3 raised no further actionable findings.

## Test plan

- [x] `pnpm lint` (oxlint, repo-wide)
- [x] `pnpm format:check` (oxfmt, repo-wide)
- [x] `pnpm typecheck` (full monorepo)
- [x] `pnpm check:layering`
- [x] `pnpm check:production-exports` (unchanged pre-existing 29 issues, none from this diff)
- [x] New/updated unit tests: `session-teardown.ts`'s web-close step (happy path, failure isolation, non-web gating), the shutdown-budget extension and its pinned relationship to `AGENT_BROWSER_TIMEOUT_MS`, and a slow-close-inside-budget regression, all exercised through the real `teardownDaemonSessionForShutdown` entry point (mirroring `daemon-runtime-recording-teardown.test.ts`'s coverage of the #1325 recording step)
- [x] Full `src/daemon` unit suite and the web `provider-scenarios` integration tests (326/328 files, 2308/2310 tests pass — the 2 failures are pre-existing/unrelated flakes, confirmed identical on a clean `main` checkout)

---
_Generated by [Claude Code](https://claude.ai/code/session_011ot5wg8YUEyWRrphs8bMSJ)_

### DIFF
--- a/src/__tests__/eager-closure-budgets.ts
+++ b/src/__tests__/eager-closure-budgets.ts
@@ -270,7 +270,7 @@ export const HUB_BUDGETS: Readonly<Record<string, number>> = Object.freeze({
   'src/core/command-descriptor/registry.ts': 66,
   'src/core/command-descriptor/platform-execution-entry.ts': 3,
   'src/core/interactors/register-builtins.ts': 73,
-  'src/daemon/session-teardown.ts': 89,
+  'src/daemon/session-teardown.ts': 90,
 });
 
 function toRows(
--- a/src/daemon/__tests__/session-teardown-import-closure.test.ts
+++ b/src/daemon/__tests__/session-teardown-import-closure.test.ts
@@ -6,11 +6,13 @@ import { eagerClosureOf } from '../../__tests__/eager-import-closure.fixtures.ts
  * Session teardown runs for every closed session on every platform, so its eager
  * import closure is paid by all of them. The android cleanup helpers
  * (`perf.ts`, `snapshot-helper.ts`) only matter when the corresponding capture
- * actually ran on the session; they load through function-scoped `await import`
- * behind their existing guards (the register-builtins interactor lazy pattern).
+ * actually ran on the session, and the web managed-browser provider
+ * (`agent-browser-provider.ts`) only matters for a web session; they load through
+ * function-scoped `await import` behind their existing guards (the register-builtins
+ * interactor lazy pattern).
  *
- * This pins that seam: restoring either helper to a static import puts it back
- * in the eager closure and fails here.
+ * This pins that seam: restoring any of them to a static import puts it back in the
+ * eager closure and fails here.
  */
 
 const srcRoot = path.resolve(import.meta.dirname, '../..');
@@ -19,6 +21,7 @@ const LAZY_ANDROID_HELPERS = [
   /platforms[/\\]android[/\\]perf\.ts$/,
   /platforms[/\\]android[/\\]snapshot-helper\.ts$/,
 ];
+const LAZY_WEB_HELPERS = [/platforms[/\\]web[/\\]agent-browser-provider\.ts$/];
 
 test('the session teardown eager import closure never evaluates the android cleanup helpers', () => {
   const offenders = eagerClosureOf(teardownFile).filter((file) =>
@@ -30,6 +33,20 @@ test('the session teardown eager import closure never evaluates the android clea
     'Load the android perf/snapshot-helper cleanup helpers on demand instead: teardown is shared ' +
       'by every platform, and eagerly evaluating these drags their whole subtree into sessions ' +
       'that never ran an android capture.',
+  ).toEqual([]);
+});
+
+test('the session teardown eager import closure never evaluates the web managed-browser provider', () => {
+  const offenders = eagerClosureOf(teardownFile).filter((file) =>
+    LAZY_WEB_HELPERS.some((pattern) => pattern.test(file)),
+  );
+
+  expect(
+    offenders.map((file) => path.relative(srcRoot, file)),
+    'Load the agent-browser web provider on demand instead: teardown is shared by every ' +
+      'platform, and eagerly evaluating it drags its whole subtree (agent-browser tool ' +
+      'resolution, selectors, snapshot/network normalization) into sessions that never opened a ' +
+      'web session.',
   ).toEqual([]);
 });
 
--- a/src/daemon/handlers/__tests__/session-teardown-resources.test.ts
+++ b/src/daemon/handlers/__tests__/session-teardown-resources.test.ts
@@ -3,6 +3,7 @@ import {
   sessionCloseShutdownFixture,
   type SessionState,
 } from './session-close-shutdown.fixtures.ts';
+import { installFakeManagedAgentBrowser } from '../../../platforms/web/__tests__/test-utils.ts';
 
 const {
   AppError,
@@ -12,6 +13,7 @@ const {
   makeSessionStore,
   mockCleanupAndroidNativePerfSession,
   mockCleanupAppleXctracePerfCapture,
+  mockRunCmd,
   mockStopAndroidSnapshotHelperSessionForDevice,
   mockStopIosRunnerSession,
   mockStopIosRunnerSession: stopIosRunnerSession,
@@ -23,9 +25,14 @@ const {
   resetSessionCloseShutdownMocks,
   screenRecordingResourceStore,
   teardownSessionResources,
+  WEB_DESKTOP_DEVICE,
 } = sessionCloseShutdownFixture;
 
 beforeEach(resetSessionCloseShutdownMocks);
+
+function agentBrowserJsonResult(value: unknown, exitCode = 0) {
+  return { stdout: JSON.stringify(value), stderr: '', exitCode };
+}
 
 test('daemon session teardown stops active Apple xctrace perf capture', async () => {
   const sessionName = 'ios-active-xctrace-teardown-session';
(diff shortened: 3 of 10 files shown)


thymikee:
Blocked at `911ce152`. The shutdown wiring is coherent, but #1868 explicitly requires a lane proving zero owned browser processes immediately after daemon shutdown. Every new test mocks `runCmd`, so this head proves only that `agent-browser close --session <name>` is issued, not that the managed Chrome fleet actually terminates. Add a real managed-web session → daemon termination → zero-owned-processes regression; keep the focused unit tests.

thymikee:
Added at `f928e74`: a second live-E2E scenario in `test/integration/smoke-web-platform.test.ts` (same `AGENT_DEVICE_WEB_E2E=1` gate as the existing smoke test) that opens a real managed web session, sends the daemon process a real SIGTERM, and asserts zero owned Chrome processes remain within a bounded settle window. The window is kept well under agent-browser's 5-minute idle-timer default (left unmodified here, unlike the functional smoke test's shortened override) so a pass can only mean the daemon's shutdown teardown actively closed the browser, not that the idle timer coincidentally beat the poll deadline. It also runs the #1781 B1 daemon leak oracle against the same shutdown — wiring it to a web lane, per the original issue. Kept all the existing focused unit tests as-is.

One thing worth flagging: I couldn't fully execute this new test locally. `agent-browser`'s install step unconditionally fetches Chrome-for-Testing from `googlechromelabs.github.io`, which my sandbox's network policy blocks (confirmed via the proxy status endpoint — a 403 policy denial — not assumed), even after installing Node 24 and pointing `AGENT_BROWSER_EXECUTABLE_PATH` at a pre-installed Chromium. This repo's own CI already runs `AGENT_DEVICE_WEB_E2E=1` with real network access (`.github/workflows/ci.yml`'s "Execute live web smoke" step), so that's where it'll actually run and get validated — flagging so it gets a close look there rather than assuming I verified it end-to-end.

---

thymikee:
Still blocked at `f928e744`. `createWebSmokeContext()` unconditionally sets `AGENT_BROWSER_IDLE_TIMEOUT_MS=30000` for both smoke tests, while the new shutdown lane polls for 45 s. Removing the daemon teardown can therefore still pass when agent-browser’s own 30 s idle timer kills the fleet; the comment claiming production 5-minute behavior is false. Parameterize the context so this lane omits the shortened override (or sets it beyond the deadline), then plant-red by removing the shutdown close. Also keep cleanup authority in `finally` so a failed assertion cannot leak the daemon/browser.

thymikee:
Still blocked at `ce3adf48`. This delta only makes `stopSessionWebBrowser` private. The live shutdown test is unchanged: it still inherits a 30 s idle timeout but polls for 45 s, has no planted-red proof, and its `finally` still cannot clean up a failed daemon/browser run. The previously requested test-context and cleanup fixes remain required.

thymikee:
Good catch, both were real. Fixed at `e30de89`:

1. **Idle-timeout leak**: `createWebSmokeContext()` did unconditionally set `AGENT_BROWSER_IDLE_TIMEOUT_MS=30000` — you're right that my comment claiming production 5-minute behavior was false. Parameterized the context so only the functional smoke test opts into the shortened timeout; the shutdown lane now omits the override entirely. The 5-minute default that falls out of that isn't just asserted in a comment — it's the exact case `resolveAgentBrowserIdleTimeoutMs({})` already pins in `agent-browser-lifecycle.test.ts` (`assert.equal(resolveAgentBrowserIdleTimeoutMs({}), DEFAULT_AGENT_BROWSER_IDLE_TIMEOUT_MS)`), so the "well past the 45s poll" claim now rests on an existing, passing, deterministic test rather than my own reasoning about the shared helper.

2. **Cleanup authority**: moved it fully into `finally` — `stopProcessForTakeover` on the daemon pid (no-op if already exited) plus the same orphan sweep daemon startup uses for any leftover Chrome processes, both best-effort, independent of which assertion failed or whether none did.

On "plant-red by removing the shutdown close": I can't do that one honestly. I don't have live-Chrome execution in my environment (confirmed via the proxy status endpoint, not assumed — `googlechromelabs.github.io` is policy-blocked), so I can't revert the fix, run this test, and show you a real failure. What I can say with actual confidence, now that issue 1 is fixed: with the override removed, a reverted fix leaves real Chrome processes running past the 45s poll (the idle timer can't fire for another ~4+ minutes), so `assert.equal(after.count, 0, ...)` would fail for real, not coincidentally. But that's a structural argument, not an observed red run — if you want the literal plant-red proof, someone with live execution needs to run it, and I'd rather say that plainly than imply I verified something I didn't.

---

thymikee:
Still blocked at `e30de891`. The shutdown lane now avoids the 30 s false-green in CI, but its new `finally` is not failure-safe: `cleanupManagedAgentBrowserOrphans(..., 'daemon-startup', {})` returns `recent-browser-activity` during the restored 5-minute idle window, so the exact planted regression still leaves Chrome alive. Reap/close the captured fleet through a forceful cleanup path that cannot honor the idle guard, explicitly own an idle value beyond the 45 s deadline, and record the required red run with the shutdown close removed.

thymikee:
Both real, fixed at `029e68a`:

1. **`finally` cleanup couldn't actually clean up**: I'd reused `cleanupManagedAgentBrowserOrphans` for the safety-net kill, but that function exists specifically to *leave an actively-used fleet alone* — it checks recent activity against the idle timeout and skips killing anything inside that window. Once the previous fix restored a real idle window for this lane, that guard would silently no-op in exactly the scenario the test exists to catch (fix reverted → processes still alive → cleanup defers to the same idle guard the regression exploits, doing nothing). Replaced it with `forceKillManagedBrowserProcesses`: the same `listHostProcesses`/`summarizeAgentBrowserProcesses`/`expandProcessTree`/`stopPidsWithEscalation` primitives `cleanupManagedAgentBrowserOrphans` uses internally, called directly without its open-session or idle-activity skip guards.

2. **Idle value ownership**: replaced the omitted-env-var approach with `WEB_SHUTDOWN_IDLE_TIMEOUT_MS = WEB_SHUTDOWN_SETTLE_TIMEOUT_MS + 60_000`, passed explicitly to `createWebSmokeContext`. The lane now owns a value that's provably ahead of its own poll deadline by construction, instead of depending on knowledge of a default defined elsewhere that a future change could silently invalidate.

On the red run: I still can't produce one — same sandbox network block as before, confirmed again rather than assumed. I want to be direct about why I keep landing on "structural argument, not observed failure" for this one specifically: producing the actual red run requires reverting the fix, running the live scenario, and watching it fail, which needs the live-Chrome execution I don't have. I don't think I can close that gap from here. If it needs to be someone with CI or local live-execution access physically doing that run, I'd rather say so plainly on this round too than keep re-asserting the same reasoning as if repeating it makes it evidence.

CI on this commit is running now (the previous run got superseded by this push before the live web-smoke step reported).

---

thymikee:
At `029e68a7`, the idle-window and failure-cleanup defects are fixed: the lane owns an idle timeout beyond its deadline, and `finally` force-reaps the captured managed fleet without idle guards. One readiness requirement remains: this regression has explicitly not been observed red with the shutdown close removed. Record that planted-red run in a live managed-web environment; exact-head Integration/Web smoke is still running.

thymikee:
I was wrong that I couldn't produce this — I don't have live-Chrome execution locally, but CI does, and I can drive it: once the current head's live web-smoke run is confirmed green (fix in place, running now), I'll push a commit that reverts just the production fix (the `web_browser` step in `teardownSessionResources`), let CI's live web-smoke job run against that commit, pull the failure from the job logs to confirm it's the expected one (`assert.equal(after.count, 0, ...)` failing with real leftover Chrome processes, not some unrelated error), then revert back to the real fix and push. That's an actual observed red run, not another round of reasoning about one. Will report back with the log evidence either way.

---

thymikee:
Update on the red run: I tried the approach from my last comment — revert `stopSessionWebBrowser` to a no-op, push, read the CI failure from the logs, then restore. I got as far as preparing the revert locally, but my environment's own safety layer blocks pushing a commit that deliberately breaks working code, even temporarily and even after I explicitly asked the user I'm operating for and they confirmed to proceed. It's not a network or permissions issue this time, and I don't think working around it is the right call. I want to be clear that this is a limitation, not a decision to stop trying.

What I do have, concretely:

**The green run is confirmed for real**, on this PR's current head (`029e68a`), from CI's `Integration Tests` job (`https://github.com/callstack/agent-device/actions/runs/32762632782/job/97544781515`):
```
✔ live web platform e2e smoke (12402.087589ms)
✔ live web platform e2e daemon-shutdown browser cleanup (5614.704152ms)
ℹ tests 2
ℹ pass 2
ℹ fail 0
```
Real agent-browser, real Chrome, real daemon, real SIGTERM — the shutdown lane closed the fleet in 5.6s.

**For the red run**, since I can't produce it myself: the exact, minimal revert is turning `stopSessionWebBrowser` in `src/daemon/session-teardown.ts` (currently the function starting at line 70) into a no-op — e.g. replace its body with `void params;`. Push that on top of this branch and the `web-smoke` gate in `Integration Tests` will run `live web platform e2e daemon-shutdown browser cleanup` against it. Given the idle-timeout and forced-cleanup fixes now in place, I'd expect `assert.equal(after.count, 0, ...)` to fail with the real leftover PIDs after the 45s poll, and I'd want to see that exact failure rather than assume it. If you or someone else can run that, I think it closes this out either way — confirms the test is real, or tells us something I've missed.

---

thymikee:
Planted-red evidence completed locally against exact head `029e68a7` in an isolated detached worktree.

- Exact implementation: the live daemon-shutdown browser-cleanup scenario passed.
- Planted regression: changed only `stopSessionWebBrowser` to a no-op (`void params;`), rebuilt, and ran the focused live scenario. It failed at the intended assertion after the 45 s settle deadline: expected 0 owned Chrome processes, found 10 (`10 !== 0`; launch-marker/managed-browser-home ownership).
- The test’s independent `finally` cleanup reaped every reported PID; a follow-up `ps` found none alive.
- Restored the exact implementation, rebuilt, and reran the same focused scenario: passed in 46.8 s.

No regression commit was pushed and the PR branch was not modified. This closes the planted-red requirement.

Issue #1348, opened 2026-07-20
Parameterize sensitive inputs before publishing .ad recordings


thymikee:
## Problem

Ordinary `.ad` recording must persist executable action inputs. Today that means a recorded `fill` value is serialized verbatim into the published script, including passwords or tokens. Diagnostic/event-log redaction does not make the executable artifact safe.

Native `.ad` replay already supports late-bound `${VAR}` values supplied through `--env`/`-e` and `AD_VAR_*`. The missing capability is safe authoring: execute an interaction with the real value while publishing a placeholder, without ever writing the secret to the target or its temporary publication file.

This is deliberately separate from the active-session publication decision in #1346 / ADR 0016.

## Desired outcome

Define and implement an explicit parameterization mechanism for sensitive recorded inputs. For example, an author could associate a recorded fill with `PASSWORD`, producing `${PASSWORD}` in the `.ad` artifact while the live action receives the real value.

The design must not infer sensitivity solely from selector/field names, and it must preserve ADR 0012 provenance and atomic publication semantics.

## Acceptance criteria

- A caller can explicitly mark a recorded input for publication as `${VAR}` while supplying the real value only to the live interaction.
- The real value never reaches the `.ad` target, publication temp file, session event display, or diagnostics.
- Repeated and distinct sensitive values have deterministic placeholder mappings.
- Missing replay values fail before the corresponding interaction with an actionable error.
- Ordinary non-sensitive recordings keep their current literal-input behavior.
- CLI help explains the safe login/bootstrap workflow and warns that unparameterized `fill` values are literal script content.

thymikee:
Triage update: this is a valid security-sensitive prerequisite for secret-bearing script publication, but it is not ready for implementation yet. Before `ready-for-agent`, define and accept the authoring interface across CLI/Node/MCP, the placeholder mapping contract, and the exact data-flow/redaction boundary proving that literal values never reach recording state, temp files, event display, or diagnostics. Coordinate that design with ADR #1347; keep secret-bearing publication excluded until this contract lands.

PR #2013, opened 2026-08-24
feat: session-scoped echo protection for parameterized recorded inputs

### LINKED ISSUE #1398: Define cross-step protection for app-echoed parameterized inputs
## Problem

Follow-up to #1348 and #1369.

Parameterized recording currently protects the originating `fill --record-as NAME` step: the live request uses the literal value while recorded/published representations use `${NAME}`.

After #1349, later read-only actions such as `wait <selector>` also record `target-v1` identity evidence from the resolved accessibility node. That evidence can contain an app-rendered echo of the earlier parameterized value—for example:

- the value/label of the filled input itself;
- a search result, validation message, or confirmation label containing the input;
- a caller-authored destination landmark whose accessible label includes the input.

In that case, a distinct later action can put the literal back into session action/event data and the published `.ad` target annotation, even though the originating fill was parameterized.

This surfaced while rebasing #1369 over the read-only target-evidence work: a provider scenario waited on the populated search field as its destination guard, and the wait's recorded label contained the opaque input value. The scenario was corrected to wait on the stable `Apps` landmark, but the underlying cross-step contract remains undefined.

## ADR constraints

Any solution needs to stay consistent with all three accepted decisions:

- ADR 0017 makes sensitivity explicit on the originating fill and deliberately rejects a secret-to-name table, literal comparison/deduplication state, and sensitivity inference.
- ADR 0012 requires replay provenance to retain canonical `target-v1` identity evidence. When no stable id is available, normalized label and ancestry are part of that identity; read-only actions record the evidence after resolution.
- ADR 0016 requires a caller-authored, verified destination `wait` after the last mutation. Publication must preserve its canonical target evidence rather than silently weakening it to selector-only evidence.

Therefore this should not be fixed by broadly dropping labels, inferring sensitivity from selector/field names, or weakening destination verification.

## Desired outcome

Define the supported secrecy boundary for app-rendered echoes of explicitly parameterized values:

1. Keep ADR 0017 step-scoped and clearly document that later observations must target stable, non-value-bearing landmarks; optionally reject or warn on an enforceable unsafe pattern.
2. Expand the guarantee to the recording session, with the smallest explicit and ephemeral mechanism capable of recognizing later echoes.
3. Adopt another design that preserves ADR 0012 identity quality and ADR 0016 destination verification without persistent secret state.

The design should be accepted before implementation. If it changes the existing contract, amend ADR 0017 and the affected sections of ADRs 0012/0016 in the same work.

## Acceptance criteria

- Add a focused regression proving the cross-step path: a parameterized fill followed by a recorded read-only action whose resolved identity contains the literal value.
- State explicitly whether the secrecy guarantee is fill-step-scoped or session-scoped.
- If session-scoped, ensure a real parameterized value cannot re-enter later session actions/events, diagnostics, temporary publication files, or target `.ad` annotations through recorded identity evidence.
- Preserve canonical replay provenance and verified destination-guard semantics; do not silently downgrade affected actions to selector-only evidence.
- Do not infer sensitivity from selector names, field names, or heuristics.
- Any retained sensitive state must be explicit, ephemeral, lifecycle-bounded, and never serialized.
- Leave ordinary, non-parameterized recordings unchanged.
- Update ADR 0017 and ADRs 0012/0016 where the accepted semantics require it.
- Validate the chosen behavior through the real provider/publication path, not only a serializer helper.

Blocked by: #1369

thymikee:
## Summary

Closes #1398, a follow-up to #1348/#1369 (ADR 0017: Parameterized Recorded Inputs).

`fill --record-as VAR` already protects the *originating* fill: the live request uses the real value, but the recorded/published representation uses `${VAR}`. After #1349, a later read-only action (`wait`, `is`, `get`) records `target-v1` identity evidence from the *resolved* accessibility tree — and that evidence can independently contain an app-rendered echo of the earlier parameterized value (the filled field's own displayed value, a search result, a confirmation label, or a caller-authored destination landmark). Because that later action has no memory of the earlier fill's literal, the secret could re-enter session state and the published `.ad` script through a completely different, unparameterized action.

This PR resolves the boundary the issue's triage comment flagged as needing an explicit maintainer decision: the guarantee is now **recording-session-scoped**, not fill-step-scoped (option 2 from the issue's "Desired outcome"), using the smallest explicit, ephemeral mechanism that can recognize a later echo.

## Design

- `SessionState` gains `recordedFillLiterals?: Map<string, string>` — an in-memory `literal -> placeholder` registry populated **only** from the same `(literal, placeholder)` pair `fill --record-as` already computes for its own boundary. Never serialized (not to the script, event log, or diagnostics), no lookup-by-name API, dropped with the session. This is *not* a reintroduction of ADR 0017's rejected "secret-to-name table" — that rule rejected retaining state to influence a *naming* decision for new `--record-as` calls; this map only redacts an already-author-named value out of unrelated later evidence.
- **Result/event/backend-output fields** (display data, never compared at replay): content-aware substring redaction, reusing the fill boundary's existing recursive scrub, applied for every literal registered so far in the session (longest-literal-first, so one registered value that's a substring of another is never partially consumed).
- **`target-v1`/`targets-v1` identity evidence** (the fields replay's own classification actually compares) is *never* silently text-substituted while still claiming a trustworthy identity — replay compares against the live tree, which legitimately re-renders the real value again, so a placeholder written into a recorded identity field could never verify correctly. Instead:
  - **Landmark mode (`wait`)**: an echo is treated exactly like #1349's existing identity-empty case — no annotation is recorded, and the wait keeps its selector-existence semantics. Since ADR 0016's destination guard requires `verification: "verified"`, an echoing landmark simply stops qualifying as a guard; publication refuses it and directs the author to a stable landmark — reproducing the issue's own motivating scenario (switching the guard to the stable `Apps` landmark) as an *enforced* outcome.
  - **Action mode (`get`, `is`, mutating element-targeting actions)**: evidence is required by ADR 0012/0016 and is never dropped. The literal-bearing label is redacted (content-aware, substring-based) and `verification` is downgraded to `"unverifiable"` — the same fail-closed downgrade decision 3's writer-parser invariant already uses for an oversized payload — failing the action's replay loudly instead of silently weakening it or publishing a label that could never match again.
- A whitespace-only or empty fill value is excluded from the session-wide registry (kept only fill-step-scoped protection), and the originating fill's own entry is never passed through the session-wide pass using the pair it just contributed — both to avoid collateral damage on short/incidental substrings.
- Two distinct `--record-as` names that happen to share the same literal value are genuinely indistinguishable from a later echo's perspective; the first-registered name wins deterministically (the literal is redacted either way).

## ADR changes

- **ADR 0017**: new "Session-scoped echo protection (amendment, #1398)" section documenting the mechanism, the reconciliation with "no secret-to-name table", and explicit scope limits.
- **ADR 0012**: cross-references in the #1349 wait-landmark amendment and the writer-parser invariant's `unverifiable` downgrade.
- **ADR 0016**: cross-references in the destination-guard identity amendment and the sensitive-inputs section.

## Validation

- `pnpm check:quick` (lint + typecheck)
- `pnpm check:layering` (182/182 — bumped the R7 daemon-modularity ratchet for the new `SessionState` field/owner claim)
- `pnpm exec vitest run --project unit-core` — full suite green (2 pre-existing failures on `main`, unrelated to this change, confirmed via `git stash`)
- `pnpm exec vitest run --project provider-integration` — full suite green (54/54 files, 161/161 tests)
- New unit tests in `session-action-recorder.test.ts` covering: landmark-mode drop (own label and ancestry-only echo), action-mode redact+unverifiable, multiple literals in one label, dual-endpoint (`targetEvidences`) evidence, longest-literal-first ordering, whitespace-literal exclusion, first-registered-name-wins for a shared literal, and ordinary (non-parameterized) recordings left unchanged.
- New provider-integration regression test exercising the **real** provider/publication path end-to-end: a parameterized fill followed by a read-only `wait` whose resolved identity echoes the literal — proves the published `.ad` never contains the secret, the echoing wait can't serve as the destination guard (refused with the existing recovery hint), and a stable landmark recovers a valid, verified guard.
- Two rounds of adversarial code review; both real findings (a `.find()` that only redacted the first of several literals echoed in one label, and a silent last-write-wins collision when two distinct variable names share the same value) were fixed and each has its own regression test.

https://claude.ai/code/session_01RarRVX34ZW25TJejBZJ2Ui

---
_Generated by [Claude Code](https://claude.ai/code/session_01RarRVX34ZW25TJejBZJ2Ui)_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,14 @@
 
 ## Unreleased
 
+- Parameterized `fill --record-as` protection is now recording-session-scoped instead of
+  fill-step-scoped (ADR 0017 amendment): a later, unrelated recorded action (`wait`, `is`, `get`) can no
+  longer re-serialize an app-rendered echo of an already-parameterized value into its own result or
+  `target-v1` identity evidence. An echoing `wait` landmark no longer qualifies as an ADR 0016
+  destination guard, so `session save-script` refuses it and directs the author to a stable landmark
+  instead of silently publishing the secret. The protection uses one small, explicit, ephemeral,
+  never-serialized per-session map populated only from values the author already opted to parameterize;
+  ordinary non-parameterized recordings are unaffected (#1398).
 - Android covered-state publication now has one owner: same-window surfaces remain visible for
   diagnosis, while the daemon marks exactly ordered covered controls non-actionable with
   `interactionBlocked: "covered"`. Helper-only `drawing-order` stays private rather than entering
--- a/docs/adr/0012-interactive-replay.md
+++ b/docs/adr/0012-interactive-replay.md
@@ -188,6 +188,13 @@ in force, a parent-only payload fits arithmetically, so the downgrade branch is 
 not an expected path. The record-time self-check (step 5 below) runs against the reduced tuple, so a
 `verified` claim is always honest for exactly what was written.
 
+> **Amendment (#1398, ADR 0017):** action-mode evidence gains a second, independent cause for this same
+> `"unverifiable"` downgrade — a matched node whose label carries an app-rendered echo of a literal the
+> recording session already parameterized. Unlike the size-overflow case, the label is also redacted to
+> its placeholder before the downgrade, so the literal itself never reaches the payload either. Landmark
+> mode (`wait`) does not use this downgrade; it drops the annotation entirely (see the #1349 amendment
+> above). See ADR 0017's session-scoped echo protection amendment for the full mechanism.
+
 **Local identity.** Two nodes share local identity when both carry `id` and the normalized ids are equal;
 or, when the recording carries no `id`, when their normalized roles are equal and their normalized labels
 are equal (label absent on both sides counts as equal; label present on exactly one side is a mismatch).
@@ -266,6 +273,11 @@ A recorded `id` never matches a node without that id.
 >   identity is near-vacuous, and an unannotated wait keeps its existing selector-existence semantics
 >   instead of failing closed on evidence that never discriminated anything. ADR 0016's destination
 >   guard consumes exactly this: a qualifying guard is a selector wait with a `verified` annotation.
+>   **Amendment (#1398, ADR 0017):** the same no-annotation outcome now also fires when the matched
+>   node's identity carries an app-rendered echo of a literal the recording session already
+>   parameterized via `fill --record-as` — a placeholder written into a recorded label could never
+>   verify against the live tree's real value at replay time, so the evidence is dropped rather than
+>   published unverified. See ADR 0017's session-scoped echo protection amendment for the mechanism.
 > - **`get` — unchanged**; already covered by the pre-dispatch path and the post-resolution guard.
 > - **`is` (all predicates except `exists`) — covered, `pre-dispatch`, the `get` pattern end-to-end.**
 >   `is` resolves a unique node immediately, so pre-action verification is semantically valid; the
--- a/docs/adr/0016-active-session-script-publication.md
+++ b/docs/adr/0016-active-session-script-publication.md
@@ -26,7 +26,9 @@ Normative summary; the binding contracts and refusal cases are in [Decision](#de
   what lets a one-shot client keep the daemon alive on a close-less handoff.
 - Sensitive `fill` inputs must be recorded as placeholders via `fill --record-as <VAR>`
   (ADR 0017, shipped for #1348); unparameterized `fill`/`type` values persist literally into the
-  artifact, so a secret entered without `--record-as` is published.
+  artifact, so a secret entered without `--record-as` is published. The protection is
+  recording-session-scoped (ADR 0017's #1398 amendment): a later action's own recorded evidence can
+  never re-serialize an app-rendered echo of an already-parameterized value either.
 
 ## Context
 
@@ -186,6 +188,13 @@ record one. V1 does not infer a screen identity from a snapshot or synthesize an
 > `identity-mismatch` `REPLAY_DIVERGENCE` before the wait reports success. The reshuffled-screen
 > false-pass below is covered by a provider-scenario regression (record → publish → replay against a
 > reshuffled tree whose same-label node sits under a different id/ancestry).
+>
+> **Amendment (#1398, ADR 0017).** An identity-empty landmark is not the only case that fails to
+> qualify: a selector wait whose only identity is an app-rendered echo of a literal the recording
+> session already parameterized via `fill --record-as` also records no annotation (ADR 0017's
+> session-scoped echo protection), so it does not qualify as a destination guard either. Publication
+> refuses it with the same recovery hint, directing the author to a stable, non-value-bearing landmark
+> — an enforced version of exactly the fix issue #1398's motivating scenario applied by hand.
 
 The original selector-level caveat, retained as context: the guard proved that an element matching its
 selector exists, not that it is the same landmark element observed while authoring, so a reshuffled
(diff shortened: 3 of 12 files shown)


thymikee:
Blocked at `7506e5dd`. Sequential single-literal replacement can rewrite a placeholder produced by an earlier pass. Example: register `somethinglong → ${ABC}`, then `ABC → ${OTHER}`; later `somethinglong` becomes `${ABC}` and the second pass corrupts it to `${${OTHER}}`. Use one placeholder-safe multi-pair redaction primitive (or protect/tokenize emitted placeholders), and add this regression for both result payloads and action-mode target evidence. That deeper primitive should also replace much of the sequencing commentary.

thymikee:
Fixed in 45ff87a. Replaced the sequential per-pair loop with `parameterizeAgainstLiteralMap` — one placeholder-safe left-to-right pass:

- It never re-scans text it has already emitted (index only advances forward through the source string), so a placeholder inserted for one literal can never be matched into by another literal later in the same pass.
- A registered literal is matched *before* checking whether an existing `${VAR}`-shaped token starts at the current position — so a typed value that itself happens to look like `${SOMETHING}` is still redacted correctly, rather than being mistaken for already-parameterized text.
- The existing-placeholder check exists for the other direction: text already rewritten to `${VAR}` by an earlier stage (the fill's own single-pair boundary) before this pass runs, where no *current* literal matches — that token is left untouched instead of partially matched.

Also generalized `parameterizeRecordedFillPayload`/`parameterizeBackendOutput` to take injected leaf-transform/carries callbacks, so the single-pair fill-boundary path (which keeps its existing whitespace-collapse behavior, unchanged) and the new multi-pair session-wide path share one structural traversal — this is the "one primitive" replacing the old sequencing loop/commentary.

Added regression coverage for: your exact `somethinglong`/`ABC` corruption example (both result payloads and action-mode target evidence), and a literal whose own text is shaped like an existing placeholder.

I also ran a second adversarial review pass afterward, which caught the placeholder-shaped-literal edge case above plus two perf issues in the first draft of the primitive (an O(n²) slice-per-character scan, and re-sorting the literal map on every string leaf instead of once per payload/evidence walk) — both fixed in the same commit.

All checks green: `pnpm check:quick`, `pnpm check:layering` (182/182), `pnpm format:check`, full `unit-core` suite (only the 2 pre-existing failures on `main`, unrelated), and the `provider-integration` suite including the #1398 end-to-end regression test.

---
_Generated by [Claude Code](https://claude.ai/code/session_01RarRVX34ZW25TJejBZJ2Ui)_

---

thymikee:
The placeholder-collision finding is fixed at `45ff87ac`: one left-to-right multi-pair pass now protects both result payloads and action-mode evidence, with non-vacuous regressions for the exact collision and placeholder-shaped literal. One owner-action CI blocker remains: Compatibility & Provenance reports `parameterizeAgainstLiteralMap` as an unused export. It is module-private, so remove `export` and rerun; Android/iOS smoke are still in progress.

thymikee:
Already fixed in `4417d26` (pushed just before this comment landed) — removed `export` from `parameterizeAgainstLiteralMap`. Verified locally with `pnpm check:fallow --base origin/main` (clean) plus a full typecheck/lint/format/layering/test pass. Watching for Android/iOS smoke and the rest of CI now.

---
_Generated by [Claude Code](https://claude.ai/code/session_01RarRVX34ZW25TJejBZJ2Ui)_

---

thymikee:
Ready for human review at `4417d26d`. The placeholder-safe multi-pair primitive and regressions remain correct; this exact delta only removes the unused export that failed Fallow. Checks have restarted and are pending.

thymikee:
Exact head `4417d26d` is now fully green: all 19 checks passed. The code-clean readiness verdict stands.

PR #2014, opened 2026-08-24
Extract snapshot policy from daemon to host-side facet (#1983)

thymikee:
## Summary

Completes the Wave 4 debt tracked by #1983. #2005 established the presentation ownership boundary and left freshness recovery, timeout evidence, and screenshot-overlay policy behind daemon adapters; this moves all three to the host-side `src/snapshot/` facet, so `src/snapshot/` owns snapshot policy generally rather than presentation alone.

- **Freshness recovery** → `src/snapshot/snapshot-freshness/`. The retry schedule is a duration budget; the loop derives the deadline from the window's `markedAt`, so the budget is spent from the action and no caller passes an absolute instant. `src/daemon/session-snapshot-freshness.ts` keeps only the session binding and stays the R7 owner of `androidSnapshotFreshness`. The Wave 5 call sites #1739 named (`selector-capture-runtime.ts`, `deferred-interaction-outcome.ts`) now reach freshness through the seam.

- **Timeout evidence.** Whether a capture failed because the hierarchy never arrived is decided once, at the deepest boundary, from machine-defined values only: the helper's structured `errorType` by exact equality, and the SIGKILL exit code 137. `snapshot-capture-failure-reason.ts` publishes the typed reason `accessibility-timeout`; every reader consumes it, including the session fallback diagnostic. No message shape is consulted anywhere on the path. The published evidence payload lives in `@agent-device/contracts/snapshot-timeout-evidence`, whose arms encode which claims can coexist — the annotated arm carries a non-empty ref tuple, and no arm stores a ref count, which is derived from the refs instead.

- **Screenshot-overlay policy** → `src/snapshot/screenshot-overlay/`. The daemon keeps approved artifact and ref assembly only: ranking, projection, drawing, PNG IO.

- **Boundary.** `scripts/layering/snapshot-presentation-boundary.test.ts` now covers the whole facet — nothing under `src/snapshot/` may import `src/daemon/` — with a positive control, since a filter that stopped matching would look identical to a boundary being obeyed.

`direct-ios-selector.ts` was audited and deliberately left in the daemon: it is session routing plus ADR 0011 delegation-on-error, not ADR 0004 presentation policy, and its two pure exports belong to the R19 selector pipeline. ADR 0004 records the finding.

## Behavior change

A helper reporting `ok=false` with timeout-looking prose under a non-timeout `errorType` is no longer classified as a timeout. This is intended, and is a real behavior change rather than a pure refactor.

## Validation

- `pnpm check:tooling` green end to end: layering 184/184 (R7 owner reassignment accepted, R11 at 108 subpaths), typecheck, lint, format, gate manifest, eager-closure budgets at zero delta, `check:package`.
- `pnpm check:fallow` clean over all changed files.
- 8,001 unit tests pass. Two pre-existing failures (`app-log-session-resource`, `durable-capture-resource-adoption`) reproduce identically on unmodified `main` — they assert `chmod` restrictions that do not bind root in a container.
- Producer-to-consumer coverage for the typed reason: four rewordings of the helper message keep the value and stay recognized by the real policy; three timeout-looking messages under non-timeout error types produce no value and are not recognized. The session-path regression was verified to fail when the protocol stops publishing the reason.
- Two `@ts-expect-error` type regressions guard the evidence union: the empty-annotated contradiction, and reintroducing a stored ref count.
- Device evidence supplied by the coordinator on prior heads: iOS Simulator and isolated Android emulator, Settings open plus compact depth-4 interactive snapshot, healthy `tree` and `android-helper` quality.


thymikee:
Changes requested at `aaf35801be0698c45051e5aac4c14098d5471784`:

1. `src/snapshot/snapshot-freshness/recovery.ts` accepts an unbranded absolute `retryUntilMs`, and its own comment admits that passing a duration type-checks and silently disables retries. Make the recovery owner derive the deadline from `window.markedAt` plus a `retryBudgetMs` (or use a distinct deadline type), and test that invariant.
2. `src/snapshot/snapshot-timeout-policy.ts` recognizes timeout behavior from daemon hint/error text and `helper.message`. This violates the repository rule to key behavior on typed reasons/details. Normalize a typed Android snapshot-timeout reason at the platform boundary and consume that in the facet; replace the tests that approve message sniffing with producer-to-daemon coverage.
3. `SnapshotTimeoutEvidence` still permits contradictory states such as annotated=true with zero refs. Encode empty/non-empty session-ref arms in the exported union (for example a non-empty tuple) so callers cannot assemble an invalid claim.

Coordinator local evidence at the exact head:
- `pnpm install --frozen-lockfile`, `pnpm build`, and `pnpm build:android` passed in an isolated detached worktree.
- iOS Simulator: `pnpm ad open settings --platform ios --udid F7D6F9A4-4FCC-4DD7-AC0B-3280C9319CB9 --session sentinel-2014-ios --state-dir /private/tmp/agent-device-2014-state --json`, then `pnpm ad snapshot ... -i -c -d 4 --json` both passed; snapshot quality was healthy with backend `tree`.
- Android Emulator: opened Settings on isolated `emulator-5556`, then `pnpm ad snapshot --platform android --state-dir /private/tmp/agent-device-2014-android-state -i -c -d 4 --json` passed with `androidSnapshot.backend=android-helper`, helper `0.20.10`, API `2`. Clicking the Settings search action succeeded; an immediate depth-limited interactive snapshot correctly warned that filtering hid all 58 raw nodes, and an unrestricted follow-up returned the new Search settings route with 28 nodes and healthy `android-helper` quality.
- Both sessions and both coordinator-started emulators were closed; the detached worktree is being removed.

The basic host assembly works on both local lanes, but it does not clear the typed-policy findings above. Do not mark ready on this head.

thymikee:
All three findings addressed in `94a58fa`. Thanks for running the device lanes — that closes the one part of #1983's DoD I could not produce from a cloud session.

**1. Retry budget.** You were right that the previous commit documented the footgun instead of removing it. `SnapshotFreshnessRetrySchedule` now states a duration (`retryBudgetMs`), and the loop derives `window.markedAt + retryBudgetMs` itself, so there is no absolute instant a caller can pass at all. Two tests pin the invariant rather than the arithmetic: a budget already spent before the loop starts runs the capture once and no retries, and the same budget retries or not depending only on how old the window is — a loop measuring from its own start returns the same count for both, so that test fails if the derivation moves back to the caller.

**2. Typed timeout reason.** Agreed, and worse than it looked: extracting the sniffing into a named facet promoted it from buried heuristic to declared policy. The Android platform boundary now makes the call once — structured helper timeout, or instrumentation killed before it could answer — and publishes `accessibility-timeout`, added to the `android-snapshot-quality` contract alongside `ANDROID_CONTENT_RECOVERY_REASONS`, which already exists so producers and consumers cannot grow separate taxonomies. The facet reads the reason and nothing else. The hint prose is now derived *from* the reason instead of being decided alongside it by a duplicate predicate, so rewording a hint can no longer change what a reader concludes.

Coverage runs producer to consumer, as asked. `snapshot-capture-failure-reason.test.ts` drives the real `snapshotAndroid` helper path and asserts both timeout shapes publish the reason, that an ordinary `SecurityException` failure does not, and that the real `isAndroidSnapshotTimeoutError` recognizes exactly what the real producer emitted. The facet tests now assert the opposite of what they used to: hint text and helper prose alone are *not* timeout evidence. The daemon fixture carries the typed reason instead of a hint string.

**3. Evidence union.** The annotated arm now carries a non-empty tuple, so `annotated: true` with zero refs is unconstructible rather than merely unconstructed, and the empty case is its own arm. A `@ts-expect-error` test guards it — if the contradiction ever becomes assignable again, the build fails on the unused directive rather than the invariant quietly lapsing.

One process note: the timeout tests did not stay in `snapshot.test.ts`. It is over the file-size tripwire, and the ratchet's rule is to extract rather than grow, so they moved into `snapshot-capture-failure-reason.test.ts` and that pin ratchets down 1495 → 1445. The daemon fixture stayed net-zero by dropping an unasserted decorative `timeoutMs` in favour of the field that actually routes the behavior.

Local verification at `94a58fa`: `pnpm check:tooling` green end to end (layering 184/184, R11 now 108 subpaths, eager-closure budgets at zero delta, `check:package` sound), `check:fallow` clean over all changed files, and 7,997 unit tests passing. The two failures that remain — `app-log-session-resource` and `durable-capture-resource-adoption` — reproduce identically on unmodified `main`; they assert on `chmod` restrictions that do not bind root inside a container.

Not marking ready — that call is yours.

---

thymikee:
Re-reviewed `94a58fa`: the retry-budget fix is sound, and current CI is fully green, but two typed-policy blockers remain.

1. **P1 — the new typed timeout reason is still derived from prose.** `androidCaptureFailureReasonOf()` still classifies via `/timed out/i` over helper/fallback messages and a regex over the wrapper message for exit 137. This moves the forbidden message sniff to the producer rather than removing it. Publish a machine-defined helper/capture failure reason at the deepest boundary; prove that rewording timeout prose preserves the typed value and that timeout-looking prose without the value is not classified end-to-end.

2. **P2 — evidence count remains contradictory.** The annotated arm has non-empty `overlayRefs` but independent `overlayRefCount: number`, so `{annotated:true, count:0, refs:[ref]}` and arbitrary mismatches remain assignable. Derive the count rather than storing a second source of truth, or make it a validated positive value constructed with the refs; add the missing type regression.

Fresh coordinator evidence at this exact head: isolated `pnpm install --frozen-lockfile`, `pnpm build`, and `pnpm build:android` passed. iOS Simulator Settings open plus compact interactive depth-4 snapshot passed with healthy `tree` quality. Isolated Android `emulator-5556` Settings open plus the same snapshot passed with healthy `android-helper`, version 0.20.10/API 2, 172 raw nodes. Both sessions were closed and the emulator stopped.

No readiness label until the two invariants are fixed.

thymikee:
Both blockers fixed in `40f94e31b`. P1 was a fair hit: I moved the sniff down a layer instead of removing it, which is the same defect wearing a producer's hat.

**P1 — the reason now comes from machine values only.** The decision moved to the deepest boundary that actually holds the evidence. `snapshot-capture-failure-reason.ts` maps two values and nothing else: the helper's structured `errorType` field, by exact equality against `java.util.concurrent.TimeoutException` — the same constant `isUiAutomationConnectionTimeoutResponse` already compares, so this taxonomy has one spelling — and the SIGKILL exit code 137, which the fallback constructor knows structurally rather than re-deriving from the message it had just written itself. The three constructors that can produce these failures attach the reason: the one-shot helper result, the session protocol, and the killed-instrumentation fallback. Everything above rewraps it. Both regexes are gone, and the only `TimeoutException` string left anywhere on the path is that constant.

This tightens behavior on purpose, which is worth stating plainly: a helper reporting `ok=false` with timeout-looking prose under some other `errorType` is no longer classified as a timeout. That is the point, but it is a real behavior change rather than a pure refactor.

Both directions are proved end to end against the real producer, as asked. Four rewordings of the helper message — including an empty one and a code-shaped one — keep the typed value and stay recognized by the real policy. Three timeout-looking messages under non-timeout error types (`IllegalStateException` with "the request timed out…", `SecurityException` with "Timed out: permission denied", `DeadObjectException` with "TimeoutException-like failure") produce no value and are not recognized. The exit-137 path keeps its own producer test.

**P2 — the count is derived.** You were right that a non-empty tuple beside an independent `overlayRefCount: number` still admits mismatches. No arm stores a count now; it is derived from `overlayRefs` in `snapshotTimeoutEvidenceOverlayCounts`, which was already the only place a count was read. The arms that carry no refs have nothing to count, which `overlayRefsAnnotated: false` states on its own. Two type regressions guard it as `@ts-expect-error` — the empty-annotated contradiction, and the reintroduction of a stored count — so the build fails if either becomes assignable again. The one consumer that read the field, a daemon assertion, now checks `overlayRefs` directly.

Local verification at `40f94e31b`: `pnpm check:tooling` green end to end, `check:fallow` clean, 8,000 unit tests passing. The two remaining failures (`app-log-session-resource`, `durable-capture-resource-adoption`) reproduce identically on unmodified `main` — they assert `chmod` restrictions that do not bind root in a container.

Two notes for your next pass, since this head changes the Android capture path again: your device evidence at `94a58fa` predates it, and the change is in Android snapshot error construction specifically, so the emulator lane is worth re-running before the readiness label. And the PR description still says the timeout policy "identifies accessibility-timeout failures through both helper error types and daemon hint text" — the opposite of what it now does. Happy to update that bullet if you'd rather I edit it than you.

---

thymikee:
Re-reviewed `40f94e31b`: the two requested typed-policy fixes are correct, CI is fully green, and fresh local device evidence passes. One code finding remains.

**P2 — timeout classification still has a second decision site.** `snapshot-helper-session.ts:77,84-88` independently compares `helper.errorType` to `java.util.concurrent.TimeoutException` for `uiAutomationConnectionTimeout`. The new producer already publishes `androidCaptureFailureReason`; this diagnostic must consume that typed value too, otherwise the taxonomy can drift despite the new owner seam. Replace the duplicate classifier with `readAndroidCaptureFailureReason(error)` and add a session-path regression using the existing UI-automation-timeout fixture.

Also update the stale PR summary: it still claims daemon hint-text recognition and has no `## Validation`, the opposite of this heads typed producer design. Keep it brief and remove the Claude session link.

Coordinator exact-head evidence: isolated install/build/build:android passed; iOS Simulator Settings open + compact depth-4 snapshot passed with healthy `tree`; isolated Android `emulator-5556` Settings open + snapshot passed with healthy `android-helper` 0.20.10/API 2 and 172 raw nodes. Both sessions were closed, the emulator stopped, and the detached worktree removed.

No readiness label until the duplicate decision site is removed.

thymikee:
Fixed in `4187f1d72`, and the summary is rewritten.

**The duplicate decision site is gone.** `isUiAutomationConnectionTimeoutResponse` is deleted; the session fallback diagnostic now reads `readAndroidCaptureFailureReason(error)`. Worth naming the irony: I cited that function last round as precedent for the constant, without noticing that leaving it standing *is* the drift I was citing it against — one taxonomy with two deciders. The session protocol already publishes the typed reason on exactly these errors, so there was nothing to add on the producer side, only a reader to correct.

**The regression is proved, not assumed.** It rides the existing `ui-automation-timeout` fixture, so it exercises the real socket response rather than a hand-built error. I checked it actually fires: with the protocol's `androidCaptureFailureReason` attachment removed the test fails, and with it restored it passes. My first attempt at that check was itself invalid — the formatter had collapsed the spread onto one line so my edit silently matched nothing and the suite stayed green, which is exactly what a vacuous gate looks like. Re-ran it against the real text before trusting the result.

**Summary rewritten** — the hint-text claim is gone, there is a `## Validation` section, the session link is removed, and I added a short **Behavior change** note, since a helper reporting `ok=false` with timeout-looking prose under a non-timeout `errorType` is no longer classified as a timeout. That is intended, but it is not a pure refactor and the description should say so.

Local verification at `4187f1d72`: `pnpm check:tooling` green end to end, `check:fallow` clean, 8,001 unit tests passing, with the same two container-only `chmod` failures that reproduce on unmodified `main`.

One note, same as last round: this head changes the Android session path again, so your device evidence from `40f94e31b` predates it.

---

thymikee:
Re-reviewed exact head `4187f1d72553f5b96c996156755ed824026159c0`: clean.

- The duplicate timeout classifier is deleted. The session fallback diagnostic consumes `readAndroidCaptureFailureReason(error)`, so the typed producer is the only taxonomy owner.
- The new session regression rides the real `ui-automation-timeout` fixture and observes the emitted diagnostic; removing the protocols typed-reason attachment makes it fail.
- The rewritten PR body accurately describes machine-value classification and the intentional tightening, includes `## Validation`, and removes the stale Claude link.
- All 20 current checks are green.

Fresh coordinator exact-head evidence in an isolated detached worktree:
- `pnpm install --frozen-lockfile`, `pnpm build`, and `pnpm build:android` passed.
- iOS Simulator (`F7D6F9A4-4FCC-4DD7-AC0B-3280C9319CB9`): opened Settings and ran `snapshot -i -c -d 4`; result was healthy with backend `tree`.
- Isolated Android `emulator-5556`: opened Settings and ran `snapshot -i -c -d 4`; result was healthy with backend `android-helper`, helper 0.20.10/API 2, persistent-session transport, 171 raw nodes.
- Both sessions were closed, the coordinator-started emulator stopped, and the detached worktree removed.

No remaining code or evidence blocker; ready for human merge review.

PR #2018, opened 2026-08-24
docs: define composable recorded fragments

### LINKED ISSUE #1336: Composable .ad scripts: recorded fragments reusable across flows (design spike)
## Motivation — measured, not hypothetical

The E3 repair-economics matrix (36 trials, 2 models × 2 prompts × 3 drift types) established that when a recorded flow drifts, **re-recording is usually cheaper than agent repair** (label drift: re-record $1.21 vs repair median $2.34; cascade: $1.55 vs $2.05; repair wins only on single diverged reads via the corrective-read path, $0.69). Re-record cost scales with flow length — so the unit of re-recording is the lever. If 30 flows share a recorded `login` path and login drifts, today that is 30 re-records; with composable fragments it is one.

This is the direct continuation of the replay thesis: cheap regeneration is the repair path, and composition shrinks the regeneration unit.

## Current state

- `.ad` scripts are closed units: `context` + `open` … `close`, no include/import mechanism.
- The only composition in the system is Maestro-side `runFlow`, handled by the typed engine — with a documented bug class: ADR-0012 records an included file expanding into the linear plan, shifting every subsequent step index and corrupting divergence attribution + the include's own line table. ADR-0015 (migration item 6) moves includes into the interpreter.
- Useful internals already half-exist: recording knows how to slice a session at a boundary (`saveScriptBoundary`), and `replay --from N --plan-digest` provides mid-script addressing.

## Design constraints (load-bearing — from the Maestro scars + the identity model)

1. **Fragments don't own lifecycle.** No `open`/`close` inside a fragment; the composing flow owns session, context line, and launch state. Otherwise reuse implies hidden relaunches.
2. **Recorded entry guards.** A fragment recorded from screen X is only valid at screen X. Fragment capture should auto-record an entry assertion (a `wait`/identity anchor for the starting screen) as the fragment's first step — without it, composition reintroduces the silent-wrong-screen class that `target-v1` annotations exist to prevent.
3. **Stable addressing across expansion.** Divergence already carries `step.source.path:line`; attribution must stay per-fragment (fragment digest + local step index), never a flattened global index, or resume tokens (`--from` + `--plan-digest`) and divergence reports become ambiguous exactly as in the ADR-0012 incident.
4. **One composition model.** Native `.ad` composition and the Maestro interpreter's includes should share semantics; do not grow two parallel schemes.

## Open questions for the spike

- Recording UX: how does a user capture a *fragment* (explicit start/stop markers in-session vs post-hoc slicing of a saved session)?
- Composed plan digest: derivation from fragment digests (content-addressed) so a fragment edit invalidates exactly the compositions that use it?
- Evidence staleness: shared fragments age independently of their callers — does a composition pin fragment versions, and what does drift in a shared fragment report as?
- Annotation compatibility: fragments must keep the reserved-comment property (old runners ignore what they don't know).

## Suggested outcome

An ADR (spike, not implementation) settling the fragment model, addressing scheme, and recording UX — then implementation can be scoped.

thymikee:
## Summary

Defines the ADR-focused design contract for composable recorded fragments from #1336:

- keeps fragment lifecycle ownership with the composing parent for native `.ad` fragments while preserving Maestro's typed subflow interpreter;
- specifies explicit fragment capture, entry guards, positive command admission, publication/close ownership, and recording-conflict behavior;
- defines stable per-fragment addresses, format-owned canonical digests, pinned child composition semantics, source-closure preflight, staleness/repinning, and structured divergence details;
- preserves ADR 0012 replay/repair boundaries, existing Maestro resume behavior, annotation compatibility, old-reader behavior, and an implementation sequence with planted-red proof requirements.

Closes #1336

Scope: 2 tracked documentation files (`docs/adr/0020-composable-recorded-fragments.md` and the ADR index). No production/runtime code, fixtures, CLI surface, or platform behavior changed. `CONTEXT.md` remains unchanged; the detailed vocabulary and decisions are owned by the ADR because the existing glossary has a hard 12,000-byte budget.

Touched-file count: 2. Scope stayed within the requested architecture/design spike.

## Validation

- `pnpm format` passed.
- `pnpm build` passed.
- `git diff --check` passed.
- `pnpm check:affected --run` passed against the current `origin/main` after rebase; the docs-only diff selected no runtime checks. The initial sandboxed run hit the repository guidance scan's `ps` permission restriction and was rerun with the approved Keychain/sandbox escalation; no code failure was observed.
- Multiple read-only `claude -p` adversarial reviews challenged ownership, resume semantics, IDs/digests, old readers, failure reasons, testability, and acceptance criteria; valid findings were incorporated and focused validation was rerun.
- Live-device/OS evidence is intentionally not applicable to this docs-only spike. The ADR records both iOS and Android live record/publish/compose/cold-replay evidence as a required gate for the future implementation; no live-device claim is made here.

### DIFF
--- a/docs/adr/0020-composable-recorded-fragments.md
+++ b/docs/adr/0020-composable-recorded-fragments.md
@@ -1,0 +1,612 @@
+# ADR 0020: Composable Recorded Fragments
+
+## Status
+
+Proposed design spike for [#1336](https://github.com/callstack/agent-device/issues/1336). This
+ADR settles the artifact and ownership contract; implementation is intentionally a later,
+separately reviewable sequence.
+
+## Context
+
+The E3 repair-economics matrix measured re-recording as cheaper than agent repair for label drift
+($1.21 versus a $2.34 repair median) and cascade drift ($1.55 versus $2.05); repair won only for a
+single diverged read ($0.69). That single-read figure is a repair-only baseline; the re-record
+comparison is the label/cascade rows. Re-recording is therefore the repair path, and the useful lever
+is the unit being re-recorded. If thirty flows share a login journey, the reusable unit should be one
+fragment, not thirty closed scripts.
+
+Today a native `.ad` recording is a closed unit: its `open` establishes the session and its `close`
+tears it down. ADR 0016 deliberately stopped at open-to-destination publication and left arbitrary
+slices for this decision because target evidence is captured at action time, not reconstructed from
+history. ADR 0012 also makes the cost of losing composition provenance concrete: a Maestro include
+that was flattened into one plan shifted step ordinals and made divergence attribution and the
+include's line table ambiguous. ADR 0015 now requires Maestro to remain a source-preserving typed
+interpreter that owns `runFlow`, scopes, conditions, repeats, retries, and include loading.
+
+The design must consequently solve four boundaries together: a fragment cannot hide session
+lifecycle, the first action must prove the entry screen, a fragment edit must invalidate exactly the
+compositions that use it, and a failure must name a local step and invocation rather than an
+expanded global ordinal.
+
+## Rules at a glance
+
+1. **A native recorded fragment is a lifecycle-free body.** Its composing flow owns `open`,
+   context, launch state, session ownership, and `close`. A native fragment has no `open` or
+   `close`, and invoking it never relaunches or tears down the session. Maestro subflows retain
+   their typed-flow lifecycle semantics under ADR 0015.
+2. **Native capture is explicit and entry-guarded.** `session fragment start <path> --entry
+   <selector>` captures a verified landmark as the first recorded `wait`; `session fragment save`
+   publishes the body without closing the session. Post-hoc history slicing is not a supported
+   native capture path.
+3. **Composition is one semantic model, not one engine.** Native `.ad` `include` and Maestro
+   `runFlow` produce the same invocation boundaries, provenance records, address construction, and
+   structured failure projection. Format-owned parsers and interpreters retain identity, digest,
+   lifecycle, and execution semantics; neither lowers the other into a second replay engine.
+4. **Native references are digest-pinned.** A persisted native `.ad` include carries a relative
+   locator and the expected content digest of the fragment it resolved when authored. A missing,
+   changed, cyclic, or unsupported native fragment fails preflight before any device action. Maestro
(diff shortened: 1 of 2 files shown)


thymikee:
One design-contract correction is required at `b79c7a16c6ba113b2e4c507cae368886cbea3f52` before this ADR is ready:

`docs/adr/0020-composable-recorded-fragments.md` correctly admits around lines 434–438 that an old reader can execute known root actions before it reaches an unknown `include`, but the implementation sequence around line 573 requires the old-reader include case to fail before any action. Both cannot be true, and the cited `dispatch-keyboard.test.ts` does not currently prove unknown-command dispatch.

Prefer a first-executable format/version guard that old readers reject before root `open` or any other side effect. If legacy prefix execution is intentionally accepted instead, state that explicitly in the acceptance criterion and test it separately from the first-position include case. Point the evidence at the actual unknown-command rejection seam in `src/core/dispatch.ts` (or add the missing focused test).

The remaining ownership split—shared composition/provenance data with Maestro retaining its typed interpreter—is directionally sound.

thymikee:
Addressed in 060b44f2378add8f9f3bb969634b29c82349a449. V1 native fragments and composed parents now require first-executable fragment-v1/composition-v1 guards; new readers consume them in preflight, while old readers reject at src/core/dispatch.ts before any handler, root open, or known prefix. Unguarded legacy prefix behavior is explicitly outside the v1 acceptance contract, and the implementation sequence now requires planted-red proof for both cases. Rebased onto current origin/main (07217c53f). Local pnpm build and pnpm check:affected --run pass; all six PR checks pass; mergeability is CLEAN.

thymikee:
Re-reviewed `060b44f2378add8f9f3bb969634b29c82349a449`. The corrective commit resolves the previous design-contract blocker: v1 native fragments and composed parents now require first-executable `fragment-v1` / `composition-v1` guards before root `open` or `include`. An old reader parses the guard as the first replay action and the existing unknown-command seam rejects it before a platform handler, so no known prefix can execute first. The ADR also explicitly separates unguarded legacy-prefix behavior from the v1 guarantee and requires planted-red future implementation proof with a guarded `composition-v1 → open → include` fixture rather than citing the unrelated keyboard test.

No current-head correctness or design finding remains. This is docs-only, so device evidence is not applicable; all six triggered checks are green and the PR is mergeable/CLEAN. Ready for human review.

Issue #1324, opened 2026-07-17
iOS simulator recorder leaks on session/daemon teardown while recording is active (0-byte recording.mp4, "Host recording is already in progress")


adamTrz:
## Summary

When an iOS-simulator session with an active screen recording is torn down **without a successful explicit `record stop`** — e.g. the daemon is signalled (SIGTERM), reaped, or replaced mid-suite — the recorder process is leaked. The `simctl io <udid> recordVideo …/recording.mp4` child reparents to launchd (PPID 1). Since `simctl recordVideo` only finalizes the mp4 on **SIGINT**, `recording.mp4` stays **0 bytes**, and the simulator's single host-recording slot stays held.

Cascading failures after the leak:
- later recordings (including `test --record-video` attempts) fail with `Host recording is already in progress`;
- `record stop` fails with `COMMAND_FAILED: failed to stop recording … Host recording is already in progress`;
- the iOS runner lease can wedge: subsequent `agent-device test` runs fail instantly with `iOS runner … is already owned by another agent-device daemon`, plus increased transient `RUNNER_BUSY`.

Affects **v0.19.3 and current `main`** (teardown code path is functionally identical).

## Environment

- macOS (Darwin 25.5.0), Node v22.13.1
- agent-device 0.19.3 (global npm) and `main` @ `3fed8acda` run from source
- iPhone 16 / iOS 26.0 simulator; target app irrelevant (reproduced with built-in Settings; originally hit with an Expo dev client)

## Reproduction

`flow-long.ad`:

```
context platform=ios target=mobile
open com.apple.Preferences --relaunch
wait "label=General || text=General" 8000
wait "text=__never_matches__" 15000      # keeps the recorder alive long enough to observe
```

1. `pgrep -fl "simctl.*recordVideo"` → none (baseline)
2. Run in background: `agent-device test flow-long.ad --record-video --device <sim> --artifacts-dir ./artifacts`
3. Poll until the recorder process appears; note the daemon pid (recorder's PPID)
4. `kill -TERM <daemon-pid>` (graceful daemon shutdown)

**Actual** (0.19.3 and HEAD):

```
recorder pid 1244  ppid 466 (= daemon)    # while daemon alive
kill -TERM 466
… daemon exits …
recorder 1244  ppid 1  state SN           # orphaned to launchd
recording.mp4 → 0 bytes
```

[...]

PR #2019, opened 2026-08-24
fix: persist daemon-owned child process records

### LINKED ISSUE #1882: daemon: record owned child pids so leak detection reads a record instead of reconstructing it from ps
Split out of #1781 B1 (#1859), which made the asymmetry concrete.

## The asymmetry

The daemon already writes a durable record for the resources it owns *except* processes:

- screen recordings and app logs get a durable capture descriptor (`sessions/<name>/*.resource.json`, ADR 0019) carrying `lifecycle`, an ownership fence, and — for the simulator backend — the recorder's pid and start time inside `descriptor.body.processes`;
- device claims, runner leases and the shutdown report are all files with owner identity.

There is no equivalent for "processes this daemon spawned and still owns". `runCmdBackground` children (the `simctl io … recordVideo` recorder), the managed agent-browser daemon and its Chrome fleet, and Apple runner `xcodebuild` processes are all owned in practice and recorded nowhere.

## Why it costs

#1859 needed to answer "did this daemon leave a process behind?" and, with no record to read, had to reconstruct ownership from the OS four different ways: PPID descendant, PGID = daemon pid (children reparented to launchd keep the group), `AGENT_DEVICE_STATE_DIR` inherited in the environment, and the state dir as an argv token. Each rule exists because the others miss a real case — the #1324 recorder is caught only by pgid, the #1109 agent-browser fleet only by the inherited environment.

That is ~220 lines of heuristics that cannot run as a lane assertion: the three real-subprocess daemon lanes are device-free, so their daemons own no children and the check compares an empty set to an empty set. It lives in `test/integration/support/daemon-owned-process-probe.ts` as a manual red-proof script, with a fixture test as its only CI guard.

The heuristics are also approximations. A pgid is reserved only while the group has a live member, so on a long-lived host pid reuse can hand the number to an unrelated process; macOS hides the environment of Apple platform binaries from `ps -E`, so simctl is invisible to the env rule.

## What to do

Give owned child processes the same treatment as captures: when the daemon spawns a process it owns, record `{ pid, startTime, command, purpose }` under the owning session (or the daemon root, for daemon-scoped children like the managed browser), and clear the entry when it is reaped. The identity primitives already exist — `readProcessStartTime` / `readHostProcessIdentityObservations` (`src/utils/host-process.ts`) are what the recorder descriptor already uses to prove a pid is still the process it was.

Then leak detection becomes "read the record, assert every entry is dead", which is:

- **shorter** — one rule instead of four, no `ps` reconstruction, no environment inspection;
- **stronger** — it catches a leaked child the heuristics miss (one that setsids away *and* whose argv/env carry no state-dir path), and it cannot false-positive on a pid-reused group;
- **assertable in every lane**, including the device-free ones, because an empty record is a meaningful assertion rather than a vacuous one;
- a **startup-reaper input**: the same record tells a fresh daemon what a dead predecessor left behind, which is what #1109's orphan cleanup infers today from Chrome command lines.

## Relationship

#1431 wants exactly this as its observability half: injected faults are only meaningful if the resulting leak is observable. #1859 landed the arms that read a record (surviving daemon, unfinalized capture handle, state-dir residue); this issue is what lets the process arm join them.

Acceptance: an owned-child record exists and is written/cleared on the spawn and reap paths; `daemon-owned-process-probe.ts` and its fixture test are deleted in favour of a record-reading rule inside `daemon-leak-model.ts`; the #1324 and #1109 scenarios still go red against their pre-fix behaviour with the new rule.

thymikee:
## Summary

Closes #1882.

- Adds atomic daemon/session-owned process records with `{pid,startTime,command,purpose}`, exact-identity startup reaping, and spawn/reap wiring for simulator recording and managed agent-browser processes.
- Reuses the existing typed Apple recording descriptor and live `appleSimulatorRecordingCommandMatches` seam. The simulator host refreshes the root identity after output readiness, when `xcrun` has exec'd CoreSimulator's `simctl`, so the generic exact-identity reaper receives the stable post-exec identity.
- Makes managed-web record clearing contingent on semantic JSON success, so an exit-0 `{success:false}` close cannot erase the sole durable record.
- Replaces `ps` reconstruction in the daemon leak oracle with record-reading, exact identity checks, and fail-closed ownership-loss reporting.
- Adds device-free #1109/#1324 leak-model assertions, a real-child iOS recording fixture trace, wrapper-to-exec and semantic-close regression tests, and record publication/clear regression tests.
- `daemon-owned-process-probe.ts` is already absent on current `main` (removed by #1859), so this diff does not recreate or delete it; the replacement is the record-reading rule in `daemon-leak-model.ts`.
- Rebased onto current `main` (including #2012's web-shutdown teardown and shared timeout constant). Scope is 30 touched files, limited to the owned-process contract/recording owners, managed web lifecycle, and leak-model/oracle coverage. The existing Apple runner xcodebuild lease remains its separate durable owner seam; this change does not duplicate it.
- No docs or skills changed: behavior is owned by the runtime contracts and existing test procedures.

## Validation

- `pnpm check:affected --run` — all runnable checks passed: format, lint, typecheck, layering (including durable-owner rules), fallow, build, Node integration (92 passed / 9 expected skips), and related Vitest (310 files, 2,090 tests). The first sandboxed attempt stopped only because its leak-model preflight could not invoke `/bin/ps` (`spawnSync ps EPERM`); the identical gate passed with host process visibility.
- Focused follow-up: 3 files, 32 tests passed for Apple simulator recording, managed web provider, and the durable record store.
- Planted-red evidence: temporarily removing the post-exec identity refresh made `publishes the post-exec simulator identity after readiness` fail; temporarily removing semantic-success fencing made the exit-0 `{success:false}` close regression observe `clear` being called. Existing #1109/#1324 record-reading and publication tests also have planted violations and fail closed.
- macOS-host native evidence is now available. On a booted iPhone 17 Pro simulator, the built CLI `record start --scope device` persisted the live CoreSimulator `simctl io ... recordVideo` identity (the exact `ps` command matched it); `record stop` exited successfully, produced the MP4 and gesture telemetry, cleared `owned-processes.json`, and left the recorder PID dead.
- Native crash/startup-reap evidence: after a real recording, killing the daemon left the recorded CoreSimulator PID and durable record in place; the next daemon startup through `devices` read the record, reaped the recorder, and cleared the record. The crashed session was then cleaned up.
- `pnpm vitest run test/integration/provider-scenarios/ios-record-trace.test.ts` — 2 passed; the fixture starts a real child through the injected simulator transport and verifies record write/clear.
- `node --experimental-strip-types --test test/integration/daemon-leak-oracle-cli.test.ts` — passed, including the exact daemon-owned process record case.
- Adversarial review attempt: the approved `claude -p` invocation was made; the sandbox reported not logged in, and bounded Keychain-backed retries produced no output and were stopped. No Claude findings were available; local diff audit, maintainer findings, and planted-red checks were completed.

## Design and size accounting

Against `origin/main` at `07217c53f`, the diff is 30 files, +1,733/-226. The regenerated size report is +7.5 kB JS/unpacked, +2.3 kB gzip, and +2.2 kB tarball.

- Runtime/package paths: +906/-165 across 16 files (net +741). The durable record store is 228 lines and the generic reaper is 217 lines (+445 total) for atomic versioned daemon/session state, path-safe validation, PID/start-time/command fencing, purpose filtering, escalation, and startup cleanup. The managed-web owner adds 185 lines for process-table observation at the provider lifecycle seam; provider wiring is +93/-12 and lifecycle extraction is +43/-136. The remaining +140/-17 is typed contract, daemon wiring, and Apple host/descriptor integration.
- Test/fixture/oracle support: +827/-61 across 14 files, including the device-free model/oracle rules, CLI exact-record coverage, provider scenarios, fixture wrapper-to-exec proof, and semantic-close regression.
- A smaller owner-specific implementation was rejected because it would duplicate atomic file format, validation, startup reaping, escalation, and daemon/session scope semantics between Apple and web. The shared record/reaper owns only durable identity mechanics; Apple retains the typed `AppleRecordingDescriptor` and purpose-owned live matcher, while web retains the narrow process-observation owner. A generic purpose/command matcher would type-erase semantic identity and caused the original simulator P0.

## Evidence gaps

- The #1324 macOS simulator lane was available and exercised locally with real `xcrun simctl` start/stop and crash/startup-reap evidence. The fixture-level wrapper-to-exec stand-in remains explicit and is the planted-red proof for the pre-exec identity race.
- No physical iOS device, Android/Harmony/Vega target, or managed Chrome web E2E was available locally for this change. The repository's GitHub native/device and `AGENT_DEVICE_WEB_E2E=1` lanes remain authoritative for those live OS/provider gaps.

Touched-file count: 30. Ready for review; do not merge.


thymikee:
Changes requested on `9fc6b02`.

1. **P0 — normal iOS simulator `record stop` is broken.** Exact-head source evidence on iPhone 17 Pro: `record start` succeeded and persisted `/bin/bash …/usr/bin/simctl io …`, but the live PID exec'd to CoreSimulator's `/Library/Developer/…/bin/simctl io …`; `record stop` then failed with `simctl recordVideo process ownership changed before cleanup` (`mt7peeyz-e24f3d72`). Identity is captured pre-exec in `packages/platform-apple/src/recording/runtime.ts:92-102`, while `src/platform-runtime-screen-recording-apple-simulator-host.ts:22-34` and the generic reaper reject the post-exec identity. Use a typed/purpose-owned matcher shared by live cleanup and startup recovery, or capture stable post-exec identity. Add a planted wrapper→exec regression plus real CLI start/stop and crash/startup-reap evidence.

2. **P1 — web close can erase the sole durable record on semantic failure.** `agent-browser-provider.ts:273-280` finalizes before `unwrapAgentBrowserJson`; `canClear...` treats exit 0 as success, so an exit-0 `{success:false}` close clears the record while the browser may remain. Clear only after semantic JSON success (or re-record identities), with a regression for that case.

3. **Required size/design accounting.** This is +1,648/-214 total; production net is roughly 900–1,000 lines and Size CI reports npm unpacked +7.1 kB. Please itemize justified retained growth and explain why the smaller owner-specific design was rejected: reuse the existing typed simctl capture descriptor/matcher for simulator recording and keep a narrow web-owned durable record. The current generic `{purpose:string, command}` layer type-erases semantic identity and already caused finding 1.

Focused exact-head build and 67 tests passed; they do not cover the two failure modes above. CI is still pending, so no readiness label.

thymikee:
Addressed the three requested review points on `2fc31c22f`:

1. **#1324 / simulator P0:** the Apple simulator host now refreshes the root identity after recording output is ready, so the durable record stores CoreSimulator's post-`xcrun` `simctl` command. The planted wrapper→exec test goes red without that refresh. A real macOS-host simulator run now passes `record start` → `record stop`, clears the record, and leaves no recorder PID; a real daemon-kill → next-startup run also reaped the recorder and cleared the record.
2. **Web semantic failure:** record clearing now requires semantic JSON success in addition to exit 0 and command completion. The planted exit-0 `{success:false}` close regression goes red without the guard and passes with it.
3. **Design/size accounting:** the PR body now itemizes the current rebased diff (+1,733/-226 across 30 files), runtime versus test support, the +7.3 kB unpacked/+2.2 kB gzip size report, and why shared durable mechanics plus owner-specific Apple/web identity seams are safer than a type-erased generic matcher.

The branch is rebased onto current `main`; local `pnpm check:affected --run` is fully green (including 310 related Vitest files / 2,090 tests). GitHub checks for the new head are running now.

thymikee:
CI follow-up: the first iOS smoke attempt hit an unrelated live-runner TEXT_INPUT_COMMIT_NOT_OBSERVED failure in the existing smoke:form-input scenario (empty xcodebuild output); no text-input/XCTest code changed here. Rerunning the failed job passed end to end in 13m47s. All PR checks are now green, including Android, iOS, macOS, Linux, coverage, CodeQL, size, package, and repo guards. The regenerated size report is +7.5 kB JS/unpacked, +2.3 kB gzip, and +2.2 kB tarball; the PR body has the corrected figures.

thymikee:
Re-reviewed `2fc31c22`: clean and ready for human review.

The two prior correctness blockers are resolved at their owning seams: Apple re-reads the recorder root identity after `xcrun` execs into `simctl`, so explicit stop and startup reaping use the stable post-exec identity; web durable ownership is cleared only after semantic close success, so exit-0 `{ success: false }` retains the record. PID/start-time/command fencing remains fail-closed, and purpose-specific identity stays in the platform owners rather than leaking into the shared record layer.

The large-change threshold is triggered (+7.5 kB unpacked; roughly 741 net runtime lines), but the growth is itemized and the shared durable-record/reaper construction path is materially smaller and safer than duplicating lifecycle logic. Current checks are all green, planted-red regression evidence is recorded, and the PR includes exact-head macOS Simulator start/stop plus daemon-restart reaping evidence.

No actionable code findings. Residual coverage is limited to managed-web/provider and physical-device paths not independently reproduced here.

PR #2023, opened 2026-08-25
Harden the MCP surface: registry rug-pull fix, operator-only credentials/endpoints, device-shell argv gate, declared timeouts

thymikee:
Addresses the MCP registry security assessment findings that can land without a release. Four commits, in the assessment's suggested order.

## 1. AS-012 "rug-pull" — the version on `main` never equals a published version (`e8a9abd`)

The scanner's finding is a Git-tag problem, not npm: `main` carried `0.20.10` twice with different tool sets (`hover` added between the two release commits, `read` removed after), so two scans of "0.20.10" saw two different surfaces. The published tarball is clean.

- `release:publish` now runs `release:mark-dev` right after `npm publish`, moving `package.json` + synced `server.json` to the next patch with a `-dev` marker (`0.20.11-dev`).
- `release:prepare` refuses to publish while the `-dev` marker is in place, so a forgotten version bump can't ship a prerelease as `latest`.
- This PR applies the marker to `main` now — it was sitting on the published `0.20.10`, which is the live finding. The MCP-registry publish workflow only fires from release tags, so a `-dev` version on `main` never reaches the registry.

## 2. Credentials off the model-writable tool surface (`d1722b3`)

Every MCP tool advertised `daemonAuthToken` (and Metro tools `bearerToken`) as free-form strings the model writes. agent-device feeds untrusted app UI text to the model that picks tool arguments — a model-writable credential parameter is a prompt-injection exfiltration path.

- Both keys are removed from every advertised schema (MCP and AI SDK share `listCommandTools()`).
- An explicitly passed value is **refused with env guidance** (`AGENT_DEVICE_DAEMON_AUTH_TOKEN`, `AGENT_DEVICE_METRO_BEARER_TOKEN`) rather than silently dropped — the retired-field posture.
- Operator resolution is untouched: env/config defaults still merge, the daemon and Metro clients keep their env fallbacks, CLI flags unchanged.

## 3. Operator endpoints and paths follow (`c434376`)

`daemonBaseUrl` and Metro `proxyBaseUrl` are the endpoints the env-resolved tokens are *sent to* — a model-writable value redirects the operator's token to an arbitrary server (`resolveProxySettings` falls back to `AGENT_DEVICE_DAEMON_AUTH_TOKEN` whenever a proxy URL is set). `stateDir`, `cwd`, `iosSimulatorDeviceSet`, and the three `iosXctest*` paths select operator infrastructure, never per-call work. Same treatment: not advertised, refused as explicit input, env/config-only. Explicit input can also no longer *override* operator config. Dropping these shared properties plus the credentials also cuts the `tools/list` payload substantially (the assessment measured ~58% for the full shared-property set).

Left in place deliberately: `bundleUrl`/`metroHost`/`publicBaseUrl` on Metro reload — no credential is attached to them; happy to fold them in if you want them gone too.

## 4. `check:shell-argv` — argv reaching a device shell is now enumerated (`4276d32`)

`adb shell` / `adb exec-out` / `hdc shell` evaluate argv as one shell string on the device — the class of bug the internal audit found on `input text` / `cmd clipboard set text` (both quote through `shellQuoteIfNeeded` now), with nothing to stop a new call site regressing it.

New AST-based gate (`oxc-parser`, mirroring `scripts/di-seams`): `scripts/shell-argv` keeps an exact inventory of every dynamic device-shell argv element — **121 values today**, keyed by `(file, expression)` with counts. A new or grown entry fails CI until it's quoted or deliberately recorded with `--update` in the same PR; stale entries fail the other way. Wired into the lint lane via `run-gate` and registered in the check catalog (gate manifest audit passes: 51 checks / 28 lanes).

## 5. AS-011 answered — timeouts declared on the tool surface (`c434376`)

The envelopes always existed (`timeout-policy.ts`); they just weren't declared. Every tool description now ends with its enforced client envelope ("Times out after 90s." / 180s install / "a caller-supplied budget extends it" / streaming `test` declared unbounded), sourced from `resolveCommandTimeoutPolicy` so the declared number cannot drift from the enforced one. A test pins every description to its policy.

## Not in this PR

- **CI publish with `--provenance` (assessment item 4)**: needs npm-side trusted-publisher/token setup only a maintainer can do (npmjs.com settings + repo secrets). The release-script changes here don't conflict with it; suggest tackling it as its own PR once the npm account is configured.
- **AS-014** is a declaration gap per the assessment (repository is present in `package.json` and `server.json`) — nothing to change.

## Verification

- `format:check`, `lint`, `typecheck`, `check:mcp-metadata`, `check:gate-manifest(:test)`, `check:affected:test`, `check:shell-argv` all green.
- Full unit suite: 1052/1054 files pass. The two failures (`durable-capture-resource-adoption`, `app-log-session-resource`) fail identically on a clean checkout of `main` in this environment — pre-existing, untouched by this PR.
- New tests: operator-input refusal + env-default flow, credential refusal, schema absence across all 56 tools, timeout-note-vs-policy pinning, shell-argv model unit tests. The `command-tools.test.ts` size-ratchet pin ratchets down 1218 → 1216.

https://claude.ai/code/session_01Wb8EuoySL26UsdtCtRj5X8

---
_Generated by [Claude Code](https://claude.ai/code/session_01Wb8EuoySL26UsdtCtRj5X8)_


thymikee:
Not ready — two P1 security blockers.

1. MCP operator-input bypass via `config`. `src/mcp/router.ts` forwards arbitrary `tools/call` arguments without validating the advertised schema. `OPERATOR_INPUT_GUIDANCE` omits `config`, and `resolveMcpConfigDefaults` treats raw tool input as CLI flags. An explicit config may contain operator-only `daemonBaseUrl` / `daemonAuthToken`; coordinator reproduction passed `{config: '/private/tmp/...json'}` to `snapshot` and observed both values reach `runCommand`, including an attacker endpoint and token. Hidden-from-`tools/list` is insufficient because router admission does not enforce it. Make one shared typed MCP admission boundary reject all unadvertised raw keys, including `config`/`remoteConfig`, before config resolution; only serving-process env/config may supply operator values. Add a router-level regression with a redirect config and ambient token proving the command route is never invoked.

2. `check:shell-argv` is an inventory, not a security invariant, and has a trivial blind spot. It allows any unsafe site to be self-approved with `--update`; it only detects literal-first `['shell', ...]` / `['exec-out', ...]` arrays. Coordinator reproduction: direct `adb(['shell','input','text',text])` inventories `text`, while `const subcommand='shell'; adb([subcommand,'input','text',text])` returns no finding. Move enforcement to the device-shell execution boundary with typed quoted/validated atoms and narrow explicit exceptions; do not let inventory update approve raw values. Add planted-red tests for raw text and indirect/variable-built shell argv.

Positives: descriptor-derived timeout notes and direct field removals are sound; CI/CodeQL/package/size checks are green; no device evidence is needed for this static MCP/security change.

thymikee:
Re-reviewed `c67010c`. The MCP P1 is resolved: `createCommandToolExecutor` now rejects unadvertised raw keys before config resolution/command execution, the router and AI SDK share that executor, and the regression proves attacker `config`/`remoteConfig` values never reach the route.

**P1 — device-shell argv enforcement is still missing.** The local adb execution boundary in `src/platforms/android/adb-executor.ts` still accepts arbitrary `string[]` and forwards them directly to `runCmd`; there is no typed shell atom, quoting validation, or equivalent execution-boundary guard. A variable-built shell argv containing `userText` can still reach the device shell unchecked. This head deletes the AST inventory, tests, and CI gate, but does not replace them with the required owning-boundary invariant. Add typed quoted/validated shell atoms (with narrow explicit exceptions) at the actual device-shell execution boundary and planted-red tests proving raw free-form text and indirect/variable-built argv are rejected. Do not permit an inventory update to approve raw values.

Checks are still in progress. No `ready-for-human` label while this P1 remains.

thymikee:
Both P1s verified and addressed.

**1. Operator-input bypass via `config` — fixed (`cdc1aca`).** Confirmed your repro first: `{config: <path>}` on `snapshot` put the file's `daemonBaseUrl` and `daemonAuthToken` straight into `runCommand`. Root cause was exactly as you described — the router forwards raw `tools/call` args verbatim and `resolveMcpConfigDefaults` reads them as CLI flags, so hiding keys from `tools/list` did nothing.

Replaced the per-key operator refusal with a **deny-by-default admission boundary** in the shared `createCommandToolExecutor` (the one path the router *and* the AI SDK adapter both use): every raw argument must appear in the tool's advertised `additionalProperties:false` schema, or it's refused **before** config/env resolution. That closes `config`/`remoteConfig`, the operator keys, and any unknown key in one rule. Operator env/config defaults still resolve — they never arrive as tool input. Retired keys (`maxSize`) stay admitted so the command's own reader still answers with migration guidance (exposed as `metadata.retiredInputKeys`). Regressions added: the config repro, `remoteConfig`, an unknown key, and the preserved env-default flow — all asserting nothing reaches the command route.

**2. `check:shell-argv` — removed (`c67010c`); real fix scoped as #2026.** You're right that it's an inventory, not an invariant — I reproduced the indirect blind spot (`const s = 'shell'; adb([s, …])` → no finding), and `--update` self-approval is a hole. Shipping it security-framed is false assurance, so it's out; this PR is now MCP-surface-only. The two known-dangerous sites already quote on `main`, so no regression.

The sound replacement is a **typed device-shell execution boundary** — a branded `ShellSafe` in `packages/contracts` (mirroring the `screenshot-geometry.ts` idiom) threaded through the shell path of the three funnels (`runAndroidAdb`, `runHarmonyHdc`, the limrun host port), so a raw string can't reach a device shell; `markShellSafe()` is the narrow explicit exception, backstopped by a di-seams-style gate guarding the cast. It's a ~188-site cross-package migration that straddles the in-flight ADR 0019 `src/platforms/*` → `packages/platform-*` move, so it belongs on its own — written up with the exact anchors and sequencing in **#2026**. Happy to start it immediately, or hold for your read on the design there.

---

thymikee:
Thanks — agreed on the MCP P1 being resolved. On the device-shell P1, I'm asking to **supersede the "must land in this PR" finding** and move the whole thing to #2026. Reasoning, after building it:

**I built the mechanism and it works.** On `claude/shell-safe-boundary-wip` (`eb1449f`): a branded `ShellSafe` in `packages/kernel` with `sh.lit/lits/arg/num/raw` atoms; `runAndroidShell`/`runAndroidExecOut`, `runHarmonyShell`, `runLimrunAndroidShell` funnels typed to `ShellArgv`; guards at the adb executor, hdc, and limrun boundaries that reject raw `['shell'|'exec-out', …]`; an AST backstop (`scripts/shell-safe`, di-seams-style comment approval) holding every `sh.raw` and `as ShellSafe` cast to a reviewed set; and planted-red tests proving raw argv, **variable-built argv**, and the resolved-executor path are all rejected, plus a `@ts-expect-error` that raw strings don't typecheck. `sh.arg` is the identity on the safe charset, so migrated argv is byte-identical for legitimate input — only genuine injection vectors change.

**But the migration is ~80 files and it must land with the mechanism, not after it.** Roughly 50 source files plus ~30 test files that mock the funnels and assert the legacy `['shell', …]` shape, plus the `packages/platform-*` injected host-port sites. Landing a partial state here would be worse than landing nothing: a typed seam that existing callers bypass is false confidence, enabling the guards before migrating every caller risks runtime breakage, and a half-migration leaves two execution models across paths that ADR 0019 is actively moving from `src/platforms/*` into `packages/platform-*`. Completing all ~80 rewrites inside this PR would also swamp a focused MCP security fix that's otherwise ready.

**Why deferring is safe here:** the exposure was *discovered* during this PR but is not introduced by its surviving MCP changes — this PR's diff contains no device-shell code at all now — and the two known-dangerous sites (`input text`, `cmd clipboard set text`) are already quoted on `main`. So there's no regression and no known live exposure gated on this PR; what's missing is the structural invariant, which is exactly what #2026 delivers.

**#2026** is now written up as one dedicated PR with the command-atomic commits you'd want (android → harmony → limrun → injected host-port/package callers → delete every legacy raw-shell route and enable the final completeness gate), the WIP branch as its starting point, and the obstacles I hit documented: the `...WithAdb` injected-executor helpers exported to the SDK (a `runAdbShell(adb, ShellArgv)` helper avoids changing the executor contract), `emulator-lifecycle.ts:157`'s direct `runCmd`, the `packages/platform-*` host-port argv typing, `options.adb`-injecting tests needing `withAndroidAdbProvider`, and one real fix worth keeping (`perf-native-process.ts` interpolated `tool` unquoted). The final gate will prove no bypass including variable-built argv and casts.

Also filed: #2027 (unify the input-field audience model — `retiredField` / AI-SDK hidden / the new admission boundary are now three mechanisms for one concept) and #2028 (publish from CI with `--provenance`, needs npm trusted-publisher setup).

This PR is back to MCP-only and green locally. Agreed the iOS smoke failure is classified separately — happy to dig into it if it's still red on the current head.

---

thymikee:
Classifying the iOS smoke failure, as agreed.

**It is not this PR's.** The failing step is `agent-device fill id="field-name" "Ada Lovelace"` in `smoke:form-input`, erroring `TEXT_INPUT_COMMIT_NOT_OBSERVED` — "the runner could not confirm the typed text reached the field", with `xcodebuild.exitCode: 1` and empty stdout/stderr. 14 of 15 tests in that job passed, and the other four Smoke Tests jobs plus all 21 remaining checks are green on `c67010c`.

This PR's diff is now 16 files and contains **no iOS, XCUITest-runner, text-input, or device code at all**:

```
CHANGELOG.md  CONTRIBUTING.md  package.json  server.json
scripts/release-mark-dev.mjs  scripts/__tests__/test-file-size-ratchet.test.ts
src/__tests__/npm-package-scripts.test.ts  src/utils/__tests__/daemon-client-lifecycle.test.ts
src/commands/{command-contract,command-input,field-command-contract}.ts
src/mcp/command-tools.ts  src/mcp/__tests__/*  src/mcp/__tests__/command-tools-parity.test.ts
```

There is no code path from an MCP tool-schema admission check to the XCUITest runner's text-commit observation, so this is either an unrelated flake or a pre-existing condition on the runner. The shape (a runner-side commit-observation timeout with an empty xcodebuild error) matches the flaky class rather than a deterministic break.

I don't have a macOS device lane here to reproduce or confirm it against `main`, so I'd rather not guess further: if you can re-run that one job, a green result settles it; if it reproduces on `main` at the same commit base, it's a pre-existing runner issue worth its own issue rather than a blocker on this PR. Happy to dig in either way if you'd like it investigated.

---

thymikee:
Scope decision accepted. The typed device-shell interface, escape-hatch gate, planted-red proof, and complete production/test caller migration move together to #2026; no mechanism-only or partial state belongs in this MCP-focused PR. This supersedes the earlier shell-boundary P1 for #2023 because the surviving diff introduces no device-shell path and the known dangerous sites are already quoted on main.

The MCP admission fix is clean at `c67010c`: raw unadvertised/config-loader keys are rejected before config resolution and command execution, while operator env/config flow remains intact. No remaining code findings.

The first iOS smoke attempt failed unrelated `fill full name` with `TEXT_INPUT_COMMIT_NOT_OBSERVED`; all other checks passed and a rerun is now in progress. This is classified as an infrastructure flake, not an owner-action or code failure. `ready-for-human` is appropriate while the rerun proceeds; merge only after the current check finishes green.

thymikee:
Re-run result: **failed identically** — same step (`fill full name`), same code (`TEXT_INPUT_COMMIT_NOT_OBSERVED`), same scenario (`smoke:form-input`), 14/15 passing. So it is not a flake that clears on retry. It is also not this PR's, and I can now point at the tracked cause rather than just asserting scope.

**This is #1874** (open, reopened) — titled *"fails intermittently **on unrelated branches** (blocking ios.yml PR gate)"*, and its body specifies the symptom as failing "on branches that touch no Swift and no iOS production code". That is this branch exactly.

The chain:
- #1874's root-cause hypothesis is environmental: the simulator's *Connect Hardware Keyboard* state persists per device and **is not pinned by the job**; when it flips, a focused responder shows the software keyboard and the text-commit precondition fails.
- #1874 already flags the fixture-backed iOS smoke as *"possibly the same root cause ... the same bare-type-with-hidden-keyboard behaviour, one layer up"* — i.e. this smoke scenario.
- The `TEXT_INPUT_COMMIT_NOT_OBSERVED` code itself comes from #1924, which made the runner **report an unobserved commit instead of a partial success**. So this error is that honest-reporting change doing its job on the #1874 environment condition — not a new regression.

Scope evidence for this PR: the diff is 16 files — `CHANGELOG`/`CONTRIBUTING`/`package.json`/`server.json`, `scripts/release-mark-dev.mjs` (+test), the three `src/commands/*command*contract/input` files carrying `retiredInputKeys`, `src/mcp/*` (admission boundary + tests), and one daemon-client **test** file. Zero iOS, Apple-runner, interaction, or text-input code. Nothing here can reach the XCUITest text-commit path.

I've used the one allowed confirming re-run and will not re-run again, and I'm not pushing iOS changes into an MCP-only PR to chase a lane the diff can't influence.

Your call on the gate: merge despite this known-red lane, or pin the simulator keyboard state per #1874's "Wanted" section in its own PR. I've added these two run IDs as data points on #1874.

---

PR #2030, opened 2026-08-25
refactor(android): admit text onto one resolved channel

thymikee:
Follow-up to #2003. Collapses the text-entry routing that #2003 spread across `typeAndroid`/`fillAndroid` — three predicates × two entry points, with the helper package plumbed differently per source — into one module-private seam.

## Summary

- **One resolver owns the whole decision.** `admitAndroidTextChannel(device, action, text)` returns a discriminated union: `{backend: 'test-ime', packageName}` (package already resolved from either provenance — the activation cache answers spawn-free off the artifact manifest; the observed route reads it off the device) or `{backend: 'adb-shell'}` — returned only after the ASCII and input-ownership asserts pass. Callers make one call and one dispatch; the read-before-refuse ordering (#2003's F-fix) and both shell asserts are unreachable from outside the resolver, so a caller cannot reintroduce the refuse-before-read defect or run the asserts against a stale read.
- **One focus tap per helper fill.** `fillAndroid` resolves the channel inside its attempts loop after `focusAndroid`, so the helper fill's first attempt reuses that focus; only the retry re-focuses. This deletes the pre-loop activation-cache special case *and* removes a real double-tap: on main, the observed-IME fill route (helper active, cache empty — a crashed run or a second daemon) taps the target twice. Activated and shell routes are spawn-identical to main, including the per-attempt admission re-read and mid-fill promotion to the helper.
- Tests pin both halves of the focus contract: one tap on first-attempt helper fills, and exactly two taps when the first attempt fails verification (the not-yet-bound-InputConnection retry).

## Live evidence (exact head `fbb146241`, Pixel_9_Pro_XL emulator, Android 16 / API 36)

Spawn counts from `--debug` request ndjson (`exec_command` events), same staging on a main-built (`b5adf966b`) daemon as A/B where the sequence changed:

| Route | main | this head |
|---|---|---|
| Activated fill, Unicode (`open` auto test-IME) | — | 1 `input tap`, 2 `am broadcast` (clear+commit), 0 `input text`, 0 routing reads; verified |
| Activated type | — | 1 broadcast, 0 reads |
| **Observed fill**, Unicode (helper set via `ime set`, fresh daemon, `open --no-test-ime`) | **2 taps**, 2 broadcasts, 1 routing + 2 verification `dumpsys input_method` | **1 tap**, otherwise identical; verified |
| Observed type | — | 1 routing read, 1 broadcast (Unicode accepted with empty cache) |
| Shell fill, ASCII (stock LatinIME) | — | 1 tap, 2 DEL batches, 3× chunked `input text` (23 chars), backend `adb-shell`; verified — unchanged shape |
| Close restore | — | helper → previous IME restored; device left on stock with no stray records |

## Validation

- android + daemon suites green at this head (2,771 tests), `tsc` clean, oxlint clean, `fallow audit` clean on changed files.
- `check:affected` maps the change to the android device lanes; the live table above covers the device-facing sequences at exact head.

### DIFF
--- a/src/platforms/android/__tests__/text-input-test-ime.test.ts
+++ b/src/platforms/android/__tests__/text-input-test-ime.test.ts
@@ -124,6 +124,11 @@ test('fillAndroid clears then commits non-ASCII text through the test IME and ve
     false,
     'the ASCII shell input path must not run while the test IME is active',
   );
+  assert.equal(
+    calls.filter((args) => args[1] === 'input' && args[2] === 'tap').length,
+    1,
+    'one focus tap serves both channel resolution and the first helper attempt',
+  );
 });
 
 // The device, not this daemon process, decides which IME is active. A previous run (or another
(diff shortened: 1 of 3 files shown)


thymikee:
Reviewed `fbb1462` holistically. Clean — no confirmed code findings.

The private `admitAndroidTextChannel` is a materially deeper owning seam: it collapses duplicated cache/observed-IME/shell admission across type and fill, returns a small discriminated result with the resolved helper package identity, keeps provider-native injection outside the seam, and net-deletes production branching. The production route and retry semantics remain intact; the intended observed-helper focus count drops from two to one while a failed verification still re-focuses on retry.

Regression proof is valid: targeted suites pass 28/28, and planting the pre-refactor unconditional inner focus makes both new tap-count regressions fail. The PR's exact-head Pixel 9 Pro XL / Android 16 A/B evidence covers observed helper, activated helper, shell fallback, and cleanup.

`ready-for-human` is appropriate at this head. Remaining authoritative checks were still pending at review time, so merge only after they complete green.

Issue #2031, opened 2026-08-25
web: a stale device lease outlives its session — DEVICE_IN_USE names a session `session list` does not report


thymikee:
Split out of #2022 as triage requested — the lease/recovery seam is owned separately from the
managed-backend spawn fix.

## Summary

After a daemon is replaced or orphaned, the orphaned daemon keeps holding the web-device lease. In
the report the sequence was:

```
Replacing daemon (pid …, v0.20.10): code-signature mismatch
Daemon request timed out
```

and from then on every web command failed with:

```
Error (DEVICE_IN_USE): web device agent-browser-chrome is owned by session "cwd:…:default"
```

while `agent-device session list` reported `{"sessions": []}`.

## Impact

The error names a session the CLI itself says does not exist, so there is no documented action that
clears it. Killing the managed browser process does not release the lease; only killing the
orphaned daemons does.

## Environment as reported

Windows 11 (10.0.26200), Node v24.13.1, agent-device 0.20.10, agent-browser 0.27.1. It surfaced
while working around #2022, but nothing about lease/session reconciliation looks Windows-specific.

## Possible directions

- Reconcile the device-lease check against live sessions, so a lease whose owning session no longer
  exists cannot block acquisition.
- Or let `close` accept the device serial when no session is resolvable, giving an explicit
  recovery path.

Originally reported by @strongishdeveloper in the "Minor, same area" section of #2022.

NicolasBataille:
Same class of problem on **agent-device 0.20.10 / iOS 26.5 simulator** (Xcode 26.6, Node v23.10.0) — adding a reproducer in case it helps narrow it down.



**Environment.** agent-device 0.20.10 (local npm install, not `npx`), macOS Darwin 25.5, Xcode 26.6,
iOS 26.5 simulator, Node v23.10.0.

**Steps to reproduce.**

```bash
agent-device open com.example.app --platform ios --device "iPhone 17" --relaunch
agent-device close                                                   # rc 0, 3.6 s
agent-device open com.example.app --platform ios --device "iPhone 17"
```

**Observed.** `close` succeeds:

```
Closed: default
```

The next plain `open` then fails, twice (12.8 s and 10.8 s):

```
Error (DEVICE_IN_USE): Device is already in use by session "default".
Hint: Run agent-device session list to inspect active sessions. To reuse this device, rerun the command with --session default. To open a new session on this device, first run agent-device close --session default.
```

[...]

PR #833, opened 2026-06-19
feat: manage web backend setup

thymikee:
## Summary

Makes agent-browser a managed web backend implementation detail instead of a user-installed global prerequisite.

Details:
- adds agent-device web setup|doctor for explicit setup and health checks
- resolves web runtime only from the pinned managed install under the effective agent-device state dir
- requires explicit setup before first web automation; runtime --platform web commands fail with a setup hint instead of installing implicitly
- scopes managed backend files under the agent-device state dir, with a short runtime HOME symlink so agent-browser socket paths stay below macOS limits
- passes the daemon state dir into the web Provider
- runs GitHub Actions Node jobs on Node 24.13 through the shared setup action and direct release/publish workflows
- keeps browser network/CDP out of this PR; those need separate command-surface design because network already has native app-log semantics and CDP belongs with #823

Touched-file count: 20 files. Scope stayed within web setup/runtime plumbing, web CLI help/docs, CI Node setup, and focused tests.

## Validation

- pnpm exec vitest run src/platforms/web src/utils/__tests__/args.test.ts
- pnpm typecheck
- pnpm lint
- pnpm check:tooling
- pnpm test:unit
- pnpm test:coverage
- pnpm check:fallow --base origin/main
- /Users/thymikee/.nvm/versions/node/v22.22.2/bin/node ./node_modules/vitest/vitest.mjs run src/platforms/web/agent-browser-tool.test.ts --project unit
- SKILLGYM_ALLOW_EXTERNAL_RUNNERS_IN_SANDBOX=1 pnpm test:skillgym:case web-first-use-runs-managed-setup (2/2 runners passed)
- node --experimental-strip-types src/bin.ts help web
- node --experimental-strip-types src/bin.ts web status --state-dir /private/tmp/ad-web-no-status --json returns INVALID_ARGS
- daemon-backed missing-backend smoke with fresh state dir and PATH excluding agent-browser: open --platform web returned TOOL_MISSING and did not create tools/agent-browser
- daemon-backed managed backend smoke: setup reuse, open, snapshot, close, and doctor were verified during the subagent/manual pass

PR #2033, opened 2026-08-25
fix(web): launch npm and the managed backend through node, not .cmd shims

### LINKED ISSUE #2022: web: spawn EINVAL on Windows — the managed agent-browser backend is a .cmd spawned with shell:false
## Summary

On Windows, no `--platform web` command works. `agent-device web setup` fails to
install the managed backend, and once the backend is installed by hand, every
web command — including `agent-device web doctor` — fails with `spawn EINVAL`.

Both failures are Windows-only and neither is caused by a broken environment:
the backend runs perfectly when invoked directly.

## Environment

| | |
|---|---|
| OS | Windows 11 (10.0.26200) |
| Node | v24.13.1 |
| agent-device | 0.20.10 (latest at time of writing) |
| agent-browser | 0.27.1 (the pinned managed backend) |
| Shells tried | Git Bash, PowerShell 5.1 — same result in both |

## Bug 1 — `web setup` cannot find npm

```
$ agent-device web setup
Setting up managed agent-browser backend (downloads if needed)...
Error (TOOL_MISSING): npm not found in PATH
Hint: Install required platform tooling and ensure it is available in PATH.
```

npm is on PATH and resolvable:

```
PS> (Get-Command npm.cmd).Source
C:\Program Files\nodejs\npm.cmd
```

On Windows npm ships as `npm.cmd` / `npm.ps1` plus an extensionless `npm` shell
script that Windows cannot execute. `dist/src/agent-browser-tool.js` installs the
backend with:

```js
await i(`npm`, [`install`, `--prefix`, t, `--no-audit`, `--no-fund`, `--no-save`, `${L}@${I}`], …)
```

A bare `npm` cannot be spawned on Windows without either PATHEXT resolution or a
shell, so this never succeeds regardless of PATH.

## Bug 2 — `spawn EINVAL` on every web command

Installing the backend manually into the managed location (see workaround below)
gets past bug 1, and then:

```
$ agent-device web doctor
Error (UNKNOWN): spawn EINVAL
Hint: Unexpected internal error. Retry with --debug and report the diagnostics log if it persists.
```

**The backend itself is fine** — invoked directly it answers immediately:

```
$ ~/.agent-device/tools/agent-browser/0.27.1/package/node_modules/.bin/agent-browser.cmd --version
agent-browser 0.27.1
```

### Root cause

`dist/src/agent-browser-tool.js` resolves the backend binary to a `.cmd` on
Windows:

```js
function Z(e){
  let t = d.join(e, `package`);
  return d.join(t, `node_modules`, `.bin`,
    process.platform === `win32` ? `agent-browser.cmd` : `agent-browser`);
}
```

and it is spawned through the shared exec helper in
`dist/src/sdk-android-adb.js`, which passes `shell: false`:

```js
let s = m(r, t, { cwd: n.cwd, env: n.env, stdio:[`pipe`,`pipe`,`pipe`],
                  detached: n.detached, windowsHide: !0, shell: !1 });
```

Since the fix for **CVE-2024-27980** (Node 18.20.2 / 20.12.2 / 21.7.3 and later),
`child_process.spawn` on Windows **throws `EINVAL` for a `.bat` or `.cmd`
command unless `shell: true`**. So this combination cannot work on any currently
supported Node.

That also explains why the error surfaces as `UNKNOWN` rather than something
actionable — `EINVAL` from `spawn` is not being mapped to a diagnosis.

## Suggested fix

Two options, the second is cleaner:

1. **Spawn `.cmd`/`.bat` with `shell: true` on win32**, quoting the command path
   (it lives under the state dir, which can contain spaces via the user profile).
   This is what the workaround below does.

2. **Do not use the `.cmd` shim at all.** Resolve the backend to its JS entry
   point inside `package/node_modules/agent-browser` and spawn
   `process.execPath` with that script as the first argument. No shell, no
   cmd.exe quoting, and identical behaviour across platforms — which also
   removes the win32 branch from `Z()`.

Option 2 avoids reintroducing any of the argument-quoting hazards that the Node
CVE fix exists to prevent.

For bug 1, resolve npm the same way — spawn `npm.cmd` on win32, or invoke
`process.execPath` against npm's `npm-cli.js`.

## Workaround currently in use

Install the pinned backend manually into the managed location, exactly as
`web setup` would:

```bash
npm install --prefix "$HOME/.agent-device/tools/agent-browser/0.27.1/package" \
  --no-audit --no-fund --no-save agent-browser@0.27.1
```

then write the marker `web setup` writes, because `installed` is
`binary is X_OK && manifest.json exists`:

```jsonc
// ~/.agent-device/tools/agent-browser/0.27.1/manifest.json
{ "package": "agent-browser", "version": "0.27.1", "node": "v24.13.1", "installedAt": "…" }
```

and patch the exec helper so a `.cmd` is spawned through a shell — wrapping the
`spawn` import once rather than editing each call site:

```js
import { spawn as __adSpawnOrig, spawnSync as h } from "node:child_process";
function m(c, a, o) {
  if (process.platform === 'win32' && typeof c === 'string'
      && /\.(cmd|bat)$/i.test(c) && (!o || o.shell !== true)) {
    const q = /\s/.test(c) ? '"' + c + '"' : c;
    return __adSpawnOrig(q, a, Object.assign({}, o || {}, { shell: true }));
  }
  return __adSpawnOrig(c, a, o);
}
```

`adb.exe` and `emulator.exe` do not match the extension test, so Android
behaviour is unchanged.

### Confirmed working

With the manual install plus that patch, on the same machine that produced the
`EINVAL` above:

```
$ agent-device web doctor
Web backend is healthy.

$ agent-device open "http://localhost:5174" --platform web
Opened: http://localhost:5174

$ agent-device snapshot -i --platform web
Page: http://localhost:5174
Snapshot: 8 nodes
@e4 [text-field] "Email" [editable]
@e6 [button] "Sign in"

$ agent-device fill @e4 "someone@example.com" --platform web
Filled 26 chars

$ agent-device find text "Sign in" exists --platform web
Found: true
```

So the diagnosis holds: nothing is wrong with the backend or the environment —
only the way the `.cmd` is spawned.

Note that `shell: true` makes Node emit `DEP0190` ("args are not escaped, only
concatenated") on every web command, which is why option 2 above is the better
fix rather than simply flipping the flag.

### Minor, same area

After patching, the first command fails with
`Replacing daemon (pid …, v0.20.10): code-signature mismatch` followed by
`Daemon request timed out`. That is reasonable behaviour on its own, but the
orphaned daemons then hold a stale web-device lease which surfaces as:

```
Error (DEVICE_IN_USE): web device agent-browser-chrome is owned by session "cwd:…:default"
```

while `agent-device session list` reports `{"sessions": []}`. Killing the
managed browser process does not clear it; only killing the daemons does. A
`DEVICE_IN_USE` error naming a session that `session list` says does not exist
is hard to act on — it might be worth having the lease check reconcile against
live sessions, or having `close` accept the device serial when no session is
resolvable.

thymikee:
## Summary

On Windows no `--platform web` command worked, and `agent-device web setup` could not install the
managed backend at all. Both failures came from spawning Windows console shims with `shell: false`:

- `web setup` ran a bare `npm`. On Windows npm ships as `npm.cmd`/`npm.ps1` plus an extensionless
  shell script Windows cannot execute, so the spawn failed with `TOOL_MISSING: npm not found in
  PATH` no matter how npm was installed.
- Every web command — including `web doctor` — spawned
  `<install>/package/node_modules/.bin/agent-browser.cmd`. Since the CVE-2024-27980 fix
  (Node 18.20.2 / 20.12.2 / 21.7.3 and later) `child_process.spawn` refuses a `.bat`/`.cmd`
  command without `shell: true`, so this failed with `spawn EINVAL` on every supported Node.

**`runManagedAgentBrowser(args, options)` is now the only path that executes the backend.** Entry
resolution, the Node runtime, the managed environment, and the spawn all live behind it; setup,
doctor, and the provider go through it, so no caller is in a position to reintroduce the shim. The
entry comes from the installed package's declared `bin` (`bin/agent-browser.js` in 0.27.1) rather
than a hard-coded path. No shell, no `cmd.exe` quoting, no `DEP0190`.

**npm is untouched on macOS and Linux**, which were never broken: setup still spawns `npm` from
PATH. Only Windows resolves npm's own `npm-cli.js`, from exactly two places — an `npm_execpath` that
really is npm's launcher (pnpm and yarn advertise theirs through the same variable), else the copy
bundled beside `node` — and otherwise fails with the existing actionable `TOOL_MISSING`. Setup also
pins `--no-global`, so an ambient `npm_config_global` cannot redirect the install out of the managed
prefix, and an install that exits 0 without leaving a runnable entry is reported as that rather than
as the not-installed-yet hint telling you to run the command you are already running.

**The published status contract is unchanged.** `binaryPath` stays in `AgentBrowserToolStatus`,
`web setup --json`, `web doctor --json`, and `web setup`'s text output, still naming npm's console
shim; it is now informational. `entryScript` and `packageDir` are additive. The one behaviour change
is that a managed install counts as present only when the backend **package** is present, not when a
`node_modules/.bin` shim is — npm always writes both, so existing installs are unaffected.

One nuance: the backend now runs under the same Node runtime as agent-device rather than whichever
`node` the shim's shebang resolved from `PATH`. That is the deterministic choice (setup already
requires Node 24+), but it does pin the pair together.

Closes #2022

## Validation

`pnpm typecheck`, `pnpm lint`, `pnpm format:check` and `pnpm check:affected --run` are **fully
green** at `99baeb6`, after a fresh `pnpm install --frozen-lockfile && pnpm build` on the rebased
base (#2021 changed a workspace manifest).

Regression coverage, written as interface tests rather than layered unit tests:

- **npm resolution matrix**, driven through `setupManagedAgentBrowser`: POSIX spawns bare `npm`;
  Windows uses a validated `npm_execpath`; Windows falls back to the npm bundled beside `node`;
  Windows with neither fails `TOOL_MISSING` and spawns nothing.
- **`managed agent-browser runs its JS entry with the current Node runtime, not the shim`** — the
  fixture writes the `.bin` shim npm links, so this proves the shim is present and still not the
  spawned command, from a state dir whose path contains spaces.
- **`src/__tests__/cli-web.test.ts`** pins the published JSON of `web setup --json` and
  `web doctor --json` (`binaryPath`, `entryScript`, `packageDir`, `installDir`, `installed`, and
  `socketDir` still absent). Both go red if `binaryPath` leaves the status.
- Managed env, the install marker, and the unusable-install error each keep one test. Provider tests
  assert logical agent-browser arguments; the spawn shape is owned and asserted by the tool module.

Live evidence on Linux (Node 22 host, state dir path containing a space), running the shipped
resolution and spawn path against the real registry package:

```
npm install --prefix "…/state dir/tools/agent-browser/0.27.1/package" --no-global … agent-browser@0.27.1
npm exit=0 added 1 package in 1s
entryScript=…/state dir/…/node_modules/agent-browser/bin/agent-browser.js
backend spawn exit=0 stdout="agent-browser 0.27.1"
```

**Accepted residual risk:** there is no Windows host in this environment, so the Windows path is
contract-tested rather than exercised live. The claim rests on the mechanism — no `.cmd` is spawned
anywhere on the path — plus the reporter's own confirmation in #2022 that the backend answers when
it is not launched through its shim. Web smoke coverage
(`test/integration/smoke-web-platform.test.ts`) is CI/host-gated and did not run here.

## Scope

12 files, +659/−156; production is +207/−88 across four files. The change stays inside the web
command family.

An earlier revision of this PR factored npm and JS-entry resolution into a generic
`src/utils/node-package-cli.ts`. That module had exactly one production consumer, which is a
hypothetical seam rather than earned reuse, so it and its test are gone (−405 lines); the package
manifest `bin` lookup that is worth keeping is now private to the web module. Installation (npm
invocation and manifest write) sits in `agent-browser-install.ts` so `agent-browser-tool.ts` stays
inside the repo's 300-line target; it is web-private, takes plain parameters, and exports nothing
generic.

A second commit pinning the workflow help-card budget was dropped during the rebase: #2021 raised
that budget to `< 9100` on `main`, which resolves the base-branch Coverage failure this PR was
previously blocked behind. Nothing in this PR touches it now.

**Follow-up worth its own change:** `agent-device react-devtools` and `agent-device agent-cdp` still
run `runCmdStreaming('npm', ['exec', …])`, the same bare-`npm` spawn that fails on Windows. They are
a different command family with their own test surface, so they are not in this diff.

**Not in scope**, per triage on the issue: the stale web-device lease that survives an orphaned
daemon (`DEVICE_IN_USE` naming a session `session list` does not report) is filed separately as
#2031. Also left alone: `spawn`'s *synchronous* throws (which is why the reporter saw
`UNKNOWN: spawn EINVAL` rather than a classified error) bypass `runSpawnedCommand`'s error mapping.
No path in this diff can produce one any more, and repairing that boundary means auditing every
producer of a spawn error.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,18 @@
   `-dev` prerelease marker so the version on `main` never equals a published version (registry
   scanners diff the tool surface per version string, and a moving surface under a released number
   reads as a republish). `release:prepare` refuses to publish while the `-dev` marker is in place.
+- Windows `--platform web` works again. `agent-device web setup` no longer fails with
+  `npm not found in PATH`, and every web command — including `web doctor` — no longer fails with
+  `spawn EINVAL`. The managed `agent-browser` backend is now launched as `node <js-entry>` on every
+  platform instead of through its `node_modules/.bin` console shim, which is a `.cmd` file on
+  Windows that `child_process.spawn` refuses without a shell (CVE-2024-27980 hardening);
+  `shell: true` would only trade that for argument-quoting hazards and a `DEP0190` warning on every
+  command. Setup spawns `npm` from PATH unchanged on macOS and Linux, and runs npm's own
+  `npm-cli.js` under the current Node only on Windows, where a bare `npm` is not spawnable. A
+  managed install now counts as present only when the backend package itself is, and
+  `web setup --json` / `web doctor --json` gain `entryScript` and `packageDir`; the published
+  `binaryPath` is unchanged and still names npm's console shim, now informational rather than the
+  spawned command (#2022).
 - Parameterized `fill --record-as` protection is now recording-session-scoped instead of
   fill-step-scoped (ADR 0017 amendment): a later, unrelated recorded action (`wait`, `is`, `get`) can no
   longer re-serialize an app-rendered echo of an already-parameterized value into its own result or
--- a/docs/agents/web-backend.md
+++ b/docs/agents/web-backend.md
@@ -8,6 +8,12 @@ Web automation uses a managed `agent-browser` backend as an implementation detai
 - Use `agent-device web doctor` to run the backend health check.
 - The managed install respects `--state-dir` / `AGENT_DEVICE_STATE_DIR`.
 - Web automation requires Node 24+ while the rest of agent-device keeps its Node 22 baseline.
+- Every backend call spawns the package's declared `bin` entry with the current Node runtime,
+  never the `node_modules/.bin` console shim: Windows ships that shim as `.cmd`, which
+  `child_process.spawn` refuses without a shell (CVE-2024-27980 hardening), and a shell would
+  reintroduce argument-quoting hazards. `runManagedAgentBrowser` is the only path that executes
+  the backend. Setup spawns `npm` from PATH as before, except on Windows, where a bare `npm` is
+  not spawnable and npm's own `npm-cli.js` runs under the current Node instead.
 
 Default first-run flow:
 
--- a/src/__tests__/cli-web.test.ts
+++ b/src/__tests__/cli-web.test.ts
@@ -1,0 +1,108 @@
+import assert from 'node:assert/strict';
+import fs from 'node:fs';
+import { test } from 'vitest';
+import { runCliCapture } from './cli-capture.ts';
+import { mkdtempForTestSync } from './test-utils/tmp-dir.ts';
+import { withCommandExecutorOverride } from '../utils/exec.ts';
+import {
+  installFakeManagedAgentBrowser,
+  withNodeRuntime,
+  writeFakeManagedAgentBrowserPackage,
+} from '../platforms/web/__tests__/test-utils.ts';
+
+type SpawnedCommand = { cmd: string; args: string[] };
+
+// `binaryPath` has been in the published `web setup`/`web doctor` JSON since #833.
+// The Windows spawn fix (#2022) moves execution to `node <entryScript>` and adds
+// `entryScript`/`packageDir`, but the released field stays in the contract.
+test('web doctor --json keeps the published status fields and spawns the JS entry', async () => {
+  const stateDir = mkdtempForTestSync('agent device cli web doctor ');
+  try {
+    const install = installFakeManagedAgentBrowser(stateDir);
+    const spawned: SpawnedCommand[] = [];
+
+    const result = await withCommandExecutorOverride(
+      async (cmd, args) => {
+        spawned.push({ cmd, args });
+        return { stdout: 'ok', stderr: '', exitCode: 0 };
+      },
+      async () =>
+        await runCliCapture(['web', 'doctor', '--json'], {
+          env: { AGENT_DEVICE_STATE_DIR: stateDir },
+        }),
+    );
+
+    const status = parseStatus(result.stdout);
+    assert.equal(status.binaryPath, install.binaryPath);
+    assert.equal(status.entryScript, install.entryScript);
+    assert.equal(status.packageDir, install.packageDir);
+    assert.equal(status.installDir, install.installDir);
+    assert.equal(status.installed, true);
+    assert.equal(status.socketDir, undefined);
+    assert.deepEqual(spawned, [
+      { cmd: process.execPath, args: [install.entryScript, 'doctor', '--offline', '--quick'] },
+    ]);
+    assert.equal(result.calls.length, 0);
+  } finally {
+    fs.rmSync(stateDir, { recursive: true, force: true });
+  }
+});
+
+test('web setup --json keeps the published status fields after installing', async () => {
+  const stateDir = mkdtempForTestSync('agent device cli web setup ');
+  try {
+    let install: ReturnType<typeof writeFakeManagedAgentBrowserPackage> | undefined;
+    let stdout = '';
+
+    await withNodeRuntime({ version: '24.13.0' }, async () => {
+      const result = await withCommandExecutorOverride(
+        async (_cmd, args) => {
+          // Stand in for the npm run that writes the managed package tree.
+          if (args.includes('--prefix')) install = writeFakeManagedAgentBrowserPackage(stateDir);
+          return { stdout: '', stderr: '', exitCode: 0 };
+        },
+        async () =>
+          await runCliCapture(['web', 'setup', '--json'], {
+            env: { AGENT_DEVICE_STATE_DIR: stateDir },
+          }),
+      );
+      stdout = result.stdout;
+    });
+
+    const status = parseStatus(stdout);
+    assert.equal(status.binaryPath, install?.binaryPath);
+    assert.equal(status.entryScript, install?.entryScript);
+    assert.equal(status.packageDir, install?.packageDir);
+    assert.equal(status.installed, true);
+    assert.equal(status.socketDir, undefined);
+  } finally {
+    fs.rmSync(stateDir, { recursive: true, force: true });
+  }
+});
+
+function parseStatus(stdout: string): Record<string, unknown> {
+  const payload: unknown = JSON.parse(firstJsonDocument(stdout));
+  assert.ok(isRecord(payload) && payload.success === true, stdout);
+  const data = payload.data;
+  assert.ok(isRecord(data), stdout);
+  const status = data.status;
+  assert.ok(isRecord(status), stdout);
+  return status;
+}
+
+function isRecord(value: unknown): value is Record<string, unknown> {
+  return typeof value === 'object' && value !== null;
+}
+
+/**
+ * The CLI dispatches `web` inside its top-level try, and the capture harness
+ * turns `process.exit` into a throw, so the command's payload is followed by the
+ * CLI's own report of that synthetic exit. Only the first document is the
+ * command's own output; real runs exit for real and print it once.
+ */
+function firstJsonDocument(stdout: string): string {
+  const lines = stdout.split('\n');
+  const end = lines.indexOf('}');
+  assert.ok(end >= 0, stdout);
+  return lines.slice(0, end + 1).join('\n');
+}
--- a/src/daemon/handlers/__tests__/session-teardown-resources.test.ts
+++ b/src/daemon/handlers/__tests__/session-teardown-resources.test.ts
@@ -351,8 +351,10 @@ test('daemon session teardown closes an open web session immediately, not on age
   // fleet right away, the same way an explicit `session close` does, instead of leaving the
   // Chrome processes to agent-browser's own multi-minute idle timer.
   expect(mockRunCmd).toHaveBeenCalledTimes(1);
-  const [, args] = mockRunCmd.mock.calls[0] as [string, string[]];
-  expect(args).toEqual(['close', '--json', '--session', sessionName]);
+  const [cmd, args] = mockRunCmd.mock.calls[0] as [string, string[]];
+  // `node <entry> close ...`: the managed backend never runs through its `.bin` shim (#2022).
+  expect(cmd).toBe(process.execPath);
+  expect(args.slice(1)).toEqual(['close', '--json', '--session', sessionName]);
 });
 
 test('daemon session teardown surfaces a web close failure through the cleanup-failure channel', async () => {
--- a/src/daemon/server/daemon-runtime-web-close-teardown.test.ts
+++ b/src/daemon/server/daemon-runtime-web-close-teardown.test.ts
@@ -129,7 +129,14 @@ test('daemon shutdown closes an open web session immediately, without waiting fo
   expect(stderrChunks.join('')).toBe('');
   expect(sessionStore.get(session.name)).toBeUndefined();
   const closeCall = mockRunCmd.mock.calls.find(([, args]) => (args as string[]).includes('close'));
-  expect(closeCall?.[1]).toEqual(['close', '--json', '--session', session.name]);
+  // `node <entry> close ...`: the managed backend never runs through its `.bin` shim (#2022).
+  expect(closeCall?.[0]).toBe(process.execPath);
+  expect((closeCall?.[1] as string[] | undefined)?.slice(1)).toEqual([
+    'close',
+    '--json',
+    '--session',
+    session.name,
+  ]);
 });
 
 test('daemon shutdown reports a web close failure on stderr instead of losing it silently', async () => {
--- a/src/platforms/web/__tests__/test-utils.ts
+++ b/src/platforms/web/__tests__/test-utils.ts
@@ -8,19 +8,73 @@ const TEST_AGENT_BROWSER_VERSION = '0.27.1';
 type FakeManagedAgentBrowserInstall = ReturnType<typeof expectedManagedAgentBrowserInstall>;
 
 export function installFakeManagedAgentBrowser(stateDir: string): FakeManagedAgentBrowserInstall {
-  const install = expectedManagedAgentBrowserInstall(stateDir);
-  fs.mkdirSync(path.dirname(install.binaryPath), { recursive: true });
-  fs.writeFileSync(install.binaryPath, '#!/bin/sh\nexit 0\n');
-  fs.chmodSync(install.binaryPath, 0o755);
+  const install = writeFakeManagedAgentBrowserPackage(stateDir);
   fs.writeFileSync(path.join(install.installDir, 'manifest.json'), '{}');
   return install;
 }
 
+/** The package tree `npm install` leaves behind, without the setup manifest. */
+export function writeFakeManagedAgentBrowserPackage(
+  stateDir: string,
+): FakeManagedAgentBrowserInstall {
+  const install = expectedManagedAgentBrowserInstall(stateDir);
+  fs.mkdirSync(path.dirname(install.entryScript), { recursive: true });
+  fs.writeFileSync(install.entryScript, 'process.exit(0)\n');
+  // npm links a console shim beside the package; the fixture carries it so tests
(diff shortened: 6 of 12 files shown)


thymikee:
Merged `main` (`608bf7a`) into the branch — the only conflict was the `## Unreleased` changelog block, resolved by keeping both entries. `main` touched none of the files this PR changes.

One pre-existing failure to flag, not from this diff: `src/__tests__/cli-help.test.ts` → *"help workflow prints the compact workflow card…"* fails at **9003 bytes against its `< 9000` budget**. I reproduced it on a detached checkout of `origin/main` with no changes applied, so it arrived with #2023's help-topic additions and is red on the base branch too. This PR adds no CLI help text, so I have not touched it — it needs ~4 bytes trimmed from the workflow card by whoever owns that budget.

Everything else is green on the merged head: `pnpm typecheck`, `pnpm lint`, `pnpm format:check`, and `pnpm check:affected --run` (297 of 298 test files, the one failure above). The web/daemon suites this PR touches are 70/70.

---

thymikee:
**P1 — preserve the released web status contract.** `AgentBrowserToolStatus`, `web setup --json`, and `web doctor --json` have exposed `binaryPath` since #833 and through v0.20.10. This head removes it and substitutes `entryScript`, but the Windows spawn fix does not require a breaking JSON/API rename. Keep `binaryPath` in the published status shape (it may remain informational while execution uses `process.execPath + entryScript`) and add CLI-level JSON regressions for setup and doctor; `entryScript` can be additive. If a breaking removal is intended, it needs an explicit breaking-release decision rather than landing inside this fix.

The JS-entry spawn design and focused regressions otherwise look sound. Current Coverage (2) is a confirmed base-branch help-card budget failure, iOS smoke is pending, and merge-readiness still needs exact-head Windows evidence for `web setup`, `web doctor`, and one real web command with a spaced state dir.

thymikee:
P1 addressed in `93bc27a` — no breaking rename any more.

`binaryPath` is back in `AgentBrowserToolStatus`, `web setup --json`, and `web doctor --json`, naming npm's console shim exactly as it did before; `web setup`'s text output prints `agent-browser available at: <binaryPath>` again. `entryScript` and `packageDir` are purely additive. Execution is unchanged — `process.execPath` plus the JS entry — so the shim is now informational and never spawned.

CLI-level JSON regressions in the new `src/__tests__/cli-web.test.ts`:

- `web doctor --json keeps the published status fields and spawns the JS entry` — pins `binaryPath`, `entryScript`, `packageDir`, `installDir`, `installed`, asserts `socketDir` still does not leak, and asserts the spawn was `node <entryScript> doctor --offline --quick`.
- `web setup --json keeps the published status fields after installing` — same field pinning after a stubbed install.

Both go red when `binaryPath` leaves the status; I checked by removing it and re-running.

The web fixture now also writes the `.bin` shim npm links, so the spawn regressions prove the shim is *present and still not used*, rather than merely absent.

On the two open items you named:

- **Coverage (2)** is the base-branch help-card budget (9003 vs `< 9000`), reproduced on a detached `origin/main` for both `cli-help.test.ts` and `cli-schema/cli-help-topics.test.ts`. Untouched here.
- **iOS Smoke** failed on `cb25b00` at `fill id="field-name"` with `TEXT_INPUT_COMMIT_NOT_OBSERVED` on the live simulator — a lane this diff does not touch. This push re-runs it; if it reproduces I will report rather than widen the PR into iOS text input.
- **Windows evidence** I cannot produce: no Windows host in this environment. Merge-readiness for that claim needs someone on Windows to run `web setup`, `web doctor`, and one real web command against a spaced state dir. Everything else is stated in the PR body as residual risk rather than claimed as verified.

---

thymikee:
Re-reviewed exact head `93bc27a307208f57c0559543309dde935e9b41c1`. The prior P1 is resolved: released `binaryPath` is restored across the status type, `web setup --json`, `web doctor --json`, and text output; `entryScript`/`packageDir` are additive, while execution remains `process.execPath <JS entry>`. Regression tests pin the public JSON and prove the `.bin` shim exists but is not spawned. No remaining code finding. All four smoke lanes pass; Coverage (2) is the known base workflow-help 9003/<9000 failure. Readiness still needs exact Windows evidence for `web setup`, `web doctor`, and one real web command with a state directory containing spaces. No `ready-for-human` label until that evidence is posted and the confirmed failing CI is resolved.

thymikee:
Thanks for the re-review. Force-pushed a tidied branch — history only, no change to the reviewed diff, plus the budget fix you approved.

**`6a403d9`** — the whole web fix as one commit, rebased onto current `main` (`52ac5da`). The merge commit and the follow-up `binaryPath` commit are gone; the tree is identical to what you reviewed at `93bc27a`, carried forward onto the newer base. The only conflict was the `## Unreleased` changelog block, resolved as a pure insertion after `main`'s entries.

**`b548587`** — the workflow help-card budget, raised to 9003 as you allowed. Both assertions (`cli-help.test.ts` and `cli-schema/cli-help-topics.test.ts`) now compare against a named `WORKFLOW_CARD_BUDGET_BYTES = 9003` and use `<=`, so it is pinned at exactly today's size rather than given headroom — the next addition to the card fails there and is a deliberate decision, the same shape as the test-file size ratchet. The comment records that #2023 pushed it past the old 9000 target and that raising it was approved.

`pnpm check:affected --run` is now **fully green** on `b548587` — no remaining failures, base-branch or otherwise. `pnpm typecheck`, `pnpm lint` and `pnpm format:check` are clean, after a fresh `pnpm install --frozen-lockfile && pnpm build` on the new base.

Still outstanding and still not something I can produce: the Windows evidence. There is no Windows host in this environment, so `web setup`, `web doctor`, and a real web command against a spaced state dir need to come from someone on Windows before `ready-for-human`. Everything else you listed is resolved.

---

thymikee:
Re-reviewed exact head `b548587c645a459317c6e0a878f61f78337b7cb3`. The web tree is unchanged from reviewed `93bc27a` apart from rebase/history cleanup; released `binaryPath` remains preserved and there is no new web-code finding. The additional help-budget commit repairs the base-owned 9003-byte failure by pinning exactly 9003 with no growth headroom, so the next byte still fails. I cannot independently confirm the comment/commit’s claim that this specific increase was maintainer-approved from the PR record, but the resulting gate remains conservative. Windows evidence is still absent: `web setup`, `web doctor`, and one real web command with a spaced state directory on Windows remain required. Fresh CI is still in progress, so no `ready-for-human` label yet.

thymikee:
On the approval provenance — you are right that it is not verifiable from the PR record, so here it is explicitly.

The approval came out-of-band, not on GitHub: @thymikee, driving this change in the Claude Code session that produced the PR, said *"you can increase the budget to 9003, I allow"* after I reported the 9003/`< 9000` failure as a base-branch blocker I would not touch on my own. That is the whole basis for `b548587`. I should have recorded where it came from rather than writing "approved by the maintainer" with no pointer — treat this comment as that pointer, and a one-word confirmation here from @thymikee puts it on the record properly.

Two ways to settle it, whichever you prefer:

1. **Keep it here.** The gate stays pinned at exactly 9003 with `<=`, so the next byte fails.
2. **Drop it from this PR.** The budget is `main`'s concern, not this fix's — I am happy to `git revert` that commit and send it as its own small PR against `main`, which is arguably where a base-owned gate change belongs. That returns this PR's Coverage lane to red until the separate PR lands.

Say which and I will do it. Nothing about the web fix changes either way, and the Windows evidence gap is unaffected and still open.

---

thymikee:
Coordinator update for exact head `b548587c645a459317c6e0a878f61f78337b7cb3`: the repository has no Windows host, so the requested live Windows run is accepted as an explicit residual risk rather than a readiness blocker. The Windows-safe design is structurally covered: both npm and agent-browser execute as `process.execPath <JS entry>` with no shell or `.cmd` spawn, spaced-path regressions pass, and the released JSON contract is preserved. Code review is clean. All completed checks are green; iOS smoke is still in progress. Marking `ready-for-human` under the project rule that clean code may be labeled while CI is still running.

thymikee:
Simplified per review — the generic package-launch subsystem is gone. Force-pushed to `e098e34`; net **−393 lines** against the previously reviewed `b548587`.

**`src/utils/node-package-cli.ts` (172 lines) and its 233-line test are deleted.** One production consumer never earned a shared seam.

**`runManagedAgentBrowser(args, options)` is now the sole execution owner.** Entry resolution, `process.execPath`, the managed environment, and the spawn all live behind it; setup, doctor, and the provider all go through it. The exported `resolveAgentBrowserTool()` command/args/env struct and the `AgentBrowserTool` type are gone, so no caller is in a position to reintroduce the shim.

**npm is untouched on POSIX.** `{ command: 'npm', args }` exactly as before — no prefix scanning, no PATH traversal, no Debian-symlink resolution, no `npm_execpath` package-manager discrimination. Only Windows resolves a JS entry, from just two places: a validated `npm_execpath` ending in `npm-cli.js`, or `node_modules/npm/bin/npm-cli.js` beside `process.execPath`; otherwise the existing actionable `TOOL_MISSING`.

**The package-manifest `bin` lookup stays**, now private to the web module — that is the contract worth reusing, and hard-coding `dist/cli.js` would be brittle.

Tests are interface-level now: a four-case npm-resolution matrix (POSIX bare `npm`, Windows advertised, Windows bundled, Windows missing) driven through `setupManagedAgentBrowser`; one "shim exists but `node <entry>` is spawned" test on a spaced path; the setup/doctor JSON compatibility tests in `cli-web.test.ts`. Provider tests assert logical agent-browser arguments — the executor strips the node prefix without re-asserting the command shape, which the tool module owns.

One structural note: rather than let `agent-browser-tool.ts` grow to 366 lines (past the repo's 300-line target), I split installation out to `agent-browser-install.ts` (93 lines) — npm invocation and manifest write. It is web-module-private, takes plain parameters, and exports nothing generic. Tool file is 309 lines.

Preserved properties, all still covered: no shell, no `.cmd` execution, spaced paths work, `binaryPath` remains compatible. `pnpm check:affected --run` is fully green on `e098e34`, as are typecheck, lint and format.

---

thymikee:
Re-checked exact head `e098e34f58e85cb095046dbe6fdea42ee6eb64fe` after #2021 merged. The code review and all required checks remain clean, but GitHub now reports the branch `DIRTY`/conflicting against current `main`. This is no longer merge-ready: rebase onto `main`, resolve the conflict without broadening the change, rerun the affected/full required checks, and request re-review on the new head. The accepted residual remains that the Windows-specific path is contract-tested rather than exercised on a live Windows host.

thymikee:
Rebased onto `c77bc40` and pushed as `99baeb6`. GitHub now reports the branch **clean**, not conflicting. Ready for re-review on this head.

**How the conflict resolved.** The web fix commit replayed cleanly with no conflict. The only conflict was in the second commit, the help-card budget — and it conflicted because #2021 had already raised that budget on `main` to `< 9100`. So I dropped my commit rather than override `main`'s choice with my stricter pin. That is the whole resolution: **the branch is now a single commit, and nothing in this PR touches the help-card budget any more.** The base-branch Coverage failure this PR was previously blocked behind is resolved by `main` itself.

Nothing was broadened: the diff against the new base is the same web change, 12 files, +659/−156, production +207/−88.

**Re-validated on the new base**, after a fresh `pnpm install --frozen-lockfile && pnpm build` (#2021 changed a workspace manifest, so the affected selector fails open to the wider gate set): `pnpm check:affected --run` is fully green, as are typecheck, lint and format. CI is re-running on `99baeb6`.

The PR body is updated to match this head — it still described the generic `node-package-cli.ts` module from before the simplification.

Residual risk is unchanged and as you accepted it: the Windows path is contract-tested, not exercised on a live Windows host.

---

PR #2035, opened 2026-08-25
fix(ios): grant the text-entry commit wait time against progress

thymikee:
## Summary

`type` and `fill` on iOS stop reporting `TEXT_INPUT_COMMIT_NOT_OBSERVED` over text that is still
landing. The synthesized commit wait used a flat 3s deadline, which cannot tell a *throttled*
simulator input pipeline (characters keep arriving, slowly) from a *wedged* one (nothing arrives) —
so it condemned both at the same instant. That is what turned a simulator-load episode into a red
`Smoke Tests (iOS simulator)` / targeted-XCTest lane on branches touching no iOS code.

- **The commit wait now grants time against progress** (`SynthesizedCommitBudget`). The observed
  value's expected-prefix length — the same length-only evidence `logCommitCadence` already emits —
  decides: while it grows, the wait continues, up to a 10s ceiling. A pipeline making no progress
  expires at exactly the 3s the flat deadline used, so nothing that fails today starts passing
  merely by waiting longer, and a wedge is condemned no later than before.
- **Text-entry readiness stops guessing from a wall clock.** `waitForTextEntryReadiness`'s
  hardware-keyboard fallback returned a possibly-unfocused element after 0.35s of "no software
  keyboard seen". It now returns only when the target element itself confirms keyboard focus, and
  re-arms that window otherwise; unconfirmed, the wait runs its normal readiness timeout — what it
  would have spent with no fallback at all.
- **The keyboard-hidden precondition of `testBareTypeUsesTappedInputWhenSoftwareKeyboardIsHidden`
  is now `XCTSkipIf`, not `XCTAssertFalse`.** An environment flip must not read as a product
  regression on an unrelated PR.

Refs #1874 — deliberately not a closing keyword. This is a mitigation: it removes the two
mechanisms that were actually firing in CI, but the underlying question that issue tracks —
why the simulator input pipeline throttles at all — is untouched here, and the 40-second
episodes it records would still exhaust the ceiling. #1874 stays the open `needs-info`
tracker for that.

### Two corrections to the issue's diagnosis

**Neither observed failure reaches `waitForTextEntryReadiness:407`.** Tracing both paths:

- `testBareTypeUsesTappedInputWhenSoftwareKeyboardIsHidden` types through the tap witness, and
  `rememberedTextEntryTarget()` returns before readiness is ever consulted.
- `smoke:form-input` → `fill full name` never reaches readiness either: it fails inside
  `runSynthesizedReplacementRoute`, which `shouldUseSynthesizedFirstResponderReplacement` gates on
  `hasResolvedElement == false` — the penalized coordinate route builds that target directly in
  `executeTypeCommand`, bypassing `focusTextInputForTextEntry` entirely.

**And the smoke half is not the deadline either — it is the fixture.** This PR's own iOS lane
reproduced it ([run 32889322172](https://github.com/callstack/agent-device/actions/runs/32889322172)),
and the trace settles it:

```
[DEBUG-1874] wait start expectedLen=12 route=replacement     19:35:00.652
   (no poll lines at all)
[DEBUG-1874] wait outcome=notObserved elapsedMs=3608 route=replacement   19:35:04.260
```

The wait never polled. It returned from the `textMatchesPlaceholder` guard, which refuses *before*
polling because an empty text field renders its placeholder as its accessibility value, so a match
cannot prove a commit. `expectedLen=12` is `"Ada Lovelace"` — and `field-name`'s placeholder was
`"Ada Lovelace"`. `field-email` had the same collision with `ada@example.com`: twelve fills across
eight files, every checkout-form flow in the repository. The 3.6s is just the AX resolution for the
placeholder read that precedes the guard.

It looked intermittent because the synthesized-replacement route is gated on
`xCTestChannelPenalized` — it fires when the host is loaded. That also explains the datapoint the
issue found most puzzling: `rerun_failed_jobs` on the same commit reproducing *identically*. A
deterministic guard on a load-gated route looks exactly like a sticky per-device flake.

The collision additionally made the read-back assertions vacuous — `assertJsonContains(name, 'Ada
Lovelace')` is satisfied by an empty field rendering the placeholder — so fixing it strengthens the
smoke rather than just quieting it. Fixed in the fixture, not in the values, because frozen
replay-compat corpora carry the same fills. No static guard: a regex over source cannot enumerate
"every fill", so it would sit there green while its claim was false — the Swift tests prove a
placeholder-equal AX value is unobservable, and the live smoke lane is what caught this one.

So the three changes address three different things: the budget is what the targeted XCTest lane
needs, the fixture is what the smoke lane needs, and the readiness fix closes the latent
sensitivity the issue describes without being what fired in either.

**Pinning the simulator keyboard preference would have been a no-op, so this PR does not do it.**
Measured on a dedicated simulator, headless `simctl boot`, iOS 26.2, then deleted:

| per-device `ConnectHardwareKeyboard` | Settings search field, after tap |
| --- | --- |
| `true` (hardware keyboard "connected") | full software keyboard |
| `false` | full software keyboard |

`com.apple.iphonesimulator` is Simulator.app's preference domain; with no Simulator.app running,
CoreSimulator does not consult it. Writing it would have added a step, a guard test and a green
assertion that proved only that `defaults write` ran. The useful half of that finding is the
reassuring one: a headless lane has no hardware keyboard to inherit, so the software keyboard
always shows and the ambient state the issue names is not a variable in these lanes at all — which
independently agrees with the 2026-08-21 diagnosis comment.

## Validation

Local, iPhone 17 Pro / iOS 26.2 simulator, arm64 slice. The Swift tree is byte-identical to the
validated one — `main` moved under this branch mid-work (#2021) and the rebase touched nothing
under `apple/runner/`, and the static gates were re-run on `ba7e03f885` after it:

- The full `ios.yml` PR `-only-testing:` list: **75 passed, 0 failures**.
- Red-first for the budget, through the shipped waits: reverting `SynthesizedCommitBudget.record`
  to a no-op turns `testCommitWaitOutlivesTheFlatDeadlineWhileTheExpectedPrefixGrows` and
  `testCommitWaitCeilingStopsAnIndefinitelyThrottledPipeline` red, while
  `testCommitWaitCondemnsAFrozenPipelineAtTheStallBudget` and
  `testCommitWaitTreatsARetreatingValueAsNoProgress` stay green — they encode the behavior that
  must not move.
- `testHardwareKeyboardResponderConfirmsItsOwnKeyboardFocus` was written red first and caught a
  wrong implementation: `XCUIElement.descendants(matching:)` does **not** include the receiver, so
  an element-rooted focus query answers `0` for a leaf text field. Instrumented on-device
  (`selfDescendants=0`, `appFocused=1` naming the tapped field at `keyboardVisible=0`) and rebuilt
  against the app-wide predicate with an identity check. That XCTest behavior is now pinned by the
  test rather than assumed.
- The keyboard A/B above; probe device deleted and `com.apple.iphonesimulator` restored from a
  pre-probe export afterwards, agent-device sessions closed.
- The placeholder collision was found by reading this PR's own failed smoke lane; the fix is
  covered by that lane going green on the next head, not by a static guard.

Not verified locally: the macOS host XCTest lane, blocked on this host by the known
Accessibility/Automation permission gap. The budget tests are platform-neutral and were executed on
the simulator instead, where the same `#if AGENT_DEVICE_RUNNER_UNIT_TESTS` bundle runs them. CI on
the head is the authority for the host and smoke lanes.

Unrelated pre-existing red on `main` at the same commit, not introduced here:
`cli-help-topics.test.ts` / `cli-help.test.ts` fail because the `workflow` help card is 9003 bytes
against a 9000-byte cap.

## Design notes and residual risk

- **The budget lives inside the wait.** `awaitSynthesizedCommitOutcome` and its replacement
  counterpart hold it as a local `var` and advance it from the same observation the progress check
  already reads, with the clock injected alongside the existing observation and pacing seams. An
  earlier revision made it a detached object shared between two escaping closures and guarded that
  sharing with a TypeScript parser; the guard was evidence the seam was wrong, so both are gone.
  What replaced them is four sequence tests that drive the real waits.
- **A failing `fill` can now take up to 10s instead of 3s** when the value keeps converging without
  ever matching (a formatter rewriting input, say). Bounded well inside the daemon's 45s
  per-command budget. A dropped-character corruption still fails at ~3s, because its prefix freezes.
- **`fill` still cannot verify a value equal to the field's placeholder**, and that asymmetry is
  route-dependent: the ordinary `xctest-element` route reads the raw value back and passes, while
  the penalized route refuses. Both are right on the evidence each has, but the same command can
  therefore report differently under load. Worth its own issue against #1924's three-state
  `PlaceholderCommitEvidence` model; not expanded into here.
- **This reduces exposure rather than eliminating it.** The 40-second episode traced in the
  2026-08-21 comment would still exhaust the ceiling and report honestly. Why the simulator input
  pipeline throttles at all is unchanged by this PR; the `[DEBUG-1874]` cadence traces stay in place
  for it.

**Size**: the growth is Apple runner source, because the whole `AgentDeviceRunnerUITests` bundle
ships as source; JS shrinks. `SynthesizedCommitBudget` keeps its own file only to hold
`RunnerTests+SynthesizedTextEntry.swift` inside the 500-line extraction rule — folding it back
would save nothing packed. The review rework removed the detached test file and the TypeScript
guard, so the final figure is below what the last size report measured.

7 files: the iOS text-entry path and the form fixture's two placeholders. No docs change:
`TEXT_INPUT_COMMIT_NOT_OBSERVED`'s meaning and recovery advice in `website/docs/docs/commands.md`
are unchanged — only how long the runner looks before reporting it.


thymikee:
The +6.5 kB npm unpacked is the two new Swift files (the whole UITests bundle ships as source); JS shrinks by 216 B. Itemized in the PR body, with why a smaller design was rejected.

thymikee:
Re-reviewed `ba7e03f8854503666fd08931365a4ba2f5b4f7b8`. The progress-aware behavior, focus confirmation, live iOS evidence, and all green checks are sound. One P2 maintainability issue remains, especially given the +6.5 kB unpacked growth: `SynthesizedCommitBudgetTests` exercise a detached class, while a new 61-line TypeScript parser proves that two escaping Swift closures happen to share it. The paragraph-long explanation that this tuple must retain a textual shape for another source parser is evidence the seam is wrong, and a reference type does not prevent the closures from being wired to different instances. Move the progress budget into the executable Swift polling path (`awaitSynthesizedCommitOutcome` and replacement counterpart) with the existing injected observation/pacing seams plus an injected clock. Make red-first Swift sequence tests exercise progress past 3s, no-progress expiry at 3s, and the 10s ceiling through the real wait; then delete the TypeScript wiring guard and detached scaffolding. Re-review after the head changes.

thymikee:
Addressed in `e5bdccbfbb`.

The budget is now a local `var` inside `awaitSynthesizedCommitOutcome` and its replacement counterpart, advanced from the same observation the progress check already reads, with the clock injected alongside the existing observation and pacing seams. Recording progress and asking whether time is up are two statements in one loop, so there is no cross-closure coupling left — the reference type went back to a value type, and the TypeScript wiring guard and the detached tests are deleted.

In their place, four sequence tests drive the shipped waits through a hand-driven clock: a growing prefix outlives the flat 3s deadline (settles at t=8s), a frozen prefix is condemned at exactly 3s, an indefinitely throttled pipeline stops at the 10s ceiling, and a value churning between two lengths buys no time. Red-first verified: the two progress tests fail against a no-op `record`; the two unchanged-behavior tests stay green, which is what they are for. The ten existing commit-wait tests moved to the same clock rather than an `isExpired` predicate.

Re-validated on iPhone 17 Pro / iOS 26.2: the full `ios.yml` PR list is 75 passed, 0 failures. Net effect on size is a reduction against the last report — one Swift file and one TypeScript file deleted.

thymikee:
Re-review at `e5bdccbfbbd287c85ee42d4c45fe60bd6fc083ec`: the P2 is resolved. Budget state is now local to each executable commit-wait loop; the detached Swift policy tests and TypeScript source parser are gone; clock-driven tests exercise the shipped append/replacement waits and prove progress past 3s, frozen-prefix expiry at 3s, the 10s ceiling, and churn buying no time. The exact-head iOS smoke and all required checks are green. No remaining code findings; ready for human review.

thymikee:
Max-effort review of the branch found one real bug, now fixed in `9485e8052e`.

**The budget clock started before the wait, not at it.** `waitStartedAt` was stamped before `synthesizedCommitPollingIngredients`, whose `placeholderValue` read is an AX round-trip — and on the loaded host this whole change exists for, AX reads were measured in seconds. With a 3.5s placeholder read the first observation already exceeded the 3s stall budget, so `type` returned `TEXT_INPUT_COMMIT_NOT_OBSERVED` after a single poll. The pre-PR code stamped its deadline *inside* the ingredients function, i.e. after that read, so it gave the field a full 3s. The change was making `type`/`fill` give up sooner than the code it replaced, in exactly the condition it targets.

The budget is now two durations, and only the poll loop starts it, from its own first `now()` — a pre-loop timestamp is no longer expressible at the call site. `testCommitWaitBudgetStartsAtTheLoopRatherThanBeforeIt` pins it (60s of setup must still leave the whole stall budget), verified red against a deadline started outside the loop.

Two smaller ones fixed alongside: the poll took two clock samples where one was meant, and `record`'s doc still described a `-1` unreadable sentinel that no caller passes any more.

Re-validated: policy set 28 passed, full `ios.yml` PR list 75 passed, 0 failures.

thymikee:
The max-effort review turned up something bigger than the three findings I posted, and it came from this PR's own failing iOS lane.

**The `smoke:form-input` half of #1874 is not the commit deadline — it is the fixture.** [Run 32889322172](https://github.com/callstack/agent-device/actions/runs/32889322172) failed with the exact issue signature, and the trace settles it:

```
[DEBUG-1874] wait start expectedLen=12 route=replacement     19:35:00.652
   (no poll lines at all)
[DEBUG-1874] wait outcome=notObserved elapsedMs=3608         19:35:04.260
```

Zero poll lines. The wait never entered its loop — it returned from the `textMatchesPlaceholder` guard, which refuses *before* polling because an empty text field renders its placeholder as its accessibility value, so a match cannot prove a commit. `expectedLen=12` is `"Ada Lovelace"`, and `field-name`'s placeholder **was** `"Ada Lovelace"`. `field-email` had the same collision with `ada@example.com`. Twelve fills across eight files — every checkout-form flow in the repo asks `fill` for the one thing it documents it cannot verify. The 3.6s is just the AX resolution for the placeholder read ahead of the guard.

It looked intermittent because the synthesized-replacement route is gated on `xCTestChannelPenalized`, i.e. it fires when the host is loaded. That also explains the datapoint the issue found most puzzling — `rerun_failed_jobs` on the same commit reproducing *identically*. A deterministic guard on a load-gated route looks exactly like a sticky per-device flake.

It also made the read-back assertions vacuous: `assertJsonContains(name, "Ada Lovelace")` is satisfied by an empty field rendering the placeholder. So this strengthens the smoke rather than quieting it.

Fixed in `a36b9ec91f` by changing the two fixture placeholders — not the filled values, because frozen replay-compat corpora carry the same fills. `fixture-fill-placeholder-collision.test.ts` guards the class and was planted-red: before the change it lists all twelve.

Residual, flagged not fixed: `fill` still cannot verify a value equal to its field's placeholder, and the verdict is route-dependent — the ordinary `xctest-element` route reads the raw value back and passes where the penalized route refuses. Both are right on the evidence each has, but the same command can report differently under load. That belongs against #1924's three-state `PlaceholderCommitEvidence` model, in its own issue.

thymikee:
[P2] Remove `src/__tests__/fixture-fill-placeholder-collision.test.ts`; keep the two placeholder changes and the existing Swift/live regression coverage. The 83-line test is a source-reconstruction guard, not a durable fixture invariant: it regex-parses JSX and only two literal fill syntaxes, duplicates Swift trim/equality semantics in TypeScript, and already misses the multiline real `fill` in `test/integration/android-emulator-e2e/live-form-scenario.ts` (plus expressions, spreads, `.tsx`, variables, typed clients, and unlisted roots). It can therefore stay green while its “Every fill” claim is false. The owning evidence already exists: Swift tests prove placeholder-equal AX values are unobservable, and live `smoke:form-input` failed on the prior head for exactly this fixture collision. If a global static contract is ever needed, expose fixture metadata and use the real `.ad` parser rather than regex-parsing source.

Current iOS smoke is still running; this is a code-quality blocker independent of that result.

thymikee:
Agreed and done in `adcc35b8db` — guard deleted, the two placeholder changes kept.

You are right that a regex over source cannot enumerate "every fill", so it would sit there green while its claim was false; expressions, variables, typed clients and unlisted roots are all outside it, and duplicating the Swift trim/equality rule in TypeScript is a second source of truth for a rule that already has one. One correction for the record, since it does not change the conclusion: the multiline `fill` in `android-emulator-e2e/live-form-scenario.ts` *was* matched — `\s*` spans newlines and `test/integration` is in the scanned roots — so the guard was wrong in principle rather than demonstrably incomplete at that call site.

I took the same pass over the rest of the change while I was in there. The commit deadline was a budget value type, a nested `Deadline` type and a factory method; it is now one flat struct the poll loop constructs, with the two durations as defaulted parameters. Production call sites name no budget at all, tests name one only when they are asking about time, and both `SynthesizedCommitBudget.standard` and the tests' `unboundedCommitBudget` disappear.

Re-validated on iPhone 17 Pro / iOS 26.2: policy set 28 passed, full `ios.yml` PR list 75 passed, 0 failures. (One local run reported `Executed 0 tests` — the known silent-green contention flake, not a result; re-run gave 75.) Static gates green.

thymikee:
P1: change `Closes #1874` to a non-closing reference. This PR is a valid mitigation, but #1874 remains the canonical open `needs-info` tracker for the unidentified simulator input-throttle mechanism; the PR itself says it reduces exposure rather than eliminating it. As written, merge would auto-close an unresolved tracker. No code changes requested; once the closing keyword is removed, the code/evidence review is clean and this can be marked ready-for-human.

thymikee:
Done — `Closes #1874` is gone from both surfaces.

The PR body now reads `Refs #1874`, with a sentence on why: this removes the two mechanisms that were actually firing in CI, but the input-throttle question that issue tracks is untouched here, so it stays the open `needs-info` tracker.

The body alone would not have been enough. This repo allows squash-merge only, and `squash_merge_commit_message` is `COMMIT_MESSAGES` — so the squash body is built from the individual commit messages, not the PR body, and `ba7e03f885` still carried `Closes #1874`. I rewrote that message and rebased the branch onto it (force-push, head now `a1cd050d52`). Messages only: the tree is byte-identical to `adcc35b8db`, `git diff adcc35b8db HEAD` is empty, and no commit in the range carries a closing keyword any more.

Since only commit messages changed, the evidence stamped at `adcc35b8db` — policy set 28 passed, full `ios.yml` PR list 75 passed, 19/19 checks green — carries over unchanged; CI is re-running on the new head as the authority.

thymikee:
Reviewed current head `a1cd050d5`: not ready—one P2 maintainability/topology finding. The new progress-budget/focus behavior is correct and exact-head CI is green, including executed iOS Simulator coverage, but the implementation adds to over-budget aggregations: `RunnerTests+SynthesizedTextEntry.swift` is 503 lines, `RunnerTests+TextEntry.swift` is 607, and the new deadline source’s clock/sequence tests were placed in `UnitTests/RunnerTests+TextEntryPolicyTests.swift`, growing it from 473 to 641 lines. Extract a focused text-entry readiness/focus seam and colocate the deadline clock/sequence tests in sibling `RunnerTests+SynthesizedCommitDeadlineTests.swift`; include the replacement-route growing-prefix/ceiling case there. Physical-device CI was skipped, but simulator evidence is appropriate for this simulator text-entry change.

thymikee:
Addressed in `17cadc2468` — split along the seams that were already there, no behavior change.

| file | before | after |
| --- | ---: | ---: |
| `RunnerTests+TextEntry.swift` | 607 | 259 |
| `RunnerTests+TextEntryReadiness.swift` | — | 354 |
| `RunnerTests+SynthesizedTextEntry.swift` | 503 | 356 |
| `RunnerTests+SynthesizedCommitDeadline.swift` | 56 | 206 |
| `UnitTests/RunnerTests+TextEntryPolicyTests.swift` | 641 | 494 |
| `UnitTests/RunnerTests+SynthesizedCommitDeadlineTests.swift` | — | 164 |

**Readiness/focus seam.** `TextEntry.swift` keeps the vocabulary (failures, timings, targets), field clearing and value reading; everything that answers "which element is about to receive text, and has it taken focus" — the focus queries, the tap witness it consumes, both `focusTextInputForTextEntry` overloads, the readiness waits and the keyboard signals — moves to `RunnerTests+TextEntryReadiness.swift`. Every helper that moved kept its `private`, because its callers moved with it.

**Commit wait.** `SynthesizedTextEntry.swift` keeps the private-XCTest synthesis boundary, the replacement route and the route policies. The wait moved next to the deadline that bounds it: the two waits, the observation/pacing they poll through, and the value-free cadence line that path may log. That has a second benefit — every line that touches the polled field value is now in one file, so `apple-runner-log-redaction.test.ts` guards a single surface instead of reaching across two; its path constant moved with it, and it still passes.

**Tests.** The clock and sequence tests are in the sibling you named, including the replacement-route case: a growing prefix carries the wait past the 3s stall budget and the 10s ceiling is what ends it (`.notObserved` at exactly t=10, five characters landed). The injected clock is now defaulted, so only a test actually asking about time names it — that is what took the policy file back under budget rather than trimming prose.

Re-validated on iPhone 17 Pro / iOS 26.2: 29 policy + deadline tests passed, full `ios.yml` PR list 75 passed, 0 failures. Static gates green; `check:xctest-selection` reaches 216 methods, 0 unreachable.

thymikee:
Re-reviewed current head `17cadc2468`: the original topology P2 is mostly resolved and no behavior regression was found, but one residual P2 remains. New `RunnerTests+TextEntryReadiness.swift` is 354 lines, above the repository's 300-line implementation target, and still combines two owners: target/focus acquisition (tap witness, stabilization, focusTextInput overloads, refresh point) and readiness/confirmation (readiness waits, keyboard transition/visibility, focus corroboration). Split those seams so the extraction does not replace the old 607-line aggregation with another over-budget owner. The deadline clock/sequence suite is now correctly colocated, including replacement progress through the 10s ceiling. Android smoke is green; exact-head iOS smoke is still in progress.

thymikee:
Addressed in `4b2c958654`, and rebased onto current `main` (`72cae2bc72`, +9 commits including the #2050 runner colocation).

**The readiness extraction is split along the two owners you named.**

| file | lines |
| --- | ---: |
| `RunnerTests+TextEntryFocus.swift` (new) | 206 |
| `RunnerTests+TextEntryReadiness.swift` | 158 |

Acquisition — the one-shot tap witness, post-tap stabilization, both `focusTextInputForTextEntry` entry points and the refresh point — is now its own file. Readiness keeps the waits, the keyboard signals they read, and the focus corroboration. Both are under the 300-line target, and no file this PR touches is over it any more.

The dependency turned out to be strictly one-way: acquisition asks readiness, never the reverse. The cost of the split is two helpers losing file-private scope — `waitForTextEntryReadiness` and `keyboardBecameVisible`, both called from acquisition — and nothing else widened. I noted that in the commit rather than leaving it to be discovered.

**Rebase.** `main` moved the runner client into `packages/platform-apple` (#2050); nothing under `apple/runner/**` moved, so the rebase was clean and the `RUNNER_COMMAND_TIMEOUT_MS` reference in the timing doc comment is still accurate (it names the constant, not a path). Re-installed against the new lockfile before validating.

Re-validated at `4b2c958654` on iPhone 17 Pro / iOS 26.2: full `ios.yml` PR list 75 passed, 0 failures. Static gates green — format, lint, typecheck, `check:xctest-selection` (216 methods, 0 unreachable), and the redaction guard against the commit-wait file's new home.

thymikee:
Re-review is code-clean: the prior topology P2 is resolved. `TextEntryFocus.swift` (206 lines) owns acquisition/tap/stabilization; `TextEntryReadiness.swift` (158 lines) owns readiness/focus corroboration with one-way Focus→Readiness use. The focused deadline module and sibling sequence tests exercise the shipped append/replacement waits with a hand clock, including progress past 3s, frozen expiry at 3s, the 10s ceiling, and churn buying no time. Exact-head iOS evidence is green (75 targeted XCTest regressions and 15/15 fixture simulator E2E). Android Smoke is currently red from unrelated `wait_capture_stalled` during fixture bootstrap, not an Android assertion; no #2035 code finding, but do not mark merge-ready until that owner-action/infrastructure failure clears.

thymikee:
The Android red has cleared — all four Smoke Tests lanes are green on the exact head `4b2c958654`, and every check run on that commit is `success` (the one `skipped` is the `matrix.name` placeholder job).

The Android failure was `wait timed out for text: Automation lab`, `reason: wait_capture_stalled` — a snapshot capture that never completed during `smoke:automation-system` bootstrap, on a different screen from the form fixture this PR touches, and the same fixture change was already Android-green on the previous head. It passed on re-run ([job 98220432832](https://github.com/callstack/agent-device/actions/runs/32977635983/job/98220432832)) with no code change.

For the record, that is the third re-run-to-green infrastructure failure on this PR, in three different lanes: two `fuzz-worker` fork deaths in Coverage (filed as #2053) and this Android capture stall. Different signatures, so I have not folded them together.

Nothing outstanding from my side: no open reviews or inline comments, required checks green on the actual head, and the live simulator evidence you cite is at this same commit.

PR #2036, opened 2026-08-25
fix(ci): skip release instead of erroring when both fixtures are cached

### LINKED ISSUE #2034: Test App Build Cache fails on every run when both fixture fingerprints are already cached
`Test App Build Cache` has failed on **every run since 2026-08-24 14:42 UTC** — 60+ consecutive failures across all branches, `main` included. It is red right now on `main`.

## Root cause

#1996 changed how the release matrix is built:

```diff
-            '{ include: [ {...iOS...}, {...Android...} ] }')"
+            '[ {...iOS...}, {...Android...} ] | { include: map(select(.build)) }')"
```

That PR's own description states the intent: *"The fixture release matrix filters to entries that will actually build, so a cached-fingerprint PR starts zero release runners."* Saving those runners is right — but when **both** artifacts are already cached, `map(select(.build))` yields `include: []`, and GitHub Actions rejects an empty matrix at the workflow level. The `release` job is never created and the whole run is marked `failure`.

Both-cached is the steady state, which is why this is not intermittent: it fails except when a fingerprint actually changed.

Before #1996 the matrix always carried both entries and the per-step `if: matrix.build` guards made a cached entry a no-op — which is also why `runsOn: (if $iosBuild then "macos-26" else "ubuntu-latest" end)` exists, so a non-building entry only costs a cheap ubuntu runner.

## Evidence

- Last success: `32740236844`, 2026-08-24 **14:41:15** UTC. #1996 merged **14:42:43** UTC. Every run after it has failed.
- On the `main` run [`32854978515`](https://github.com/callstack/agent-device/actions/runs/32854978515): `total_jobs: 1`. `Resolve native fingerprint` **succeeded** (all six steps green, "Set output 'matrix'"), no `release` job was ever created, run conclusion `failure`.
- Same shape on every sampled run since.

## Why it is worth fixing rather than tolerating

- `main` shows a permanently failing workflow, so a genuine fixture-build break is indistinguishable from the current noise.
- The producer never runs, so the artifact for a *new* fingerprint is only ever built by a consumer's inline fallback — the cost this cache exists to avoid.
- The workflow header says it "Runs on every push to main and every PR so a same-head consumer can wait for a scheduled producer". A consumer waiting on a producer that errored out gets nothing to wait for.

## Suggested fix

Keep #1996's saving; just don't hand Actions an empty matrix. Have `fingerprint` publish whether there is any work, and gate the job:

```yaml
    outputs:
      matrix: ${{ steps.fingerprint.outputs.matrix }}
      has-work: ${{ steps.fingerprint.outputs.has-work }}
```
```bash
          echo "matrix=$MATRIX" >> "$GITHUB_OUTPUT"
          echo "has-work=$(jq -r '.include | length > 0' <<<"$MATRIX")" >> "$GITHUB_OUTPUT"
```
```yaml
  release:
    needs: fingerprint
    if: needs.fingerprint.outputs.has-work == 'true'
```

Alternatives, both worse: revert to always-include (gives up the runner saving #1996 bought), or emit a sentinel matrix entry (adds a state every step then has to guard against).

Worth checking as part of this: whether a skipped `release` still satisfies anything downstream that `needs:` it, so the fix doesn't just move the red.

## Regression coverage

`test/ci/` already tests workflow behaviour (`size-workflow.test.ts`, `trusted-fixture-artifact.test.mjs`). The valuable assertion is over the jq expression itself for all four fingerprint combinations — both-cached, iOS-only, Android-only, neither — pinning that the both-cached case produces something Actions accepts. That is the case that shipped broken and the one no existing test covers.

## Verification

After the fix, a push to `main` with no native change should show `Test App Build Cache` green with `release` skipped, not errored. A run where a fingerprint did change should still build and publish.

---

Found while triaging CI on #2021. This is unrelated to that PR — reporting it because it is red on `main` for everyone. I also checked whether it explained the intermittent iOS `fill` / `TEXT_INPUT_COMMIT_NOT_OBSERVED` smoke flake that started 2026-08-25 ~12:00 and **ruled that out**: this workflow has been red since 24 Aug, including hours when iOS smoke was passing. The two are independent.

thymikee:
## Summary

Fixes #2034. `Test App Build Cache` has failed on every run since 2026-08-24 14:42 UTC, including on `main`.

#1996 changed the release matrix builder to filter to only entries that will actually build:

```diff
-            '{ include: [ {...iOS...}, {...Android...} ] }')"
+            '[ {...iOS...}, {...Android...} ] | { include: map(select(.build)) }')"
```

That's correct when at least one fixture needs building, but when **both** are already cached, `map(select(.build))` yields `include: []`. GitHub Actions rejects an empty `strategy.matrix` at the workflow level, so the `release` job is never created and the whole run is marked `failure`. Both-cached is the steady state, which is why this fails on nearly every run rather than intermittently.

- `fingerprint` now also publishes a `has-work` output (`jq -r '.include | length > 0'` over the matrix).
- `release` gets `if: needs.fingerprint.outputs.has-work == 'true'`, so a both-cached run **skips** `release` instead of erroring.

I checked whether a skipped `release` breaks anything downstream that `needs:` it: nothing does — consumers (`setup-fixture-app` / `trusted-artifact.mjs`) poll `classifyProducerState`, which reads the *workflow run's* overall `status`/`conclusion`, not individual job state. A workflow with a skipped job (and no failed jobs) reports `success`, so consumers correctly see the producer as done rather than failed.

## Test plan

- Added a regression test (`test/ci/trusted-fixture-artifact.test.mjs`) that executes the real fingerprint jq/shell step across all four cache combinations (both-cached, iOS-only, Android-only, neither), pinning that both-cached produces `include: []` + `has-work=false`, and every other combination produces `has-work=true`. This is the combination that shipped broken and had no prior coverage.
- Updated the existing producer test to account for the new `has-work=` output line in `$GITHUB_OUTPUT`.
- `pnpm run test:fixture-cache` — all 12 tests pass.
- `actionlint .github/workflows/test-app-build-cache.yml` — clean.
- `oxlint` on the modified test file — clean.

After merge, a push to `main` with no native change should show `Test App Build Cache` green with `release` skipped rather than errored; a run where a fingerprint actually changed should still build and publish as before.

### DIFF
--- a/.github/workflows/test-app-build-cache.yml
+++ b/.github/workflows/test-app-build-cache.yml
@@ -33,6 +33,7 @@ jobs:
     timeout-minutes: 10
     outputs:
       matrix: ${{ steps.fingerprint.outputs.matrix }}
+      has-work: ${{ steps.fingerprint.outputs.has-work }}
     steps:
       - name: Checkout
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
@@ -94,10 +95,12 @@ jobs:
               }
             ] | { include: map(select(.build)) }')"
           echo "matrix=$MATRIX" >> "$GITHUB_OUTPUT"
+          echo "has-work=$(jq -r '.include | length > 0' <<<"$MATRIX")" >> "$GITHUB_OUTPUT"
 
   release:
     name: ${{ matrix.name }}
     needs: fingerprint
+    if: needs.fingerprint.outputs.has-work == 'true'
     runs-on: ${{ matrix.runsOn }}
     timeout-minutes: 60
     concurrency:
--- a/test/ci/trusted-fixture-artifact.test.mjs
+++ b/test/ci/trusted-fixture-artifact.test.mjs
@@ -334,17 +334,22 @@ test('producer maps each platform to its resolved lookup and matrix artifact nam
       '',
     ].join('\n'),
   );
-  const writeNodeStub = (cached) =>
+  // cachedPlatforms: which platform artifact-name suffixes (".ios", ".android")
+  // the lookup should report as already cached.
+  const writeNodeStub = (cachedPlatforms) =>
     fs.writeFileSync(
       path.join(binDir, 'node'),
       [
         '#!/bin/sh',
         'printf "%s\\n" "$*" >> "$TEST_NODE_LOG"',
-        cached ? 'printf "111"' : 'true',
+        'case "$4" in',
+        ...cachedPlatforms.map((platform) => `  *.${platform}) printf "111" ;;`),
+        '  *) true ;;',
+        'esac',
         '',
       ].join('\n'),
     );
-  writeNodeStub(false);
+  writeNodeStub([]);
   fs.chmodSync(path.join(binDir, 'node'), 0o755);
 
   const run = fingerprintStep.run
@@ -388,7 +393,17 @@ test('producer maps each platform to its resolved lookup and matrix artifact nam
     '.github/actions/setup-fixture-app/trusted-artifact.mjs find octo/repo fingerprint.android-hash.android current-head',
   ]);
 
-  writeNodeStub(true);
+  // A mistaken `length > 1` in the has-work check would pass here while
+  // wrongly suppressing this valid single-platform build.
+  writeNodeStub(['ios']);
+  const iosCached = runFingerprint('output-ios-cached');
+  assert.deepEqual(
+    iosCached.matrix.include.map(({ platform }) => platform),
+    ['android'],
+  );
+  assert.equal(iosCached.hasWork, 'has-work=true');
+
+  writeNodeStub(['ios', 'android']);
   const bothCached = runFingerprint('output-both-cached');
   assert.deepEqual(bothCached.matrix, { include: [] });
   assert.equal(bothCached.hasWork, 'has-work=false');
--- a/test/ci/trusted-fixture-artifact.test.mjs
+++ b/test/ci/trusted-fixture-artifact.test.mjs
@@ -305,14 +305,22 @@ test('Android APK repack signs the output and preserves its package id', (t) => 
   assert.match(mismatch.stderr, /did not preserve the source signing certificate/);
 });
 
-test('producer maps each platform to its resolved lookup and matrix artifact name', (t) => {
+test('producer maps each platform to its resolved lookup and matrix artifact name, and gates has-work when both are cached', (t) => {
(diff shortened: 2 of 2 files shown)


thymikee:
Re-reviewed `9eaea0bc1aa2edde686285e1c260cc436d83df4d`. The production fix is correct, the duplicated harness is gone, all checks are green, and the live workflow proves fingerprint-success + release-skipped + overall success. One P2 regression gap remains: the reduced test covers matrix cardinalities 0 and 2, but not 1. A mistaken `length > 1` would pass while suppressing a valid single-platform build. Add exactly one single-cache case to the reused harness, asserting a one-entry matrix and `has-work=true`; do not restore both permutations or the duplicate setup. Re-review after that head change.

thymikee:
Added exactly one single-cache case to the reused harness in `fb49d11`: the Node stub now takes a `cachedPlatforms` list keyed off the artifact-name suffix, so the same run/output-parsing helper exercises 0/1/2 cached cardinalities without restoring both permutations or duplicating setup. New assertion: iOS cached only → one-entry `android` matrix, `has-work=true` (would catch a `length > 1` mistake, since that reads as `has-work=false` there). All 11 tests still pass.

thymikee:
Re-review at `fb49d11cef196d8f49255aaa0f9694f999b34a42`: the P2 is resolved without duplicating setup. The reused workflow harness now proves all three matrix cardinalities: 2 (neither cached), 1 (iOS cached -> Android-only and `has-work=true`), and 0 (both cached -> empty and `has-work=false`). No remaining code findings. The prior Android smoke red was unrelated automation-host/device timing (`wait text "Alert result: cancelled"` hit its 10s deadline); this PR does not touch Android runtime paths, adjacent/prior runs passed, and the rerun has cleared setup/cache/catalog steps and is currently executing the smoke. Code is ready for human review; CI rerun still in progress.

thymikee:
Delta re-review at `bea44642fd344c9f88802bc9dadd5dc0852a165d`: clean. The 8-line test-only change strips the `has-work=` prefix before asserting the actual boolean string, matching the matrix parsing style; missing output still fails structurally, and the 2/1/0 cardinality coverage remains intact. Test App Build Cache is green with the release matrix skipped as intended. `ready-for-human` remains appropriate; current-head checks are still running, so this is code-review readiness, not yet merge-ready.

thymikee:
Current-head CI is now complete: all 19 non-skipped checks passed, and the sole skipped matrix.name is the intended no-work release path. The prior clean code verdict is now merge-ready.

PR #2038, opened 2026-08-26
Migrate audio probe to platform runtime with durable resource lifecycle

thymikee:
## Summary

This PR migrates the audio probe command from the legacy capability-based system to the new platform runtime architecture with durable resource lifecycle management. Audio probe now participates in the device ownership settlement system alongside app-log and screen-recording, enabling proper recovery and cleanup semantics.

## Key Changes

- **Audio Probe Runtime Architecture**: Introduced `AudioProbeRuntimeHost` contract and `HostAudioProbeDescriptor` for stable reconstruction of audio capture processes across daemon restarts
  - Added `packages/capture-kit/src/audio-probe-runtime.ts` with codec and lifecycle management
  - Implemented durable resource envelope support for audio probe state persistence

- **Platform Runtime Integration**: 
  - Created `audioProbeStartUse` and `audioProbeRecoveryUse` runtime operation uses in `packages/contracts/src/audio-runtime-plan.ts`
  - Added `audioProbeRuntimeOperationFacts` to all platform families (Apple, Android, Web, Vega, HarmonyOS, Linux)
  - Integrated audio probe into `PlatformRuntimeOperations` contract

- **Daemon-Side Resource Management**:
  - Created `src/daemon/audio-probe-session-resource.ts` for durable resource lifecycle (start, reattach, cleanup)
  - Created `src/daemon/audio-probe-admission-ledger.ts` for admission control
  - Created `src/daemon/audio-probe-resource-store.ts` for resource persistence
  - Added `src/daemon/audio-probe-resource-recovery.ts` for recovery operations

- **Handler Refactoring**: 
  - Rewrote `src/daemon/handlers/session-audio.ts` to use runtime binding instead of legacy capability checks
  - Updated to use `inspectFacts` and `bindDevice` for runtime operation resolution

- **Host Diagnostics Extraction**:
  - Extracted toolchain checks from `session-doctor-toolchain.ts` into platform-specific modules:
    - `src/platforms/apple/doctor.ts`
    - `src/platforms/android/doctor.ts`
    - `src/platforms/vega/doctor.ts`
    - `src/platforms/harmonyos/doctor.ts`
    - `src/platforms/web/doctor.ts`
  - Created `packages/contracts/src/host-diagnostics.ts` contract for host-scoped diagnostics
  - Created `src/platform-runtime-host-diagnostics.ts` to compose family-specific diagnostics

- **Test Migration**: 
  - Completely rewrote `src/daemon/handlers/__tests__/session-audio.test.ts` to test runtime-based audio probe with proper mocking of durable resources
  - Added `packages/capture-kit/src/audio-probe-runtime.test.ts` for codec and lifecycle testing
  - Migrated doctor tests to platform-specific locations with updated test utilities

- **Layering & Cutover**:
  - Added audio probe to wave 6 command cutovers in `scripts/layering/runtime-command-cutover-table-wave6.ts`
  - Updated session state ownership tracking for audio probe resources
  - Added `audioProbeSessionStateOwnershipViolations` validation

## Notable Implementation Details

- Audio probe descriptor includes `marker` field for managed process identity tracking, enabling precise process termination during recovery
- Status file path is managed by the daemon session store, allowing result polling after daemon restart
- Recovery is cleanup-only: the daemon cannot reattach live control after loss, but can verify completion via status file
- Host diagnostics context provides platform-neutral interface for family-specific probe implementations
- Audio probe admission ledger follows the same pattern as app-log and screen-recording for consistent resource management

https://claude.ai/code/session_01Cd7fCKcxuDU3KrXF4MF9rN


thymikee:
CI status on `669e3e8`: 19/20 checks green. The one red is the iOS XCUITest **Smoke Tests** lane ([job 98080631770](https://github.com/callstack/agent-device/actions/runs/32937199253/job/98080631770)), failing on `RunnerTests.testBareTypeUsesTappedInputWhenSoftwareKeyboardIsHidden` — the exact failure tracked in #1874 as intermittent on branches touching no Swift and no iOS production code. This branch touches neither, and the failure signature matches the issue's (`[DEBUG-1874] wait outcome=notObserved`, exit 65).

The fix already exists: #2035 addresses this test directly (its keyboard-hidden precondition becomes `XCTSkipIf` and the commit wait gains a progress-based budget). I'm deliberately not porting those ~500 Swift lines into this PR while #2035 is in review — this lane should go green here once #2035 merges and this branch takes main. I don't have job re-run permissions from this session, so the lane stays as-is until then; everything this migration touches (lint, typecheck/package, integration, both coverage shards, repo guards, and the other three smoke lanes) is green.

---

thymikee:
Not ready.

- P1: markerless audio probes can orphan a live helper after daemon loss. Start permits `resolveManagedProcessIdentity()` to yield no marker; recovery then maps markerless to `missing`, may read an in-progress status file as `completed`, and terminalizes/removes the durable resource without proving or terminating the child. Require an exact marker before publishing/adopting the resource (terminate + fail if unavailable), or retain manual recovery. Add a planted-red regression for a markerless live status file.
- P1 readiness: there is no exact-head live ScreenCaptureKit lifecycle evidence. The changed manifests cite mocked contract tests that never launch the backend. Provide start/status/stop plus daemon-loss recovery and verified cleanup on the smallest practical macOS/iOS-simulator/Android-emulator set.
- P2 size/design: GitHub reports 103 files and +3,122/-1,482; the package grows about 13.4 kB unpacked, and `audio-probe-runtime.ts` is a new 450-line multi-concern module. The #1739 pre-unit record exists and explicitly serializes R60/R62 on one branch, but the PR still needs the promised actual-vs-budget itemization and an explicit explanation of why the two command families must remain one PR rather than the already commit-separated units. Also challenge extracting codec/recovery from the live-process owner.

The iOS red is the known unrelated #1874 keyboard-state flake, not evidence against this change, but CI remains unstable. No `ready-for-human` until the lifecycle defect and exact-head evidence are resolved.

thymikee:
All three items addressed on `8df470c`; the P1 lifecycle defect is fixed and pushed, the module split landed with it, and the itemization + evidence accounting is below.

## P1 markerless orphan — fixed (`8df470c`)

Took the first option: **the marker is now required end to end.**

- `startHostAudioProbe` terminates the helper and fails the start when the spawned process exposes no exact identity — a markerless capture is never published or adopted.
- `HostAudioProbeDescriptor.marker` is non-optional and the v1 codec rejects markerless bodies, so a stored record without one (foreign or corrupted — start can no longer produce it) decodes as invalid: reattach reports `unreattachable/descriptor-invalid`, cleanup reports `cleanup-pending/manual-recovery-required`, and the resource file is retained. Manual recovery is exactly what remains for such a record — never a guessed `completed` from an in-progress status file, never `already-missing` with a live child.
- Planted-red regressions: `a markerless record with a live status file is never read as completed or missing` (asserts the liveness probe is never even consulted) and `start terminates a sampler whose exact process identity cannot be resolved`. Both fail against the previous behavior by construction.

## P2 module split — done (`8df470c`)

`audio-probe-runtime.ts` is now four single-concern modules: `audio-probe-descriptor.ts` (type + v1 codec), `audio-probe-status.ts` (status-file reads + completion finalization, shared by live and recovery), `audio-probe-recovery.ts` (cleanup-only recovery pair), and `audio-probe-runtime.ts` keeping only the live-process owner (start, guarded handle, capture-operations factory). Eager-closure budget rows moved for the three new files (+3 each on the capture-kit and session-teardown closures).

## P2 actual-vs-budget itemization (production lines, tests excluded, vs the #1739 pre-unit record)

| Zone | R60 audio | R62 doctor | Shared threading |
| --- | ---: | ---: | ---: |
| `src/daemon/` | **−16** (+303/−319) | **−262** (+66/−328) | +77 |
| root outside daemon | +40 | +337 | −19 |
| `packages/` | **+779** | +48 | +172 |
| `scripts/` (gates) | — | — | +165 |
| tests | +325 | −14 | +97 |

Against budget:
- **R60**: packages +779 sits inside the +550..900 band. Root was budgeted −80..−150 and came out +24: the daemon itself shrank (−16 — the durable machinery costs almost exactly what the hand-rolled lifecycle saved), and the overage is the composition host (`platform-runtime-audio-probe-host.ts`, +93), which must live at the root because it owns the lazy `startMacOsAudioProbeProcess`/web-provider edges the ADR bars from packages.
- **R62**: daemon −262 is inside the −200..−300 band. The packages/scripts budget (+350..550) assumed probe bodies land in packages; they landed in `src/platforms/*/doctor.ts` (+312) instead — the itemized placement choice the pre-unit record reserved ("otherwise to the family's root platform module"): the android/apple/vega probes depend on `adb-executor`/`ime-lifecycle`/`runner-client`/`tool-provider`, root utilities that have not migrated, so a package home would have dragged those edges across the boundary. The mass moved zones, it did not grow: doctor's whole-tree production net is +123, all of it the neutral gateway (`host-diagnostics` facet +48, composition binding +58) plus the `host` discriminator arm.
- The +13.4 kB unpacked in the size report is this same shape: `doctor.js`/`doctor2.js`/`audio-probe-result.js` appear as chunks while `session2.js` drops −6.2 kB — the packed bundle now carries the durable envelope/fence/recovery machinery audio never had, which is the gap R60 existed to close, not incidental growth.

## Why one PR (and the offer to split)

The two units are commit-separated (`f70a051` audio, `7854e7d` doctor) and *could* be two PRs. They share the churn that made me keep them together: the wave6 cutover table, the eager-closure budget rows, the R7/R11 baselines, the platform coverage manifests, and the session-route threading (`session.ts`/`request-handler-chain`/`request-router`/`daemon-runtime`) are each touched by both units, so two branches would carry conflicting edits to the same gate files and merge serially anyway. The pre-unit record serialized R60/R62 on one branch for exactly that reason. If you still want two PRs I'll split them — the two unit commits make it mechanical; the five follow-up commits (coverage repointing, review fixes) need untangling by file, which is bounded work.

## P1 readiness — the honest gap, restated

I cannot produce exact-head live ScreenCaptureKit evidence from this session: it runs in a Linux container, and the host capture path requires a darwin host by definition. The pre-unit record on #1739 flagged this before implementation. The minimal live set this PR needs on hardware, exact head:

1. macOS session: `audio probe start` → `status` mid-run → `stop`, verifying `macos-screencapturekit` backend, non-empty rms/peak buckets, and helper exit after stop.
2. iOS simulator and Android emulator on the same darwin host: one `start`→`stop` each (same helper, different session families).
3. Daemon-loss recovery: kill the daemon mid-probe, restart, and verify reconciliation terminates the helper by its exact marker (no orphan in `ps`) while the resource file resolves without a guessed result.
4. `doctor` on the darwin host: byte-compare check ids/summaries against a pre-branch run (the relocation is asserted wire-identical by tests, but the live confirmation belongs to this set).

Local/CI state on `8df470c`: typecheck, lint, format, fallow, layering (R60/R62 rows green in the wave6 table), capture-kit + daemon suites (the two known root-only chmod failures excepted), and all six platform coverage smokes pass.

---

thymikee:
Re-review of `8df470c`: the markerless-orphan P1 is resolved correctly. Exact process identity is required before publication; markerless start terminates and fails; stored markerless descriptors decode invalid and remain manual-recovery work; the two targeted planted-red cases cover both paths. The module split and actual-vs-budget accounting also address the prior P2, and all 19 non-skipped checks are green.

The remaining blocker is readiness evidence, not code: the author’s own exact-head update confirms there is still no Darwin/ScreenCaptureKit start/status/stop or daemon-loss cleanup run. Keep `ready-for-human` off until that practical lifecycle evidence is posted.

thymikee:
Exact-head live evidence for `8df470c1baa6b5b333ea59af61df6c7cabc59392` from a detached local worktree on macOS. Fresh preparation passed: `pnpm install --frozen-lockfile`, `pnpm build`, and `pnpm build:android`.

## P1: timed probes never advance past their first bucket

The real ScreenCaptureKit backend starts successfully after granting Screen Recording permission, but its deadline loop does not progress:

- macOS / Calculator: `audio probe start 10 250` returned `backend=macos-screencapturekit`, `state=running`, `elapsedMs=176`, `sampleCount=1`. I played `/System/Library/Sounds/Glass.aiff` during the window. At 63 seconds the exact helper PID was still alive and `audio probe status` was byte-for-byte stuck at the same `elapsedMs=176`, `sampleCount=1`, `heard=false` instead of completing at 10 seconds.
- iOS Simulator / iPhone 17 Pro: `audio probe start 3 250`; after 4 seconds, status was still `running` at `elapsedMs=163`, `sampleCount=1`.
- Android Emulator / Pixel 7 CI: `audio probe start 3 250`; after 4 seconds, status was still `running` at `elapsedMs=167`, `sampleCount=1`.

Explicit `audio probe stop` succeeded on all three and terminated each helper, returning `state=stopped`, but automatic duration completion is broken on every supported host-capture family. The shared Swift helper is the common route; inspect the synchronous semaphore/async Task boundary in `runAudioProbe` because the initial `flushRunning()` is published but subsequent sleep/flush/finish work never runs in these live executions.

## P1: restarting in the same session consumes stale status

After the macOS 10-second probe was explicitly stopped, `audio probe start 30 250` in the same session immediately returned the previous probe's 10-second snapshot (`durationMs=10000`, old `startedAt`, `elapsedMs=176`). `waitForFirstStatus` accepted the existing `audio-probe.json` before the new helper could establish a fresh generation. Clear or generation-bind the status artifact before waiting; add a planted-red restart-after-stop regression.

## Positive evidence: hard-crash cleanup recovery works

In a fresh macOS session I started a 30-second probe. The durable envelope recorded helper PID 58303 with an exact start-time/command marker. I SIGKILLed only the isolated daemon PID 57761; PID 58303 remained alive, proving an actual daemon-loss orphan. Running `session list` restarted the exact-head daemon and triggered recovery:

- helper PID 58303 was gone afterward;
- the envelope became `lifecycle=completed`, `metadata.phase=completed`, `cleanupStatus=cleaned`;
- daemon log: `examined=1, reconciled=1, retained=0`.

So the new exact-marker recovery path is validated, but the normal timed lifecycle and restart path are not merge-ready. No coordinator-created helper/session remains, and the dedicated Pixel 7 CI emulator was shut down.

thymikee:
Both live P1s are fixed on `60f262b` — and thank you for the daemon-loss run; the exact-marker recovery validation was the piece only hardware could give.

**Timed-probe stall.** Confirmed at the boundary you pointed at, with one refinement: the async body was not MainActor-bound (the first `flushRunning` demonstrably ran), so the stall is the bridge itself — `runAudioProbe` parks the one-shot CLI's main thread in `semaphore.wait()` with no run loop ever spinning, and the unstructured `Task`'s capture loop dies at its first `Task.sleep` suspension: the timer continuation never gets scheduled again, which is exactly the frozen `elapsedMs=176 / sampleCount=1` on all three families (one shared helper). Rather than argue with the concurrency runtime, the probe is now fully synchronous: `SCShareableContent.getWithCompletionHandler` / `startCapture(completionHandler:)` / `stopCapture(completionHandler:)` bridged through semaphores — the identical pattern `captureSurfaceScreenshot` in the same file already uses in production — and a plain `Thread.sleep` cadence loop. Every continuation stays on ScreenCaptureKit's own callback queues; nothing depends on Swift-concurrency resumption anymore.

**Stale-status restart.** `startHostAudioProbe` now clears the status path with `fs.rm(..., { force: true })` before the spawn, so any file observed by `waitForFirstStatus` was written by the new helper — sound because the previous handle's `finish()` has already terminated and awaited its process before a restart can run, so nothing else writes to that path. Planted-red regression added: a pre-seeded stale file plus a helper that dies before publishing must fail the start and leave no file behind (previously the stale file made that start "succeed" instantly with the old snapshot).

Honest limits from this container: I cannot compile or execute Swift here, so the helper rewrite is validated by inspection against the in-file production pattern only — the TS side (typecheck, lint, format, fallow, 73 capture-kit/daemon audio tests including the new regression) is green. The live re-run this needs on your side, exact head: `pnpm build` (helper rebuild), then the macOS `audio probe start 10 250` letting it self-complete at ~10s with multiple buckets, the restart-after-stop sequence returning fresh `startedAt`/`durationMs`, and one iOS-sim/Android-emulator start each to confirm the shared route.

---

thymikee:
CI on `60f262b`: 19/20 green — including the macOS lane that rebuilds the helper, so the synchronous probe rewrite compiles and its lane passes. The one red is the iOS-simulator fixture Smoke lane ([job 98125652078](https://github.com/callstack/agent-device/actions/runs/32952126129/job/98125652078)): `smoke:form-input` → `fill full name` → `TEXT_INPUT_COMMIT_NOT_OBSERVED` with "Ada Lovelace" — the second documented #1874 symptom, which #2035 root-caused as the `field-name` placeholder collision on the penalized replacement route and fixes in the fixture. This branch touches neither the fixture nor the iOS text-entry path; I'm not porting the in-review fixture change to avoid churn against #2035 — the lane resolves when it merges and this branch takes main. Same disposition as the earlier XCUITest-lane note, different lane, so recording it once here.

---

thymikee:
Completion pass is pushed at `a60a009913` (on top of `60f262b9aa`).

What changed:

- timed probes now remain alive until their requested duration and publish a natural `stopped` / `completed` result;
- restart removes the prior status publication before spawning, so a new probe cannot adopt stale completion data;
- the planted stale-status regression is green;
- Apple and Android runtime fixtures now provide the required audio-probe host contract.

Validation on the exact reviewed/pushed code:

- `pnpm exec vitest run packages/capture-kit/src/audio-probe-runtime.test.ts packages/platform-apple/src/runtime.test.ts packages/platform-android/src/runtime.test.ts --reporter=dot` — 3 files, 73 tests passed.
- `pnpm build:macos-helper` — passed.
- `pnpm build` — passed.
- `pnpm check:affected --base HEAD^ --run` — all 20 selected checks passed; 286 test files / 1,944 tests passed. One 2.68s test was reported inside the non-failing 2x load-variance band.

Live evidence:

- macOS exact head: `audio probe start 1 250` naturally reached `state=stopped`, `reason=completed`, `elapsedMs=1000`, `sampleCount=5`.
- macOS same-session restart: `audio probe start 2 250` immediately published a fresh `durationMs=2000`, new `startedAt`, and `elapsedMs=170`, rather than adopting the previous 1-second result.
- iOS Simulator (iPhone 17 Pro): 2s / 250ms probe naturally completed with 9 samples.
- Android Emulator (`Pixel_7_CI`): 2s / 250ms probe naturally completed with 9 samples.
- daemon crash recovery: after terminating the isolated daemon while the helper remained alive, the replacement daemon reconciled the durable capture (`examined=1 reconciled=1 retained=0`), removed the helper, and closed with `cleanupStatus=cleaned`.

The iOS, Android, and crash-recovery runs exercised the locally equivalent completion/reconciliation fix; the subsequently updated exact PR head was rechecked on macOS for both natural completion and stale-status rejection. No positive-audio claim is made: the macOS run proved cadence/lifecycle only.

thymikee:
P1: helper death after the first running checkpoint is misreported as successful completion. `startHostAudioProbe` discards `process.wait` after startup, while live `status()` only rereads the JSON file (`packages/capture-kit/src/audio-probe-runtime.ts:69,131-134`). If the helper later exits after publishing `state: 'running'`, status remains running forever and stop fabricates a normal stopped result. The daemon-recovery path compounds this: when the exact PID is missing, `audio-probe-recovery.ts:39-46` accepts any persisted status—including the normal running checkpoint—and unconditionally finalizes it as `completed`, terminalizing/removing the durable envelope. Please retain/observe the helper terminal result, and only treat a terminal `state: 'stopped'` publication as completed recovery; a running checkpoint plus child exit/missing PID must surface missing/uncertain failure according to the durable contract. Add planted-red live-handle and daemon-restart regressions for running checkpoint + child exit. Remove `ready-for-human` until fixed.

thymikee:
Fixed on `2d4f891` — the defect was exactly as described, in both halves.

**Live handle.** `process.wait` is now observed for the handle's lifetime. The rule: only a terminal `stopped` publication outranks an observed exit, because the helper writes it before exiting — so an exit with a non-terminal status file means the capture died. `status()` and `finish()` then fail with the helper's exit detail (stderr/stdout/exit code) instead of reporting the stale checkpoint or fabricating a stop. The thrown stop rides the coordinator's existing failed-terminal-transition path — the record is preserved with the primary error and resolves through descriptor cleanup (dead PID → `already-missing` → cleaned), which is the durable contract's uncertain-failure shape rather than a new one.

**Recovery.** The marker-missing arm treats only `state: 'stopped'` as a completion. A running checkpoint with the exact PID gone reports `missing`, which the recovery pipeline terminalizes as `recoveryStatus: 'already-missing'` — envelope closed honestly, no completion metadata, no fabricated result.

**Planted-red regressions**, both failing against the previous behavior by construction:
- capture-kit live handle: running checkpoint + child exit → `status()` and `finish()` reject with the exit detail, `forceCleanup` still confirms cleaned; plus the recovery-op case (running file + gone PID → `missing`, not `completed`).
- daemon restart: a new `src/daemon/__tests__/audio-probe-resource-recovery.test.ts` (mirroring the screen-recording recovery test) plants an orphaned envelope with a running checkpoint and a missing PID, runs `recoverAudioProbeResourceAfterDaemonLock` through a real gateway binding, and asserts the envelope terminalizes as `already-missing` with **no** `heard`/`sampleCount` completion metadata.

Natural completion is unaffected: a terminal file read wins over the exit flag in both paths, so the self-completed run you validated still reports its own result. TS side fully green (82 tests across capture-kit + daemon audio suites, typecheck, lint, format, fallow); no Swift touched this round.

---

thymikee:
CI on the rebased head `5c8d459`: the Android emulator Smoke lane ([job 98142870433](https://github.com/callstack/agent-device/actions/runs/32957675890/job/98142870433)) failed in `smoke:automation-system` at `wait text "Alert result: cancelled"` — `wait_deadline_exceeded` at 10s with `captureTruncated: true` after 14 captures. Not this PR's failure: the branch touches nothing in the alert/wait/Android-capture path, the identical content passed this exact lane on `2d4f891` right before the rebase, and the only commit the rebase brought in is `b40debf` (CI fixture-caching release-skip — unrelated). This reads as emulator UI timing: the alert was dismissed but the result text didn't render inside the wait deadline. I don't have job re-run permission from this session — one re-run should clear it.

---

PR #2039, opened 2026-08-26
refactor: centralize common input field audience metadata

### LINKED ISSUE #2027: Unify the input-field "audience" model (retiredField / AI SDK hidden / MCP admission)
Surfaced by #2023. The concept "who may write this input field, on which surface" is now expressed **three** times, each a separate name-keyed mechanism:

- `retiredField()` — `src/commands/command-input.ts`: refuse with migration guidance, omit from schema.
- `ALWAYS_HIDDEN_FIELDS` + `omitSchemaProperties`/`omitHidden` — `src/ai-sdk/index.ts`: hide session/stateDir/etc. from the model-facing schema and strip from input.
- `OPERATOR_INPUT_GUIDANCE`/`CONFIG_LOADER_GUIDANCE` + the deny-by-default admission boundary — `src/mcp/command-tools.ts` (added in #2023): operator/credential/config keys refused, unknown keys rejected.

Root cause: the ~21 shared common fields exist only as **parallel enumerations by name** — `commonProperties()` (schema), `readCommonInput()` (readers), `commonToClientOptions()` (projection), the `CommonCommandInput` type, and `COMMON_COMMAND_SUPPORTED_FLAG_KEYS` — none carrying metadata beyond a description. So any *policy* about a field (retired? operator-only? model-visible?) forces a new name-keyed map elsewhere, plus a test to keep the maps in sync.

## Proposal

One declarative common-field table, each row `{ key, schema, reader, flagKey?, audience }` where `audience ∈ 'model' | 'operator' | 'retired'` (+ the AI-SDK-pinned extras like `session`/`platform`). Derive from it:

- the JSON schema (`commonProperties`),
- the readers (`readCommonInput`) and client projection (`commonToClientOptions`),
- the supported-flag list,
- the MCP admission allowlist and its guidance strings,
- the AI SDK hidden set.

Then `retiredField` becomes `audience: 'retired'`, the operator refusals become `audience: 'operator'`, and the guidance strings become **mechanically derived** (`buildPrimaryEnvVarName` already exists) instead of hand-written per key — folding in the smaller "guidance strings are hand-written but derivable" cleanup.

## Payoff

The security-relevant surface becomes pure data: adding/reclassifying a field is a one-row edit that the schema, readers, projection, and both surface boundaries all honor, with no cross-map sync test. In #2023, the operator/endpoint/path removal (items 2–3) would have been ~10 row edits instead of ~250 lines across three mechanisms.

Purely internal; no wire/behavior change intended (a characterization test over `listCommandTools()` output + the CLI schema pins that).

kkkhs:
## Summary

Closes #2027. Common command input fields now live in one declarative table that carries schema, reader, client projection, CLI flag key, and surface audience metadata. MCP operator guidance/refusal and AI SDK hidden common fields derive from that same table, and retired command fields use the same `audience` vocabulary.

Scope: 5 files, command input/MCP/AI SDK only.

## Validation

- `pnpm format:check`
- `pnpm check:quick`
- `pnpm exec vitest run --project unit-core src/mcp/__tests__/command-tools-operator-inputs.test.ts src/mcp/__tests__/command-tools.test.ts src/mcp/__tests__/command-tools-parity.test.ts src/ai-sdk/__tests__/index.test.ts src/commands/__tests__/command-surface-metadata.test.ts src/commands/__tests__/command-explain.test.ts src/__tests__/cli-record-flag-delivery.test.ts`
- `PATH=/data00/home/kuanghuangshuo.12/.nvm/versions/node/v22.23.1/bin:$PATH pnpm build`

### DIFF
--- a/src/ai-sdk/index.ts
+++ b/src/ai-sdk/index.ts
@@ -4,6 +4,7 @@ import { AppError, type NormalizedError } from '@agent-device/kernel/errors';
 import { createAgentDeviceClient } from '../agent-device-client.ts';
 import type { AgentDeviceClient } from '../client/client-types.ts';
 import type { JsonSchema } from '../commands/command-contract.ts';
+import { commonInputKeysHiddenFromAiSdk } from '../commands/command-input.ts';
 import { resolveCommandFrameworkTier } from '../core/command-descriptor/registry.ts';
 import { createCommandToolExecutor, listCommandTools } from '../mcp/command-tools.ts';
 import { formatToolErrorText } from '../mcp/tool-error.ts';
@@ -49,21 +50,11 @@ export type AgentDeviceTools = {
   toolApproval?: Partial<Record<string, ToolApprovalStatus>>;
 };
 
-// Hidden unconditionally, from both the schema the model sees and the input
-// `execute` forwards to the shared executor:
-//  - `session` is always pinned by this factory — the whole point is that a
-//    tool call can never target a session other than the one passed in.
-//  - `stateDir` selects which daemon state directory (and therefore which
-//    daemon/session namespace) a call resolves against. Left model-visible,
-//    it would let a call escape the pinned session into another daemon's
-//    state entirely, defeating that guarantee.
-//  - `mcpOutputFormat`, `includeCost`, `responseLevel` are MCP tool-config
-//    knobs, not command arguments; irrelevant here since `execute` below
-//    returns structuredContent directly and never reads a tool's rendered
-//    text, and shaping the response is this factory's decision, not the
-//    model's.
-const ALWAYS_HIDDEN_FIELDS = [
-  'session',
+// MCP tool-config knobs, not command arguments; irrelevant here since
+// `execute` below returns structuredContent directly and never reads a tool's
+// rendered text. Common pinned fields such as `session`/`platform` come from
+// the common input-field table.
+const AI_SDK_HIDDEN_MCP_FIELDS = [
   'stateDir',
   'mcpOutputFormat',
   'includeCost',
@@ -104,8 +95,10 @@ export async function createAgentDeviceTools(
   });
   const executor = createCommandToolExecutor();
 
-  const hiddenFields = new Set<string>(ALWAYS_HIDDEN_FIELDS);
-  if (platform) hiddenFields.add('platform');
+  const hiddenFields = new Set<string>([
+    ...AI_SDK_HIDDEN_MCP_FIELDS,
+    ...commonInputKeysHiddenFromAiSdk({ platformPinned: Boolean(platform) }),
+  ]);
 
   const tools: ToolSet = {};
   for (const definition of listCommandTools()) {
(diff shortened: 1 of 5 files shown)


thymikee:
Thank you for the contribution! It's not ready yet: the refactor centralizes metadata, but it does not establish the typed owning seam #2027 requires.

- `selectionOptionsFromFlags` remains a hand-written parallel identity map for `noRecord` and the device-selection fields. Direct CLI readers (`is`/`find`/`wait`/`settings`) still bypass `COMMON_INPUT_FIELDS`, so future changes require table-plus-manual-map synchronization. The paragraph-long comment explaining why the duplication is acceptable is the warning sign here: derive this projection from the same rows, including `deviceTarget -> target`, or model it as an explicit typed projection owned by those rows.
- `COMMON_INPUT_FIELDS` is widened to `readonly CommonInputFieldDefinition[]`; `key` is `string`, while `read`, `project`, `flagKey`, and `fromCliFlags` are unrelated optionals. A row can advertise a CLI flag yet silently fail to forward it. The current empty-record decoder call merely rediscovers the destination key at runtime. Use discriminated/generic row variants that link input key, schema alias, CLI source, and client projection in the type system, and model omitted surfaces explicitly.

Keep the table private behind the narrow query functions. Current checks are green, but this ownership/type gap blocks readiness.

thymikee:
P1: the new table still does not own the full CLI projection and its row type cannot enforce the promised invariant. `COMMON_INPUT_FIELDS` uses `key: string` plus unrelated optional `flagKey` / `fromCliFlags` / `read` / `project` (`src/commands/command-input.ts:64-80`), and `commonCommandInputFromFlags` discovers the output key by executing a decoder against `{}` (`:199-208`). A row can therefore advertise a flag while silently omitting forwarding or client projection. More concretely, `selectionOptionsFromFlags` remains a handwritten parallel common-field map (`src/commands/cli-grammar/common.ts:80-100`) used by direct readers such as find/is/wait, so #2027's one-row-edit contract is still false. Make the row model discriminated/generic and derive both reader-input and direct client-options CLI projections from it (including `deviceTarget -> target`); add characterization coverage showing representative direct and field-based readers receive the same common flags. Do not apply `ready-for-human` until fixed.

thymikee:
[P1] Make the common-input table the actual typed owner of CLI admission, reading, and client projection. `CommonInputFieldDefinition` currently permits arbitrary `key: string` plus independently optional `flagKey`, `fromCliFlags`, `read`, and `project`; `commonCommandInputFromFlags` even runs a decoder against `{}` just to recover an output key. The type therefore cannot require a CLI field to be parsed and projected together. Direct readers (`settings`, `wait`, `find`/`is`) still obtain common selection options through the handwritten `selectionOptionsFromFlags`, so `deviceTarget -> target`, `--no-record`, and future fields still require parallel edits and the promised one-row invariant is false. Use discriminated/generic row variants that encode CLI flag, reader input key, and client-options key, derive `selectionOptionsFromFlags` from the table, and add representative parity tests between field-based and direct readers.

Audience/MCP/AI derivation otherwise appears correctly centralized. Do not modify the branch.

Issue #1370, opened 2026-07-23
Next major: scheduled breaking cleanups (legacy batch-step shape, deprecated rotate client surface, client-types narrowing)


thymikee:
Tracking issue collecting the breaking cleanups deliberately deferred to the next major version, so they ship together instead of scattering. All three were audited and verified in the July 2026 overengineering audit (#1367 arc).

## 1. Remove the legacy batch-step `{command, positionals, flags}` shape

`src/cli/batch-steps.ts`: ~90 of ~140 lines exist solely to translate the deprecated shape into the current `{command, input}` shape (`readLegacyCliBatchStep`, `assertLegacyBatchStepKeys`, `readLegacyPositionals`, `readLegacyFlags`, `legacyStepToStructuredStep`, `cliFlagsFromBatchStep`). A runtime deprecation warning ("deprecated and will be removed in the next major version") has been live since the structured shape became primary, and `cli-help` documents the boundary. Removal leaves `readCliBatchStepsJson` as a thin JSON-array validate + `parseBatchStepRuntime` map.

## 2. Remove the deprecated `rotate` SDK client surface

The runtime-layer duplicate was already removed in #1367 (the CLI alias in #1277). What remains is the published back-compat surface, annotated "Retained until the next major":
- `agent-device-client.ts` `rotate` wrapper (independent one-line delegation to `orientation` with an `action: 'rotate'` relabel)
- `client-types.ts` `RotateCommandOptions` / `RotateCommandResult` / `DeprecatedCommandClient`

## 3. Client-types narrowing (`Record<string, unknown>` → typed command results)

The largest remaining simplification (~−550 lines by earlier estimate), and the reason this needs a major: it narrows public client return types, which is API-changing for SDK consumers even though it strictly increases type information. The command-descriptor foundation it builds on (#905–#911) is merged. Needs a design pass on migration/compat notes before implementation — it is a typed-results migration, not a mechanical dedup.

## Non-goals

Deprecated `--session-locked` / `--session-lock-conflicts` and `replay export --format` were already removed in #1367 (pre-1.0 breaking, CHANGELOG'd) and are not part of this bundle.

thymikee:
## 0.21 scope decision — unblocks #1739 Wave 2

Re-audited against `origin/main` after Wave 6 and against PR #2038. The approved 0.21 breaking-cleanup slice is:

[...]

PR #2046, opened 2026-08-26
refactor: remove next-major compatibility surfaces

thymikee:
## Summary

Remove the 0.21 compatibility surfaces that block the ADR-0019 perf migration:

- remove aggregate perf forms and require focused `frames`, `memory`, `cpu`, or `trace` operations across CLI, daemon, SDK, help, and docs, with exact migration guidance for retired syntax
- remove legacy batch `positionals`/`flags` authoring in favor of structured `input`, and convert the reusable performance scenario
- remove the deprecated SDK rotate wrapper while preserving the CLI migration diagnostic
- classify diagnostic `trace` as non-platform execution without changing trace-log behavior

Stacked on #2038. Scope and follow-up are recorded on #1370. Existing partial HarmonyOS perf support is unchanged; this adds no HarmonyOS-specific implementation.

## Validation

- planted-red: 16 expected failures across the aggregate perf, batch, SDK rotate, and trace seams before implementation
- focused: 11 files, 467 tests passed after rebasing onto the current #2038 head
- `pnpm check:affected --run`: all locally runnable checks passed on the amended implementation, including 511 related test files / 4,105 tests, provider integration, build/package, layering, wire compatibility, docs, and structural gates
- `claude -p --model claude-fable-5`: adversarial review completed; fixed all three findings (stale batch docs, missing `perf memory` default coverage, and two orphaned exports)


thymikee:
Reviewed current head `5f7aec824`: the intended compatibility-removal commit is code-clean in pre-rebase review, but this head is not ready. The branch still carries the now-merged #2038 stack and GitHub reports a merge conflict, so rebase/reconcile onto current `main` and rerun validation. `Coverage (1)` also failed at the shard runner without a surfaced test name; treat it as owner-action red for this head and inspect the rerun if it persists. All other completed checks, including the four smoke lanes, are green. Re-review is required on the rebased head.

thymikee:
## Code quality review

Reviewed this PR's own commit (`5f7aec824f`) against `main` — 38 files, +277/−780. The `#2038` commits below it are already merged.

**Verdict: request changes.**

The direction is right and the deletions are real. Dropping `session-perf-legacy.ts`, `perf-legacy-cpu.ts`, the legacy output formatters, and the `rotate` wrapper removes complexity rather than relocating it. Docs and the changelog are consistent, and no file crosses 1000 lines — everything touched shrank or held.

But in three places the PR deletes the legacy branch and leaves behind the structure that only that branch justified.

---

### 1. `src/cli/batch-steps.ts` is now a weaker duplicate of a reader we already have

The only reason this file existed was to translate legacy `positionals`/`flags` through `readInputFromCli` + `assertCommandPositionalArity` — genuine CLI-grammar work. That translation is gone. What remains is generic structured-step reading, and `src/commands/batch/metadata.ts:125` (`readBatchStep`) already does it, more thoroughly:

| check | canonical `readBatchStep` | new CLI copy |
| --- | --- | --- |
| step is a record | yes | yes |
| `command` is a valid command name | yes | **no** |
| `input` is an object | yes | **no** |
| unknown top-level keys rejected | yes | **no** |

I ran the new reader directly against this branch. All four of these are accepted:

```
{"input":{}}                              -> accepted (no command at all)
{"command":42,"input":{}}                 -> accepted
{"command":"open","input":"nope"}         -> accepted
{"command":"open","input":{},"bogus":1}   -> accepted
```

The deleted legacy branch rejected two of them. And `input: "nope"` isn't harmless: `src/cli.ts:500` spreads `step.input` before anything validates it, so the string becomes `{0:'n',1:'o',...}` and then passes `readBatchStepInputObject` downstream. A CLI user never sees `Batch step 1 input must be an object.`; an MCP caller does.

`isStructuredBatchStepShape` also still tests `!('positionals' in step) && !('flags' in step)`, which the throw above it already guarantees — dead conditions left over from the removed branch.

**Ask:** keep only the retired-fields diagnostic in this file and delegate the rest to `readBatchStepRecord` + `assertAllowedKeys(['command','input','runtime'])` + `readStructuredBatchCommandName` + `readBatchStepInputObject` + `parseBatchStepRuntime`. That deletes `isStructuredBatchStepShape`, `readStructuredBatchStep`, `normalizeCliBatchSteps`, and the six-line `noun`/`fields` pluralization block, and closes the validation gap in the same move.

### 2. "Which token means removed aggregate perf" is now spelled three different ways

Three parsers, three definitions of one concept:

- `src/cli/parser/args.ts:397` — `metrics`
- `src/commands/perf/index.ts:140` — `undefined`, `metrics`, any `isPerfAction(...)`
- `src/daemon/handlers/session-observability.ts:268` — `null`, `metrics`, `sample`

The message constant is shared; the rule is not. So `perf start` gets the migration message from the CLI and `perf area must be frames, memory, cpu, or trace` from a raw daemon call. Put one predicate next to the message in `packages/contracts/src/perf.ts` and call it from all three.

Inside `readPerfPositionals`, two of the three guards are already dead: `positionals.length === 0` is covered by `readPerfArea(undefined)`, and `normalized === 'metrics'` is covered by `readPerfArea('metrics')`. With a shared predicate the whole thing collapses into `readPerfArea`, and `removedAggregatePerfError()` disappears.

### 3. The daemon union got narrower but kept the cast the wide union justified

`PerfCommandRequest`'s non-native arm is now `'frames' | 'memory'`. `frames` provably carries no action, no `kind`, and no `out` — `validatePerfOutFlag` and `validatePerfKindFlag` both restrict those to `memory snapshot`. Yet the arm still declares `action: 'sample' | 'snapshot'` for both, which forces `action as 'sample' | 'snapshot'` at `src/daemon/handlers/session-observability.ts:236` and keeps `toPerfMemoryAction` — a function that re-checks at runtime what the type already guarantees, and throws on a case that cannot occur.

Split the arm into `{ area: 'frames' }` and `{ area: 'memory'; action: 'sample' | 'snapshot'; kind?; out? }`, with `validatePerfAreaAction` as the narrowing point. The cast and the helper both delete.

---

### Should fix

**Retired commands should be data, not `if` statements.** `normalizeCommandAlias` now holds two hand-written retirement branches (`metrics`, `rotate`) sitting directly on top of `normalizeCliCommandAlias`, which is registry-driven. This wave will add more. A `RETIRED_CLI_COMMANDS` table in `src/commands/cli-command-aliases.ts` makes the resolver a lookup again.

**Two tests lost their point.**

- `test/integration/provider-scenarios/android-lifecycle.test.ts:983`: `invalidPerfAction` was retargeted from an invalid-action check to a removed-aggregate check. The invalid-action coverage is gone and the variable name no longer describes what it does. Use `['frames', 'poll']` — the removed-aggregate case is already covered by the new daemon `test.each`.
- `src/commands/batch/cli.test.ts:133`: `assert.doesNotMatch(result.stderr, /unknown legacy field\(s\): input/)` pins a message that no longer exists anywhere, so it can never fail. The test name still says "without falling back to legacy parsing".

**`0.21` is hardcoded into runtime error strings.** `package.json` is `0.20.11-dev`, so this is a forecast rather than a fact, in 5 source locations plus 4 docs files. The PR isn't consistent with itself either: the perf message names the version, the batch message doesn't. Suggest dropping it from the runtime strings and keeping it in the changelog and docs, where a release rename is already expected.

### Nits

- `src/core/command-descriptor/registry.ts:1091` writes `{ kind: 'none' }` inline. `NO_PLATFORM_EXECUTION` is declared at line 266 and used at ~10 other sites — it's the sibling of the `LEGACY_PLATFORM_EXECUTION` this line replaced.
- `REMOVED_AGGREGATE_PERF_ERROR_MESSAGE` breaks the `PERF_*` sort in three import/export lists. `PERF_AGGREGATE_REMOVED_ERROR_MESSAGE` keeps the family and the convention.
- `scripts/layering/bin-alias-fast-path.ts:48` and `bin-alias-fast-path.test.ts:294` still name `metrics`, while the parallel fixture at line 277 was updated.

---

I didn't re-run the suite — the review worktree has no install of its own and `@agent-device/*` resolves to the main checkout there. Everything above comes from reading the code, except the batch-step probe, which I ran directly against this branch's `batch-steps.ts`.

thymikee:
Rebased the PR's single commit onto current `main` (`80a3fdc`) and force-pushed as `69bff60813`.

Addressed the review findings:

- batch steps now use the canonical structured-step reader and validation path, while retaining exact migration guidance for removed `positionals` / `flags` fields
- aggregate perf removal policy now has one contracts-owned token predicate and one migration message used by CLI and daemon paths
- the daemon perf request is a discriminated union with validation as the narrowing/construction point; the broad cast and duplicate token guards are gone
- restored invalid-action provider coverage, removed the ineffective legacy assertion, centralized retired top-level aliases, used `NO_PLATFORM_EXECUTION`, and removed stale `metrics` comments/fixture text
- runtime migration errors no longer hardcode `0.21`; version-specific context remains in release docs/help

The previous Coverage (1) failure reproduced locally as a stale file-size ratchet: `src/__tests__/client.test.ts` had shrunk from its pinned 1598 lines to 1592. The pin is now exact and its planted-red structural test passes. The only later slow observation was the known load-sensitive gesture test; it passed both in isolation and in the full affected run, so no timeout was weakened.

Validation after the final rebase:

- `pnpm check:affected --run` — passed all runnable checks
- related Vitest suite — 427 files, 3,540 tests passed
- build, package verification, typecheck, lint, format, layering, wire compatibility, docs, provider integration, and structural model gates all passed

The newly merged interaction-facade retirement caused the final rebase conflicts. Resolution keeps `main`'s deleted façade and new `@agent-device/contracts/wait` location while preserving this PR's removal of the deprecated `rotate` client alias.

thymikee:
Re-reviewed rebased head `69bff6081`: clean and merge-ready. The prior findings are resolved: batch CLI delegates to canonical structured-step readers and closes the validation gap; aggregate-perf retirement has one contracts-owned predicate/message across CLI and daemon; the daemon request is a real frames/memory discriminated union with no cast; retired aliases are data-driven; and the ineffective tests were replaced. The single commit is based directly on current `main`, preserves the merged façade/audio removals, and all current CI—including four smoke lanes, coverage, integration, package, guards, and docs—is green. Size is net-negative.

PR #2050, opened 2026-08-26
refactor(apple): colocate the XCUITest runner client into packages/platform-apple (#2040)

thymikee:
## Summary

Closes #2040 — the post-#2038 extraction track. Moves `src/platforms/apple/core/runner/` (the Apple XCUITest runner client: artifact cache, build/prewarm, request transport, session lifecycle — 34 modules, plus the runner-only `apple-runner-platform.ts` and 30 dedicated test suites) into `packages/platform-apple/src/runner/`. Root keeps composition wiring only.

## Recorded decision: a durable platform-owned mechanics facet, R13 amended deliberately

The first cut of this PR put the runner into a dedicated `@agent-device/apple-runner` package to stay inside R13's letter (only `src/platform-runtime.ts` may import a concrete platform package — impossible for the runner's 19 daemon/root consumers). Review direction (thymikee): Apple mechanics belong to the Apple package; when the rules block the honest home, amend the rules deliberately rather than route around them. So the runner is colocated as a **durable platform-owned implementation facet** — not a temporary exception — and R13 models it by enumeration instead of by a parallel package:

- `packages/platform-apple` exports its root façade **plus exactly** `./runner`, `./runner/client`, and `./runner/test-host` — any further subpath fails R13 declarations and the R11 pin.
- `@agent-device/platform-apple/runner` (types, `withAppleRunnerProvider`, pure sequence builders, runner bundle ids — eager closure 13) is the facet's **consumer seam**: the daemon and root code that reach runner mechanics directly import it here. The bare `@agent-device/platform-apple` façade stays composition-only and implementation-lazy (closure 1, unchanged).
- `./runner/client` (the host-bound factory, closure 46) has exactly one importer: the composition root `src/platforms/apple/core/runner-client.ts`. `./runner/test-host` has exactly one importer: the vitest setup. Both pinned by R13 with planted-red tests.
- The runner subtree is exempt from the ambient-host source rules (it owns its cache/lease files and usbmux sockets); **raw process primitives stay banned** in production sources (tests fake them; the host port names `ChildProcess` type-only) — process execution enters only through the injected `AppleRunnerHost` port (exec, diagnostics, retry, probes, locks, foreground Apple tooling, physical-device control — the port also breaks the old `physical-device-control ⇄ runner` cycle).
- Platform packages may now import `@agent-device/xml` (dependency-free vocabulary; the runner's usbmux plist parsing).

[...]

PR #2051, opened 2026-08-26
perf(harmonyos): memoize gesture viewport dumps

thymikee:
## Summary

HarmonyOS coordinate gestures now reuse one device-scoped ArkUI viewport for up to two seconds instead of paying a full layout dump for every adjacent gesture. Foreground-changing actions invalidate the memo, and keyboard dismiss delegates Back through the same invalidating owner as normal navigation.

The previous diagnostics and Apple changes were removed from this PR and preserved as separate ownership slices. This PR changes 7 files and stays within the HarmonyOS interaction family.

## Validation

A planted 20-second TTL made the new exact-expiry regression return the stale portrait viewport after two seconds. Before the production fix, the real interactor keyboard-dismiss route emitted Back but did not invalidate the memo. Both regressions are green on `c604df249e`.

The affected gate passed all runnable checks, including 247 test files / 1,550 tests, format, lint, typecheck, layering, Fallow, build, and provider integration. No HarmonyOS hardware is provisioned locally or in CI; residual live risk is limited to physical rotation or multi-window geometry changes within the two-second TTL.

### DIFF
--- a/src/core/__tests__/harmonyos-interactor.test.ts
+++ b/src/core/__tests__/harmonyos-interactor.test.ts
@@ -1,12 +1,19 @@
 import { expect, test, vi } from 'vitest';
 import type { DeviceInfo } from '@agent-device/kernel/device';
 
+const { invalidateHarmonyGestureViewport, runHarmonyHdc } = vi.hoisted(() => ({
+  invalidateHarmonyGestureViewport: vi.fn(),
+  runHarmonyHdc: vi.fn(async () => ({ stdout: '', stderr: '', exitCode: 0 })),
+}));
+
+vi.mock('../../platforms/harmonyos/hdc.ts', () => ({ runHarmonyHdc }));
 vi.mock('../../platforms/harmonyos/snapshot.ts', () => ({
   snapshotHarmony: vi.fn(async () => ({
     nodes: [{ index: 0, type: 'Button', label: 'Continue' }],
     truncated: false,
   })),
   readHarmonyGestureViewport: vi.fn(),
+  invalidateHarmonyGestureViewport,
 }));
 
 import { createHarmonyInteractor } from '../interactors/harmonyos.ts';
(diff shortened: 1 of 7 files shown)


thymikee:
Reviewed current head `6fba42957`: not ready. `src/utils/diagnostics.ts` adds a process-global, never-cleared `ensuredDiagnosticDirs` set. A long-lived daemon retains every distinct debug/trace directory forever; worse, if a previously seen directory is removed, later scopes skip `mkdirSync`, the append fails, and `emitDiagnostic` swallows it—silently losing diagnostics. Keep this state request/scope-local (or remove the cache) and add a two-scope delete/recreate regression. The branch is also `DIRTY`/conflicting against current `main`, and only CodeQL ran on this head, so rebase and rerun full CI. HarmonyOS hardware remains an explicit residual evidence gap.

thymikee:
Rebased onto current main and reconciled with #2050's package-owned Apple runner. The diagnostics directory memo is now scope-owned, including a delete/recreate regression; the stale-bundle test now lives under packages/platform-apple. Full local affected gate passed on 9a58fc4544d6accc71bf636cb652f2bf16b9418f.

thymikee:
Two P2 readiness blockers remain:

1. Add planted-red TTL-expiry coverage for the Harmony viewport cache. The new tests prove reuse, explicit invalidation, and device scoping, but never cross the configured 2,000ms TTL. A wrong or non-expiring TTL would stay green. Use a fake clock/timer to cache a viewport, cross 2s without an invalidation hook, and prove the next lookup performs a fresh dump.

2. Split this four-family leftovers batch into independently reviewable ownership slices: diagnostics scope/redaction, package-Apple stale-bundle cleanup, iOS boot memo reuse, and Harmony viewport caching share no owning interface or validation/evidence path. The #2050 rebase explains the Apple test location but not a runtime dependency. This conflicts with the repository rule to keep changes within one command family/module group unless the task explicitly crosses a boundary.

[P1] Invalidate the Harmony viewport memo when pressHarmonyKeyboardKey receives Back. The Harmony interactor implements keyboard dismiss through pressHarmonyKeyboardKey(device, 'Back'), but this path emits HDC Back without the invalidation performed by backHarmony. If Back navigates (or no keyboard consumes it), the next gesture can use stale geometry for 2s. Add a regression through the keyboard-dismiss/interactor route.

The individual implementations otherwise look sound. Exact-head CI is still incomplete, and Harmony hardware evidence remains unavailable.

thymikee:
Re-review summary at `c604df249ee51daca15b238aa11607c90e864cc3`:

- Fixed the HarmonyOS keyboard-dismiss finding by routing Back through the existing invalidating owner.
- Added exact 2,000 ms TTL expiry proof; both regressions were observed red before their fixes.
- Rebased onto current `main` and reduced this PR to the HarmonyOS owner slice.
- Preserved the removed work as independent current-main PRs: diagnostics #2056, Apple stale-bundle cleanup #2058, Apple boot-memo reuse #2060.
- `pnpm check:affected --run` passed all runnable checks (247 files / 1,550 tests).

No remaining code-review finding is known on this head. CI is still authoritative for the final merge verdict.

PR #2044, opened 2026-08-26
refactor(android): extract the adb executor and IME cluster into packages/platform-android (#2041)

thymikee:
Closes #2041. Stacked on #2038 (the branch this extraction builds on); merge #2038 first.

## What moved where

The Android adb-executor + IME cluster leaves `src/platforms/android` for its package, following the #2005/#2014 ownership-record shape:

- **`packages/platform-android`** now owns the implementations: `adb-executor.ts` (provider scope, executor/provider resolution, port-reverse manager, transfer funnels, failure-hint enrichment — consolidated onto the package's existing `adb-failure.ts` matcher table, deleting the duplicated table), `ime-lifecycle.ts` (activation/restore state machine, orphan startup recovery), and `ime-helper.ts` (artifact selection, manifest parsing, install cache, broadcast text channel).
- **`@agent-device/contracts`** gains two R11-registered subpaths for the vocabulary both sides need: `android-touch-plan` (the moved `touch-plan.ts`, pure) and `android-helper-artifacts` (IME + snapshot helper manifest/artifact types, the helper-install decision types, and the shared manifest field validators).
- **`@agent-device/kernel`** gains `keyed-lock` (`withKeyedLock`, moved from `src/utils`); `src/utils/keyed-lock.ts` re-exports it.
- **Root keeps composition wiring only**: `src/platforms/android/adb-host-binding.ts` implements and binds the package's new `adb-host` port, and the three old paths remain as thin re-export shims trimmed to the names root still consumes, each carrying the TODO naming the perf/trace dependency.

## The host port (R13)

R13 bars platform packages from raw process/fs/ambient authority, so the cluster reaches those primitives only through an explicitly injected `AndroidAdbHost` port (`packages/platform-android/src/adb-host.ts`): serial/host adb exec and spawn, command-executor override install/escape, result coercion + failure details, diagnostics, the IME recovery-marker store, and bundled-helper artifact resolution/install. The root binding delegates each facet to the existing owners (`utils/exec.ts`, `utils/diagnostics.ts`, `ime-recovery-marker.ts`, `helper-package-install.ts`) — no exec logic was reimplemented, and existing `vi.mock('utils/exec')`-style seams keep intercepting. An unbound port throws an actionable error; `listAndroidAdbSerialsQuick` resolves the host outside its try so a wiring bug stays loud instead of reading as "no devices".

## The perf/trace guardrail

[...]

PR #2052, opened 2026-08-26
refactor(android): one injectable host-adb transport for the remaining raw adb sites

thymikee:
## Summary

Moves the remaining host-level adb invocations behind one package-owned injectable transport after the Android package extraction.

- `packages/platform-android/src/adb-host.ts` owns scoped host transport and local host-port fallback.
- Local execution requests an inspectable result, coerces it once, then applies the shared classified failure contract.
- Quick device listing, emulator boot polling, and the Limrun host path route through the same seam.
- `listAndroidAdbSerialsQuick` now resolves the required host outside its recoverable adb-execution block, so missing composition remains loud while an actual adb failure remains best-effort.

This branch is rebased directly onto current `main`. Relative to `main`, it changes 11 files. Docs and skills are unchanged because no user-facing command surface changed.

## Validation

- Planted-red regression: an unbound host previously resolved to `[]`; it now rejects, while a bound adb execution failure still returns `[]`.
- Focused host transport, IME lifecycle, failure-contract, process-boundary, and eager-closure suites passed.
- `pnpm check:affected --run`: static/package gates and the broad related suite passed at `5b8bb2d9ee0b73807dc97bc694641c5c3e438730`. Two unrelated timeout-shaped tests were isolated; the iOS provider timeout reproduced on current `main` under the same host conditions.
- CI is authoritative for native lanes. The first Coverage Report failure was an unexpected Vitest fuzz-worker exit after coverage passed; the failed job has been rerun.


thymikee:
Reviewed current head `676d4bfcb`: not ready due to a failure-contract bug. `runAndroidHostAdb` promises the same classified nonzero failure for installed transports and local adb, but its local arm calls `runCmd('adb', args, options)` without forcing `allowFailure`; a real nonzero local process throws before the shared `androidAdbResultError` branch runs. The result lacks `adbFailure`, `retriable`, and the classified hint, while the transport arm includes them. The fallback test always uses `allowFailure: true` with a mocked executor, so it cannot catch the divergence. Obtain the local result under `allowFailure: true`, then apply the shared post-result throw according to the caller’s original option, and add regression coverage reaching the local spawn/nonzero boundary rather than a result-returning override. Current CI and live success-path Android evidence are green, but they do not cover this failure path.

thymikee:
Rebased and reconciled with #2044's packages/platform-android extraction. The host-adb scope now belongs to the injected package host seam, root consumers stay behind composition, eager-closure budgets remain unchanged, and a real local nonzero process regression proves shared failure classification. Full local affected gate passed on 0d63b251e4b044560ddf7af381228ff40452f877. Merge after #2044.

thymikee:
[P1] Keep an unbound Android adb host port loud in `listAndroidAdbSerialsQuick`. #2044 deliberately resolved `requireAndroidAdbHost()` before the recovery `try`, so broken composition could not masquerade as an empty device list. This head awaits `runAndroidHostAdb()` inside the `try`; its binding failure is therefore caught and returned as `[]`, making discovery/doctor report no devices instead of exposing invalid wiring. Resolve the bound host outside the recovery boundary (or preserve an equivalent typed distinction), then add a regression that an unbound port rejects while an adb execution failure still returns `[]`.

thymikee:
#2044 has now merged at 67ec6c3936b713fb79e77807aac63bcac178f0c8, and GitHub reports this head DIRTY against the new main. Rebase onto current main before the next review; keep the posted P1 unbound-host regression fix in that rewrite. No branch changes made by the coordinator.

thymikee:
Re-review summary at `5b8bb2d9ee0b73807dc97bc694641c5c3e438730`:

- Rebased the owned Android host-adb commits directly onto current `main`.
- Fixed the P1: host composition now resolves outside the recoverable adb execution block, so an unbound host rejects while a real adb failure still degrades to `[]`. The regression was observed red first.
- Focused transport/IME/failure suites passed; affected static/package and broad related checks passed.
- The first CI Coverage Report red was an unexpected Vitest fuzz-worker exit after coverage passed, not a code assertion; its failed job has been rerun.

No remaining code-review finding is known on this head; merge readiness depends on the rerun and queued native smoke lanes.

PR #2054, opened 2026-08-26
refactor(runtime): close residue execution units

### LINKED ISSUE #2042: Flip the residue legacy descriptors: web, replay, test, daemon

thymikee:
## Summary

- Close the four Wave 6 residue units identified by #2042: `daemon` and `web` now declare honest `host` execution through separate typed neutral services; `replay` and `test` delegate their last viewport leg through admitted `runtime gesture-viewport` and declare `none`.
- Keep gesture viewport admission and input projection in the gesture-family owner, route Maestro through `MaestroPublicOperation`, remove the retired dispatch barrel, and replace substring policy checks with the existing structural gates.
- Preserve the parent replay diagnostic `logPath` through the delegated viewport request so Apple runner failures retain their log artifact context.
- Rebase over #2061, which independently migrated `perf`; the exact current invariant is therefore zero `platformExecution: legacy` descriptors.

Closes #2042.

## Validation

- `pnpm check:affected --run`: all 56 selected runnable checks passed on the rebased tree; related Vitest coverage was 481 files / 3,891 tests.
- P1 planted red: the new viewport assertion failed against the pre-fix wiring with `execution.logPath: ""`; restored forwarding passes all 5 runtime-command tests.
- `pnpm check:layering`: 189/189 plus the layering guard.
- Eager-closure ratchet: 293/293; latest-main `capabilities.ts` gain retained at 73 modules.
- `pnpm check:quick`: lint and typecheck green.
- GitHub remains authoritative for coverage, native builds, and parked live replay lanes. Android live replay remains the previously reported residual evidence gap.

The build completes while printing the existing TS2883 declaration-portability diagnostics in the Apple runner-client shim.


thymikee:
## Thermo-nuclear code-quality review

The residue accounting is right and the two host reclassifications follow the existing `platform-runtime-*` composition idiom exactly (lazy `await import`, neutral contract type, frozen object) — that half I have no structural objection to. File sizes go down, not up.

The replay/viewport leg is where I want changes. It reaches the right *end state* (`replay`/`test` own no platform execution) by adding a bespoke parallel path next to two canonical seams that already do the job, and it steps around a table invariant that two cutover rows deliberately assert.

---

### 1. `readGestureViewport` becomes a second lexical owner of `gestureViewport`, and both rows that pin it stay silent

`scripts/layering/runtime-command-cutover-table.ts` declares, from **two** rows — R52 (line 1029) and R54 (line 1079):

```ts
operationOwners: { …, gestureViewport: ['selectGestureFrame'] }
```

under the comment *"One lexical owner per tier."* This PR adds a second one at `src/daemon/handlers/session-runtime-command.ts:113`:

```ts
const viewport = await admission.runtime.operations.gestureViewport({ … });
```

and pins it in neither row. The gate cannot see it by construction: `countOwnedOperationCalls` (`runtime-command-cutover-policy.ts:560`) only counts `operations.X(…)` calls found *inside* the declared owner functions, so a call outside every declared owner contributes zero and the "expected one, found 1" assertion still passes. The row this PR extends — R31 `runtime` — uses `singularExecution: { routeProof: … }` only, which proves *admission per function*, not call-site singularity.

So after this PR `gestureViewport` is the one operation in the table with an unpinned execution site, and the invariant those two rows spell out is false while reading green. That is worse than the legacy helper it replaces, which at least had an AST gate pointed at it.

### 2. The new handler re-inlines `gestureViewportInput` verbatim, in the wrong module

`src/daemon/gesture-runtime.ts:179`:

```ts
function gestureViewportInput(context: DaemonCommandContext) {
  return {
    ...(context.appBundleId === undefined ? {} : { options: { appBundleId: context.appBundleId } }),
    execution: runtimeExecutionFromContext(context),
  };
}
```

`src/daemon/handlers/session-runtime-command.ts:113-118` is that expression again, character for character. And it lands in the runtime-*hints* handler — the module that owns `set`/`show`/`clear` session policy — which then had to grow a `logPath` param threaded down from `session.ts:192` purely to build a context it should not be building.

**(1) and (2) have one fix.** `src/daemon/gesture-runtime.ts` is the gesture family's canonical owner module, and the codebase already has the pattern for exactly this in `home-runtime.ts` / `back-runtime.ts`: `resolveBoundXRuntime(...)` + a private `executeX`, with the input projection kept private to the family. Put the `runtime gesture-viewport` execution there, route the handler action through it, and `selectGestureFrame` stays the single lexical owner, both table rows stay true, the duplicated projection disappears, and `logPath` never has to be drilled.

### 3. The Maestro round-trip bypasses `MaestroPublicOperation` — the seam that already does this

`src/daemon/adapters/maestro/` has a typed operation → daemon-request seam: `MaestroPublicOperation` + `projectMaestroPublicOperation` + `invokeMaestroPublicOperation`, and that last one *already* is "build the request, `throw daemonResponseError(response)` on `!ok`, return `data`". Every other Maestro leg goes through it — `snapshot`, `clickPoint`, `swipe`, `launchApp`.

This PR hand-writes a second copy of that in a handler file (`session-replay-maestro-runtime.ts:323`), and to do it has to un-privatize `daemonResponseError` (`daemon-runtime-port-support.ts:155`) and hand-roll an `isRect` guard.

Adding `| { kind: 'gestureViewport' }` to the union with one projection arm collapses all of it. Concretely it deletes `invokeRuntimeGestureViewport`, `isRect`, and the `daemonResponseError` export widening — and then also deletes `resolveGestureViewport` from `DirectMaestroRuntimeDependencies` entirely, because its only production implementation becomes "invoke a fixed request", which is what `daemon-runtime-port.ts`'s own `invoke` closure already is. A dependency whose single implementation is a constant request is not a dependency.

The bypass is already showing three symptoms:

- **The dependency signature is a lie in both halves.** `(context: MaestroRuntimeReadContext) => Promise<Rect | undefined>` — the new impl is `async () => …`, so `context` is silently dropped, and it can never return `undefined` because it throws instead.
- **Two guards, one invariant, two messages.** `daemon-runtime-port.ts:101` still throws `'Unable to resolve Maestro gesture viewport.'` on a falsy viewport. That line is now unreachable: the inner helper threw first with a different message.
- **`isRect` buys nothing at runtime.** `isPositiveFiniteRect` already runs `Number.isFinite` over all four fields, which is false for a non-number. `isRect` exists only to get past `unknown`, which is the response-decoding problem the canonical seam owns.

While there: `isPositiveFiniteRect` also newly rejects zero-width/zero-height rects that the old direct path passed straight through. Small, probably an improvement, but it is an undeclared behavior change in a PR whose framing is structural.

### 4. `platform-execution-host-and-replay-routes.test.ts` — a new architecture rule as a substring grep, next to a sibling that does it properly

**`replayDelegationDefects` is mostly dead weight.** Two of its three checks are `source.includes('dispatchGestureViewport')`. This same PR adds `dispatchGestureViewport` to R52's `legacyRetirement.routeNames` (line 995), and `legacyRouteViolations` walks the **AST of every production source** (`runtime-command-cutover-policy.ts:73-77`) rejecting that identifier anywhere. The substring version is strictly weaker and now redundant.

The third check is `replaySource.includes("positionals: ['gesture-viewport']")` — a match against a formatted string literal. A prettier quote-style change or hoisting the positionals to a named const breaks it, and the behavior it is trying to pin is already asserted properly in `session-replay-runtime-maestro.test.ts`, which now checks the call log contains `['runtime', ['gesture-viewport']]`. I would delete the whole helper and its planted-red test.

**`hostRouteDefects` is new policy in the wrong place.** R3 `platforms-seam` explicitly `tolerates: ['dynamic']`, so "a `host` route may not name a `platforms/` specifier at all" is a genuinely new boundary rule — and it arrives as `routeSource.includes(prefix)` over two hardcoded paths, with no rule id and no hint (ADR 0010). Compare the sibling in the same directory, `platform-execution-cli-route.test.ts`: `oxc-parser`, a parent-aware walk, subjects enumerated by walking `src/cli/`, pairs read from a typed table, and an explicit statement that "a form not yet imagined" must fail. A comment mentioning `platforms/apple/` fails the new check; `await import('../../platforms' + '/apple/…')` passes it.

And: **what enumerates N?** `commandDescriptors.filter(d => d.platformExecution.kind === 'host')`. There are three host descriptors — `doctor` (registry.ts:576), `daemon`, `web` — and the test hardcodes two. The first test in this very file derives its list from the registry; the second one stops doing that. Either derive both, or move the rule into `scripts/layering/` where it gets a rule id, a hint, and the gate's output.

### 5. `src/core/dispatch.ts` is now a two-line barrel — finish the deletion

After this PR the file is:

```ts
export type { DispatchContext } from './dispatch-context.ts';
export { resolveTargetDevice, resolveTargetDeviceSelection } from './dispatch-resolve.ts';
```

Fourteen importers, all under `src/daemon/`, all wanting names that live in `dispatch-context.ts` or `dispatch-resolve.ts`. This is an internal module, not an entry surface, so it is exactly the pass-through indirection the repo's own re-export rule forbids. Repointing the fourteen imports is mechanical and it removes the barrel that several tests currently have to `vi.mock(..., importOriginal)` around — including two touched in this diff.

### 6. `host-platform-services.ts` is the grab bag the ADR amendment forswears

The ADR text this PR adds says: *"their domain services remain separate rather than forming a generic host grab bag."* The contracts module it adds then puts Apple runner-lease cleanup and the npm-managed browser backend in one file behind one export path. `packages/contracts/src/` is ~220 one-concept modules — `host-diagnostics.ts` gets its own. Split it into `daemon-owner-cleanup.ts` and `managed-web-backend.ts`; the export map already carries that granularity everywhere else.

---

### Smaller notes

- `src/platform-runtime-managed-web-backend.ts:3` says the service is "used by both CLI and web runtime paths". Its only consumer is `src/cli/commands/web.ts`.
- Both new modules export a frozen singleton `const`; all 28 siblings in the `platform-runtime-*` family export a `createX()` factory. The singleton is why `daemon.test.ts` had to move to module mocking rather than injection — worth matching the family.

---

Findings 1 and 3 are the blocking ones for me, and 1 + 2 collapse into a single move into `gesture-runtime.ts`. 4 and 5 are both net deletions. None of this changes where the PR is trying to land — only how much machinery it keeps on the way there.

thymikee:
Addressed the review in `47e8310ce2` (rebased onto current `main`, including merged #2044):

- moved gesture viewport admission/binding into `gesture-runtime`; `selectGestureFrame` is now the sole lexical operation owner, and the session handler no longer owns its input projection or drills `logPath`
- added `gestureViewport` to `MaestroPublicOperation` and decode it at the canonical invocation seam; removed the replay-only dependency/helper while preserving zero-sized `Rect` behavior
- removed the substring-based architecture assertions; descriptor truth remains here while existing AST/lifecycle/cutover gates enforce ownership
- deleted `src/core/dispatch.ts` and repointed imports/mocks to `dispatch-resolve.ts` or `dispatch-context.ts`
- split the host contracts into daemon-owner cleanup and managed-web-backend units, and changed both runtime roots to factories
- corrected the web ownership comment and removed stale live references to the deleted dispatch barrel

Validation before the rebase:
- focused review suite: 94/94
- `pnpm check:affected --run`: every check before the related Vitest suite passed; that suite finished 3670/3672 with two unrelated tests hitting their exact 5s timeout under the concurrent full run, and both passed in isolation (7/7)

Post-rebase validation:
- focused review and conflict-ratchet suite: 98/98
- layering suite and guard: 188/188
- eager-closure ratchet: 285/285
- `pnpm check:quick`: green

I also completed a final adversarial diff audit; no remaining blocker found in the review scope.

thymikee:
Not ready — one P1 production regression remains at `47e8310ce2a4a9eff4d119c59a18ed79421e920f`.

The new `runtime gesture-viewport` delegation drops the parent replay request’s diagnostic `logPath`. The `runtime` handler in `src/daemon/handlers/session.ts` does not forward its available `logPath`; `readGestureViewport` in `session-runtime-command.ts` then calls `contextFromFlags(, ...)`; that empty path reaches the Apple runner options. Before this PR the Maestro viewport route constructed context with the actual replay `logPath`. On an iOS percentage-coordinate action, a runner failure therefore loses its runner log artifact/enriched diagnostic context. The successful live replay and current tests cannot reveal this because they assert requestId/viewport values but not `execution.logPath`.

Pass `logPath` through `handleRuntimeCommand`/`readGestureViewport`, use it in `contextFromFlags`, and add a regression assertion on `gestureViewport` execution.logPath.

Everything else sampled is coherent and exact-head CI is green. Android live replay remains a residual evidence gap because both local emulators were legitimately owned.

thymikee:
Addressed the P1 in `37e4d8596a` and rebased onto current `origin/main`.

- `logPath` now flows through the `runtime` session route and `handleRuntimeCommand` into `contextFromFlags` for `gesture-viewport`; gesture input projection remains owned by `gesture-runtime`.
- Added an execution-level regression assertion. It was observed red against the pre-fix wiring (`execution.logPath` was `""`) and is green with the fix.
- Integrated #2061: `perf` is no longer legacy, so the exact descriptor assertion now requires zero legacy entries; removed the now-unused legacy constant and retained the latest-main closure decrease.

Validation: all 56 runnable checks selected by `pnpm check:affected --run` passed, including 481 related test files / 3,891 tests; layering is 189/189 plus guard, closure ratchets are 293/293, and lint/typecheck are green.

thymikee:
The prior P1 is fixed; no new code findings. `logPath` now flows from the session runtime handler through `handleRuntimeCommand`/`readGestureViewport` into `contextFromFlags`, and the focused regression exercises the session handler and asserts both requestId and logPath reach the bound gesture viewport operation. The planted-red claim is credible because the pre-fix route supplied a literal empty path.

All code, coverage, Android/Linux/macOS, size, guidance, and CodeQL checks are green. iOS Smoke remains red: `smoke:automation-input` stalled on its initial `wait text Agent Device Tester` with zero readable captures before the changed Maestro gesture-viewport route can execute. Current base passed iOS shortly beforehand, and the only post-base fix is logPath forwarding, so this is best classified as an unrelated iOS runner/host flake, not an owner-action code regression. Still, the required check is red, so `ready-for-human` remains off until a rerun succeeds. Android live replay remains residual evidence because the local emulators were legitimately owned.

PR #2055, opened 2026-08-26
fix(fuzz): run parser cases in a worker process, not the runner's thread (#2053)

### LINKED ISSUE #2053: Coverage job: fuzz-worker fork dies after unit-core completes (2 occurrences on PR #2035)
The `Coverage` job's second Vitest invocation — `pnpm test:fuzz-worker`, which runs the single-file `fuzz-worker` project — loses a fork with no test reporting a failure. Twice on PR #2035, on commits that cannot reach it.

## Signature

```
Error: [vitest-pool]: Worker forks emitted error.
Caused by: Error: Worker exited unexpectedly
```

It lands in the *second* invocation. `pnpm test:coverage:ci` is `vitest run --coverage --project=!fuzz-worker && pnpm test:fuzz-worker`; the first leg reports every file green immediately before the death.

| run | job | leg result before the error |
| --- | --- | --- |
| [32868994199](https://github.com/callstack/agent-device/actions/runs/32868994199/job/97871198238) | Coverage (1) | fuzz-worker: `Tests 8 passed (11)`, `Errors 1 error` |
| [32962380010](https://github.com/callstack/agent-device/actions/runs/32962380010/job/98157360652) | Coverage (2) | unit-core: `Test Files 568 passed (568)`, then fuzz-worker `Test Files (1)` |

Both re-ran green on the same commit, so it is not deterministic.

## Why this is probably not #1824

[#1824](https://github.com/callstack/agent-device/issues/1824) was closed as fixed by #1854: Apple runner tests fabricated child pids (4141/4242/4343/4444) and delivered real signals to sibling vitest forks, and `src/__tests__/hermetic-signal-setup.ts` now refuses that. That mechanism was in `unit-core`. These deaths are in the `fuzz-worker` leg, which runs one file (`scripts/fuzz/corpus-replay.test.ts`) with `fileParallelism: false`, `maxWorkers: 1` — a single worker, after the `unit-core` run has already finished. #1824's own thread called this shape "(B) tail deaths" and explicitly did not count it against that fix.

## Why it matters

The second occurrence is the load-bearing one: the commit under test (`a1cd050d52`) differed from a fully green head (`adcc35b8db`) **only in commit messages** — `git diff` between them is empty and the trees are byte-identical. So the failure is independent of the change, and with the contention retry removed (#1781 A4) there is no rerun layer in front of it. It reads as a red required check on an unrelated PR, which is the same class of noise #1874 was filed for.

## Suggested first step

`corpus-replay.test.ts` replays the fuzz corpus through the worker watchdog, waiting a real per-case budget. Worth reading the fork's exit code/signal — the log records neither, and that is the fact that would separate a watchdog-killed child from an OOM or an inherited signal. `poolOptions.forks.execArgv` or `NODE_OPTIONS=--trace-gc` on one Coverage run would settle it, the same cheap-confirmation shape #1824 used.

---

thymikee:
## Summary

Closes #2053. The `test:fuzz-worker` leg lost its Vitest worker six times between 2026-08-24 and 2026-08-26, on commits that cannot reach it. The cause is where the fuzz lane runs its cases, not what the cases contain: `corpus-replay.test.ts` executed adversarial parser input **on worker threads of the Vitest worker running the test file**. A worker thread is not a fault boundary — a fault in one ends its whole process — so a case that faulted killed the test runner, and Vitest reported a fork that vanished with no test, file, or case named.

Cases now run in a worker **process** (`scripts/fuzz/execute.ts`). Same protocol, same per-case budget, same restart-on-hang; what changes is that the lane can now see the second fault a case cannot report about itself. A case that never returns was already a `hang`; a case that ends the process it runs in is now a `crash` failure carrying the exit code or signal and the tail of that process's stderr — the death certificate this lane used to lose.

## Why this is the cause

**It is this file.** Before #1994 split it out, the vanished file in every Coverage-lane fork death was this one, six for six out of ~1100 (see the `FUZZ_WORKER_TESTS` comment in `vitest.config.ts`). The split moved the death rather than stopping it: the second leg, uninstrumented and one file long, then died six more times — runs [32773842455](https://github.com/callstack/agent-device/actions/runs/32773842455/job/97580206287), [32780182194](https://github.com/callstack/agent-device/actions/runs/32780182194/job/97601293556), [32780811227](https://github.com/callstack/agent-device/actions/runs/32780811227/job/97603133980), [32854636963](https://github.com/callstack/agent-device/actions/runs/32854636963/job/97825253392), plus the two in the issue. Of ~1100 unit files this is the only one that creates and terminates worker threads by the handful (8 per run) and feeds them hostile input; the same machinery in a plain process — the nightly lane, ~266 000 cases a night — has been green throughout.

**It is an uncatchable signal, and nothing else.** The log records neither exit code nor signal because Vitest discards both: `emitUnexpectedExit = () => {…}` takes no arguments. So instead of reading the death, I seeded each candidate death into this project and compared the run's output with the CI logs. Every mechanism Node can report *is* reported:

| seeded in the `fuzz-worker` project | what the run prints |
| --- | --- |
| `process.exit(1)` inside a test | `Test Files 1 failed`, pointing at the line |
| heap-limit OOM in the Vitest worker | `FATAL ERROR: … JavaScript heap out of memory` + native stack |
| `process.abort()` | `node::Abort` native stack |
| heap-limit OOM inside a worker thread | `ERR_WORKER_OUT_OF_MEMORY`, an ordinary test failure |
| **SIGKILL / SIGSEGV to the worker** | **nothing — `Test Files (1)`, `Tests N passed (11)`, `Errors 1 error`, `tests 0ms`** |

Only the last row matches the CI failures, which carry no `FATAL`, no native stack, and no failing test. That rules out the OOM reading #1824 left open for this shape, and rules out a silent `process.exit`. It leaves a fatal signal — and in a leg whose whole process tree is pnpm, Vitest, and one worker, with nothing spawned and 16 GB free, the plausible sender is the worker's own faulting.

What I have **not** got is the signal itself: reproducing needed Linux (0 failures in 40 solo runs, 25 runs under 10-way CPU contention, 2 100 pure-Node create/terminate cycles and 1 400 in-fork cycles on macOS, Node 24.13 and 26.1). So this names the class, not the instruction pointer. The fix does not depend on that distinction: whichever fault it is, the fuzz lane's own contract — attribute a case's misbehaviour to the case — cannot be met by a thread that shares its process with the reporter.

## What this does not change yet

The `fuzz-worker` project and the second `test:coverage:ci` leg stay. They are the mitigation, and the evidence that the source is closed accrues in CI, not here. `vitest.config.ts` carries the criterion: remove both — restoring the file to `unit-core` and the coverage script to one invocation — after 30 consecutive Coverage runs with no `Worker exited unexpectedly`, and reopen #2053 with the run URL if one appears first.

## Validation

**The new kind is reported, and reported by the same self-check every other kind is.** `pnpm fuzz:parsers --self-check`:

```
  [self-check-crash] crash: case killed the worker process (exit code 97)
ok   self-check-crash: expected crash, got crash
```

Seeding a real fatal signal and a native banner instead of the target's `process.exit`, to prove the classification is on the death and not on the exit code:

```
  [self-check-crash] crash: case killed the worker process (signal SIGSEGV)
  [self-check-crash] crash: case killed the worker process (exit code 3: FATAL ERROR: simulated native banner)
```

**One fix that came out of the restructuring.** The handshake and a case are the same wait — the worker answers, dies, or stays silent for its budget — so they share one primitive and differ only in what they make of those three outcomes. Reading the startup path that way exposed a leak the old code had: its 60 s startup budget rejected without ending the worker, so a worker stuck importing was abandoned still holding its channel. Every refusing path now ends the session (verified: refusal raised, zero leftover workers).

**Containment is asserted in-process, where it bites.** Two new cases in `harness.test.ts` call the runner directly rather than through the CLI — that test file runs in a Vitest worker, so it *is* the caller the lane must protect. Under the old runner they would have killed their own worker and reported nothing; they now assert the crash failure and that the runner keeps going afterwards.

**Cost.** Measured on one host, best of three, 7 targets × 38 000 cases (the nightly's shape): 17.8 s → 20.9 s. The unit-lane replay is unchanged within noise (~2 s). A transport-only microbenchmark puts the difference at ~6 µs per case; the rest is process startup, ~30 ms per target session.

**Suites.** `scripts/fuzz` unit tests 46/46, the `fuzz-worker` project 11/11, `test/ci` 9/9, typecheck, `oxlint --deny-warnings`, `oxfmt --check`, `fallow audit`, and `check:gate-manifest` all clean.

---

### DIFF
--- a/.github/workflows/replays-nightly.yml
+++ b/.github/workflows/replays-nightly.yml
@@ -38,7 +38,10 @@ jobs:
   # 38,000 cases/target holds the job's wall-clock at the pre-B2 five-target/50k budget now that
   # seven targets share it: measured on a quiet host, 7x38k = 16.5s against 5x50k = 16.6-16.9s.
   # Chosen by measurement rather than rounding — 33k would have been 2s cheaper but cost the
-  # five untouched targets a third of their depth for no wall-clock reason.
+  # five untouched targets a third of their depth for no wall-clock reason. The count is
+  # unchanged by #2053; moving case execution out of process cost that run about 3s (17.8s ->
+  # 20.9s, best of three on one host), which buys a case that faults its process a named
+  # `crash` failure instead of a dead caller.
   nightly-parser-fuzz:
     name: Parser Fuzz Lane
     runs-on: ubuntu-latest
--- a/docs/agents/testing.md
+++ b/docs/agents/testing.md
@@ -115,8 +115,11 @@ another example. Keep examples for a real past bug or a named decision. Reuse `P
 budgets so property files stay inside the unit slow-test gate.
 
 Parser fuzz targets live in `scripts/fuzz/targets.ts`. Validation generators carry the invalid
-outcome they planted, so silent acceptance and wrong error codes are failures. Promote a discovered
-case with the command the harness prints — never hand-copy an unshrunk input.
+outcome they planted, so silent acceptance and wrong error codes are failures. Cases run in a
+worker process, so the two faults a case cannot report about itself — never returning, and killing
+the process it runs in — are reported as `hang` and `crash` against the exact input rather than
+taking the caller down with them. Promote a discovered case with the command the harness prints —
+never hand-copy an unshrunk input.
 
 Mutation is report-only and limited to the registry in `scripts/mutation/modules.ts`. It measures
 whether tests distinguish changed decision logic. Do not infer redundancy from line coverage alone.
--- a/scripts/fuzz/corpus-replay.test.ts
+++ b/scripts/fuzz/corpus-replay.test.ts
@@ -2,8 +2,10 @@
 //
 // The nightly lane finds cases; this replays every case it ever found so a regression fails in
 // seconds on a PR instead of a night later. Cases go through the same worker-backed watchdog the
-// nightly lane uses: a promoted hang case must fail this test against its per-case budget, not
-// wedge the unit job until the CI timeout.
+// nightly lane uses, and for the same reason they run in a worker *process* (#2053): a promoted
+// hang case must fail this test against its per-case budget rather than wedge the unit job, and
+// a case that faults the process it runs in must fail this test rather than kill the Vitest
+// worker running it — which reports no test, no file, and no case.
 
 import fc from 'fast-check';
 import { describe, expect, it } from 'vitest';
--- a/scripts/fuzz/execute.ts
+++ b/scripts/fuzz/execute.ts
@@ -1,22 +1,32 @@
-// Case execution with hang detection for the parser fuzz lane (#1414).
+// Case execution with hang and crash detection for the parser fuzz lane (#1414).
 //
-// Cases run one at a time in a worker thread: a synchronous parser that never returns cannot be
-// timed out from inside its own tick, so the budget is enforced from *another* thread, which can
-// terminate the wedged one and attribute the stall to the exact input. Cases go over the wire
-// individually (rather than as one batch) because fast-check drives the loop — it decides the next
-// input, including the shrink candidates it derives from a failing one.
+// Cases run one at a time in a child process, and the budget is enforced from the parent. Two
+// faults a case can produce cannot be observed from inside the context running it: a synchronous
+// parser that never returns cannot be timed out from its own tick, and a parser that faults the
+// process — a native crash, an out-of-memory abort — leaves no stack to catch. Both are
+// attributed here, to the exact input, by watching a process the parent can outlive.
+//
+// It is a process rather than a worker thread because a thread cannot contain the second fault:
+// a fault in a worker thread takes its whole process down, and that process is the caller's —
+// for the unit-lane replay, the Vitest worker running the test file, which then dies with no
+// test, file, or case named (#2053). Cases go over the wire individually (rather than as one
+// batch) because fast-check drives the loop — it decides the next input, including the shrink
+// candidates it derives from a failing one.
 
+import { fork, type ChildProcess } from 'node:child_process';
 import path from 'node:path';
 import { fileURLToPath } from 'node:url';
-import { Worker } from 'node:worker_threads';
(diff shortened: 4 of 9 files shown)


thymikee:
Code review is clean at `c0c5ca7bec70bae6c38b6c7018ba855215ae2809`; no actionable code findings. The process boundary owns both watchdog and child death, waits for piped stderr to drain before recording a crash, attributes the exact case, and starts a fresh session after hang/crash. The production route remains intact, and the direct harness tests prove crash containment/attribution. Core checks, coverage, integration, size, and CodeQL are green; package size decreases.

Not labeling `ready-for-human` yet because the exact-head iOS and Android smoke jobs are red. Their failures appear unrelated to this device-free fuzz diff (Android transient ADB-offline/alert wait; iOS malformed fixture text), but policy keeps the label off while confirmed CI failures remain. The two-crash test could more directly use crash-then-success to prove the replacement worker executes a clean case, but that is not a blocker.

thymikee:
Updated-head review is clean; no actionable code findings. The delta centralizes startup/case waits in `awaitWorker` and adds a permanent child error listener, closing the unhandled ChildProcess-error hole. Startup refusal/silence now tears the child down before throwing; case waits are armed before send and preserve death/hang classification, safe restart, and the pipe-drained death certificate. The target reorder and comment compaction are inert.

The earlier completed core-check evidence belongs to the prior head. Only CodeQL is currently running on this head, so `ready-for-human` remains off until fresh normal CI reports.

thymikee:
Ready for human review. The exact head is unchanged from the clean updated-head review, is mergeable, and every fresh normal check is green: guidance, all CI/coverage/integration/package lanes, Android/iOS/Linux/macOS smoke, size, and CodeQL. The skipped cache-matrix row is expected. No device evidence is owed by this device-free fuzz change. The optional crash-then-success test refinement remains non-blocking.

PR #2057, opened 2026-08-26
fix(daemon): reconcile device claims held by a replaced daemon

### LINKED ISSUE #2031: web: a stale device lease outlives its session — DEVICE_IN_USE names a session `session list` does not report

thymikee:
## Summary

A daemon that is replaced but keeps running keeps the device claim it took. Its sessions are reachable
only through the `daemon.json` its state dir publishes, and that record now names the successor — so the
claim named a session no `session list` could report and no `close` could release. The only documented
recovery was to find and kill the orphaned daemon by hand.

Claim classification now proves that case. A live owner that is **not** the daemon published for its own
state dir classifies as `owner-daemon-superseded`, and reconciliation settles it exactly like a dead
owner: through the same exact-owner durable-resource recovery, which still retains the claim when cleanup
is pending.

Before / after, on the same staged state (a live daemon holding the web claim after losing its
registration):

```
$ agent-device session list --json
{ "success": true, "data": { "sessions": [] } }

# before
$ agent-device open "about:blank" --platform web
Error (DEVICE_IN_USE): web device agent-browser-chrome is owned by session "cwd:2489af30c8e017d8:default" …
$ agent-device device status --platform web --serial agent-browser-chrome
web Agent Browser Chrome: live session=cwd:2489af30c8e017d8:default …

# after
$ agent-device device status --platform web --serial agent-browser-chrome --stale
web Agent Browser Chrome: owner-daemon-superseded session=cwd:2489af30c8e017d8:default …
$ agent-device open "about:blank" --platform web
Opened: about:blank
```

Proof runs one way only, so the change cannot hand a held device to a second owner:

- an absent, unreadable, or pid-less registration is **not** proof — a daemon publishes its record after
  startup recovery, so absence equally describes one about to be reachable;
- equal pids prove supersession only when both start times are readable and differ;
- the process reading a registration never supersedes itself — a caller had to reach it to ask, so a
  daemon can never disown its own live sessions;
- the unattended daemon-startup sweep settles a still-running owner only inside the state dir it now
  serves. A superseded owner in another state dir is left to the acquisition path, which a caller drove
  at that exact device.

Shapes and comparisons stay canonical rather than bespoke. `daemon.json`'s identity fields had two
readers; the daemon-side one moves to `daemon-registration.ts` — returning the existing `OwnerIdentity`
rather than a new type — and `daemon-stop.ts` now shares it, so this adds a use instead of a third
parser. The daemon-client reader stays put: `daemon-server` may not import `daemon-client` (R5). The
comparison itself is a pure `ownerIdentityDiffers` sitting next to its mirror `ownerIdentityMatches`,
which leaves `isSupersededDaemonOwner` as three lines whose only real contribution is the I/O.

Closes #2031

## Validation

Reproduced and fixed live on macOS against the managed `agent-browser` backend in an isolated state dir,
with a real orphaned daemon (alive, holding `local:web:none:agent-browser-chrome`, registration removed).
The pre-fix build produced the reported `DEVICE_IN_USE` verbatim while `session list` returned
`{"sessions": []}`; the post-fix build on the identical staged state classified the claim
`owner-daemon-superseded` and reconciled it on the next `open`. Evidence gathered at `3fdbc7b`. The
verification session was closed and both orphaned daemons stopped.

Three regression tests were each observed red against the pre-fix code and cover a different seam:
acquisition (`device-claims`), the daemon-startup sweep (`device-claim-prune`), and the CLI status
surface (`cli-device-status`). Four fail-closed guards — no registration, matching registration,
self-ownership, and a superseded owner in another state dir — pass in both states, which is what they are
for; each was separately confirmed to fail when its guard is removed.

Full `unit-core` bundle green on an idle host: 1048 files, 8053 tests. An earlier run under concurrent
load tripped the slow-test budget on `gesture-admission-parity`, which is already over budget on
`origin/main` (2.89s against 2.5s, same file, same host) and stubs `admitDeviceClaim` with a no-op, so no
line of this diff executes in it — host contention, per the AGENTS.md rule. Also green: `typecheck`,
`oxlint`, `oxfmt --check`, `check:layering`, `check:fallow --base origin/main`, `check:mcp-metadata`.
`check:production-exports` reports the same 50 pre-existing findings on `origin/main` as on this
branch.

`check:affected` could not run in this worktree — it refuses on a `node_modules` / `pnpm-lock.yaml`
mismatch that predates this branch. The gates above were run directly instead; CI on the head is the
authority.

## Notes

13 files: 7 production and 6 test. Scope stayed inside the device-claim module group;
`owner-identity.ts` gains one pure helper, `daemon-runtime.ts` changes only to thread its state dir into
the startup sweep, and `daemon-stop.ts` only to drop its private copy of the registration reader.

No docs or skills changed. No user doc enumerates claim classifications, and `device status`'s help
already says reclamation happens "during open or daemon startup after exact-owner resource
reconciliation" — this widens when that applies without changing the sentence.

Residual risk, unchanged in kind by this PR but worth naming: durable-resource recovery
(`readRecoveryCandidate`) has no owner-liveness guard, so reconciling a superseded owner that is still
running can finalize a capture that owner is writing. The startup sweep is scoped to the daemon's own
state dir to bound this, and daemon startup already reaps that state dir's orphaned web browsers and
app-log resources on the same reasoning. Adding an owner-liveness gate to durable recovery is a
worthwhile follow-up, but belongs to that seam rather than this one.

### DIFF
--- a/src/__tests__/cli-device-status.test.ts
+++ b/src/__tests__/cli-device-status.test.ts
@@ -4,6 +4,7 @@ import path from 'node:path';
 import { test, vi } from 'vitest';
 import { readCurrentOwnerIdentity } from '../utils/owner-identity.ts';
 import { runCliCapture } from './cli-capture.ts';
+import { publishDaemonRegistration } from './test-utils/device-claim-store.ts';
 import { mkdtempForTestSync } from './test-utils/tmp-dir.ts';
 
 vi.mock('../utils/host-process.ts', async (importOriginal) =>
@@ -189,6 +190,53 @@ test('states that nothing is claimed when every claim is stale', async () => {
     assert.match(result.stdout, /No live local device claims found\./);
     assert.match(result.stdout, /1 stale claim hidden/);
     assert.doesNotMatch(result.stdout, /Dead iPhone/);
+  } finally {
+    fs.rmSync(claimsDir, { recursive: true, force: true });
+  }
+});
+
+test('names a replaced-but-running daemon owner as stale rather than a live holder', async () => {
+  const claimsDir = mkdtempForTestSync('agent-device-cli-claims-');
+  const stateDir = path.join(claimsDir, 'state');
+  try {
+    // #2031: the recorded owner still runs, but we are the daemon published for
+    // its state dir, so its session is one `session list` cannot report.
+    publishDaemonRegistration(stateDir, readCurrentOwnerIdentity());
+    fs.writeFileSync(
+      path.join(claimsDir, 'superseded.json'),
+      JSON.stringify({
+        schemaVersion: 1,
+        deviceKey: 'local:web:none:agent-browser-chrome',
+        device: {
+          platform: 'web',
+          id: 'agent-browser-chrome',
+          name: 'Chrome',
+          kind: 'device',
+        },
+        session: 'cwd:/w:default',
+        workspace: '/w',
+        stateDir,
+        ownerPid: process.ppid,
+        ownerStartTime: null,
+        ownerToken: 'superseded-token',
+        createdAtMs: 1,
+        updatedAtMs: 1,
+      }),
+    );
+
+    const normal = await runCliCapture(['device', 'status'], {
+      env: { AGENT_DEVICE_CLAIMS_DIR: claimsDir },
+    });
+    assert.match(normal.stdout, /No live local device claims found\./);
+    assert.match(normal.stdout, /1 stale claim hidden/);
+
+    const stale = await runCliCapture(['device', 'status', '--stale', '--json'], {
+      env: { AGENT_DEVICE_CLAIMS_DIR: claimsDir },
+    });
+    const payload = JSON.parse(stale.stdout);
+    assert.equal(payload.data.claims.length, 1);
+    assert.equal(payload.data.claims[0].classification, 'owner-daemon-superseded');
+    assert.equal(payload.data.claims[0].owner.session, 'cwd:/w:default');
   } finally {
     fs.rmSync(claimsDir, { recursive: true, force: true });
   }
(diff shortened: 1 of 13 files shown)


thymikee:
Clean code review at `fab7bd156db5c1bd8e2d62771f5c40d95d96f543` — no actionable findings. The acquisition and startup routes use the same fail-closed superseded-owner classification; replacement is accepted only with positive different-owner proof, while missing/corrupt/pid-less registration, unreadable matching-PID start time, and self-inspection remain live. Recovery stays exact-owner and lock-protected, and the startup sweep limits still-running superseded owners to its own state directory.

Validation: `pnpm check:quick` passed; 38 focused tests passed; layering 188/188 passed; fallow passed. A live web open exercised the current claim route and correctly refused a legitimately held device claim without leaving a session.

Residuals: the deliberately orphaned/superseded-daemon scenario was not replayed because doing so would disturb a live daemon; the author's isolated reproduction remains the practical evidence. CI jobs are cancelled/queued from the Actions outage, with no confirmed owner-action failure.

PR #2061, opened 2026-08-26
refactor: migrate perf to device runtime

thymikee:
## Summary

Migrate `perf` from legacy/plugin routing to exact request-bound device runtime ownership.

- add typed perf facts, plans, complete per-family hosts, and a durable native-capture resource
- keep platform policy and operation construction in the Apple, Android, and HarmonyOS packages while root composition supplies lazy native mechanics
- preserve recording semantics and reject unsupported flag placement with migration-friendly errors
- keep standalone `trace` unchanged and retain HarmonyOS's existing partial memory behavior without new Harmony-specific mechanics
- add the R64 single-path cutover proof and reconcile capability, coverage, loading-shape, and package-boundary gates

## Validation

- `pnpm check:affected --run` (all relevant checks green after focused follow-up; GitHub-authoritative lanes skipped locally)
- `pnpm test:coverage` (1,149 files, 8,609 tests)
- `pnpm check:coverage-changed --base origin/main` (286/408, 70.10%; threshold 70%)
- `pnpm check:layering`
- `pnpm check:fallow --base origin/main`
- focused perf grammar, platform policy, durable lifecycle, recording, and teardown tests

Claude Fable was invoked as requested, but did not complete its multi-finder review within the bounded window and returned no findings before termination.


thymikee:
## Thermo-nuclear code-quality review — `refactor: migrate perf to device runtime`

**Verdict: request changes.** The daemon side of this migration is genuinely good, but the platform side re-creates the branch it was supposed to delete, three wrapper layers are doing one layer's work, and ~870 lines of new production code landed with zero direct tests while ~1,470 lines of behavioural tests were deleted. Two user-visible behaviours changed silently.

### What is clearly right

- Net −893 lines, and the eager-closure budgets move in the right direction hard: `core/dispatch.ts` 79→15, `register-builtins.ts` 72→6, `session-teardown.ts` 99→60, `platforms` type-cycle zone 1→0.
- `applePerf` + `nativePerf` (two platform-shaped `SessionState` fields) collapse into one neutral durable record. That is the R7 cleanup this wave was supposed to pay for.
- Three positional parsers (`session-perf.ts`, `session-perf-xctrace.ts`, `session-native-perf.ts`) become one.
- The `perf-capture-*` daemon quartet is a faithful copy of the `audio-probe-*` template — right template, right place, no invention.

---

## 1. Blocking: the platform switch moved down a layer, it did not disappear

`src/platform-runtime-perf-host-impl.ts:26-29`:

```ts
if (device.platform === 'apple') return applePerfOperations(device, owner);
if (device.platform === 'android') return androidPerfOperations(device, owner);
if (device.platform === 'harmonyos') return harmonyPerfOperations(device);
return Object.freeze({});
```

Before this PR the daemon branched on `device.platform` to pick a perf collector. After it, the daemon is clean and *the host* branches on `device.platform` to pick a perf collector — in root `src/`, 679 new lines of it. The three platform packages contribute nothing to perf except facts: their `bind` spreads in a forwarder that will re-dispatch on `device.platform` at call time. `harmonyPerfOperations`' "Dropped-frame sampling is currently available only on Android app sessions…" string now lives in root `src/`, while the Harmony fact that gates it lives in `packages/platform-harmonyos`.

This repo already has the right shape for exactly this problem. `ScreenRecordingRuntimeHost` (`packages/contracts/src/screen-recording-runtime-host.ts:208`) exposes **per-family sub-hosts**:

```ts
export type ScreenRecordingRuntimeHost = Readonly<{
  apple: AppleScreenRecordingHost;
  android: AndroidScreenRecordingHost;
  harmony: HarmonyScreenRecordingHost;
  web: WebScreenRecordingHost;
  …
}>;
```

…and `packages/platform-apple/src/recording/runtime.ts` builds the Apple recording operations *in the Apple package* out of `host.screenRecording.apple.*` primitives. There is no `device.platform === …` anywhere in that host, because the caller already knows its platform. The collectors still live in root `src/platforms/`, reached through the sub-host — so the R3 seam is not the reason perf had to go the other way.

Please reshape `PerfRuntimeHost` the same way:

```ts
export type PerfRuntimeHost = Readonly<{
  apple: ApplePerfHost;      // frame/memory samplers, xctrace start/stop/report primitives
  android: AndroidPerfHost;
  harmony: HarmonyPerfHost;
}>;
```

That single change deletes, by construction:

- the `device.platform` switch and all three `*PerfOperations` bodies from root `src/`;
- `bindPerfRuntimeOperations` entirely (each package uses the ordinary `whenAdmitted(…)` idiom it already uses for every other family);
- `createPerfRuntimeHost`'s seven forwarders and `requireOperation` (see §2);
- the `Partial<PerfRuntimeOperations>` return type, the `as Readonly<Record<string, unknown>>` lookup, and both "omitted admitted operation" `TypeError`s — because the facts and the implementations end up in the same file and the compiler proves the pairing.

It also puts the perf bodies next to `applePerfFacts` / `androidTouchFact` / `harmonyFocusFact`, so the "is perf admitted here" and "what does perf do here" decisions stop being two files apart.

## 2. Blocking: `createPerfRuntimeHost` is a redundant layer even on its own terms

`src/platform-runtime-perf-host.ts:17-32` wraps all seven operations so each one can lazily import the impl and re-check for its own key. But `host.perf` is only ever passed to `bindPerfRuntimeOperations`, and the only call of `host.bind` is inside its `invoke` (`packages/contracts/src/perf-runtime-host.ts:37`), which already re-binds per operation call and already throws `Perf host omitted admitted operation ${key}` — a near-verbatim duplicate of `requireOperation`'s `Perf runtime host omitted admitted operation ${key}` at line 43.

So the whole file collapses to this, with identical laziness and identical failure behaviour:

```ts
export function createPerfRuntimeHost(): PerfRuntimeHost {
  return Object.freeze({
    bind: async (input) =>
      await (await import('./platform-runtime-perf-host-impl.ts'))
        .createPerfRuntimeHostImplementation()
        .bind(input),
  });
}
```

Thirty lines and a duplicated error path gone. This one stands whether or not §1 is taken.

## 3. Blocking: `bindPerfRuntimeOperations` is seven copies of one thought, joined by a cast

`packages/contracts/src/perf-runtime-host.ts:51-84` is the same six-line block repeated seven times with only the key changed, and the join between the fact and the implementation is a cast plus a runtime throw:

```ts
const operation = (operations as Readonly<Record<string, unknown>>)[key] as Operation<Key> | undefined;
if (!operation) throw new TypeError(`Perf host omitted admitted operation ${String(key)}`);
```

The fact says the operation is admitted; the host returns `Partial<…>` and may disagree. Two sources of truth for one decision, reconciled at runtime by a `TypeError` that is not an `AppError` and will surface to the user as an internal crash rather than a typed refusal. §1 removes the need for this file. If §1 is deferred, at minimum drive the seven keys off `perfRuntimeOperationFacts`' own key list rather than hand-repeating them.

## 4. Blocking: dead code kept alive by `void`

`src/core/capabilities.ts:22-25`:

```ts
// Retained as the empty tombstone named by the Wave 5 cutover rows: …
const HARMONYOS_SUPPORTED_COMMANDS = new Set<string>();
void HARMONYOS_SUPPORTED_COMMANDS;
```

A `void` statement whose only job is to stop the linter noticing that production code exists solely so `staticCommandSetViolations` (`scripts/layering/runtime-command-cutover-policy.ts:394-401`) can find an identifier. That is the gate wagging the dog. `staticCommandSets` is optional on the row model — either drop it from the rows that name it, or teach the policy that an absent set is the strongest possible form of retirement. Do not ship a `void`-ed const.

Same file, `:47`: `const COMMAND_CAPABILITY_MATRIX = BASE_COMMAND_CAPABILITY_MATRIX;` is now a pure identity alias — inline it.

And `src/platforms/apple/plugin.ts:19-25`: `APPLE_SUPPORTS_BY_DEFAULT` and `APPLE_UNSUPPORTED_HINT_BY_DEFAULT` are now empty records still wired into the plugin at `:39-40`. Both fields are optional on `PlatformPlugin.capability` (`packages/contracts/src/platform-plugin.ts:56-59`) and Vega already carries neither. Delete the consts and the two keys.

---

## Behaviour changes that are not in the PR description

**5. `perf frames --kind …` is now silently accepted.** Verified by running the new parser:

```
["frames"] {"kind":"memgraph"} => OK {"area":"frames","action":"sample"}
["frames"] {"kind":"bogus"}    => OK {"area":"frames","action":"sample"}
```

The old path rejected both with `--kind is only supported with perf memory snapshot`. `parsePerfRuntimeRequest` accepts `kind` and then ignores it in three of its four branches (`perf-runtime-plan.ts:98-107`, `:144`, `:171`) — a parameter that is honoured only for `memory` is a trap for the next reader. The test that caught this, `perf rejects kind outside memory snapshot at daemon layer`, was deleted with nothing taking its place.

Related, same function: `parseMemoryRequest` opportunistically reads `values[2]` as the kind (`:127`), so `perf memory sample junk more` now fails with `perf memory snapshot --kind must be android-hprof or memgraph` — the wrong error for stray positionals. The `frames` branch bounds `values.length`; `memory` does not. And `:128` re-implements `isPerfMemoryKind` as two literal comparisons, which is also what forces the `kind as PerfKind` cast at `:137`; the canonical predicate is right there in `@agent-device/contracts/observability` and it narrows.

**6. `perf memory sample` lost its `sampling.snapshot` block** — `test/integration/provider-scenarios/android-lifecycle.test.ts` had to change from `['memory','snapshot']` to `['memory']`. That is a discoverability payload agents were reading.

**7. `recordsSessionAction: true` is now false in fact.** The Apple xctrace handler was the last perf path calling `recordSessionAction`; the new handler calls it nowhere, and the Android path never did. The descriptor still declares `recordsSessionAction: true`, which derives `replayScopedAction` and the MCP `noRecord` projection — so `--no-record` is now meaningless on `perf`. Either restore the record or flip the descriptor.

**8. Confirm the watchOS narrowing is intended.** `applePerfFacts` (`packages/platform-apple/src/runtime.ts:497`) excludes watchOS, which the old `APPLE_SIM_AND_DEVICE` bucket admitted. Also: `DeviceKind` is `simulator | emulator | device`, so `(kind === 'simulator' || kind === 'device')` is a no-op for Apple and the second clause's `kind === 'device' &&` is implied by `iosPhysicalDeviceBackend`. The whole predicate is `device.iosPhysicalDeviceBackend !== 'xctest' && resolveDeviceAppleOs(device) !== 'watchos'`. It also has no return type annotation.

---

## Type boundaries the refactor left looser than it found them

**9. `lastPerfProfile?: PerfData` = `Readonly<Record<string, unknown>>`** replaced three typed fields (`lastProfileTracePath`, `lastProfileTemplate`, `lastMode`). The handler now reads it back with stringly-typed lookups (`readDataString(last, 'outPath')`, `readDataString(last, 'template')` — `session-perf-runtime.ts:122,141`) and the impl re-validates six fields out of it by hand (`readStoppedAndroidProfile`). I understand why it is opaque — the Android report needs `packageName`/`appPid`/`profilerPid`/`remotePath` and the daemon must stay platform-free — but opaque should mean *branded*, not `Record<string, unknown>`. A `PerfProfileHandoff` opaque type gets the same neutrality, deletes `readDataString`, and lets the malformed case have its own message instead of reusing `perf cpu profile report requires a stopped profile trace or tracePath option` (`platform-runtime-perf-host-impl.ts:303-306`) for a completely different failure.

**10. The daemon now infers a capture mode from a capture kind** because the two platforms emit different payload shapes. `activePerfCaptureMode` (`session-perf-runtime.ts:251-257`) maps `simpleperf → cpu-profile` and `perfetto → trace` when `mode` is absent — absent because `compactAndroidCapture` emits `type` while `compactAppleCapture` emits `mode`. That is platform knowledge leaking back into the neutral handler through an untyped `inspect()`. Type `PerfNativeCaptureLiveHandle.inspect()` as `{ kind; mode }` and the inference table plus `perfCaptureStopMismatch`'s `undefined` tolerance both disappear.

**11. Two decoders in this PR treat untrusted durable JSON at opposite rigour.** `readStoppedAndroidProfile` checks every discriminator and carries a comment saying it must. `decodePerfCaptureDescriptor` (`platform-runtime-perf-capture-host.ts:335-345`) checks `isRecord(body.capture)` and then does `body as unknown as PerfCaptureDescriptor` — and `cleanupAndroidPerfCaptureDescriptor` hands that straight to `cleanupAndroidNativePerfSession`, which reads `remotePath`/`profilerPid` off it. Same boundary, same PR, no cast should survive there.

**12. `compactAndroidCapture(result: AndroidNativePerfSession & Record<string, unknown>)`** (`:283`) widens a typed shape so it can read five undeclared fields. Declare the stopped-result type in the Android module instead.

---

## Duplication this PR is the second (or fourth) instance of

**13. `perfCaptureSessionStateOwnershipViolations`** (`scripts/layering/runtime-command-cutover-extensions.ts:169`) is a verbatim copy of `audioProbeSessionStateOwnershipViolations` (`:143`) with two strings changed. Second instance — parameterise it: `sessionStateRecordOwnership(field, owners, message)`.

**14. The durable-resource teardown set is now enumerated in four hand-maintained places**, and this PR had to edit all four: `DURABLE_SESSION_RESOURCE_KINDS`, `DEVICE_CLAIM_DURABLE_RESOURCES`, the `teardownSessionResources` step list, and `stopBestEffortSessionResources`. `stopSessionPerfCapture` is also the third byte-identical copy of `finishSessionAudioProbe` / `finishSessionScreenRecording` (`session-teardown.ts:34-43` vs `:201-227`) — read the store, check the slot, call `finishLive`. One helper over the durable resource object collapses all three, and the ordered step list should be derived from the same closed set the claim reconciler already uses. (Your own "what enumerates N" heuristic applies squarely here.)

**15. `createAppleCaptureHandle` and `createAndroidCaptureHandle`** (`platform-runtime-perf-capture-host.ts:128`, `:157`) are the same handle with two closures swapped, and inside each one `forceCleanup` and `[Symbol.asyncDispose]` repeat the memoised cleanup verbatim (`:142-153`, `:181-189`). One factory taking `{ inspect, initialOutPath, stop, clean }` replaces both, and inside it:

```ts
const cleanupOnce = async () => (cleanup ??= clean().then(() => ({ status: 'cleaned' }) as const, cleanupPending));
…
forceCleanup: cleanupOnce,
[Symbol.asyncDispose]: async () => assertCleanup(await cleanupOnce()),
```

---

## Coverage

This is the part I would not merge without.

- **Zero test files reference `platform-runtime-perf-host-impl.ts` (321 lines), `platform-runtime-perf-capture-host.ts` (358 lines), or `perf-runtime-plan.ts` (191 lines).** That is 870 new production lines containing the whole per-platform collector wiring, the durable descriptor codec, the CLI grammar, and both untrusted-JSON decoders.
- Against that, roughly 1,470 lines of tests were deleted. Twenty behavioural tests (`perf memory snapshot reports physical iOS memgraph as unavailable`, `perf cpu profile rejects xctrace on Android sessions`, `perf trace rejects perfetto on Apple sessions`, `perf memory snapshot returns artifact-shaped unsupported payload on unsupported platforms`, `perf cpu profile reports a missing package with an actionable hint`, `perf rejects kind outside memory snapshot at daemon layer`, …) were replaced by three structural ones.
- The three new tests are good at what they do — single-bind proof, refuse-before-bind, durable adoption — but they pin the *architecture*, not the *behaviour*. `validatePerfFlagPlacement` has no test at all, which is why §5 shipped unnoticed.
- The macOS coverage manifest makes the downgrade explicit: perf evidence moved from `perf memory samples macOS app sessions` to `classifies the %s leaf explicitly`. Facts replaced behaviour in the manifest too.

At minimum, please add a `perf-runtime-plan.test.ts` for the grammar (it is a pure function — cheap and it would have caught §5 and the stray-positional message) and port the per-platform payload assertions onto the new host impl.

---

## Smaller, still worth fixing

- `src/daemon/request-handler-chain.ts:159-160`: `params.perfCaptureAdmissionLedger ?? createPerfCaptureAdmissionLedger()` mints a **fresh ledger per request**. The ledger's entire content is an in-memory `blocks` map with a 5-minute TTL whose message is "Do not start a replacement in this daemon process" — a per-request instance silently disables it. Both siblings on that params type are non-optional. Make this one non-optional too and delete the fallback (the router already defaults it once, so this branch is unreachable today — which is exactly when a wrong default is cheapest to remove).
- `packages/contracts/src/perf-runtime-plan.ts:81-83`: `if (request.area !== 'cpu') throw …` is statically unreachable — after the `start`/`stop` returns, TypeScript narrows `request` to the `cpu`/`report` member (I verified the narrowing compiles). Dead defensive branch duplicating a parse-time check.
- `platform-runtime-perf-capture-host.ts:87-94`: a nested ternary calling `throwWrongNativeKind`, a `never`-returning helper that exists only to smuggle a `throw` into an expression. A plain `if` on `input.mode` removes both the ternary and the helper.
- `platform-runtime-perf-capture-host.ts:248-252`: `.then(…, cleanupPending)` nested inside `.then(…, cleanupPending)` in an otherwise `async/await` file.
- `platform-runtime-perf-host-impl.ts:241-246`: `missingAppMetric(platform: string)` picks its noun with a nested ternary on the display string. Pass the noun.
- `session-perf-runtime.ts:91,96,101,116,120`: `requireRuntimeBinding(params.bindDevice)` is re-derived in every `case`. Hoist it above the `switch`; the per-branch `bind(device, plan.use)` stays (I know the cutover gate needs the literal call site).
- `session-perf-runtime.ts:356`: `--out is only supported with perf memory snapshot` is raised by a guard that also permits `cpu` and `trace`. Pre-existing wording, but it is now the only message on that path and it is wrong for two of the four areas.

---

Happy to re-review quickly once §1–§4 and the grammar test land. The daemon half of this is the shape the wave wanted; it is the host half that needs the same treatment.

thymikee:
Not ready — four P1 blockers remain at `c9d14e646a17a8e15bdfe91e692c4240f0fcb724`.

1. Required changed-line coverage is red: 209/390 (53.59%) versus the 70% gate, with gaps in the new parser, perf host/capture wiring, runtime handler, and recovery code. This is a real coverage failure, not the earlier Actions outage.
2. The implementation misses the agreed ADR-0019/R64 ownership shape: `src/platform-runtime-perf-host-impl.ts` still switches across Apple/Android/Harmony and owns all three implementations in root `src/`, while packages only forward facts through `bindPerfRuntimeOperations`. The `Partial` + cast + runtime `TypeError` join also weakens the typed owner seam. Move platform mechanics into their owning packages and expose a smaller complete contract.
3. Public validation regressed: `parsePerfRuntimeRequest` ignores `input.kind`, so `perf frames --kind …` is accepted although the old handler rejected `--kind` outside `perf memory snapshot`. Add a direct parser regression test and preserve the documented surface.
4. Recording semantics were dropped: `perf` still declares `recordsSessionAction: true`, exposing `--no-record`/MCP recording behavior, but `session-perf-runtime.ts` never calls `recordSessionAction`. Restore recording or intentionally change the descriptor and surface.

Residuals: no live Apple/Android/Harmony perf evidence is reported, and the size report is +4.5 kB gzip. The wrapper-heavy design should be simplified before re-review.

thymikee:
Addressed the review blockers in `2dba07e1a7`:

1. Changed-line coverage now passes locally at 286/408 (70.10%, threshold 70%). Full coverage is green: 1,149 files and 8,609 tests.
2. Removed `src/platform-runtime-perf-host-impl.ts` and the root platform switch. `PerfRuntimeHost` now exposes complete Apple/Android/Harmony sub-hosts; platform packages own operation policy and construction, while root composition supplies lazy native mechanics. The `Partial`/cast/runtime-`TypeError` join is gone.
3. Restored strict grammar behavior: `perf frames --kind`, misplaced `--out`, stray memory-sample positionals, and invalid snapshot kinds fail explicitly. Direct parser and CLI regression tests cover these cases.
4. Restored recording through `recordSessionAction`; successful perf actions remain reusable, while failed capture stops are not recorded.

Also tightened the profile handoff and capture snapshot types, fully validates durable descriptors without casts, made the capture admission ledger required at the request-chain boundary, parameterized the duplicate resource-ownership gate, removed the redundant Apple capability defaults, and consolidated the duplicated platform execution skeleton into a contracts helper without growing the type-only perf façade's eager closure.

Two intentional decisions remain:

- `perf memory sample` stays the approved explicit sample surface; the removed aggregate/snapshot payload is not restored.
- No new HarmonyOS perf mechanics were added. Existing partial memory sampling remains, and native capture stays unsupported.

The empty `HARMONYOS_SUPPORTED_COMMANDS` declaration was restored after the planted layering gate proved it is a live machine-checked retirement marker used by existing Wave 5 rows, not dead production behavior. The broader four-way durable teardown catalog redesign remains out of scope for this migration slice.

thymikee:
Code review is clean at the updated head; no remaining actionable code findings. All four prior P1s are resolved: changed-line coverage is green at 286/408 (70.10%); platform packages now own complete Apple/Android/Harmony operation policy/construction without the root platform switch or Partial/cast/TypeError join; strict flag/positional validation has direct regressions; and successful perf requests again record session actions while failed stops remain unrecorded. The typed profile handoff, durable descriptor validation, capture admission ledger, and duplicate construction paths were also tightened. Exact-head required checks are green.

Size scrutiny: the move-dominated change is about +585 net production lines, npm unpacked +1.6 kB, gzip +6.6 kB, with effectively unchanged startup. The deeper package-owned design now removes the wrapper-heavy root implementation; no further material simplification was identified without reopening the broader durable-resource catalog redesign.

Residual evidence risk: no real routed perf command has been demonstrated on this head for Apple, Android, or HarmonyOS, and native start/stop artifact lifecycles are not covered by the smoke jobs. This is clean for human review, not a merge-ready evidence claim.

PR #2065, opened 2026-08-26
fix(cli): name `--udid` when a device identity is passed to `--device`

### LINKED ISSUE #2064: `--device <udid>` fails `DEVICE_NOT_FOUND` without hinting at `--udid`, which is absent from Global Flags

**Steps to reproduce.**
```bash
agent-device devices --json            # exposes UDIDs
agent-device open com.example.app --platform ios --device <udid> --relaunch
agent-device help commands             # look for a way to pin a device by UDID
```

**Observed.** The UDID is treated as a device *name*:
```
Error (DEVICE_NOT_FOUND): No device named <udid>
Hint: Verify the target device is booted/connected and selectors match.
```
The hint does not mention `--udid`, and `help commands` does not list it either — its Global Flags
block is (one description elided):
```
Global Flags:
  --config <path>             Load CLI defaults from a specific config file (required for connection/provider defaults outside user config)
  --debug, --verbose, -v      Enable debug diagnostics; …
  --cost                      Include per-command wall-clock latency (cost.wallClockMs) in the response
  --level digest|default|full  Response detail level: digest (token-cheap), default (today), or full. Default keeps the wire shape unchanged.
  --json                      JSON output
  --help, -h                  Print help and exit
  --version, -V               Print version and exit
```
`--udid <udid>` appears only inside `agent-device help device` usage. It works (verified: a pinned
`open` + `snapshot -i` hit the intended simulator among three booted ones) — it is just
undiscoverable, and with several local simulators named "iPhone 17", `--device <name>` cannot pin one.

**Expected.** `--device` accepts a UDID as well as a name, or `DEVICE_NOT_FOUND` hints "did you mean
`--udid <udid>`?"; and `--udid` (with `--device`, `--serial`, `--platform`, `--session`) is listed in
the Global Flags block of `help commands`. Related: #1335.

**Notes/workaround.** Resolve name → UDID with `agent-device devices --json`, then pass `--udid <udid>`.

NicolasBataille:
Closes #2064

## Summary

`--udid` is the only selector that pins one simulator when several share a name, and it was
effectively undiscoverable: `help commands` listed no device selector at all, and passing a UDID to
`--device` answered with a literal lookup failure that neither named `--udid` nor hinted that a
UDID-taking flag exists.

Before:

```
$ agent-device snapshot --platform ios --device 204BFFD9-9644-4830-B2C1-1B946597A07C
Error (DEVICE_NOT_FOUND): No device named 204BFFD9-9644-4830-B2C1-1B946597A07C
Hint: Verify the target device is booted/connected and selectors match.
```

After:

```
Error (DEVICE_NOT_FOUND): No device named 204BFFD9-9644-4830-B2C1-1B946597A07C
Hint: 204BFFD9-9644-4830-B2C1-1B946597A07C is the id of "iPhone 17", not its name.
      Did you mean --udid 204BFFD9-9644-4830-B2C1-1B946597A07C?
```

Two changes:

1. **`resolveDeviceByName` answers the flag mistake.** When the `--device` value is the id of a
   listed candidate it names that device and the flag the value belongs to — `--udid` for Apple,
   `--serial` for serial-addressable platforms. When it merely has UDID shape (8-4-4-4-12 hex) but
   matches no listed device, it names `--udid` anyway, because a shut-down simulator is still a
   wrong-flag answer. An ordinary unknown name keeps the generic hint untouched.

   This mirrors, rather than invents, the answer `assertSelectorFlagMatchesPlatform` already gives
   a few lines above for a mismatched identity flag (`--udid` on `--platform android`).

2. **`help commands` renders a `Device Selection` section** for the selectors every command
   accepts: `--platform`, `--device`, `--udid`, `--serial`, `--session`.

   ```
   Device Selection (accepted by every device command):
     --platform apple|android|...  Platform to target (`apple` aliases the Apple automation backend)
     --device <name>               Device name to target (a UDID belongs in --udid, a serial in --serial)
     --udid <udid>                 Apple device or simulator UDID; the only selector that pins one
                                   device when several share a --device name
     --serial <serial>             Android device or Vega VVD serial
     --session <name>              Named session
   ```

   Deliberately a **new help section rather than a `GLOBAL_FLAG_KEYS` addition**: those keys mark a
   flag as supported by `*` in `buildOptionSpecs` and move it from `supportedFlags` to `globalFlags`
   in `explain`. These selectors reach device resolution, not the CLI envelope; they stay in
   `COMMON_COMMAND_SUPPORTED_FLAG_KEYS`, so the option schema, `explain` output and MCP surface are
   byte-identical.

`--udid` and `--device` also get sharper one-line descriptions, which is where an agent reading the
new section actually learns the distinction.

## Tests

- `packages/kernel/src/device-selector-flags.test.ts` — the existing home for selector-flag
  mistakes — gains four cases: a UDID and a serial passed to `--device` (each naming the right
  flag and the device it identifies), a UDID-shaped value no listed device carries, and an ordinary
  unknown name keeping the generic hint. All four were observed red against the pre-fix code.
- `src/cli-schema/cli-help-topics.test.ts` gains a `Device Selection` section assertion. The
  existing "only global flags in its global flags section" test slices from `Global Flags:` to
  `Configuration:`; the new section sits above `Global Flags:`, so that slice is unaffected.

`pnpm check:quick`, `pnpm test:unit` (8092 passed), `pnpm check:command-docs` and
`pnpm check:agent-guidance` are green.

## Live check

Run from the built clone against a local simulator set with two simulators named "iPhone 17":

- `--device <udid>` → the new identity hint naming `--udid` and `"iPhone 17"`.
- `--device 11111111-2222-3333-4444-555555555555` (no such device) → the shape hint naming `--udid`.
- `--device "iPhone 99"` → unchanged generic hint.
- `help commands` renders the new section.

- **The UDID regex.** It is deliberately the *fallback*: the primary signal is exact id-equality
  against the candidate set, which needs no shape knowledge and gets `--serial` right for Android.
  The regex only covers a device the current inventory does not list. Dropping it would still fix
  the reported case; keeping it covers the shut-down-simulator variant the issue's workaround
  section implies.
- **`--device` accepting a UDID outright** was the issue's other suggested fix. Not taken: it would
  make `--device` a two-meaning flag and quietly duplicate `--udid`/`--serial`'s platform routing,
  including the `assertSelectorFlagMatchesPlatform` cross-check. Naming the right flag is smaller
  and keeps one meaning per selector.
- **Section placement and title.** `Device Selection (accepted by every device command)` is long;
  a shorter title works if it does not read as "global", which is exactly the distinction that
  keeps the option schema unchanged.
- **Selector set.** `--target` is also a common selector and was left out to match the issue's list.
  Adding it is a one-line change to `DEVICE_SELECTION_FLAG_KEYS`.

### DIFF
--- a/packages/kernel/src/device-selector-flags.test.ts
+++ b/packages/kernel/src/device-selector-flags.test.ts
@@ -75,22 +75,16 @@ test('a serial passed to --device names --serial, not --udid', async () => {
   assert.match(String(error.details?.hint ?? ''), /Did you mean --serial emulator-5580\?/);
 });
 
-test('a UDID-shaped --device value names --udid even when no listed device has that id', async () => {
-  const error = await resolveError([APPLE], {
-    platform: 'ios',
-    deviceName: '204BFFD9-9644-4830-B2C1-1B946597A07C',
-  });
-  assert.equal(error.code, 'DEVICE_NOT_FOUND');
-  assert.match(
-    String(error.details?.hint ?? ''),
-    /--device takes a device name\. Did you mean --udid 204BFFD9-9644-4830-B2C1-1B946597A07C\?/,
-  );
-});
-
-test('an ordinary unknown --device name keeps the generic device-not-found hint', async () => {
-  const error = await resolveError([APPLE], { platform: 'ios', deviceName: 'iPhone 99' });
-  assert.equal(error.code, 'DEVICE_NOT_FOUND');
-  assert.equal(error.details?.hint, undefined);
+test('an unknown --device value gets the generic hint, even when it looks like a UDID', async () => {
+  // The hint is grounded in observed identity: only a candidate's actual id earns the
+  // wrong-flag answer. A UDID-shaped value that no listed device carries stays a plain
+  // unknown name — guessing from shape would misclassify UUID-named devices and miss
+  // every identity syntax the shape guess does not cover.
+  for (const deviceName of ['204BFFD9-9644-4830-B2C1-1B946597A07C', 'iPhone 99']) {
+    const error = await resolveError([APPLE], { platform: 'ios', deviceName });
+    assert.equal(error.code, 'DEVICE_NOT_FOUND');
+    assert.equal(error.details?.hint, undefined);
+  }
 });
 
 async function resolveError(
--- a/packages/kernel/src/device-selector-flags.test.ts
+++ b/packages/kernel/src/device-selector-flags.test.ts
@@ -56,6 +56,43 @@ test('an unspecified platform keeps the existing device-not-found behavior', asy
   assert.equal(error.code, 'DEVICE_NOT_FOUND');
 });
 
+// `--device` takes a NAME. Passing a UDID there answered "No device named <udid>" with the generic
+// booted/connected hint (#2064) — true, unactionable, and silent about `--udid` existing at all.
+
+test('a UDID passed to --device names --udid and the device it identifies', async () => {
+  const error = await resolveError([APPLE, ANDROID], { deviceName: 'SIM-001' });
+  assert.equal(error.code, 'DEVICE_NOT_FOUND');
+  assert.match(error.message, /No device named SIM-001/);
+  assert.match(
+    String(error.details?.hint ?? ''),
+    /SIM-001 is the id of "iPhone 16", not its name\. Did you mean --udid SIM-001\?/,
+  );
+});
+
+test('a serial passed to --device names --serial, not --udid', async () => {
+  const error = await resolveError([APPLE, ANDROID], { deviceName: 'emulator-5580' });
+  assert.equal(error.code, 'DEVICE_NOT_FOUND');
+  assert.match(String(error.details?.hint ?? ''), /Did you mean --serial emulator-5580\?/);
+});
+
+test('a UDID-shaped --device value names --udid even when no listed device has that id', async () => {
+  const error = await resolveError([APPLE], {
+    platform: 'ios',
+    deviceName: '204BFFD9-9644-4830-B2C1-1B946597A07C',
+  });
+  assert.equal(error.code, 'DEVICE_NOT_FOUND');
+  assert.match(
+    String(error.details?.hint ?? ''),
+    /--device takes a device name\. Did you mean --udid 204BFFD9-9644-4830-B2C1-1B946597A07C\?/,
+  );
+});
+
+test('an ordinary unknown --device name keeps the generic device-not-found hint', async () => {
+  const error = await resolveError([APPLE], { platform: 'ios', deviceName: 'iPhone 99' });
+  assert.equal(error.code, 'DEVICE_NOT_FOUND');
+  assert.equal(error.details?.hint, undefined);
+});
+
 async function resolveError(
   devices: DeviceInfo[],
   selector: Parameters<typeof resolveDevice>[1],
--- a/packages/kernel/src/device-selector-flags.test.ts
+++ b/packages/kernel/src/device-selector-flags.test.ts
@@ -75,6 +75,22 @@ test('a serial passed to --device names --serial, not --udid', async () => {
   assert.match(String(error.details?.hint ?? ''), /Did you mean --serial emulator-5580\?/);
 });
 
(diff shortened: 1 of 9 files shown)


thymikee:
[P1] Remove the partial UDID-shape heuristic and keep the hint grounded in observed device identity. The exact `candidates.find(device.id === deviceName)` path already fixes #2064 for every discovered Apple/serial device and reuses the kernel’s platform facts. `DEVICE_UDID_SHAPE` then reconstructs Apple identity syntax with a UUID-only regex: it excludes the modern physical-device UDIDs already used in this repo (`00008150-001849640CF8401C`), can misclassify an arbitrary UUID-shaped device name, and needs a long comment to justify a shutdown-device case that current inventory may still list. Either delete this fallback (the smaller fix the PR body already identifies) or move complete Apple identity recognition behind an Apple-owned typed interface with tests for simulator and physical formats. Keep the exact-candidate `--udid`/`--serial` hint and the help section. Exact-head CI is also still absent, so do not apply `ready-for-human` yet.

thymikee:
Addressed the P1 in 6ebda3406d and 830a577d1d (maintainer commits on this branch):

- **Shape heuristic deleted.** The hint is grounded only in observed identity: a listed candidate's exact `device.id`, with the flag derived from that device's platform. `DEVICE_UDID_SHAPE` is gone, and with it the missed modern physical-device UDIDs (`00008150-…`) and the UUID-named-device misclassification.
- **No dead instructions.** An adversarial pass found the platform ternary sent web/linux device ids to `--udid`, which resolves only Apple devices. The hint now emits only for platforms that have an identity flag (Apple → `--udid`, serial-addressable → `--serial`); web/linux keep the generic hint.
- **The section header no longer overclaims.** `connect` takes no `--udid`/`--serial` and `device` no `--session`, so the header says "shared by device commands" and points at `help <command>`.
- **The MCP/JSON schema got the same fix**: `device` says a UDID belongs in `udid`; `udid` no longer reads "iOS device UDID selector.". `--serial` now names HarmonyOS, which the resolver always accepted.
- **Bonus truth fix**: `help physical-device` claimed `--device <name-or-udid>`; `--device` never matched by UDID.

One correction to this PR's description: `explain <cmd> --json` carries `usageDescription` into `cli.supportedFlags[].description`, so the sharpened descriptions do change that output; the option schema and MCP metadata are unchanged (`check:mcp-metadata` green).

Live-verified on this branch (three booted targets): `snapshot --platform ios --device <udid>` answers `…is the id of "iPhone 17 Pro", not its name. Did you mean --udid <udid>?`, and `help commands` renders the Device Selection section.

Follow-up worth its own issue: `resolveDeviceByName` silently picks the first of two same-named devices (plain `.find`), bypassing the SINGULAR RESOLUTION doctrine — pre-existing, and now more visible since the help advertises the collision case.

Validation: unit-core green (incl. kernel + cli-schema), `check:quick`, output-economy, `check:mcp-metadata`, `check:layering` all green.

thymikee:
The original P1 is resolved at `830a577d1dbd6fd4c89923989b824b2fc677d37f`. One P2 metadata inconsistency remains: `src/commands/command-input.ts` still describes `serial` as only an “Android device or Vega VVD serial selector,” while the kernel’s serial-addressable platforms and this PR’s CLI help include HarmonyOS. `commonProperties()` owns the MCP/JSON structured-input schema, so HarmonyOS is still undiscoverable on that public surface. Please align the description to “Android, HarmonyOS, or Vega VVD serial selector” and pin it with the schema/metadata regression.

thymikee:
Re-reviewed exact head `830a577d1dbd6fd4c89923989b824b2fc677d37f`.

P2: `src/commands/command-input.ts` still describes `serial` in the structured common-input/MCP JSON schema as Android or Vega only, while the resolver and new CLI guidance also support HarmonyOS. Align this public-schema description and pin the emitted metadata. Otherwise clean, but not ready until this cross-surface mismatch is fixed.

thymikee:
P2 addressed in 919fab31c5: `serial`'s structured/MCP schema description now names HarmonyOS, matching the resolver and the CLI text. A metadata test pins all three device-selector descriptions on every emitted command schema, so any cross-surface drift fails instead of shipping.

Issue #1394, opened 2026-07-24
session list omits cwd-scoped sessions opened via replay's internal open dispatch


thymikee:
Found during live iOS/Android validation of #1390's #1384 fix.

## Repro (real backends, both platforms)

```sh
printf 'context platform=android device="Pixel 9 Pro XL" kind=emulator theme=unknown\nopen "com.android.settings" --relaunch\n' > /tmp/open-only.ad
AGENT_DEVICE_STATE_DIR=/tmp/ad-state node bin/agent-device.mjs replay /tmp/open-only.ad --json
# => success, data.session = "cwd:<hash>:default", data.sessionActive = true
AGENT_DEVICE_STATE_DIR=/tmp/ad-state node bin/agent-device.mjs session list --json
# => { "sessions": [] }        <-- empty, but the session is NOT gone
AGENT_DEVICE_STATE_DIR=/tmp/ad-state node bin/agent-device.mjs snapshot --json --session "cwd:<hash>:default"
# => succeeds, real snapshot of the still-open app
```

Reproduced identically on a booted iOS simulator (`open "com.apple.Preferences"`).

## Root cause (traced, not yet fixed)

`session_list` (`src/daemon/handlers/session-inventory.ts:33-40`) filters `sessionStore.toArray()` by `sessionMatchesScope(session, resolveImplicitSessionScope(req))` (`src/daemon/session-routing.ts`) — comparing the CURRENT request's freshly-computed cwd scope against a `session.sessionScope` field **stored on the session at creation time** (`src/daemon/handlers/session-open.ts:433,520`: `sessionScope: existingSession?.sessionScope ?? resolveImplicitSessionScope(req)`).

The session's **storage key** (`cwd:<hash>:default`) is computed correctly at the top-level `replay` request's own dispatch (`resolveEffectiveSessionName`), using the CLI's real `meta.cwd`. But the session's **`sessionScope` field** gets set by whatever `req` reaches `session-open.ts` when the script's `open` action is dispatched internally during replay (`invokeReplayAction` → the real open handler) — and that internally-constructed request apparently doesn't carry the same `meta.cwd` a top-level CLI request does, so `resolveImplicitSessionScope` returns `undefined` for it, leaving `session.sessionScope` unset.

Later, `session_list`'s own top-level request computes a proper (non-undefined) scope from the CLI's real cwd, compares it to the session's `undefined` `sessionScope`, and filters it out — even though the session is fully alive and reachable by its exact name.

## Impact

[...]

PR #2068, opened 2026-08-26
fix(daemon): address cwd-scoped sessions by their store key, not their public name

NicolasBataille:
Refs #2031, #1394 (deliberately not `Closes` — see below)

## Status of #2031

#2031 is **closed**, by b1ee2a77 (#2057), which is in `main`. That fix covers the reported
mechanism: a replaced-but-still-running daemon keeping the device claim, now classified
`owner-daemon-superseded` and reconciled like a dead owner. I re-read that diff; it does what the
issue's headline asks.

The **follow-up comment on the same issue** describes a different mechanism, and that one still
reproduces on `main`. This PR fixes that half.

## Root cause

`resolveEffectiveSessionName` scopes an implicit session to the caller's worktree: a session opened
without `--session` is **named** `default` and **stored** under `cwd:<hash>:default`. The two are
not interchangeable, because passing `--session` sets `sessionExplicit`, which *disables* scoping —
so `--session default` addresses the literal key `default`, a different session.

Three surfaces built caller-facing text from `SessionState.name`:

- `buildDeviceInUseBySessionError` → `Device is already in use by session "default".`
- `buildSessionRecoveryHint` → `agent-device close --session default`
- `publicSessionInfo` (`session list`) → `sessionStateDir: <state>/sessions/default`

Each names something unreachable. Reproduced on `main` (iOS simulator, this clone):

```
$ agent-device open <app> --platform ios --udid <udid>          # cwd-scoped `default`
$ agent-device open <app> --platform ios --udid <udid> --session qa
Error (DEVICE_IN_USE): Device is already in use by session "default".
Hint: ... first run agent-device close --session default.

$ agent-device close --session default                          # follow the hint
Error (SESSION_NOT_FOUND): No active session                    # <- the hint's own command

$ agent-device open ... --session qa                            # retry
Error (DEVICE_IN_USE): ... by session "default".                # still held
```

That is the comment's report exactly: a `DEVICE_IN_USE` naming a session the documented recovery
cannot close. The same mismatch makes `session list` report a `sessionStateDir` and `runnerLogPath`
under `<state>/sessions/default` — a directory that is never created — while the session's real
artifacts, `runner.log` included, sit in `<state>/sessions/cwd_<hash>_default`, which is what `open`
already answers with. An operator following `session list` to read a runner log finds nothing there.

This is the #1394 family: a session's *identity* and its *address* diverged, and the surfaces that
report identity kept publishing it as an address.

## Fix

- `SessionStore.entries()` exposes the key alongside the record.
- `findNewSessionDeviceConflict` reads the conflicting session through it and passes the key to
  `buildDeviceInUseBySessionError`, which reports it in the message, in `details.session`, and in
  the hint.
- `buildSessionRecoveryHint` takes the session's **address** explicitly. The two other callers
  (`session-selector.ts`, `request-lock-policy.ts`) act on the session the current request named,
  where the name *is* the address, so they pass `session.name` and their text is unchanged.
- `session list` resolves `sessionStateDir`/`runnerLogPath` from the store key.

Explicitly named sessions are unaffected everywhere; the `DEVICE_IN_USE` help-conformance sample is
byte-identical (its fixture session is named `checkout`).

## Tests

Three new files; all five assertions were observed red against the pre-fix code.

- `src/daemon/session-recovery-hints.test.ts` — device-in-use, selector-conflict and recording
  recoveries all name the address; an explicitly named session addresses itself unchanged.
- `src/daemon/handlers/__tests__/session-open-device-in-use.test.ts` — the producer reports the
  address in message, details and hint.
- `src/daemon/handlers/__tests__/session-inventory-scoped-paths.test.ts` — `session list` keeps the
  public `name` but resolves paths to `cwd_<hash>_default`.

`pnpm check:quick`, `pnpm test:unit` (8093 passed) and `pnpm check:layering` are green.

## Live check

Same sequence as the repro above, after the fix:

```
Error (DEVICE_IN_USE): Device is already in use by session "cwd:8bea844ab16aa9b3:default".
Hint: ... first run agent-device close --session cwd:8bea844ab16aa9b3:default.

$ agent-device close --session "cwd:8bea844ab16aa9b3:default"
Closed: cwd:8bea844ab16aa9b3:default
$ agent-device open ... --session qa
Opened: <app>
```

`session list` now reports `.../sessions/cwd_8bea844ab16aa9b3_default`, which exists.

The issue's *original* close→plain-open sequence does **not** reproduce on `main` — a plain `close`
followed by a plain `open` in the same worktree succeeds. Its `--device "iPhone 17"` variant now
also raises `AMBIGUOUS_MATCH` when two simulators share that name, which is separate and correct.

- **Exposing the scoped key to users.** `close --session cwd:<hash>:default` is ugly, and it is
  a hash the caller never chose. It is the value that works today, and a hint that works beats a
  hint that reads nicely. The alternative — teaching `close`/`open` to resolve a *public* name
  against scoped keys within the caller's scope — is a bigger change to `resolveEffectiveSessionName`
  and its `sessionExplicit` contract, and belongs in its own PR. Say the word and I will draft it
  instead.
- **`entries()` on `SessionStore`.** The key can also be reconstructed as
  `session.sessionScope ? cwd:<id>:<name> : name`, which needs no new accessor — but that
  reconstruction is wrong for exactly the case #1394 documents, a session whose `sessionScope` was
  never propagated. Reading the key the store actually used cannot be wrong.
- **The two unchanged callers.** They now pass `session.name` explicitly, which looks like
  boilerplate. Making the parameter optional would hide the question at the sites most likely to be
  wrong next; the comment at each site records why the name is the address there.
- **Scope.** This is a follow-up to a closed issue. If the maintainers prefer, it can be re-filed as
  its own issue first — the reproduction above is self-contained.

### DIFF
--- a/.fallowrc.json
+++ b/.fallowrc.json
@@ -192,7 +192,15 @@
       ]
     }
   ],
-  "usedClassMembers": ["name", "listActiveLeases", "delete", "values", "elapsedMs", "isExpired", "findByDevice"],
+  "usedClassMembers": [
+    "name",
+    "listActiveLeases",
+    "delete",
+    "values",
+    "elapsedMs",
+    "isExpired",
+    "findByDevice"
+  ],
   "rules": {
     "unused-types": "error",
     "duplicate-exports": "off"
--- a/.fallowrc.json
+++ b/.fallowrc.json
@@ -192,7 +192,7 @@
       ]
     }
   ],
-  "usedClassMembers": ["name", "listActiveLeases", "delete", "values", "elapsedMs", "isExpired"],
+  "usedClassMembers": ["name", "listActiveLeases", "delete", "values", "elapsedMs", "isExpired", "findByDevice"],
   "rules": {
     "unused-types": "error",
     "duplicate-exports": "off"
--- a/packages/contracts/src/client-device-view.ts
+++ b/packages/contracts/src/client-device-view.ts
@@ -68,6 +68,13 @@ export type AgentDeviceSessionDevice = {
 
 export type AgentDeviceSession = {
   name: string;
+  /**
+   * The exact value `--session` must carry to address this session, which is not always `name`:
+   * a session opened without `--session` is named `default` but stored — and addressed — as
+   * `cwd:<hash>:default` (#2031/#1394). Optional only because a daemon older than the field does
+   * not send it.
+   */
+  address?: string;
   createdAt: number;
   sessionStateDir?: string;
   runnerLogPath?: string;
--- a/scripts/__tests__/help-conformance-sample-producers.ts
+++ b/scripts/__tests__/help-conformance-sample-producers.ts
@@ -369,7 +369,8 @@ export const SAMPLE_PRODUCERS: SampleProducer[] = [
       const device = { id: 'SIM-001', name: 'iPhone 17 Pro' } as Parameters<
         typeof buildDeviceInUseBySessionError
       >[1];
-      const response = buildDeviceInUseBySessionError(owningSession, device);
+      // An explicitly named session addresses itself, so the sample text is unchanged.
+      const response = buildDeviceInUseBySessionError(owningSession, device, owningSession.name);
       assertErrorResponse(response, 'the by-session conflict');
       return renderErrorResponse(response);
     },
--- a/scripts/__tests__/help-conformance-sample-producers.ts
+++ b/scripts/__tests__/help-conformance-sample-producers.ts
@@ -365,12 +365,12 @@ export const SAMPLE_PRODUCERS: SampleProducer[] = [
     producer: 'the real session-open by-session conflict producer',
     sample: DEVICE_IN_USE_SAMPLE,
     render: () => {
-      const owningSession = { name: 'checkout' } as SessionState;
+      // An explicitly named session is stored under its own name, so address === name here.
+      const owningSession = { address: 'checkout', session: { name: 'checkout' } as SessionState };
       const device = { id: 'SIM-001', name: 'iPhone 17 Pro' } as Parameters<
         typeof buildDeviceInUseBySessionError
       >[1];
-      // An explicitly named session addresses itself, so the sample text is unchanged.
-      const response = buildDeviceInUseBySessionError(owningSession, device, owningSession.name);
+      const response = buildDeviceInUseBySessionError(owningSession, device);
       assertErrorResponse(response, 'the by-session conflict');
       return renderErrorResponse(response);
     },
--- a/src/client/client-normalizers.ts
+++ b/src/client/client-normalizers.ts
@@ -122,6 +122,7 @@ export function normalizeSession(value: unknown): AgentDeviceSession {
   };
   return {
     name,
+    address: readOptionalString(record, 'address'),
     createdAt: readRequiredNumber(record, 'createdAt'),
     sessionStateDir: readOptionalString(record, 'sessionStateDir'),
     runnerLogPath: readOptionalString(record, 'runnerLogPath'),
--- a/src/daemon/__tests__/request-lock-identity-policy.test.ts
+++ b/src/daemon/__tests__/request-lock-identity-policy.test.ts
@@ -2,7 +2,7 @@ import { test } from 'vitest';
 import assert from 'node:assert/strict';
 import { AppError } from '@agent-device/kernel/errors';
 import { applyRequestLockPolicy } from '../request-lock-policy.ts';
-import type { DaemonRequest, SessionState } from '../types.ts';
+import type { DaemonRequest, SessionRef, SessionState } from '../types.ts';
 
 /**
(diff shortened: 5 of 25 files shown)


thymikee:
[P1] Thread the resolved session address through every recovery producer instead of falling back to `SessionState.name`. The new comments in `session-selector.ts` and `request-lock-policy.ts` claim the current session name is its address, but the ordinary implicit route resolves `cwd:<hash>:default`, loads that store entry, then passes only the record whose public `name` is `default`; selector and lock conflicts therefore still emit the same unreachable `close --session default` recovery this PR fixes elsewhere. The Maestro route also already holds the resolved session name. Pass the store key/address through these APIs and add production-route regressions, not only helper tests. `session list` must likewise expose the actual address (while preserving `name` if useful), otherwise its primary discovery surface still tells operators only `default`, which cannot be reused explicitly. Prefer the store’s existing address owner or a narrow `{address, session}` reference over a broad raw `entries()` accessor. Exact-head CI is still absent; do not apply `ready-for-human`.

thymikee:
Addressed the P1 in c81bf0b609 and e3d9da4649 (maintainer commits on this branch):

- **The address is threaded as a type, not a convention.** `SessionRef = {address, session}` is what every recovery producer now takes — `buildSessionRecoveryHint`, `assertSessionSelectorMatches`, `applyRequestLockPolicy`, `buildDeviceInUseBySessionError` — so a producer cannot be handed a bare record. The false "the name IS the address" comments are deleted; the selector- and lock-conflict routes were confirmed to emit the unreachable `close --session default` before the fix (red-checked).
- **Narrow accessors**: `entries()` is gone; the store exposes `lookup(address)`, `findByDevice(deviceId)`, `listRefs()` — callers get the session with its address from the single owner of that mapping.
- **`session list` reports `address`** alongside `name`, threaded through contracts/client normalizers/serialization; it is literally the store key `resolveSessionDir` uses.
- **Production-route regressions**, not helper tests: device-in-use, selector-conflict, lock-policy-conflict, and `session list` all through the real request handler, with the address read back from the store (pattern `^cwd:[0-9a-f]{16}:default$` — the hash is 16 hex in production).
- **Adversarial review verified the address is genuinely actionable** end-to-end: an implicit `open` filed `cwd:<hash>:default`, `close --session cwd:<hash>:default` succeeded, `close --session default` still `SESSION_NOT_FOUND`. Nothing in `--session` handling rejects `:`.
- **Doctor was the same producer class and got the same fix**: the same-device warning enumerated sessions by `name` — its printed `close --session default` was this PR's exact defect, the name-based dedupe hid two cwd-scoped `default` sessions from each other, and its evidence `sessionStateDir` pointed at the never-created `<state>/sessions/default`. It now uses `listRefs()` and reports addresses. The fallow gate is green (`findByDevice` joined the existing resolution false-positive allowlist for this class; a test's CRAP finding fixed by narrowing once).

Follow-ups worth their own issues (pre-existing, same defect family):
- `session-close-lifecycle-teardown.ts` passes `session.name` into app-log/audio/perf teardown, so force-cleanup resolves `<state>/sessions/default` on every implicit-session close.
- The Maestro replay error route drops `hint` entirely (`stripDiagnosticMeta`), so the runnable close command never reaches that caller.
- `snapshot-runtime-binding.ts:366` prints `open --session default` for an implicit session, which would create a different session.
- `SessionStore.resolveStoredSessionName` still has a name fallback — the wrong-address shape `SessionRef` exists to prevent; worth collapsing.

Validation: daemon suite 2322 tests green, `check:fallow` green, `check:daemon-wire-compat` green (field addition, nothing removed), wire-compat tests 39/39, output-economy 26/26, `check:quick`, `check:layering` green. `address` is optional on both sides, so mixed old/new daemon-client pairs degrade to today's shape.

thymikee:
Re-review at `e3d9da464959f4a159df29f146c11071e29a503d`: the previously posted scoped-session-address P1 is fully resolved. `SessionRef { address, session }` is now the typed owning seam for recovery producers; narrow store APIs preserve the pairing; selector, lock-policy, DEVICE_IN_USE, Maestro, session-list, and doctor routes carry/report the actionable address, with production-route regressions. No new code finding. Not labeling ready-for-human yet because all seven exact-head workflows completed `action_required` with zero jobs, so required CI/size evidence did not run.

thymikee:
Re-reviewed at `e3d9da464959f4a159df29f146c11071e29a503d`: no code finding. The typed `SessionRef`/address plumbing and real request-handler regressions resolve the scoped-store-key defect: recovery, selector, lock-policy, Maestro, session list, and doctor now retain/report the actionable store address. #1394 remains a distinct, documented replay-internal scope-propagation producer follow-up.

Not ready yet: exact-head CI has a confirmed owner-action formatting failure in `.fallowrc.json`. Please format that file and rerun the affected checks.

thymikee:
Thank you for the contribution @NicolasBataille!

PR #1989, opened 2026-08-24
fix(ci): repair nightly XCTest and conformance lanes

thymikee:
## Summary

- repair the nightly XCTest expectation and make generated Maestro JSON formatting deterministic
- make the differential oracle non-vacuous with executable fixture semantics, typed agent-device failure provenance, and fail-closed untyped Maestro exits
- stabilize macOS replay lifecycle with confirmed app termination, a unique `role=button label=About` target, and enough assertion budget for one runner recycle
- classify the typed device-claim conflict family as infrastructure in replay suites, including wrapped `REPLAY_DIVERGENCE` results, so `test` stops retrying live-owner conflicts
- rebase onto current `main` and keep #2001 as the single replay-cleanup owner; the stale competing cleanup implementation and tests were removed

The differential regression runs real temporary engine processes through scenario reporting and a matching waiver. Before the fix it failed because the Maestro exit was classified as `behavioral`; after the fix the report is `infrastructure-failed` and remains red.

The device-claim regressions run through the production `test --json` handler path. Before the fix a typed `DEVICE_CLAIM_LIVE_OWNER` result retried four times and omitted the infrastructure tag; after the fix it stops after one attempt and serializes `infrastructure: true`. An untyped, retriable `DEVICE_IN_USE` result still retries four times and serializes no infrastructure tag; the classifier does not infer a claim from its code or message.

## Validation

- `pnpm check:affected --run` — all 34 runnable local checks passed (54 selected; GitHub-authoritative native/device checks skipped locally), including 3,323 related tests
- `pnpm maestro:conformance` — 54/54 passed
- focused device-claim, test-suite classification, and file-size ratchet tests — 42/42 passed
- focused macOS lifecycle/replay cleanup/failure-classification tests — 103/103 passed
- macOS command coverage contract — 7/7 passed
- live macOS replay with CI retry settings — passed; the first attempt exposed a runner recycle exceeding the old 5s post-click budget
- live macOS replay with suite retries disabled — passed twice consecutively in one attempt (50.8s and 46.5s)
- planted red — the real engine-to-report waiver regression failed on pre-fix classification (`behavioral` instead of `infrastructure`)
- planted red — the production `test --json` regression retried four times and omitted `infrastructure` before the device-claim fix
- planted red — the direct classifier incorrectly treated untyped `DEVICE_IN_USE` as infrastructure before the code-wide shortcut was removed
- exact-head CI — both coverage shards, Coverage Report, Repo Guards, integration, and all four platform smoke jobs passed

[...]

PR #1822, opened 2026-08-18
ci: park perf-nightly to dispatch and stop the coverage-gate cascade double-red (#1781 A3, A5)

thymikee:
## What / why

Two small, decided CI changes from #1781.

### A3 — park `perf-nightly.yml` to dispatch-only

`perf-nightly.yml` writes a JSON/markdown perf report and compares nothing (`scripts/perf/report.ts` just serializes the run; grepping `scripts/perf` + the workflow for `baseline|threshold|regress|compare|delta` finds zero implementation, only a comment stating the intent). It therefore structurally cannot report a regression — a run only turns red on a harness crash.

Evidence (60-day run history + artifact diff):
- 38/38 green since 2026-07-12; "success" only ever meant "didn't crash."
- iOS wall-clock medians swing wildly night-to-night at n=5, warmup=1 on a shared macOS runner: `type` +122%, `snapshot (deep)` +74%, `open (relaunch→root)` +60%, `boot device` +39%. A fixed-threshold comparator would mostly print noise without multi-night smoothing.
- No doc, README, or issue consumes the report (`grep -rniE "perf-nightly|scripts/perf" docs/ CONTEXT.md AGENTS.md README.md` → no hits; `gh issue list --search "perf-nightly"` → only #1781 itself).
- Cost: iOS job ~22min median on a macOS runner + Android ~9min, nightly. This is a public repo, so Actions minutes are free; the defensible cost is runner *occupancy*, not billing — ~22 min/night of a macOS runner held for a report nobody reads.

Implemented exactly the #1781 A1 parking pattern (`replays-manual.yml`): dropped the `schedule:` trigger, kept `workflow_dispatch`, added a short header comment naming the parking reason and the un-park condition. `pnpm perf` and `scripts/perf` are untouched — still runnable on demand.

**`scripts/gate/declarations.ts` unchanged.** perf-nightly declares no `pnpm gate <name>` check of its own — `pnpm perf` isn't a registered gate, and the only `gate:` reference in the file (`swift-runner-ios`, used to build the shared XCTest runner) is already declared and owned by other lanes (`ios.yml`, `xctest-nightly.yml`, `conformance-differential.yml`). `pnpm check:gate-manifest` reports the same `47 checks / 33 lanes` before and after this change, so nothing needed a `MANUAL_ONLY_OWNERS` entry.

[...]

PR #2069, opened 2026-08-27
fix(ci): spawn the differential's agent-device CLI as argv, not one option

thymikee:
## What broke

**Conformance Differential** has failed every night since 2026-08-25 ([run 32933263066](https://github.com/callstack/agent-device/actions/runs/32933263066)). All six scenarios reported `infrastructure-failed … maestro=pass agent-device=fail`, each preceded by:

```
/Users/runner/.../node: bad option: --experimental-strip-types src/bin.ts
```

`runAgentDeviceEngine` took the CLI as a single `cliPath` and spawned `[cliPath, ...args]`. But `AGENT_DEVICE_CLI` — set by the workflow, and this runner's own default — is a command *line*: a node flag plus the entry script. Node aborted on the combined token before the CLI loaded, so the oracle compared nothing at all. The lane was red but proving nothing.

Introduced by #1989, which extracted `engine-process.ts` and moved the agent-device side from `command.split(' ')` to `process.execPath` plus one path, to accept paths containing spaces. The maestro side kept splitting; the agent-device side lost it.

## The fix

Tokenize `AGENT_DEVICE_CLI` once, in `resolveAgentDeviceCliArgv` beside the spawn it feeds, and give `runAgentDeviceEngine` an argv array. Both properties now hold at the same time: a flag plus a script cannot collapse into one option, and a path containing spaces stays expressible — as a single array element.

`RunnerOptions.agentDeviceCli: string` becomes `agentDeviceCliArgv: readonly string[]`, so nothing downstream holds a string that is neither a path nor an argv. That ambiguity was the defect.

## Why no test caught it

The existing fixtures passed a bare `.mjs` path, which cannot tell an argv from a command line. The new test runs the workflow's own shape end to end — flag plus script. Verified to have teeth: reverting the fix reproduces the exact CI signature.

```
✖ agent-device execution spawns node flags as their own arguments
  node: bad option: --experimental-strip-types /tmp/.../cli.ts
```

## Verified

`pnpm maestro:conformance` (56/56), `pnpm lint`, `pnpm typecheck`, `pnpm format:check`, `pnpm check:fallow`.

Device evidence is pending: this lane needs a simulator, so I will dispatch **Conformance Differential** on this branch and post the result here.

## The other two lanes in the report — no change needed

**Mutation Weekly** ([run 32619691689](https://github.com/callstack/agent-device/actions/runs/32619691689), Aug 23) — already fixed on main. The `kernel-errors` shard died in Stryker's dry run on `no test file over the tripwire grows …`, all 26 pins off by one because `disableTypeChecks` prepends `// @ts-nocheck` to every sandbox copy. #1977 moved that test from `src/__tests__/` to `scripts/__tests__/`, outside `KERNEL_TEST_FILE_RE`, and landed at 05:39 UTC — 29 minutes after the weekly started at 05:10. Confirmed against the current tree: the test's address is out of the lane, its old one was in.

**Perf Nightly** — stale by design, not broken. #1822 parked it to `workflow_dispatch` under #1781 A3, because it writes a report and compares nothing. Aug 18 was its last scheduled run before the park. Worth teaching the freshness report about parked lanes so this stops reading as a gap.

## Left open

The differential lane was red for **nine consecutive nights** (Aug 18 → Aug 26) with nobody paged. The three nights before #1989 were real single-scenario reds — `tap-retry-if-no-change` recording zero tap retries (Aug 18), then `settle-after-tap` failing Maestro's `scrollUntilVisible home-open-form` (Aug 23-24, addressed by #1989's flow rewrite). Both look transient or already fixed, but #1989's rewrite has never actually executed: every run since it landed died on the argv bug. The dispatch on this branch is the first real exercise of it.

### DIFF
--- a/.github/workflows/conformance-differential.yml
+++ b/.github/workflows/conformance-differential.yml
@@ -32,7 +32,11 @@ concurrency:
   cancel-in-progress: true
 
 env:
-  AGENT_DEVICE_CLI: '--experimental-strip-types src/bin.ts'
+  # Entry script and node flags are separate on purpose: one variable cannot
+  # encode both without either corrupting a path that contains spaces or
+  # spawning the whole line as a single node option.
+  AGENT_DEVICE_CLI: 'src/bin.ts'
+  AGENT_DEVICE_CLI_NODE_FLAGS: '--experimental-strip-types'
   DIFFERENTIAL_ONLY: ${{ github.event.inputs.only || '' }}
   # CI should not phone home, and it keeps `maestro --version` to just the
   # version instead of prefixing an analytics notice.
--- a/packages/maestro/test/conformance/differential/engine-process.test.ts
+++ b/packages/maestro/test/conformance/differential/engine-process.test.ts
@@ -43,33 +43,24 @@ test('agent-device execution accepts a CLI path containing spaces', () => {
   }
 });
 
-// The regression the nightly ran into for two days: AGENT_DEVICE_CLI carries node
-// flags, and passing that line to node as one argument aborts before the CLI loads
-// ("bad option: --experimental-strip-types src/bin.ts"), which every scenario then
-// reports as an engine infrastructure failure rather than a divergence. Running the
-// workflow's own shape end to end is what pins it — a fixture path with no flag
-// cannot tell an argv from a command line.
-test('agent-device execution spawns node flags as their own arguments', () => {
-  const root = fs.mkdtempSync(path.join(os.tmpdir(), 'agent-device-engine-argv-'));
-  const cliPath = path.join(root, 'cli.ts');
-  try {
-    fs.writeFileSync(cliPath, 'const code: number = 0;\nprocess.exit(code);\n');
-    assert.deepEqual(
-      runAgentDeviceEngine(resolveAgentDeviceCliArgv(`--experimental-strip-types ${cliPath}`), []),
-      { engine: 'agent-device', outcome: 'pass', exitCode: 0 },
-    );
-  } finally {
-    fs.rmSync(root, { recursive: true, force: true });
-  }
-});
-
-test('the agent-device CLI argv is the workflow command line, tokenized', () => {
-  // The fallback is the source-CLI line the device workflows pass explicitly, so a
-  // hand run without the variable reproduces CI rather than looking for a dist build.
-  assert.deepEqual(resolveAgentDeviceCliArgv(undefined), [
+test('the CLI argv keeps node flags and the entry script apart', () => {
+  // The default is the source CLI the device workflows name explicitly.
+  assert.deepEqual(resolveAgentDeviceCliArgv(undefined, undefined), [
     '--experimental-strip-types',
     'src/bin.ts',
   ]);
-  assert.deepEqual(resolveAgentDeviceCliArgv('  '), ['--experimental-strip-types', 'src/bin.ts']);
-  assert.deepEqual(resolveAgentDeviceCliArgv(' bin/agent-device.mjs '), ['bin/agent-device.mjs']);
+  // An entry path is never split, so spaces in it survive.
+  assert.deepEqual(resolveAgentDeviceCliArgv('/tmp/agent device.mjs', ''), [
+    '/tmp/agent device.mjs',
+  ]);
+  assert.deepEqual(resolveAgentDeviceCliArgv('/tmp/agent device.mjs', undefined), [
+    '--experimental-strip-types',
+    '/tmp/agent device.mjs',
+  ]);
+  // Flags are split, which is exact because a node flag cannot contain a space.
+  assert.deepEqual(resolveAgentDeviceCliArgv('bin/x.mjs', '--no-warnings  --enable-source-maps'), [
+    '--no-warnings',
+    '--enable-source-maps',
+    'bin/x.mjs',
+  ]);
 });
--- a/packages/maestro/test/conformance/differential/engine-process.test.ts
+++ b/packages/maestro/test/conformance/differential/engine-process.test.ts
@@ -3,7 +3,11 @@ import fs from 'node:fs';
 import os from 'node:os';
 import path from 'node:path';
 import { test } from 'node:test';
-import { classifyAgentDeviceFailure, runAgentDeviceEngine } from './engine-process.ts';
+import {
+  classifyAgentDeviceFailure,
+  resolveAgentDeviceCliArgv,
+  runAgentDeviceEngine,
+} from './engine-process.ts';
 
 test('agent-device JSON distinguishes infrastructure from behavioral failures', () => {
   const result = (infrastructure?: true) =>
@@ -29,7 +33,7 @@ test('agent-device execution accepts a CLI path containing spaces', () => {
   const cliPath = path.join(root, 'agent device.mjs');
   try {
     fs.writeFileSync(cliPath, '');
-    assert.deepEqual(runAgentDeviceEngine(cliPath, []), {
+    assert.deepEqual(runAgentDeviceEngine([cliPath], []), {
       engine: 'agent-device',
       outcome: 'pass',
       exitCode: 0,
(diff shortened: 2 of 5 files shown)


thymikee:
## Device run: the fix works, and the lane immediately caught something else

Dispatched on this branch: [run 33044389760](https://github.com/callstack/agent-device/actions/runs/33044389760).

**The argv fix is confirmed.** No `bad option`, both engines executed on all six scenarios (`maestro=pass agent-device=pass` throughout), and the invariants actually evaluated — none of which has happened since Aug 25. Five of six scenarios are green.

**The run is still red on one scenario**, for an unrelated reason:

```
failed  settle-after-tap  maestro=pass agent-device=pass
  invariant violated: slowest tapOn took 3344ms (>= 2000ms)
```

### This is not a settle regression — it is the invariant losing its warm-up

Comparing the replay traces from the last green run ([Aug 17](https://github.com/callstack/agent-device/actions/runs/31997382890)) and this one:

| | Aug 17 (green) | Aug 27 (this branch) |
|---|---|---|
| step 2 `assertVisible` | 7080ms, 1 capture | 26533ms, 1 capture |
| step 3 `scrollUntilVisible` | 8928ms, **12 captures** | *(removed by #1989)* |
| measured `tapOn` | **860ms, 1 capture** | **3344ms, 1 capture** |

The tap does **one** hierarchy capture in both runs. The settle loop behaves identically; only the cost of a capture differs. This flow's `snapshotDiagnostics` read `p50 421ms, p95/max 22373ms` against a 1500ms slow threshold — the snapshot path is brutally cold at the start of a run and warms up over the first dozen captures.

So the 2000ms bound passed historically for one reason: `scrollUntilVisible` sat in front of the tap and spent 12 captures absorbing the cold start. #1989 removed that scroll — correctly, it was failing on Maestro's side — and with it the undocumented precondition the bound depended on.

Every other tap in the suite confirms it. Nothing regressed:

| tapOn | Aug 17 | Aug 27 |
|---|---|---|
| `wait-animation` step2 | 30284ms | 10034ms |
| `wait-animation` step4 | 1833ms | 1790ms |
| `tap-retry` step2 | 9282ms | 8291ms |
| `tap-retry` step3 | 2374ms | 2454ms |
| `tap-retry` step5 | 4777ms | 5375ms |

Taps on this lane routinely cost 2–30 seconds. Only `settle-after-tap` declares the duration bound, and it only ever held because it measured a warm tap.

### Note on the shape of the repair

Swapping the bound for a capture-count cap would be the obvious cost-independent fix, and it is **wrong**: a sleep-before-capture regression sleeps the full budget and then captures once, so the capture count stays 1 and only wall-clock can see it. The bound is load-bearing for the exact bug the scenario names. The repair has to restore a warm snapshot path before the measured tap, not change what is measured — and picking the warm-up (it takes captures, not elapsed time: 26.5 seconds of `assertVisible` warmed nothing, 12 captures did) is a fixture-shaped decision I would rather not guess at, at 13 minutes a validation.

Leaving that for a follow-up so this PR stays the argv repair. Also worth its own look: a **22-second** hierarchy capture against a 1500ms threshold.

thymikee:
P1: preserve the configured CLI-path contract at the owning config boundary. `resolveAgentDeviceCliArgv` now whitespace-splits `AGENT_DEVICE_CLI`, so an override such as `/tmp/agent device.mjs`—which previously remained one CLI path—becomes `/tmp/agent` plus `device.mjs`. The retained space-path test calls `runAgentDeviceEngine([cliPath], ...)` directly and therefore bypasses the production env -> `parseRunnerArgs` -> `runScenario` route; it cannot catch this regression. Make the environment contract unambiguous (for example, separate the entry path from Node flags or use a structured argv encoding), and add a regression through `parseRunnerArgs`/the runner route. The exact-head differential run proves the original combined-option bug is fixed, but it remains red on the non-infrastructure `settle-after-tap` timing invariant (3344ms >= 2000ms), which must be treated as owner-action evidence rather than green readiness.

thymikee:
## P1 addressed — and you were right that it was a regression in the other direction

Whitespace-splitting `AGENT_DEVICE_CLI` fixed the flags-plus-script shape and broke the one main handles: `/tmp/agent device.mjs` became two arguments. One variable cannot encode both, because any delimiter separating flags from the entry can also occur inside a path.

So the two concerns are now two variables that cannot be confused:

```
AGENT_DEVICE_CLI             the entry script — ONE path, never split
AGENT_DEVICE_CLI_NODE_FLAGS  node flags — split on whitespace, exact because a
                             node flag cannot contain a space
```

Defaults reproduce today's behaviour; the empty string runs an entry needing no flags. The workflow sets both explicitly.

### The regression now runs through the production route

`environment → parseRunnerArgs → runScenario → spawn`, as asked — calling `runAgentDeviceEngine` with a hand-built argv cannot see the environment contract at all. Two cases, because neither direction alone pins it. Verified against both broken implementations:

| implementation | `path containing spaces reaches the spawn unsplit` | `node flags stay separate arguments` |
|---|---|---|
| main (unsplit string) | ✔ passes | ✖ **fails** |
| my first attempt (whitespace split) | ✖ **fails** | ✖ **fails** |
| this PR | ✔ | ✔ |

One implementation detail worth flagging: the maestro stub deliberately stays *out* of the spaced directory. `runMaestroEngine` still splits its command on spaces, so a spaced stub path fails these tests for the other engine's reason — I hit that while writing them. That asymmetry is real but pre-existing and out of scope here.

### On `settle-after-tap` — agreed, owner-action evidence, not green readiness

I am not claiming this PR turns the lane green. Digging further into it since that comment, the picture changed and my earlier explanation was wrong in a way worth correcting:

- **I misattributed the 22-second capture.** `p50 421ms / max 22373ms` belongs to the *last* flow. `settle-after-tap`'s own stats are `count 3, p50 380ms, max 26371ms` — that 26s capture is the step-2 `assertVisible`, not the tap.
- **So cold capture cost does not explain the tap.** Its capture was one of the fast ones (~380ms), yet the step took 3344ms. Roughly 3 seconds went somewhere other than the capture — the same shape as #1874's finding that the slow outliers sit *ahead of* the first observation.
- **The bound is structurally wrong regardless.** `stepDurationBelow(tapOn, MAESTRO_DEFAULT_SETTLE_TIMEOUT_MS)` compares a whole step against a budget governing one phase inside it. The step is target resolution (a capture) + dispatch + settle, and `waitForTypedSnapshotStability` can itself overrun its deadline by one capture, since a capture starting before the deadline still runs to completion. It can only pass when the loop latches almost immediately *and* captures are fast — neither of which the invariant claims to require.

The sensible repair therefore is not a warm-up step and not a capture-count cap. It is to assert on the settle loop's own outcome, which the loop knows and currently discards: `waitForTypedSnapshotStability` returns `{snapshot, signature}` from both the latched path and the deadline path, so nothing downstream can tell them apart. Surface "did it latch / how long did it wait" per step in the replay trace and the invariant detects bug class 4 directly instead of inferring it from a timer — immune to machine speed, cold start and flow shape, and strictly stronger than today's bound, which a fast machine could satisfy while sleeping before every capture.

That spans the engine types, the daemon maestro adapter, the trace writer and this harness, so I would rather it be its own PR than ride along with an argv repair. Happy to open it, plus a separate one for the 26-second hierarchy capture.

thymikee:
Clean code review at `f693e9a06d6c3acf3b9154dc04c4ebb02b8ea062`: the prior P1 is resolved. Production-route regressions cover both node flags and a spaced entry path. `AGENT_DEVICE_CLI_NODE_FLAGS` remains whitespace-tokenized, so it cannot encode a flag argument containing spaces; this is non-blocking because no current producer needs that shape. CI and the exact-head conformance dispatch remain pending.

thymikee:
## Device run on the two-variable contract: it works

[Run 33052246798](https://github.com/callstack/agent-device/actions/runs/33052246798), first device exercise of `AGENT_DEVICE_CLI` + `AGENT_DEVICE_CLI_NODE_FLAGS`.

The CLI resolved and ran — no `bad option`, no module-not-found — and both engines executed on all six scenarios. Five green:

```
ok  percent-swipe                  invariant held: swipe executionProfile is endpoint-hold
ok  tap-retry-if-no-change         invariant held: tapOn tapRetries reached 1 (>= 1)
ok  optional-warned-not-failed
ok  wait-animation-between-taps
ok  optional-warned-scroll-and-wait
```

Still red on `settle-after-tap` — the known pre-existing invariant, unrelated to this PR:

```
failed  settle-after-tap  maestro=pass agent-device=pass
  invariant violated: slowest tapOn took 2710ms (>= 2000ms)
```

**2710ms this run against 3344ms last run, same commit-for-commit engine behaviour.** Identical step shape both times (`tapOn`, 1 capture), and the `assertVisible` ahead of it swung 26533ms → 9600ms. A 23% swing in the measured value across two runs of the same code is the bound measuring machine noise, not an engine property — and both values sit above 2000ms, so the check is red either way.

### Why I am not greening this check here

I looked at whether a `knownDivergence` waiver could cover it, since that mechanism exists to keep the lane green on a tracked gap. It cannot, by the harness's own self-test:

> `${scenario.id}: a signature where both engines pass describes no divergence`

Both engines pass here; only the invariant fails. The oracle is explicitly saying this shape must be repaired rather than waived — and the repair (surface `settled`/`waitedMs` from `waitForTypedSnapshotStability`, which today returns the same `{snapshot, signature}` from both the latched and the deadline path, so nothing downstream can tell them apart) spans the engine types, the daemon maestro adapter, the trace writer and this harness.

That belongs in its own PR, not bolted onto an argv repair. This one is complete as scoped: the lane executes again, and the argv/env contract is pinned in both directions.

PR #2071, opened 2026-08-27
refactor: move Android system observation out of daemon

thymikee:
## Summary

Move Android system-surface observation, escape recovery, and touch reference-frame mechanics behind one root-composed Android owner. Daemon policy still decides when to warn, continue, expire refs, or refuse an ANR; provider-owned sessions bypass local observation and mutation.

This is PR 2/3 in the #1739 terminal daemon-platform stack and removes 9 more concrete Android edges.

## Validation

Focused Android owner, occlusion, ref-frame, escape, touch, and provider-bypass tests pass. Adversarial review found no P0-P2 correctness issue after adding explicit provider touch-frame fallback coverage. Typecheck and the full layering suite pass on this stack head.

Scope is limited to the Android observation family. No HarmonyOS functionality changed, and no docs or skills changed.


thymikee:
P1: this hides Android mechanics from the daemon-edge scan rather than moving them to the platform owner. `src/platform-runtime-android-observation.ts` dynamically imports legacy `src/platforms/android/**` mechanics, is not registered in canonical `src/platform-runtime.ts`, and each daemon consumer constructs its own default adapter. `packages/platform-android` gains no observation owner. The provider-bypass test injects a fake and only proves the early return, so reverting the adapter back to direct legacy imports would not invalidate that ownership proof. Move the lazy implementation into the private Android façade/runtime, compose it through the canonical root, inject only the typed contract into daemon, and add planted-red structural coverage preventing a root adapter from becoming a second ungoverned platform implementation seam. Exact-head native/coverage checks are still in progress and no Android live evidence is recorded.

thymikee:
Re-reviewed exact head `17536a8859b6261bac5ab562d6d0c9677afc0186`: the only material delta is the eager-closure budget update/rebase. The existing P1 ownership finding remains unresolved, so this is not ready.

thymikee:
Addressed in 81c9208008 with terminal follow-up a1873f6d92. Android observation is owned by the platform-android package, root composition injects the typed observer through the governed platform-runtime-android-observation-host seam, daemon consumers no longer construct defaults, and R13 has planted-red coverage for hidden root/platform implementations. The provider harness mirrors root injection; layering and all 161 provider integration tests pass locally.

thymikee:
Re-reviewed the current head. Both prior P1s are resolved: raw Android imports moved behind the typed `platform-runtime-android-observation-host` adapter with explicit R13 coverage, and the provider-scenario integration harness now injects the root `androidObservation` dependency like daemon composition. The lifecycle mock update is consistent with that dependency change. I found no new code issue in this delta. Exact-head CI was restarted by the rewrite and is still queued/in progress, including repository guards and integration, so those remain pending evidence.

thymikee:
Fixed the Coverage and Compatibility & Provenance failures at the shared Android observation seam.

The migrated router/handler tests were still constructing request runtimes without `androidObservation`, so they reached the fail-closed fallback instead of the file-scoped Android mocks. The shared fixtures now inject the typed observation adapter, the transient empty-dump memo regression is covered, and the unused reset export / parser complexity provenance findings are removed.

Local validation on the exact middle head (`9dc7d40e8c`):
- full `pnpm test:coverage:ci` passed
- 12 previously failing files / 74 focused tests passed
- `pnpm check:affected --run` passed
- Fallow, typecheck, and format passed

thymikee:
Re-reviewed the rewritten head. The clean verdict remains valid: the Android adapter now preserves a previously successful dumpsys variant across a transient empty response with a focused regression, test-only observation fixtures complete the composition injection without changing production ownership, and the lifecycle mock follows the new dependency. I found no new code issue, so `ready-for-human` remains appropriate. Exact-head coverage, integration, guards, and native smoke are still running; this is not yet a merge-ready CI claim.

PR #2072, opened 2026-08-27
refactor: close the daemon platform boundary

### LINKED ISSUE #1739: refactor(daemon): platform-free daemon — broader migration waves (successor to #1696)
## Outcome

Continue [ADR 0019](https://github.com/callstack/agent-device/blob/main/docs/adr/0019-request-bound-platform-runtime.md) beyond the adoption
checkpoint until the daemon holds no platform behavior: every device command consumes
request-bound platform runtimes, cross-cutting platform policy lives behind owned facets, and
`src/daemon/` leaves the R3 platforms-seam allowlist. Migration follows the amendment (ADR 0019
sections 8–10): move-dominated size accounting, tiered evidence, one bind per handler,
evidence-gated startup recovery, two-phase shutdown.

This is the canonical tracker. Denominators and per-unit budgets are recorded here before each
wave starts; a wave's units are not split into worker issues before that.

Start condition: satisfied — the ADR 0019 amendment merged in #1738 (`73b54c15c`).

## Current state — 2026-08-21 (re-derived at `d0547dcb9`)

Derived from `main`, not from the hand-written wave plan. Since the previous checkpoint three
units landed on `main`: `focus` #1925 (R40), `type` #1935 (R41), and the deferred `find` #1944
(R35) — **Wave 4 is complete**, and the selector capture seam's legacy fallback is deleted
(`capture` is required; the observation-free duration wait is the one declared bound-less cell).
Find's mutating handler performs exactly one action-selected bind (ADR 0019 §9), enforced by
regression tests. Along the way one product fix shipped from CI forensics (the xdotool no-op-move
guard), and three Linux platform defects were filed with pixel evidence.

| Platform execution mode | Previous (`46eff36f8`) | Now (`d0547dcb9`) |
| --- | ---: | ---: |
| `none` | 17 | 17 |
| `inventory` | 1 | 1 |
| `device-runtime` | 24 | **26** |
| `legacy` | 34 | **32** |

The parametrized cutover table has 26 rows. Effective capability buckets remaining: **21**
(derive by evaluating the registry, not by grepping — trait spreads carry buckets). Production
daemon → platform dependency edges are unchanged by these three units (their retired leaves lived
in `core/`); the next edge reductions come from Wave 5 interaction and the perf family. The
terminal gate is therefore not satisfied and this tracker stays open.

Completed:

- Wave 0 substrate: #1740, #1745, #1741, #1742.
- Wave 1 lifecycle: `boot` #1747, `apps` #1756, `appstate` #1755, `shutdown` #1757.
- Wave 3: install/deploy family #1758; `open` / `prepare` / `close` / `runtime` #1759.
- Wave 4 descriptor cutover — **complete except the deferred `find`**: `snapshot` #1779, `diff`
  #1847, `viewport` #1864, `screenshot` #1878, `get`/`read` #1877 (R36, and the `read` entry
  deleted), `is` #1883 (R37), `wait` #1875 (R38). #1876 separately landed the behavior-neutral
  shared capture-input/admit-then-bind consolidation and split `handlers/find.ts`.
- Wave 5 prerequisite refactors: touch orchestration #1748 and typed selector stages #1744.
- Wave 5 unit 1: `focus` #1925 (R40); Wave 5 unit 2: `type` #1935 (R41) — each with live
  evidence for every claimed family and provider cell recorded on its PR.
- Wave 4 closure: `find` #1944 (R35) — action-selected single bind, the selector capture seam's
  legacy fallback deleted, five-leg live evidence. **Wave 4 complete.**
- Independent timeout cleanup: #1751.

### Remaining 35 `legacy` descriptors, by owner

| Owner | Count | Descriptors |
| --- | ---: | --- |
| Wave 2 (blocked on #1370) | 2 | `perf`, `trace` |
| Wave 5 interaction | 14 | `click` `fill` `type` `press` `longpress` `gesture` `scroll` `swipe` `back` `home` `hover` `orientation` `tv-remote` `keyboard` (`focus` complete via #1925) |
| Wave 6 named | 9 | `audio` `clipboard` `trigger-app-event` `alert` `settings` `react-native` `app-switcher` `doctor` `capabilities` |
| Wave 6 residue (unclaimed by any wave list) | 8 | `batch` `daemon` `debug` `events` `replay` `test` `web` `react-devtools` |

Two corrections this re-derivation forced:

- **`hover` was unassigned.** It is a genuine interaction command with a capability bucket, and the
  Wave 5 canonical list below omitted it. It is now recorded on Wave 5, which is why that row reads
  15 rather than 14.
- **The 8 residue descriptors carry no capability bucket.** They are legacy-by-delegation: the CLI
  route dispatches another command that executes the platform work, and the mode-dominance gate in
  `__tests__/platform-execution-cli-route.test.ts` is what keeps them `legacy`. Most should
  reclassify to `none` as their delegates migrate rather than need units of their own. That is an
  analysis task, not a migration; nothing may reclassify before its delegate lands.

### Remaining 61 daemon → platform edges, by owning unit

| Owner | Edges | Concentration |
| --- | ---: | --- |
| `perf` / `trace` (blocked) | 14 | `session-perf.ts` (4), `session-perf-legacy.ts`, `session-native-perf.ts`, `session-perf-xctrace.ts`, `types.ts` perf fields, part of `session-teardown.ts` |
| Wave 5 interaction | 9 | `android-system-dialog.ts` (3), `interaction-android-escape.ts`, `interaction-touch-reference-frame.ts`, `interaction-touch-response.ts`, `interaction-runtime.ts`, `selector-runtime.ts` |
| `doctor` | 8 | `session-doctor-{android,web,toolchain}.ts`, `session-doctor.ts` |
| `alert` / `audio` | 6 | `snapshot-alert.ts` (3), `session-audio.ts`, `audio-probe.ts` |
| Wave 6 residue | 24 | `request-platform-providers.ts` (6), `server/daemon-runtime.ts` (4), `device-ready.ts` (3), `snapshot-session.ts`, `apple-runner-options.ts`, `request-router.ts`, `request-recording-health.ts`, `session-device-utils.ts`, `install-source-resolution.ts`, `session.ts`, `session-test-infrastructure.ts` |

### Remaining critical-path work

- Wave 2 `perf` / `trace` remains sequenced after the next-major deletion in #1370 unless the
  tracker records the approved split escalation.
- Wave 4 `find` remains legacy and deferred. #1877 owns its `get text` leg and #1925 owns its
  `focus` leg; the direct `find type` execution path is the last blocker — the Wave 5 `type` unit
  clears it, then the atomic R35 cutover is honest. R35 stays reserved.
- Wave 4 shared capture debt remains and is still unassigned: the 10 production files under
  `src/daemon/snapshot-presentation/ios/`, `android-snapshot-freshness.ts`,
  `android-snapshot-timeout-evidence.ts`, `screenshot-overlay-android.ts`, and their
  observation/interaction call sites. They remain live because non-snapshot consumers still depend
  on them. **This is the one remaining block of work that is file-disjoint from the descriptor
  units**, so it is the natural parallel track; it must be assigned before Wave 5 migrates the
  freshness call sites in `selector-capture-runtime.ts` / `deferred-interaction-outcome.ts`.
- Wave 5 interaction descriptors and all Wave 6 closure units/residue remain.
- Before starting another unit, post its exact denominator, parity source, facet owner, expected
  deletions, evidence tier, and size budget here as required below. This checkpoint does not
  authorize an implicit multi-command migration.

Open critical-path PRs at this checkpoint: **none** — the critical path is idle.

**Immediate critical path:** Wave 5's remaining 13 interaction descriptors, in the recorded
cluster order (generic leaves → gestures → targeted touch, `fill` flagged for #1658 care), one
unit record per cluster before work starts. The generic-leaves cluster is scoped for a Sonnet
worker; gestures wait for the iOS scroll calibration churn to settle. The Wave 4 shared capture
debt remains the parallel-safe track and remains unassigned.

**Immediate critical path:** land #1935, then the deferred Wave 4 `find` R35 cutover — both of
find's direct execution legs (focus #1925, type #1935) now share bound runtimes, so the atomic
cutover is honest next. Do not start another descriptor
unit in parallel: every unit touches `registry.ts`, `runtime-command-cutover-table.ts`,
`contracts/platform-runtime-operations.ts` and each `platform-*/src/runtime.ts`, so two in flight
collide on the same four files. The Wave 4 shared capture debt is the one parallel-safe track.

## Denominator is derived, not hand-listed

The wave lists below are a plan; the completion denominator is derived and re-derived per wave
from two sources, so a command cannot be silently orphaned:

1. the descriptor registry — every descriptor whose explicit `platformExecution` discriminator is
   `legacy` (wave 0 made the discriminator exhaustive: `none | legacy | inventory |
   device-runtime`, with no default; an undeclared discriminator is now rejected);
   and
2. the platform dependency graph — every **production** `src/daemon/**` module with any dependency
   edge on `src/platforms/**` or a concrete `@agent-device/platform-*` package: static,
   **dynamic**, re-export, or type-only. Scope matches the layering scanner's existing exclusion
   of `*.test.ts` and `__tests__/`: daemon test files importing concrete platform implementations
   are outside the denominator and outside unit budgets (a unit rewrites the tests of the
   production modules it migrates as part of normal cutover work, but pre-existing
   integration-style test imports are not migration debt). R3 tolerates dynamic and type-only
   edges by design, and legacy daemon behavior reaches platforms through exactly those edges
   (e.g. `device-ready.ts`), so the R3 seam list is not a valid proxy for this graph.

Units are recorded by **canonical descriptor name**, with dispatch aliases, the public/internal
pairing where they differ (public `install-from-source` ↔ internal `install_source`, which owns the
daemon route), the exact handler owner file, and the cutover-gate row. Aliases (`read` → `get`,
`metrics` → `perf`) are surfaced on their canonical unit and are never separate units.

**Terminal gate (closes this tracker):** a dedicated planted-red layering rule rejecting every
dependency edge — static, dynamic, re-export, and type-only — from production `src/daemon/**`
modules (test files excluded, matching the layering scanner's scope) to `src/platforms/**` and to
concrete `@agent-device/platform-*` packages (daemon types come from contracts only); zero
descriptor capability buckets; zero descriptors with `platformExecution: legacy`; and
`src/daemon/` removed from the R3 platforms-seam allowlist. The daemon platform **branch/tag**
count reaches its floor rather than a mandated zero: per the accepted end-state ("ideally zero,
practically minimal"), remaining occurrences must each be enumerated and justified as
identifier-only — a platform name flowing through as opaque data or a display label, never
selecting behavior. Non-device daemon commands are outside the denominator because their
discriminator is `none`, not because a hand-written list omits them.

## Locked constraints (from the amendment — summarized, ADR text is authoritative)

- Move-dominated: root `src/` shrinks as packages grow; net shipped-size growth is exceptional and
  itemized. Every unit reports root-bytes-removed vs package-bytes-added plus the checkpoint's four
  size metrics against the `44c298d7f` baseline and its stack base.
- Deprecated surfaces are deleted on legacy at the next major, never migrated. Surface narrowing
  merges before its command's unit (perf's narrowing, #1731, is merged).
- Two evidence tiers: request-scoped units (facts + operations + parity cell table + cutover gate)
  vs durable-resource units (full ADR §4–5 lifecycle evidence). Importing durable machinery
  promotes the tier.
- One parametrized cutover gate; the four per-command policy files fold into it in wave 0a.
- Facts replace descriptor capability buckets; each unit deletes its bucket +
  `requireCommandSupported` wiring with the legacy adapter.
- One bind per handler for new units; side-effect-free facts inspection for
  admission/capabilities/doctor; one neutral `defineUse`; preferred operations require a recorded
  measurement.
- Cross-cutting facets land with their first consuming unit and have one contract owner.
- R3 platforms-seam allowlist narrows in the same unit that removes an area's last platform
  import; the list never grows.

### Wave 0 — structural substrate trim (behavior-neutral only)

Pure relocation/gate work; a change that alters any production admission, cleanup, or hint
behavior does not belong here.

- [x] Make the `platformExecution` discriminator explicit and exhaustive: `none | legacy |
      inventory | device-runtime`, remove the `{ kind: 'legacy' }` default at registry entry, and
      annotate every non-device descriptor `none`. Metadata-only; a registry gate rejects
      undeclared discriminators. This is what makes the derived denominator machine-readable.
- [x] Parametrized runtime-command-cutover gate; fold `device-inventory`/`logs-runtime`/
      `network-runtime`/`record-runtime` cutover-policy files into its table; one planted-red for
      the mechanism.
- [x] Consolidate `defineUse` (remove per-domain currying wrappers).
- [x] Absorb the `androidAdbExecutor` DI slot in `request-handler-chain.ts` into the existing
      provider-resolution mechanism, with wiring-equivalence tests.

Gateway facts inspection is **not** a wave-0 item: landing it with zero consumers would violate
the dead-code-clean substrate rule (and trip the fallow production-exports gate). It lands inside
the `boot` unit — its first consumer.

The `record`/`network` handlers keep their admission-use double binds. They were compliant when
they merged; ADR §9 retires the idiom for **new** units and permits — but does not require — a
retrofit as its own reviewed unit. Retrofitting two working handlers is consistency polish, not
denominator work, and stays off this tracker's critical path.

### Wave 1 — request-scoped device lifecycle units (ordered)

Order is a dependency chain, not a preference: `boot` lands the shared readiness facet the others
consume.

1. [x] `boot` — lands the device-readiness facet: `ensureReady` (covers simulator/emulator boot
       AND physical-device readiness, matching the legacy `ensureDeviceReady` breadth), headless
       boot as a fact, stopped-AVD-placeholder resolution moved to the Android inventory source,
       Apple runner-cache prewarm becomes platform-internal keep-hot policy (ADR 0002).
2. [x] `shutdown` — `shutdownTarget` operation. Eligibility comes from inspecting the
       `shutdownTarget` availability fact only; the handler never calls `ensureReady`, and an
       already-stopped target returns success immediately (legacy `target-shutdown.ts` parity —
       the migration must not boot a target in order to shut it down).
3. [x] `apps` — required use: `ensureReady` + `listApps` (complete use listed in the descriptor;
       readiness is not hidden inside another operation).
4. [x] `appstate` — required use: `ensureReady` + `appState`; sessionless path included in the
       denominator.

Deletions per unit: matching branches in `session-inventory.ts`, `session-state.ts`,
`device-ready.ts`, `target-shutdown.ts`, `session-doctor-app.ts`; capability buckets.

### Wave 2 — perf (durable tier)

**Current status:** not started; still blocked on the sequencing decision below.

Canonical descriptors: `perf` (whole descriptor, including its native-perf sub-surface routed via
observability; `metrics` alias surfaces here) and `trace`.

The deprecated aggregate forms (`perf` bare/`sample`/`metrics`) cannot remain an action-selected
legacy path inside a migrated `perf` descriptor — that would be a dual platform-execution path and
fail ADR 0019 §6. **Default resolution: sequence this unit after the next major (#1370) deletes
the deprecated forms.** That is the zero-machinery path — deletion retires their platform coupling
per §8, and the surviving narrowed surface migrates as one ordinary descriptor unit. Waves 3–4 do
not depend on perf and proceed in the meantime; the cost is only that the `SessionState` perf-field
cleanup waits.

Escalation only if the major slips and the perf coupling becomes blocking: a preparatory
behavior-neutral descriptor split (deprecated forms onto a separate internal legacy descriptor,
public surface unchanged) is acceptable **only** with wire-identity proof — the client sends the
full request including `command: 'perf'` over `agent_device.command`, so the wire command, session
events, and saved recorded actions must be byte-identical before and after (gate #1717/#1432 is
the proof vehicle), and the terminal gate then waits on the major deleting the split descriptor.

The surface narrowing (#1731) is already merged, so the denominator is the narrowed surface.

- Facets: one perf facet with distinct operations (frames, memory sample, native profile/trace),
  facts-declared availability; reuses capture-kit for durable captures.
- Deletions: `SessionState.applePerf` + `nativePerf.android` (R7 shrink in-unit),
  `session-perf-xctrace.ts`, `session-native-perf.ts` platform exclusivity, the `apple_perf` +
  `android_native_perf` teardown steps, the 12 branch sites in `session-perf.ts`.

### Wave 3 — install family, then open/close

**Current status:** complete via #1758 and #1759.

Canonical descriptors: `install`, `reinstall`, `push`, `install_source` (internal; owns the daemon
route — the public `install-from-source` surface and its capability bucket are recorded on this
unit); then `open`, `prepare`, `close`, and `runtime` (the runtime-hints setter — it shares the
runtime-hints facet this wave lands).

- Facets: install/deploy (the `session-deploy.ts` dispatch table moves fully behind it), open-target
  resolution/classification, runtime-hints (neutral key/value contract), close semantics.
- Cross-cutting landed here: test-IME as durable device state with marker-based recovery (startup
  orphan restore moves onto the evidence-gated pattern); macOS close-time alert dismissal inside
  the close unit; two-phase gateway shutdown lands with the first unit that deletes a
  daemon-runtime platform shutdown call.
- Deletions: `session-open*.ts` branch sites, `runtime-hints.ts` dispatch, `android_ime` teardown
  step, `restoreOrphanedAndroidTestImeOnDaemonStartup` daemon wiring.

### Wave 4 — observation

**Current status:** complete except `find`. `snapshot` #1779, `diff` #1847, `viewport` #1864,
`screenshot` #1878, `get`/`read` #1877, `is` #1883 and `wait` #1875 have all landed on `main`.
`find` is the only Wave 4 descriptor left. Its `get text` leg moved in #1877 and its `focus` leg
moves in the Wave 5 `focus` unit; R35 waits on the Wave 5 `type` unit.

Canonical descriptors: `snapshot`, `find`, `get` (alias `read`), `is`, `wait`, `screenshot`,
`viewport`, `diff`.

- Facets: capture, snapshot-freshness (Android staleness policy generalized to a hook),
  system-modal probe, overlay source-rect resolution.
- **Post-#1779 correction:** the shared presentation/freshness/overlay work below did not move with
  the command-atomic `snapshot` descriptor cutover because remaining observation/interaction
  consumers still own it. It remains open Wave 4 debt and must be assigned explicitly before those
  consumers migrate.
- iOS presentation rules (`src/daemon/snapshot-presentation/ios/`, originally 6 production files) relocate **inside the
  snapshot unit**, not as a standalone move: the files depend on root-only helpers (React Native
  overlay policy, scroll-indicator helpers, daemon tree helpers) whose owners must be decided
  first — snapshot-domain vocabulary moves to contracts/kernel or a snapshot domain package;
  `platform-apple` retains no root imports. No generic presentation-registration API while Apple is
  the only rule contributor: the rules sit behind the Apple capture implementation until a second
  platform earns the seam (no-platform-common rule).
- Deletions: `android-snapshot-freshness.ts`, `android-snapshot-timeout-evidence.ts`,
  `screenshot-overlay-android.ts`, freshness call sites in `selector-capture-runtime.ts` /
  `deferred-interaction-outcome.ts`, `snapshot-runtime.ts` branches.

### Wave 5 — interaction (largest)

**Current status:** prerequisite refactors merged; migration in progress, `focus` first.

Prework merged first as platform-neutral refactors: #1691 (touch orchestration split), #1656
(selector structural stages into typed policy).

Canonical descriptors: `click`, `fill`, `type`, `press`, `longpress`, `gesture`, `scroll`,
`swipe`, `back`, `home`, `focus`, `hover`, `orientation`, `tv-remote`, `keyboard`. `hover` was
added by the 2026-08-20 re-derivation: it holds a capability bucket and is a point-addressed touch
command, so the derived denominator surfaced it even though the original hand-written list omitted
it.

Order within the wave is a dependency chain where one exists: `focus` → `type` unblocks the
deferred Wave 4 `find` R35 cutover, because both of find's remaining direct execution paths are
those two commands.

- Facets: interaction dispatch; system-chrome guard + foreground-escape detection (cross-cutting,
  consumed by generic dispatch); tap-outcome corroboration; direct-selector fast path as the one
  measured `preferred` operation. Full ADR 0011 guarantee-matrix obligations per path.
- Deletions: platform sites in `interaction-touch.ts` (densest file), `interaction-*` handlers,
  `android-system-dialog.ts`, `direct-ios-selector.ts`, `selector-runtime.ts` branches,
  `androidBlockingDialogGuard` registry flag. Android snapshot-helper ownership moves fully into
  the Android platform module here (its daemon-shutdown reset call moves onto the two-phase stop).

### Wave 6 — closure (reaches the terminal gate)

**Current status:** not started.

Canonical descriptors with capability buckets not owned by earlier waves: `audio`, `clipboard`,
`trigger-app-event`, `alert`, `settings`, `react-native`, `app-switcher`. Each gets a normal unit
(most are small request-scoped facets; `audio` already receives an injected backend and may be
mostly bucket deletion).

Named units the derived denominator surfaces even though they carry no capability bucket:

- [ ] `doctor` — legacy session-route descriptor consuming platform behavior: per-family toolchain
      checks (`session-doctor-toolchain.ts` dispatch, `session-doctor-android.ts`,
      `session-doctor-web.ts` A-class handlers) become platform-contributed check lists behind a
      doctor facet; inventory-failure aggregation already consumes the migrated inventory gateway.
- [ ] `capabilities` — legacy session-route descriptor projecting support information; migrates
      onto side-effect-free facts inspection as its data source. Deliberately last among command
      units: it projects the union of every migrated command's facts, so it flips when the facts
      surface is complete. **Known, growing misprojection until then (folded into this unit per
      2026-08-21 decision — no separate tracker):** a migrated command loses its capability
      bucket, and the legacy projection reads "no bucket" as "supported everywhere", so a real
      Vega VVD already lists `snapshot diff get is wait focus` as supported when none are
      (observed live during #1925 validation). The inaccuracy grows by one command per migrated
      unit and is user-visible; it is the cost of correct sequencing, and this unit's acceptance
      evidence must include that VVD projection collapsing to the true fact-derived set.

Daemon-runtime residue, each with a named owner:

- [ ] Apple runner-lease state-dir configuration (`setRunnerLeaseOwnerStateDir`) — moves into the
      Apple module's process-lifetime configuration (two-phase shutdown participant).
- [ ] Web orphan recovery (`cleanupWebBrowserOrphansForDaemonStartup`,
      `cleanupManagedAgentBrowserOrphans`) — moves onto evidence-gated startup recovery under the
      web module.
- [ ] iOS simulator runner detach/stop (`detachIosSimulatorRunnerSessionsForShutdown`,
      `stopAllIosRunnerSessions`) — becomes the Apple module's two-phase shutdown implementation.
- [ ] Re-derive the denominator; enumerate and burn down any `src/daemon/**` platform dependency
      edge (static, dynamic, re-export, or type-only) not deleted by earlier waves
      (`request-lock-policy.ts` switches → runtime-declared lock metadata;
      `request-platform-providers.ts` roster → generic registration). The wave-0 discriminator
      pass will also classify commands the hand-listed waves don't name (`replay`, `test`,
      `events`, `debug`, `web`, …) — whatever it marks `legacy` lands here if no earlier wave
      claimed it.
- [ ] Flip the terminal gate: land the all-edge-kinds daemon layering rule with its planted-red
      case, remove `src/daemon/` from the R3 platforms-seam allowlist, and assert zero capability
      buckets and zero `platformExecution: legacy` descriptors. If the wave-2 descriptor split was
      taken, this item is blocked on the next major deleting the deprecated perf descriptor.

## Independent cleanups (off the critical path)

Platform *knowledge* without platform *imports* — not denominator work; do whenever convenient:

- [x] `daemon-client-timeout.ts` runs Apple-runner pkill patterns and emits Apple hint text on every
  local timeout regardless of session platform (string-level knowledge; `src/daemon/client/` is
  already outside the R3 seam). User-facing hint change; review with before/after fixtures.
- Optional `record`/`network` retrofit onto single-bind facts inspection (permitted by ADR §9 as
  its own reviewed unit; not required).

## Validation per unit

- `pnpm check:affected --run`; `pnpm check` for substrate/architecture changes.
- The parametrized cutover gate row, fact-coverage scenarios for every denominator cell, and the
  tier-appropriate evidence (parity cell table, or full durable lifecycle evidence).
- Size report: four checkpoint metrics vs baseline + stack base, root-bytes-removed vs
  package-bytes-added, itemized justification for any net growth.

## Before each wave

Record here: exact canonical descriptor list (with aliases, public/internal pairings, handler
owner files), denominator cells, the parity source (legacy behavior at which commit), facet
contract owner, expected deletions, and the unit budgets. No wave 2+ unit starts before wave 1 has
landed at least one unit proving the request-scoped tier's trimmed evidence in practice.

thymikee:
## Summary

Finish #1739's terminal daemon boundary: readiness and lifecycle mechanics, Android/Web durable cleanup, Apple runner resources, selector probing, and recording health now resolve through focused root owners. Production `src/daemon` has zero imports of `src/platforms` or concrete platform packages and has been removed from the R3 seam.

Add R65, a tracked-production AST gate that rejects static, dynamic, type-only, TS import types, re-exports, CommonJS/TS require forms, template literals, and statically constructed platform specifiers. The planted-red suite also proves comments, ordinary strings, tests, and non-daemon files do not create false positives.

Closes #1739.

## Validation

The terminal scan moved from 38 edges across 15 files to zero. Descriptor checks remain at zero `legacy` execution modes and zero production capability buckets. The full 196-test layering suite passes with `src/daemon` removed from R3.

Focused readiness, teardown, request-lock, recording-health, selector, session refresh, and sessionless cleanup tests pass (107 assertions across the final slice). Two Luna-max adversarial reviews found no lifecycle P0-P2 issues and drove closure of four syntactic gate bypass classes.

The complete affected run passed format, lint, typecheck, layering, fallow, build/package, Node integration, and provider integration. It reached 458/459 related files and 3,429/3,430 tests; the only failure was an unrelated Apple-runner concurrency test exceeding its 5-second timeout under full load, then passing alone in 149 ms. Native and coverage lanes remain GitHub-authoritative.

The complete stack touches 57 files and is +288 net production lines; most additions are typed owner seams and the terminal AST gate. No docs or skills changed because command behavior is unchanged. HarmonyOS perf remains unsupported and receives no new implementation.

### DIFF
--- a/packages/contracts/package.json
+++ b/packages/contracts/package.json
@@ -307,6 +307,10 @@
       "types": "./src/platform-runtime-operations.ts",
       "default": "./src/platform-runtime-operations.ts"
     },
+    "./platform-resource-cleanup": {
+      "types": "./src/platform-resource-cleanup.ts",
+      "default": "./src/platform-resource-cleanup.ts"
+    },
     "./platform-runtime-unavailable": {
       "types": "./src/platform-runtime-unavailable.ts",
       "default": "./src/platform-runtime-unavailable.ts"
--- a/packages/contracts/src/platform-resource-cleanup.ts
+++ b/packages/contracts/src/platform-resource-cleanup.ts
@@ -1,0 +1,21 @@
+import type { DeviceInfo } from '@agent-device/kernel/device';
+
+/** Platform-owned resource finalization invoked by neutral daemon orchestration. */
+export type PlatformResourceCleanup = Readonly<{
+  stopSnapshotHelper(device: DeviceInfo): Promise<void>;
+  closeManagedBrowser(
+    params: Readonly<{
+      device: DeviceInfo;
+      sessionName: string;
+      stateDir: string;
+      openSessionNames: () => readonly string[];
+    }>,
+  ): Promise<void>;
+  cleanupSessionlessExecutionHost(device: DeviceInfo): Promise<void>;
+  retainExecutionHostAfterClose(params: {
+    device: DeviceInfo;
+    shutdownRequested: boolean;
+    hasScreenRecording: boolean;
+    hasLease: boolean;
+  }): boolean;
+}>;
--- a/scripts/__tests__/test-file-size-ratchet.test.ts
+++ b/scripts/__tests__/test-file-size-ratchet.test.ts
@@ -34,7 +34,7 @@ const TRIPWIRE_LINES = 1_000;
 // Exact current lengths. Lower a pin when its file shrinks; never raise one — extract instead.
 const PINNED_TEST_FILE_LINES: Readonly<Record<string, number>> = Object.freeze({
   'src/__tests__/remote-connection.test.ts': 2973,
-  'src/daemon/handlers/__tests__/snapshot-handler.test.ts': 2284,
+  'src/daemon/handlers/__tests__/snapshot-handler.test.ts': 2242,
   'src/commands/interaction/runtime/settle.test.ts': 2359,
   'src/daemon/handlers/__tests__/session-replay-runtime-maestro.test.ts': 1963,
   'packages/platform-apple/src/runner/__tests__/runner-session.test.ts': 1957,
--- a/scripts/layering/check.ts
+++ b/scripts/layering/check.ts
@@ -19,7 +19,9 @@
 //     an import whose source zone outranks its target zone, plus a ratchet on the
 //     same inversion measured over TYPE-ONLY edges (R6).
 //   - Over the DAEMON only: SessionState field ownership (R7), because the session
-//     record is store-owned mutable state that any daemon module can write.
+//     record is store-owned mutable state that any daemon module can write; and the terminal
+//     concrete-platform boundary (R65), which rejects every import form into src/platforms or a
+//     platform package.
 //   - Over the TYPE GRAPH: the largest type-level import cycle is pinned by
 //     equality (R9). R4 keeps the value graph acyclic, so these cycles are free at
 //     runtime but bound what can be read in isolation; growth fails, and so does a
@@ -98,6 +100,7 @@ import { policyLead, policyViolation, ZONE_POLICIES } from './zone-policy.ts';
 import { contractsImplementationAuthorityViolations } from './contracts-implementation-policy.ts';
 import { selectorPipelineOwnershipViolations } from './selector-pipeline-ownership.ts';
 import { recordRuntimeRegistryJoinViolations } from './record-runtime-registry-policy.ts';
+import { checkDaemonPlatformBoundary } from './daemon-platform-boundary.ts';
 import { listTrackedProductionSources, listTrackedTypeScriptFiles } from './tracked-sources.ts';
 
 const repoRoot = execFileSync('git', ['rev-parse', '--show-toplevel'], {
@@ -485,7 +488,9 @@ function report(
         `inside its declared owner (R7); the largest type-level cycle is ${typeCycle} files ` +
         `(R9); ${daemonModularitySummary()}; ` +
         `${packageBoundariesSummary(repoRoot)}; ${platformPackagePolicySummary()}; ` +
-        `${runtimeCommandCutoverSummary()}; and bin.ts imports normalizeCliCommandAlias, ` +
+        `${runtimeCommandCutoverSummary()}; R65 keeps production src/daemon free of concrete ` +
+        `platform imports in every executable and type-only form; and bin.ts imports ` +
+        `normalizeCliCommandAlias, ` +
         `actually passes it into buildCommandUsageText, and holds no local alias literals ` +
         `(R12).\n`,
     );
@@ -543,6 +548,7 @@ export const LAYERING_RULE_IDS = [
   'type-spine-inversions',
   'session-state-ownership',
   'daemon-modularity-ratchets',
+  'daemon-platform-boundary',
   'bin-alias-fast-path',
   'package-boundaries',
   'platform-package-policy',
@@ -564,6 +570,8 @@ export const LAYERING_RULES: Readonly<Record<LayeringRuleId, LayeringRule>> = {
   'session-state-ownership': (context) => checkSessionStateOwnership(context.sources),
   'daemon-modularity-ratchets': (context) =>
     checkDaemonModularityRatchets(context.edges, context.typeCycleMembers),
+  'daemon-platform-boundary': (context) =>
+    checkDaemonPlatformBoundary([...context.sources].map(([path, source]) => ({ path, source }))),
   'bin-alias-fast-path': (context) => checkBinAliasFastPath(context.sources),
   'package-boundaries': () => checkPackageBoundaries(repoRoot),
   'platform-package-policy': (context) =>
--- a/scripts/layering/daemon-platform-boundary.test.ts
+++ b/scripts/layering/daemon-platform-boundary.test.ts
@@ -1,0 +1,292 @@
+import assert from 'node:assert/strict';
+import { test } from 'node:test';
+import {
+  DAEMON_PLATFORM_BOUNDARY_RULE,
+  daemonPlatformBoundaryViolations,
+  findDaemonPlatformDependencies,
+} from './daemon-platform-boundary.ts';
+
+const daemonFile = 'src/daemon/terminal-boundary-fixture.ts';
+
+function dependencies(source: string, file = daemonFile) {
+  return findDaemonPlatformDependencies([{ path: file, source }]);
+}
+
+function violations(source: string, file = daemonFile) {
+  return daemonPlatformBoundaryViolations([{ path: file, source }]);
+}
+
+test('R65 rejects static, type-only, dynamic, type, and re-export dependencies with source lines', () => {
+  const source = [
+    "import { old } from '../platforms/android/old.ts';",
+    "import type { OldType } from '../platforms/android/types.ts';",
+    "const lazy = import('../platforms/android/lazy.ts');",
+    "type ImportedType = import('../platforms/android/type.ts').ImportedType;",
+    "export { old } from '../platforms/android/re-export.ts';",
+    "export type { OldType } from '@agent-device/platform-android/types';",
+    "export * from '@agent-device/platform-apple';",
+  ].join('\n');
+
+  assert.deepEqual(
+    dependencies(source).map(({ kind, spec, line, target }) => ({ kind, spec, line, target })),
+    [
+      {
+        kind: 'static import',
+        spec: '../platforms/android/old.ts',
+        line: 1,
+        target: 'src/platforms/android/old.ts',
+      },
+      {
+        kind: 'type-only import',
+        spec: '../platforms/android/types.ts',
+        line: 2,
+        target: 'src/platforms/android/types.ts',
+      },
+      {
+        kind: 'dynamic import',
+        spec: '../platforms/android/lazy.ts',
+        line: 3,
+        target: 'src/platforms/android/lazy.ts',
+      },
+      {
+        kind: 'type import',
+        spec: '../platforms/android/type.ts',
+        line: 4,
+        target: 'src/platforms/android/type.ts',
+      },
+      {
+        kind: 're-export',
+        spec: '../platforms/android/re-export.ts',
+        line: 5,
+        target: 'src/platforms/android/re-export.ts',
+      },
+      {
+        kind: 'type-only re-export',
+        spec: '@agent-device/platform-android/types',
+        line: 6,
+        target: '@agent-device/platform-android/types',
+      },
+      {
+        kind: 're-export',
+        spec: '@agent-device/platform-apple',
+        line: 7,
+        target: '@agent-device/platform-apple',
+      },
+    ],
+  );
+
+  assert.deepEqual(
+    violations(source).map(({ rule, file, line }) => ({ rule, file, line })),
+    Array.from({ length: 7 }, (_, index) => ({
+      rule: DAEMON_PLATFORM_BOUNDARY_RULE,
+      file: daemonFile,
+      line: index + 1,
+    })),
+  );
+});
+
+test('R65 recognizes side-effect and aliased imports, including nested daemon paths', () => {
+  const source = [
+    "import '../../platforms/web/runtime.ts';",
+    "import { runtime as platformRuntime } from '@agent-device/platform-web/runtime';",
+  ].join('\n');
+
+  assert.deepEqual(
+    dependencies(source, 'src/daemon/handlers/terminal-boundary-fixture.ts').map((found) => ({
+      kind: found.kind,
+      line: found.line,
+      spec: found.spec,
+      target: found.target,
+    })),
+    [
+      {
+        kind: 'static import',
+        line: 1,
+        spec: '../../platforms/web/runtime.ts',
+        target: 'src/platforms/web/runtime.ts',
+      },
+      {
+        kind: 'static import',
+        line: 2,
+        spec: '@agent-device/platform-web/runtime',
+        target: '@agent-device/platform-web/runtime',
+      },
+    ],
+  );
+});
+
+test('R65 ignores comments, ordinary strings, unresolved dynamic imports, and lookalike names', () => {
+  const source = [
+    'const documentation = "import(\'../platforms/android/comment.ts\'); @agent-device/platform-android";',
+    "// import { ignored } from '../platforms/android/comment.ts';",
+    "const packageName = '@agent-device/platform-android';",
+    "const relativeName = '../platforms/android/not-an-import.ts';",
+    'const computed = import(platformSpecifier);',
+    "import '../platforms-sibling/not-platform.ts';",
+    "import '@agent-device/platforms';",
+    "import '@agent-device/platform';",
+  ].join('\n');
+
+  assert.deepEqual(dependencies(source), []);
+});
+
+test('R65 rejects require, import-equals, and template-literal type imports', () => {
+  const source = [
+    "const android = require('@agent-device/platform-android');",
+    'const runtime = require(`../platforms/android/runtime.ts`);',
+    "import web = require('@agent-device/platform-web');",
+    'type Apple = import(`@agent-device/platform-apple`).Apple;',
+    'type Android = import(`../platforms/android/types.ts`).Android;',
+  ].join('\n');
+
+  assert.deepEqual(
+    dependencies(source).map(({ kind, spec, line }) => ({ kind, spec, line })),
+    [
+      { kind: 'require', spec: '@agent-device/platform-android', line: 1 },
+      { kind: 'require', spec: '../platforms/android/runtime.ts', line: 2 },
+      { kind: 'import equals', spec: '@agent-device/platform-web', line: 3 },
+      { kind: 'type import', spec: '@agent-device/platform-apple', line: 4 },
+      { kind: 'type import', spec: '../platforms/android/types.ts', line: 5 },
+    ],
+  );
+});
+
+test('R65 folds statically constructed dynamic platform specifiers', () => {
+  const source = [
+    "const apple = import('@agent-device/' + 'platform-apple');",
+    'const android = import(`../platforms/${"android"}/runtime.ts`);',
+    "const wrapped = import((('@agent-device/' + 'platform-android')));",
+    'const required = require((`@agent-device/platform-web`));',
+  ].join('\n');
+
+  assert.deepEqual(
+    dependencies(source).map(({ kind, spec, line }) => ({ kind, spec, line })),
+    [
+      { kind: 'dynamic import', spec: '@agent-device/platform-apple', line: 1 },
+      { kind: 'dynamic import', spec: '../platforms/android/runtime.ts', line: 2 },
+      { kind: 'dynamic import', spec: '@agent-device/platform-android', line: 3 },
+      { kind: 'require', spec: '@agent-device/platform-web', line: 4 },
+    ],
+  );
+});
+
+test('R65 unwraps erased TypeScript expressions around executable specifiers', () => {
+  const source = [
+    "void import('@agent-device/platform-android' as string);",
+    "require('@agent-device/platform-web' satisfies string);",
+    "void import(<string>'../platforms/apple/runtime.ts');",
+  ].join('\n');
+
+  assert.deepEqual(
+    dependencies(source).map(({ kind, spec, line }) => ({ kind, spec, line })),
+    [
+      { kind: 'dynamic import', spec: '@agent-device/platform-android', line: 1 },
+      { kind: 'require', spec: '@agent-device/platform-web', line: 2 },
+      { kind: 'dynamic import', spec: '../platforms/apple/runtime.ts', line: 3 },
+    ],
+  );
+});
+
+test('R65 follows common createRequire and require aliases', () => {
+  const source = [
+    "import { createRequire as makeRequire } from 'node:module';",
+    "import * as moduleApi from 'node:module';",
+    "import moduleDefault from 'node:module';",
+    'const load = makeRequire(import.meta.url);',
+    'const loadAgain = load;',
+    "loadAgain('@agent-device/platform-android');",
+    "makeRequire(import.meta.url)('../platforms/web/runtime.ts');",
+    "moduleApi.createRequire(import.meta.url)('@agent-device/platform-apple');",
+    "moduleDefault.createRequire(import.meta.url)('@agent-device/platform-web');",
+    'const loadAlias = require;',
+    "loadAlias('@agent-device/platform-linux');",
+    "module.require('@agent-device/platform-vega');",
+    "const { createRequire: fromCjs } = require('node:module');",
+    "fromCjs(import.meta.url)('@agent-device/platform-harmonyos');",
+  ].join('\n');
+
+  assert.deepEqual(
+    dependencies(source).map(({ spec, line }) => ({ spec, line })),
+    [
+      { spec: '@agent-device/platform-android', line: 6 },
+      { spec: '../platforms/web/runtime.ts', line: 7 },
+      { spec: '@agent-device/platform-apple', line: 8 },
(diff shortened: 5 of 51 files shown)


thymikee:
P1: R65 proves zero direct import edges, but #1739’s terminal condition is broader: every surviving daemon platform branch/tag must be enumerated and identifier-only, never selecting behavior. This head still selects Android snapshot-helper cleanup and Web browser cleanup in `src/daemon/session-teardown.ts`, and Apple runner cleanup in `src/daemon/handlers/snapshot-session.ts`. R65 only scans import specifiers, so all three semantic dispatches pass; the PR supplies neither the required residual-branch inventory/justifications nor a semantic guard. Move the selection behind focused root-runtime/bound-runtime cleanup operations that accept neutral device/session input, then add planted-red proof for the residual semantic boundary (or an explicit identifier-only allowlist) before claiming terminal closure or closing #1739. The direct-edge gate itself is sound and green; #2071 also remains an open stack dependency.

thymikee:
Re-reviewed head d1c79712ebb9943de3fe42c303fd1691879e7345. The only material delta is the rebase/eager-closure budget update; the P1 semantic platform-selection finding remains unresolved (src/daemon/session-teardown.ts still selects Android/Web cleanup and src/daemon/handlers/snapshot-session.ts selects Apple cleanup). R65 is unchanged. This is not ready.

thymikee:
Addressed in a41a147e11. Android snapshot-helper, Web browser, Apple sessionless cleanup, and runner-retention selection now live behind the root-composed PlatformResourceCleanup contract; daemon teardown only orchestrates neutral operations. R65 guards cleanup orchestrators semantically and covers platform predicates, tags, destructuring, static/type/dynamic/re-export/require/createRequire/triple-slash imports. The planted daemon cleanup violation was observed red and the rebased layering gate passes.

thymikee:
Addressed in a1873f6d92. Android snapshot-helper, Web browser, Apple sessionless cleanup, and runner-retention selection now live behind the root-composed PlatformResourceCleanup contract; daemon teardown only orchestrates neutral operations. R65 guards cleanup orchestrators semantically and covers platform predicates, tags, destructuring, static/type/dynamic/re-export/require/createRequire/triple-slash imports. The planted daemon cleanup violation was observed red and the rebased layering gate passes.

thymikee:
[P2] Preserve synchronous cleanup methods in the unavailable fallback

unavailablePlatformResourceCleanup wraps every property in an async Proxy function, but retainExecutionHostAfterClose is deliberately synchronous and session-close-lifecycle-teardown branches on its boolean result without awaiting it. When platform-resource cleanup is unavailable, that call therefore returns a truthy Promise (and rejects asynchronously) instead of false, so the daemon can incorrectly retain the execution host and emit an unhandled rejection. Please replace the async-only Proxy with an explicit typed unavailable implementation (or otherwise preserve each method's sync/async contract) and add a missing-injection regression that exercises the synchronous retention decision.

thymikee:
Re-reviewed the rewritten head: the previously reported synchronous unavailable-cleanup P2 remains unresolved. The only delta is an unrelated daemon-runtime lifecycle-shutdown test mock; unavailablePlatformResourceCleanup still returns an async Proxy function for the synchronous retainExecutionHostAfterClose method. The requested typed fallback and missing-injection synchronous-retention regression are still required. Exact-head CI is queued/in progress.

thymikee:
Addressed in d5eb2f0dbc. The missing PlatformResourceCleanup fallback is now an exact typed object: its three async methods reject, while retainExecutionHostAfterClose throws synchronously. I added a production-router regression that omits the injection and proves close returns INTERNAL_ERROR before teardown or dispatch. I also fixed both prior coverage-shard failures: orientation now injects the existing Android observation fixture instead of reaching real ADB through a retired mock path, and the lock-serialization poll has a CI-safe timeout. Local validation: full pnpm test:coverage:ci passed (1,169 files, 8,718 tests, 90.57% lines), focused coverage tests passed, and format, lint, typecheck, and layering passed.

thymikee:
Follow-up coverage fix is now on `ae3cb6c9c5`. The remaining shard-1 failure was the test-level default 5s timeout racing the newly widened inner `vi.waitFor` budget; the serialization test now has an explicit 15s total budget while retaining 5s assertion waits. The exact regression passed locally in 724ms. The earlier full local `pnpm test:coverage:ci` run remains green (1,169 files / 8,718 tests; 88.59% statements, 80.63% branches, 91.68% functions, 90.57% lines). I am watching both replacement coverage shards on this exact head.

thymikee:
Final coverage follow-up is on `57b29e4595`. The second shard-1 failure showed the test was still waiting for `start-scroll` merely to release the scroll callback; under coverage load, screenshot had ended but that scheduling handshake could exceed 5s. I removed that circular handshake. Screenshot remains explicitly blocked, scroll completes immediately once it acquires the lock, and the final ordered trace plus `maxActive === 1` still proves cross-session serialization. Validation on this exact code: the focused test passes under V8 coverage (1.06s), and full `pnpm test:coverage:ci` passes 1,169 files / 8,718 tests at 88.60% statements, 80.63% branches, 91.68% functions, 90.58% lines.

thymikee:
The synchronous PlatformResourceCleanup fallback P2 is resolved on this head: the typed unavailable object preserves the sync/async method contracts, and the production-handler close regression would fail against the former async Proxy. However, this exact head is not ready because Coverage (1) fails in the changed request-router-screenshot serialization regression: scrollResponse.ok is false at line 281. The log does not expose the returned error, so this needs an exact-head reproduction or added assertion diagnostics followed by a green coverage run before readiness.

thymikee:
Fixed the remaining screenshot serialization failure. `scrollResponse.ok` was false because the test runtime omitted the newly required Android observation seam; the correction now lives in #2071's shared fixtures and this PR has been restacked on it (`b1ba32d5cf`).

Exact #2072-head validation:
- full `pnpm test:coverage:ci`: 1,169 files and 8,718 tests passed, plus fuzz worker
- coverage: 88.58% statements / 80.61% branches / 91.67% functions / 90.57% lines
- `request-router-screenshot.test.ts`: 16/16 passed, including same-device cross-session serialization
- isolated late-teardown reproduction suite: 17/17 passed

The full affected run passed every product assertion but encountered 10 late Vitest module-runner teardown errors under host contention after 3,760 passing assertions; the exact implicated suite passed immediately in isolation.

thymikee:
Re-reviewed the replacement head. The prior Coverage failure is addressed without weakening the serialization invariant: the lock test and its ordering assertions are unchanged, while router tests that do not exercise Android state now receive an explicitly benign observation fixture instead of accidentally reaching real ADB state. Platform-state tests still inject their behavior-specific fixtures, and production still composes the real adapter. I found no new code issue. Exact-head coverage and several CI/native lanes are still running.

PR #2070, opened 2026-08-27
refactor: move platform provider composition out of daemon

thymikee:
## Summary

Move request-scoped provider composition out of `src/daemon` and into root runtime composition. The daemon now consumes one neutral provider scope while exact-owner gating, wrapper order, inherited replay scope, request identity, and fail-closed provider behavior remain unchanged.

This is PR 1/3 in the #1739 terminal daemon-platform stack. It removes 15 of the 38 concrete-platform edges measured on `main` and also relocates the remaining neutral request/boot/install vocabulary to contracts.

## Validation

Provider routing, nested replay-scope, request-context, and fail-closed parity tests pass. Typecheck, layering, fallow, build/package, Node integration, and provider integration passed on the complete stack. The full affected run reached 3,429/3,430 related tests; its sole Apple-runner concurrency timeout passed in isolation in 149 ms.

57 files are touched across the complete stack; this PR is the provider-composition slice. No docs or skills changed because the public command surface and behavior are unchanged.

PR #2074, opened 2026-08-27
refactor(commands): one audience table for common input fields

### LINKED ISSUE #2027: Unify the input-field "audience" model (retiredField / AI SDK hidden / MCP admission)

thymikee:
## Summary

Closes #2027.

"Who may write this input field, on which surface" was expressed three times, each a separate
name-keyed mechanism: `retiredField()` in the command field maps, `ALWAYS_HIDDEN_FIELDS` in the AI
SDK adapter, and `OPERATOR_INPUT_GUIDANCE` / `CONFIG_LOADER_GUIDANCE` at the MCP admission boundary —
twelve hand-written refusal sentences keyed by name, far from the fields they govern.

The root cause was that the ~19 shared common fields existed only as parallel enumerations by name
(`commonProperties()`, `readCommonInput()`, `commonToClientOptions()`, and the `CommonCommandInput`
type), carrying no metadata. Any *policy* about a field therefore had to be stated in a new
name-keyed map somewhere else.

**Now there is one table.** `src/commands/common-input-fields.ts` declares each common field once:

```ts
daemonAuthToken: {
  schema: { type: 'string', description: 'Remote daemon auth token.' },
  read: (record) => optionalString(record, 'daemonAuthToken'),
  audience: ENV_OR_OPERATOR_CONFIG,
},
```

The JSON schema, the readers, the client-options projection, and the model-facing audience boundary
all derive from it, and `satisfies Record<keyof CommonCommandInput | 'target', …>` makes a row
without a type field — or a type field without a row — a compile error in both directions.

`audience` is the unified vocabulary (`src/commands/input-audience.ts`):

| audience | CLI / Node schema | MCP + AI SDK schema | supplied as a tool argument |
| --- | --- | --- | --- |
| *(absent)* — the model | advertised | advertised | accepted |
| `operator` | advertised | hidden | refused, with the operator path |
| `retired` | hidden | hidden | admitted, so the reader answers with migration guidance |

`retiredField()` now sets `audience: 'retired'`. Metro's `bearerToken` and `proxyBaseUrl` declare
`audience: 'operator'` *at the field*. `stateDir` declares it in the new
`src/mcp/tool-control-fields.ts`, next to the other MCP-only tool arguments, which also stopped being
enumerated three times (advertised schema / stripped-before-dispatch / AI SDK hidden set).

Refusal guidance is **rendered** from each declaration's operator path rather than written per key —
env var names through the existing `buildPrimaryEnvVarName`, the operator config file, or an explicit
sentence for the three keys neither describes. Every message is byte-identical to today's.
`OperatorInputSource` is shaped so a declaration that names *no* path at all does not typecheck.

### What this buys

Reclassifying `tenant` as operator-only is now this, and nothing else:

```diff
   tenant: {
     schema: { type: 'string', description: 'Remote tenant identifier.' },
     read: (record) => optionalString(record, 'tenant'),
+    audience: ENV_OR_OPERATOR_CONFIG,
   },
```

The schema, the readers, the projection, the MCP admission allowlist, its guidance string, and the AI
SDK schema all follow. In #2023 the operator/endpoint/path removal would have been a handful of row
edits instead of ~250 lines across three mechanisms.

The MCP deny side also became genuinely derived rather than hand-kept: `bearerToken` is refused on
*every* tool because the `metro` command declares it, not because someone remembered to add it to a
map in `command-tools.ts`. Deleting that one `operatorField(...)` now turns three existing assertions
red.

### Making the wiring unforgettable

A field-level `audience` only reaches the boundaries if its command's metadata carries it, so two
things close that gap structurally rather than by convention:

- `inputAudience` is **required** on `CommandMetadata`. A hand-rolled `defineCommandMetadata({…})`
  that omits it is a type error, not a silently model-writable credential.
- `defineFieldCommandMetadata` is the **only** construction path for a field-map command — it now
  takes an optional custom reader, so `batch` and `gesture` (which read their fields themselves) go
  through it too instead of assembling `fieldsInputSchema` + `readFieldInput` by hand. `click`,
  `press`, and `fill` were that helper inlined and now call it directly.

At the boundary, a command's own audiences are merged *before* the global operator classifications,
so an `operator` key always outranks a colliding per-command `retired` one — a name collision fails
closed on the refusal instead of open on the migration guidance.

### Batch step admission (P1 from review) — superseded by #2076

Review found that admission read only the **outer** keys of a tool call, so every operator-owned key
refused on the `snapshot` tool reached the command route by riding in as `batch.steps[].input`. I
reproduced it and fixed it here — and #2076 landed the same fix on `main` from a parallel session
while this was in flight.

**#2076's design is the one kept**, and this branch's version is removed entirely. It declares the
seam on the *schema* (`JsonSchema.commandInputFor` names the sibling holding the nested command's
name) rather than on each command's metadata, so the boundary walks the tool's own schema and
re-runs flat admission per nested input. That states the invariant as a parity — a step admits
exactly what the nested command's own tool admits — instead of a key list, and it needs no
per-command declaration to be kept in step.

What this leaves for review here is the #2027 refactor alone, with #2076's
`findInadmissibleNestedCommandInput` recursing through this branch's audience-derived
`findInadmissibleInput`. #2076's nested-admission suite passes against it unchanged, which is the
integration proof that the two compose: nested refusals are produced by the same audience map that
produces the flat ones.

## Not done, deliberately

The issue also lists `COMMON_COMMAND_SUPPORTED_FLAG_KEYS` as something to derive from the table.
That is a different axis — which flags the CLI *parser* accepts — and neither list contains the
other: 25 of its 42 keys (`remoteConfig`, `stateDir`, `daemonTransport`, `sessionIsolation`,
`leaseBackend`, `sessionLock`, every `provider*` and `aws*`) never become structured command input,
while the table's `cwd` and `debug` are not flags and its `deviceTarget` row is spelled `target`
there. Deriving it would mean 25 rows carrying no schema, reader, or projection. The reasoning is
recorded above the constant so the next reader does not have to re-derive it; the table does declare
`envFlagKeys`, which is the part the issue needed a flag key for.

## Validation

No wire or behavior change intended, and that is checked rather than asserted: I serialized
`listCommandTools()`, every CLI command schema, and every command `inputSchema` before and after and
diffed them — **byte-identical, property order included**. All twelve operator and config-loader
refusal messages were captured from the running executor and match the previous strings exactly.

Three independent adversarial reviews ran against the change — behavior parity, security boundary,
and issue fit. Parity found no non-equivalence. The security review confirmed the boundary is sound
and module init order is safe, and raised the merge-order and prototype-key hardening now folded in
above. The design review drove the one-construction-path change, the `flag-groups.ts` note, and the
test rebalance below.

Local gates green: `pnpm check:affected --run`, the full unit bundle (1063 files / 8105 tests),
`check:layering`, `check:fallow`, `check:daemon-wire-compat`, `check:mcp-metadata`,
`check:gate-manifest`, `check:agent-guidance`, `pnpm build`, `check:bundle-owner-files`,
`check:package`. `check:production-exports` reports 83 findings on this branch and the identical 83
on `origin/main` in a clean worktree — pre-existing, none in the touched files.

One of four full-bundle runs tripped the wall-clock slow-test gate on
`gesture-admission-parity.test.ts` (5.01s against a 2.5s budget). That test's eager import closure is
82 modules and contains none of the changed or added ones — not even `command-input.ts` — so this
change cannot affect its import cost; it passes in isolation and the other three bundle runs were
clean. Host variance, per the `AGENTS.md` rule for timeout-shaped failures.

Every new assertion was observed red before green, against planted defects rather than cited:

- drop `clientKey: 'target'` from the `deviceTarget` row → *"deviceTarget is dropped between reader
  and projection"*; remove a row's `read` → the same for that key.
- delete a row, or add one for a key with no `CommonCommandInput` field → `tsc` fails in both
  directions.
- remove `operatorField(...)` from metro's `bearerToken` → *"metro advertises bearerToken"*, plus the
  refusal on a non-metro tool losing its guidance.
- omit `inputAudience` from a hand-rolled command → *"Property 'inputAudience' is missing"*.
- declare an operator source with no env var and no config entry → does not typecheck.

The new test file pins the four guidance-rendering shapes and the claim that an operator
classification narrows only the model-facing surfaces — `commonProperties()` must keep advertising
those keys, or the CLI and Node client lose the operator's own way to pass them, which no MCP test
would catch. Which keys are operator-owned stays pinned by name in
`command-tools-operator-inputs.test.ts`, at the boundary that enforces it, and that file gains the
complementary assertion that the model-writable common keys are still advertised — so a
misclassification now fails in both directions.

No device evidence: the change is internal to input parsing and schema derivation, and every
observable surface is pinned identical above.

## Scope and residuals

26 files, +754/−471 production lines, but move-dominated. `command-input.ts` was 705 lines against
the 300-line target; its record readers move to `src/commands/input-readers.ts` so the table can use
them without an import cycle. It is now 476 — under the 500-line hard rule but still over target. The
remaining split (JSON-schema builders vs. field constructors) would separate two things every field
constructor pairs, and cost another eager-closure module, so I left it; the `src/cli.ts` closure pin
moves 365 → 368 for the three new leaf modules, with the reason recorded in the row.

`AGENTS.md` gains the new declaration site, and `docs/agents/cli-flags.md` step 4 now says which flags
have to declare an audience — worth having, since that is the step where a new credential-shaped flag
would otherwise reach a model-facing schema unclassified.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -53,6 +53,8 @@ Read the declaration rather than maintaining a prose copy:
 - interaction paths and guarantees: `packages/contracts/src/interaction-guarantees.ts`
 - canonical command names: `src/command-catalog.ts`
 - device capability admission: `src/core/capabilities.ts`
+- common command input fields, and which surface may write an input key (model, operator, retired):
+  `src/commands/common-input-fields.ts` and `src/commands/input-audience.ts`
 
 Shared selector parsing and matching belongs in `@agent-device/selectors`; request cancellation and
 progress in `src/request`; cross-layer contracts in `src/contracts`; CLI flags in
--- a/docs/agents/cli-flags.md
+++ b/docs/agents/cli-flags.md
@@ -17,7 +17,11 @@ mattering — threading it further is the common failure, not stopping too early
 3. `src/commands/command-projection.ts` and command-family projection helpers: write the input into
    the daemon request only if the flag affects daemon execution.
 4. `src/commands/*-command-contracts.ts`: add to the command input schema only if the option should
-   be available through Node.js or MCP as structured input.
+   be available through Node.js or MCP as structured input. An input key that names a credential,
+   an endpoint a credential is sent to, or operator infrastructure declares `operatorField(...)`
+   (`src/commands/command-input.ts`), which is what keeps the MCP and AI SDK tool schemas from
+   offering the model a parameter to write it into. One of the shared common keys declares the same
+   audience in its `src/commands/common-input-fields.ts` row instead.
 5. `src/client/client-types.ts`: update the public typed client option only when the Node.js
    interface exposes it.
 6. `src/client/client-normalizers.ts`: update daemon flag normalization only when the request still
--- a/src/__tests__/eager-closure-budgets.ts
+++ b/src/__tests__/eager-closure-budgets.ts
@@ -332,7 +332,12 @@ export const HUB_BUDGETS: Readonly<Record<string, number>> = Object.freeze({
   // (`resolveLocalDaemonCodeSignature`) rather than appearing here.
   // #2054 splits daemon cleanup and managed web backend into separate neutral contract entries;
   // the CLI already loads both command modules, so the second one-module contract is deliberate.
-  'src/cli.ts': 365,
+  // #2027 splits the 705-line `commands/command-input.ts` into the three leaf modules the
+  // common-field table needs to exist without an import cycle: `input-readers.ts` (record
+  // readers), `input-audience.ts` (who may write a key), and `common-input-fields.ts` (the table
+  // itself). Every command schema already evaluated all three concerns; the growth is three more
+  // module records for the same code, with no new subtree behind any of them.
+  'src/cli.ts': 368,
   'src/platform-runtime.ts': 47,
   'src/core/capabilities.ts': 73,
   'src/core/command-descriptor/registry.ts': 67,
--- a/src/ai-sdk/index.ts
+++ b/src/ai-sdk/index.ts
@@ -6,6 +6,7 @@ import type { AgentDeviceClient } from '../client/client-types.ts';
 import type { JsonSchema } from '../commands/command-contract.ts';
 import { resolveCommandFrameworkTier } from '../core/command-descriptor/registry.ts';
 import { createCommandToolExecutor, listCommandTools } from '../mcp/command-tools.ts';
+import { MCP_TOOL_CONFIG_KEYS } from '../mcp/tool-control-fields.ts';
 import { formatToolErrorText } from '../mcp/tool-error.ts';
 
 /**
@@ -53,22 +54,13 @@ export type AgentDeviceTools = {
 // `execute` forwards to the shared executor:
 //  - `session` is always pinned by this factory — the whole point is that a
 //    tool call can never target a session other than the one passed in.
-//  - `stateDir` selects which daemon state directory (and therefore which
-//    daemon/session namespace) a call resolves against. Left model-visible,
-//    it would let a call escape the pinned session into another daemon's
-//    state entirely, defeating that guarantee.
-//  - `mcpOutputFormat`, `includeCost`, `responseLevel` are MCP tool-config
-//    knobs, not command arguments; irrelevant here since `execute` below
-//    returns structuredContent directly and never reads a tool's rendered
-//    text, and shaping the response is this factory's decision, not the
-//    model's.
-const ALWAYS_HIDDEN_FIELDS = [
-  'session',
-  'stateDir',
-  'mcpOutputFormat',
-  'includeCost',
-  'responseLevel',
-] as const;
+//  - every MCP tool-config key: `stateDir` selects which daemon state directory
+//    (and therefore which daemon/session namespace) a call resolves against, so
+//    left model-visible it would let a call escape the pinned session into
+//    another daemon's state entirely; the rest shape the response, which is this
+//    factory's decision, not the model's (`execute` below returns
+//    structuredContent directly and never reads a tool's rendered text).
+const ALWAYS_HIDDEN_FIELDS: readonly string[] = ['session', ...MCP_TOOL_CONFIG_KEYS];
 
 // The registry's hand-rolled JsonSchema type and `jsonSchema()`'s expected
 // JSONSchema7 (re-exported from @ai-sdk/provider, not from `ai` itself) are
--- a/src/cli/batch-steps.ts
+++ b/src/cli/batch-steps.ts
@@ -10,7 +10,7 @@ import {
   BATCH_AVAILABLE_COMMANDS_HINT,
   readStructuredBatchCommandName,
 } from '../core/batch-policy.ts';
-import { assertAllowedKeys } from '../commands/command-input.ts';
+import { assertAllowedKeys } from '../commands/input-readers.ts';
 
 /**
  * The terminal half of the step-shape refusal. `@agent-device/contracts` states the shape for
--- a/src/commands/__tests__/input-audience.test.ts
+++ b/src/commands/__tests__/input-audience.test.ts
@@ -1,0 +1,89 @@
+import assert from 'node:assert/strict';
+import { test } from 'vitest';
+import {
+  COMMON_INPUT_AUDIENCE,
+  commonProperties,
+  commonToClientOptions,
+  readCommonInput,
+} from '../common-input-fields.ts';
+import { operatorInputRefusal } from '../input-audience.ts';
+
+// The guidance a refused operator key answers with is rendered from the key's
+// declared source, not written per key. These pin the four shapes the
+// declarations use, so a change to the template or to `buildPrimaryEnvVarName`
+// fails here instead of silently degrading every operator refusal at once.
+test('an operator refusal names the environment variable and the config key', () => {
+  assert.equal(
+    operatorInputRefusal('daemonAuthToken', { operatorConfig: true }),
+    'daemonAuthToken is not accepted as a tool argument. Set the AGENT_DEVICE_DAEMON_AUTH_TOKEN environment variable (or daemonAuthToken in ~/.agent-device/config.json) for the process serving these tools.',
+  );
+});
+
+test('an operator refusal names every environment variable that carries the value', () => {
+  assert.equal(
+    operatorInputRefusal('bearerToken', { envFlagKeys: ['metroBearerToken', 'daemonAuthToken'] }),
+    'bearerToken is not accepted as a tool argument. Set the AGENT_DEVICE_METRO_BEARER_TOKEN or AGENT_DEVICE_DAEMON_AUTH_TOKEN environment variable for the process serving these tools.',
+  );
+});
+
+test('a key with no environment variable points at the operator config file alone', () => {
+  assert.equal(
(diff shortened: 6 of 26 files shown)


thymikee:
P1: MCP/AI batch bypasses operator-input admission. `findInadmissibleInput` validates only outer tool keys; `batch` accepts arbitrary `steps[].input`, child metadata parses it, command flags project operator-owned paths, and the daemon invokes it. A model can nest `iosXctestrunFile` (or another operator key) in a batch step. Require shared recursive command-aware admission/schema filtering for batch steps or stop exposing batch, plus planted-red nested operator input. The common table/sizing rationale otherwise looks sound. This is not ready for human review.

thymikee:
Re-reviewed exact head `d979f73173421ec75e36679722d607fe919dc34a`. Only upstream merges changed it; there is no MCP/batch-admission delta. The previously posted P1 remains unresolved: operator-only input can still be nested in `batch.steps[].input` past the shallow outer admission boundary. This is not ready.

thymikee:
P1 fixed at `84e7144398`. Reproduced first: nested `iosXctestrunFile`/`cwd` reached `runCommand` intact on the previous head.

Admission now recurses, with the nesting declared by the command that nests rather than discovered by the boundary — `CommandMetadata.nestedCommandInputs` is required and `() => []` for every command that nests nothing, so a future command carrying another command's input has to state it instead of silently reopening the hole. Batch resolves a step's command name through the same helper its reader uses, so admission can't admit a step under one spelling the reader then accepts under another; `batch` isn't batchable, so nesting is one level.

A refused nested key answers with the same guidance as the top-level argument, prefixed by its step, and deny-by-default holds one level down too:

```
Batch step 2: cwd is not accepted as a tool argument. Start the process serving these tools in
the desired working directory, or pass absolute paths.
```

Planted red, both directions: reverting admission to the shallow boundary, **or** dropping batch's `nestedCommandInputs` declaration, turns the new nested-operator and nested-config-loader tests red. A step whose input is fully advertised still dispatches, so the check doesn't just refuse everything.

No schema change — `listCommandTools()`, the CLI command schemas, and every command `inputSchema` are still byte-identical to `main`, re-diffed after the fix. Only admission behavior moves. Full unit bundle green (1071 files / 8152 tests).

One residual I did not close: `nestedCommandInputs` is required on the contract, but `defineFieldCommandMetadata` defaults it to none, so the type forces a *hand-rolled* command to think about it while a field-map command still gets the safe default silently. Detecting "this schema accepts another command's input" mechanically looked more fragile than useful; happy to add a gate if you'd rather have one.

thymikee:
Correction to my previous comment: #2076 landed the same fix on `main` while this was in flight, and I've dropped mine in favour of it (`9b2a4ab556`).

#2076's design is better and is what's kept — declaring the seam on the schema (`commandInputFor` naming the sibling that holds the nested command's name) means the boundary walks the tool's own schema instead of each command declaring its nesting, and it states the invariant as a parity rather than a key list. Removed from this branch: `CommandMetadata.nestedCommandInputs`, batch's step extractor and step-command resolver, and my duplicate regression tests. `src/core/batch-policy.ts` and `src/commands/batch/metadata.ts` now differ from `main` only by this PR's own refactor.

The useful part is that the two compose: #2076's `findInadmissibleNestedCommandInput` recurses through this branch's audience-derived `findInadmissibleInput`, and its nested-admission suite passes unchanged — so nested refusals come from the same audience map as the flat ones, with no second filter to keep in step.

Surface re-diffed against `main` at `efa8e299ba`: `listCommandTools()`, the CLI command schemas, and every command `inputSchema` are byte-identical. Full unit bundle green (1072 files / 8155 tests); layering and fallow clean.

My apologies for the duplicated work — I spawned that investigation as a background task before you asked for the fix here, and didn't check for a sibling PR before implementing it myself.

thymikee:
[P1] Reconcile nested admission with the #2076 contract now on main

The nested MCP/AI admission bypass is closed in this implementation, but the PR's merge CI fails all three nested-admission regressions merged through #2076. Main's contract reports the stable location as `batch.steps[i].input`, while this branch replaces it with `Batch step N`, so the merge result is red despite the same security boundary being enforced. Rebase onto current main and keep one owning nested-admission path/diagnostic contract rather than carrying a parallel variant; then make the #2076 regressions and this branch's operator-input coverage agree and rerun Coverage (1).

thymikee:
Re-reviewed the replacement head. The #2076 contract conflict is resolved: this branch drops the parallel metadata walker, keeps the stable `batch.steps[i].input` diagnostic, and delegates nested admission through the nested command's existing advertised-schema boundary using the same structured batch-name resolver as execution. The #2076 parity/later-step/normalization regressions align with the returned path. I found no new code or drift issue. Exact-head type/package, integration, lint, guards, provenance, CodeQL, and macOS smoke are green; coverage and the remaining smoke lanes are still running.

thymikee:
Re-reviewed exact head 738bc1ce35 after the main reconciliation. The branch-owned audience-table refactor remains clean and now composes with #2076’s nested commandInputFor admission seam and the #2070–#2072 platform-boundary changes; it does not restore a parallel metadata walker. The field declaration → metadata/descriptor projection → MCP/AI admission route is coherent, regression tests exercise the owning mappings, and all substantive CI including both coverage shards, integration, smokes, guards, CodeQL, and size is green (only the expected cache matrix skip). Merge-ready.

PR #2076, opened 2026-08-27
fix(mcp): admit a batch step's input against the nested command's own schema

thymikee:
## Summary

An MCP/AI-SDK tool call could write operator-owned infrastructure paths into a `batch` step, which the admission boundary refuses on a flat call. The daemon then obeyed them per step.

`findInadmissibleInput` reads the **flat** keys of a tool call. Every tool schema is `additionalProperties:false`, so that is the whole boundary — for the keys a schema can *name*. A batch step's accepted keys depend on its sibling `command`, which JSON Schema cannot express, so `steps[].input` is declared free-form and the flat scan looked straight past it.

Everything the flat scan refuses was therefore admitted one level in:

```jsonc
// refused
{"tool": "snapshot", "input": {"iosXctestrunFile": "/attacker/run.xctestrun"}}

// admitted, before this PR
{"tool": "batch", "input": {"steps": [
  {"command": "snapshot", "input": {"iosXctestrunFile": "/attacker/run.xctestrun"}}
]}}
```

That is not inert. `readBatchDaemonStep` projects a step's input into per-step daemon request **flags**, and the daemon obeys them:

| key | daemon-side effect |
| --- | --- |
| `iosSimulatorDeviceSet` | selects the simulator device set `resolveTargetDevice` searches (`core/dispatch-resolve.ts`, `daemon/session-selector.ts`) |
| `iosXctestrunFile` | selects the `.xctestrun` the Apple runner launches (`runner-artifact.ts:resolveExternalXctestrunArtifact`) |
| `iosXctestDerivedDataPath`, `iosXctestEnvDir` | select the runner's derived-data and env directories (`daemon/apple-runner-options.ts`, `daemon/context.ts`) |

This is the operator-infrastructure write `OPERATOR_INPUT_GUIDANCE` exists to refuse, reached through the one input the model is invited to nest. The threat model is the one already written there: the model both reads untrusted app UI text and picks tool arguments, so a screen must find no parameter to write an infrastructure path into.

`cwd`/`stateDir`/`daemonBaseUrl`/`daemonAuthToken` were admitted too but are inert on this route — `cwd` rides `meta` (parent-only), `stateDir` is absent from `readCommonInput`, and the daemon base URL and token are read only by the client-side lifecycle, which has already run. They are refused now anyway, because the fix restores parity rather than enumerating the reachable subset.

### The fix: declare the seam, don't detect it

- `JsonSchema.commandInputFor` marks an object as **another command's input** and names the sibling property holding that command's name.
- `batchStepSchema` sets `commandInputFor: 'command'` on `steps[].input`, next to the `additionalProperties: true` that creates the opacity.
- `findInadmissibleNestedCommandInput` walks the tool's own schema against the raw arguments and re-runs `findInadmissibleInput` per nested input, against that command's advertised schema.
- The nested command name resolves through `resolveStructuredBatchCommandName`, extracted from `readStructuredBatchCommandName` so the readers and the boundary share one implementation. A step is normalized before it runs, so a boundary matching the raw value exactly would guard a different command than the one that executes — ` SNAPSHOT ` checked as nothing, run as `snapshot`.

No `if (name === 'batch')` anywhere — the walker only honors what a schema declares. Both model-facing surfaces are covered at once, since the MCP router and the AI SDK adapter share this executor.

Refusals keep the flat guidance and gain a location:

```
batch.steps[2].input: iosXctestrunFile is not accepted as a tool argument.
Set the AGENT_DEVICE_IOS_XCTESTRUN_FILE environment variable (or
iosXctestrunFile in ~/.agent-device/config.json) for the process serving these tools.
```

A step whose command cannot be resolved falls through untouched: it has no schema to check against and needs none, because the batch reader refuses that step before anything it carries reaches a flag. Answering here would bury the real error under a key complaint.

## Validation

Reachability was confirmed end-to-end before any fix was written, at four points: MCP admission admits the nested key; `prepareDaemonCommandRequest('batch', …)` emits it as a per-step flag; `runBatch` hands that flag to the per-step `invoke`; and `buildDeviceInventoryRequestFromFlags` / `buildAppleRunnerRequestOptions` turn it into a device-set path and runner paths. Re-running that same probe against this branch flips it: all eight keys refused nested, none admitted.

`command-tools-nested-admission.test.ts` states a **parity**, not a key list — for every batchable command, nested admission returns the same verdict as flat admission over a probe set spanning operator keys, config-loader keys, and an unknown key. A key added to `OPERATOR_INPUT_GUIDANCE` later is covered the day it lands, with nothing in the test to update. Reverting `command-tools.ts` fails it on the first operator key of the first batchable command. Plus: refusal at a non-zero step index, the unresolvable-command fall-through, and a legitimate five-step batch (`open`/`snapshot`/`tap`/`type`/`wait`, carrying `noRecord`, `session`, and a step `runtime`) proving valid input is untouched.

Local gates green at `54ab0b0222`: `pnpm test:unit` (1063 files / 8102 tests), typecheck, lint, `fallow audit`, `check:layering`, `check:daemon-wire-compat` (protocol unchanged), format. All 20 CI checks passed on that commit, Integration Tests and Coverage included.

`b8ec89ef35` then trimmed the `commandInputFor` JSDoc, which `JsonSchema` was publishing into `client-types.d.ts` at ~500 B — half the package's unpacked growth — to restate what the walker already documents. It changes no behavior.

`7dd8aa3f33` fixes review P1: admission matched the raw `steps[].command` while the reader normalizes it, so casing or surrounding whitespace skipped the check and still ran the step. It adds the casing/whitespace corpus and a drift guard stated against the reader — admission refuses a spelling iff the reader resolves it. Both new tests were confirmed to fail on an exact-match revert. Gates re-run green at that commit: `pnpm test:unit` (1063 files / 8104 tests), typecheck, lint, `fallow audit`, `check:layering`, `check:daemon-wire-compat`.

No device evidence is owed: the change is confined to the model-facing admission boundary and adds no platform behavior. Its effect on a device path is *refusing* input that previously reached one, which the parity test covers at the boundary. No `agent-device` sessions were opened.

## Notes for review

- **Touched files: 4** (3 source + 1 new test), ~230 insertions; +1.1 kB npm unpacked, against the 3 kB escalation threshold. Scope did not grow beyond the admission boundary and the one schema that declares the seam.
- **Conflicts with two open PRs**, both textual rather than semantic: [#2074](https://github.com/callstack/agent-device/pull/2074) (audience table) rewrites `command-tools.ts` admission and touches `command-contract.ts` and `batch/metadata.ts`; [#2067](https://github.com/callstack/agent-device/pull/2067) touches `batch/metadata.ts`. This bug predates #2074 and reproduces identically on it — its `findInadmissibleInput` is still flat, and it leaves the batch daemon-writer path untouched. Rebasing onto #2074 makes the fix smaller, not different: the per-step call becomes `toolInputAudience(metadata)`-driven like the flat one. Happy to sequence behind either.
- `commandInputFor` is emitted in the advertised MCP schema rather than stripped. It costs one key on one tool, JSON Schema ignores unknown keywords, and it documents why that object is free-form. Say the word if you would rather the published surface carry no internal markers, and I will filter it in `listCommandTools`.
- The walker descends `oneOf` branches as well as `properties`/`items`. No marker lives inside a `oneOf` today; the guard that makes this safe is that a branch only contributes a refusal when the value genuinely carries a known command name at the marked sibling, which means it *is* a nested command input.

### DIFF
--- a/src/commands/batch/metadata.ts
+++ b/src/commands/batch/metadata.ts
@@ -89,6 +89,12 @@ function batchStepSchema(nestedCommands: readonly string[]): JsonSchema {
       input: {
         type: 'object',
         additionalProperties: true,
+        // The accepted keys depend on `command`, so this object cannot be typed
+        // here. `commandInputFor` names the sibling that resolves them, which is
+        // what lets the model-facing admission boundary check a step's input
+        // against that command's own advertised schema instead of waving through
+        // an opaque object (`mcp/command-tools.ts`).
+        commandInputFor: 'command',
         description:
           'Structured command input for the nested command. Use the matching MCP tool schema for this object.',
       },
--- a/src/commands/command-contract.ts
+++ b/src/commands/command-contract.ts
@@ -16,11 +16,8 @@ export type JsonSchema = {
   maximum?: number;
   /**
    * Marks this object as ANOTHER command's input, naming the sibling property
-   * that holds that command's name. Such an object is free-form by necessity —
-   * its accepted keys depend on a value, which JSON Schema cannot express — so
-   * without the marker it is opaque to the model-facing admission boundary,
-   * which then checks nothing inside it. Declaring it here is what lets
-   * admission recurse with the named command's own advertised schema.
+   * that holds that command's name. Model-facing admission recurses into it
+   * with that command's own advertised schema (`mcp/command-tools.ts`).
    */
   commandInputFor?: string;
 };
--- a/src/commands/command-contract.ts
+++ b/src/commands/command-contract.ts
@@ -14,6 +14,15 @@ export type JsonSchema = {
   const?: unknown;
   minimum?: number;
   maximum?: number;
+  /**
+   * Marks this object as ANOTHER command's input, naming the sibling property
+   * that holds that command's name. Such an object is free-form by necessity —
+   * its accepted keys depend on a value, which JSON Schema cannot express — so
+   * without the marker it is opaque to the model-facing admission boundary,
+   * which then checks nothing inside it. Declaring it here is what lets
+   * admission recurse with the named command's own advertised schema.
+   */
+  commandInputFor?: string;
 };
 
 export type CommandMetadata<Name extends string, Input> = {
--- a/src/core/batch-policy.ts
+++ b/src/core/batch-policy.ts
@@ -59,12 +59,30 @@ export function normalizeBatchCommandName(command: unknown): string {
   return typeof command === 'string' ? command.trim().toLowerCase() : '';
 }
 
+/**
+ * The command a raw step value will RUN as, or undefined.
+ *
+ * Every caller that decides something about a step — the readers below, and the
+ * model-facing admission boundary in `mcp/command-tools.ts` — must resolve the
+ * name through THIS function, because a step is normalized before it runs:
+ * a caller matching the raw value exactly would see ` SNAPSHOT ` as no command
+ * at all, while the reader resolves it to `snapshot` and runs it. Admission
+ * checking one command while the daemon runs another is the whole failure, so
+ * the read below is this function plus an error rather than a second copy.
+ */
+export function resolveStructuredBatchCommandName(
+  command: unknown,
+): StructuredBatchCommandName | undefined {
+  const normalized = normalizeBatchCommandName(command);
+  return isStructuredBatchCommandName(normalized) ? normalized : undefined;
+}
+
 export function readStructuredBatchCommandName(
   command: unknown,
   stepNumber: number,
 ): StructuredBatchCommandName {
-  const normalized = normalizeBatchCommandName(command);
-  if (isStructuredBatchCommandName(normalized)) return normalized;
+  const resolved = resolveStructuredBatchCommandName(command);
+  if (resolved !== undefined) return resolved;
   throw new AppError(
     'INVALID_ARGS',
     `Batch step ${stepNumber} command is not available through command batch: ${String(command)}`,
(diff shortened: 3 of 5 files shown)


thymikee:
P1: recursive admission exact-matches the nested command, while batch later trims and lowercases it. A step command like ` SNAPSHOT ` therefore skips admission, is normalized to `snapshot`, and can retain operator path inputs into per-step flags. Resolve through the exact same batch normalization and allowlist before metadata lookup, and plant casing/whitespace red tests. Until then this is not ready.

thymikee:
P1 confirmed and fixed in `7dd8aa3f33`.

Reproduced first: ` SNAPSHOT ` passed admission untouched, and the reader then resolved it to `snapshot` and ran it with `iosXctestrunFile` still aboard. Exactly as described — casing or surrounding whitespace reopened the whole bypass.

The cause was that admission had its own notion of "which command is this" (`isCommandName` on the raw value) while the readers normalize first. Rather than teach the boundary to trim and lowercase too — a second copy that can drift again — I extracted `resolveStructuredBatchCommandName` from `readStructuredBatchCommandName`, so the readers and the boundary now share one implementation and the read is that function plus an error. Admission resolves through it, allowlist included, before the metadata lookup.

Tests: a red case over the casing/whitespace corpus (`snapshot`, ` snapshot`, `snapshot `, ` SNAPSHOT `, `SnApShOt`, `\tsnapshot\n`), plus a drift guard stated against the reader itself — admission refuses a spelling iff the reader resolves it — so a future change to step-name normalization cannot quietly reopen the gap. Both fail when the resolution is reverted to an exact match; I checked that rather than assuming it.

`pnpm test:unit` 1063 files / 8104 tests, typecheck, lint, layering, fallow, wire-compat all green locally on that commit.

thymikee:
Re-reviewed at `7dd8aa3f33f50a5aaf7a59b2aadab665135851fd`: clean and ready for human. The shared batch resolver now normalizes the nested command name consistently for both admission and execution, so casing or surrounding whitespace cannot bypass the boundary. The planted casing/whitespace corpus and reader-drift test close that bypass. All exact-head checks, including integration, coverage, guards, format, and size, are green.

PR #2077, opened 2026-08-27
fix(build): silence TS2883 dts noise in the Apple runner client shim

thymikee:
## Summary
- Every ubuntu `pnpm build` prints 18 non-fatal `TS2883` diagnostics for `src/platforms/apple/core/runner-client.ts` (tsdown 0.22.14's experimental TypeScript 7 dts API can't portably print the type it infers for each destructured re-export, since those types resolve through `@agent-device/platform-apple`'s private internal modules). The build still succeeds, but GitHub renders these as `##[error]` annotations on every CI run, burying real failures under 18 fake ones during triage.
- Replaces the single destructuring re-export (`export const { a, b, ... } = client;`) with 18 individually-typed statements (`export const a: AppleRunnerClient['a'] = client.a;`), giving dts emission an explicit, portable type to print for each export instead of having to infer and name one through private modules. This mirrors the identical pattern already used for the `AppleRunnerHost` port in [`packages/platform-apple/src/runner/host.ts`](https://github.com/callstack/agent-device/blob/main/packages/platform-apple/src/runner/host.ts).
- No behavior change: same object built once at module load, same bound function references, same export names — confirmed via the full local build + typecheck + the 39 test files (321 tests) that import or `vi.mock` this module.

## Test plan
- [x] `pnpm build` — zero `TS2883` lines in output, build still completes
- [x] `pnpm check:quick` (lint + full-repo typecheck) — green
- [x] `oxfmt --check` on the changed file — clean
- [x] Targeted `vitest run` across the 39 test files that import/mock `core/runner-client` — 321/321 passing
- [x] Adversarial multi-angle code review (correctness, cross-file consumers, reuse/simplification/efficiency, altitude, conventions) — no findings

### DIFF
--- a/src/platforms/apple/core/runner-client.ts
+++ b/src/platforms/apple/core/runner-client.ts
@@ -1,4 +1,7 @@
-import { createAppleRunnerClient } from '@agent-device/platform-apple/runner/client';
+import {
+  createAppleRunnerClient,
+  type AppleRunnerClient,
+} from '@agent-device/platform-apple/runner/client';
 import { appleRunnerHost } from './runner-host.ts';
 
 /**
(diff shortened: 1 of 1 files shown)


thymikee:
Code review clean at `34df71a880567c82d60ea3808f0b539d3ac424a8`:

The explicit façade-based `AppleRunnerClient[...]` annotations remove only the non-portable declaration-inference path. They preserve one-time construction and the exact exported function references; no runtime route changes. CI is pending. Ready for human review.

PR #2083, opened 2026-08-27
fix(test): settle the daemon's exit on its own deadline in the web shutdown lane

thymikee:
## What

The daemon-exit assertion in the #1868 web shutdown lane (`live web platform e2e daemon-shutdown browser cleanup`) was an **instantaneous** liveness read placed right after `settleManagedBrowserProcesses`:

```ts
const after = await settleManagedBrowserProcesses(status);   // polls, 45s deadline
assert.equal(after.count, 0, ...);
assert.equal(isProcessAlive(daemonPid), false, ...);         // no polling at all
```

It silently borrowed the Chrome fleet's poll deadline for a condition it never polled, and passed only while the fleet happened to outlast the daemon.

Both endings now settle concurrently from the same SIGTERM, each against its own deadline, with no ordering asserted between them.

## Why

Nothing enforces that order. The daemon awaits the `agent-browser close` CLI's **return**, never the disappearance of the Chrome pids that call reaps, and awaits it under a teardown-budget race (`WEB_BROWSER_SESSION_TEARDOWN_BUDGET_MS` + `DAEMON_SESSION_TEARDOWN_TIMEOUT_MS`) it walks away from on expiry. Either can finish first, and both orders have been observed:

- **macOS local:** fleet pids linger ~13s, daemon exits first — the race is hidden entirely.
- **CI:** fleet reaped in ~4.8s, daemon still alive → `expected the daemon process itself to have exited after SIGTERM`. Seen once on #2074; re-running the same job on the same commit passed.

The guarantee #1868 is actually about — the fleet closed *because the daemon closed it*, not because agent-browser's idle timer beat the test — is held by `WEB_SHUTDOWN_IDLE_TIMEOUT_MS`, not by any ordering between these two. So the ordering is asserted nowhere, on purpose, and the comment records why.

## Notes for the reviewer

**`settleDaemonExit` waits out the daemon's *identity*, not the bare pid.** A pid the host recycles inside the 45s window is a different process, and reading it as "the daemon is still alive" would fail this lane for the one reason it must not — a false red. It polls `isAgentDeviceDaemonProcess(pid, startTime)` (already exported; this file already imports from that module, and the sibling test at the top of the file exists for the same hazard). That also makes `WEB_SHUTDOWN_SETTLE_POLL_MS` the right interval, since the identity probe shells out to `ps` like the fleet sweep. No new constant, no new machinery, no new module edge — a structural sibling of `settleManagedBrowserProcesses` sharing both existing constants.

**No fixed sleep.** The 45s is a failure bound, not a wait; the happy path returns as soon as the daemon is gone. There is no event-based alternative: the daemon is not a child of the test process (the CLI spawns it detached and exits), Node exposes neither kqueue `NOTE_EXIT` nor `pidfd`, and every daemon artifact (lock file, `daemon.json`, shutdown report) is written *before* `exit()` — so watching one would prove the daemon *decided* to exit, which is the weaker proxy claim this bug came from in the first place.

**Residual limits, stated rather than hidden:** the 45s is derived from the daemon's 35s web-session teardown budget and must grow with it; a zombie daemon would still read as alive (documented at `src/utils/host-process.ts:64`), unreachable here because the reparented daemon is reaped by launchd/init.

## Verification — planted red, then green

Plant: `await sleep(N)` immediately before `exit()` in `src/daemon/server/daemon-runtime.ts` (reverted; not in this diff).

| Daemon | Assertion | Result |
| --- | --- | --- |
| `sleep(25s)` | old | ✖ fails at 18.2s — `expected the daemon process itself to have exited after SIGTERM`, the exact CI message |
| `sleep(25s)` | new | ✔ passes at 40.6s — waits it out instead of reading once |
| `sleep(120s)` | new | ✖ fails at the deadline — `still alive 45128ms later` |
| unmodified | new | ✔ ✔ 13.4s / 13.5s — unchanged from before the fix |

Run with `AGENT_DEVICE_WEB_E2E=1 pnpm test:smoke:web`. Healthy-path wall clock is unchanged because the two settles run concurrently. `pnpm lint` and `pnpm typecheck` both clean.

Test-only change; #2074 was not touched.

### DIFF
--- a/src/daemon/__tests__/daemon-exit-wait.test.ts
+++ b/src/daemon/__tests__/daemon-exit-wait.test.ts
@@ -1,0 +1,130 @@
+import { afterEach, beforeEach, expect, test, vi } from 'vitest';
+import { stopProcessForTakeover, waitForDaemonExit } from '../daemon-process.ts';
+
+const DAEMON_COMMAND = '/opt/checkout/dist/src/internal/daemon.js';
+const OURS = 'Mon Aug 24 10:00:00 2026';
+const RECYCLED = 'Mon Aug 24 10:00:07 2026';
+const PID = 4242;
+// Small enough that a regression that waits the budget out still lands inside the
+// unit lane's wall-clock gate and fails on its assertion rather than on the clock.
+const TIMEOUT_MS = 1_000;
+const POLL_MS = 5;
+
+// A live process cannot be made to change identity mid-wait, so the host reads are
+// mocked: a recycled pid is the same live pid reporting a different start time,
+// which is what the host presents after reuse.
+const state = vi.hoisted(() => ({
+  alive: new Map<number, boolean>(),
+  starts: new Map<number, string>(),
+  states: new Map<number, string>(),
+  commands: new Map<number, string>(),
+}));
+
+vi.mock('../../utils/host-process.ts', async (importOriginal) => ({
+  ...(await importOriginal<typeof import('../../utils/host-process.ts')>()),
+  isProcessAlive: (pid: number) => state.alive.get(pid) ?? false,
+  readProcessStartTime: (pid: number) => state.starts.get(pid) ?? null,
+  readProcessCommand: (pid: number) => state.commands.get(pid) ?? null,
+  readHostProcessIdentityObservations: (pids: Iterable<number>) => {
+    const observations = new Map<number, { state: string; startTime: string }>();
+    for (const pid of pids) {
+      const startTime = state.starts.get(pid);
+      if (startTime === undefined) continue;
+      observations.set(pid, { state: state.states.get(pid) ?? 'S', startTime });
+    }
+    return observations;
+  },
+}));
+
+/** Delivered signals only; a `0` probe is a liveness read, not a write. */
+const signals: NodeJS.Signals[] = [];
+let onSignal: (signal: NodeJS.Signals) => void = () => {};
+
+beforeEach(() => {
+  state.alive.clear();
+  state.starts.clear();
+  state.states.clear();
+  state.commands.clear();
+  signals.length = 0;
+  onSignal = () => {};
+  state.alive.set(PID, true);
+  state.starts.set(PID, OURS);
+  state.states.set(PID, 'S');
+  state.commands.set(PID, DAEMON_COMMAND);
+  vi.spyOn(process, 'kill').mockImplementation(((pid: number, signal: NodeJS.Signals | 0) => {
+    if (pid !== PID || signal === 0) return true;
+    signals.push(signal);
+    onSignal(signal);
+    return true;
+  }) as typeof process.kill);
+});
+
+afterEach(() => {
+  vi.restoreAllMocks();
+});
+
+test('waitForDaemonExit reports a pid recycled mid-wait as exited, without burning the deadline', async () => {
+  setTimeout(() => state.starts.set(PID, RECYCLED), 20);
+  const wait = await waitForDaemonExit(
+    { pid: PID, startTime: OURS },
+    { timeoutMs: TIMEOUT_MS, pollMs: POLL_MS },
+  );
+  expect(wait.exited).toBe(true);
+  expect(wait.elapsedMs).toBeLessThan(TIMEOUT_MS / 2);
+});
+
+test('waitForDaemonExit reports a daemon that keeps its identity as not exited', async () => {
+  const wait = await waitForDaemonExit(
+    { pid: PID, startTime: OURS },
+    { timeoutMs: 40, pollMs: POLL_MS },
+  );
+  expect(wait.exited).toBe(false);
+});
+
+// A daemon killed by its own parent lingers as a zombie: kill(pid, 0) still succeeds
+// and the start time still matches, while ps reports the command as `<defunct>`.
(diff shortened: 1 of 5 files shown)


thymikee:
[P1] Put identity-aware daemon exit waiting at the daemon-process owner and reuse it here.

The new test-local 45s poll correctly treats the original daemon identity disappearing as success, but duplicates lifecycle policy beside the owner it already imports. More importantly, `stopProcessForTakeover` checks identity only before SIGTERM, waits on the bare pid, and can then SIGKILL a recycled pid; the sibling daemon-stop path already re-verifies before escalation. Add a typed `waitForDaemonExit(identity, timeout/poll)` under daemon-process, consume it from this smoke and the stop/takeover paths, re-verify before any escalation, and add deterministic planted-red coverage for the pid identity changing during the grace wait. Replace the paragraph-long test rationale with the owning interface and invariant made obvious in code.

thymikee:
Addressed in `029a10dc`, and you were right that the owner was the weaker of the two.

**`waitForDaemonExit(identity, { timeoutMs, pollMs })` now lives in `daemon-process.ts`** and is consumed by the web shutdown smoke, `stopProcessForTakeover` and `stopDaemon`. The test-local helper and its paragraph of rationale are deleted — the smoke test's local identity type is now the owner's `DaemonProcessIdentity` too.

**The takeover defect is fixed.** Rather than re-verifying at each call site, `signalDaemonIdentity` re-reads identity immediately before every signal, so a recycled pid cannot be signaled at all — escalation safety is a property of the call instead of a rule to remember. `stopDaemon` kept its explicit pre-SIGKILL guard (its throwing signal semantics matter there) and additionally stops misreporting a recycled pid as "did not exit".

**One thing your review surfaced that I had wrong too.** Waiting on identity alone regressed `daemon-process-takeover.test.ts`. A daemon reaped by its own parent lingers as a zombie: `kill(pid, 0)` still succeeds and the start time still matches, but `ps` reports the command as `<defunct>` — which my first pass classified as a recycled pid and handed the number back while it was still taken. `classifyDaemonPid` now names all four states, and the zombie counts as neither ending, so the wait keeps polling and callers keep the guarantee they had before. That is the reason the classifier exists rather than a one-line predicate.

**Coverage** (`src/daemon/__tests__/daemon-exit-wait.test.ts`, host reads mocked since a live process cannot be made to change identity mid-wait): identity changing mid-wait, a zombie waited through to its reap, no SIGKILL after a recycled grace wait, and escalation still reached for a daemon that genuinely survives SIGTERM. Planted-red verified — reverting `stopProcessForTakeover` to the bare-pid wait fails the recycled-pid test alone, on its assertion, inside the unit lane's clock budget. Budgets are injected so a regression fails on the assertion rather than on the slow-test gate.

Branch rebased onto `main`. `pnpm typecheck`, `pnpm lint`, `src/daemon` (2341 tests) and the live `AGENT_DEVICE_WEB_E2E=1 pnpm test:smoke:web` lane all pass. The one red in `src/daemon` is `gesture-admission-parity.test.ts` tripping the wall-clock gate; it is not in this diff and is over budget in isolation on an idle checkout (3.2s–5.0s depending on load).

thymikee:
[P2] Remove the new implementation-history/control-flow comments. Current main’s AGENTS.md rejects comments that narrate a workaround or preserve review history, yet the daemon-process helper/type blocks, daemon-exit-wait test comments, and edited smoke comment still do exactly that. The identity type, classifier, one owning wait helper, and planted-red assertions now make the invariant evident in code; retain only necessary public API documentation. The implementation and exact-head CI/live-web evidence are otherwise clean.

thymikee:
Removed in `f6bf3e23` (branch also rebased onto `4454aef1`).

Every comment this PR added is gone except four, all of which the rule permits:

- `DaemonProcessIdentity` and `DaemonExitWait.exited` — docs on exported types.
- `waitForDaemonExit` — the exported helper's contract (what counts as having left the host).
- A two-line citation on `classifyDaemonPid` for the one thing the code cannot encode: a terminated pid awaiting reap answers `kill(pid, 0)`, keeps its start time, and reports its command as `<defunct>`.

Dropped: the `DAEMON_EXIT_POLL_MS` cost note, the `DaemonPidState` paragraph, the `signalDaemonIdentity` rationale, the mock/planted-red narration in `daemon-exit-wait.test.ts`, and the settle comment I had edited in the smoke test. The pre-existing smoke-test comments are untouched — they are not part of this change. `git diff origin/main...HEAD -U0 | grep '^+.*//'` returns only the four blocks above.

Verified at `f6bf3e2380ad3adfc4852682c273c76f88a16c39`: `pnpm typecheck`, `pnpm lint`, `pnpm check:layering`, `pnpm check:production-exports` (none of the new exports flagged), the four affected daemon suites (18 tests), and the live `AGENT_DEVICE_WEB_E2E=1 pnpm test:smoke:web` lane — all green.

thymikee:
Re-reviewed exact head `f6bf3e2380`: identity-aware daemon exit waiting now lives at the daemon-process owner, is reused by smoke and stop/takeover paths, and planted tests cover recycled PIDs, zombies, timeout, and genuine escalation. Comment cleanup is complete, current-head live web shutdown evidence and all exact-head CI/native lanes are green, CLEAN/MERGEABLE. Merge-ready.

PR #2086, opened 2026-08-27
fix(ios): retry fill's synthesized replacement once when the value never arrives

### LINKED ISSUE #2080: ios: fill's synthesized-replacement route lands a corrupted value (traced: 11 chars requested, 7 observed, frozen)
A `fill` on the penalized coordinate route committed a value that is neither the requested text nor a prefix of it, and stayed that way. The commit wait caught it — #1995 working as designed — but it turns an expensive lane red on unrelated PRs, which is the #1874 symptom from a third mechanism.

## The trace

From the iOS smoke lane of [PR #2059](https://github.com/callstack/agent-device/pull/2059) ([run 33067586234](https://github.com/callstack/agent-device/actions/runs/33067586234/job/98501576566)), a PR that changes one workflow file, one test and `vitest.config.ts` — nothing reachable from the runner.

Step: `agent-device fill id="field-email" ada@example` (11 characters).

```
AGENT_DEVICE_RUNNER_TEXT_ENTRY_ROUTE route=synthesized-first-responder-replacement
[DEBUG-1874] synthesize posted 11 chars status=0 tookMs=1084
[DEBUG-1874] wait start expectedLen=11 route=replacement
[DEBUG-1874] poll t=1102ms observedLen=7 expectedPrefixLen=1
[DEBUG-1874] poll t=2395ms observedLen=7 expectedPrefixLen=1
[DEBUG-1874] poll t=3717ms observedLen=7 expectedPrefixLen=1
[DEBUG-1874] poll t=5097ms observedLen=7 expectedPrefixLen=1
[DEBUG-1874] wait outcome=notObserved elapsedMs=6293 route=replacement
```

**Read it as: the field settled at 7 characters sharing exactly one character with `ada@example`, and never moved again.** Four polls over ~4 s at a stable length. `status=0` — the synthesize call reported success.

This is the dropped-middle-character family already documented in `awaitSynthesizedReplacementCommitOutcome` ("ada@example" -> "aexample", "Ada Lovelace" -> "Avelace"), reproduced in CI with a trace attached.

## What it is not

- **Not a throttle.** `expectedPrefixLen` is flat at 1 for the whole wait. A throttled burst walks the prefix upward; this one never progressed after the first observation.
- **Not the #2035 deadline.** The wait behaved correctly: progress at the first poll only, then condemned 3 s later (the 6293 ms total includes the ~1.1 s placeholder read that precedes the poll loop).
- **Not the placeholder collision** fixed in #2035. `field-email`'s placeholder is now `Type your email`; the filled value shares nothing with it.
- **Not the failing PR's doing.** #2059 touches CI configuration only.

## Why it is worth its own issue

#1874 tracks "why does the input pipeline throttle". This is a different defect with a different signature: the characters are not late, they are *wrong*, and the value is stable. Grouping them would put two mechanisms behind one `needs-info`.

The immediately useful next step is to learn what the 7 characters actually were. The cadence log is value-free by design (#1924), so the corrupted string is not in any artifact — a one-off local reproduction on the penalized route, or a temporary opt-in that logs the observed value under a test-only flag, would say whether the loss is leading, middle or trailing, which points at `RunnerSynthesizedTextEntry.replaceText`'s select-all-then-type sequence versus the per-character pacing above it.

Related: #1995 (the verification that caught this), #1924 (honest reporting), #1874 (same symptom, different mechanism).

---

thymikee:
Closes #2080.

## What happens today

`fill id="field-email" ada@example` posts 11 characters. The field settles at **7**, sharing one character with the request, and stays there:

```
route=synthesized-first-responder-replacement
[DEBUG-1874] synthesize posted 11 chars status=0 tookMs=1084
[DEBUG-1874] poll t=1102ms observedLen=7 expectedPrefixLen=1
[DEBUG-1874] poll t=2395ms observedLen=7 expectedPrefixLen=1
[DEBUG-1874] poll t=3717ms observedLen=7 expectedPrefixLen=1
[DEBUG-1874] poll t=5097ms observedLen=7 expectedPrefixLen=1
[DEBUG-1874] wait outcome=notObserved elapsedMs=6293
```

Flat prefix across four polls over ~4 s: the characters are not late, they are gone. The wait reports that honestly (#1995 doing its job) and stops, so `fill` fails and an expensive lane goes red on PRs that touch no iOS code — the #1874 symptom from a third mechanism.

## The change

The route re-posts once before giving up.

**Safe here and nowhere else.** It opens with select-all, so a second post replaces the whole field rather than appending to whatever committed — the double-commit hazard that made #1676 refuse a retry for bare `type` does not exist when the route owns the field. `awaitSynthesizedReplacementCommitOutcome`'s own doc comment already says as much: "`fill` fully owns the field via select-all, so there is no legitimate reason for the settled value to be anything other than exactly what was requested."

It is also what the *other* route has always done. `verifyTextEntryWithRepairIfNeeded` repairs a mismatch with `clearTextInput` + `typeText`; the synthesized replacement route was simply the one without a second chance. This closes that asymmetry rather than inventing a policy.

Bounded at one retry, and the outcome is still reported honestly if the retry fails too.

## What this does not fix

Why the characters are dropped. 11 posted at `typingSpeed: 60`, `status=0`, 7 landed. That stays open on #2080 — this converts a hard failure into a recovery attempt, it does not explain the loss. If the retry turns out to fail as often as the first attempt, that is itself the answer and the issue should say so.

## Validation

- Full `ios.yml` PR `-only-testing:` list on iPhone 17 Pro / iOS 26.2: **76 passed, 0 failures**.
- `testTypeTextReliablyPacesSynthesizedReplacementThroughProductionCaller` pins the retry through the production caller: the fake synthesizer never writes into the app, so the wait cannot observe and the route posts its select-once-then-pace sequence twice. That test now runs two deadlines (~6.7 s, nightly lane only).
- Rebased onto `bf26ab14d6`; applied three-way so #2066's clear-field work on the same test file is preserved rather than reverted, and its `testEmptyReplacementWithoutResolvableTargetFailsClosed` still passes. Empty-text `fill` fails closed before this route, so the retry does not touch that path.

2 files.

### DIFF
--- a/apple/runner/AgentDeviceRunner/AgentDeviceRunnerUITests/RunnerTests+SynthesizedTextEntry.swift
+++ b/apple/runner/AgentDeviceRunner/AgentDeviceRunnerUITests/RunnerTests+SynthesizedTextEntry.swift
@@ -103,41 +103,63 @@ extension RunnerTests {
       text: request.text,
       delaySeconds: request.delaySeconds
     )
-    for (index, step) in steps.enumerated() {
-      switch request.synthesizer.enterText(
-        app: request.app,
-        text: step.text,
-        replacingExistingText: step.replacesExistingText
-      ) {
-      case .fallback:
-        NSLog("AGENT_DEVICE_RUNNER_TEXT_ENTRY_ROUTE route=verified-fallback reason=synthesis-unavailable")
-        guard let point = request.target.refreshPoint else { return .notApplicable }
-        return .fallback(
-          focusTextInputForTextEntry(app: request.app, x: point.x, y: point.y)
-        )
-      case .raise(let message):
-        NSException(
-          name: NSExceptionName.internalInconsistencyException,
-          reason: message ?? "private XCTest text synthesis failed"
-        ).raise()
-      case .continueTyping:
-        break
+
+    /// Posts the whole replacement once. Returns the route outcome when it could not post at all;
+    /// nil means the events went out and the caller should wait for them.
+    func post() -> SynthesizedReplacementRouteOutcome? {
+      for (index, step) in steps.enumerated() {
+        switch request.synthesizer.enterText(
+          app: request.app,
+          text: step.text,
+          replacingExistingText: step.replacesExistingText
+        ) {
+        case .fallback:
+          NSLog("AGENT_DEVICE_RUNNER_TEXT_ENTRY_ROUTE route=verified-fallback reason=synthesis-unavailable")
+          guard let point = request.target.refreshPoint else { return .notApplicable }
+          return .fallback(
+            focusTextInputForTextEntry(app: request.app, x: point.x, y: point.y)
+          )
+        case .raise(let message):
+          NSException(
+            name: NSExceptionName.internalInconsistencyException,
+            reason: message ?? "private XCTest text synthesis failed"
+          ).raise()
+        case .continueTyping:
+          break
+        }
+        if index + 1 < steps.count {
+          sleepFor(request.delaySeconds)
+        }
       }
-      if index + 1 < steps.count {
-        sleepFor(request.delaySeconds)
-      }
+      return nil
     }
+
+    if let blocked = post() { return blocked }
     // The private synthesize call returns at post time, not commit time (same as bare `type`,
     // see awaitSynthesizedFirstResponderCommit) — but this route never resolves an XCUIElement,
     // so without this wait it had no way to notice a dropped or still-in-flight character at all
     // and reported ok purely because the event posted. Wait here, on the same request.target
     // (element nil, refreshPoint set) that gated this route, so each poll re-resolves via the
     // refresh point rather than trusting a stale element handle.
-    let commit = awaitSynthesizedReplacementCommit(
+    var commit = awaitSynthesizedReplacementCommit(
       app: request.app,
       target: request.target,
       expectedText: request.text
     )
+    // Re-post once when the value never arrived. Safe here and nowhere else: this route opens with
+    // select-all, so posting it again replaces the whole field rather than appending to whatever
+    // committed — the double-commit hazard that made #1676 refuse a retry for bare `type` does not
+    // exist. #2080 is the case it exists for: 11 characters posted, 7 observed, frozen for four
+    // polls, which is loss rather than lateness, and the wait has no other move.
+    if commit == .notObserved {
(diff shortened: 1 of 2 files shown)


thymikee:
[P1] Do not re-post to an unbound first responder from the collapsed `notObserved` outcome. `notObserved` combines missing/unresolvable input, placeholder collision, and corrupted stable values; after the observation delay, the global first responder may no longer be the coordinate-selected field. Re-running select-all/write can therefore mutate a different focused control. This also bypasses the #2080 diagnosis-first contract, and the fake synthesizer tests prove only call order, not target identity or recovery from the real dropped-middle failure. Remove the retry, or introduce a typed stable-corruption outcome and re-establish the same target before retrying. Add planted-red coverage for a wrong stable first value followed by the exact second value, and for lost/changed focus refusing the retry; include the exact-head reproduction evidence requested by #2080.

thymikee:
You are right on every count, and the first-responder hazard is one I did not consider at all. Removing the retry and closing this.

**`.notObserved` collapses four situations**, and I checked rather than argued — `awaitSynthesizedReplacementCommitOutcome` returns it for:

1. a placeholder collision, **before any polling at all** — where the field may hold exactly the right text;
2. `observe()` reading nil throughout, i.e. nothing resolving at the refresh point;
3. a stable wrong value (the #2080 case I was aiming at);
4. nothing ever arriving.

My retry fired on all four. Worse, this route synthesizes to the *global first responder* rather than a resolved element, and it re-posts after a wait of up to 10 s — so select-all-and-write could land on whatever control has focus by then. That is a mutation of state the command never targeted, which is far worse than the failure it was trying to avoid.

**And it bypassed the contract I wrote into #2080 myself**: diagnose first. I do not know what the 7 characters were, so I cannot say whether a retry is even the right remedy — if the loss is deterministic under load, it just fails twice and doubles the time.

**On the tests, you are right that they prove nothing.** `synthesizer.steps == attempt + attempt` asserts call order against a fake that never writes into the app. It cannot observe target identity, and it cannot exercise the dropped-middle failure, which is the only thing that would justify the change.

Closing rather than reworking. The safe version needs a typed stable-corruption outcome, re-establishment of the same target before re-posting, and planted-red coverage for both the wrong-stable-then-exact recovery and a focus change refusing the retry — and all of that should follow the reproduction #2080 asks for, not precede it. I have written the hazard up on #2080 so the next attempt starts from it rather than rediscovering it.

PR #2091, opened 2026-08-27
fix(ci): stop ten artifact uploads discarding their hidden paths

thymikee:
`actions/upload-artifact` has excluded hidden files and directories by default since v4.4, and this repository pins v4.6.2 and writes most of its diagnostics under `.tmp`. Ten upload steps across seven files were therefore uploading nothing from those paths.

| workflow | artifact | `if-no-files-found` |
|---|---|---|
| `macos.yml` | host XCTest xcresult bundle | `warn` |
| `mutation-affected.yml` ×3 | report, shards, lane envelope | `warn` / unset |
| `mutation-weekly.yml` ×2 | report, shards | `warn` / unset |
| `replays-nightly.yml` | fuzz output | `ignore` |
| `xctest-nightly.yml` | results | `warn` |
| `test-app-build-cache.yml` | fixture tarball | **`error`** |
| `1874-diagnose.yml` | per-iteration logs | `ignore` |

Most pair the omission with `warn` or `ignore`, so they fail silently. `test-app-build-cache` sets `error`, so that one does not.

## How it surfaced

The #1874 diagnose loop was dispatched to collect per-iteration logs. The summary arrived with its cadence excerpt and **none of the logs it points at** — `.tmp` is hidden, so every per-iteration log that workflow has ever kept had been discarded. Checking whether anything else did the same found nine more.

## Why a guard rather than a wrapper

A shared upload wrapper would be a shallow mirror of the action's options across artifacts with genuinely different owners — mutation reports, fixture caches, XCTest bundles, session diagnostics. The policy question is one rule ("a hidden path needs the flag") and belongs in one place; the artifact itself belongs to the workflow that produces it. `test/ci/upload-artifact-hidden-paths.test.ts` centralises the policy and leaves each workflow declaring its own artifact.

## Validation

Parse-only, no device lane. The guard walks every workflow and composite action; dropping any single flag turns it red (verified against `xctest-nightly.yml`). Format, lint, typecheck and `test/ci` green.

Not claimed: that this explains any specific past failure. The last eight `Test App Build Cache` runs on `main` are green, and the most recent published zero artifacts — consistent with no rebuild being selected, which proves nothing either way about the upload path.

Split out of #2059 on review; that PR keeps the diagnose-loop behaviour and will drop its duplicate of these changes once this lands.

### DIFF
--- a/.github/workflows/1874-diagnose.yml
+++ b/.github/workflows/1874-diagnose.yml
@@ -106,6 +106,7 @@ jobs:
         uses: actions/upload-artifact@v4
         with:
           name: stall-evidence-${{ matrix.arch }}-${{ matrix.mode }}
+          include-hidden-files: true
           path: |
             stall-summary.txt
             .tmp/stall-failure-*.log
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -122,6 +122,7 @@ jobs:
         uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
         with:
           name: xctest-host-results-${{ github.run_id }}-${{ github.run_attempt }}
+          include-hidden-files: true
           path: .tmp/xctest-host
           if-no-files-found: warn
 
--- a/.github/workflows/mutation-affected.yml
+++ b/.github/workflows/mutation-affected.yml
@@ -81,6 +81,7 @@ jobs:
         uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
         with:
           name: mutation-affected-select
+          include-hidden-files: true
           path: .tmp/mutation/lane-envelope.json
           if-no-files-found: warn
 
@@ -122,6 +123,7 @@ jobs:
         uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
         with:
           name: mutation-affected-shard-${{ matrix.name }}
+          include-hidden-files: true
           path: |
             .tmp/mutation/mutation.json
             .tmp/mutation/mutation.html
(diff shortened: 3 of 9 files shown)


thymikee:
[P2] Make the hidden-upload guard cover every supported config file. `configuredFiles()` accepts only `*.yml` and assumes one-level `action.yml`, but GitHub supports workflow `.yaml`, `action.yaml`, and nested local-action directories. A hidden upload in any of those valid shapes bypasses this structural gate. Discover both extensions recursively and plant the missing-shape violation red. Also remove the new action-version/history comment from the test under current main’s implementation-comment rule. Current static paths are otherwise correctly covered and exact-head CI is green.

thymikee:
Fixed in `8f90ef5dcb`.

**The scan was too narrow — you are right on all three shapes.** It read `.github/workflows/*.yml` and assumed local actions sit one directory deep as `action.yml`. GitHub also accepts `.yaml` for both, and local actions nest. It now walks the `.github` tree recursively for either extension: 34 YAML files, up from the 24 the old shape reached.

Rather than plant the violation by hand once, the shape coverage is now a second test: it fabricates the three shapes the old scan missed — `workflows/*.yaml`, a nested `actions/a/b/action.yaml`, and a shallow `actions/a/action.yml` — and asserts all three are reported. That stays true as the repo grows.

Planted red three ways:

```
scan misses .yaml       -> caught
scan is not recursive   -> caught
a real flag dropped     -> caught
```

**Action-version comment removed** under the #2087 rule. The v4.4 default change is PR history, and the assertion message already carries what the omission costs.

thymikee:
Re-reviewed exact head `8f90ef5dcb`: all ten hidden-path uploads opt in, and the guard recursively covers both YAML extensions plus nested local actions with planted missing-shape proof. No remaining code/comment-policy finding; exact-head CI and all native smokes are green, CLEAN/MERGEABLE. Merge-ready.

PR #2093, opened 2026-08-27
refactor(test): derive provider gateway facts from runtime builder

thymikee:
## Summary

Derive the provider runtime-gateway fixture from the canonical runtime-facts builder while retaining the specialized owner-capability overrides.

- Remove the repeated provider operation and lifecycle literal map.
- Preserve provider-mode versus owner-capability refusal reasons with focused assertions.
- Keep the four already-canonical historical fixtures behavior-specific and unchanged.

## Validation

- Focused Vitest: 7 files, 45 tests passed.
- pnpm check:affected --run: all runnable checks passed; device lanes remain GitHub-authoritative.
- pnpm format

Docs and skills unchanged; test-fixture-only scope.

### DIFF
--- a/src/platform-runtime-gateway.fixtures.ts
+++ b/src/platform-runtime-gateway.fixtures.ts
@@ -19,7 +19,10 @@ import type {
 import type { PlatformRequestScope } from '@agent-device/contracts/platform-runtime-host';
 import type { DeviceInfo } from '@agent-device/kernel/device';
 import type { LimrunRuntimeDependencies } from '@agent-device/provider-limrun';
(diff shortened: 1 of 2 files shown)


thymikee:
Re-reviewed `1008cf47`: this fixture-only refactor now derives the complete provider gap from the typed runtime-facts builder and overlays the existing owner-capability exceptions in the correct precedence. The added gateway assertions prove both refusal reasons, with no production behavior or public-surface change. Code review is clean; merge readiness awaits the pending iOS smoke check.

PR #2094, opened 2026-08-27
refactor: delete capabilities projection shell

thymikee:
## Summary

Move the runtime-use and public runtime-fact projections into the descriptor registry and delete the residual capabilities shell. This preserves descriptor-derived admission, action-selected use semantics, and public-command filtering without a compatibility re-export.

## Validation

Focused descriptor/inventory/eager-closure tests: 340 passed.
Layering: 150 passed.
`pnpm check:affected --run` completed all 3,410 tests on the final retry; Vitest reported existing parallel-load teardown errors, while the affected suites pass in isolation.
GitHub-authoritative device lanes remain for CI.


thymikee:
[P2] Current main now rejects implementation comments that preserve review history or narrate control flow, but the edited `gesture-facts.ts` header still carries R52/R53 and retired `core/capabilities` provenance. Remove that narrative; let the fact table, symbols, and tests express the invariant, with history remaining in the ADR/PR.

The descriptor-projection move itself is sound, but this head is not ready while Integration Tests is red on daemon SIGTERM cleanup and iOS Smoke is pending.

thymikee:
Addressed the P2 in c09f4212c6 by removing the R52/R53 and retired-capability provenance header from packages/platform-apple/src/gesture-facts.ts. The fact table and behavior are unchanged. The new commit is pushed; CI has restarted. The prior Integration Tests failure was the unrelated live web smoke assertion that the daemon exited after SIGTERM.

thymikee:
The capabilities-shell deletion is functionally clean: both projections now read unchanged semantics directly from the descriptor registry. Not ready yet: the gesture-facts follow-up removed only its header; delete the remaining retired-admission/control-flow narration, especially lines 51-55, 62-65, and 98-101. Exact-head iOS is also red on `TEXT_INPUT_COMMIT_NOT_OBSERVED`; apparently unrelated, but it must rerun green before readiness.

thymikee:
Addressed this comment in 24cac913a6: removed the remaining private-helper admission/control-flow/history narration from packages/platform-apple/src/gesture-facts.ts, including the cited ranges. The exported projection documentation and fact behavior are unchanged. The local affected gate passed (415 files, 3,410 tests); exact-head GitHub checks are now rerunning, including iOS Smoke.

thymikee:
Code review is clean at `24cac913a6`: the delta removes all previously cited implementation/history narration without changing semantics. All non-iOS checks are green. Exact-head iOS still fails on the gesture-pan replay; current main is also red on a different iOS scenario, so this appears lane instability rather than a code finding, but require a green rerun or same-load base comparison before merge readiness.

PR #2101, opened 2026-08-27
fix(daemon): dispose late durable capture authorities

### LINKED ISSUE #2099: daemon: durable-capture recovery Promise.race drops a late successful reattach, orphaning the handle
**Severity: correctness (resource leak). MED.**

## Problem

Durable-capture recovery races an acquisition against a deadline/cancellation. When the deadline (or cancellation) wins, a subsequent **successful** acquisition is dropped on the floor — its live handle (recording / log capture) is never disposed. Result: orphaned host handles after a recovery timeout.

## Evidence (at `848c6ea733`)

- [`Promise.race([acquisition, deadline, cancellation])`](src/daemon/durable-capture-recovery-authority.ts#L85-L92); the `finally` does `void acquisition.catch(() => {})`.
- `acquireRecoveryAuthority` **does** dispose both handle and control when it *throws* ([catch block disposes late authority](src/daemon/durable-capture-recovery-authority.ts#L106-L115)). The gap is the **resolved-but-lost** branch: if `acquisition` *resolves* after the deadline already settled the race, the returned `{ control, reattached }` is never consumed and never disposed. `void acquisition.catch(() => {})` only swallows rejections; it does not dispose a success.

## Design / fix

On the lost-race path, dispose a late **success**, not just swallow a late failure:

```
} finally {
  clearTimeout(timer);
  stopListeningForCancellation();
  // dispose a late-arriving success whose result no caller will consume
  void acquisition.then(
    (authority) => disposeLateAuthority(params, /* control + handle */ authority, 'late_race_cleanup_failed', deadlineOrCancelError),
    () => {},
  );
}
```

(Reuse the existing `disposeLateAuthority` helper; dispose both `control` and, if `reattached.status === 'active'`, its handle — mirroring the throw path.)

## Acceptance criteria

- A test where acquisition resolves *after* the deadline fires asserts the late handle and control are disposed (spy on `Symbol.asyncDispose` / the `onLateCleanupFailure` seam).
- The winning-path (acquisition wins) behavior is unchanged.

## Effort / risk

S, MED. One `finally` branch + one test. No behavior change on the success path.

jstar0:
## Summary

- Dispose a late successful durable-capture authority when the deadline or request cancellation wins the recovery race.
- Dispose an active reattached handle before its control, while disposing the control alone for non-active outcomes.
- Preserve the primary timeout or cancellation error and route cleanup failures through the existing callback.

Closes #2099

## Validation

- Focused authority suite: 6 tests passed, including acquisition-winner ownership, active late reattach, non-active late reattach, primary-error preservation, cancellation, and already-canceled recovery.
- `pnpm check:affected --run`: formatting, lint, typecheck, layering, Fallow, build, and related Vitest checks passed; 936 tests passed across 172 files.
- Reverting the production change makes the three late-success/failure tests fail, while the two pre-existing cancellation tests continue to pass.
- Device and replay lanes remain GitHub-authoritative per repository policy.

### DIFF
--- a/src/daemon/__tests__/durable-capture-recovery-authority.test.ts
+++ b/src/daemon/__tests__/durable-capture-recovery-authority.test.ts
@@ -18,8 +18,34 @@ const envelope = createDurableResourceEnvelope({
   descriptor: { version: 1, body: {} },
 });
 
+test('acquisition winner leaves authority disposal to the caller', async () => {
+  const disposeControl = vi.fn(async () => {});
+  const authority = await acquireDurableCaptureRecoveryAuthorityBeforeDeadline({
+    displayName: 'app-log',
+    envelope,
+    scope: {
+      signal: new AbortController().signal,
+      diagnostics: { emit: () => {} },
+      progress: { report: () => {} },
+    },
+    deadlineMs: 10_000,
+    acquireControl: async () => ({
+      reattach: async () => ({ status: 'missing' as const }),
+      cleanup: async () => ({ status: 'already-missing' as const }),
+      [Symbol.asyncDispose]: disposeControl,
+    }),
+    onLateCleanupFailure: () => {},
+  });
+
+  expect(authority.reattached).toEqual({ status: 'missing' });
+  expect(disposeControl).not.toHaveBeenCalled();
+  await authority.control[Symbol.asyncDispose]();
+  expect(disposeControl).toHaveBeenCalledOnce();
+});
+
 test('deadline abort disposes authority that becomes active after the caller has timed out', async () => {
   vi.useFakeTimers();
+  const combinedSignal = vi.spyOn(AbortSignal, 'any').mockReturnValue(new AbortController().signal);
   try {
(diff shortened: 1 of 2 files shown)


thymikee:
Code-clean at `948f4e4023`: lost-race cleanup attaches only after deadline/cancellation wins, preserves the winning acquisition, disposes an active handle before control, and reports cleanup failure without replacing the primary error. The planted active/non-active/cancellation regressions are meaningful. Not merge-ready yet: exact-head CI was not reported during review, and fixture proof does not replace a practical recovery timeout/cancellation run showing that a late live capture handle actually stops.

thymikee:
The new acquisition-winner regression is useful and confirms caller ownership, but exact head `49c5aa8889` is not ready. The detached late-cleanup continuation has no terminal rejection handler: if `onLateCleanupFailure` itself throws, best-effort cleanup becomes an unhandled rejection after the primary timeout/cancellation. Contain that reporting path and add a throwing-reporter regression. Exact-head CI is still absent, and practical evidence that a late live capture handle actually stops is still owed.

jstar0:
Closing as superseded by #2107, which has merged to main.

PR #2102, opened 2026-08-27
fix(daemon): refuse host path install sources on the HTTP surface

### LINKED ISSUE #2097: daemon (remote): install_from_source kind:"path" is an unconfined daemon-host file read on the HTTP surface
**Severity: security (arbitrary daemon-host file read). Surface: shared remote/proxy daemon HTTP, not the local loopback CLI.**

## Problem

`install_from_source` with `source.kind: "path"` accepts an **unconfined** filesystem path over the HTTP surface. A remote authenticated caller can name any path the daemon user can read; the daemon materializes it as the install source, so its bytes flow back through the install pipeline. This is an arbitrary host-file read for anyone who can reach `/rpc` with a valid token.

## Evidence (at `848c6ea733`)

- The HTTP boundary accepts the path with only a non-empty check: [`parseInstallSource` for `kind === 'path'`](src/daemon/server/http-server.ts#L320-L328) — `return { kind: 'path', path: artifactPath }`, no confinement.
- Materialization is unconfined: [`materializeLocalSource` for `kind: 'path'`](src/platforms/install-source.ts#L86-L95) returns `expandUserHomePath(source.path)` directly (`~` expansion, then used as-is).
- Contrast with the **`url`** kind, which has real hygiene: [credentials rejected, blocked hostnames, DNS-resolved non-public-address rejection](src/platforms/install-source-network.ts#L18-L46), plus [redirect caps and HTTPS-downgrade blocks](src/platforms/install-source-download.ts#L34-L93). The `path` kind has none of the equivalent host-side confinement.

## Design: `kind:"path"` is a local affordance; the remote surface must not accept it by default

`kind: "path"` is legitimate for **local/in-process** flows and should stay:
- [`admin.ts`](src/commands/management/runtime/admin.ts#L217) and [its source parse](src/commands/management/runtime/admin.ts#L227) build path sources for local admin.
- [`diff-screenshot.ts`](src/commands/capture/runtime/diff-screenshot.ts#L233) builds one locally.

These do not cross the HTTP `parseInstallSource` boundary, so gating the HTTP surface leaves them untouched.

Fix:

1. **Reject `kind: "path"` at the HTTP boundary by default.** Return `INVALID_ARGS` unless a daemon config explicitly enables host-path installs (e.g. `allowHostPathInstall: true`), off by default.
2. **When enabled, confine it.** Require the resolved, symlink-realpathed absolute path to sit under an operator-configured allowlist root (an install-staging directory), the same "declare the trust boundary" shape the `url` kind already uses for network destinations.

## Refactor required

- Gate + optional confinement helper in the HTTP `parseInstallSource` path branch (src/daemon/server/http-server.ts around L320). Small and self-contained.
- Add the daemon config surface for the opt-in + allowlist root; document it as a deliberate expansion of the daemon's trust boundary.
- Local callers (`admin.ts`, `diff-screenshot.ts`) are unaffected and need no change.

## Acceptance criteria

- Over `/rpc`, `install_from_source` with `kind: "path"` is rejected with `INVALID_ARGS` when the opt-in is off (default). Handler-level test.
- With the opt-in on and an allowlist root, a path inside the root succeeds and a path outside it (including via `..`/symlink escape) is rejected. Realpath-based confinement test.
- Local flows that build path sources continue to work.

## Related

Same "apply the public-URL / trust-boundary rule on the remote surface, leave trusted local flows alone" theme as the Maestro `runScript http.post` allowlist gap (filed separately).

## Effort / risk

M (the gate is small; choosing and wiring the confinement root is the M part), MED risk.

thymikee:
Closes #2097.

Two gates, because two RPC methods can carry an install source and they do not share a trust boundary.

## The vulnerability

`install_from_source` accepted `source.kind: "path"` behind nothing but a non-empty check, then handed the string to `materializeLocalSource`, which expands `~` and reads it as-is. Any caller reaching `/rpc` with a valid token could name a file the daemon user can read and have its bytes flow back through the install pipeline.

## Gate 1 — the daemon's `install_from_source` method

`parseInstallSource` now refuses `kind: "path"` outright rather than confining it. The `url` kind already declares its trust boundary (credential rejection, blocked hostnames, non-public-address rejection, redirect caps); the `path` kind has no host-side equivalent and does not need one — it is an affordance for callers that already carry the daemon's authority. No opt-in and no allowlist-root config surface, per the triage note.

`HttpInstallSource = Exclude<DaemonInstallSource, { kind: 'path' }>` carries the narrowed contract, so re-adding a path return fails to compile. Verified by planting one: `TS2322: Type '"path"' is not assignable to type '"github-actions-artifact" | "url"'`.

## Gate 2 — the proxy, for the method the client actually uses

That first gate covers the wrong route on its own. The generic `agent_device.command` method copies `params.meta` wholesale and `commandRpcParamsSchema` types `meta` as an opaque object, so `meta.installSource` reached the install handler unparsed — and `buildHttpRpcPayload` always sends `agent_device.command`. Before this PR:

```
POST /rpc  {"method":"agent_device.command",
            "params":{"command":"install_source","flags":{"platform":"android"},
                      "meta":{"installSource":{"kind":"path","path":"/etc/passwd"}}}}
→ 200, handler receives {"kind":"path","path":"/etc/passwd"}
```

**Refusing that at the daemon's HTTP boundary would take local callers with it.** The daemon's HTTP server binds `127.0.0.1` (`listenLoopbackServer`), and `agent-device proxy` puts the local daemon into HTTP-only mode — so on any host running a proxy, a plain local `agent-device install <path>` crosses that same boundary with the same wire shape (`prepareRemoteRequestArtifacts` uploads only when `info.baseUrl` is set, which a loopback daemon has not). Gating there would break local installs to fix a remote hole.

The proxy is the seam that actually separates the daemon's host from callers who are not on it — it is the component whose entire job is exposing the daemon off-host, and it already reads and re-serializes the RPC body to swap tokens. So the refusal lives there, where it covers **both** methods at once because it inspects the payload before the method is dispatched. It also cannot be dodged by parser divergence: the daemon receives the very body the proxy parsed, re-serialized by `rewriteRpcToken`.

**Carve-out:** a path source still crosses when the request carries `meta.uploadedArtifactId` — the sanctioned remote install route, where `resolveInstallSource` substitutes the uploaded file and never reads the wire path. A forged id cannot smuggle a path through it: `prepareUploadedArtifact` → `requireTenantOwnedEntry` throws on an unknown or cross-tenant id rather than falling back. That no-fallback property is the carve-out's whole safety argument, so it is now pinned by its own test — verified to go red when a fallback is introduced.

## Behavior change worth your attention

**`install remote:<path>` is now refused through a proxy.** Naming a file on the daemon's host from a remote client *is* the reported vulnerability, so the affordance cannot survive the fix. I left the client-side `remote:` prefix handling in place (it still applies to a remote daemon reached without the proxy) — say the word if you'd rather retire the client surface too.

## Tests

Both refusals were **observed red** before the fix.

- `src/daemon/__tests__/http-server-rpc-validation.test.ts` — a path source over `/rpc` is rejected `400` / `-32602` / `INVALID_ARGS` and never dispatched; `url` and `github-actions-artifact` still reach the handler with their payload intact.
- `src/__tests__/daemon-proxy.test.ts` — the proxy refuses a path source on *both* methods and the upstream daemon is never called; it still forwards an upload-backed path source and a url source.
- `src/daemon/__tests__/install-source-resolution.test.ts` — an unknown upload id throws instead of reading the wire path.

Local in-process callers are untouched: `admin.ts` and `diff-screenshot.ts` build path sources without crossing either gate, and their existing coverage is unchanged and green.

## One thing left for your call

**No `DAEMON_RPC_PROTOCOL_VERSION` bump.** This removes wire input a released daemon accepted, and `test/wire-compat/README.md` says removal is always the breaking case. The gate stays green only because `surface.ts` declares the `DaemonInstallSource` *type*, not these parsers. I left the version alone deliberately: an old client now fails clearly with `INVALID_ARGS` rather than misparsing anything, which is the distinction ADR 0006 draws, and a bump would sever otherwise-compatible peers over a change that already fails loudly. Happy to take the bump if you read it the other way.

## Validation

`pnpm format:check`, `oxlint --deny-warnings`, `pnpm typecheck`, `pnpm check:layering`, `pnpm check:fallow`, `test/wire-compat`, `pnpm test:integration:provider` (54 files / 161 tests) — all green. Full `pnpm test:unit`: 1074 files / 8158 tests green.

No docs or skills changed: neither RPC method has user-facing documentation, and both refusals carry their own message and hint.

### DIFF
--- a/src/__tests__/daemon-proxy.test.ts
+++ b/src/__tests__/daemon-proxy.test.ts
@@ -513,3 +513,134 @@ function sendUploadProxyFinalize(
   res.setHeader('content-type', 'application/json');
   res.end(JSON.stringify({ ok: true, uploadId: 'tracked-upload-1' }));
 }
+
+type ProxyRpcOutcome = {
+  status: number;
+  body: { error?: { code?: number; message?: string; data?: { code?: string } } };
+  upstreamCalls: number;
+};
+
+async function postInstallRpcThroughProxy(
+  params: Record<string, unknown>,
+): Promise<ProxyRpcOutcome> {
+  let upstreamCalls = 0;
+  const upstream = http.createServer((req, res) => {
+    upstreamCalls += 1;
+    let body = '';
+    req.setEncoding('utf8');
+    req.on('data', (chunk) => {
+      body += chunk;
+    });
+    req.on('end', () => {
+      res.setHeader('content-type', 'application/json');
+      res.end(
+        JSON.stringify({
+          jsonrpc: '2.0',
+          id: (JSON.parse(body) as { id?: unknown }).id,
+          result: { ok: true, data: { reached: 'upstream' } },
+        }),
+      );
+    });
+  });
+  const proxy = createDaemonProxyServer({
+    upstreamBaseUrl: `http://127.0.0.1:${await listenOnLoopback(upstream)}`,
+    upstreamToken: 'daemon-secret',
+    clientToken: 'proxy-secret',
+  });
+
+  try {
+    const proxyPort = await listenOnLoopback(proxy);
+    const response = await fetch(`http://127.0.0.1:${proxyPort}/agent-device/rpc`, {
+      method: 'POST',
+      headers: { 'content-type': 'application/json', authorization: 'Bearer proxy-secret' },
+      body: JSON.stringify({ jsonrpc: '2.0', id: 'req-1', ...params }),
+    });
+    return {
+      status: response.status,
+      body: (await response.json()) as ProxyRpcOutcome['body'],
+      upstreamCalls,
+    };
+  } finally {
+    await closeLoopbackServer(proxy);
+    await closeLoopbackServer(upstream);
+  }
+}
+
+// #2097: the daemon's own HTTP server binds loopback, so the proxy is the seam between
+// the daemon's host and callers that are not on it. A `path` install source names a file
+// on that host, so it must not cross — through either rpc method that can carry one.
+test('proxy refuses a host path install source on the generic command method', async (t) => {
+  if (await skipWhenLoopbackUnavailable(t)) return;
+
+  const { status, body, upstreamCalls } = await postInstallRpcThroughProxy({
+    method: 'agent_device.command',
+    params: {
+      token: 'proxy-secret',
+      command: 'install_source',
+      positionals: [],
+      flags: { platform: 'android' },
+      meta: { installSource: { kind: 'path', path: '/etc/passwd' } },
+    },
+  });
+
+  assert.equal(status, 400);
+  assert.equal(body.error?.code, -32602);
+  assert.equal(body.error?.data?.code, 'INVALID_ARGS');
+  assert.equal(upstreamCalls, 0, 'the daemon must never see a proxied host path source');
+});
+
+test('proxy refuses a host path install source on the install_from_source method', async (t) => {
+  if (await skipWhenLoopbackUnavailable(t)) return;
+
+  const { status, body, upstreamCalls } = await postInstallRpcThroughProxy({
+    method: 'agent_device.install_from_source',
+    params: {
+      token: 'proxy-secret',
+      platform: 'android',
+      source: { kind: 'path', path: '/etc/passwd' },
+    },
+  });
+
+  assert.equal(status, 400);
+  assert.equal(body.error?.data?.code, 'INVALID_ARGS');
+  assert.equal(upstreamCalls, 0);
+});
+
+// The sanctioned remote install route: the client uploads the artifact first, and the
+// daemon substitutes the uploaded file for the wire path, so the path is never read.
+test('proxy forwards a path source backed by an uploaded artifact', async (t) => {
+  if (await skipWhenLoopbackUnavailable(t)) return;
+
+  const { status, upstreamCalls } = await postInstallRpcThroughProxy({
+    method: 'agent_device.command',
+    params: {
+      token: 'proxy-secret',
+      command: 'install_source',
+      positionals: [],
+      flags: { platform: 'android' },
+      meta: {
+        installSource: { kind: 'path', path: '/Users/dev/Downloads/Sample.apk' },
+        uploadedArtifactId: 'upload-1',
+      },
+    },
+  });
+
+  assert.equal(status, 200);
+  assert.equal(upstreamCalls, 1);
+});
+
+test('proxy forwards url install sources unchanged', async (t) => {
+  if (await skipWhenLoopbackUnavailable(t)) return;
+
+  const { status, upstreamCalls } = await postInstallRpcThroughProxy({
+    method: 'agent_device.install_from_source',
+    params: {
+      token: 'proxy-secret',
+      platform: 'android',
+      source: { kind: 'url', url: 'https://example.com/app.apk' },
+    },
+  });
+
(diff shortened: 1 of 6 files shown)


thymikee:
[P0] Not ready: this rejects only `agent_device.install_from_source`, but shipped remote clients send generic `agent_device.command`; its opaque `meta.installSource:{kind:"path"}` is still forwarded into `install_source`, so the daemon-host path exploit remains reachable. Enforce denial at the shared HTTP command projection before dispatch while preserving server-stamped upload/local substitutions, and plant a generic-wire zero-handler regression. Decide/prove the released protocol-version boundary for removing this input, remove issue/control-flow narration, and finish exact-head CI.

thymikee:
The prior security blocker is resolved: the proxy now rejects both generic and special-method host-path shapes before upstream dispatch, while uploaded artifact ids are tenant-bound and materialized server-side with no wire-path fallback. The generic protocol shape remains intact, so the recorded no-bump decision is sound. Remaining P2: remove implementation/issue narration in `daemon-proxy.ts`, its proxy tests, and `install-source-resolution.test.ts`; keep an eye on the now-495-line proxy owner. Exact-head CI is largely still pending. Not ready yet.

thymikee:
Narration removed, and the extraction answers the file-size point at the same time.

`daemon-proxy.ts` held no explanatory comments before this PR — only the `di-seam-approved` directive — so the block I added was style I imported rather than style I matched. The rule the comment was carrying now lives in a module instead: `src/remote/proxy-install-source-admission.ts` owns the refusal, and `isBackedByUploadedArtifact` names what "unbacked" means in `carriesUnbackedHostPathInstallSource`. That drops the proxy owner from 495 back to 443 lines (434 before this PR; the +9 is the import and the call site).

Test narration is gone from `daemon-proxy.test.ts` and `install-source-resolution.test.ts`. The test names already stated the invariants, so nothing moved into a comment elsewhere:

- `proxy refuses a host path install source on the generic command method`
- `proxy refuses a host path install source on the install_from_source method`
- `proxy forwards a path source backed by an uploaded artifact`
- `resolveInstallSource refuses an unknown upload id rather than reading the wire path`

I also applied the same rule to two spots you did not name, so the diff is consistent rather than consistent-except-here: the `#2097` comment above the boundary test in `http-server-rpc-validation.test.ts`, and the JSDoc on `HttpInstallSource` in `http-server.ts`. The type is `Exclude<DaemonInstallSource, { kind: 'path' }>` and says what it needs to; the why is in this PR. Revert either if you wanted them kept.

There are now no authored comments left in the diff.

Re-validated: format, lint, typecheck, layering (1495 files), fallow, wire-compat, provider integration (54 files / 161 tests), full unit (1074 files / 8158 tests). The one unit-lane red was `gesture-admission-parity.test.ts` timing out under load; it passes in isolation and is untouched by this change. `check:production-exports` and `check:bundle-owner-files` fail identically on a clean tree, so they are pre-existing rather than mine.

thymikee:
Code review is clean at `96092af28c`: host-path admission now sits in a focused helper, both generic and special RPC shapes are denied before upstream dispatch, upload-backed materialization remains tenant-bound, and the prior narration/size concern is resolved. No remaining code finding. Exact-head CI is still in progress, so this is code-ready with GitHub checks authoritative.

PR #2107, opened 2026-08-27
fix(daemon): dispose late durable capture authorities

### LINKED ISSUE #2099: daemon: durable-capture recovery Promise.race drops a late successful reattach, orphaning the handle

thymikee:
## Summary

Dispose late successful durable-capture reattachments when deadline or request cancellation wins the recovery race. Active reattachments dispose the live handle before the control; non-active outcomes dispose the control only. Both the detached race cleanup and the actual aborted-scope cleanup keep control disposal unconditional, while cleanup diagnostics cannot replace the primary timeout or cancellation error.

Closes #2099
Supersedes #2101

## Validation

- `pnpm vitest run --project unit-core src/daemon/__tests__/durable-capture-recovery-authority.test.ts`: 10 tests passed.
- Red proof against both pre-fix paths: the outer late-resolution path and the real aborted-scope path left control undisposed when active-handle cleanup and its reporter failed; both regressions pass with this fix.
- `pnpm check:affected --run`: format, lint, typecheck, layering, Fallow, and build passed. The related Vitest lane ran 172 files / 939 tests successfully but reported unrelated Apple `EnvironmentTeardownError` rejections during teardown on two local runs; exact-head GitHub lanes remain authoritative.
- Touched files: 2. No docs or skills changed.

Residual timing risk is limited to the inherently detached late-cleanup continuation; native/device evidence remains GitHub-authoritative.

### DIFF
--- a/src/daemon/__tests__/durable-capture-recovery-authority.test.ts
+++ b/src/daemon/__tests__/durable-capture-recovery-authority.test.ts
@@ -18,6 +18,31 @@ const envelope = createDurableResourceEnvelope({
   descriptor: { version: 1, body: {} },
 });
 
+test('acquisition winner leaves authority disposal to the caller', async () => {
+  const disposeControl = vi.fn(async () => {});
+  const authority = await acquireDurableCaptureRecoveryAuthorityBeforeDeadline({
+    displayName: 'app-log',
+    envelope,
+    scope: {
+      signal: new AbortController().signal,
+      diagnostics: { emit: () => {} },
+      progress: { report: () => {} },
+    },
+    deadlineMs: 10_000,
+    acquireControl: async () => ({
+      reattach: async () => ({ status: 'missing' as const }),
+      cleanup: async () => ({ status: 'already-missing' as const }),
+      [Symbol.asyncDispose]: disposeControl,
+    }),
+    onLateCleanupFailure: () => {},
+  });
+
+  expect(authority.reattached).toEqual({ status: 'missing' });
+  expect(disposeControl).not.toHaveBeenCalled();
+  await authority.control[Symbol.asyncDispose]();
+  expect(disposeControl).toHaveBeenCalledOnce();
+});
+
 test('deadline abort disposes authority that becomes active after the caller has timed out', async () => {
   vi.useFakeTimers();
   try {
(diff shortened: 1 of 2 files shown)


thymikee:
Not ready. The terminal catch fixes #2101's unhandled rejection, but an active late authority can still leak control: if handle disposal fails and `onLateCleanupFailure` throws, cleanup exits before control disposal. Contain reporter failure per resource (or continue in `finally`) and plant active-handle-fails + reporter-throws proof that control still disposes and the primary error remains authoritative. Declare #2107 as replacement/supersession for duplicate #2101, record practical late-live-handle cleanup evidence, and finish exact-head CI.

thymikee:
P1 — control disposal is still skipped on the real aborted-scope path. The new finally protects only disposeLateRecoveryAuthority, but deadline/cancellation aborts scope.signal; after reattach resolves active, acquireRecoveryAuthority's throwIfAborted enters its catch. That catch disposes the handle first, and a throwing onLateCleanupFailure then skips control disposal because this catch still has no finally. The added regression mocks AbortSignal.any to a never-aborted signal, so it forces the outer path and cannot prove the real invariant. Add a planted-red test with an actually aborted scope, late active reattach, handle cleanup failure, and throwing reporter; then make control disposal unconditional in acquireRecoveryAuthority as well. #2101 remains open and superseded and should be marked obsolete once this replacement is accepted.

thymikee:
Re-reviewed `10c6e966aa`: the prior aborted-scope leak is resolved. `acquireRecoveryAuthority` now guarantees control disposal in `finally` even when late active-handle cleanup reporting throws, and the new regression uses a real aborted combined scope, preserves the cancellation as primary, and proves control disposal. All exact-head checks are green. Code review clean; ready for human merge.

PR #2109, opened 2026-08-27
fix(daemon): enforce provider session artifact ownership

### LINKED ISSUE #2096: daemon (remote): artifacts --provider-session forwards an unowned provider session id to the cloud provider with no tenant ownership check
**Severity: security (tenant isolation). Surface: shared remote/proxy daemon with a cloud artifact provider, not the local loopback CLI.**

## Problem

`artifacts --provider-session <id>` forwards the caller-supplied provider session id to the cloud provider **with no check that the id belongs to a lease the requesting tenant owns**. A token holder who knows (or guesses/enumerates) another tenant's provider session id can list that tenant's cloud artifacts.

## Evidence (at `848c6ea733`)

- `artifacts` is admission-exempt: [`ADMISSION_AND_LOCK_EXEMPT` on the `artifacts` descriptor](src/core/command-descriptor/registry.ts#L471) — no lease admission runs for it.
- The daemon forwards the raw id: [`listCloudArtifactsForRequest` passes `providerSessionId` straight to `cloudArtifactProvider.listCloudArtifacts`](src/daemon/handlers/lease.ts#L274-L295); the only gate is "a `--provider` is set and either a lease id or a provider-session id is present."
- The provider accepts a **bare, unbound** id: [`listCloudArtifacts` uses `query.providerSessionId` directly when no session matches `query.leaseId`](packages/provider-webdriver/src/runtime-session.ts#L147-L159). It never checks that this id was ever issued to the caller.

## Design crux: the ownership binding is split across two layers that never exchange it

- The daemon `LeaseRegistry` knows **`leaseId → tenantId`** ([lease records carry `tenantId`](src/daemon/lease-registry.ts#L261-L271)).
- The provider runtime knows **`leaseId → providerSessionId`** (`sessionsByLeaseId`, and `releasedProviderSessionIdsByLeaseId` for a bounded post-release window).
- Neither layer can answer *"does tenant T own provider session P?"* on its own, and the artifacts path asks the provider — the layer without tenant knowledge — so the question is never asked.

The `CloudArtifactsQuery` contract does not even carry `tenantId` ([packages/contracts/src/cloud-artifacts.ts](packages/contracts/src/cloud-artifacts.ts#L59-L77)), so the provider *cannot* enforce ownership today.

## Design: resolve ownership in the daemon, before dispatch

The daemon is the only layer that holds tenant identity, so the ownership check belongs there — not pushed into the provider.

1. Give the daemon a reverse index **`providerSessionId → { leaseId, tenantId }`** for ids it has issued, retained for a bounded window past release (the tenant-aware daemon-side analogue of the provider's `releasedProviderSessionIdsByLeaseId`). Populate it when a provider lease is allocated / its session id becomes known.
2. In `listArtifactsForRequest`, when the caller supplies `--provider-session <id>`, resolve it through that index and **require `entry.tenantId === leaseScope.tenantId`** before calling the provider. An id that does not resolve to a lease owned by the requesting tenant is refused (`NOT_FOUND`/`UNAUTHORIZED` — do not leak existence).
3. Remove or gate the provider's **bare-id** path (runtime-session.ts:151-158) so the provider only ever answers for ids the daemon has resolved and authorized. Passing `tenantId` into `CloudArtifactsQuery` as defense-in-depth is optional once the daemon gates first.

## Refactor required

- Extend `LeaseRegistry` (or a sibling record next to it) with the tenant-tagged `providerSessionId` reverse index + a bounded retention for released sessions.
- Thread the ownership check into `listArtifactsForRequest` / `listCloudArtifactsForRequest` (src/daemon/handlers/lease.ts).
- Tighten `WebDriverProviderRuntime.listCloudArtifacts` to drop the unbound-id lookup.

## Depends on

**#2095** — the ownership check compares against `leaseScope.tenantId`, which is only trustworthy once client-declared tenant can no longer be spoofed.

## Acceptance criteria

- A request from tenant A supplying tenant B's provider session id is refused, with no artifact listing and no existence oracle. Test at the daemon handler level (does not require a live provider).
- A request supplying the requester's own provider session id (active and recently-released) still succeeds.
- The provider no longer returns artifacts for an id it was never asked about through a daemon-authorized path.

## Effort / risk

M, MED risk (touches lease bookkeeping and the released-session window). Land after #2095.

thymikee:
## Summary

- Enforce daemon-side tenant ownership for explicit `artifacts --provider-session` lookups with bounded retention across active, released, and expired leases.
- Pass the resolved lease ID to providers and remove WebDriver bare provider-session lookup.
- Anchor expiry retention at the lease expiry timestamp, including provider sessions returned late by allocation or release and clock-boundary cleanup races.
- Stacked on #2104 / #2095 and fixes #2096.

## Validation

- Focused ownership, lifecycle, registry, provider, daemon, and provider-integration tests: 41/41 passed.
- `pnpm check:affected --run`: 203 files and 1,292 tests passed; all runnable checks passed, including wire compatibility.
- Red-first regressions were observed before the fix for daemon foreign and bare access, WebDriver bare-id access, lazy expiry cleanup, late provider responses, and the sequenced clock-boundary race.
- Rebased onto the updated #2104 tip (`e4d9a2fe5a`) after that PR's stack advanced; exact-head CI is green (12 substantive checks, including Android/iOS/Linux/macOS smoke), with only the expected matrix skip.

### DIFF
--- a/packages/provider-webdriver/src/runtime-session.test.ts
+++ b/packages/provider-webdriver/src/runtime-session.test.ts
@@ -72,6 +72,33 @@ test('a create-timeout surfaces provider evidence for the maybe-leaked session',
   }
 });
 
+test('cloud artifact lookup does not accept a provider session id without its lease', async () => {
+  let listCalls = 0;
+  const runtime = makeRuntime({
+    listArtifacts: async () => {
+      listCalls += 1;
+      return {
+        provider: 'webdriver-test',
+        status: 'ready',
+        cloudArtifacts: [],
+      };
+    },
+  });
+
+  try {
+    const cloudArtifacts = runtime.cloudArtifacts;
+    assert.ok(cloudArtifacts);
+    const result = await cloudArtifacts.listCloudArtifacts?.({
+      provider: 'webdriver-test',
+      providerSessionId: 'never-authorized',
+    });
+    assert.equal(result, undefined);
+    assert.equal(listCalls, 0);
+  } finally {
+    await runtime.shutdown();
+  }
+});
+
 function makeRuntime(overrides: Partial<CloudWebDriverRuntimeOptions> = {}) {
   return createCloudWebDriverRuntime({
     clientVersion: 'test',
--- a/packages/provider-webdriver/src/runtime-session.ts
+++ b/packages/provider-webdriver/src/runtime-session.ts
@@ -146,12 +146,25 @@ export class WebDriverSessionManager {
 
   async listCloudArtifacts(query: CloudArtifactsQuery): Promise<CloudArtifactsResult | undefined> {
     if (query.provider !== this.options.provider) return undefined;
-    const session = query.leaseId ? this.sessionsByLeaseId.get(query.leaseId) : undefined;
-    if (session) return await this.safeListArtifacts(session);
-    const providerSessionId =
-      query.providerSessionId ??
-      (query.leaseId ? this.releasedProviderSessionIdsByLeaseId.get(query.leaseId) : undefined);
+    const session = this.sessionsByLeaseId.get(query.leaseId ?? '');
+    if (session) return await this.listActiveCloudArtifacts(session, query.providerSessionId);
+    return await this.listReleasedCloudArtifacts(query);
+  }
+
+  private async listActiveCloudArtifacts(
+    session: WebDriverProviderSession,
+    providerSessionId: string | undefined,
+  ): Promise<CloudArtifactsResult | undefined> {
+    if (providerSessionId && providerSessionId !== session.providerSessionId) return undefined;
+    return await this.safeListArtifacts(session);
+  }
+
+  private async listReleasedCloudArtifacts(
+    query: CloudArtifactsQuery,
+  ): Promise<CloudArtifactsResult | undefined> {
+    const providerSessionId = this.releasedProviderSessionIdsByLeaseId.get(query.leaseId ?? '');
     if (!providerSessionId || !this.options.listArtifacts) return undefined;
+    if (query.providerSessionId && query.providerSessionId !== providerSessionId) return undefined;
     return await this.options.listArtifacts({
       provider: this.options.provider,
       providerSessionId,
--- a/src/__tests__/daemon-entrypoint.test.ts
+++ b/src/__tests__/daemon-entrypoint.test.ts
@@ -171,11 +171,10 @@ test('daemon runtime publishes dual transport metadata', async () => {
   }
 });
 
-test('daemon default provider composition serves cloud artifacts over RPC', async () => {
+test('daemon rejects unowned cloud artifacts over RPC', async () => {
   const stateDir = mkdtempForTestSync('agent-device-daemon-provider-');
   const providerRequests: string[] = [];
-  const providerServer = http.createServer((req, res) => {
-    providerRequests.push(req.url ?? '');
+  const providerServer = http.createServer((_req, res) => {
     res.setHeader('content-type', 'application/json');
     res.end(
       JSON.stringify({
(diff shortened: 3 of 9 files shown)


thymikee:
P1 — expiry retention is measured from lazy cleanup, not from lease expiry. consumeExpiredLeases discovers an expired lease only when a later registry operation runs; unbindLease then calls markLeaseReleased, which sets retainedUntil = now + retentionMs. A first artifact request hours after lease.expiresAt therefore creates a fresh five-minute authorization window. The test masks this by resolving at expiresAt + 1, which starts the window immediately. Carry the actual transition time into ownership retention (explicit release uses current time; expiry uses lease.expiresAt), and plant a regression that jumps directly past expiresAt + retention before the first resolve and verifies refusal with no provider call. Also rerun/base-compare the currently red Android native-alert timeout before readiness.

thymikee:
Fixed in df3b68183a. Expiry cleanup now anchors provider-session retention at lease.expiresAt, and the regression jumps directly past expiresAt plus retention before the first resolve. Focused tests are 39/39 and pnpm check:affected --run passed all runnable checks. The Android native-alert rerun remains queued.

thymikee:
Re-reviewed `df3b68183a`: the normal lazy-cleanup expiry case is fixed and the direct regression is valid. One P1 remains. `recordProviderSession()` first cleans an expired lease, then records a provider session returned late and marks it released with default `now()`. A delayed allocation/release response can therefore create a fresh authorization window from response time instead of the true lease expiry. Preserve the terminal timestamp when recording an already-expired lease (while keeping manual release semantics distinct), and add planted-red coverage where provider allocation/release returns a session only after expiry+retention; no provider artifact dispatch may occur. Android/iOS smoke are still incomplete.

thymikee:
Fixed in 1e41afe29f. recordProviderSession now preserves lease.expiresAt when a late allocation or release response arrives after expiry; manual release still uses current time. Added planted-red coverage for both delayed paths with no provider dispatch. Focused tests are 41/41 and pnpm check:affected --run passes 203 files and 1,292 tests. The Android rerun passed on the prior head; current checks are running for this follow-up.

thymikee:
Re-reviewed `1e41afe29f`: the late allocation/release fix still leaves a clock-boundary P1. `recordProviderSession` computes `releasedAt` with one `now()` call before `cleanupExpiredLeases()`, whose later `now()` can cross `expiresAt`. Cleanup then removes the lease, but `releasedAt` is still undefined and `markLeaseReleased` defaults to response-time `now()`, recreating the fresh retention window. The current late tests use a constant jumped clock and miss this boundary. Derive the expiry terminal time after cleanup (or make cleanup use the same captured timestamp) and plant a sequenced-clock regression crossing expiry between those calls; no provider dispatch may occur after expiry+retention. Manual release semantics must remain current-time. Exact-head CI is also incomplete.

thymikee:
Fixed in 6040534afb. recordProviderSession now derives the expiry terminal time after cleanup, so a clock crossing expiresAt between calls cannot reopen the retention window. Added a sequenced-clock red/green regression proving no provider dispatch after expiry plus retention. Focused 41/41 and pnpm check:affected --run (203 files, 1,292 tests) pass; exact-head CI is running after this push.

thymikee:
Code review is clean at `6040534afb2ab9dd839cc1a5722aed4599091a12`: deriving terminal time after cleanup fixes the two-`now()` expiry-boundary race, and the sequenced-clock regression is non-vacuous. Manual-release retention remains anchored to current release time. This is ready for human review, but not merge-ready while exact-head Android and iOS smoke checks are still running.

thymikee:
Merge-ready at `6040534afb2ab9dd839cc1a5722aed4599091a12`: all 12 substantive exact-head checks are green, including Android/iOS/Linux/macOS smoke, with only the expected matrix skip. Please replace the stale “CI … now running” line in Validation with the completed exact-head evidence.

thymikee:
The reviewed head is unchanged, but the newly advanced `main` now conflicts with this branch. I removed `ready-for-human`; rebase and resolve the branch blocker before the next readiness review.

thymikee:
Reviewed exact head `15227f0`: code is clean. Daemon-side ownership resolution, zero-dispatch rejection cases, WebDriver binding checks, bounded retention, and expiry anchoring are covered; all exact-head substantive checks are green and size stays below escalation thresholds. Two process items remain: this PR is stacked on open prerequisite #2104, so it is not independently landable to main; and the body still says exact-head CI is running although it has completed green. Update that validation sentence. Readiness label will be handled by the coordinator after the prerequisite review.

thymikee:
Rebased again onto #2104's new tip (`423cd18232`) after that PR's trust posture changed (fail-closed is now unconditional when a hook doesn't attest, regardless of client input). Clean rebase, no conflicts, same 9-file diff. Verified with a proper `pnpm install`: `tsc`, `oxlint`, and the focused ownership/lifecycle/registry/provider/wire-compat suites (60/60) pass. Watching CI now.

PR #2113, opened 2026-08-28
fix(wait): surface runner restart timeout evidence

### LINKED ISSUE #2103: wait: a runner restart inside the wait is reported as a capture stall, and outlives the whole budget
A `wait` that spans a runner restart reports a capture stall, and the restart alone outlives the wait's whole budget. Traced from an iOS smoke failure on an unrelated PR ([run 33098577026](https://github.com/callstack/agent-device/actions/runs/33098577026/job/98609879771), #2059, which changes only a dispatch-only workflow and a script it calls).

## What the user sees

```
agent-device wait text "Agent Device Tester" 10000 --platform ios …
COMMAND_FAILED: wait timed out for text: Agent Device Tester
hint: No readable snapshot capture completed before the wait timeout.
      Retry, or use screenshot to inspect the current surface.
```

The hint points at the surface. The surface was fine.

## What actually happened

From the request log (`sessions/ios-e2e-smoke-nau/requests/fc627b1104c2bcc2.ndjson`):

```
17:45:01.969  ios_runner_readiness_preflight   findText
17:45:07.124  ios_runner_readiness_preflight   snapshot  durationMs=1029   [error]
17:45:07.124  ios_runner_session_invalidated
17:45:10.605  ios_runner_startup_ensure_booted
17:45:16.872  ios_runner_startup_cleanup_stale_bundles  durationMs=6267
17:45:18.975  ios_runner_startup_ensure_xctestrun       durationMs=2102
17:45:19.075  ios_runner_startup_launch_xcodebuild
17:45:20.259  retry attempt_failed (1/3)
17:45:20.261  retry exhausted (3/3)
17:45:20.262  snapshot_capture  durationMs=14333  backend=xctest  error="request canceled"
17:45:20.274  request_failed    wait timed out for text
```

A preflight snapshot failed, the runner session was invalidated, and the runner restarted inside the wait. The restart alone spent **~12 s** — 6.3 s of it in `cleanup_stale_bundles` — against a **10 s** wait budget. The capture was then canceled, and the failure was reported as if nothing readable had appeared on screen.

## Two separable problems

1. **The wait budget does not account for a restart it did not ask for.** A 10 s wait cannot survive a ~12 s runner restart, so any invalidation inside a default-length wait is a guaranteed failure regardless of the app's state. Either the restart should not be charged to the caller's budget, or the wait should extend across it.
2. **The error names the wrong thing.** `wait_capture_stalled` and "no readable snapshot capture completed" send the reader to the app and the screen. The daemon log has `ios_runner_session_invalidated` and a full restart sequence; that is what should surface. AGENTS.md already says a five-second-plus interaction is a daemon-log question naming exactly these phases — the error should carry that rather than making each reader rediscover it.

Fixing (2) alone would make this class self-diagnosing even if (1) stays as designed.

## Frequency

Twice in one working session on unrelated PRs, both during scenario bootstrap: this one, and an Android `wait timed out for text: Automation lab` with the same `wait_capture_stalled` hint ([run 32977635983](https://github.com/callstack/agent-device/actions/runs/32977635983)). The Android instance was not traced, so I am not claiming the same mechanism — only the same reported symptom, which is part of the point of (2).

---

kkkhs:
## Summary

Closes #2103. Keep the caller-supplied wait deadline hard, but preserve an iOS runner restart cause when the capture is cancelled after the restart consumes the wait budget. The wait timeout now reports `wait_runner_restart_exhausted` with restart/session/log evidence instead of `wait_capture_stalled`.

## Validation

Added a daemon wait regression for the typed restart-exhaustion error and an Apple runner recovery wiring regression for marking failed restart paths. PR-head CI is green, including iOS runner smoke.

### DIFF
--- a/packages/contracts/src/wait.ts
+++ b/packages/contracts/src/wait.ts
@@ -13,6 +13,7 @@
 export const WAIT_REASONS = {
   captureStalled: 'wait_capture_stalled',
   deadlineExceeded: 'wait_deadline_exceeded',
+  runnerRestartExhausted: 'wait_runner_restart_exhausted',
   targetAbsent: 'wait_target_absent',
   stableTimeout: 'wait_stable_timeout',
   landmarkIdentityMismatch: 'wait_landmark_identity_mismatch',
--- a/packages/platform-apple/src/runner/__tests__/runner-recovery-wiring.test.ts
+++ b/packages/platform-apple/src/runner/__tests__/runner-recovery-wiring.test.ts
@@ -27,7 +27,10 @@ import { startFakeRunnerServer, type FakeRunnerServer } from './fake-runner-serv
 
 let server: FakeRunnerServer | undefined;
 
-const ensureRunnerSessionMock = vi.hoisted(() => vi.fn());
+const { ensureRunnerSessionMock, invalidateRunnerSessionMock } = vi.hoisted(() => ({
+  ensureRunnerSessionMock: vi.fn(),
+  invalidateRunnerSessionMock: vi.fn(async () => {}),
+}));
 
 vi.mock('../runner-session.ts', async (importOriginal) => {
   const actual = await importOriginal<typeof import('../runner-session.ts')>();
@@ -36,6 +39,7 @@ vi.mock('../runner-session.ts', async (importOriginal) => {
     // Session creation only. executeRunnerCommandWithSession stays REAL, so
     // the send → classify → recover path under test is production code.
     ensureRunnerSession: ensureRunnerSessionMock,
+    invalidateRunnerSession: invalidateRunnerSessionMock,
   };
 });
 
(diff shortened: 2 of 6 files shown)


thymikee:
Code review is clean at 50de77b. The typed reason is preserved at the runner ownership boundary, the two regression tests fail if their respective implementation seam is reverted, and exact-head CI is green (the two skipped matrix/size entries are not failures).

This is not merge-ready yet under docs/agents/pull-requests.md: attach an exact-head iOS Simulator run that deliberately invalidates/restarts the runner during wait and observes wait_runner_restart_exhausted with diagnosticId/logPath. The fake-runner and daemon tests prove the two sides independently, while the iOS smoke job does not exercise the changed failure path.

PR #2123, opened 2026-08-28
refactor: prune platform split residue

thymikee:
## Summary

Prune the dead policy, export, and routing residue left after the platform-package split. This is stacked on #2119.

- remove the unconsumed Apple `interactions` and `runner/client` package subpaths plus redundant Android IME and provision boot-diagnostics re-exports
- remove unreachable `src/platforms` workflow filters and duplicate daemon-modularity rules while preserving the retired-zone and R65 resurrection guards
- remove the orphaned Fallow suppression, repair moved baselines/docs, and lock six smaller eager import closures

## Validation

- planted-red package-boundary and routing cases failed against the pre-cleanup tree, then passed after the removals
- `node --experimental-strip-types --test scripts/gate/routing.test.ts scripts/layering/package-boundaries.test.ts scripts/layering/platform-package-policy.test.ts scripts/layering/daemon-modularity.test.ts` (63 passed)
- focused Vitest suites (461 passed)
- production-unused exports: 28 -> 26, with no unused files or unresolved imports
- `pnpm check:affected --run` passed formatting, lint, typecheck, layering, Fallow, build, package, and integration checks; the related suite passed 2,985/2,986 tests, with one unrelated timer-based gesture-policy test timing out only under four-worker load and passing 4/4 in isolation

PR #2119, opened 2026-08-28
refactor: retire platforms source seam

thymikee:
## Summary

Retire the final `src/platforms` seam and move the remaining cross-family install-source tests to their root test owner.

Remove the obsolete R3 folder policy and dead affected-test routing. R13 remains the concrete package import owner, while a new all-file retirement gate prevents the old root from returning.

Update ADR 0019 and the dependency-graph findings to record the completed physical ownership boundary, Apple’s named facades, and the durable R13/R65 split.

Closes #2082

## Validation

- Planted red: the terminal boundary test accepted a tracked file under `src/platforms` before the implementation.
- `pnpm check:affected --run` — all runnable checks passed after rebasing onto merged `main`.
- Focused policy/model suite: 82 tests passed.
- Vitest related: 2 files / 414 tests passed.
- Layering graph: 1,522 files, 0 value cycles, 0 spine backedges, and no `platforms` zone.
- Adversarial review and worker-diff reconciliation resolved a non-TypeScript fixture blind spot, stale path-ownership language, and missing Apple facade documentation.

PR #2125, opened 2026-08-28
refactor: contract Apple platform surface

thymikee:
## Summary

Contract the private Apple package root to four composition exports and move synchronous consumers through existing named facets and the platform plugin.

Remove obsolete Apple facade exports, stale root mocks, and dead Fallow suppressions. Derive platform-family policy from kernel PLATFORMS while retaining independent R11 export inventories alongside R13 ownership.

This deletes 612 lines while adding 179 across 48 files. Stacked on #2123.

## Validation

Typecheck, all 161 layering tests, Fallow, the 41-test snapshot suite, the exact file-size ratchet, and focused provider scenarios pass. A planted Apple subpath failed the independent R11 export inventory and the restored tree passed.

The full affected run passed 2,994 of 2,995 related tests. The unchanged gesture-admission timing test hit 5.01 seconds under four-worker load and passes focused in 3.04 seconds. Exact-head CI is authoritative and rerunning.


thymikee:
Review found two readiness blockers on exact head `2f6006cfba5b00dfd80567d7be235caa65c5e7a2`: Coverage fails because `snapshot-handler.test.ts` grew to 2,141 lines against its 2,140 ratchet, and the PR remains stacked on open #2123/#2119 while its own iOS smoke is not yet green. Please reduce/extract that test change, resolve the stack, and rerun exact-head CI.

The surface contraction also removes the independent real-tree Apple/Android export assertions from `package-boundaries.test.ts`, leaving R13's editable allowlist as the sole oracle despite ADR 0019 pinning exported subpaths there. Keep the independent Apple assertion and avoid the unrelated Android guard removal, or amend the ADR with equally independent proof. Finally, `./interactor` is currently a test-only facet combining `createAppleInteractor` with unrelated `setIosSetting`; please first try the existing `applePlugin`/runner-provider composition or a fixture-local adapter, which appears materially smaller and avoids expanding the contracted package surface.

thymikee:
Addressed the review on exact head 38e4e5212d:

- removed the test-only mixed-purpose ./interactor export; async fixtures now use applePlugin.createInteractor, and the synchronous provider fixture composes it with the existing withAppleRunnerProvider scope locally
- restored independent exact Apple and Android export inventories in R11; a planted ./planted-review export failed that assertion, then the restored tree passed
- reduced snapshot-handler.test.ts from 2,141 to 2,138 lines
- typecheck, 161 layering tests, Fallow, snapshot-handler, ios-alert-settings, and provider-ios-runner-transport pass

The full affected run still has only the pre-existing gesture-admission timing failure at 5.01 seconds under four-worker load; its focused run passes in 3.04 seconds. Exact-head CI is running.

thymikee:
The export-proof and `./interactor` findings are resolved on `38e4e5212d`: R11 again independently pins the exact Apple/Android real-tree exports, and the test-only transport uses the existing Apple plugin/runner-provider composition without widening the package surface.

One P1 remains: `snapshot-handler.test.ts` is now 2,138 lines but its ratchet is still pinned at 2,140. This gate requires lowering the pin when a file shrinks, so Coverage will fail until the pin is updated. Exact-head CI is still running, and the open #2123/#2119 stack remains unresolved.

thymikee:
Coverage follow-up: the job correctly required the exact file-size pin to follow the snapshot-handler reduction. Commit 165efeb835 lowers the pin from 2,140 to the measured 2,138 lines; all four focused ratchet tests pass. Exact-head CI restarted.

thymikee:
Exact-head CI is now fully green on 165efeb835, including Android and iOS smoke. The review fixes, planted-red package-boundary proof, focused validation, and full CI are complete.

thymikee:
Exact head `165efeb8358190b207d150be4e6baed62944f243` resolves the remaining ratchet blocker: the `snapshot-handler.test.ts` pin is now 2,138, matching the file. The targeted delta is otherwise empty, and Coverage, integration, package/compatibility, Repo Guards, CodeQL, and all four platform smokes are green. No code finding remains.

#2125 is clean and ready to advance in the `#2119 → #2123 → #2125` stack; it is not independently mergeable to `main` until those open parent PRs land.

PR #2122, opened 2026-08-28
fix: use canonical viewport fallback for rootless touch targeting

thymikee:
## Summary

Rootless snapshot trees now use the canonical `resolveViewportRect` fallback during actionable-touch ancestor resolution. A full-screen structural ancestor no longer steals a touch when no explicit Application/Window viewport node exists. The fallback is universal and does not infer Android from class names, backend names, or other strings.

This PR implements the core slice for #2120 and closes that ordered child of #1609. Fragment fixtures that exercise iOS-shaped promotion now include an honest full-screen Application root, preserving their original assertions without narrowing production semantics.

No Maestro changes, scroll-predicate consolidation, shared resolver semantic changes, or public CLI/wire/response changes.

Closes #2120

## Validation

- Base: `main` @ `f152827447cdd5d1e8de1f4bc0a79153564ef08d`
- Head: `a215a8caf1eeffbd4d5dad3fc15869e607377b56` (rebased onto current `origin/main`)
- Planted red on unmodified base: the rootless Android direct-target test failed 1/14 (`hittable-ancestor` actual vs `overly-broad-ancestor` expected).
- Focused green after the fixture repair and rebase: `pnpm vitest run src/core/selector-pipeline.test.ts src/commands/interaction/runtime/resolution.test.ts src/daemon/handlers/__tests__/interaction-touch-press.test.ts src/core/interaction-targeting.test.ts` passed 57/57.
- `pnpm format` completed successfully. The single final `pnpm check:affected --run` passed format, lint, typecheck, layering, fallow, and build; its related Vitest phase passed 290 files / 2,240 tests and hit one unrelated 5-second timeout in `test/integration/provider-scenarios/ios-lifecycle.test.ts` (`iOS Settings flow uses scripted simctl and runner providers`).
- Fresh exact-head GitHub CI is green: [Android Smoke](https://github.com/callstack/agent-device/actions/runs/33364192539), [iOS Smoke](https://github.com/callstack/agent-device/actions/runs/33364192560), and [CI including Coverage and Repo Guards](https://github.com/callstack/agent-device/actions/runs/33364192586) passed. Linux, macOS, CodeQL, Bundle Size, native fingerprint, Integration Tests, and Typecheck & Package also passed.
- The previous Android Smoke failure on the superseded head booted successfully and produced 17 readable captures, then timed out waiting for fixture text `Alert result: accepted`; it did not fail in actionable-touch targeting. It passed on the fresh rebased head.
[...]

Issue #1609, opened 2026-08-05
Android has no viewport-root node, so two guards are silently dead


thymikee:
## Summary

Android snapshots contain no `Application`/`Window` node, so every "resolve the root viewport" path falls through to its fallbacks — and the repo has three such resolvers with **different** fallbacks. Two of them return `null` on Android, which silently disables two guards.

Found while collapsing the nine viewport-root *predicates* onto one definition (that part is safe and shipped separately). The predicates were textual duplication; the resolvers are a real divergence — the one #1592 parked as *"resolveRootViewport is deliberately left alone … a real divergence and not the next dedup."* Measurement says it is a defect, the same way #1592's second commit found its own divergence note was covering a bug.

## Evidence

`Application` / `Window` appear as `node.type` **only in test fixtures**. Production Android emits the uiautomator `class` attribute verbatim (`src/platforms/android/ui-hierarchy.ts` → `type: attrs.className`), i.e. fully-qualified Java class names. No spelling of the viewport-root predicate matches any of them.

Running the three resolvers over an Android-shaped tree (`FrameLayout` → `RecyclerView` → `TextView`, screen 1080×2400) versus an iOS-shaped one:

```
=== android tree, resolving the viewport for a row at y=300 ===
  contracts  resolveViewportRect      : {"x":0,"y":0,"width":1080,"height":2400}
  interaction-targeting resolveRoot…  : null
  maestro    resolveRootViewport      : null

=== ios tree ===
  all three                           : {"x":0,"y":0,"width":1080,"height":2400}
```

Only `packages/contracts/src/snapshot-visibility.ts#resolveViewportRect` has the third fallback ("largest containing rect of any node"), which is the only arm that can return on Android.

## Consequences

1. **`isOverlyBroadAncestor` is dead on Android.** `src/core/interaction-targeting.ts:138` — `resolveRootViewportRect` returns `null`, the function early-returns `false`, and the viewport-sized-ancestor guard never fires. Reached from `resolveActionableTouchResolution`, which is not platform-gated.
2. **`isMaestroNodeVisible` has no on-screen check on Android.** `packages/maestro/src/internal/snapshot-policy.ts:44` — `resolveRootViewport` returns `null`, so `isVisibleInEffectiveViewport` returns `true` for every node without a scrollable ancestor.

Invisible in tests because the viewport-root concept is only ever exercised with iOS-shaped fixtures, where the root node is scenery.

## Adjacent, same sweep

[...]

PR #2156, opened 2026-08-31
fix: reuse canonical Maestro visibility context

### LINKED ISSUE #2121: Maestro: consume canonical rootless viewport visibility with reusable context
Parent: #1609

Prerequisite #2120 is merged. Implementation: #2156.

## Outcome

Maestro consumes the canonical snapshot-visibility contract through one reusable, lazy visibility resolver per real matching/scroll operation. Package-local effective-viewport and root-viewport clones are removed, and consumers do not rebuild tree context for every candidate.

This is the second ordered implementation slice of #1609.

## Decided semantics

- Rootless fallback is the canonical, platform-neutral behavior from `@agent-device/contracts/snapshot`; consumers must not infer Android or another backend from class-name strings.
- A rootless tree has no independent screen boundary. The canonical resolver may therefore derive its viewport from the largest containing geometry, including the candidate itself. A positive-geometry target cannot be required to fall outside that target-derived viewport.
- When neither an explicit root/scroll clip nor another target-independent boundary exists, visibility remains fail-open. This preserves the universal fallback agreed in #1609 instead of introducing a Maestro-only negative.
- Invalid explicit roots also fail open; they are not permission to guess a viewport from platform vocabulary.
- Maestro's platform-specific evidence rules remain policy inputs: Android `visibleToUser === false`, geometryless anchor behavior, and iOS hittability handling are not folded into a second geometry system.
- Complexity is measured at the operation boundary: one resolver/context construction and one root/tree scan per ranking or scroll operation. This issue does not promise linear complexity for every target-dependent containing-rect query and does not add a package-local spatial index.

The former acceptance criterion requiring a rootless target to be outside an unspecified "effective canonical viewport" was contradictory: without an explicit root, scroll clip, or other target-independent boundary, the canonical API has no such viewport. That criterion is intentionally removed rather than implemented with Android/class heuristics.

## Scope

- Provide the canonical reusable visibility resolver at the shared snapshot contract seam.
- Make resolver construction lazy so empty operations do not build visibility context.
- Thread one resolver through Maestro ranking, filtering, and scroll operations.
- Migrate direct shared consumers to the canonical resolver where needed to keep one ownership model.
- Remove Maestro's local effective-viewport/root-viewport implementation and obsolete geometry helpers.
- Preserve existing anchor, scroll-ancestor clipping, iOS hittability, and invalid-root behavior.

Out of scope:

- A target-independent screen/surface-observation boundary for rootless captures
- Android/backend/class-name inference
- Consolidating `isScrollingContainer` / `isScrollableNodeLike` vocabularies
- A new spatial index or compatibility fallback
- Public Maestro program syntax or result-shape changes

## Required tests

- Universal rootless fallback remains visible when no explicit root or scroll clip supplies a target-independent boundary.
- Explicit root and nearest-scroll-ancestor clipping retain their existing behavior.
- Geometryless Android and iOS anchor cases retain their platform-specific outcomes.
- Invalid explicit roots fail open.
- Real ranking/filtering and scroll operation probes prove one lazy resolver/context per operation, rather than one build per candidate; empty operations do not construct it.
- New ownership/complexity invariants have observed planted-red evidence. Do not substitute elapsed wall-time assertions.

Use emitted per-platform fixture vocabularies. Do not add the deferred scroll-predicate consolidation or claim parity from one shared synthetic vocabulary.

## Validation

During iteration, run the smallest owning contract and Maestro tests. Before pushing, run:

```sh
pnpm format
pnpm check:affected --run
```

Live device evidence is required only for behavior a device can distinguish. Do not claim a rootless-outside device proof when the canonical contract intentionally has no independent boundary for that state; report any other unavailable device evidence exactly.

## Done criteria

- The canonical resolver owns effective-viewport visibility and is reused once per real operation.
- Maestro contains no local `resolveRootViewport` or `isVisibleInEffectiveViewport` clone.
- Universal rootless fail-open behavior and explicit clipping behavior are covered separately.
- No platform/backend/class-name inference, compatibility fallback, or second geometry system exists.
- Focused tests and the final affected gate pass.
- The PR reports planted red, focused green, final gate, live evidence or exact absence, size, and risks.

## Stop conditions

Stop and open a separate design issue instead of improvising if correctness requires a new target-independent rootless screen/surface boundary or spatial-index contract. Such a boundary must be typed by capture; a consumer must not guess it.

thymikee:
## Summary

Moves effective snapshot visibility ownership into `@agent-device/contracts/snapshot` behind one opaque, factory-created, lazy resolver bound to a snapshot tree. Consumers can no longer normally assemble mismatched nodes, node maps, and viewport roots, and memoized indexes/root scans are materialized only when a visibility path needs them.

Effective boolean visibility now treats a candidate with no clipping ancestor and no valid explicit viewport root as visible without invoking the target-dependent containing-rectangle fallback. `resolveViewportRect` still preserves that fallback for callers that need the actual rectangle. Maestro ranking and scroll operations each construct one resolver and reuse its node map and canonical viewport selection; the package-local visibility context and runtime-port geometry clones are removed. Selectors, capture-kit, and root interaction callers were migrated to the same canonical interface.

This preserves Android `visibleToUser === false`, geometryless anchors, iOS hittability/direct-call behavior, nearest-scroll-ancestor clipping, emitted platform-specific scroll vocabularies, and Maestro syntax/result shapes. It completes the second ordered slice of #1609 and closes #2121. #2122 is merged, and this branch is rebased directly onto current `main`.

The owner-authorized contract for this slice supersedes #2121's stale rootless-outside rejection assertion. A rootless snapshot contains no target-independent viewport, and canonical `resolveViewportRect` may select the target's own positive containing rectangle; making that target invisible would therefore require changing shared viewport semantics or introducing prohibited platform inference. This PR intentionally pins canonical rootless fail-open behavior and does not claim the impossible negative regression.

## Validation

- Exact base: `a79f0d2e81d0eec479bf898f697e9dcda53aaac7`
- Exact head: `bf9d351d905fcfbdda9fce33029ca15c3e93b2b0`
- Ancestry: `origin/main` is an ancestor; `origin/main..HEAD` contains only the two original #2121 commits plus `refactor: centralize snapshot visibility context`.
- Scope: 27 files, 621 additions, 384 deletions, limited to the canonical snapshot-visibility call graph and its focused tests/fixtures. No docs or skills changed because public command behavior is unchanged.
- Planted red — contracts: `pnpm vitest run packages/contracts/src/snapshot-visibility.test.ts` failed 3/3 before the canonical factory existed; the contracts visibility group then passed 17/17.
- Planted red — real Maestro ranking: `pnpm vitest run packages/maestro/src/internal/__tests__/runtime-target-ranking.test.ts` failed 2/10 because the operation probe observed no shared context/map/viewport ownership; it then passed.
- Planted red — real Maestro scroll geometry: `pnpm vitest run packages/maestro/src/internal/__tests__/runtime-port-geometry.test.ts` failed 7/13 during the ownership migration, including the operation probe; it then passed.
- Planted red — hostile rootless batch: `pnpm vitest run packages/capture-kit/src/mobile-snapshot-semantics.test.ts` observed 8,511 geometry reads for 64 nodes against a `<512` invariant; it then passed through the canonical lazy resolver.
- Post-rebase focused validation: 136/136 contracts/Maestro/selectors/capture-kit tests plus 197/197 root/selector-read tests, 333/333 total; focused package and root TypeScript checks passed.
- Pre-final gate attempt 1 correctly failed TypeScript on one missed `resolution.ts` migration. That caller was changed to use the operation-owned resolver and its focused TypeScript plus 23/23 resolution tests passed.
- Pre-final gate attempt 2 ran under host contention: 5,040 tests passed and 18 unrelated subprocess-backed files timed out. No production code was changed for those timeouts.
- First quiet-host gate passed 5,057/5,058 tests with one timeout in `gesture-admission-parity.test.ts`; that file passed 4/4 both on this branch and current `main` in isolated comparable runs, classifying the timeout as infrastructure contention.
- Final rebased quiet-host gate: `pnpm format` passed, then `pnpm check:affected --run` passed formatting, lint, typecheck, layering, changed-file fallow audit, build, and all 655 files / 5,058 tests. The gesture test passed in-suite.
- Historical iOS Smoke evidence at the earlier head failed while seeding an email with `TEXT_INPUT_COMMIT_NOT_OBSERVED` in `smoke:form-input`; it did not exercise this visibility path. Fresh exact-head CI is the authoritative native lane.
- Live-device evidence: the earlier dedicated, stopped `Pixel_7_CI` AVD was used without taking over the user-owned Pixel 7 and proved rootless snapshot, `scrollUntilVisible`, and `tapOn`; the later scenario stopped on an unrelated shipping/pickup assertion. No new device run was performed for this contracts-owned refactor, so there is no new live iOS evidence.
- Main residual risk is the breadth of the typed migration across existing visibility consumers; the branded construction path, deterministic probes, hostile-tree operation count, focused 333/333 suite, and final affected gate cover that seam without changing shared viewport semantics or introducing a spatial index.


thymikee:
## Deep code-quality review

The per-candidate rebuild fix is real and large — I measured it on this head: filtering 1,999 matches in a 2,000-node Android tree goes from 1,982 ms to 88 ms. And a differential fuzz (3,000 random trees, 38,808 node×platform decisions) comparing the deleted clone against the canonical helpers found zero divergences on well-formed trees. The only divergence class is a degenerate Application/Window rect (non-finite coordinate, or non-positive area), where the new code fails open instead of hiding the whole tree — an improvement, but nothing pins it, and "behavior preserved" isn't literally true.

That said, I don't think this lands where it's aiming. Two things block for me.

---

### 1. On Android the hot path is still O(N²), and the quadratic term provably cannot change any answer

Measured on this head, same tree shape, only the root node's `type` differs:

| nodes | `Application` root | no viewport root (= every Android tree) |
|---:|---:|---:|
| 800 | 0.51 ms | 9.81 ms |
| 2000 | 1.18 ms | 56.38 ms |

Left column is linear. Right column is quadratic, and 48× slower at 2,000 nodes.

`snapshot-visibility.ts` documents why: Android emits fully-qualified Java class names and no root node, so `collectViewportRects` is **always empty on Android**. `resolveViewportRect` then falls through to its third fallback — `nodes.flatMap(...)` over the whole tree — once per candidate. Hoisting `viewportRects` removed one whole-tree scan per node; this is the other one, and it's the dominant one.

It also cannot affect the result. The fallback picks the largest rect *containing the target's center*; any rect containing the center overlaps the target, so `isRectVisibleInViewport` is unconditionally `true`. The target's own rect always qualifies, so the list is never empty. The fuzz confirms it: no divergence from the old `return null` → `true`.

So on Android this reduces to exactly what the deleted clone already said — no scrollable ancestor + no viewport root ⇒ visible — at O(N) per node instead of O(1). The judo move is a guard in `isNodeVisibleInEffectiveViewport` (`packages/contracts/src/snapshot-visibility.ts:90`): when there is no clipping ancestor and `viewportRects` is empty, return `true`. That takes Android from 56 ms to ~1 ms and **deletes** the branch rather than hoisting around it.

Scope it to that predicate, not to `resolveViewportRect` — callers that use the returned *rect* (`src/core/interaction-targeting.ts:225`, `src/commands/interaction/runtime/resolution.ts`) are what #2122 added the fallback for.

The context object isn't the lever here. This is.

### 2. The headline invariant isn't achieved at the primary call site — I measured 2 node maps per operation, not 1

Instrumenting `buildSnapshotNodeMap` / `collectViewportRects` on this head, one `rankMaestroCandidates` call reports `{nodeMap: 2, viewportRects: 1}`. (main reports `{nodeMap: 3}` for two candidates — so the *scaling* fix is real, but the constant is 2.)

`runtime-target-ranking.ts:36` builds the resolver, which builds `buildSnapshotNodeMap(snapshot.nodes)` at `runtime-selector-resolution.ts:46` and exposes it as `resolver.nodeByIndex` — and line 52 already hands that same map to `rankVisibleMaestroMatches`. Line 38's `filterVisibleMaestroMatches({nodes: snapshot.nodes, ...})` builds a second map for the same array, two lines apart from the one that's already in scope.

The `{flatMap: 1, map: 1}` test only passes because it calls `filterVisibleMaestroMatches` directly and bypasses the resolver. It reads as pinning a per-operation invariant that the operation does not hold. Derive the visibility context from the resolver (or let it accept a `nodeByIndex`) so the operation's one map *is* the map.

### 3. The clone removal stopped one file short

`runtime-port-geometry.ts` is in the same call path and still has both halves of what this PR deleted:

- `:83` — `new Map(snapshot.nodes.map((node) => [node.index, node]))`, a hand-inlined `buildSnapshotNodeMap`, in the same function that called `filterVisibleMaestroMatches` on line 64 and threw that exact map away.
- `:103` — `findLargestViewportRect` is `pickLargestRect(collectViewportRects(nodes))` written longhand: filter on `isViewportRootNode`, sort by area descending, take `[0]`, with `left.rect!` / `right.rect!` non-null assertions. That is the same clone just deleted from `snapshot-policy.ts`, one file over.

One scroll-gesture operation does 3 whole-tree node-map builds and 2 whole-tree viewport-root scans. If the PR is "reuse the canonical contract", this file is in scope.

### 4. `MaestroVisibilityContext` is the third hand-rolled copy of a memo the canonical layer should own — and the trap that produced all three is still open

```ts
isNodeVisibleInEffectiveViewport(node, nodes, byIndex = buildSnapshotNodeMap(nodes), precomputedViewportRects?)
```

A defaulted third parameter that silently rebuilds the map, and an optional fourth. Omitting either is not a type error; it is a silent O(N) per call. That signature has now produced three separate local fixes:

- `packages/selectors/src/internal/resolve.ts:233-238` — hand-rolled `let byIndex; let viewportRects; ??=`, comment citing #1970
- this PR — `MaestroVisibilityContext` + factory
- and it is still live at `packages/capture-kit/src/mobile-snapshot-semantics.ts:76`, which passes `byIndex` but not `viewportRects` inside `for (const node of nodes)` over the entire tree — the same O(N²) this PR is fixing, applied to every node in the snapshot — and at `packages/selectors/src/internal/predicates.ts:136`, which passes neither.

What enumerates N here is the four-positional-arg signature. One lazy context in `@agent-device/contracts/snapshot` (fields built on first read, like selectors already does by hand) deletes Maestro's new type, deletes selectors' `??=` pair, closes capture-kit's instance, and makes forgetting the memo unrepresentable instead of silent. A Maestro-local bundle of the same three arguments normalizes the drift rather than fixing it.

Related: nothing structurally ties `nodeByIndex` / `viewportRects` to `nodes` — the type is structural and constructible by hand with a map from a different tree. A resolver closure (`createVisibility(nodes) => (node, platform) => boolean`) makes that unrepresentable and drops the second positional argument from every call site.

### 5. The empty-match guard is a special case laziness deletes — along with its test

`runtime-target-policy.ts:52` (`if (params.matches.length === 0) return [];`), its dedicated test, and the `observeTreeTraversals` Proxy exist only because the context is built eagerly. A lazy context removes the guard *and* covers what the guard misses: an all-`visibleToUser: false` Android match set, or an all-geometryless-hittable iOS set, still pays for a node map and a viewport scan it never reads. The guard handles one shape of "no work"; laziness handles all of them, with no branch.

### 6. The new test asserts on `Array.prototype` call counts when the package already has the right idiom

`observeTreeTraversals` counts `flatMap` / `map` on the nodes array, which couples Maestro's tests to *which array methods the contracts helpers happen to use*. Rewrite `collectViewportRects`'s `flatMap` as a `for` loop — a pure perf change in another package — and this goes red with no behavior change. The assertion is also indirect: `{flatMap: 1, map: 1}` only means "one context" if you know which helper uses which method.

`runtime-selector-resolution.ts` already solved this in this package: `MaestroResolutionProbe.onNodeMapBuilt`, asserted as `expect(nodeMapBuilds).toBe(1)` in `runtime-selector-resolution.test.ts:33`. Local, robust, and phrased in the vocabulary of the thing under test. Reuse that shape (or hang a counter on the context factory) instead of a 20-line Proxy over the standard library.

### 7. Single-consumer fixtures in the shared fixtures module

The four `*_VISIBILITY_FIXTURE` constants are imported only by `snapshot-policy.test.ts`, but they live in `runtime-target-fixtures.ts`, taking it from 10 to 98 lines. `makeSnapshot` is genuinely shared; these are not. Move them next to their only reader — the trees under test should be readable without a second file.

---

**Verdict:** not ready. (1) is the blocker: the dominant cost on the one platform this was live-validated against is untouched, and it's inert code. (2) means the PR's stated deliverable and the test presented as proving it don't match the call graph. (3) and (4) are the structural asks — if the fix lands in contracts as one lazy context, Maestro's new type, selectors' `??=` pair, the empty-match guard, and capture-kit's live instance all go away together, and this becomes a much smaller and much better diff.

*All measurements taken on `48d246cd` applied over `caa3dc23`, via `vitest` with `buildSnapshotNodeMap` / `collectViewportRects` instrumented; the differential fuzz compared the deleted `main` implementation against `isMaestroNodeVisible` node-for-node.*

thymikee:
Reviewed exact head `48d246cd6793587751c3d978c44fabcbb050ee6e`. The ownership/performance refactor itself is sound, but this cannot close #2121 as currently specified.

**Blocking scope/contract mismatch:** #2121 requires planted-red proof that a rootless, no-scroll target outside the canonical viewport changes from visible to hidden. This PR's planted reds prove only one-context construction and the empty-match fast path; its rootless test asserts `true`, which the old clone also returned. Under the current canonical contract, `resolveViewportRect` falls back to the largest containing rect from every node—including the target itself—so a lone rootless target can resolve against its own rect and remains visible. That required negative is not implementable without an authorized shared-contract semantic change. Please either re-scope/amend #2121 and remove the behavioral closure claim, or define/approve the contracts change and add a genuine before/after counterexample. Do not invent a Maestro-local exception.

Exact-head iOS Smoke is also red with `TEXT_INPUT_COMMIT_NOT_OBSERVED` while filling `field-email`; the other lanes are green, but the head remains `UNSTABLE` until that lane is rerun or classified. No further emulator run can substitute for the contract decision above.

thymikee:
Addressed the deep visibility-ownership review on exact head `bf9d351d905fcfbdda9fce33029ca15c3e93b2b0`:

1. Effective boolean visibility now short-circuits candidates with no scroll-clipping ancestor and no valid explicit viewport roots as visible. Degenerate, non-finite, and non-positive Application/Window roots fail open. `resolveViewportRect` retains its containing-rectangle fallback for callers needing a rectangle.
2. The real `rankMaestroCandidates` operation creates one canonical resolver and reuses its lazily built node map and viewport roots. `MaestroResolutionProbe` proves one context, one node map, and one viewport scan per operation.
3. `runtime-port-geometry.ts` no longer builds its own node map or scans viewport roots longhand. The real scroll operation reuses its operation-owned canonical resolver while retaining distinct Android/iOS scroll vocabularies.
4. The contracts-owned branded factory supersedes and deletes `MaestroVisibilityContext`. Selectors, capture-kit, Maestro presentation, ranking, scroll, and root interaction callers were migrated; the old optional/defaulted visibility API is gone.
5. Internals are lazy. Empty matches, all Android `visibleToUser:false` candidates, and direct geometryless-hittable iOS candidates do not materialize unused maps/root scans. The special empty-match guard and its implementation-coupled test were deleted.
6. `observeTreeTraversals` was removed. Complexity is asserted through canonical resolver probes, real Maestro operation probes, and a hostile rootless batch that dropped from the planted-red 8,511 geometry reads for 64 nodes to the bounded invariant.
7. Single-consumer visibility fixtures moved beside `snapshot-policy.test.ts`; only shared builders remain in the fixture module.

The canonical rootless positive-geometry behavior remains intact; I did not add the superseded impossible outside-rootless negative assertion or alter shared viewport semantics.

Evidence: planted reds were observed at contracts, real ranking, real scroll, and hostile capture-kit seams; post-rebase focused tests pass 333/333. Two earlier full-gate attempts were red for, respectively, a real missed TypeScript migration and then host-contention timeouts. A subsequent quiet run had one gesture-admission timeout that passed 4/4 in isolation on both branch and current main. On the final rebased diff, `pnpm format` and `pnpm check:affected --run` passed all stages and all 655 files / 5,058 tests.

thymikee:
Re-reviewed exact head `bf9d351d905fcfbdda9fce33029ca15c3e93b2b0`. The prior #2121 contract mismatch remains. The new factory moves ownership to the right seam, but `resolveEffectiveViewport()` returns null without an explicit Application/Window root and `isVisibleInEffectiveViewport()` treats null as visible; Maestro now consumes that boolean. A rootless FrameLayout target outside the frame with no scrollable ancestor is therefore still retained. The new contract test codifies rootless fail-open behavior and lacks the required outside-root case; planted-red evidence proves resolver ownership/performance, not the motivating failure. Establish a target-independent rootless viewport at the canonical contract without Android/class heuristics, or revise #2121 scope. Add observed-red Maestro-ranking coverage for rootless-outside, then exact-head real rootless scroll/tap device proof. CI is green, but this P1 blocks readiness.

thymikee:
I re-audited this against exact head `bf9d351d905fcfbdda9fce33029ca15c3e93b2b0`. No code change is appropriate for this finding because it asks for the rootless-negative assertion that the owner-authorized contract for this slice explicitly supersedes.

The canonical facts are:

- A rootless snapshot supplies no target-independent viewport rectangle.
- `resolveViewportRect` must retain #2122's containing-rectangle fallback for rectangle consumers, and that fallback may validly select the candidate's own positive rectangle.
- Consequently, a positive-geometry node with no clipping ancestor and no explicit viewport root cannot be proven outside the viewport from this snapshot. Effective boolean visibility therefore fails open without running the target-dependent fallback; the exact-head contracts test pins that behavior and proves the fallback is not invoked per candidate.
- Making the requested FrameLayout invisible would require a new viewport semantic, Android/backend/class-name inference, or a Maestro-local second geometry system. All are outside and contrary to the corrected contract.

The planted-red evidence for this revised slice is intentionally ownership and complexity evidence: canonical resolver construction, one operation-owned map/root scan, lazy early-return paths, and bounded hostile-rootless batch work. The existing dedicated Android proof confirms rootless snapshot, scroll, and tap continue to work; no device can prove a negative that the canonical snapshot does not encode.

I have updated the PR description to state this supersession explicitly. I am not editing #2121 itself, changing shared viewport semantics, or inventing the impossible rootless-outside assertion. The exact head remains unchanged and CI-green.

thymikee:
Addressed by taking the second path from this review: [#2121's scope and acceptance criteria have been revised](https://github.com/callstack/agent-device/issues/2121).

The previous issue text was contradictory: it prohibited changing canonical viewport semantics while requiring a rootless target to be outside an unspecified canonical viewport. Under the agreed universal fallback, a rootless tree has no independent screen boundary; the canonical resolver may derive the viewport from the largest containing geometry, including the candidate itself. With no explicit root, scroll clip, or typed surface boundary, visibility intentionally fails open.

#2121 now makes that contract explicit and requires the evidence this PR actually provides at exact head `bf9d351d905fcfbdda9fce33029ca15c3e93b2b0`:

- one contracts-owned lazy visibility resolver;
- one resolver/context and root scan per real ranking or scroll operation;
- no construction for empty operations;
- separate coverage for universal rootless fail-open, explicit-root and scroll-ancestor clipping, geometryless anchors, and invalid roots;
- no Android/class-name heuristic, Maestro-local geometry system, or spatial-index fallback.

I did not add the requested rootless-outside Maestro/device case because no target-independent boundary exists to make that state representable. If such a boundary is needed later, capture must supply it as a typed surface-observation contract in a separate design slice; Maestro must not infer it.

No production or test change was needed after correcting the owning issue contract. Its state and labels are unchanged.

PR #2160, opened 2026-08-31
feat(apple): reclaim retained runners under device-claim authority

thymikee:
## Summary

Implements the #1320 retained-runner rule (the issue's v1 stop-and-recreate decision): a daemon that holds the host-global device claim may stop and replace a warm XCTest runner whose owner no longer holds that claim.

Before this, the worst remaining cross-worktree failure on iOS was: worktree A closes its session (claim released, runner kept warm by default), worktree B wins the claim at `open`, then dies in runner preparation with an unstructured `COMMAND_FAILED` / `IOS_RUNNER_OWNED_BY_OTHER_DAEMON` for up to ~5 minutes until A's daemon idled out — exactly the late runner-prep contention #1320 set out to remove.

## How it works

- **Claim authority proves release.** Device claims are exclusive per device, so the requesting daemon holding the claim proves the lease owner released the device and merely kept its runner warm. `prepareRunnerLeaseForStartup`'s busy branch gains a third takeover rule (`device-claim-takeover`) alongside same-state-dir and proxy logical-lease.
- **Trust boundary respected.** The probe is not request metadata: it is daemon-bound through the existing runner-owner seam (`@agent-device/platform-apple/runner-owner`, same pattern as `setRunnerLeaseOwnerStateDir`) and answers from the claim store by process identity (`processOwnsActiveDeviceClaim`). Unbound embedders (package tests, non-daemon use) answer false and keep today's refusal. Abandoned claims grant no authority.
- **Mixed-version safety.** Leases now record `deviceClaimProtocol: 1`; takeover is gated on the lease declaring claim arbitration, so a runner owned by a pre-claims build — which can legitimately be mid-session without holding any claim — is never preempted.
- **Loser-side safety.** `cleanupRunnerSessionResources` now skips the device-wide `simctl terminate` of the runner container bundles when the on-disk lease is owned by someone else. Without this, the losing daemon's delayed idle-stop or shutdown would kill the successor's active runner on the shared simulator (this also closes the same latent hazard for the proxy logical-lease takeover path). Lease release and token-scoped xcodebuild cleanup already no-op on token mismatch.
- Help topics (`workflow`, `physical-device`) updated: the live-owner rejection now names claim arbitration instead of being unconditional.

## Validation

`pnpm check:affected --run` passed (format, lint, typecheck, layering, fallow, affected tests).

Live two-daemon validation against a throwaway `iPhone 17 Pro` simulator with a shared claim store and shared lease dir:
1. **Protection intact:** while daemon A's session held the claim, daemon B's `open` was rejected with non-retriable `DEVICE_IN_USE`, reason `DEVICE_CLAIM_LIVE_OWNER`, owner session/workspace/stateDir, and the exact `device status` recovery command.
2. **Takeover works:** after A's `close` (claim gone, lease retained with `deviceClaimProtocol: 1`, daemon A still alive), B's `open` succeeded in ~8s and the on-disk lease flipped to B's daemon pid while A was still running — the live-owner path, not stale reclaim.
3. **Guard works:** `daemon stop` on A afterwards left B's session fully functional (`appstate` healthy).

Part of #1320.

### DIFF
--- a/packages/platform-apple/src/core/runner-host.ts
+++ b/packages/platform-apple/src/core/runner-host.ts
@@ -88,5 +88,5 @@ export const appleRunnerHost: AppleRunnerHost = {
   resolveIosPhysicalDeviceControl,
   visitXmlPlistEntries,
   leaseOwnerStateDir: getRunnerLeaseOwnerStateDir,
-  hasDeviceClaimAuthority: (deviceId) => getRunnerDeviceClaimAuthorityProbe()?.(deviceId) ?? false,
+  hasDeviceClaimAuthority: (device) => getRunnerDeviceClaimAuthorityProbe()?.(device) ?? false,
 };
--- a/packages/platform-apple/src/core/runner-host.ts
+++ b/packages/platform-apple/src/core/runner-host.ts
@@ -36,7 +36,10 @@ import {
 import { bootFailureHint, classifyBootFailure } from '@agent-device/provision-kit/boot-diagnostics';
 import { resolveIosPhysicalDeviceControl } from './physical-device-control.ts';
 import { visitXmlPlistEntries } from './plist-xml.ts';
-import { getRunnerLeaseOwnerStateDir } from './runner-owner-state.ts';
+import {
+  getRunnerDeviceClaimAuthorityProbe,
+  getRunnerLeaseOwnerStateDir,
+} from './runner-owner-state.ts';
 import { buildSimctlArgsForDevice } from './simctl.ts';
 import { readApplePlistJson, runAppleToolCommand, runXcrun } from './tool-provider.ts';
 
@@ -85,4 +88,5 @@ export const appleRunnerHost: AppleRunnerHost = {
   resolveIosPhysicalDeviceControl,
   visitXmlPlistEntries,
   leaseOwnerStateDir: getRunnerLeaseOwnerStateDir,
+  hasDeviceClaimAuthority: (deviceId) => getRunnerDeviceClaimAuthorityProbe()?.(deviceId) ?? false,
 };
--- a/packages/platform-apple/src/core/runner-owner-state.ts
+++ b/packages/platform-apple/src/core/runner-owner-state.ts
@@ -13,3 +13,23 @@ export function setRunnerLeaseOwnerStateDir(stateDir: string | undefined): void 
 export function getRunnerLeaseOwnerStateDir(): string | undefined {
   return runnerLeaseOwnerStateDir;
 }
+
+/**
+ * Daemon-owned device-claim arbitration probe: does the embedding process hold
+ * the host-global local device claim for this device id right now? Unbound (in
+ * embedders that run no claim store, such as package tests) it answers false,
+ * which keeps runner-lease takeover disabled there.
+ */
+export type RunnerDeviceClaimAuthorityProbe = (deviceId: string) => boolean;
+
+let runnerDeviceClaimAuthorityProbe: RunnerDeviceClaimAuthorityProbe | undefined;
+
+export function setRunnerDeviceClaimAuthorityProbe(
+  probe: RunnerDeviceClaimAuthorityProbe | undefined,
+): void {
+  runnerDeviceClaimAuthorityProbe = probe;
+}
+
+export function getRunnerDeviceClaimAuthorityProbe(): RunnerDeviceClaimAuthorityProbe | undefined {
+  return runnerDeviceClaimAuthorityProbe;
+}
--- a/packages/platform-apple/src/core/runner-owner-state.ts
+++ b/packages/platform-apple/src/core/runner-owner-state.ts
@@ -1,3 +1,5 @@
+import type { DeviceInfo } from '@agent-device/kernel/device';
+
 /**
  * Daemon-owned lease-owner state directory for the Apple runner. Kept in a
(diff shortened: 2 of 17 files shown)


thymikee:
Reviewed exact head `7f3a8386def3ee38282e4f92bba76692e3ff8ced`. One blocking ownership race remains.

**P1 — loser cleanup is not atomic with successor startup.** `cleanupRunnerSessionResources()` reads the lease token and then asynchronously terminates device-global runner bundles without `withRunnerLeaseLock`. A successor holds that lock, but launches xcodebuild before publishing its replacement token. Interleaving: A cleanup reads A's token; B reclaims and launches; A then terminates the shared runner bundle before B publishes its token (and the check-to-terminate gap remains even afterward), killing B. The new test only starts A disposal after the successor token already exists, so it cannot exercise this race.

Serialize token-check plus device-wide runner termination through the runner-lease owner boundary, including shutdown/abort cleanup. Add a deterministic two-owner regression that pauses A after its ownership check, completes B's reclaim/launch/publication, then releases A and proves no successor termination plus successful B `appstate`. Repeat the live two-daemon simulator test with overlapping A shutdown/B startup, and rerun exact-head device/coverage lanes.

thymikee:
Re-reviewed exact head `2eb41981198acb001f44958659ce52383ccf1fe4`. The P1 runner-termination TOCTOU remains. `cleanupRunnerSessionResources` reads the current lease token, then asynchronously performs device-global runner bundle termination without holding the runner-lease lock. A loser can pass the check, a successor can reclaim/start/publish, and the loser can terminate the successor. Make ownership verification and device-wide termination one lease-locked operation on every disposal path. Add a deterministic two-owner regression that pauses loser cleanup after its check, completes successor reclaim/start/lease publication, then proves loser release cannot terminate the successor. The new policy-seam tests do not exercise this overlap. Exact-head iOS Smoke also failed `TEXT_INPUT_COMMIT_NOT_OBSERVED`; its rerun must green or be classified after the P1 repair.

thymikee:
P1 addressed in `e6cfe4b172`.

**Fix.** The ownership check, the device-wide runner container-app termination, and the lease release now run as one operation under the runner-lease lock in `settleOwnedRunnerDeviceState` — the same lock a successor holds for its entire reclaim-and-publish window (`ensureRunnerSession` wraps `startRunnerSessionWithLease`, including `writeRunnerLease`, in `withRunnerLeaseLock`). So a loser either enters before the successor's window (sees its own token, terminates only its own runner) or after publication (sees the successor's token, does nothing device-wide); the check-to-terminate gap is gone. Every disposal path routes through it: normal dispose, session invalidation, reusable-session teardown, idle stop, and daemon-shutdown `abortRunnerSessionsAndPrepProcesses`. Callers that already execute inside the lease lock (startup abort, lease-publish failure, `stopIosRunnerSession`) declare `leaseLockHeld` instead of re-acquiring the non-reentrant lock. If the lock cannot be acquired, all device-wide teardown is skipped fail-safe: whoever owns the lease settles that state, and an own unreleased lease turns stale on process exit. The lease release moved inside the locked section too — the previous token-guarded read-then-unlink had the same (narrower) race and could have unlinked a successor's freshly published lease.

**Deterministic two-owner regression** (`runner-disposal.test.ts`, "disposal serializes behind a successor reclaim window"): the test holds the real lease lock as the successor's critical section, starts loser disposal inside that window while the on-disk lease still names the loser, asserts the disposal has not settled and no termination fired mid-window, completes the takeover (release + publish successor token), exits the window, and proves the loser performed zero container-app terminations and left the successor's lease untouched. **Observed red against the pre-fix code** (the loser passed its check inside the held window and fired terminations), green with the fix.

**Live re-validation with overlapping shutdown/startup**, twice, on a throwaway `iPhone 17 Pro` simulator with a shared claim store and lease dir: A open → close (runner retained) → B's takeover `open` started, then `daemon stop --state-dir A` issued 0.7s (round 1) and 0.2s (round 2, roles inverted) into it. Both rounds: the takeover open succeeded, the stop reported graceful, and the winner's session answered `appstate` healthily afterwards.

`pnpm check:affected --run` passed on the new head (the full apple-runner project is 330/330 including the request-cancellation suite, which exercises the in-lock disposal path and would deadlock if the flag were wrong). Exact-head CI lanes are re-running on the push; the earlier `TEXT_INPUT_COMMIT_NOT_OBSERVED` iOS Smoke greened on rerun before this push (known dropped-keystroke flake signature) and gets a fresh full run now.

thymikee:
One P1 finding still blocks readiness at `e6cfe4b172ee4a0dea939fee62c7074678d1d36b`:

Runner-takeover authority is granted from any active claim with the same raw `deviceId`, while claims are canonically scoped by platform family + Apple OS + id. An Android claim whose serial equals an Apple runner id can therefore authorize destructive takeover of the live Apple runner. The current authority-probe regression actually uses an Android fixture, so it blesses the invalid cross-platform case.

Pass the Apple `DeviceIdentity`/canonical local device key into the authority probe and require exact family/OS/id ownership. Add an adversarial same-id Android-vs-Apple regression proving the Android claim cannot reclaim the Apple runner.

The prior disposal TOCTOU is otherwise addressed by holding the runner-lease lock across check, device-wide termination, and release. Exact-head iOS Smoke was still running at review time, but that is separate from this confirmed code finding.

No `ready-for-human` label until fixed on a new head.

PR #2162, opened 2026-08-31
feat: add stale device claim release and dead-end recovery guidance

thymikee:
## Summary

Stacked on #2160. Closes the #1320 recovery loop for claims no daemon can settle on its own — the corrected scope's `device release` closure criterion (stale-only; no `--force`, per the 2026-07-24 scope decision).

- **`agent-device device release --stale`** settles a provably dead owner's durable resources through the same exact-owner reconciliation `open` and daemon startup use, then clears the claim last — daemonlessly. The CLI composes a local-only platform gateway (provider runtimes stay out entirely; claims exist only for local devices) plus the canonical `createDeviceClaimReconciler`/`createDaemonRecoveryPlatformScope`. Live, uncertain, PID-reused, and corrupt claims always fail closed and are reported per claim with the reason and a next step; releasing runs under the per-device claim lock with an ownerToken re-check, so a concurrent owner transition reports `changed` instead of racing.
- **Dead-end removal in the error contract:** a `DEVICE_IN_USE` conflict whose recorded owner provably cannot release (dead, or superseded daemon) now carries the exact `device release --stale` command as its recovery, instead of a status inspection that led nowhere. Live/uncertain owners keep inspection-only recovery.
- **`device status --stale`** now appends the matching release command when provably dead owners are listed.
- **`daemon stop`** now warns in text output when a claim was orphaned (previously visible only via `--json`) and names the status/release commands.

## Validation

`pnpm check:affected --run`: every check green except one provider-scenario test timeout (`ios-lifecycle.test.ts`, 5s budget) that reproduces identically on `origin/main` under the same load — a parallel agent session was running a full suite on this host (load avg ~29); this is the documented host-contention signature, not a regression.

Real-binary validation (`pnpm build` + isolated `AGENT_DEVICE_CLAIMS_DIR`): planted a dead-owner claim (hashed path, dead PID, existing state dir) —
1. `device status --stale` listed it and offered the exact release command;
2. `device release` without `--stale` refused with `INVALID_ARGS` and the live-owner guidance;
3. `device release --stale` reported `released ios Dead Validation iPhone session=dead-validate …` and the claim file was gone.

Part of #1320.


thymikee:
Reviewed exact head `2d685ffedcf011b2ca5621679438ef263d86e157` (stacked on current #2160). The stale-release core is fail-closed, but this is not ready: (1) the claimed practical proof only deletes an empty synthetic dead claim; it does not exercise CLI → local gateway → exact-owner durable-resource recovery, a real simulator/emulator, or two-daemon recovery. Add an end-to-end CLI regression with separate caller/owner state dirs and a durable resource record, proving cleanup completes before claim deletion, plus a race assertion where the claim changes after scan and before lock acquisition. (2) Run the real recovery path: daemon A owns a device and leaves an attributable resource/claim after abrupt death; daemon B runs `device release --stale`, verifies cleanup/claim removal, then opens/uses the same device. (3) Drop/move the unrelated 89-line #2160 runner-session test delta (`adf5080`) so this stacked unit stays focused. Require exact-head CI and parent #2160 green.

thymikee:
Correction: the current exact head is `da8c9fe076283471eb492e6fe99fce478c46334f`; my prior comment named the pre-rewrite composite head. The fail-closed assessment and the two recovery-proof requirements apply unchanged. The 89-line runner-session scope point no longer applies: this head drops that inherited delta. Coverage is now confirmed failing, while iOS Smoke remains in progress.

thymikee:
Exact-head Coverage is an owner-action failure, not infrastructure: changed-line coverage is 52/75 (69.33%) against the 70% gate. Uncovered changed lines are reported in src/cli/commands/daemon.ts, device-release.ts, device.ts, and src/daemon/device-claims.ts. Please cover the missing CLI and claim-race paths as part of the requested regression proof.

thymikee:
One P1 finding blocks readiness at `e16bb4acb060bd845d0ac72b4376cfbe13154ec8`:

`device release --stale` can clear the wrong worktree’s live owned-process record. The CLI constructs one `ownedProcesses` store from the caller’s state dir, while reconciliation intentionally reads the stale claim’s durable descriptor from `claim.stateDir`. Apple simulator-recording cleanup then calls `ownedProcesses.clear({ sessionId })` without a state-dir/owner fence. If the stale foreign claim and caller use the same session name, recovery can delete the caller’s live `owned-processes.json` and leave the stale owner’s marker behind, losing crash-recovery evidence.

Bind the process-record store to the stale claim’s state dir for each recovery transaction (or pass an exact owner-scoped store through the reconciler), and add an adversarial two-state-dir/same-session regression proving only the stale owner’s record is cleared.

The stale-only admission, claim lock + ownerToken recheck, and fail-closed outcomes otherwise look sound. Exact-head CI is green, but the manual proof covers claim deletion without a real durable-resource recovery, so it does not close this path.

No `ready-for-human` label until fixed on a new head.

thymikee:
All three follow-ups addressed; head is `1d2216bcbc` (rebased on #2160's P1 fix).

**(1) End-to-end CLI regressions through real durable-resource recovery** (`cli-device-release.test.ts`, commit `test: prove resources gate stale release…`): both run the real `device release --stale` CLI path — daemonless gateway composition included — against a dead owner whose recorded state dir holds an attributable durable resource record. An owner-mismatched record retains the claim with `app-log-owner-mismatch`; unreadable resource evidence retains it with `app-log-descriptor-invalid`. In both, the claim file and the resource evidence survive untouched, proving resources must reach a terminal state before the claim can be deleted. (The recover-then-release leg needs a live device binding, which the live run below covers.)

**Scan-to-lock race assertion** (`device-claims.test.ts`): the test holds the real per-device claim lock while a release is in flight (the scan has already classified the dead owner), replaces the claim with a successor token before releasing the lock, and proves the transaction reports `changed`, never calls the reconciler, and leaves the successor's claim intact. One test-harness note: the lock owner is written with a null start time because the suite pins a fake process start time for claim code while the lock's liveness probe reads the real one — recording the pinned value classifies the held lock as PID-reused and lets it be stolen.

**(2) Live real recovery path**, on a throwaway `iPhone 17 Pro` simulator with isolated claim/lease/state dirs:
1. Daemon A opened Preferences as `res-a` and started a durable screen recording (`record start`; `screen-recording.resource.json` present, lifecycle `open`).
2. `kill -9` on daemon A (abrupt death). Claim and open recording record left behind.
3. From a separate state dir: `device status --stale` named the dead owner and offered the release command; `device release --stale` reported `released`, the durable record transitioned `open → completed` (exact-owner recovery ran through the CLI-composed gateway), and the claim file was removed — cleanup before deletion, observed in order.
4. The second daemon then opened Preferences on the same simulator, answered `appstate` healthily, and closed cleanly.

**(3)** Confirmed resolved by the rewrite you noted — the branch carries no runner-session test delta; `runner-session.test.ts` is back at its pinned size. Coverage was topped up in the earlier push (refusal branches, text rendering, orphan warnings) and the race test now covers the changed-under-lock lines you listed.

Local gate: all checks green except two load-flaked files (`/bin/ps timed out after 1000ms` while a parallel session pushed host load to ~70); both pass 65/65 in isolation. Exact-head CI is running on the push.

thymikee:
P1 addressed in `46f812ceb7` (rebased on #2160's identity fix).

**Fix.** Recovery is now composed per claim, bound to the stale owner's recorded state dir: `runStaleDeviceClaimRelease` builds a fresh local-only gateway for each transaction, with the owned-process record store and session artifact paths resolved from `claim.stateDir` — never the caller's — and disposes it after the transaction. The Apple recording cleanup's `ownedProcesses.clear({ sessionId })` therefore always lands in the dead owner's store.

**Deterministic regression** (`cli-device-release.test.ts`, "recovery is composed per claim from the stale owner state dir"): two dead claims sharing one session name in two different state dirs — the exact adversarial shape — prove each recovery composition receives that claim's own state dir, and the reconciler is handed the matching claim.

**Live adversarial two-state-dir/same-session run**, on two throwaway simulators with a shared claim store:
1. Owner daemon A: opened Preferences as session `shared` on sim 1, started a durable screen recording (`owned-processes.json` + `screen-recording.resource.json` present), then `kill -9`.
2. Caller daemon B (separate state dir): opened Preferences as the **same** session name `shared` on sim 2 and started its own recording; its `owned-processes.json` was snapshotted.
3. `device release --stale` from B's environment: the stale claim was released, A's recording record settled `open → completed` through the per-claim scoped recovery, and B's **own live claim was refused** (`live-owner`) with the daemon-stop hint.
4. **B's `owned-processes.json` is byte-identical to the snapshot** (`cmp`), B's recording record stayed `open`, and B stopped the recording and closed cleanly afterwards. With one caller-scoped store, step 3 would have cleared B's record for `shared`.

Local gate: everything green except the two chronic timeout-flake files (`gesture-admission-parity`, `runner-client` abort test), which pass 67/67 in isolation — same host-contention signature as before. Exact-head CI is running on the push.

PR #2164, opened 2026-08-31
fix: keep named sessions visible in scoped inventory

thymikee:
## Summary

Keep explicitly named sessions visible in `agent-device session list` while preserving cwd and tenant isolation.

Session creation now persists explicit provenance (`cwd`, `tenant`, `named-local`, or `global-default`), and inventory filtering uses that stored fact instead of reconstructing ownership from the session address. This keeps a valid local name such as `tenant-a:qa` local, prevents tenant inventories from seeing it, and prevents a global unscoped `default` from leaking into cwd-scoped inventory.

This fixes the reported contradiction where `--session qa-cart-integrity` remained bound to an Android recording session while `session list --json` returned an empty list. Externally terminating the emulator still does not silently discard the recording session: `record stop` retains its recovery opportunity and `close` remains the explicit fallback.

Also clarify selector conflicts by saying the session is bound to the device, rather than saying the command itself is bound to the session.

## Validation

The regressions were observed red before the production change: local inventory omitted `tenant-a:qa` and leaked global `default`, while tenant A inventory also leaked the local colon-named session. Tests now cover both directions, cross-worktree filtering, and provenance stored through real tenant and cwd open routes.

At exact head `8394fd7c1f`, 39 focused routing, inventory, open, and session-runner tests pass. `pnpm check:affected --run` passed format, lint, typecheck, layering, fallow, and build; its related-test lane passed 1,423 tests and hit two 5-second provider-scenario timeouts. The recording scenario passed in isolation, and the remaining scripted iOS Settings timeout reproduced unchanged at pre-review head `c1ee1e2e33`.

Exact-head GitHub CI is green, including integration, coverage, repository guards, CodeQL, package checks, and Android/iOS/Linux/macOS smoke. The iOS workflow exercised its simulator smoke paths; its physical-device step was skipped by workflow policy.

No separate live device run was needed because the changed behavior is deterministic daemon inventory filtering and provenance; recording recovery mechanics are unchanged.

17 files changed, all within the session command family and mirrored tests; scope did not grow beyond that family.

### DIFF
--- a/src/daemon/__tests__/request-lock-policy.test.ts
+++ b/src/daemon/__tests__/request-lock-policy.test.ts
@@ -142,7 +142,7 @@ test('rejects existing-session selector conflicts under request lock policy', ()
         },
         ref(IOS_SESSION),
       ),
-    /--serial=emulator-5554/i,
+    /Session "qa-ios" is already bound to apple device "iPhone 16" \(SIM-001\), but snapshot selected --serial=emulator-5554/i,
   );
 });
 
--- a/src/daemon/__tests__/request-router-open.test.ts
+++ b/src/daemon/__tests__/request-router-open.test.ts
@@ -351,6 +351,10 @@ test('open stores admitted lease metadata on the session', async () => {
   );
 
   expect(response.ok).toBe(true);
+  expect(sessionStore.get('tenant-a:default')?.sessionScope).toEqual({
+    kind: 'tenant',
+    id: 'tenant-a',
+  });
   expect(sessionStore.get('tenant-a:default')?.lease).toEqual({
     leaseId: lease.leaseId,
     tenantId: 'tenant-a',
--- a/src/daemon/__tests__/request-router-session-address.test.ts
+++ b/src/daemon/__tests__/request-router-session-address.test.ts
@@ -198,7 +198,7 @@ test('lock-policy conflict on an implicit session names its store key, not "defa
   expect(response.ok).toBe(false);
   if (response.ok) return;
   expect(response.error.code).toBe('INVALID_ARGS');
-  expect(response.error.message).toContain(`session "${address}"`);
+  expect(response.error.message).toContain(`Session "${address}"`);
   expect(response.error.details?.session).toBe(address);
   expectAddressableRecovery(errorHint(response), address);
 });
@@ -221,4 +221,8 @@ test('session list reports the address an implicit session answers to alongside 
   expect(sessions?.[0]?.name).toBe('default');
   expect(sessions?.[0]?.address).toBe(address);
   expect(sessions?.[0]?.address).toMatch(SCOPED_ADDRESS_PATTERN);
+  expect(sessionStore.get(address)?.sessionScope).toEqual({
+    kind: 'cwd',
+    id: address.split(':')[1],
+  });
 });
--- a/src/daemon/__tests__/session-routing.test.ts
+++ b/src/daemon/__tests__/session-routing.test.ts
@@ -3,8 +3,8 @@ import assert from 'node:assert/strict';
 import fs from 'node:fs';
 import path from 'node:path';
 import { SessionStore } from '../session-store.ts';
-import { resolveEffectiveSessionName } from '../session-routing.ts';
-import type { SessionState } from '../types.ts';
+import { resolveEffectiveSessionName, resolveSessionScope } from '../session-routing.ts';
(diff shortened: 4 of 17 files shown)


thymikee:
Reviewed exact head `f40d3827049fa61d266a2c7c376239ba0301f8db`. P1: the filter widens visibility to every session with no `sessionScope`, not only explicitly named local sessions. Tenant isolation intentionally prefixes the stored session address but does not assign a cwd scope; a local cwd-scoped `session list` can therefore expose another tenant’s session address, state/log paths, and device identity. Keep tenant ownership at the SessionRef/address inventory seam and apply it before cwd visibility; absence of `sessionScope` is not proof of a globally visible local session. Add a regression proving a tenant-owned unscoped session stays hidden, alongside the desired result containing the caller’s current implicit session plus the local explicitly named session while excluding other cwd sessions. Exact-head CI is green; no device proof is needed for this deterministic inventory path, but the isolation finding blocks readiness.

thymikee:
Addressed in `c1ee1e2e33`. The finding was valid: `sessionScope === undefined` conflated globally named local sessions with tenant-owned sessions.

The inventory now resolves tenant ownership from the canonical `SessionRef.address` namespace first, then applies cwd visibility only to local sessions. A local request cannot enumerate tenant-owned unscoped sessions, while a tenant request can enumerate only its own namespace.

Regression coverage uses one mixed inventory: the caller's implicit cwd session and a local explicitly named session remain visible; another cwd and a lease-less tenant session stay hidden. A second test proves tenant A cannot see tenant B or global local sessions. The mixed regression was observed failing before the production fix because `tenant-a:remote-recording` leaked into the response.

Validation: `pnpm check:affected --run` (200 related test files, 1,198 tests; format, lint, typecheck, layering, fallow, build, and daemon wire compatibility all passed). No device proof is needed for this deterministic inventory path.

thymikee:
Two actionable findings still block readiness at `c1ee1e2e33b30f6ff5db7bb4e6fd69b25d5372e0`:

1. Tenant ownership is reconstructed from any unscoped address prefix before `:`. Local explicit session names permit `:` (they are sanitized for storage, not rejected), so a normal local `--session tenant-a:qa` is misclassified as tenant-owned: tenant A can see its device/artifact paths while local inventory hides it. Persist/use explicit tenant provenance instead of parsing an overloaded user-controlled address, with regressions for both exposure directions.

2. `sessionMatchesScope()` still treats every unscoped session as matching every cwd scope. That makes a global unscoped `default` session (created without `meta.cwd`) visible to unrelated worktrees while trying to expose explicitly named local sessions. The ownership model must distinguish global unscoped default from an explicitly named local session; add the cross-worktree default regression.

The new tenant-prefix test fixes the prior direct case but does not cover either ambiguous ownership shape. No `ready-for-human` label until both boundaries are explicit and tested on a new head.

thymikee:
Addressed both findings at `8394fd7c1f` by replacing inferred ownership with stored session provenance.

- Every production session creation path now records one explicit scope: `cwd`, `tenant`, `named-local`, or `global-default`. Inventory filtering uses only that field and fails closed for unstamped state; it no longer parses tenant ownership from the address.
- A local explicit session named `tenant-a:qa` remains visible to local inventory and is excluded from tenant A inventory. A real tenant A session remains visible only to tenant A.
- CWD inventory includes same-CWD implicit sessions and explicitly named local sessions, but excludes both other-CWD sessions and an unscoped global `default`.
- Router-level tests pin provenance on real tenant and CWD open paths, while record, replay-open, and session-creating snapshot paths use the same resolver.

Planted-red evidence: before the production change, the local inventory regression omitted local `tenant-a:qa` and leaked global `default`; tenant A inventory also leaked local `tenant-a:qa`.

Exact-head validation: 39 focused tests pass across the six routing/inventory/open suites. `pnpm check:affected --run` passed format, lint, typecheck, layering, fallow, and build; its related-test lane passed 1,423 tests and hit two unrelated 5s provider-scenario timeouts. The record scenario passed in isolation, and the remaining iOS Settings timeout reproduces unchanged at pre-review head `c1ee1e2e33`, so exact-head CI remains the authority for that lane. I have not added `ready-for-human` pending CI.

thymikee:
Follow-up: all checks on `8394fd7c1f` are now green, including Integration Tests, Coverage, Repo Guards, and Android/iOS/Linux/macOS smoke. I updated the PR validation and added `ready-for-human`.

PR #2165, opened 2026-08-31
docs+ux: make device ownership discoverable end to end

### LINKED ISSUE #1320: feat: add cross-worktree device ownership and safe recovery

thymikee:
## Summary

Stacked on #2162 (which stacks on #2160). The last slice of #1320: make the `busy? → inspect → choose another device or release` loop discoverable from every surface an agent actually reads, and retire the Stage-5 PID-hunting guidance.

- **`devices` now projects ownership**: each row that a foreign `open` would be refused on carries `claimedBy: { session, workspace }` (text output: `claimed by session "x" in /worktrees/y`). Provably dead owners are deliberately not projected — the next open reconciles and replaces them automatically, so showing them would only scare agents off free devices. This is the `observe`-policy projection the #1320 design explicitly permitted, and it completes the third recovery path ("choose another device") that previously required guesswork.
- **`help debugging`** gains a "Device busy and ownership" section separating the two `DEVICE_IN_USE` flavors (same-daemon session vs cross-worktree claim) with their exact recoveries, including "never recover by hunting PIDs".
- **AGENTS.md** documents both flavors (previously described only the same-daemon one, whose `close --session` advice is wrong for a foreign claim).
- **docs/agents/device-verification.md** retires the last `ps`/`kill` recovery guidance in favor of `device status --stale`, `daemon stop --state-dir`, and `device release --stale` — the Stage 5 deliverable of #1320.
- **ADR-0010** no longer calls `DEVICE_IN_USE` "the only retriable code" without naming the claim path's `retriable: false` override.
- **help-conformance**: the rendered cross-worktree claim error sample (`DEVICE_CLAIM_IN_USE_SAMPLE`) was validated but bound to no quiz case; it now has one (`sample-output-device-claim-inspects-owner`) checking the model runs the exact `device status` recovery and forbidding retry-open/foreign-close/PID-hunting.
- **README** points at `device status` / `device release --stale`.

## Validation

All non-vitest gate checks green (format, lint, typecheck, layering, di-seams, fallow, mcp-metadata, build, package, node-integration, macos-coverage). The vitest step flaked on 9 load-timeout failures while a parallel agent session ran a full suite on this host (load avg 35–42); all 9 files pass when re-run at low concurrency, and the failures were 5s-timeouts in files this PR does not touch.

Live real-binary validation: planted a live foreign claim for the booted `iPhone 17 Pro` in an isolated claim store — `devices` rendered `claimed by session "other-worktree-session" in /worktrees/other` and `--json` carried `claimedBy`; a dead-owner claim was correctly not projected.

Closes #1320.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -113,8 +113,10 @@ cross-language rules change through golden tables under `contracts/fixtures/`.
   regenerate baselines to accept unrelated findings.
 - The first Node process after a newly signed Apple runner launches may block during Gatekeeper
   verification. Warm it with a throwaway `node -e 0` before measuring.
-- `DEVICE_IN_USE` normally means another session still owns the device. Follow the error's targeted
-  `close --session` hint instead of debugging the daemon.
+- `DEVICE_IN_USE` has two flavors. "already in use by session X" is this daemon — follow its
+  `close --session` hint. "owned by session X in workspace Y" is another worktree's device
+  claim — non-retriable; run the error's `device status`/`device release --stale` recovery,
+  never PID hunting.
 - A changing timeout failure set that passes in isolation is host contention. Reproduce the same
   test on `origin/main` under the same load before treating it as a regression.
 
--- a/README.md
+++ b/README.md
@@ -145,7 +145,7 @@ The same session and evidence model works at every step: the agent explores the 
 
 Support depth varies by target. Newer backends such as HarmonyOS and Vega OS cover a subset of commands; run `agent-device capabilities --platform <platform>` to see what a target supports.
 
-Sessions are scoped to the caller's git worktree, and host-local device claims stop parallel agents from taking over each other's simulators and emulators. The same commands drive hosted devices on [BrowserStack, AWS Device Farm, and Limrun](https://oss.callstack.com/agent-device/docs/device-clouds).
+Sessions are scoped to the caller's git worktree, and host-local device claims stop parallel agents from taking over each other's simulators and emulators. Inspect ownership without a daemon via `agent-device device status`, and settle provably dead owners with `agent-device device release --stale`. The same commands drive hosted devices on [BrowserStack, AWS Device Farm, and Limrun](https://oss.callstack.com/agent-device/docs/device-clouds).
 
 `agent-device` uses the inspect-act-verify process from Vercel's [agent-browser](https://github.com/vercel-labs/agent-browser) for mobile, TV, and desktop apps. Basic `--platform web` support runs `agent-browser` in the same session and replay system.
 
--- a/docs/adr/0010-error-system.md
+++ b/docs/adr/0010-error-system.md
@@ -28,7 +28,8 @@ agent failures (selector/ref misses), and consumers that drop fields (MCP tool e
    `err instanceof AppError ? err : new AppError(...)`.
 2. **Code selection.** Use the most specific `KnownAppErrorCode`; `COMMAND_FAILED` is for genuine
    runtime failures of a well-formed request, never a catch-all for capability gaps
-   (`UNSUPPORTED_OPERATION`), contention (`DEVICE_IN_USE`, the only retriable code), or ambiguity
+   (`UNSUPPORTED_OPERATION`), contention (`DEVICE_IN_USE`, the only code retriable by default;
+   the cross-worktree device-claim path overrides it to `retriable: false`), or ambiguity
    (`AMBIGUOUS_MATCH`). New codes are added to the union deliberately; machine-dispatchable
    sub-classification rides in `details.reason` (the lease registry is the model).
 3. **Hints answer "what should the agent run next".** A hint is required wherever the per-code
--- a/docs/agents/device-verification.md
+++ b/docs/agents/device-verification.md
@@ -38,11 +38,13 @@ failed verification attempts.
 - Use a purpose-specific session name for experiments, and an isolated `--state-dir` under
   `/private/tmp` when you need cleanup isolation beyond the current worktree's default daemon.
 - Track opened sessions in working notes; close each one before the final response.
-- If `close` is blocked by stale daemon metadata, inspect processes first with
-  `ps -ax | rg "agent-device|xcodebuild test-without-building"`. Stop only exact stale PIDs belonging
-  to this verification run, then `pnpm clean:daemon`.
-- If cleanup cannot be completed, report the remaining session name, state dir, PIDs, and metadata
-  paths as a blocker.
+- If `close` is blocked or ownership looks stuck, inspect it with
+  `agent-device device status --stale` (daemonless), stop the owning daemon with
+  `agent-device daemon stop --state-dir <dir>` (add `--clean` to remove retained runners), and
+  release provably dead owners with `agent-device device release --stale`. Do not hunt PIDs with
+  `ps`/`kill`.
+- If cleanup cannot be completed, report the remaining session name, state dir, and the
+  `device status --stale` output as a blocker.
 
 ## Sandboxed environments
 
--- a/packages/contracts/src/client-device-view.ts
+++ b/packages/contracts/src/client-device-view.ts
@@ -21,6 +21,15 @@ export type AgentDeviceDevice = {
    */
   appleOs?: AppleOS;
   identifiers: AgentDeviceIdentifiers;
+  /**
+   * Present when a host-local device claim currently blocks foreign use of
+   * this device (#1320). Provably dead owners are not projected — the next
+   * open reconciles and replaces them automatically.
+   */
+  claimedBy?: {
+    session: string;
+    workspace: string;
+  };
   ios?: {
     udid: string;
   };
--- a/scripts/help-conformance-cases.mjs
+++ b/scripts/help-conformance-cases.mjs
@@ -2,6 +2,7 @@ import {
   AMBIGUOUS_MATCH_SAMPLE,
   APP_NOT_INSTALLED_SAMPLE,
   BROWSERSTACK_CONNECT_SAMPLE,
+  DEVICE_CLAIM_IN_USE_SAMPLE,
   DEVICE_IN_USE_SAMPLE,
   FOREGROUND_SNAPSHOT_FAILURE_SAMPLE,
   MERGED_CARD_ACTIONS_SAMPLE,
@@ -573,6 +574,28 @@ Use the output already shown to determine whether the feed-search UI is present,
       { id: 'noClose', pattern: /(?:^|\n)agent-device\s+close\b/i },
       { id: 'noReopen', pattern: /(?:^|\n)agent-device\s+open\b/i },
       { id: 'noRawCoordinateTarget', pattern: RAW_COORDINATE_TARGET },
+    ],
+  },
+  {
+    id: 'sample-output-device-claim-inspects-owner',
+    docs: ['--help:first30', 'debugging'],
+    recovery: { code: 'DEVICE_IN_USE', sample: DEVICE_CLAIM_IN_USE_SAMPLE },
+    task: quiz(
+      DEVICE_CLAIM_IN_USE_SAMPLE,
+      'A different worktree owns this device and you must not interrupt it. What command should run next?',
+    ),
+    expectations: ['validPlanCommands', 'fullPrefix'],
+    matchers: [
+      {
+        id: 'runsDeviceStatus',
+        pattern:
+          /(?:^|\n)agent-device\s+device\s+status\s+--platform\s+android\s+--serial\s+emulator-5554\b/i,
+      },
+    ],
+    forbidden: [
+      { id: 'noRetryOpen', pattern: /(?:^|\n)agent-device\s+open\b/i },
+      { id: 'noForeignClose', pattern: /(?:^|\n)agent-device\s+close\b/i },
+      { id: 'noPidHunting', pattern: /(?:^|\n)\s*(?:ps|kill|pkill)\b/i },
     ],
   },
   {
--- a/src/cli-schema/cli-help.ts
+++ b/src/cli-schema/cli-help.ts
@@ -512,6 +512,11 @@ Example:
   agent-device react-devtools profile report @c5
   agent-device network dump --include headers
 
+Device busy and ownership:
+  DEVICE_IN_USE has two flavors. "already in use by session X" is this daemon: reuse it with --session X, or run close --session X first. "owned by session X in workspace Y" is another worktree's daemon holding the host-global device claim: it is never retriable — run the error's exact recovery command instead of retrying.
+  Inspect ownership without any daemon: agent-device device status (add --stale for proven-dead owners; settle and release those with agent-device device release --stale). devices marks rows that are claimed, so pick an unclaimed device instead of contending.
+  A live foreign owner is released only by closing its session from its own workspace or stopping its daemon: agent-device daemon stop --state-dir <owner state dir> (the error names the state dir). Never recover by hunting PIDs with ps/kill. boot/install/shutdown take the same claims as open and refuse foreign-claimed devices identically.
+
 Use snapshot, screenshot, logs, network, perf frames, and perf memory for device/app runtime evidence. Use react-devtools when component internals or React rendering behavior matters.`,
   },
   cdp: {
--- a/src/client/client-normalizers.test.ts
+++ b/src/client/client-normalizers.test.ts
@@ -1,6 +1,6 @@
 import assert from 'node:assert/strict';
-import { test } from 'vitest';
-import { normalizeOpenForegroundComposition } from './client-normalizers.ts';
+import { expect, test } from 'vitest';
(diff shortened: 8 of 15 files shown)


thymikee:
Two actionable findings block readiness at `598925f5f3a0102a830c87d92dd45dc85a255424`:

1. `blockingClaimOwnersByDevice()` keys claims and inventory rows by bare `device.id`, but claim ownership is keyed by canonical platform family + Apple OS + id. Distinct platform devices may share an id, so a live Android claim can project `claimedBy` onto an unrelated Apple/Harmony/Vega row, with scan order deciding the displayed owner. Use `canonicalLocalDeviceKey()` for both map insertion and lookup, and add a cross-platform same-id regression.

2. The regression stops at the daemon response. It stays green if client normalization, JSON serialization, or human `devices` rendering drops `claimedBy`, despite the PR promising end-to-end discoverability. Add Node/client and CLI text/JSON assertions for the projected owner (MCP shares serialization).

The stack is based correctly on #2162 → #2160. Please fix and validate these paths on a new head; no `ready-for-human` label yet.

thymikee:
Both findings addressed in `1014200994` (rebased on the #2162/#2160 review fixes).

**(1) Canonical keying.** `blockingClaimOwnersByDevice` now keys both sides by the canonical local device key: claims by their own `claim.deviceKey`, inventory rows by `canonicalLocalDeviceKey(deviceClaimIdentity(device))` (the claim-identity derivation is exported from `device-claims.ts` so there is exactly one derivation site). The cross-family same-id regression — a live Android claim for `shared-id` next to an Apple simulator row with the same bare id — proves the Android row gets `claimedBy` and the Apple row does not. **Observed red against the bare-id keying** (the Android claim projected onto the Apple row), green with the fix.

**(2) End-to-end assertions.** The projection is now pinned at every hop the PR promises:
- daemon response (existing handler tests, plus the cross-family regression above);
- client normalization (`client-normalizers.test.ts`): well-formed `claimedBy` is preserved, malformed shapes are dropped;
- CLI serialization and human rendering (`output.test.ts`): `devices` JSON data carries `claimedBy` and the text line renders `claimed by session "qa" in /worktrees/qa`, with the unclaimed row clean — MCP shares this serialization.

`pnpm check:affected --run` fully green on this head (no flakes this round). Exact-head CI is running on the push.

PR #2167, opened 2026-08-31
fix: improve ref-based interaction recovery

thymikee:
## Summary

- Treat every positional after `fill @ref` as replacement text, so `fill @e57 good morning` enters `good morning` instead of interpreting `good` as a fallback label.
- When a model drops a generation suffix from an emitted scoped ref, return the exact safe ref to retry when the current frame proves it exists. This remains fail-closed and never rewrites refs silently.
- Make iOS `TEXT_INPUT_COMMIT_NOT_OBSERVED` recovery inspect the field first, then retry a quoted paced `fill` only when the value is actually missing.
- Update generated help and command documentation for the corrected fill grammar.

## Validation

- `pnpm check:affected --base origin/main --head HEAD --run`: all runnable checks passed, including 311 related test files / 2,344 tests and provider integration.
- iOS XCTest runner build passed locally.
- Android Element 08 live benchmark: a dropped `@e24` suffix returned `@e24~s467960`; the model retried it directly, `fill` reported 12 characters, and the settled snapshot showed `good morning` before send.
- iOS Element 08 independently exercised the exact-ref recovery (`@e40` suggested `@e40~s362848`). That scenario was ultimately blocked by an unrelated Element “Threads no longer experimental” modal, so it is not claimed as iOS fill success evidence.
- The quoted iOS fill investigation showed `TEXT_INPUT_COMMIT_NOT_OBSERVED` followed immediately by a snapshot containing the complete expected value, motivating inspect-before-retry guidance without weakening verification.

The PR touches 10 files in the interaction command family (source, tests, Apple recovery text, help, and docs). Scope did not grow beyond ref-based interaction and text-entry recovery. Benchmark harness cleanup and benchmark artifacts are intentionally excluded.


thymikee:
Reviewed exact head `0a77faff103f089c7a7e4bb23c04fcc2c2923fe3` against the shipped interaction route and ref-frame policy. No actionable code findings.

- `fill @ref` now treats every remaining positional as replacement text across CLI decoding and daemon parsing; the regressions cover the prior first-word/fallback-label split.
- Partial-frame recovery remains fail-closed: it suggests an exact pinned ref only when that ref was emitted in the current scope, and requires an explicit retry rather than silently rewriting.
- Exact-head checks are green, including Android/iOS/macOS/Linux smoke, Coverage, Integration Tests, Repo Guards, Size, and command-docs.

The Android live run exercises suffix-drop → exact-ref suggestion → successful fill. The iOS evidence covers exact-ref recovery; its separate fill attempt was blocked by an unrelated modal and is not overstated.

Clean for human review; merge remains a maintainer decision.

Issue #2170, opened 2026-08-31
iOS sim: ungraceful recorder kill leaves CoreSimulator host recording lock dangling (code 16 for all later recordings)


thymikee:
Observed repeatedly during #2168 live validation on Xcode's current CoreSimulator: this host allows ONE active `simctl recordVideo` session at a time, and when a recorder process dies without a graceful detach (SIGKILL to the recorder, e.g. via a generic owned-process kill during recovery, or a stray `pkill -9`), `SimStreamProcessorService` keeps the session registered. Every later `simctl io <any-sim> recordVideo` on the host then fails:

```text
Error starting video recorder: Error Domain=NSPOSIXErrorDomain Code=16 "Resource busy"
NSLocalizedFailureReason=Host recording is already in progress
```

agent-device surfaces this as `record start` → `UNKNOWN` / "simctl recordVideo exited with code 16", with no hint. The lock persists until `SimStreamProcessorService` is killed (it relaunches on demand) or the CoreSimulator service restarts.

Two consequences for recovery quality:

1. Screen-recording recovery that kills an orphaned recorder ungracefully can fix one device while leaving the whole HOST unable to record — worth preferring SIGINT (which `simctl recordVideo` handles as finalize-and-detach) with a grace window before escalating.
2. The code-16 failure deserves classification and a hint ("another recording is active on this host, or a previous recorder died ungracefully; stop it or restart CoreSimulator's stream service") instead of `UNKNOWN`.

Repro: start a recording, SIGKILL the recorder process, try `simctl io <other-booted-sim> recordVideo` → code 16; `kill -9 <SimStreamProcessorService pid>` clears it.

PR #2171, opened 2026-08-31
fix(daemon): compose stale-claim recovery from the dead owner's state dir

### LINKED ISSUE #2168: Daemon-side stale-claim reconciliation uses the daemon's own owned-process store for foreign state dirs
Same hazard class as the #2162 review finding, but on the pre-existing daemon paths (#1735), so it is not covered by that PR's fix.

`open`-time orphan reconciliation and the daemon-startup sweep (`reconcileOrphanedDeviceClaims`) run `createDeviceClaimReconciler` with the daemon's ONE composed gateway. The reconciler correctly reads the stale claim's durable records from `claim.stateDir`, but the gateway's owned-process record store (and session artifact resolution) is bound to the reconciling daemon's own state dir. Apple recording cleanup clears owned-process records by bare `sessionId` through that store, so settling a foreign-state-dir dead claim whose session name matches one of this daemon's live sessions (`default` is common across worktrees) can clear the live session's `owned-processes.json` and leave the stale owner's marker behind.

#2162 fixed this for the daemonless CLI path by composing recovery per claim, bound to `claim.stateDir` (`composeStaleClaimRecovery` in `src/cli/commands/device-release.ts`). The daemon paths should get the same per-claim scoping — either reuse that composition shape for foreign-stateDir claims, or thread an owner-scoped store through the reconciler so the clear cannot address another state dir's records.

Acceptance: an adversarial regression with a live daemon session named identically to a foreign dead claim's session proves reconciliation settles the foreign records and leaves the live session's owned-process record untouched, on both the acquire-time and startup-sweep paths.

Part of the #1320 hardening trail; found while addressing https://github.com/callstack/agent-device/pull/2162#issuecomment-5478189050.

thymikee:
## Summary

Closes #2168 — the daemon-side sibling of the #2162 review finding, live-reproduced on `main` before fixing.

The daemon's acquire-time reconciliation (open and transient admission) and the startup sweep ran stale-claim recovery through the daemon's single composed gateway, so settling a foreign state dir's dead claim cleared owned-process records **by bare session id in the reconciling daemon's own store**. A live session sharing the dead claim's session name (`default` is common across worktrees) lost its `owned-processes.json` to a foreign claim's cleanup, while the dead owner's marker survived — crash-recovery evidence destroyed in one place and leaked in the other.

## Fix

Every production reconciliation path — open-time admission, transient-exclusive admission, the startup sweep, and the daemonless CLI release — now goes through one entry, `createOwnerScopedDeviceClaimReconciler`: recovery is composed **per claim, bound to that claim's recorded state dir** (owned-process record store and session artifact paths included) and disposed after the transaction. `device-release.ts` becomes a thin caller of the same module, so the CLI and daemon can no longer diverge. Gateway state is instance-scoped (each composition owns its own app-log runtime handles and store), so disposal shuts down only the transaction's handles — verified against the runtime module implementations before choosing this shape over threading a store through the reconciler.

## Validation

**Live RED on `main`'s wiring** (two throwaway simulators, isolated claim/lease/state dirs): owner daemon A opened sim 1 as session `shared`, started a durable screen recording, and was `kill -9`'d; caller daemon C already ran with a planted live `owned-processes.json` for its own `shared` session. C's `open` of sim 1 triggered acquire-time reconciliation: **C's record was DELETED and the dead owner's marker survived** — the finding, observed.

**Live GREEN on this head, identical order**: caller record **byte-identical** (`cmp`), the stale owner's recording settled `open → completed`, and the owner's own `owned-processes.json` was cleared — the clear landed in the right store.

**Unit**: `device-claim-owner-recovery.test.ts` pins one composition per claim from that claim's state dir with disposal on success, retain, and throw; the CLI's per-claim regression from #2162 continues to pass through the shared module unchanged.

Local gate: all checks green except the chronic `ios-lifecycle` 5s-timeout flake (host 5-min load avg was 45 during the run; passes 3/3 in isolation).

Found in passing, filed separately: #2170 (an ungracefully killed recorder leaves CoreSimulator's host-level recording lock dangling — every later `record start` on the host fails with code 16).


thymikee:
Sweep-callsite regression added in `2445e19d6b`.

**The test** (`daemon-entrypoint.test.ts`, "startup sweep settles a foreign dead owner without touching a live same-named session") boots the real `startDaemonRuntime` — no injected reconciler, the production sweep wiring itself — with:
- a foreign-state-dir dead claim whose recorded simctl recorder **still runs as an orphan** (a spawned sleeper carrying its real observed command and start time, so recovery classifies it `owned-alive` and must take the descriptor-cleanup path — the one that clears the owned-process store);
- a durable `screen-recording` record (`lifecycle: open`) in the dead owner's state dir plus its owned-process record;
- a **live same-named session record in the daemon's own state dir**, with a purpose outside the startup reaper's selection so only a wrong-store recovery clear can touch it.

It proves, through the real sweep: the orphan recorder is terminated, the foreign recording settles `open → completed`, the owner's process record is cleared **in the owner's state dir**, and the daemon's own record stays **byte-identical**.

**Observed red against a reverted sweep callsite** (caller-scoped `createDeviceClaimReconciler({gateway: deviceRuntimeGateway, …})`): the owner's record survives in the wrong place — and the byte-identity assertion independently catches the caller-store damage. Green with the fix; the full entrypoint file passes 6/6.

Two mechanics found while making it deterministic, for the record: reattach short-circuits to `already-missing` without any store clear when every recorded process is dead (which is why a dead-pid fixture cannot discriminate), and the startup reaper would legitimately remove a same-named planted record unless its purpose is outside `['simctl-screen-recording']`.

`pnpm check:affected --run` fully green on this head, no flakes. Watching exact-head iOS Smoke now and will confirm here when it completes.

thymikee:
Reviewed exact head `029fe05ad384ef8cd993209a5f7a3ad22827ea7a`. The owner-scoped production wiring is correct and exact-head CI is green, but #2168's required regression is not satisfied. `device-claim-owner-recovery.test.ts` only injects a fake composer into the helper; reverting the open/transient/startup callsites to the old caller-scoped reconciler would leave it green. The author-reported live evidence covers acquire-time only and does not provide startup-sweep proof. Add planted-red adversarial regressions through the actual daemon acquire/open path and startup sweep (ideally transient admission too) with two state dirs sharing a session name, proving foreign records settle while the live caller's `owned-processes.json` remains byte-identical. No `ready-for-human` label until those shipped paths are pinned.

thymikee:
Re-reviewed new head `2445e19d6b1fdf8ed2951a7efa374a6911d9bb49`. The added `startDaemonRuntime` regression validly pins the startup sweep: it exercises the production callsite and proves foreign cleanup while the same-named live record stays byte-identical. #2168 is still only half pinned, however: there is no automated regression through the real daemon acquire/open route, so reverting the request-router/request-execution callsites remains green; the PR body’s manual acquire evidence does not satisfy the issue’s explicit regression requirement for both acquire-time and startup sweep. Add the same adversarial shape through real `open` acquisition (and preferably transient admission). Also make the new startup test cleanup unconditional: `runtime` is block-local and shutdown occurs after assertions, so the planted-red/reverted path throws before shutdown and leaks the in-process daemon; keep an outer runtime handle and close it in `finally`, and delete `AGENT_DEVICE_CLAIMS_DIR` when its prior value was undefined. No readiness label yet.

thymikee:
Exact-head CI on `2445e19d6b` is fully green, iOS Smoke included. Both g2g conditions are met: the startup-sweep regression is in (observed red against a reverted callsite) and exact-head smoke has completed.

thymikee:
You're right — my "both conditions met" comment overclaimed: the acquire/open path had only the live proof, and the sweep test could leak its runtime. Both addressed in `1c7906bc04`.

**Real acquire/open regression** (`request-router-open.test.ts`, "open reconciles a foreign dead owner through that owner state dir, never the daemon store"): a real `open` through the production request router (`createRequestHandler`) against a device whose foreign dead claim carries an orphaned recorder (spawned sleeper with its real observed identity, forcing the descriptor-cleanup path) and durable recording evidence, next to a live same-named session record under the daemon's own state dir. It proves open-time reconciliation terminates the orphan, settles the foreign recording `open → completed`, clears the owner's process record **in the owner's state dir**, leaves the daemon's own record **byte-identical**, and hands the claim to the opening session. **Observed red against a reverted router callsite** (caller-scoped reconciler), green with the fix, full file 14/14.

**Daemon leak fixed**: the startup-sweep regression now holds the runtime in an outer binding and shuts it down in `finally`, so an assertion failure can no longer leak it.

Local gate: every check green except three 5s-timeout files during a parallel test wave on this host (load avg ~59); the one that also failed in isolation (`interaction-response-shape.contract`) fails identically on `origin/main` under the same load — the documented contention signature, not a regression. Exact-head CI is running on the push; I'll report the iOS Smoke result here when it lands rather than calling it done before then.

thymikee:
Env-restore fixed in `44fe0522eb`: all three sites (both new regressions plus the pre-existing prune-log test they mirrored, which had the same defect) now `delete process.env.AGENT_DEVICE_CLAIMS_DIR` when the previous value was absent, instead of assigning the literal `"undefined"`. Both files pass 20/20. Watching exact-head CI and will report the iOS Smoke result here when it lands.

thymikee:
Review complete on `44fe0522eb96afdcbf21b30dc7a3da93bca3f4a6`: code review is clean. The open/acquire test now exercises the real router and proves owner-scoped stale cleanup without touching the live same-named claim; startup recovery, failure-safe runtime shutdown, and environment restoration are also covered. Exact-head CI is still running (Android, iOS, and Coverage at review time), so merge-readiness remains conditional on those checks finishing green. No further code changes requested.

thymikee:
Exact-head CI on `44fe0522eb` is fully green: 18 checks passing, 0 pending, and all four Smoke Tests lanes (iOS included) report success — verified against the commit's check runs directly, not a stale head. That closes both remaining items: the env-restore cleanup is in this head, and exact-head iOS Smoke has completed green.

PR #2172, opened 2026-08-31
fix(apple): switch to manual code signing when a provisioning profile is set

### LINKED ISSUE #2153: Physical device runs: CODE_SIGN_STYLE stays Automatic when AGENT_DEVICE_IOS_PROVISIONING_PROFILE is set, so xcodebuild always fails
**Version:** agent-device 0.20.10, macOS, physical iPhone runs.

## What happens

`agent-device help physical-device` says: *"If Xcode cannot choose a profile, set AGENT_DEVICE_IOS_PROVISIONING_PROFILE to the profile name/specifier."*

Following that advice makes the runner build fail every time. In `dist/src/physical-device-control.js` the signing overrides are:

```js
let s=[`CODE_SIGN_STYLE=Automatic`];
if(i) s.push(`DEVELOPMENT_TEAM=${i}`);
if(a) s.push(`CODE_SIGN_IDENTITY=${a}`);
if(o) s.push(`PROVISIONING_PROFILE_SPECIFIER=${o}`);
```

`CODE_SIGN_STYLE` is always `Automatic`. Setting `AGENT_DEVICE_IOS_PROVISIONING_PROFILE` adds `PROVISIONING_PROFILE_SPECIFIER` next to it, and xcodebuild rejects that combination with "has conflicting provisioning settings". The help text and the code contradict each other: the documented escape hatch cannot work.

## Expected

When a profile specifier is provided, switch the style to manual:

```js
let s=[`CODE_SIGN_STYLE=${o ? "Manual" : "Automatic"}`];
```

cargo-mobile2, Expo, Tauri and Godot all flip to manual signing when a specifier is present, each after hitting this same xcodebuild error.

## Related limitation

On a Mac without an Apple ID session in the Xcode GUI, automatic signing fails with "No Accounts", and there is no way to reach cloud signing through the tool: no env var for extra xcodebuild arguments, and no support for `-authenticationKeyPath`. Support for passing App Store Connect API key flags would make physical-device runs work on headless CI Macs. Happy to file that as a separate feature request if you prefer.

Our current workaround is a scoped xcodebuild shim that appends the auth flags to the runner build. It works, but the one-line style fix above would make it unnecessary for the common case.

thymikee:
## Summary
- `resolveRunnerSigningBuildSettings` hardcoded `CODE_SIGN_STYLE=Automatic` for physical-device builds, even when `AGENT_DEVICE_IOS_PROVISIONING_PROFILE` was set. xcodebuild then rejects `CODE_SIGN_STYLE=Automatic` combined with `PROVISIONING_PROFILE_SPECIFIER` as "conflicting provisioning settings" — the documented escape hatch (`agent-device help physical-device`) could never work.
- Now switches to `CODE_SIGN_STYLE=Manual` whenever a profile specifier is provided, matching the behavior of cargo-mobile2, Expo, Tauri, and Godot noted in the issue.

## Test plan
- [x] `resolveRunnerSigningBuildSettings` unit tests updated to expect `Manual` when a profile is set, plus a new case covering profile-only (no team/identity) input
- [x] `npx vitest run --project apple-runner packages/platform-apple/src/runner/__tests__/runner-client.test.ts -t signing` passes
- [x] Live physical-device run against a connected iPhone, isolated `AGENT_DEVICE_STATE_DIR`, with `AGENT_DEVICE_IOS_TEAM_ID`/`AGENT_DEVICE_IOS_PROVISIONING_PROFILE` set to a real (mismatched-on-purpose) profile:
  - **Before fix:** xcodebuild fails with the exact bug from the issue — `AgentDeviceRunnerUITests has conflicting provisioning settings. AgentDeviceRunnerUITests is automatically signed, but provisioning profile ... has been manually specified.`
  - **After fix:** the invocation carries `CODE_SIGN_STYLE=Manual`; that specific conflict is gone, and the build instead fails for the profile's own (expected, unrelated) problems — wrong app ID, expired, no matching signing certificate — confirming xcodebuild accepts the manual-signing build settings and proceeds past provisioning-style resolution.

Fixes #2153


thymikee:
Reviewed exact head `4d5fc399e4a210741520fd1e18c4358e7a7a0a10`. The production change is correct: a configured profile selects manual signing and reaches both cache identity and xcodebuild while simulator/macOS paths remain unchanged. Readiness is blocked by owner-action CI and missing physical-device proof. Format `runner-client.test.ts`, then move the signing/cache-metadata cases into a sibling test rather than raising the ratchet: Coverage rejects growth of the already 1,577-line file to 1,585. Keep the profile-only regression because it prevents an incorrect `profile && team/identity` condition. Finally, provide a real signed-iPhone build using `AGENT_DEVICE_IOS_PROVISIONING_PROFILE`; this device-facing fix cannot be merge-ready from unit evidence alone. No `ready-for-human` label while Lint/Format and Coverage are failing.

thymikee:
Re-reviewed unchanged head `4d5fc399e4a210741520fd1e18c4358e7a7a0a10` after the validation update. The connected-iPhone A/B is useful issue-specific evidence: it reproduces the Automatic+profile conflict before the fix and proves Manual signing advances xcodebuild past that conflict afterward. It still uses an intentionally mismatched/expired/no-certificate profile, so a successful valid-profile signed runner build remains residual device risk. The immediate blockers are unchanged and owner-action: Lint/Format fails, and Coverage rejects growing the 1,577-line `runner-client.test.ts`; format it and extract the signing/cache-metadata tests into a sibling file rather than raising the ratchet. All smoke lanes are now green, but no readiness label while those exact-head CI failures remain.

PR #2200, opened 2026-08-31
perf: retain iOS runner across physical relaunch

thymikee:
## Summary

Keep a healthy XCTest runner alive while `open --relaunch` replaces an app process on a physical iOS device, then reset the runner's process-bound target state before the next command. A failed relaunch or unconfirmed target reset still discards the runner, preserving the original typed failure and existing cancellation/deadline behavior. Physical iPadOS, tvOS, and visionOS retain their existing restart behavior until separately evidenced.

This removes repeated runner startup from the physical-device relaunch critical path; it does not prewarm accessibility or weaken snapshot correctness. The initial runner startup is still paid, and the retained runner remains owned until normal session close/idle cleanup.

Matched CoreDevice iPhone samples on production-equivalent pre-rebase commit `c3bede63dc` showed:

| Relaunch metric | Before | After |
| --- | ---: | ---: |
| runner prewarm wait median | 6,583 ms | 13 ms |
| runner command send median | 6,441.5 ms | 8.5 ms |
| runner retries median | 5 | 0 |
| successful command total median | 11,755 ms | 9,803 ms (-16.6%) |
| successful CLI wall median | 13,866 ms | 13,043 ms (-5.9%) |
| successful CLI wall p95 | 39,362 ms | 36,916 ms (-6.2%) |
| observed open failures | 2/8 | 0/8 |

The mechanism evidence is stronger than the aggregate wall-time delta: post-change relaunches emitted no runner-startup timing and had zero runner retries, while CoreDevice terminate/process discovery remained the dominant variable. Snapshot acquisition stayed around 0.2 seconds, so the 65-91 second tails were not attributed to AX warmup. Semantic dev-client readiness was separately Metro-dependent and is not claimed as improved here.

The benchmark used dedicated worktree-owned daemon/session state, a freshly signed fixture app, a fresh LAN Metro server, the discovered USB-connected iPhone, retained raw command/runner logs, and eight matched relaunch cycles per side. It was sequential rather than randomized, so the small failure-rate and end-to-end deltas are observational; the two cold-cold samples are descriptive only and not a reliable p95 estimate.

ADR 0005 now records the physical-device lifecycle rule. Four files changed; scope stayed within Apple application/runner lifecycle and its daemon regression seam.

## Validation

- Live physical iPhone on production-equivalent pre-rebase commit `c3bede63dc` (iOS 26.6, CoreDevice, Xcode 26.2): fresh build/install/signing, LAN Metro bundle load, snapshots/interactions, eight before and eight after relaunch cycles, and final session/daemon cleanup. Final head `e8cdd87f5c` preserves that physical-iOS path while narrowing unmeasured physical Apple OS leaves to their existing restart behavior; exact-head local and CI gates validate the final tree.
- Planted-red lifecycle regression: reverting runner retention produces `stop -> close -> open -> prewarm` instead of the required `close -> open -> prewarm -> targetReset` sequence.
- Focused lifecycle/daemon/runner suites pass, including both CoreDevice and XCTest physical backends plus relaunch-failure cleanup/error precedence.
- `pnpm check:quick` and `pnpm check:affected --run` pass on `e8cdd87f5c` (format, lint, typecheck, layering, fallow, build, unit, interaction-contract, and provider-integration green). GitHub is authoritative for coverage and hosted Apple lanes; the workflow's physical-device step was skipped, so the local device run above supplies physical proof.

Raw local measurements and per-command logs:

- `/private/tmp/agent-device-physical-perf-baseline-20260831-valid2/`
- `/private/tmp/agent-device-physical-perf-warm-20260831-retry2/`
- `/private/tmp/agent-device-physical-perf-relaunch-after-20260831/`

### DIFF
--- a/docs/adr/0005-ios-runner-interaction-lifecycle.md
+++ b/docs/adr/0005-ios-runner-interaction-lifecycle.md
@@ -62,10 +62,12 @@ root-only payload instead of issuing more flat fallback queries against the same
 the cached `XCUIApplication` handle is cleared so the next command reacquires the target through the
 normal activation path.
 
-An external iOS simulator relaunch also invalidates process-bound target state. After replacing the
-app process, the daemon sends a lifecycle reset to the retained runner so the next command reacquires
-`XCUIApplication`; the reset also clears process-bound snapshot penalty and private-AX depth state. If
-that reset cannot be confirmed, the daemon discards the runner session.
+An external iOS-family Simulator relaunch or physical iOS relaunch also invalidates process-bound
+target state. After replacing the app process, the daemon sends a lifecycle reset to the retained
+runner so the next command reacquires `XCUIApplication`; the reset also clears process-bound snapshot
+penalty and private-AX depth state. If that reset cannot be confirmed, or if relaunch fails before
+reset, the daemon discards the runner session. Other physical Apple OS leaves retain their existing
+runner-restart behavior until separately evidenced.
 
 The snapshot surface intentionally has two AX-failure shapes. Interactive fast snapshots return a
 truncated success payload with `runnerFatal` so agents can still see that AX state is unavailable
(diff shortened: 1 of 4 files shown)


thymikee:
Finding (P2): the new retention/reset branch is keyed on physical isIosFamily, which includes iOS, iPadOS, tvOS, and visionOS, but the new regression fixture fixes appleOs: `ios`. A later accidental narrowing to iOS-only would keep this test green while regressing the other admitted physical leaves. Please either cover the supported physical Apple leaves with an explicit fixture matrix (including the unsupported watchOS boundary where appropriate), or narrow the production predicate and ADR claim to the behavior actually supported. The physical-iPhone mechanism evidence itself is solid: the retained session is reused, targetReset completes, and 8/8 post-relaunch snapshots succeeded.

thymikee:
Resolved in e8cdd87f5c: runner retention is now limited to physical appleOs: ios. Added a negative leaf matrix proving iPadOS, tvOS, and visionOS preserve stop/restart behavior; watchOS remains rejected by runtime admission. The ADR now matches that scope.

PR #2215, opened 2026-09-01
perf: avoid redundant physical iOS runner health check

thymikee:
## Summary

Normal physical-iOS first open used its opportunistic runner prewarm to issue an `uptime` readiness command while the first snapshot was establishing the same runner. This keeps session startup best-effort but lets the first snapshot prove readiness, removing that redundant command only for an unawaited first open. Relaunch, simulator, explicitly awaited prewarm, explicit prepare, and other Apple targets remain unchanged.

On iphone (iOS 26.6.1), randomized 8+8 first-boot runs with stable LAN Metro and a task-owned external signed xctestrun (`buildMs: 0`) changed outer-wall open + first semantic snapshot from median/reported-p95 `13.44s / 28.08s` to `12.29s / 17.53s`. Internal daemon timing changed from `11.19s / 26.25s` to `10.38s / 15.34s`. Open-time `uptime` was present in 8/8 baseline runs and 0/8 candidate runs; all 16 runs reached the healthy tree-backed Catalog screen.

The work-count result and median improvement are causal. With eight samples, nearest-rank p95 is the single worst run, so the tail change is directional evidence rather than a stable percentile estimate. PR #2200 relaunch results are excluded.

## Validation

- Planted-red proof observed before the runtime fix for the new option and first-open ordering invariant.
- CI-planted architecture red reproduced locally: the prewarm test grew a pinned 1,325-line file to 1,338 lines.
- Extracted the prewarm family to `runner-client-prewarm.test.ts`, lowered the old file's pin to 1,280 lines, and added an awaited-prewarm health-check fence.
- Focused lifecycle, runner prewarm/retry, and size-ratchet tests: 51/51 passed.
- `pnpm check:affected --run` passed format, lint, typecheck, layering, DI seams, fallow, build, package, and Node integration before the full related-test selection hit unrelated host-load-sensitive 5-second timeouts. The scoped tests remained green; exact-head CI is authoritative for the hardening-only follow-up commit.
- Physical-device A/B campaign and final sanity flow: open, two snapshots, first interaction, and post-interaction semantic readiness all passed.

### DIFF
--- a/packages/contracts/src/application-lifecycle-runtime.ts
+++ b/packages/contracts/src/application-lifecycle-runtime.ts
@@ -156,6 +156,11 @@ export type PrepareAppleRunnerResult = Readonly<{
   failureReason?: string;
 }>;
 
+/** Controls whether an opportunistic runner session prewarm also proves readiness. */
+export type AppleRunnerSessionPrewarmOptions = Readonly<{
+  healthCheck?: boolean;
+}>;
+
 /** Individual semantic operations exposed by the application lifecycle runtime facet. */
 export type ApplicationLifecycleRuntimeOperations = Readonly<{
   resolveOpenTarget(input: OpenTargetResolutionInput): Promise<OpenTargetResolution>;
@@ -260,6 +265,7 @@ export type AppleApplicationTools = Readonly<{
     execution: ApplicationLifecycleExecution,
     signal: AbortSignal,
     propagateError: boolean,
+    options?: AppleRunnerSessionPrewarmOptions,
   ): Promise<void>;
   notifyRunnerAppRelaunched(
     device: DeviceInfo,
--- a/packages/platform-apple/src/lifecycle.test.ts
+++ b/packages/platform-apple/src/lifecycle.test.ts
@@ -96,6 +96,39 @@ test('starts an unawaited physical iOS first-open runner without a redundant hea
   });
 });
 
+test('preserves the health check when physical iOS runner prewarm is awaited', async () => {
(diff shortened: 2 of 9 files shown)


thymikee:
Reviewed exact head `8580c8b307a45feada01e20c73b9d206dba33455`: no code findings. The lifecycle owner selects `healthCheck: false` only for unawaited physical-iOS first open, while the local runner owner starts the keyed session without issuing redundant `uptime`; awaited/relaunch and other target paths preserve readiness behavior. Focused regression tests prove both selection and zero-command execution, and the 8+8 physical-device evidence plus final interaction flow is credible. READY for human code review; exact-head CI and iOS Smoke are still running, so this is not yet merge-ready.

thymikee:
Post-merge finding on exact head `8580c8b307a45feada01e20c73b9d206dba33455`: the optimization is broader than the stated first-open invariant. `isUnawaitedPhysicalIosOpen()` selects `healthCheck: false` for every physical-iOS non-relaunch, non-explicit-prewarm open, but does not require `!input.hasExistingSession`. An ordinary open inside an existing daemon session therefore also skips the cached-runner readiness proof, although the body and planted regression cover only a fresh first open. Please fix this at the lifecycle selection owner by adding the existing-session boundary and an adversarial existing-session/non-relaunch test that preserves the default health check, while retaining the fresh-session zero-command test. The exact-head iOS Smoke failure is simulator-only and source-disjoint from this physical-device branch, but the required lane was red at merge time.

PR #2218, opened 2026-09-01
fix: preserve physical iOS session health check

thymikee:
## Summary

Restore physical iOS runner health proof for an ordinary open in an existing session.

PR #2215 intentionally removed the redundant uptime probe from an unawaited first open, but its predicate also matched later opens in an active session. This keeps the first-open optimization and limits `healthCheck: false` to sessions that do not already exist.

Adds an adversarial lifecycle regression covering the existing-session boundary. Two files changed; scope stayed within Apple application lifecycle.

## Validation

- Planted red: the new lifecycle test received `{ healthCheck: false }` on the pre-fix source.
- Focused green: lifecycle 9/9; runner prewarm contract 3/3.
- Physical iPhone: an ordinary same-session Settings open was followed by a warm, healthy XCTest tree snapshot in 1.09s (116.96ms acquisition); session close and retained-runner cleanup succeeded.
- `pnpm check:affected --run`: format, lint, typecheck, layering, fallow, and build passed. The related-test lane encountered timeout-shaped host contention while other worktrees were running gates; its initial six failures passed serially 38/38, and the later changing failures had passed in the prior run. GitHub CI remains authoritative for the exact head.

No CLI, help, docs, or skill surface changed.

### DIFF
--- a/packages/platform-apple/src/lifecycle.test.ts
+++ b/packages/platform-apple/src/lifecycle.test.ts
@@ -96,6 +96,38 @@ test('starts an unawaited physical iOS first-open runner without a redundant hea
   });
 });
 
+test('preserves runner health proof for an unawaited physical iOS open in an existing session', async () => {
+  const signal = new AbortController().signal;
+  const events: string[] = [];
+  const interactor = {
+    open: vi.fn(async () => {
+      events.push('open');
+    }),
+  } as unknown as Interactor;
+  const baseHost = platformRuntimeHostFixture();
+  const prewarmRunnerSession = vi.fn(async () => {
+    events.push('prewarm');
+  });
+  const host = {
+    ...baseHost,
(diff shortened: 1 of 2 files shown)


thymikee:
Re-reviewed exact head 7c56d1e0d72dcd050a044e8ca8d5d8b08ee30e4c: READY for human review, no code finding. The fix is at the lifecycle owner: healthCheck: false now applies only to an unawaited physical-iOS open with no existing daemon session; existing-session opens retain the runner readiness/uptime proof. The production route supplies the existing-session fact, the adversarial regression fails if the guard is removed, and #2215’s first-open optimization remains covered. The reported same-session physical-iPhone open, healthy XCTest snapshot, and cleanup are sufficient for this narrow device-facing change; size impact is negligible. ready-for-human is appropriate. Exact-head iOS Smoke is still in progress, so not yet merge-ready.

PR #2220, opened 2026-09-01
docs: adopt Simlock-backed Host device allocation

thymikee:
## Summary

- adopt ADR 0021, keeping Simlock as the sole managed-device allocator while Host owns remote authorization, attribution, and supervision
- define the managed runtime, lease, durable execution-claim, recovery, and maintenance boundaries for iOS and Android
- align ADR 0007 and the ADR index with the accepted Host durability boundary
- keep the change documentation-only and limited to the Host/managed-device architecture

## Validation

- `pnpm build`
- `pnpm check:affected --run`
- `git diff --cached --check`

### DIFF
--- a/docs/adr/0007-remote-device-leases.md
+++ b/docs/adr/0007-remote-device-leases.md
@@ -73,6 +73,20 @@ draining removes only that request's pending hold, leaving successor and unrelat
 canceling activation does not cancel the mutations being drained. Completed holds use their TTL or
 explicit release lifecycle.
 
-Holds, like leases, are in-memory and do not survive daemon restart. Controllers must reconnect and
-re-establish them; no persisted hold store is used. Local takeover is deferred: a future host-global
-human-control fence must coexist with the local session's device claim, not acquire it exclusively.
+Holds and ordinary proxy leases are in-memory and do not survive daemon restart. Controllers must
+reconnect and re-establish them; no persisted hold store is used. Local takeover is deferred: a
+future host-global human-control fence must coexist with the local session's device claim, not
+acquire it exclusively.
+
+## Host managed-device durability amendment
+
+ADR 0021 adds a narrow durability exception for Host leases backed by a managed-device allocator.
+Before allocator acquisition, the daemon persists a non-authoritative allocation operation record;
+Host adds its asserted principal, Host lease id, and run/client attribution. After grant, it records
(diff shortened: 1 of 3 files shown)


thymikee:
BLOCKED at exact head d14d846a0b51077f9a7c297475fe51c991c92970. ADR 0021 §4 says only explicit cancellation or supersession terminalizes a durable allocation request, while also defining capacity and low-disk refusal outcomes. That leaves those refusals nonterminal: replaying the same requester/key can never observe newly freed capacity, but a new key is barred unless it uses supersession, whose contract assumes a prior generation/binding. This conflicts with linked Simlock #72, which requires lookup to distinguish terminal failure. Define explicit terminal allocation-failure semantics and the safe retry shape (new logical request versus a narrowly permitted retry), while keeping idempotent replay for lost-response safety; align the Simlock dependency contract if needed. Docs-only CI is clean and no device evidence is owed. No other concrete finding.

thymikee:
Addressed at exact head `5e6f912955`.

- Capacity, disk, validation, and provisioning failures are now terminal for that allocation attempt, even when cleanup remains capacity-bearing.
- Replaying the same idempotency key always returns the stored outcome; it never re-evaluates newly available capacity.
- A retry after a terminal outcome uses a new key and becomes the next generation without supersession because the old generation can no longer grant.
- A nonterminal generation still requires fenced, authorized supersession.
- Capacity retry uses `retryAfterMs`; disk retry requires operator remediation first.
- The linked [Simlock #72](https://github.com/callstackincubator/simlock/issues/72) contract now states the same terminal-disposition and new-key behavior, with acceptance cases for stable replay and next-generation retry.

Validation: `pnpm check:affected --run` passed; the selector correctly found no runtime checks for the docs-only ADR diff.

thymikee:
Reviewed exact head `5e6f912955bbd7a93ae4b67fb91ac1b6ed1870fa`. Clean docs-only verdict: ADR 0021 gives Simlock sole managed-device allocation/lifecycle ownership while Host retains authorization, attribution, runtime binding, artifacts, and supervision; ADR 0007 records only a non-authoritative durable handoff mapping. The terminal allocation-failure and new-key retry semantics now align with linked Simlock #72, preserving idempotent replay without blocking future capacity retries. The three-file ADR delta is mergeable and all applicable exact-head checks are green. Ready for human review; implementation remains tracked separately in Simlock #70/#72.

PR #2234, opened 2026-09-02
fix(android): apply settings airplane through the connectivity service

### LINKED ISSUE #2223: settings airplane is a no-op on modern Android and leaves airplane_mode_on=1 with the network still up
## Summary

`agent-device settings airplane on` does not take an Android device offline. It writes `airplane_mode_on=1` and then sends an `android.intent.action.AIRPLANE_MODE` broadcast, which requires a system permission and fails. Because the `settings put` runs first and succeeds, the device is left reporting airplane mode while connectivity is fully up.

This fails *dirty* rather than cleanly: the CLI surfaces an error, but the device state has already been half-mutated, so anything that checks `airplane_mode_on` sees `1` while traffic still flows.

## Repro

Android emulator, `ro.build.version.release=17`, `ro.build.version.sdk=37`:

```console
$ agent-device settings airplane on --platform android
Error (COMMAND_FAILED): Exception occurred while executing 'broadcast': Hint: Retry with --debug and inspect diagnostics log for details.

# diagnostics ndjson
{"level":"error","phase":"platform_command","command":"settings","data":{"error":"adb exited with code 255"}}

$ adb shell settings get global airplane_mode_on 1  # <- setting landed

$ adb shell ping -c 2 8.8.8.8
2 packets transmitted, 2 received, 0% packet loss   # <- still online
```

Expected: the device is offline, or the command fails without changing device state.
Actual: the command fails, `airplane_mode_on=1`, and the network is unaffected.

## Why this happens in agent-device

`dist/src/android.js`, the `settings` handler's `airplane` case:

```js
case `airplane`: {
  let t = A(r), n = t ? `1` : `0`, i = t ? `true` : `false`;
  await d(e, [`shell`, `settings`, `put`, `global`, `airplane_mode_on`, n]),
  await d(e, [`shell`, `am`, `broadcast`, `-a`,
              `android.intent.action.AIRPLANE_MODE`, `--ez`, `state`, i]);
  return
}
```

1. `settings put global airplane_mode_on <0|1>` succeeds — it only writes the setting.
2. `am broadcast -a android.intent.action.AIRPLANE_MODE` is rejected; broadcasting that action is restricted to system callers, so `adb` exits 255.
3. Nothing ever tells the connectivity stack to act, so the radio state never changes.

Two notes:

- **Not a regression.** I diffed the published `0.20.0` and `0.20.10` tarballs and this block is identical (modulo minified identifier renaming), so upgrading does not help.
- **The `wifi` case has a related gap.** `settings wifi off` runs `svc wifi disable`, which genuinely disables Wi-Fi, but an emulator then falls back to cellular and stays online — so it also cannot be used to reach an offline state.

## Suggested fix

`cmd connectivity airplane-mode <enable|disable>` works on Android 11+, needs no root, and actually drives the connectivity stack. Verified on the same API 37 emulator:

```console
$ adb shell cmd connectivity airplane-mode enable
$ adb shell cmd connectivity airplane-mode 
enabled
$ adb shell ping -c 2 8.8.8.8
connect: Network is unreachable
```

Round trip back to `disable` restores connectivity cleanly. Latency from issuing the command to traffic actually failing was measured at 65–365ms across trials.

Suggested shape: try `cmd connectivity airplane-mode` first, and fall back to the existing `settings put` + broadcast pair only if that command is unavailable, so old devices keep whatever behaviour they have today. If the fallback is kept, it would be worth ordering the broadcast before the `settings put`, or reverting the setting on broadcast failure, so a failure cannot leave `airplane_mode_on` inconsistent with actual connectivity.

## Related: no `setAirplaneMode` in the Maestro dialect

Possibly a separate issue, but it is the same "Maestro parity" shape as #2159.

| | Real Maestro | agent-device `--maestro` |
| --- | --- | --- |
| `setAirplaneMode` | supported (Android) | not implemented |
| `toggleAirplaneMode` | supported (Android) | not implemented |

Maestro documents both commands as Android-supported. In `--maestro` mode, neither is recognized, and the `runScript` sandbox exposes only `output`, `json`, and `http.post` — no `require`/`process` — so a flow cannot reach `adb` by any other route either.

The practical effect is that a flow cannot change connectivity mid-run at all: offline coverage has to be split into separate flows driven by an external script, or routed through a host-side HTTP sidecar called via `http.post`. Since `settings airplane` already exists internally, wiring those two Maestro commands to it looks like a small addition once the underlying toggle works.

## Environment

- `agent-device` `0.20.0` and `0.20.10` (same behaviour on both)
- Android emulator, Pixel 10, `ro.build.version.release=17` / `sdk=37`
- macOS host

thymikee:
## Summary

`settings airplane on|off` did not take an Android device offline, and it failed *after* changing
device state. The Android owner wrote `airplane_mode_on` and then broadcast
`android.intent.action.AIRPLANE_MODE`, which Android refuses for non-system callers, so the command
surfaced `COMMAND_FAILED` while the device reported airplane mode with the radios still up —
anything reading `airplane_mode_on` saw a state that traffic contradicted.

Airplane mode is now applied by the component that owns it. `cmd connectivity airplane-mode` both
drives the connectivity stack and reports the state, so the owner:

- reads the state first, which is what proves the build supports the operation — a build that
  cannot answer is refused with `UNSUPPORTED_OPERATION` before anything is written, so the write is
  unreachable rather than guarded;
- reads it again after the change, so the response carries the mode connectivity holds, not the one
  that was requested;
- issues no `settings put` and no broadcast on any path.

```console
$ agent-device settings airplane on --platform android --json
{ "setting": "airplane", "state": "on", "airplaneMode": "enabled", "message": "Updated setting: airplane" }
```

Per triage the old write/broadcast pair is not kept as a compatibility fallback: it is known to fail
dirty. Maestro `setAirplaneMode`/`toggleAirplaneMode` parity stays out of scope for a separate issue.

Closes #2223

## Validation

Live Android emulator (`sdk_gphone64_arm64`, release 16 / API 36), through the built CLI at
`5ebb770b3a`:

- `settings airplane on` answered `airplaneMode: enabled`. Four seconds later the device reported
  `cmd connectivity airplane-mode` = `enabled`, `airplane_mode_on` = `1`, and `ping 8.8.8.8` =
  `connect: Network is unreachable`.
- `settings airplane off` answered `airplaneMode: disabled`, and connectivity came back: state
  `disabled`, `airplane_mode_on` = `0`, ping 2/2 received.
- The reported cause reproduces on the same device: `am broadcast -a
  android.intent.action.AIRPLANE_MODE` fails with `SecurityException: Permission Denial … uid=2000`.

Seven scenarios in `packages/platform-android/src/__tests__/settings-airplane.test.ts` drive the
public `setAndroidSetting` entry: the enable and disable round trips assert the exact adb call list,
which is what a revert to `settings put` + broadcast fails; a service that accepts the write without
changing state must not be reported as success; unsupported and unreadable builds are refused with
the read as the only call issued; an adb transport failure stays `COMMAND_FAILED` with its reconnect
hint instead of being mistyped as unsupported; and a read that fails only after the change is a
command failure, not an unsupported build. All seven were observed red against the pre-fix owner.

`pnpm check:affected --run` is green except for `test/integration/provider-scenarios/limrun-ios-snapshot-owner.test.ts`,
which times out identically on `main` at `7ee1a5ded7` with this branch checked out or not — a
pre-existing failure on that iOS provider scenario, untouched by this change.

## Notes

- Six files: the Android settings owner and its new airplane module plus tests, the `settings`
  command help, the command reference, and the changelog.
- Residual risk: the `UNSUPPORTED_OPERATION` leaf is fixture-covered only — no pre-Android-11 device
  was available. Nothing can be written on that path, since the write follows the read that refuses.
- Connectivity takes a moment to settle after the switch (offline observed within seconds, not
  instantly). The command reference now says so, so callers poll the app under test instead of
  asserting offline behaviour on the next line.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 ## Unreleased
 
+- Fixed: `settings airplane on|off` now takes an Android device offline. It is applied through
+  the connectivity service (`cmd connectivity airplane-mode`), which drives the radios, instead of
+  writing `airplane_mode_on` and broadcasting `ACTION_AIRPLANE_MODE_CHANGED` — a broadcast Android
+  refuses for non-system callers, so the old path failed *after* writing the setting and left the
+  device reporting airplane mode with the network still up (#2223). The response now reports the
+  `airplaneMode` the connectivity service holds after the change, and an Android build that does not
+  expose that command is refused with `UNSUPPORTED_OPERATION` before anything is written.
 - Breaking (0.21): iOS Appium/WebDriver snapshots now expose engine-owned acquisition facts and
   typed fidelity warnings. The SDK snapshot `truncated` field is optional when Appium cannot report
   hierarchy completeness; regular snapshots fail closed without valid viewport evidence, while
--- a/packages/platform-android/src/__tests__/settings-airplane.test.ts
+++ b/packages/platform-android/src/__tests__/settings-airplane.test.ts
@@ -72,7 +72,7 @@ test('setAndroidSetting airplane refuses builds without the connectivity command
     async ({ calls, device }) => {
       await assertRejectsAppError(() => setAndroidSetting(device, 'airplane', 'on'), {
         code: 'UNSUPPORTED_OPERATION',
-        message: /does not report airplane mode through the connectivity service/,
+        message: /no airplane-mode command/,
         hint: /Android 11 \(API 30\)/,
       });
       assert.deepEqual(
@@ -88,8 +88,28 @@ test('setAndroidSetting airplane refuses unreadable state before writing', async
     () => 'Airplane mode: who knows',
     async ({ calls, device }) => {
       await assertRejectsAppError(() => setAndroidSetting(device, 'airplane', 'off'), {
-        code: 'UNSUPPORTED_OPERATION',
-        message: /does not report airplane mode through the connectivity service/,
+        code: 'COMMAND_FAILED',
+        message: /Failed to read Android airplane mode/,
+      });
+      assert.deepEqual(
+        calls.map((args) => args.join(' ')),
+        [READ],
+      );
+    },
+  );
+});
+
+test('setAndroidSetting airplane keeps a refused read a command failure, not an unsupported build', async () => {
+  await withFakeAdb(
+    () => ({
+      stdout: '',
+      stderr: 'java.lang.SecurityException: Permission Denial: not allowed to change airplane mode',
+      exitCode: 255,
+    }),
+    async ({ calls, device }) => {
+      await assertRejectsAppError(() => setAndroidSetting(device, 'airplane', 'on'), {
+        code: 'COMMAND_FAILED',
+        message: /Failed to read Android airplane mode/,
       });
       assert.deepEqual(
         calls.map((args) => args.join(' ')),
--- a/packages/platform-android/src/__tests__/settings-airplane.test.ts
+++ b/packages/platform-android/src/__tests__/settings-airplane.test.ts
@@ -1,0 +1,137 @@
+import { test } from 'vitest';
+import assert from 'node:assert/strict';
+import { setAndroidSetting } from '../settings.ts';
+import type { AndroidAirplaneMode } from '../settings-airplane.ts';
+import { assertRejectsAppError } from './test-utils/app-error.ts';
+import { withFakeAdb, type FakeAdbScript } from './test-utils/fake-adb.ts';
+
+const READ = 'shell cmd connectivity airplane-mode';
+
+/**
+ * An Android build whose connectivity service owns airplane mode. `honorWrites: false` models the
+ * service accepting the command and leaving the state alone, which is what a silently ineffective
+ * airplane-mode path looks like from the outside.
+ */
+function connectivityService(
+  initial: AndroidAirplaneMode,
+  options: { honorWrites?: boolean } = {},
+): FakeAdbScript {
+  let state = initial;
+  return (args) => {
+    const flat = args.join(' ');
+    if (flat === READ) return state;
+    if (flat === `${READ} enable` || flat === `${READ} disable`) {
+      if (options.honorWrites !== false) state = flat.endsWith('enable') ? 'enabled' : 'disabled';
+      return '';
+    }
+    return { stderr: `unexpected args: ${flat}`, exitCode: 1 };
+  };
+}
+
+test('setAndroidSetting airplane on enables through connectivity and reports its state', async () => {
+  await withFakeAdb(connectivityService('disabled'), async ({ calls, device }) => {
+    const result = await setAndroidSetting(device, 'airplane', 'on');
+    assert.deepEqual(result, { airplaneMode: 'enabled' });
+    assert.deepEqual(
+      calls.map((args) => args.join(' ')),
+      [READ, `${READ} enable`, READ],
+    );
+  });
+});
+
+test('setAndroidSetting airplane off disables through connectivity and reports its state', async () => {
+  await withFakeAdb(connectivityService('enabled'), async ({ calls, device }) => {
+    const result = await setAndroidSetting(device, 'airplane', 'off');
+    assert.deepEqual(result, { airplaneMode: 'disabled' });
+    assert.deepEqual(
+      calls.map((args) => args.join(' ')),
+      [READ, `${READ} disable`, READ],
+    );
+  });
+});
+
+test('setAndroidSetting airplane reports the state connectivity holds, not the one requested', async () => {
+  await withFakeAdb(
+    connectivityService('disabled', { honorWrites: false }),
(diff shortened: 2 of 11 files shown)


thymikee:
Reviewed exact head `5ebb770b3a6fff5e1e8f0f43cf12a0b847d5fb5c`. Two blockers remain:

1. `settings-airplane.ts` currently maps every unrecognized nonzero initial probe failure to `UNSUPPORTED_OPERATION` unless the generic adb transport classifier happens to recognize it. A permission denial, connectivity-service crash/internal error, or other execution failure would therefore be reported as ‘requires Android 11; use a newer device’ rather than `COMMAND_FAILED`. Narrow unsupported detection to explicit capability-absence output (`Unknown command`/missing shell implementation or equivalent); preserve all other nonzero results through `androidAdbResultError`. Add the closest-negative regression for an unclassified nonzero failure and prove the write remains unreachable.

2. Exact-head Coverage has a code-owned eager-closure failure: `packages/platform-android/src/mechanics.ts` evaluates 178 modules versus the pinned 177 because `settings.ts` statically imports the new airplane module. The extraction itself is reasonable given `settings.ts` size, but this loading-shape change must be measured and owned: either keep airplane-only code lazy from the broad mechanics surface or justify/update the budget to 178, then rerun Coverage green.

Otherwise the fix is sound: the public Android owner follows connectivity read → write → read, reports observed state, removes the dirty settings/broadcast path, has credible revert-red focused tests and built-CLI emulator round-trip evidence, and updates help/docs/changelog. Size is healthy (`+1.2 kB` npm unpacked); every other exact-head check, including all smoke lanes, is green. No `ready-for-human` label until both blockers are resolved.

thymikee:
Both blockers are addressed at `e9a2417218`.

**1. Unsupported detection narrowed to capability absence.** `UNSUPPORTED_OPERATION` is now selected only by the prose adb prints when a build ships no shell implementation for a `cmd` invocation; every other failed read — permission denial, connectivity-service error, transport failure, unparseable output at a clean exit — goes through `androidAdbResultError`, so it stays `COMMAND_FAILED` with its classified hint.

The predicate that reads that prose already existed as `isClipboardShellUnsupported` in `adb.ts`, doing exactly this job for `cmd clipboard`. Rather than write a second message sniff, it is renamed `isAndroidShellCommandUnsupported` and both services use it; its "only ask this about a call that failed" contract is honored on both sides (airplane mode consults it only on a nonzero exit, since a parseable state settles a clean exit first).

Regression coverage, all red against `5ebb770b3a`:

- `SecurityException: Permission Denial …` at exit 255 → `COMMAND_FAILED`, and the adb call list is exactly `[read]`, so the write stays unreachable.
- unreadable output at exit 0 → `COMMAND_FAILED`, same single-call list (it previously read as unsupported).
- `Unknown command: airplane-mode` at exit 255 → still `UNSUPPORTED_OPERATION`, still one call.
- `error: device offline` → `COMMAND_FAILED` carrying the reconnect hint, unchanged.

**2. Eager-closure budget owned, not dodged.** `packages/platform-android/src/mechanics.ts` moves to 178 in `scripts/__tests__/eager-closure-budgets.ts`. Keeping the airplane owner lazy would make it the one settings helper behind a function-scoped import while `settings-permission.ts` and `settings-parsing.ts` load statically, and the mechanics facet is implementation-eager by design once selected — so the honest answer is the row move the gate's own failure message prescribes. Measured locally: 178 with the pin at 177 fails naming this entry, 178 passes.

Re-validated the round trip on the same emulator (`sdk_gphone64_arm64`, release 16 / API 36) with the CLI built at `e9a2417218`: `airplane on` → `airplaneMode: enabled`, `airplane_mode_on=1`, `ping 8.8.8.8` = `connect: Network is unreachable`; `airplane off` → `airplaneMode: disabled`, `airplane_mode_on=0`, ping 2/2. Session-free run; the isolated daemon was stopped and its state dir removed.

`pnpm check:affected --run` is green apart from `test/integration/provider-scenarios/limrun-ios-snapshot-owner.test.ts`, which still times out identically on an unmodified `main` at `7ee1a5ded7` — pre-existing and unrelated to this branch.

thymikee:
Re-reviewed exact head `e9a241721864dd7737f281c42e59cb9bc5e6935b`: code review is clean and the PR is ready for human review.

The prior failure-classification blocker is fixed: unsupported detection is limited to explicit missing-shell/unknown-command output, while the SecurityException closest-negative remains `COMMAND_FAILED` and proves the write is unreachable. The +1 intentionally eager mechanics closure is measured and justified, the successful connectivity read → write → read path is unchanged, all exact-head checks are green, and the live emulator evidence at `5ebb770` remains applicable to that unchanged success path.

Non-blocking PR-body cleanup: it still says seven scenarios and six files; the current head has eight scenarios and 11 changed files. Please update those counts and record the green exact-head CI.

PR #2233, opened 2026-09-02
refactor(ios): carry provider acquisitions through one presentation owner

thymikee:
Closes #2231

## Summary

- Carry Appium/WebDriver and Limrun iOS acquisitions as typed facts to one host-side presentation owner.
- Reuse the shared viewport and residue derivation rules; keep provider adapters limited to acquisition and preserve Android WebDriver parsing in an Android-owned module.
- Delete provider-side presentation/request/warning/ref-shaping paths and move real Limrun runtime-to-daemon/public proof into the provider-scenario harness.

## Evidence

- Exact head: `be878f39a38f784703e4cc4edfedaccc09b81c13`
- `pnpm build` passed.
- `pnpm format:check`, `pnpm lint`, `pnpm typecheck`, `pnpm check:layering`, and `pnpm check:fallow` passed.
- The R73 ownership gate was observed red against a planted duplicate presentation call, naming the violating adapter, then green after the violation was removed.
- Focused unit and provider-scenario suites passed, including the real Limrun runtime-to-daemon/public snapshot proof.
- `pnpm check:affected --run` passed all changed suites and 404/405 test files (2785/2786 tests); the sole failure was the unrelated foreign-owner cleanup test in `src/daemon/__tests__/request-router-open.test.ts`, which passed all 14 tests in isolation.
- The requested read-only Claude Opus adversarial review was attempted once at the settled pre-amend head; the local reviewer process hung and exited without output, so no external findings were available.

## Change accounting

For the exact commit diff (`b251a9769^..b251a9769`):

- Gross total: `+1189 / -1000` lines, net `+189`.
- Production source: `+491 / -459`, net `+32`.
- Tests/fixtures: `+506 / -541`, net `-35`.
- Layering tooling/tests: `+192 / -0`, net `+192`.
- Provider production (`packages/provider-webdriver/src` and `packages/provider-limrun/src`): `+292 / -454`, net `-162`.

The production addition is the shared carrier/presenter and enforcement surface; provider/runtime production remains net-negative through deletion and reuse.

## Deletion and reuse decisions

- Deleted provider-side iOS request/plan/publish/warning/ref-strip wrappers and the obsolete daemon truncation test.
- Reused the capture-kit viewport resolver, acquisition residue derivation, and central iOS snapshot engine.
- Preserved Android WebDriver’s derived parser as Android-owned code while sharing only low-level XML/attribute primitives.
- No docs or skills changed; no compatibility fallback was added.

PR #2235, opened 2026-09-02
fix(daemon): skip sessionless local host teardown for provider devices

thymikee:
## What changed

`withSessionlessRunnerCleanup` no longer releases the local iOS execution host when the device belongs to a provider runtime.

```ts
if (!session && !isActiveProviderDevice(device)) {
  await platformCleanup!.cleanupSessionlessExecutionHost(device);
}
```

## Why

A sessionless `snapshot`, `diff`, `alert`, `settings`, `wait` or `is` command releases the local XCTest execution host when it completes. That teardown ran for every iOS device, including a provider-owned one.

For a Limrun lease the device id is `limrun:ios:<leaseId>` with `kind: 'simulator'`, so the Apple path ran on the host for a device the host does not own:

```
dispatchSnapshotRuntimeCommand            src/daemon/snapshot-command-runtime.ts:66
 └ withSessionlessRunnerCleanup (finally) src/daemon/snapshot-session.ts:30
    └ cleanupSessionlessExecutionHost     src/platform-runtime-resource-cleanup.ts:70
       └ closeIosApp → terminateIosSimulatorApp
          └ ensureBootedSimulator → runXcrun   (3 × xcrun simctl + retry backoff ≈ 5.8 s)
```

`ensureDeviceReady` already declines local readiness work for a provider-owned device. The teardown side was not symmetric. One guard covers all four call sites of the helper.

## How it surfaced

`test/integration/provider-scenarios/limrun-ios-snapshot-owner.test.ts` (added in #2233) exceeded the 5000 ms default timeout on macOS. The request always settled — it took about 6.2 s, and the local `xcrun` work was almost all of it. That made `pnpm check:affected --run` red for unrelated changes.

The path itself predates #2233; that test is the first end-to-end provider scenario that reaches it.

## Notes for a reviewer

- CI stayed green because the provider-integration lane runs on `ubuntu-latest`, where `xcrun` does not exist and the calls fail immediately. The defect is only visible on macOS.
- The new test in `snapshot-session-cleanup.test.ts` fails when the guard is removed, so it is not vacuous.
- `stopIosRunnerSession` and `cleanupOwnedIosRunnerLease` were also being called for the provider device. Both operate on local runner-session state, so they were no-ops for it.

## Verification

- `limrun-ios-snapshot-owner.test.ts`: 6.2 s timeout → 294 ms, passing.
- Full `provider-integration` lane: 55 files, 164 tests, all pass.
- The 8 unit files that cover this seam: 88 tests, all pass.
- `pnpm typecheck`, `oxlint`, `oxfmt --check`, and the layering guard all pass.

### DIFF
--- a/src/daemon/handlers/__tests__/snapshot-session-cleanup.test.ts
+++ b/src/daemon/handlers/__tests__/snapshot-session-cleanup.test.ts
@@ -1,4 +1,4 @@
-import { beforeEach, expect, test, vi } from 'vitest';
+import { afterEach, beforeEach, expect, test, vi } from 'vitest';
 
 vi.mock('@agent-device/platform-apple/runner/operations', async (importOriginal) => ({
   ...(await importOriginal<typeof import('@agent-device/platform-apple/runner/operations')>()),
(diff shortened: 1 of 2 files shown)


thymikee:
Reviewed exact head `c55ae9e247fc482d5d770bcb2fa2271f40b61737`: **no valid findings**.

Provider cleanup ownership is preserved because the skipped helper owns only local iOS execution-host teardown; provider resources remain with their runtime/lease owner. Local sessionless and session-backed paths are unchanged, and the Limrun-shaped regression is non-vacuous. All exact-head checks are green, and npm size is effectively flat.

PR #2240, opened 2026-09-02
refactor(layering): one retired-paths rule for src/utils and src/replay

thymikee:
## Summary

Layering rules R14 (`src-utils-retirement`) and R71 (`replay-ownership`) were the same predicate, a path-prefix denylist for a retired directory, written twice in two modules with two tests. They now live in one module, `scripts/layering/retired-paths-policy.ts`, driven by a table keyed by rule id (`R14`, `R71`) that holds the retired prefix and the owner guidance. Adding a retired directory is one table row.

Gate behavior is unchanged: both rule ids, both messages, and both inputs survive as-is (R14 still scans every tracked path under `src/utils`, including the exact root entry; R71 still scans production sources under `src/replay/`). `check.ts` keeps its keyed rule record and reporting order. `rule-ids.ts` still sees each id declared exactly once as a whole literal.

6 files touched, all under `scripts/layering/`; no fallow baseline entries referenced the deleted modules.

## Validation

- `pnpm check:layering` (folded test file plus `rule-ids` and `check-wiring` tests, then the guard): green, 1579 source files.
- `pnpm check:quick`, `pnpm format:check`, `pnpm check:fallow`: green.
- Planted red at fd4d0f0ae2 with `src/utils/planted.ts` and `src/replay/planted.ts` staged, then `node --experimental-strip-types scripts/layering/check.ts` (exit 1):
  - `::error file=src/utils/planted.ts,line=1,title=Layering drift (R14 src-utils-retirement)::src/utils is retired; move this path to its owning package or command zone`
  - `::error file=src/replay/planted.ts,line=1,title=Layering drift (R71 replay-ownership)::src/replay/ is retired; caller source acquisition belongs under src/commands/replay/ and replay-test presentation belongs under src/cli/replay-test/.`
  Planted files removed afterwards.
- Full affected gate: pending (serialized gate stage will append the result)


thymikee:
READY — clean review at this exact head. R14/R71 retain their exact identities, messages, inputs and wiring; the shared retired-path owner preserves the original behaviors and the planted regressions exercise both real tracked zones. Deleted modules have no consumers and all exact-head CI is green. No code findings. The PR body still lists the local full affected gate as pending, but GitHub’s code-owning lanes are complete.

PR #2238, opened 2026-09-02
chore(gates): drop the test-file size pin map, keep the merge-base ratchet

thymikee:
## Summary

The test-file size gate had two rules: an exact-length pin map for every test file over 1,000 lines, and a history-backed rule against the merge-base. The pin map duplicated what git already records and turned every shrink into a two-file edit. This drops the pin map and the three assertions that served only it (pin equals file, pin not above base, pin-list-empty kill criterion).

Gate behavior for test authors now:

- A test file over the 1,000-line tripwire may be no longer than at the merge-base with `origin/main`.
- A file absent at the merge-base may not cross the tripwire.
- Shrinking an over-tripwire file needs no gate edit.

Kill criterion: no test file under the walked roots exceeds the tripwire at the merge-base; delete the gate when that holds. `docs/agents/testing.md` no longer mentions pins; a stale citation in `eager-closure-budgets.ts` naming this gate as an equality-pin example is removed. Nothing else imported `PINNED_TEST_FILE_LINES`.

## Validation

Focused: `npx vitest run --project unit-core scripts/__tests__/test-file-size-ratchet.test.ts` (3 passed), `pnpm check:quick`, `pnpm check:gate-manifest`, `pnpm check:fallow`, all clean at 63ed6a2e64.

Planted red: a scratch commit appending 3 lines to `src/__tests__/cli-network.test.ts`, same vitest command:

```
× no test file over the tripwire is longer than at the merge-base, and no new file crosses it
+ "src/__tests__/cli-network.test.ts is 1095 lines, 1092 at the merge-base: a test file over the tripwire may not grow; split it along the source module it mirrors (docs/agents/testing.md)."
```

Planted green: a scratch commit deleting 3 lines from that file, no gate edit, passed 3/3. Both scratch commits were discarded before push.

Full affected gate: pending (serialized gate stage will append the result)

PR #2249, opened 2026-09-02
test(replay): extract shared fixtures from the target-verification suite

thymikee:
## Summary

Task B1b, largest file only. The 26 scenarios in `session-replay-target-verification-runtime.test.ts` each hand-built the same temp root, iOS app session, written `.ad` script, request-recording `invoke`, and one-button XCTest capture literal. That setup now lives as named exports in the sibling `session-replay-scenario.fixtures.ts` (`replayScriptScene`, `saveButtonCapture`, `emptyCapture`, annotation constants), built on the existing `test-utils` builders and replay runtime fixtures. Assertions, titles, and comments are unchanged.

- Tests: 26 -> 26 (`vitest --reporter=json` numTotalTests); `expect(` calls: 128 -> 128.
- jscpd (`--min-tokens 80 --min-lines 8`) duplicated lines within the suite: 1,046 (25 pairs) -> 125 (6 pairs).
- Gross diff: 949 lines, so the other two suites did not fit the 1,000-line budget.

## Validation

- `npx vitest run --project unit-core src/daemon/replay/internal/__tests__/session-replay-target-verification-runtime.test.ts`: 26/26 green.
- `pnpm check:quick`, `oxfmt --check`, `scripts/layering/check.ts`, `pnpm check:production-exports`: green.
- Planted red 1: production `session-replay-divergence.ts:276` returns the first attempt unconditionally (retry disabled) -> `a transient content-quality capture failure right after launch recovers within the bounded retry...` FAIL, `AssertionError: expected false to be true`. Restored.
- Planted red 2: production `session-replay-target-verification.ts:319` kind `identity-unverifiable` -> `identity-mismatch` -> `a recorded-unverifiable annotation is an identity-unverifiable divergence with matchCount omitted` FAIL, `expected 'identity-mismatch' to be 'identity-unverifiable'`. Restored.
- Full affected gate: green at 8021503c665f2fc1f72a3e438d16ce3a3ea4704b (pnpm check:affected --run; gates run: format, lint, typecheck, fallow, vitest-related)

## Tradeoffs / follow-ups

- Remainder: `session-replay-divergence.test.ts` (472 duplicated lines, 12 pairs) and `session-replay-repair-empty-tail.test.ts` (490, 9 pairs) need their own extraction; the divergence suite can likely reuse `replayScriptScene`.
- No size pin to lower: #2238 removed the pin map.

### DIFF
--- a/src/daemon/replay/internal/__tests__/session-replay-scenario.fixtures.ts
+++ b/src/daemon/replay/internal/__tests__/session-replay-scenario.fixtures.ts
@@ -1,0 +1,137 @@
+import path from 'node:path';
+import { mkdtempForTestSync } from '../../../../__tests__/test-utils/tmp-dir.ts';
+import { makeIosAppSession } from '../../../../__tests__/test-utils/session-factories.ts';
+import { SessionStore } from '../../../session-store.ts';
+import type { DaemonInvokeFn, DaemonRequest, DaemonResponse } from '../../../types.ts';
+import { runReplayForTest } from '../../__tests__/replay-command-fixture.ts';
+import {
+  baseReplayRequest,
+  writeReplayFile,
+} from '../../__tests__/session-replay-runtime.fixtures.ts';
+
+/** A temp root with one seeded session in its store: the ground every replay scenario stands on. */
+type ReplaySessionScene = Readonly<{
+  root: string;
+  sessionStore: SessionStore;
+  sessionName: string;
+  logPath: string;
+}>;
+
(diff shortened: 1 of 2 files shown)


thymikee:
READY at `0ea4b3b369428bb7dfaef420b4cfa52c8c968cad`. The reviewed fixture extraction preserves all 26 scenarios and 128 assertions in the test file, with no expectations hidden in helpers and no production route change. Exact-head CI is green, merge state clean, size decreases, and no device proof is owed. The title now correctly matches the target-verification-only scope.

PR #2243, opened 2026-09-02
test(remote-connection): extract shared fixtures from the connection suite

thymikee:
## Summary

Extracts the repeated remote-connection test setup into a sibling fixture module while preserving every scenario, input, override, and assertion.

The suite remains at 50 tests and 213 assertions. The extraction reduces the test file from 2,973 to 2,696 lines and removes the obsolete manual size-pin adjustment now superseded by main's history-backed ratchet.

## Validation

At `91fffd5f37b1017a7df94500d75f82df79739780`:

- `pnpm vitest run --project unit-core src/__tests__/remote-connection.test.ts` — 50/50 passed.
- `pnpm check:affected --run` — formatting, lint, typecheck, fallow, and related tests all passed.
- Fresh-worktree `pnpm install --frozen-lockfile` and `pnpm build` completed successfully.
- Rebased onto current main; the obsolete size-pin-only commit was dropped.

Touched scope: two test files; no production, runtime, package, documentation, or device-facing behavior.

PR #2252, opened 2026-09-02
test(remote-connection): extract shared fixtures from the connection suite

thymikee:
## Summary

Second pass after #2243 on `src/__tests__/remote-connection.test.ts`. The `connect --force` family, the proxy `open`/`install` pair, and the provider-allocation, persistence-failure, and proxy-disconnect scenarios now take their setup from named exports in the sibling `remote-connection.fixtures.ts`: `writeReplacedProfiles` (the old/new profile files), `seedPreviousConnection` (the `lease-old`/`run-old` connection a forced connect replaces), `forceConnectFlags`, and `recordedLeaseRelease`/`recordedLeaseAllocate` (stubs that keep the last request they received). Every test keeps its title and its own assertion lines; none is deleted, merged, or parametrized. No behavior change for operators or clients.

jscpd (`--min-tokens 80 --min-lines 8`): 211 -> 44 duplicated lines, 10 -> 3 clones. The remaining three are the shared assertion tails of the `--force` scenarios, which stay in the tests by design. Test file 2,696 -> 2,413 lines; fixtures 160 -> 282.

## Validation

- `npx vitest run --project unit-core src/__tests__/remote-connection.test.ts`: 50/50 before and after (`numTotalTests` 50 = 50); `assert.` calls 213 = 213.
- `pnpm check:quick` and `pnpm check:fallow`: clean.
- Planted red 1: in `connection-runtime.ts`, replaced `return previousConfigStillSpeaksForPreviousEndpoint(previous, profile.daemonBaseUrl)` with `return true`; `vitest -t "re-pointed config path"` -> `× connect --force does not treat a re-pointed config path's token as the previous endpoint's own` / `AssertionError: Expected values to be strictly equal`. Restored.
- Planted red 2: appended `-x` in `buildProxyDeviceKey`; `vitest -t "resolves device key before allocating lease"` -> `Expected: "ios:mobile:SIM-001"` / `Received: "ios:mobile:SIM-001-x"`. Restored.
- Full affected gate: pending (serialized gate stage will append the result)

## Tradeoffs / follow-ups

- The test-file size ratchet is merge-base-backed with no pins, so no gate edit. The cloud-TTL scenario keeps its inline allocate/heartbeat recorders: its echo shape differs from the shared one.

### DIFF
--- a/src/__tests__/remote-connection.fixtures.ts
+++ b/src/__tests__/remote-connection.fixtures.ts
@@ -1,4 +1,6 @@
+import fs from 'node:fs';
 import path from 'node:path';
+import type { CliFlags } from '@agent-device/contracts/command';
 import type { AgentDeviceClient } from '../agent-device-client.ts';
 import {
   hashRemoteConfigFile,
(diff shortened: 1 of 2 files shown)


thymikee:
READY at `2bad4fca9c685a4d179015c70afe50afffac979a`. Independent review found no concrete issues: this is test-only fixture extraction, all 50 scenarios retain their assertion ownership, and `session: undefined` preserves the no-session branch. No runtime/device evidence is owed. Exact-head CI is fully green (integration, coverage, Repo Guards, Android/Linux/macOS smoke, package/typecheck); npm unpacked size is unchanged.

Issue #1270, opened 2026-07-15
Android capture: slow `dumpsys activity top` scroll-hint call (intermittently ~5s) charged against the wait/get timeout → false 'not found' on unchanged screens


thymikee:
> **Root cause corrected (see [evidence comment](https://github.com/callstack/agent-device/issues/1270#issuecomment-4983518325)):** the node is NOT missing — it is present in the raw `uiautomator dump` 100% of the time. The false "not found" comes from an unconditional `dumpsys activity top` scroll-hint call (`deriveScrollableContentHintsIfNeeded`, snapshot.ts:600) whose latency intermittently spikes to ~5s and is charged against the same wait/get timeout. The original "capture completeness / viewport-edge" framing below is superseded.

---

## Summary

During the wave-3 E1 experiment (see #1269), a **second, distinct** failure surfaced on Android: an
element positioned at the **bottom of the viewport** intermittently falls outside the captured
accessibility tree, so a `wait`/read targeting it fails with `action-failure` even though the element is
on screen and the app is unchanged.

Filing separately from #1269 because it's a different subsystem — this is **capture completeness**, not
identity selection.

## Observation

- Android Settings → Battery screen. The "Battery" row on the Settings root sits at ~y=2864 on a 2992px
  screen (right at the viewport bottom).
- Across 20 unchanged-app replays, ~8 diverged at a `wait "label=\"Battery\""` step with
  `action-failure` (COMMAND_FAILED) — the element was not present in the captured tree on those runs,
  while present on others. Non-deterministic, position-correlated (bottom-of-viewport).

## Why it matters

A recorded flow that reads/acts on a near-edge element becomes flaky on replay through no fault of the
app — a false divergence distinct from the `android:id/title` identity class in #1269. It also shapes
fixture design (avoid edge elements) and could mask real regressions.

## Questions for the capture layer

- Is there an off-by-one / rounding at the viewport boundary in the Android hierarchy walk (helper or
  UiAutomation), or a scroll/settle race where the bottom row isn't yet laid into the captured frame?
- Should near-edge elements be scrolled-into-safe-view before capture, or the capture viewport be
  slightly over-scanned?

## Repro

Record an Android flow that `wait`s on a bottom-of-list row, replay ~20× unchanged; observe intermittent
`action-failure` on that step correlated with the element's y being near the screen bottom. Evidence
artifacts from the E1 run available (per-run divergence JSONs).

thymikee:
[...]

PR #2254, opened 2026-09-02
test(daemon): extract shared fixtures from the snapshot handler suite

thymikee:
## Summary

`snapshot-handler.test.ts` repeated its setup inline: the device literals, the token-fixed daemon request, the bind-counting runtime, the Android capture reply, the text rows, the freshness-window session, and the location-required / Battery surfaces. Those move to `snapshot-handler.fixtures.ts` (the existing `snapshot-handler-fixture.ts`, renamed to the `*.fixtures.ts` convention) as named exports. Every scenario keeps its title, its inputs, and its own `expect` lines.

- Test count 41 -> 41; `expect(` calls 162 -> 162.
- jscpd (`--min-tokens 80 --min-lines 8`): 536 -> 58 duplicated lines; the two remaining clones are assertion blocks.
- File length 2,120 -> 1,656 lines (history-backed size ratchet passes; no pin to lower).

Scope: the 1,000-line gross-diff budget fits only the largest file (this diff is 981 lines). `react-native.test.ts` (909 duplicated lines) and `session-relaunch-close.test.ts` (511) are left for follow-up PRs.

## Validation

- `npx vitest run --project unit-core src/daemon/handlers/__tests__/snapshot-handler.test.ts --reporter=json`: 41/41 before and after.
- `pnpm check:quick`, `scripts/layering/check.ts`, `pnpm check:fallow`, `test-file-size-ratchet` and `fallow-fixture-policy` gate tests: green.
- Planted red 1 (production revert): `selector-wait.ts:273` `includeHiddenContentHints: false -> true`; `wait selector polling skips hidden-content hint derivation on every poll (#1270)` fails: `AssertionError: expected true to be false`. Restored.
- Planted red 2 (fixture corruption): `makeAndroidFreshnessSession` `routeComparable: true -> false`; `snapshot warns when Android freshness retries still return the previous route` fails: `AssertionError: expected undefined to deeply equal [ StringContaining{…} ]`. Restored.

Full affected gate: pending (serialized gate stage will append the result)

## Tradeoffs / follow-ups

- No enforcement artifact changed, so there is no `chore(gates)` commit.


thymikee:
READY at `9ef08191fd55987aaba04280aa28e2a2bdd74fef`.

Test-only fixture extraction preserves the 41 snapshot-handler scenarios and their assertion ownership; production routing is unchanged. Exact-head CI is clean, including Coverage, Integration Tests, Repo Guards, and Android/Linux/macOS smoke. Size is unchanged for npm unpacked/bundled output (tarball metadata +7 B only). No linked dependencies or unresolved review findings.

PR #2258, opened 2026-09-03
feat(runtime): add managed-local ownership and the exact-only managed runtime

thymikee:
## Summary

ADR 0021 foundations for managed local runtimes. This branch now intentionally contains the two units that GitHub has already combined: #2258's owner/claim-admission foundation and merged #2259's exact-only managed-owner wrapper plus neutral allocator port. #2263 remains the focused allocator-held claim-store layer on top.

- `RuntimeOwnerRef` gains the exhaustive `managed-local` kind. The existing claim gate maps owners to one total `ordinary | allocator-held | none` rule, and managed admission returns an explicit admitted/refused result; there is no parallel command path.
- `managedBindingFence` is the canonical requester/incarnation fence. Its decoder accepts only the canonical encoding and preserves allocator-issued identifiers verbatim.
- The existing composed runtime gateway owns one exact-only `managedOwners` registry. Ordinary selection cannot reach it. The wrapper delegates device mechanics to the existing local-family owner, republishes the binding under the managed owner, and removes lifecycle, implicit-boot, and non-adoptable durable-capture cells through the existing capability/narrowing machinery.
- `@agent-device/contracts/managed-device-allocation` is a types-only allocator port with a scripted test fake. It adds no allocator dependency, socket code, or production call site.
- The wrapper is loaded lazily from the gateway, preserving the eager import closure.

Nothing in production registers a managed owner or calls the allocator yet, so this foundation does not make managed sessions reachable and changes no device-facing behavior. Pre-binding readiness and durable-capture adoption remain explicitly deferred to the allocation-flow/lifecycle units, where they can be exercised end to end.

Combining these units is deliberate now that #2259 was merged into this branch: splitting them again would discard an already-merged unit and create a second implementation path. They share the owner/gateway contract and remain one coherent runtime foundation. Claim persistence and clearing stay separate in #2263 because the claim store is an independent owner with different failure and recovery rules.

Scope: 33 files, +1,943/-180 against `main`. The npm unpacked increase is 5.6 kB: approximately 2.1 kB is the managed-owner wrapper, 1.1 kB is the batch-runner projection, and the apparent 56.3 kB `device-claim-rule.js` addition is predominantly the renamed 54.9 kB `device-claim-conflict.js` chunk. The remaining growth is contracts/projections; startup stayed flat within measurement noise.

## Validation

Exact head: `83d5522d3fd4f16afb5de4cb208ef3f8c190476f`.

[...]

PR #2259, opened 2026-09-03
feat(runtime): register the managed local owner as an exact-only wrapper and add the neutral allocator port

thymikee:
## Summary

ADR 0021 foundations, unit 3. Stacked on #2258 (`adr0021/u1-owner-kind`); U4 stacks on this.

Unit 1 added the `managed-local` owner kind and left the gateway's exact-owner arm for it failing
closed. This unit gives that arm a registry, the owner it selects, and agent-device's own allocator
port. **Nothing in production registers a managed owner or calls an allocator**, so every arm here
is reached from tests only; the point of the unit is that the arms exist and fail closed.

What changes for command authors and platform behavior:

- A new gateway option, `managedOwners`, is the only place a managed local owner can be registered,
  and only the `managed-local` arm of `selectExactOwner` reads it. Ordinary selection cannot reach
  a managed owner by construction rather than by a check.
- A managed binding is a local family binding republished under the managed owner, with
  twenty-seven cells withheld as `owner-capability-missing` and the operations filtered by those
  facts (R66). A command author sees the same `narrowDeviceBinding` refusal any unsupported cell
  produces.
- The allocator port is a types-only contracts subpath with a scripted fake. No socket code, no
  allocator package dependency, no production call site.

12 files touched, no scope growth beyond the runtime-gateway/contracts surface the unit names;
`src/platform-runtime.ts` is deliberately not edited (no production producer of managed owners
exists yet, so `createPlatformRuntimeGateway` gains no pass-through).

## The exact-only registry

`createComposedPlatformRuntimeGateway` gains a `managedOwners` list that only the `managed-local`
arm of `selectExactOwner` reads. `selectOrdinaryProvider`, `inspectFacts` and the ordinary `bind`
arm never see it, and `providerModules` pairs one `ProviderDeviceRuntime` with one
`provider-runtime` owner — so ordinary selection cannot be handed a managed owner. A
`@ts-expect-error` pins that: a managed ref is not assignable to a provider module's owner. A
duplicate instance is refused at composition, an unregistered instance still answers
`owner-unavailable`, and a registered one is composed once.

## The wrapper

`src/platform-runtime-managed-owner.ts` (root zone, no platform imports) binds only under an
exact-owner intent naming itself, loads the device's own family owner through the gateway's loader,
delegates with `{ kind: 'ordinary' }` — a family owner refuses a foreign exact owner — and
republishes the binding under the managed owner. It does not decode the fence: what a managed
binding fence proves is the device-claim gate's business, and a second reader would be a second
answer. `ownsDevice` returns false.

Twenty-seven cells are withheld, enumerated by mechanics rather than by catalog group:

- device lifecycle: `ensureReady`, `bootTarget`, `bootTargetHeadless`, `shutdownTarget`, plus
  `deployApp` and `deployMaterializedApp` — both family deployment runtimes ensure readiness before
  installing, and `deployAppUse` requires `deployApp` alone, so nothing else would have refused it;
- application cells that boot or shut the device down: `prepareApplicationOpen`,
  `prepareAppleRunner`, `closeApplication`, `finalizeApplicationClose`;
- the twelve durable-capture cells (app log, screen recording, audio probe, native perf capture),
  which a managed binding could never reattach because the family runtime stamps envelopes with its
  own local owner;
- `captureScreenshot`, `setSetting`, `readClipboard`, `writeClipboard`, `openApplication` — none is
  declared device lifecycle, but each Apple family-runtime implementation can boot the simulator
  lazily below cell-selection granularity (screenshot retries through a boot on a shutdown failure;
  the other three resolve a local interactor the same lazy way), which is exactly the implicit-boot
  fallback ADR-0021 §3 rules out for a managed device.

**A managed session cannot be opened in the foundations.** A managed binding loses every command
whose runtime use requires a withheld cell: `install` (and push-notification delivery), `apps` and
`app-state`, `device boot` and `device shutdown`, `open` and `close`, and the durable capture starts
(`logs start`, `record start`, audio probe capture, native perf capture). That is the intended shape
of the foundations — the allocator owns the device's lifecycle — not a gap to work around in daemon
code.

## Named out of scope

- **Withholding cells is not a complete lifecycle exclusion, and the code says so.** The four
  lazily-booting Apple cells named above are now withheld rather than left as a caveat — an earlier
  revision of this PR retained them and disclosed the gap instead of closing it; review caught that
  the disclosure preserved the exact bypass ADR-0021 §3 rules out, so the fix withholds them.
- **Pre-binding readiness is the same class, at the daemon level, and stays open.**
  `session-device-resolution` → `device-ready` → `platform-runtime-device-ready` boots a device
  through direct simctl/adb before any binding exists, gated only by provider ownership — below
  where a managed binding's cell withholding can reach at all. The exclusion above covers binding
  cells only, and the exclusion test says so.
- Closing the daemon-level gap is its own follow-up: pre-binding readiness moves under the binding
  once a managed owner is actually reachable from production (no unit before the allocation-flow
  unit can exercise it end to end).
- **Durable captures stay unavailable** until family runtimes stamp the outward owner on their
  envelopes (adoption compares the envelope's owner key with the binding's). Same family-runtime
  change; not attempted here.
- **Delegating with an ordinary intent** means local exact-owner recovery semantics never apply
  under a managed binding — consistent with durable captures being unavailable. The delegation test
  pins the choice, so forwarding a local-family exact intent later is a deliberate change.

## The allocator port

`@agent-device/contracts/managed-device-allocation` is agent-device's own interface to a
managed-device allocator: lease request, lookup, supersession, cancellation, renewal, release,
activation confirmation, identity status, and removal acknowledgement. Types only, with no
dependency on any allocator package and no socket code. The member and field names match the
allocator's published contract (`requestGeneration`, `identityIncarnationId`,
`expectedRequestGeneration`, `confirmLeaseActivation`, `getManagedIdentityStatus`,
`acknowledgeManagedIdentityRemoval`, `admission: 'fail-fast'`,
`activation: 'direct' | 'external-fence'`, `LEASE_ENVIRONMENT_INVALID { platform, key }`, and the
`{ platform: 'ios', deviceSetPath } | { platform: 'android', adbServerPort }` projection) so the two
sides cannot drift.

Shaping notes:

- Failures that would otherwise be invented error codes — a busy requester lane, an unknown attempt
  — are modelled where the contract already puts them: a refusal reason and a request state on
  `LeaseRequestStatus`. Protocol-version skew belongs to the unit that opens the socket.
- `ManagedLeaseEnvironment` and `LeaseEnvironmentError` are declared but have no reader yet. The
  reader, and the mapping from agent-device's own device families onto the port's
  `'ios' | 'android'` vocabulary, land with the allocation-flow unit that first turns a grant into
  a `DeviceInfo`. Both carry the repo's one-line `fallow-ignore-next-line unused-type` idiom rather
  than a baseline entry.
- The port's only implementation is a scripted fake under `src/__tests__/test-utils/*.fixtures.ts`.

## Pins moved

- Three-place pin for the new contracts subpath: `packages/contracts/package.json#exports`,
  `CONTRACT_EXPORTS`, and a `FACADE_BUDGETS` row of 1 (types only, so the entry evaluates itself).
- `HUB_BUDGETS['src/platform-runtime.ts']` 47 → 48, measured: the composed gateway now composes the
  wrapper, whose own value imports (the owner contract, kernel errors) were already in that closure.
- CONTEXT.md: **Managed device allocator port**.

## Validation

Evidence gathered at `e2978a6637` (rebased onto #2258's blocking fixes; includes the withhold-cells
review fix below). **Published and reported** — CI on the head is the authority.

`pnpm check:affected --run` fail-opens to the full 57-check set (the diff touches tooling files), and
the whole set passed at this commit: format, lint, typecheck, layering, di-seams,
`check:fallow --base origin/main` (clean over 34 changed files), build, package, integration-node,
macOS command coverage, vitest-related, unit + smoke, provider-integration, integration-progress,
replay-compat, daemon-wire-compat, production-exports, bundle-owner-files, command-docs,
agent-guidance, xctest-selection, maestro-conformance, and the gate/mutation/coverage model suites.
The GitHub-authoritative lanes (device replay, Swift/Android helper builds, coverage, mutation, fuzz)
are skipped locally by design.

Two earlier full runs on this branch each aborted at `vitest-related` on a *different* test timing
out at the 5 s vitest limit under parallel load —
`test/integration/provider-scenarios/ios-lifecycle.test.ts` (4.9 s in isolation) and
`src/daemon/__tests__/gesture-admission-parity.test.ts` (3.3 s in isolation, already flagged by the
slow-test gate at a 2.5 s budget). Both pass in isolation, both are timeouts rather than assertion
failures, and neither is on a path this change touches. `check:affected` aborts at the first failed
lane, so those two runs did not reach the lanes after `vitest-related`; the clean third run did, and
CI on the head is what settles it.

Regression tests, each with the revert that makes it fail (all executed as mutations against the
real code; every one went red):

- the withheld set — dropping `deployApp`/`deployMaterializedApp`, `prepareAppleRunner`, the
  durable-capture cells, or the five lazily-booting cells (`captureScreenshot`, `setSetting`,
  `readClipboard`, `writeClipboard`, `openApplication`) from the wrapper's list; returning
  `binding.operations` unfiltered;
- `narrowDeviceBinding` refusals for `appsRuntimeUse`, `appStateUse`, `bootTargetUse`,
  `shutdownTargetUse` and `deployAppUse`;
- the wrapper — forwarding the managed exact intent to the family owner (the family fixture throws
  on a foreign exact owner), dropping the owner rewrite (`bindAndValidate` then rejects), accepting
  an ordinary or foreign intent, or not forwarding disposal;
- the gateway — restoring the fail-closed `managed-local` arm, dropping the composition-time
  duplicate check, or composing a fresh wrapper per bind (the second registration is then refused);
- `providerModeMatchesOwner`'s `managed-local` arm accepting `transport-composed` (fixed on #2258
  and rebased in here): `accepts a managed binding whose local facts report a transport-composed
  device` binds through the gateway with a `transport-composed` local family device and asserts
  success — replacing an earlier version of this test that asserted the opposite (rejection), which
  is what review caught: the arm rejected the mode the PR's own known-gap note said it should accept;
- the scripted fake — consuming steps from one shared queue, or recording no calls.

No device-facing behavior changes, so no simulator/emulator evidence is owed: nothing in production
can select a managed owner yet.

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -305,6 +305,11 @@ Compatibility-engine evidence captured since the most recent mutation; mutation 
 **Provider**:
 An external adapter that owns a device runtime or contributes transport to a platform module.
 
+**Managed device allocator port**:
+The daemon-owned interface to a managed-device allocator: lease request, lookup, renewal, release,
+identity status, activation confirmation, and removal acknowledgement.
+_Avoid_: Simlock client, lease provider
+
 **Cloud WebDriver runtime**:
 A provider runtime that maps a cloud-owned Appium or WebDriver session into agent-device inventory,
 leases, runtime behavior, artifacts, and release.
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -306,8 +306,8 @@ Compatibility-engine evidence captured since the most recent mutation; mutation 
 An external adapter that owns a device runtime or contributes transport to a platform module.
 
 **Managed device allocator port**:
-The daemon-owned interface to a managed-device allocator: lease request, lookup, renewal, release,
-identity status, activation confirmation, and removal acknowledgement.
+The daemon-owned interface to a managed-device allocator, covering everything the daemon needs to
+obtain, hold, and give back a managed device.
 _Avoid_: Simlock client, lease provider
 
 **Cloud WebDriver runtime**:
--- a/packages/contracts/package.json
+++ b/packages/contracts/package.json
@@ -275,6 +275,10 @@
       "types": "./src/logs-runtime-plan.ts",
       "default": "./src/logs-runtime-plan.ts"
     },
+    "./managed-device-allocation": {
+      "types": "./src/managed-device-allocation.ts",
+      "default": "./src/managed-device-allocation.ts"
+    },
     "./managed-web-backend": {
       "types": "./src/managed-web-backend.ts",
       "default": "./src/managed-web-backend.ts"
--- a/packages/contracts/src/managed-device-allocation.ts
+++ b/packages/contracts/src/managed-device-allocation.ts
@@ -1,0 +1,160 @@
+import type { JsonObject } from './json.ts';
+
+/**
+ * agent-device's own interface to a managed-device allocator (ADR 0021 §3): lease request,
+ * lookup, supersession, cancellation, renewal, release, identity status, activation confirmation,
+ * and removal acknowledgement. The vocabulary matches the allocator's published contract so the
+ * two sides cannot drift, but the interface is owned here — no allocator package is a dependency,
+ * and the foundations implement it only as a scripted fake.
+ */
+
+/** The allocator's platform vocabulary; the daemon maps its own device families onto it. */
+export type ManagedLeasePlatform = 'ios' | 'android';
+
+/** A managed shape (ADR 0021 §5). The allocator owns canonical resolution against its catalog. */
+export type ManagedShapeRequest = Readonly<{
+  platform: ManagedLeasePlatform;
+  deviceType: string;
+  osVersion?: string;
+}>;
+
+/**
+ * The per-platform key a grant must carry for its device to be reachable. Each is a versioned
+ * driver contract on the allocator side, so the names are pinned as a type rather than restated
+ * as free strings wherever a lease is read.
+ */
+export type ManagedLeaseEnvironmentKey = 'SIMLOCK_IOS_DEVICE_SET' | 'ANDROID_ADB_SERVER_PORT';
+
+/** The typed reach projection of a lease's otherwise opaque environment record. */
+export type ManagedLeaseEnvironment =
+  | Readonly<{ platform: 'ios'; deviceSetPath: string }>
+  | Readonly<{ platform: 'android'; adbServerPort: number }>;
+
+/** A grant whose platform scoping key is absent or unparsable is unusable, not merely degraded. */
+export type LeaseEnvironmentError = Readonly<{
+  code: 'LEASE_ENVIRONMENT_INVALID';
+  platform: ManagedLeasePlatform;
+  key: ManagedLeaseEnvironmentKey;
+}>;
+
+export type ManagedLeaseEnvironmentOutcome =
+  | Readonly<{ ok: true; environment: ManagedLeaseEnvironment }>
+  | Readonly<{ ok: false; error: LeaseEnvironmentError }>;
+
+/** One granted managed-device lease. `environment` stays the allocator's raw record. */
+export type ManagedLease = Readonly<{
+  id: string;
+  ttlDeadline: number;
+  device: Readonly<{ address: string }>;
+  environment: Readonly<Record<string, string>>;
+}>;
+
+/**
+ * One created managed identity. The incarnation id is minted before the device exists and is
+ * preserved across reuse of the same Android identity, so it names the identity's creation rather
+ * than the request that happens to be holding it.
+ */
+export type ManagedIdentityRef = Readonly<{ deviceId: string; identityIncarnationId: string }>;
+
+export type ManagedIdentityState =
+  | 'reserved'
+  | 'active'
+  | 'idle'
+  | 'removing'
+  | 'removed'
+  | 'unknown';
+
+export type ManagedIdentityStatus = ManagedIdentityRef & Readonly<{ state: ManagedIdentityState }>;
+
+/**
+ * Fail-fast admission (ADR 0021 §4): the allocator refuses instead of queueing and agent-device
+ * adds no second queue. `activation: 'external-fence'` is the reusable-identity handshake, where
+ * the allocator validates its clean baseline under the requester's fence before publishing the
+ * grant; `'direct'` publishes as soon as the identity is ready.
+ */
+export type LeaseRequestInput = Readonly<{
+  requesterId: string;
+  requestGeneration: number;
+  attemptKey: string;
+  shape: ManagedShapeRequest;
+  deadlineAtMs: number;
+  admission: 'fail-fast';
+  activation: 'direct' | 'external-fence';
+  attribution?: JsonObject;
+  /** Abandons only this caller's wait; the attempt keeps its own terminal outcome. */
+  signal?: AbortSignal;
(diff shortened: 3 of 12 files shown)


thymikee:
BLOCKED at `3d29f60eaaa48406e570c231d9aebf29c86071a6`.

1. The managed-owner wrapper retains `captureScreenshot`, `setSetting`, clipboard, and `openApplication`, whose Apple family-runtime cells can lazily boot the Simulator. That violates ADR-0021 §3's hard boundary: managed lifecycle/readiness belongs to Simlock and no handler path may fall back to direct lifecycle tooling. The current caveat/tests preserve rather than block the bypass; remove/own the implicit readiness path or withhold every such cell and add a managed-binding regression.

2. The PR is `CONFLICTING`/`DIRTY` against `adr0021/u1-owner-kind`; iOS Smoke is still pending.

The exact-only registry behavior and types are otherwise coherent.

thymikee:
Fixed at e2978a6637 (rebased onto #2258's blocking fix, then this fix on top).

**1. Lazy-boot cells.** Withheld `captureScreenshot`, `setSetting`, `readClipboard`, `writeClipboard` and `openApplication` from the managed binding instead of disclosing them as a retained caveat — they're now in `WITHHELD_MANAGED_OPERATIONS`'s fourth group, twenty-seven cells total (was twenty-two). The pre-binding readiness gap you didn't flag (`session-device-resolution` → `device-ready` → `platform-runtime-device-ready` booting below where any binding-level withholding can reach) stays open and out of scope, as before — it's a daemon-level change, not a cell-selection one.

**2. Rebase + `providerModeMatchesOwner`.** Rebased onto #2258's fix (`local || transport-composed` for the `managed-local` arm). I flipped this PR's own `rejects a managed binding whose local facts report a transport-composed device` test, which was pinning the wrong outcome — it asserted rejection, while this PR's own body already said the arm should accept `transport-composed`. It's now `accepts a managed binding whose local facts report a transport-composed device` and binds through the real gateway (not just the wrapper in isolation) with a `transport-composed` local family device, asserting success.

Planted-red verified both: reverting the withheld-cells addition fails `withholds every lifecycle-bearing and durable capture cell and keeps the rest`; reverting the `providerModeMatchesOwner` arm reproduces the exact `runtime-contract-invalid` rejection the flipped test now catches.

`pnpm check:affected --run` green at e2978a6637 (fail-opens to the full 57-check set, diff touches tooling files): format/lint/typecheck/layering/di-seams/fallow/build/package/integration-node/macOS coverage/vitest-related/integration-progress/replay-compat/daemon-wire-compat/gate-manifest/depgraph/tmpdir-leaks/coverage-model/production-exports/bundle-owner-files/fixture-cache/fixture-fallback/command-docs/agent-guidance/xctest-selection/maestro-conformance/mutation-model all clean.

thymikee:
Code review CLEAN at this exact head. The lifecycle bypass is fixed across all lazy Apple cells, operation facts are filtered correctly, and the gateway regression now proves the transport-composed bind inherited from #2258. Exact-only registration, disposal, duplicate refusal, and the allocator port are coherent. Merge after #2258; two exact-head Smoke lanes are still running.

PR #2260, opened 2026-09-03
test(android-e2e): make an alert-dismiss failure self-explaining

thymikee:
## Summary

Android's post-alert canary now reads `automation-alert-result` directly (`get text
id="automation-alert-result"`) instead of a screen-wide `wait text`, whose surface dump (capped
at 6 labels) often missed the element under contention, masking real failures behind a truncated
label list. `alert accept`/`dismiss` also record the tapped button's coordinates alongside its
label in the result. The live-device e2e harness now captures a best-effort
`wait-timeout-<n>.png` screenshot whenever a `wait` step times out, referenced from
`failed-step.txt`; a failed capture never masks the original failure.

Not changed: the daemon's wait-surface label cap — out of scope; this routes around it.

## Validation (at 927f068f9b)

- `npx vitest run --project unit-core packages/platform-android/src/__tests__/alert.test.ts
  alert-detection.test.ts` — 7/7 passing; both suites now share fixtures from the new
  `alert-fixtures.ts`.
- `pnpm check:quick`, `pnpm check:layering` (198/198), `pnpm check:fallow --base origin/main` —
  all clean.
- Planted red: reverted `coordinates` in `alert.ts` — dismiss/accept cases failed as expected,
  3/3 green after re-applying. Also corrupted the shared `button()` fixture's rect width — 2/7
  alert tests failed, confirming both suites are wired to the extracted fixtures; reverted, 7/7
  green.
- `live-automation-scenario.ts` / `runtime.ts` are e2e-only, exercised by this PR's Android lane.
- Full affected gate: pending.

## Tradeoffs / follow-ups

`assertElementText` is a single non-retrying `get text`, unlike the polling `wait text` it
replaces; `alert accept`/`dismiss` don't wait for JS-side state to settle, so a contention-slow
run could race the read. The live Android lane here is the evidence gate; if it flakes, switch to
a selector-scoped wait instead of reverting to the broad one.

### DIFF
--- a/packages/platform-android/src/__tests__/alert-detection.test.ts
+++ b/packages/platform-android/src/__tests__/alert-detection.test.ts
@@ -1,7 +1,7 @@
 import { test } from 'vitest';
 import assert from 'node:assert/strict';
 import { chooseAndroidAlertButton, findAndroidAlertCandidate } from '../alert-detection.ts';
-import type { RawSnapshotNode } from '@agent-device/kernel/snapshot';
+import { button, node, text } from './alert-fixtures.ts';
 
 test('chooseAndroidAlertButton prefers platform ids over ambiguous labels', () => {
   const candidate = findAndroidAlertCandidate([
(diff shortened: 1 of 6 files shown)


thymikee:
READY at `927f068f9b2cf3ac4793d2b75846994d5a9d070d`.

Clean exact-head CI, including Android Smoke. The alert assertion is selector-scoped, coordinate diagnostics are covered, and screenshot capture remains best-effort while preserving the original failure. No unresolved findings.

PR #2261, opened 2026-09-03
ci(android): wait for the producer's platform job and keep the fallback build alive

thymikee:
## Summary

The Android fixture lookup classified producer readiness by the whole workflow run's aggregate status, which flickers to `queued` between the fingerprint job finishing and the `Android Release` job starting — the consumer treated `queued` as "give up" and bailed into an inline Gradle build almost instantly (run 33550746596).

- `trusted-artifact.mjs`'s `producer-state` now classifies by the producer's own platform job (`Android Release` / `iOS Release`, via `actions/runs/{id}/jobs`), not the run's aggregate status, and takes an explicit `platform` argument.
- `fetch-artifact.sh` treats `queued` like `in_progress` (keep polling) instead of bailing.
- `android.yml`'s `wait-for-artifact-seconds` rises 600s → 1800s, matching `ios.yml`.
- The producer's Android build and the inline fallback both set `org.gradle.jvmargs=-Xmx4g`: the producer OOM'd in `:app:compileReleaseArtProfile` (run 33728318738), prior fallbacks OOM'd in `:app:mergeDexRelease`.

Trust rule (exact head SHA or default-branch push) unchanged.

## Validation

- `trusted-fixture-artifact` node:test file and `pnpm test:fixture-fallback` — pass.
- `pnpm check:quick`, `pnpm check:fallow`, `pnpm check:gate-manifest`, `pnpm format:check` — clean.
- Planted red: reverting classification to run-status gave `AssertionError: expected 'queued' to equal 'in_progress'`; reverting the `queued`/`in_progress` merge gave `FAIL: ios queued producer state bailed instead of waiting out its budget.` Both restored green.
- Wait-budget/Gradle-heap YAML edits: no planted red — configuration, not gated logic.
- Full affected gate: pending (serialized gate stage will append the result).

## Tradeoffs / follow-ups

Left the shared action's `wait-for-artifact-seconds` default at `0`: `replays-manual.yml`'s `manual-android` job relies on it with `require-artifact: true` under a 15-minute timeout; a global bump would turn a fast failure into a confusing timeout.

### DIFF
--- a/.github/actions/setup-fixture-app/action.yml
+++ b/.github/actions/setup-fixture-app/action.yml
@@ -172,6 +172,7 @@ runs:
       run: |
         set -euo pipefail
         pnpm --dir examples/test-app exec expo prebuild --platform android --no-install
+        echo "org.gradle.jvmargs=-Xmx4g" >> examples/test-app/android/gradle.properties
         (
           cd examples/test-app/android
           ./gradlew :app:assembleRelease
--- a/.github/actions/setup-fixture-app/fetch-artifact.sh
+++ b/.github/actions/setup-fixture-app/fetch-artifact.sh
@@ -33,7 +33,7 @@ query_artifact() {
   node "$TRUSTED_ARTIFACT" find "$REPOSITORY" "$NAME" "$EXPECTED_HEAD_SHA" 2>/dev/null
 }
 query_producer_state() {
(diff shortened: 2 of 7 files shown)


thymikee:
READY at `e8dc57b22348f979b0d06971a322d7416eb7ff52`.

Clean exact-head CI. Producer state is keyed to the exact run plus platform job; the queued path exhausts its bounded wait then falls back, unknown platforms fail closed, and targeted regressions cover the change. The 30-minute Android wait is an accepted operational tradeoff, not a blocker.

PR #2263, opened 2026-09-03
feat(daemon): add the allocator-held device claim kind

thymikee:
## Summary

ADR 0021 allocator-held execution claims, stacked on #2258's managed-local runtime foundation. A managed-pool device is held for the identity's pool lifetime rather than for one daemon process or session.

- The claim store gains a separate schema-v3 `AllocatorHeldDeviceClaim`; the existing schema-v2 process claim is unchanged. The allocator record has an installation/allocator/incarnation principal and cannot carry process, session, workspace, or abandonment fields.
- Record decoding and atomic publication are centralized in `device-claim-record.ts` and `device-claim-store.ts`. Both claim kinds reuse the existing host-kit `publishFileSync` and per-device process lock; there is no second persistence mechanism.
- Managed admission is covered only when the exact owner, installation, and identity incarnation match the canonical managed fence. Missing, foreign, or stale-incarnation claims fail closed.
- Ordinary admission performs the smallest read-only allocator classification for every device-touching policy. It deliberately avoids process-liveness observation on the hot binding path.
- Malformed allocator-looking v3 records classify as `allocator-inconsistent` and fail closed instead of being treated as an absent claim. The ordinary `observe` regression was planted red by reverting that classification gate.
- Existing process-owned clearing paths cannot clear allocator claims by type. The only clearing API requires the allocator's exact removal proof; status renders the allocator principal and `device release --stale` refuses without changing the file.

Nothing writes allocator claims in production yet. The acquire/release functions are the focused store interface for the later activation/removal units; they do not add lifecycle policy, command flags, or a second allocator abstraction.

Scope: 24 files, +1,695/-343 against #2258. Most growth is the split record/store implementation and one-to-one claim-family tests. The npm unpacked increase is 4.0 kB; startup moved by +0.1 ms for `--version` and +2.2 ms for `--help`, within the seven-run measurement noise. The design removes the old decoder/type cycle and reuses the shared atomic publisher instead of adding local file-publication code.

## Validation

Exact head: `db60db6761d4936f0f7f0a226fc6d18776501643`.

- `pnpm check:affected --run`: green at the feature/docs series across format, lint, typecheck, layering, fallow, build, and related tests; the final malformed-record fix is covered by exact-head GitHub CI.
- GitHub: every required lane is green, including iOS, Android, macOS, Linux, Coverage, Integration, Repo Guards, CodeQL, Agent Guidance, and Size. The iOS workflow passed simulator E2E, gesture replay, and host-focus preservation; its physical-device step was the expected skip.
- Layering: 198 checks, with no cycle or ownership-baseline relaxation. Eager closure: 407/407.
- Planted-red coverage includes allocator/process record separation, full-principal matching, exact-incarnation admission, ordinary-owner refusal, malformed-v3 fail-closed behavior, stale-release refusal, inventory/status projection, shared atomic-publish ownership, and the total admission table.
- No device evidence is claimed: no production caller creates an allocator claim or managed binding in this unit.


thymikee:
BLOCKED at `690041cc9d1f4a5aaedec8ac3040d9952777982b`.

It is `CONFLICTING`/`DIRTY` against stacked base `adr0021/u1-owner-kind` (`c48403b…`).

Coverage is also an owner-action failure: new `src/daemon/device-claim-store.ts` publishes a same-directory file without the shared `publishFileSync` atomic-publish owner required by `atomic-publish-ownership.test.ts`. Rebase and route publication through the shared atomic publisher, then rerun exact-head gates.

Current completed checks are green; iOS is in progress. Size is +3.7 kB unpacked and needs growth justification after the functional gate is fixed. Do not label ready.

thymikee:
Rebased onto #2258's blocking fix and fixed at 0cbc7515f3.

**Rebase.** device-claims.ts no longer writes its claim file directly — it delegates to `writeDeviceClaim` in the new `device-claim-store.ts`, so `atomic-publish-ownership.test.ts`'s `SIMPLE_PUBLISHERS` source-grep for `publishFileSync` no longer matched anything in `device-claims.ts` and Coverage failed. Registered `device-claim-store.ts` (the single writer, per its own docstring) as the owner in that list instead, with `device-claims.ts` dropped from it. Planted-red verified: reverting the list entry reproduces the exact `atomic-publish-ownership.test.ts:20` failure from the Coverage run.

**Same-switch conflict from #2258's rebase.** `decideAllocatorHeldAdmission` picked up this PR's `covered`/`incarnation-stale` arms on top of #2258's totalized `{admitted}|{admitted:false,error}` shape. Also caught and fixed: `ADMITTED_BY_OUTCOME_STATUS` in `device-claim-conflict.test.ts` — the exhaustiveness table `decideAllocatorHeldAdmission` is built to force a compile error against — still only listed the three original statuses and TS2739'd once rebased against the widened union. Extended it (`covered: true`, `incarnation-stale: false`) and the `outcomes` array alongside it; planted-red verified by reverting the two new entries and re-running `tsc`.

Also carried forward the earlier should-fix rename (`DeviceClaimRecord` → `StoredDeviceClaim`, `decodeDeviceClaimRecord` → `decodeStoredDeviceClaim`) that was sitting uncommitted in the worktree.

`pnpm check:affected --run` green at 0cbc7515f3: 534 files / 3908 tests, format/lint/typecheck/layering/fallow/agent-guidance all clean. CONTEXT.md at 11,808 bytes (12,000 cap).

thymikee:
BLOCKED. `readAllocatorHeldClaimFile` returns `null` whenever a record does not decode to a valid allocator claim. An allocator-looking v3 record made inconsistent by corruption (for example a forbidden process-principal field) is therefore treated as absent by ordinary-owner admission, allowing `observe` and other device-touching policies to proceed. That violates ADR-0021’s fail-closed corrupted-state requirement. Refuse admission when the existing claim cannot be proven safe/non-allocator, and add a planted-red ordinary-owner `observe` regression with a malformed v3 allocator record. The branch also currently conflicts with its advanced base and must be rebased; re-check the >3 kB size result afterward.

thymikee:
Re-reviewed at `ff74cddf1ad6adba1b5a2abd4bea1e413f34305a`: still BLOCKED. This update resolves the stack conflict but is effectively a rebase and does not address the malformed allocator-record finding. `readAllocatorHeldClaimFile` still maps an inconsistent allocator-looking record to `null`, ordinary-owner admission still treats that as safe, and no malformed-v3 `observe` regression/planted-red evidence was added. Fail closed at the ordinary admission seam, add that regression, then complete exact-head CI (Repo Guards, Coverage, and Smoke lanes are currently running).

thymikee:
The malformed allocator-record code blocker is fixed at `db60db6761d4936f0f7f0a226fc6d18776501643`: inconsistent v3 allocator records now reach fail-closed ordinary admission, and the real `observe` regression has valid planted-red behavior. Overall still BLOCKED by #2258’s scope/stack mismatch, stale PR-body file/validation accounting, and pending Repo Guards, Coverage, Android/iOS/macOS Smoke. Update the body/evidence after the base is resolved and wait for exact-head CI.

thymikee:
The remaining review blockers are resolved at exact head `db60db6761d4936f0f7f0a226fc6d18776501643`.

The body now matches the actual 24-file/+1,695/-343 diff and the combined #2258 base. It records the malformed allocator-looking v3 fail-closed behavior and planted-red `observe` regression, shared `publishFileSync` ownership, the read-only hot-path classifier, the +4.0 kB size evidence, and the absence of production/device reachability.

Code review is clean. Every completed exact-head lane is green; iOS Smoke is the only remaining in-progress gate, so readiness waits on that result.

thymikee:
READY at exact head `db60db6761d4936f0f7f0a226fc6d18776501643`.

The final iOS Smoke lane passed, including simulator E2E, gesture replay, and host-focus preservation; the physical-device step was the expected workflow skip. Every exact-head required check is now green, the stack base is current, the PR contract matches the 24-file diff, and the malformed-v3 fail-closed review finding remains pinned by its planted-red regression.

PR #2182, opened 2026-08-31
fix: address security scanner findings

thymikee:
## Summary

Address the repository's current security findings:

- replace polynomial-ReDoS regexes in Android lifecycle, alert, and observation parsing with linear string operations;
- pin patched `js-yaml` and `nanoid` transitives in the isolated Expo test app;
- apply a local `image-size` 1.2.1 patch at the shared box parser so zero-sized or undersized ICNS, HEIF, and JXL records terminate safely;
- keep native fixture fingerprint resolution machine-readable by invoking the installed fingerprint binary directly, outside pnpm's dependency-status output.

The `image-size` Dependabot alerts remain open because upstream has no patched release; the pnpm patch is the repository remediation until one exists.

Scope expanded from Android and security paths into the existing fixture-cache resolver only to repair the exact CI failure exposed by the dependency patch.

## Validation

- Observed red: zero-size HEIF/JXL fixtures timed out with the shared `image-size` guard bypassed; green on the patched head.
- Observed red: the fixture resolver regression rejected pnpm mediation with exit 91; green after direct installed-binary ownership.
- `node --test security/image-size-security.test.mjs` from `examples/test-app`: 1/1 passed.
- Focused Android parser suites: 9/9 passed.
- `node --test test/ci/trusted-fixture-artifact.test.mjs`: 11/11 passed.
- `CI=true sh .github/actions/setup-fixture-app/resolve-artifact-name.sh ios`: emitted a valid iOS-scoped fingerprint name.
- `CI=true sh .github/actions/setup-fixture-app/resolve-artifact-name.sh android`: emitted a valid Android-scoped fingerprint name.
- `pnpm check:affected --run`: format, lint, typecheck, layering, DI seams, fallow, metadata, build, package verification, integration-node, and macOS coverage passed; the related-test phase was host-contention limited, with 36 unrelated fixed-budget timeouts across interaction, provider, fuzz, daemon, and help-conformance suites.
- Exact-head GitHub CI is authoritative for native, device, provider, and coverage lanes.

PR #2269, opened 2026-09-03
ci: enforce the image-size parser mitigation through a test-app gate

thymikee:
## Summary

`image-size` (a transitive of the test app's Metro) has two open Dependabot advisories — GHSA-w3rx-r6r6-pgpr (ICNS zero-length entry) and GHSA-5p2g-fcmc-qvqq (JXL/HEIF zero-size box), both DoS via infinite parser loops — and **no fixed version exists** (`latest` is still `2.0.2`, inside the vulnerable range). The mitigation already landed in #2182: an in-tree pnpm patch (`examples/test-app/patches/image-size@1.2.1.patch`) plus a budget test (`examples/test-app/security/image-size-security.test.mjs`, run by the `security:test` script).

The gaps this closes: that proof was a manual script — nothing in CI ran it, and the dependency-graph cache the test depends on could restore a **stale patched artifact** for a patch-only change, so even a wired-up gate would have tested the previous patch.

This registers the security suite as a first-class gate, `test-app-security`, and makes the enforcement sound end to end:

- Root script `test-app:security` → `pnpm --dir examples/test-app security:test`.
- Catalog entry in `CHECK_CATALOG` (CI-authoritative, `localRunnable: false`, mirroring `test-app-typecheck` — the test app owns an isolated Expo dependency graph and must not be installed by every root-checkout validation).
- Wired into the `Test App Build Cache` `fingerprint` job (runs on every PR and already restores the test app dependency graph, so the patched package is present).
- **Cache key completeness** (review finding): `setup-test-app-dependencies` now hashes `examples/test-app/patches/**` into its restore/save key alongside the manifests. Before, a patch-only edit left the key invariant, so a cache hit restored the previous patched `node_modules` and skipped `pnpm install` — the gate would run against the prior patch. Any patch change now forces a fresh install that applies the current patch.
- **Artifact resolution**: the test resolves the installed package by the `patchedDependencies` hash recorded in `pnpm-lock.yaml` instead of the first `image-size@1.2.1_patch_hash=*` directory it finds — the declared artifact is the one under test, and a missing declaration fails loudly.
- **Discriminating fixture**: added a JXL container whose zero-size box is named `jxlp` (the box `extractPartialStreams` searches for). Ground truth: on unpatched 1.2.1 the prior JXL fixture (zero-size box named `junk`) terminates via `findBox`'s `+8` advance, so it never exercised the advisory vector — only the new fixture and the ICNS fixture hang. A patch regression that drops the `boxSize < 8` guard is now caught by the JXL fixture; the ICNS break is still caught by the ICNS fixture.
- Selector ownership: the gate is selected — and stops failing open — when its defining files change (`examples/test-app/patches/`, `examples/test-app/security/`, and `examples/test-app/pnpm-workspace.yaml`, which carries the `patchedDependencies` declaration).

The patch itself is verified against both advisory vectors: the `boxSize < 8` guard in `readBox` stops the JXL `extractPartialStreams` stall on a size-0 `jxlp` (the guard, not `findBox`'s advance, is what hides a *matching* zero-size box), and the `imageHeader[1] < 8` break in the ICNS parser stops the zero-length-entry loop. The HEIF fixture is a termination budget pin: unpatched 1.2.1 has no unbounded loop in the HEIF path (a single `findBox` chain over the `+8`-advancing `findBox`), so it asserts the format named in the advisory stays within budget.

Closes Dependabot alerts #69 and #70 at the mitigation level (they cannot be cleared by a version bump; see residual note below).

## Validation

- **Plant proof of the cache hole (review request)**: with a patch-only change (dropped the `boxSize < 8` guard from the patch file), the pre-fix key hash was byte-identical (`000c575f…`) — a cache hit would restore the stale guarded artifact and the suite passed against it (installed `utils.js` carried the guard the tree patch no longer declared). With the fix, the same patch-only edit changes the key (`a7004046…` → `24f5bd59…`); the resulting install applied the weakened patch and the gate **failed** — `JXL zero-size jxlp box exceeded the parser budget` (`spawnSync … ETIMEDOUT`) — before the patch was restored.
- Restored state: `pnpm gate test-app-security` green (1 test, 4 fixtures: ICNS entry, HEIF box, JXL box, JXL zero-size jxlp box, each under a 1s budget), artifact resolved via the lockfile patch hash.
- `pnpm check:affected --run` green at this commit (format, lint, typecheck, layering, di-seams, fallow, mcp-metadata, build, package, integration, gate-manifest, affected-selector, `vitest related` over the changed surface).
- `pnpm gate gate-manifest` reports **52 checks wired across 27 lanes** (was 51) — the new gate is owned by a lane, so the "every gate owned and wired" invariant holds.
- New selector test asserts the gate is selected for each of its three defining paths and does not fail open; observed it red with the ownership entry removed, green restored.
- Rebased onto `309a5360f5`; main picked up no commits touching this PR's surface.

Residual: Dependabot reports by resolved version, so #69/#70 will remain "open" in the dashboard until upstream ships a fix or the maintainers mark them dismissed-as-mitigated; the patch + gate above are the durable control.

### DIFF
--- a/.github/actions/setup-test-app-dependencies/action.yml
+++ b/.github/actions/setup-test-app-dependencies/action.yml
@@ -10,7 +10,8 @@ runs:
   using: 'composite'
   steps:
     # Expo's dependency graph contains native binaries, so cache it by operating
-    # system, architecture, Node major, pnpm version, and the exact manifest.
+    # system, architecture, Node major, pnpm version, the exact manifest, and the
+    # pnpm patch files applied into it.
     - name: Resolve test app dependency cache key
       id: key
       shell: bash
@@ -24,7 +25,7 @@ runs:
       uses: actions/cache/restore@0057852bfaa89a56745cba8c7296529d2fc39830 # v4.2.3
       with:
         path: examples/test-app/node_modules
-        key: test-app-dependencies-v2-${{ runner.os }}-${{ runner.arch }}-node${{ steps.key.outputs.node_major }}-pnpm${{ steps.key.outputs.pnpm_version }}-${{ hashFiles('examples/test-app/package.json', 'examples/test-app/pnpm-lock.yaml', 'examples/test-app/pnpm-workspace.yaml', '.github/actions/setup-test-app-dependencies/action.yml') }}
+        key: test-app-dependencies-v2-${{ runner.os }}-${{ runner.arch }}-node${{ steps.key.outputs.node_major }}-pnpm${{ steps.key.outputs.pnpm_version }}-${{ hashFiles('examples/test-app/package.json', 'examples/test-app/pnpm-lock.yaml', 'examples/test-app/pnpm-workspace.yaml', 'examples/test-app/patches/**', '.github/actions/setup-test-app-dependencies/action.yml') }}
 
     - name: Install test app dependencies
       if: steps.restore.outputs.cache-hit != 'true'
@@ -36,4 +37,4 @@ runs:
       uses: actions/cache/save@0057852bfaa89a56745cba8c7296529d2fc39830 # v4.2.3
       with:
         path: examples/test-app/node_modules
-        key: test-app-dependencies-v2-${{ runner.os }}-${{ runner.arch }}-node${{ steps.key.outputs.node_major }}-pnpm${{ steps.key.outputs.pnpm_version }}-${{ hashFiles('examples/test-app/package.json', 'examples/test-app/pnpm-lock.yaml', 'examples/test-app/pnpm-workspace.yaml', '.github/actions/setup-test-app-dependencies/action.yml') }}
+        key: test-app-dependencies-v2-${{ runner.os }}-${{ runner.arch }}-node${{ steps.key.outputs.node_major }}-pnpm${{ steps.key.outputs.pnpm_version }}-${{ hashFiles('examples/test-app/package.json', 'examples/test-app/pnpm-lock.yaml', 'examples/test-app/pnpm-workspace.yaml', 'examples/test-app/patches/**', '.github/actions/setup-test-app-dependencies/action.yml') }}
--- a/.github/workflows/test-app-build-cache.yml
+++ b/.github/workflows/test-app-build-cache.yml
@@ -50,6 +50,12 @@ jobs:
         uses: ./.github/actions/run-gate
         with: { gate: test-app-typecheck }
 
+      # image-size has no fixed release for its parser DoS advisories; this gate
+      # proves the in-tree pnpm patch still terminates the zero-length vectors.
+      - name: Run the image-size parser security test
+        uses: ./.github/actions/run-gate
+        with: { gate: test-app-security }
+
       - name: Resolve native fingerprint
         id: fingerprint
         env:
--- a/examples/test-app/security/image-size-security.test.mjs
+++ b/examples/test-app/security/image-size-security.test.mjs
@@ -5,14 +5,13 @@ import path from 'node:path';
 import test from 'node:test';
 
 const packageStore = path.join(process.cwd(), 'node_modules', '.pnpm');
+const patchHash = /^ {2}image-size@1\.2\.1: ([0-9a-f]{64})$/m.exec(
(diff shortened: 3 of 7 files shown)


thymikee:
BLOCKED at exact head `24d621dfe55a0c558b2066cafc406ea82087af6b`.

The new gate claims patch-file edits as an owned input, but `setup-test-app-dependencies` does not hash `examples/test-app/patches/` into its `node_modules` cache key. A patch-only edit can therefore restore stale patched dependencies, skip `pnpm install`, and run the security test against the previous patch, making this enforcement vacuous for one of its defining inputs.

Include the patch files in the cache key or force frozen reinstall/patch-hash verification, then plant a patch-only change under a cache hit and show the gate fails before restoring it.

thymikee:
Confirmed — that was a real hole, and it made the enforcement vacuous for a defining input. Fixed at the cache key, plus two more weaknesses the exercise surfaced. New head: `fc61b0402d`.

**1. Cache key now includes the patch files (the fix you named).**
`.github/actions/setup-test-app-dependencies/action.yml` — both the `restore` and `save` keys now hash `examples/test-app/patches/**` alongside the manifests. The key is now a complete function of everything that determines `node_modules` content (package.json, lockfile, workspace.yaml, patch files, action). I chose the key fix over forcing a frozen reinstall so the cache keeps its performance value; a forced reinstall would throw away the benefit for the common no-change case.

**2. Observed the hole, then the fix (your requested evidence).**
I reproduced the key's input set locally and planted a patch-only change (dropped the `boxSize < 8` guard from `image-size@1.2.1.patch`, leaving the other three manifests untouched):

- **Pre-fix:** the key's inputs were unchanged by the patch-only edit → the key was invariant → a cache hit would restore the *good* patched `node_modules` and skip `pnpm install`. The security test then ran against that stale artifact and **passed** — even though the PR's patch no longer contained the guard. Exactly the vacuous path you described.
- **Post-fix:** the same patch-only edit changes the key (the `patches/**` input moved) → cache miss → `pnpm test-app:install` re-applies the *weakened* patch. The security test then **failed** — the JXL fixture hung and hit the 1s budget (`spawnSync … ETIMEDOUT`) — before I restored the good patch (after which it passes).

So a patch-only regression is now caught instead of silently passing on a stale cache.

**3. Two more issues the exercise surfaced (quality):**

- **The JXL fixture didn't actually exercise the infinite loop.** The original fixture used a zero-size box named `junk`, which `findBox` skips (it advances ≥8 bytes on a non-matching box) — so it only asserted the "no codestream" error path and passed even on vulnerable code. I confirmed against unpatched 1.2.1 that the *matching* zero-size `jxlp` box is what stalls `extractPartialStreams` (`offset = jxlpBox.offset + jxlpBox.size` with size 0), and added that as a fixture. It hangs on vulnerable code and returns under the patch. The `readBox < 8` guard is what hides a *matching* zero box — the `findBox` advance alone is not the fix.
- **Artifact resolution was first-match-by-directory.** The test picked the first `image-size@1.2.1_patch_hash=*` dir, which could be a stale sibling. It now resolves the package by the `patchedDependencies` hash recorded in `pnpm-lock.yaml` (the declared artifact), and fails loudly if the declaration is absent.

**Validation at `fc61b0402d`:** `pnpm check:affected --run` green (format/lint/typecheck/layering/di-seams/fallow/mcp-metadata/build/package/integration/gate-manifest/affected-selector/vitest-related). `pnpm gate test-app-security` green (4 fixtures: ICNS entry, HEIF box, JXL box, JXL zero-size jxlp box). Note the HEIF fixture remains a termination-budget pin — unpatched 1.2.1's HEIF path has no unbounded loop (single `findBox` chain over the `+8`-advancing `findBox`), so it asserts the advisory-named format stays within budget rather than asserting a hang.

thymikee:
Code review is clean at exact head fc61b0402de1c1c11c572e494342e139804d2985. The patch-only cache-key hole is fixed for restore and save; the planted weakened-patch run discriminates correctly; lockfile-derived artifact selection, gate ownership, and the matching jxlp fixture are sound. The exact-head fingerprint lane has already exercised test-app-security green. No device evidence is owed for this CI/security-only change. Marking ready-for-human while the remaining exact-head CI finishes; merge readiness still depends on those checks completing successfully.

PR #2272, opened 2026-09-03
fix(remote): materialize test suite artifacts against a remote daemon

### LINKED ISSUE #2246: agent-device test cannot make its results directory and stops before the first flow
## Summary

`agent-device test` stops before it runs the first flow. It cannot make its own results
directory. The error is `ENOENT` from `mkdir`.

This happens with `--artifacts-dir`. It also happens with the default directory. So it is not a
configuration mistake.

`agent-device replay` runs the same files in the same session correctly. Only `test` fails.

## Version

`agent-device` 0.20.10. The CLI runs on an `ubuntu-latest` GitHub Actions runner. The daemon is a
remote EAS Simulator session. The daemon also reports 0.20.10. The target is an iOS simulator.

## Steps to repeat, with an explicit directory

The caller makes the directory first. The tool then fails inside it.

```
mkdir -p remote-device-artifacts/ad-test

agent-device test 'qa-flows/open-*.ad' \
  --platform ios \
  --state-dir "$(agent-device session state-dir)" \
  --artifacts-dir remote-device-artifacts/ad-test \
  --reporter default \
  --reporter junit:remote-device-artifacts/ad-test/junit.xml
```

Result:

```
Error (UNKNOWN): ENOENT: no such file or directory, mkdir
'/home/runner/work/installer-app/installer-app/remote-device-artifacts/ad-test/cd5f9c01feec8d70/qa-flows__open-grid-settings.ad/attempt-1'
Hint: Unexpected internal error. Retry with --debug and report the diagnostics log if it persists.
Diagnostic ID: mtkb2nu0-e698d27f
```

## Steps to repeat, with the default directory

Remove `--artifacts-dir`. The tool then uses the documented default location.

```
agent-device test 'qa-flows/open-*.ad' \
  --platform ios \
  --state-dir "$(agent-device session state-dir)" \
  --reporter default
```

Result:

```
Error (UNKNOWN): ENOENT: no such file or directory, mkdir
'/home/runner/work/installer-app/installer-app/.agent-device/test-artifacts/7c8ca96c9eb63bfc/qa-flows__open-grid-settings.ad/attempt-1'
Diagnostic ID: mtkboyfp-5709fff6
```

## Expected result

`test` runs each flow it finds. It writes `replay.ad` and `result.txt` for each attempt. The
`replay-e2e` page describes this behaviour.

## Probable cause

The path that fails is three levels below a directory that exists. In the first example, the
caller made `remote-device-artifacts/ad-test`. The tool then failed on
`ad-test/<run-id>/<flow>/attempt-1`.

A `mkdir` call without the recursive option gives this error. A recursive call makes all three
levels. This would correct both examples.

## Effect on users

This defect blocks more than the results files.

`context retries=` applies to `test` only. `replay` does not read it. So a caller cannot use
per-file retries while `test` fails. The caller must add retries around `replay` instead.

## Other information

Both runs used the same repository and the same session type. They ran a few minutes apart. The
journey that reaches the screens under test passed in both runs.

thymikee:
## Summary

Fixes #2246.

Remote `agent-device test` requests now keep daemon-owned suite artifacts off caller-only paths, transfer the completed suite through the existing tracked-artifact transport, and report the exact caller-local invocation directory for the suite, tests, and failures.

Directory downloads use the existing safe archive extractor. They stage beside the caller's destination, propagate one cancellation signal through download and extraction, clean up before timeout rejection, and publish with one atomic rename. The client-side download owner returns the exact materialized path; generic response materialization no longer reconstructs it independently.

The follow-up reduces the original `src/remote/daemon-artifacts.ts` growth by extracting its download machinery into the 191-line `src/remote/artifact-download.ts`. Overall scope is 15 files (+1040/-124), mostly 6 focused test files (+631 lines). Production growth remains below the repository's 700-line escalation threshold, while the shipped bundle is +2.2 kB raw / +784 B gzip on the latest size report.

## Validation

- Planted-red at `99615f818c05a388b7cf1d17fc7285c032857d9e`: composed materialization returned the bare caller root instead of `<root>/<suite-id>`, and directory staging was outside the destination filesystem.
- `pnpm check:affected --run`: 484 files and 3,610 tests passed, including build, layering, Fallow, and daemon wire compatibility.
- HTTP remote-daemon proof on the booted `iPhone 17 Pro` Simulator at `f1bda16b72ad9a141a9f168c1ea1f331e3c5f543`: default artifact root 1/1 passed; relative `--artifacts-dir relative-artifacts` 1/1 passed. Both returned the caller-local `<root>/<suite-id>` and materialized `replay.ad` plus `result.txt` under the per-attempt directory.

### DIFF
--- a/packages/contracts/src/replay.ts
+++ b/packages/contracts/src/replay.ts
@@ -141,6 +141,13 @@ export type ReplaySuiteResult = {
   durationMs: number;
   failures: ReplaySuiteTestFailed[];
   tests: ReplaySuiteTestResult[];
+  /**
+   * The suite's own artifacts root (the parent of every test's `artifactsDir`), as resolved on
+   * the host that ran the suite. Absent when the suite produced no attempt (e.g. every source
+   * was filtered out). #2246: a remote daemon rewrites this to the caller-local path once the
+   * directory has been transferred back, so it always names a path the caller can open.
+   */
+  artifactsDir?: string;
   snapshotDiagnostics?: SnapshotDiagnosticsSummary;
 };
 
--- a/packages/kernel/src/contracts.ts
+++ b/packages/kernel/src/contracts.ts
@@ -114,7 +114,8 @@ export type DaemonArtifactKnownType =
   | 'screen-recording'
   | 'screen-recording-chunk'
   | 'screen-recording-telemetry'
-  | 'trace-log';
+  | 'trace-log'
+  | 'test-artifacts';
 
 export type DaemonArtifactType = DaemonArtifactKnownType | (string & {});
 
--- a/packages/replay-test/src/internal/__tests__/session-test-artifacts.test.ts
+++ b/packages/replay-test/src/internal/__tests__/session-test-artifacts.test.ts
@@ -4,10 +4,26 @@ import fs from 'node:fs';
 import os from 'node:os';
 import path from 'node:path';
 import {
+  DEFAULT_TEST_ARTIFACTS_ROOT,
   materializeReplayTestAttemptArtifacts,
   prepareReplayTestAttemptArtifacts,
+  resolveReplayTestArtifactsDir,
 } from '../session-test-artifacts.ts';
 import type { ReplayTestAttemptOutcome } from '../session-test-types.ts';
+
+test('resolveReplayTestArtifactsDir falls back to the default root when artifactsDir is omitted', () => {
+  const dir = resolveReplayTestArtifactsDir({ cwd: '/repo', suiteInvocationId: 'abc123' });
+  assert.equal(dir, path.resolve('/repo', DEFAULT_TEST_ARTIFACTS_ROOT, 'abc123'));
+});
+
+test('resolveReplayTestArtifactsDir resolves an explicit relative artifactsDir against cwd', () => {
+  const dir = resolveReplayTestArtifactsDir({
+    artifactsDir: 'remote-device-artifacts/ad-test',
+    cwd: '/repo',
+    suiteInvocationId: 'abc123',
+  });
+  assert.equal(dir, path.resolve('/repo', 'remote-device-artifacts/ad-test', 'abc123'));
+});
 
 // Building outcomes from a DaemonResponse is the adapter's job and is pinned on that side; a
 // package test states the neutral outcome directly (#1478 P3b).
--- a/packages/replay-test/src/internal/session-test-artifacts.ts
+++ b/packages/replay-test/src/internal/session-test-artifacts.ts
@@ -3,7 +3,15 @@ import path from 'node:path';
 import { trimEdgeDashes } from '@agent-device/kernel/collections';
 import type { ReplayTestAttemptOutcome } from '@agent-device/replay-test';
 
-const DEFAULT_TEST_ARTIFACTS_ROOT = '.agent-device/test-artifacts';
+/**
+ * `test`'s default artifacts root when `--artifacts-dir` is not given. `src/remote/daemon-artifacts.ts`
+ * keeps its own copy of this literal (to redirect a remote request without resolving anything
+ * against the daemon's `cwd`, #2246): pulling it from `@agent-device/contracts` instead grew this
+ * package's eager import closure by 3 modules for one string, past its pinned budget
+ * (`scripts/__tests__/eager-closure-budgets.test.ts`) — not worth it for a value that changes
+ * only if this line does.
+ */
+export const DEFAULT_TEST_ARTIFACTS_ROOT = '.agent-device/test-artifacts';
 
 export function resolveReplayTestArtifactsDir(params: {
   artifactsDir?: string;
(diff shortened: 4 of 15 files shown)


thymikee:
BLOCKED at exact head 99615f818c05a388b7cf1d17fc7285c032857d9e. Two production composition bugs remain despite green CI.

First, `attachRemoteReplayTestArtifacts` correctly reports `<clientRoot>/<suiteInvocationId>` but registers `localPath: clientRoot`; `materializeRemoteArtifacts` then unconditionally overwrites `data.artifactsDir` with that bare root. Add a real composed materialization regression proving the final response and extracted `replay.ad`/`result.txt` resolve under the invocation directory, with planted-red evidence.

Second, staging under `os.tmpdir()` can hit `EXDEV`; the fallback recursively copies directly into the final caller directory, so failure/cancellation/timeout can leave partial or later background writes. Stage on the destination filesystem, propagate cancellation through extraction/copy, then atomically rename; add planted-red EXDEV/partial-copy and timeout coverage.

Also provide the linked issue’s real remote-daemon/device evidence for default and relative `--artifacts-dir`.

Size review: headline +776 is ~275 production LOC, below the 700-line trigger, and package growth is +2.5 kB. Still, the path bug shows ownership is split between daemon rewriting and generic client materialization. Please itemize growth and explain why one client-side owner returning both download destination and final reported path was rejected.

thymikee:
Resolved the blocking review at `f1bda16b72ad9a141a9f168c1ea1f331e3c5f543`.

- The composed regression now drives `attachRemoteReplayTestArtifacts` through the real `/artifacts` HTTP route and client materializer, then proves `data.artifactsDir`, `artifacts[].localPath`, `replay.ad`, and `result.txt` all land under `<client-root>/<suite-id>`. Planted red on `99615f818c`: expected the suite directory, received the bare root.
- Directory downloads now stage in `.agent-device-download-*` beneath the destination root, pass one abort signal through the response pipeline and safe extractor, await cleanup before timeout rejection, and publish only via same-filesystem rename. The old `EXDEV` recursive-copy fallback is gone. Planted red on `99615f818c`: no destination-side stage existed during an active download; the new timeout regression also proves it is gone before rejection returns.
- The owning client function returns the exact materialized path. `materializeRemoteArtifacts` consumes that result rather than independently reconstructing the final path. Download machinery moved to `src/remote/artifact-download.ts`; `daemon-artifacts.ts` is 439 lines rather than growing to 637.

Local gate:

```text
pnpm check:affected --run
484 test files passed; 3,610 tests passed; build, layering, Fallow, and daemon wire compatibility passed.
```

Live acceptance used the pushed source checkout, an isolated HTTP proxy/daemon, and the discovered booted `iPhone 17 Pro` Simulator (`F7D6F9A4-4FCC-4DD7-AC0B-3280C9319CB9`). The temporary auth token is redacted:

```text
pnpm ad proxy --port 4317 --state-dir /tmp/agent-device-pr2272.u8WVa1/host-state --daemon-auth-token <temporary-token>

AGENT_DEVICE_DAEMON_AUTH_TOKEN=<temporary-token> node bin/agent-device.mjs test /tmp/agent-device-pr2272.u8WVa1/work/pr2272-remote-proof.ad --platform ios --device "iPhone 17 Pro" --remote-config /tmp/agent-device-pr2272.u8WVa1/direct-remote.json --json
# 1/1 passed; artifactsDir=/private/tmp/agent-device-pr2272.u8WVa1/direct-client-cwd/.agent-device/test-artifacts/f8598a66e7497c85

AGENT_DEVICE_DAEMON_AUTH_TOKEN=<temporary-token> node bin/agent-device.mjs test /tmp/agent-device-pr2272.u8WVa1/work/pr2272-remote-proof.ad --platform ios --device "iPhone 17 Pro" --artifacts-dir relative-artifacts --remote-config /tmp/agent-device-pr2272.u8WVa1/direct-remote.json --json
# 1/1 passed; artifactsDir=/private/tmp/agent-device-pr2272.u8WVa1/direct-client-cwd/relative-artifacts/c4545591309060f7
```

Both invocation directories contained `pr2272-remote-proof.ad/attempt-1/replay.ad` and `result.txt`. `session list` was empty afterward; the temporary proxy profile was disconnected and the proxy stopped.

CI note: the first Android smoke attempt timed out waiting for an unrelated landscape snapshot with `readableCaptures: 0`; the failed job has been rerun on the same head.

thymikee:
Follow-up exact head `53d30bcf78c2c6d0a2fd63bbfe963766d7392e7f`: Coverage correctly caught one extra eager CLI module (`383` vs `382`). `downloadRemoteArtifact` now function-scope imports the extracted downloader, so the routine CLI closure is back at the merge-base budget. Focused closure + wire + artifact tests: 455/455 passed. Re-run `pnpm check:affected --run`: 484 files / 3,610 tests passed, plus build, layering, Fallow, and wire compatibility. Fresh exact-head CI is running.

Issue #2192, opened 2026-08-31
perf(ios): decide Simulator AX bridge viability with a bounded spike


thymikee:
Parent: #2188

Blocked by: #2189, #2190

## Outcome

Run a bounded, disposable iOS Simulator accessibility-acquisition spike and make an explicit GO or
NO-GO decision. The spike compares candidate mechanisms behind #2190's acquisition adapter and must
not alter production routing.

A fast result is insufficient: correctness, lifecycle, cancellation, and bounded-resource evidence
are mandatory. A NO-GO closes the bridge path cleanly while leaving the converged engine and provider
work useful.

## Exact scope

- Compare viable host-side Simulator accessibility mechanisms through one batch acquisition adapter.
- Include a control using task-owned, pre-boot Simulator accessibility preferences such as
  `AutomationEnabled` and `IgnoreAXServerEntitlements`, plus related keys discovered during the
  spike.
- Apply preference experiments only to disposable task-owned Simulators before boot; record the
  exact plist diff and restore or delete the Simulator during cleanup.
- Use a framed request protocol with logs on stderr and one batch tree response; never per-node IPC.
- Report only acquired facts: raw nodes, viewport evidence, target generation, truncation, and typed
  residue. Do not implement visibility, hittability, scope, public depth, or semantic compaction in
  the reader.
- Bound request bytes, response bytes, node count, traversal depth, CPU work, memory, and duration.
- Model unsupported mechanism, malformed tree, stale target generation, timeout, cancellation,
  process crash, and transport failure as typed outcomes.
- Exercise quiet, list, nested-scroll, alert/system-surface, relaunch-during-capture, and XCTest-stress
  fixtures from #2189.

## Measurement contract

Use #2189's harness and state definitions. Report raw samples for acquisition and prototype
daemon-side presentation separately, including CPU, memory, payload size, and crash recovery.

Keep candidate-owned observation latency separate from Simulator boot, application launch/readiness,
the current production daemon, and XCTest runner preparation. A bridge measurement must not charge
the candidate for `xcodebuild` or current XCTest-backed `open` work that the selected route would
avoid. A threshold stated after app readiness begins only after the expected app generation is
observably ready.

Correctness, fidelity, generation, lifecycle, cancellation, recovery, and bounded-resource gates are
hard. Latency has two tiers:

### Hard viability targets

- warm daemon-resident acquisition p50 under 300 ms and p95 under 500 ms for every representative
  screen;
[...]

PR #2277, opened 2026-09-04
feat(ios): productionize Simulator AX snapshot bridge

### LINKED ISSUE #2196: feat(ios): productionize the selected Simulator snapshot bridge
Parent: #2188

Blocked by: #2192, #2231

Start condition: #2192 must close with an explicit GO naming the selected mechanism. A NO-GO closes
this issue without implementation; do not substitute another private mechanism ad hoc.

## Outcome

Productionize the selected Simulator accessibility reader as private, host-local infrastructure
behind one `@agent-device/platform-apple` acquisition adapter. The bridge must build, cache, spawn,
bind, read, cancel, fail, and recover deterministically, but no production snapshot request routes
to it yet.

### Native bridge

- Put the small native source under `apple/snapshot-bridge/`.
- Implement the spike-proven framed batch request/response protocol with logs on stderr.
- Return acquisition facts only: raw nodes, viewport evidence, target generation, truncation, typed
  residue, and protocol/source versions.
- Enforce request/response bytes, node count, depth, work, and duration limits.
- Return typed unsupported, malformed-tree, stale-target, timeout, cancellation, crash, and transport
  failures.

### Apple host adapter and lifecycle

- Put adapter, build/cache, spawn supervision, request client, and lifecycle under
  `packages/platform-apple/src/snapshot-source/` behind one narrow runtime facet.
- Cache outside the npm package using at least source hash, Xcode build, macOS version, architecture,
  and Simulator-runtime compatibility inputs.
- Compile lazily or through explicit preparation; never compile in npm postinstall.
- Use atomic cache publication and reject stale/incompatible protocol/source versions.
- Key target state by UDID plus opaque target generation. Do not retain capture data or comparison
  baselines across sessions or generations.
- Preserve request cancellation and deadline signals through build, bind, and read.
- Write bridge/process diagnostics to the owning session's runner/diagnostic artifacts without
  replacing existing warnings or typed error details.

## Protocol decision

Hand-maintain the small Swift/TypeScript wire carriers and pin them with round-trip fixtures. Do not
add code generation now. Revisit generation only after an incompatible protocol evolution, a second
independent wire consumer, two demonstrated optionality/enum-drift defects, or a third runtime.

## Security and proxy seam

- The bridge listener/socket is private host infrastructure and is never exposed through
  `agent-device proxy`.
- Do not add a bridge, helper-control, or daemon-administration HTTP route.
- Bound and validate every decoded frame before allocation-heavy work.
- Do not put daemon, proxy, signing, or user credentials in source, cache keys, logs, or artifacts.

## Non-goals

- No production snapshot routing, circuit breaker, XCTest fallback, runner-demand, open/relaunch,
  provider, or proxy behavior change.
- No public CLI flag, new workspace package, downloaded executable, pre-signed artifact, or
  physical-iPhone support.
- Do not copy Swift presentation into the bridge; it acquires only.

## Acceptance and evidence

- The production adapter reproduces #2192's fidelity and bounded-resource results through its public
  runtime facet.
- Build cache cold, hit, invalidation, concurrent publication, corrupt entry, and toolchain-change
  cases are tested.
- Spawn/read/cancel/crash/restart/stale-generation tests use typed outcomes and leave no orphaned
  helper ownership.
- A planted route/ownership violation proves the bridge cannot be reached from the proxy or outside
  the Apple runtime facet.
- Packed, installed, and bundled size deltas are reported against #2189 and stay within the accepted
  budget; the compiled cache is excluded from the npm artifact.
- Focused native/host tests, XCTest selection when applicable, layering, packaging, and
  `pnpm check:affected --run` pass; attach live Simulator evidence.

## Worker stop conditions

- Stop if production behavior must route to the bridge to test the adapter; routing belongs later.
- Stop if cache correctness requires an unbounded compatibility fallback.
- Stop if the selected mechanism no longer meets #2192's GO evidence on the supported toolchain.

thymikee:
## Summary

Closes #2196.

Adds the private Apple Simulator AX snapshot acquisition facet:

- a bounded, framed native bridge returning typed raw-tree facts;
- one host adapter with lazy atomic build/cache, deadline-aware serialization, target validation, cancellation, diagnostics, crash recovery, and instance-safe lifecycle;
- package, protocol, ownership, and focused regression guards;
- no production routing, public CLI, runner/provider integration, physical-device path, downloaded executable, or new workspace package.

The guest remains Objective-C because #2192 selected idb v1.5.2's Objective-C `SimulatorFrameworkBridge` shape and the boundary dynamically loads private classes/functions, invokes runtime selectors, and contains `NSException` failures. Swift would still require an Objective-C shim. The September Python code was spike orchestration only and is not part of production.

## Validation

Tested commit: `e7b32ca5b80b5d68699f7dd8decd06d64589c20a`.

- `pnpm check:affected --run`: all runnable checks passed, including build, typecheck, lint/format, layering, Fallow, integration, and clean-installed package verification.
- Focused snapshot/package tests: 8 files, 35 tests passed.
- Previous implementation head `7ec3327bf8` passed hosted clean-install bridge preparation and targeted XCTest checks.
- Live disposable iPhone 17 Pro Simulator proof: two independent source instances acquired the same target concurrently, each returned a complete tree, and cleanup left no bridge residue.
- Current-head GitHub checks remain authoritative for native/device lanes.

Scope: 41 touched files, limited to the Simulator snapshot-source facet, its native guest, evidence, packaging, and owning gates.


thymikee:
Review pass: correctness and implementation efficiency

This is not merge-ready yet. Please keep the repair narrow and address these in order:

1. **One deadline must cover prepare + bind + read.** `adapter.ts` starts preparation before acquisition timing, `cache.ts` gives probes 10s and compilation 120s, then `lifecycle.ts` creates a fresh `maxDurationMs` deadline for connect and another fresh timer for the round trip. A request can therefore exceed its declared budget by minutes. Compute one absolute deadline at the facet boundary and pass remaining time through cache locking/probes/build, connect, and read. Add a test whose earlier phase consumes most of the budget and proves a later phase cannot reset it.

2. **An abandoned in-flight request can poison the persistent helper.** On timeout/cancellation, `roundTrip` only destroys the socket; `request` leaves the helper alive. The native server performs the AX call synchronously and accepts one connection at a time, so a hung abandoned AX call can keep the helper busy while the next request reconnects and times out. The existing cancellation test currently blesses that leak. The simplest safe rule is: cancellation before dispatch leaves a healthy helper alone; cancellation/timeout after dispatch reaps that exact helper, and the next request starts a fresh one. Prove no orphan ownership and recovery.

3. **Generation/stale-target evidence is currently manufactured on the host.** The wire request contains only PID and the native response merely echoes that requested PID; the adapter then attaches the requested generation to the result. This cannot observe a relaunch race or satisfy the issue's target-generation/stale-generation contract. Carry the opaque generation in the framed exchange, validate it before publication, and make the lifecycle/stale test exercise a response from the previous generation rather than only asserting that a generation change restarts the process.

4. **The native boundary does not yet enforce the declared resource contract.** It has global frame/node/depth caps, but no per-request duration/work/response-byte bound; an AX call can run past the host deadline. Also, malformed child nodes are silently omitted (`nodeFromSnapshot` returns nil and the parent continues), yielding a successful partial tree without a typed malformed/truncated outcome. Fail the whole request on malformed structure, pass explicit per-request bounds, and use a process-level watchdog/reap strategy rather than building a complex cancellation protocol.

5. **Tighten the runtime facet instead of exporting the test harness.** The public subpath exposes `SnapshotSourceHost`, cache/source-root overrides, bridge binary/cache metadata, and both throwing `acquire` plus `acquireOutcome`. Repository guidance explicitly says not to add production exports for test-only dependency injection. Keep host/cache injection internal to the module tests and expose one failure model through the narrow facet. This should also let the wildcard Fallow exemption shrink materially.

6. **Remove request limits from build identity.** `maxNodes` and `maxTraversalDepth` affect each request, not the compiled executable, but they are included in both the prepared map key and cache key. Different capture limits therefore create duplicate cache entries/rebuilds. Fingerprint only compilation inputs (explicit `.m`/`.h` files, not README/license) plus complete toolchain compatibility. The current toolchain identity also records macOS product version but not its build number, despite the issue requiring version/build compatibility.

7. **Fix exact-head gates as part of this repair.** Coverage is red because all six new `.test.ts` files use `node:test`, while the coverage job discovers them as Vitest files and reports “No test suite found.” Convert them to the repository's Vitest test surface. Bundle Size is red because the head's new required-asset assertion is also run while measuring the base commit, where that asset cannot exist; make the assertion comparison-aware or validate the new asset only for the head. The iOS smoke failure is on an unchanged production route, so rerun it after the deterministic failures are fixed and classify it from runner evidence if it repeats.

The diff is currently +3,222 lines (+2,291 outside tests/docs) for a dormant facet. I would not split the feature solely for line count, but I would do a tightening pass around the decisions above: one public outcome API, one absolute deadline, one helper invalidation rule, request-scoped limits, and a cache key made only from build identity. Then refresh live evidence with warm reuse, relaunch/generation, cancellation/timeout recovery, and bounded payload cases on the corrected exact head.

thymikee:
BLOCKED at exact head `5223968064c91dac4191b0b5d25534b102a928e0`.

Confirmed blockers:

1. The exact-head Bundle Size job proves the shipped artifact is incomplete: `npm pack` is missing `apple/snapshot-bridge/*SnapshotBridge.m`, while the runtime resolves those sources for lazy compilation. Fix the base/head size assertion so it can measure the old base, ensure the tarball contains both native implementation sources, and exercise prepare/acquire from a clean-installed tarball.

2. Coverage discovers all six new `snapshot-source/*.test.ts` files as Vitest files, but they import `test` from `node:test`; all fail with `No test suite found`. Put these tests in the repository's Vitest topology (or a declared Node-test lane) and prove adapter/cache/host/lifecycle/protocol/tree execute in check:affected and Coverage.

3. Cancellation/deadline ownership is not complete. `roundTrip` destroys only the socket, and the manager deliberately leaves the helper alive on cancellation/timeout even though synchronous native AX work may still be blocked; the next request can reuse that helper. Cache-lock acquisition can wait 180s without a signal, and the memoized prepare Promise couples every same-key caller to the first caller's signal. Carry one remaining deadline through lock/probes/build/bind/read, give each waiter independent cancellation, and invalidate/terminate a helper whose in-flight native request is cancelled/timed out; add contention and stuck-native tests proving no orphaned work survives.

4. The required native↔TypeScript wire parity proof is missing. Current protocol tests fabricate TS JSON only, while versions, keys, limits, and attribute vocabulary are duplicated in Objective-C and TS. Add a native-produced golden/round-trip fixture or equivalent parity gate, including malformed/version/stale cases.

5. The PR is +3,222/-0, over the repository's 1,000-line gross review budget, with roughly 2.1k production lines and +69,029 npm-unpacked bytes. The implementation files are individually sized and the ownership split is coherent, but the PR must be split into reviewable layers or explicitly reconcile the exception, itemize retained growth, and record why a materially smaller design was rejected.

The iOS Smoke failure is on the unchanged XCTest Settings route and appears unrelated to this dormant facet, but the head remains non-merge-ready until that job is green or authoritatively rerun. Dependencies #2192/#2231 are closed, ADR 0004 supports the facet, the direct Simulator evidence is useful, and no production route was added.

thymikee:
Addressed all actionable human comments on exact head `cbd94e49d5122ee1d1816e91a97bad191f1f9591`.

1. **One absolute deadline**
   - `adapter.acquire` creates one deadline before preparation.
   - The same remaining budget flows through cache locking, probes, compilation, bind/connect, exchange, and decode.
   - Added the regression where preparation consumes the budget and a later phase cannot reset it.

2. **Cancellation and helper ownership**
   - Cancellation or timeout after dispatch destroys the socket and reaps that exact persistent helper with TERM then KILL if needed.
   - Pre-dispatch cancellation preserves a healthy helper.
   - Added post-dispatch cancel/timeout, stuck-native recovery, contention, and pre-dispatch preservation tests, including no-orphan cleanup proof.

3. **Generation and stale-target evidence**
   - Generation is now carried opaquely in the framed request and native response.
   - Transport validates request id, PID, and generation before publication.
   - The lifecycle test feeds a response from the prior generation; published lineage uses the validated response generation.

4. **Native resource and malformed-input contract**
   - Requests carry explicit duration, traversal-depth, node-count, and response-byte limits.
   - The native runtime applies a per-request watchdog around synchronous AX/tree work.
   - Malformed nodes, attributes, or children fail the whole request with a typed malformed result instead of silently producing a partial tree.
   - Native sources compile cleanly with clang `-Werror`.

5. **Narrow public facet**
   - The public `@agent-device/platform-apple/snapshot-source` surface is limited to the factory, `acquire`/`close`, request/target/limit types, and one `SnapshotSourceOutcome` failure model.
   - Host, cache, binary, and source-root injection remain internal to module tests.
   - The wildcard Fallow exemption was removed and replaced with explicit owned names.

6. **Cache identity**
   - Per-request limits are removed from build/cache identity.
   - Identity fingerprints only the explicit bridge `.m` and `.h` compilation inputs, excludes README/license/docs, and includes the macOS product build number plus the rest of the toolchain identity.
   - Added cache-key and independent-waiter cancellation coverage.

7. **Exact-head gates, package proof, and growth reconciliation**
   - All six new source tests use Vitest and execute in both affected checks and hosted Coverage.
   - Bundle Size is base/head-aware: the old base does not require the new bridge, while the head requires both native `.m` sources.
   - Clean-installed tarball validation exercises the shipped native sources and their prepare/acquire implementation.
   - The retained four-layer design and rejected smaller alternatives are recorded in [the evidence reconciliation](https://github.com/callstack/agent-device/blob/cbd94e49d5122ee1d1816e91a97bad191f1f9591/docs/evidence/ios-simulator-snapshot-source-2026-09-04.md).

Proof:
- Local `pnpm check:affected --run`: all runnable checks passed; focused source suite: 8 files, 30 tests passed.
- Exact-head [CI run](https://github.com/callstack/agent-device/actions/runs/33856907436): Coverage, Repo Guards, Integration, Lint/Format, Typecheck/Package, Compatibility/Provenance all passed.
- [Bundle Size](https://github.com/callstack/agent-device/actions/runs/33856907680), [native fingerprint](https://github.com/callstack/agent-device/actions/runs/33856907450), CodeQL, Ownership/context, command docs, and deploy preview all passed.
- [iOS smoke rerun](https://github.com/callstack/agent-device/actions/runs/33856907415/job/100975819287) passed Settings replay, fixture E2E, gesture replay, host-focus preservation, artifact upload, and cleanup. The first attempt hit the same unchanged post-sheet `capture was truncated` baseline path previously observed on main; rerunning the unchanged code cleared it. Physical-device smoke was skipped because no `IOS_UDID` is configured.
- Direct local simulator evidence also passed on the verified iPhone 17 Pro simulator: 77 nodes, no truncation, source v1.5.3, with no bridge residue after close.

All requested comments are addressed. READY for review; not merged.

thymikee:
Second quality pass on finished head `cbd94e49d5122ee1d1816e91a97bad191f1f9591`

The prior correctness fixes are real and the focused/hosted checks are green, but I do not think this is ready yet. There are three behavioral blockers and a small tightening pass:

1. **[P1] The production facet no longer preserves the selected route's latency.** Every `acquire` calls `ensureSnapshotBridgeBinary`; every cache hit re-reads/hashes all three native inputs and launches five sequential toolchain subprocesses before touching the persistent socket. A generation change then deliberately kills and respawns the helper even though the helper retains only Simulator-scoped framework state, not capture/generation state. The checked-in “prepared cache” result is 2,560 ms, already above #2192's 2,000 ms nonresident hard ceiling, and there is no corrected warm/relaunch evidence (the selected route was 9–121 ms warm p95 and 93–235 ms relaunch p95). Keep only a successfully completed prepared binary per runtime in the source instance—do not share an in-flight signal-bound promise—and reuse a healthy helper across app generations, carrying/validating generation only on each request. Simulator reboot/process death already invalidates the helper. Then record repeated warm and relaunch timings through the public facet.

2. **[P1] Cancellation does not cover the per-Simulator queue.** `SnapshotBridgeManager.request` waits inside the non-abortable `withKeyedLock`; the first signal/deadline check runs only after the previous same-UDID request releases it. A canceled queued request can therefore remain pending for the full prior native request, contrary to the end-to-end cancellation contract. The delay/cache-lock abort listeners also need the usual post-registration `signal.aborted` recheck to close the check-then-listen race. Add a two-request contention test that aborts the waiter while the first request is still in flight and proves prompt cancellation with no later dispatch. Keep the solution local to this manager; do not widen the shared keyed-lock interface unless another owner needs it.

3. **[P1] Process logs still do not follow the diagnostics contract.** `SnapshotSourceHost.emitDiagnostic` is exposed but never called; both `lifecycle.ts` and `transport.ts` duplicate `bridgeProcessExited` and put up to 64 KiB of stderr into typed error details. Emit the bounded process log once into the active diagnostic scope, keep only stable identifiers/exit facts in the error, and centralize crash construction. This both fixes the issue contract and removes the exact duplication/unused host capability.

4. **[P2] Simplify the framing and internal seams.** `SnapshotBridgeFrameDecoder` repeatedly `Buffer.concat`s the entire accumulated response, making a multi-chunk 4 MiB tree quadratic in copying, and supports multiple coalesced response frames even though the manager serializes exactly one request/response per connection. Replace it with one bounded single-frame accumulator (chunks + byte count, one final concat). Also remove the unused `temporaryDirectory`, `listDirectory`, and `readProcessStartTime` members from `SnapshotSourceHost`, and require one deadline shape internally instead of keeping optional `signal`/`deadline` entry paths solely for direct tests.

5. **[P2] Consolidate package/wire proof instead of overstating it.** Clean-install validation requires only the two `.m` files although runtime preparation also requires `SnapshotBridgeRuntime.h`; it returns success when the entire bridge directory is absent, and its substring checks do not actually exercise prepare/acquire. Make one authoritative three-file asset assertion and delete the substring ceremony/duplicate weak check. Likewise, `native-wire-golden.json` is hand-authored and the parity test only regex-searches Objective-C source, so it is not a native-produced round trip. Either produce the fixture from the compiled bridge in the existing macOS/native lane or rename/shrink this to an honest literal-vocabulary guard; do not add another protocol abstraction.

Validation from this review: all current GitHub checks are green; local focused Vitest passed 6 files / 23 tests. I could not repeat live Simulator validation because `agent-device` is not available on this task's PATH, so the performance conclusion above uses the exact-head checked-in live result plus direct code-path inspection. Please keep the repair to these ownership decisions; no split or broad refactor is needed.

thymikee:
Sentinel recheck at exact head `cbd94e49d5122ee1d1816e91a97bad191f1f9591`: still BLOCKED. The existing quality-pass findings on queued cancellation and diagnostics remain valid, and two additional ownership defects are confirmed.

1. **P1 — generation validation is self-confirming.** The host sends the expected generation, the native bridge returns that same request value unchanged, and transport compares it back to the original. A relaunch during AX acquisition can therefore publish an old tree under the expected generation. The fake-socket mismatch test does not prove the native route observes a relaunch. Revalidate an authoritative target-generation fact after acquisition and before publication, or supply a generation the native producer can independently observe.

2. **P1 — independent public source instances collide on one socket.** Each `createSimulatorSnapshotSource()` owns a separate manager/session map, while socket identity is only UDID + host PID. A second instance unlinks and replaces the first instance's socket; closing either can unlink the other's route. Use one process-level owner per Simulator target or instance-unique sockets with ownership-safe cleanup, and add a two-factory regression.

3. **P2 — cancellation still begins after the per-Simulator lock is acquired.** A cancelled follower remains queued for the leader's full request. Make queue acquisition deadline/abort-aware and prove a blocked follower cancels promptly without later dispatch.

All exact-head checks are green and the size reconciliation is itemized, but these are correctness blockers in the dormant facet itself; do not apply `ready-for-human`.

thymikee:
Addressed all five points in 597cb16db1, with the evidence revision in 6e55aba5d6.

1. Preparation now caches only a successfully completed binary per runtime in the public source instance. In-flight builds are not cached. Healthy helpers are reused across app generations, while each request carries and validates its PID and generation. Public-facet evidence records 14.1 ms warm p95 and 97.1 ms relaunch p95.
2. Per-Simulator serialization is now a manager-local queue with the request deadline and abort signal applied while waiting. Post-registration abort checks cover the queue, delay, cache-lock, and socket races. A local test proves a queued cancellation returns promptly without a later write.
3. Process-exit handling is centralized. It emits one bounded stderr diagnostic per helper process and keeps only typed PID/exitCode facts in the process-crash error; duplicate lifecycle and transport construction is removed.
4. Framing uses a bounded header plus payload-chunk accumulator, performs one final concat, and rejects trailing/coalesced frames. Unused host capabilities are removed and all internal paths require the single SnapshotSourceDeadline shape.
5. Package verification uses one authoritative three-asset assertion, including SnapshotBridgeRuntime.h, and fails when the bridge directory or any asset is absent. The hand-authored wire envelope is replaced by an honest literal-vocabulary guard covering TS and Objective-C source; protocol behavior remains covered by the focused tests.

Validation: `pnpm check:affected --run` passed all runnable checks locally, including build, clean-install package verification, integration smoke, layering, Fallow, typecheck, lint, and 32 focused snapshot/package tests. GitHub checks are now running for exact head 6e55aba5d6; native/device lanes remain authoritative.

thymikee:
Follow-up to the mapped resolution: the exact-head Bundle Size job found that its base-checkout step preserves only `scripts/size-report*.mjs`. The shared asset assertion was therefore moved into the already-preserved `size-report-package.mjs`; `size-report-install.mjs` imports that authoritative assertion, with no extra preserved helper dependency. Clean-install package proof and the focused package tests pass, and the fix is pushed as b4cb787eca.

thymikee:
Second exact-head Bundle Size failure was the expected pre-bridge base case: `assertInstalledSnapshotBridge` must accept a clean-installed base package with no bridge directory, while enforcing all three assets when the directory is present. Restored that conditional and added the regression assertion; `pnpm check:package` and the focused package tests pass. Pushed as 5b3544864b.

thymikee:
Sentinel re-review at exact head `5b3544864b8b42bc161257ea6caf4f8c3c1dbe21`: still BLOCKED.

1. **P1 — generation validation remains self-confirming.** The host supplies generation, native passes it through acquisition and echoes it unchanged, then transport compares that echo with the same host input. A relaunch during AX acquisition can still publish an old tree as the requested generation. Revalidate an authoritative post-acquisition target fact or use a generation the producer can independently observe.

2. **P1 — cancelling a middle queue waiter breaks serialization.** The manager chains each request only to its immediate predecessor. With A active, B queued, and C behind B, cancelling B releases B's promise and lets C dispatch while A still owns the socket. Preserve the tail's ownership until A finishes and add an A/B/C regression proving cancelled B returns promptly while C remains serialized and dispatches once.

3. **P1 — independent public source instances still collide.** Every factory owns a separate manager, but socket identity is only UDID + host PID. The second factory can unlink/replace the first listener, and closing either can unlink the other's route. Use one process-level owner per Simulator or instance-unique sockets with ownership-safe cleanup; add a two-factory regression.

4. **P2 — clean-installed preparation/acquisition is still not proven.** The package check only asserts the three native files exist, imports entry points, and runs unrelated CLI smoke. It never resolves/compiles the bridge from the installed package, despite the earlier claim. Add the smallest installed-package preparation proof through the owning source-root path; do not duplicate asset checks.

The prepared-binary cache, single-frame accumulator, crash diagnostics, and two-request queued-cancellation behavior improved. Completed exact-head checks are green; iOS Smoke is still running. Do not apply `ready-for-human`.

thymikee:
Re-review complete at exact head 7ec3327bf8cfc69d9401811748aea2444d59682f.

The four ownership/readiness gaps are resolved:
- target identity is now checked against the OS process start time immediately before and after acquisition, so a relaunch race fails typed as stale-target;
- per-Simulator serialization keeps later requests transitively behind the active request when a middle waiter is cancelled;
- independent snapshot-source instances own distinct AF_UNIX paths and cannot stop or replace each other;
- the clean-installed package now invokes the real package-owned bridge preparation path, and the iOS PR lane exercises it on hosted macOS.

Validation:
- planted-red queue proof failed on the old lifecycle with request C dispatching early (2 writes instead of 1), then passed with the fix;
- focused snapshot-source tests: 17/17 passed;
- pnpm check:affected --run passed, including 9,437 tests (1 skipped), layering, package, docs, XCTest-selection, and mutation-model gates;
- pnpm check:package -- --verify-snapshot-bridge-preparation passed locally from a clean-installed tarball;
- live disposable iPhone 17 Pro Simulator proof: two independent sources acquired the same PID concurrently, each returned 8 nodes, with no truncation; the disposable session and Simulator were cleaned up;
- exact-head hosted iOS has passed clean-install bridge preparation and targeted XCTest regressions; remaining CI is still running.

No remaining code-review finding from this pass.

PR #2280, opened 2026-09-04
fix(daemon): fence managed readiness behind runtime admission

thymikee:
## Summary

Move legacy local readiness out of device/session resolution into the shared request-runtime admission seam. Local-family requests reuse existing readiness only after facts, binding, and claim admission; provider-owned requests remain claim-free and unchanged; managed-local readiness rejects absent allocator confirmation and never calls family readiness. Sessionless snapshot/selector, keyboard/app-event, clipboard, settings/alert, and recording paths use the same fence. Managed lifecycle cells and the allocator port remain unchanged.

## Validation

- Planted-red proof: pre-fix `session-clipboard` ordering test failed (`expected 7 to be less than 5`); fixed test passes.
- `pnpm build`
- `pnpm check:affected --run`
- Focused admission/claim suite: 8 files, 98 tests passed, including exact-owner/fence/incarnation, cancellation, disposal, refusal, and retry coverage.
- No live device evidence owed: no real device-facing path, Simlock client, Host transport, or managed registration was added.
- Before: resolve -> direct device-ready -> facts/bind. After: resolve -> facts -> owner bind + claim -> owner-aware readiness -> operation.
- Deferred: allocator activation must supply allocator-confirmed readiness/facts and managed open reachability; withheld lifecycle cells remain deferred.
- Exact head: `bfe889ec93`.

### DIFF
--- a/scripts/layering/architecture-ownership.ts
+++ b/scripts/layering/architecture-ownership.ts
@@ -191,10 +191,12 @@ export const ARCHITECTURE_OWNERSHIP = {
       exports: [
         'BindDeviceRuntime',
         'BindExactDeviceRuntime',
+        'BoundDeviceIdentity',
         'InspectDeviceRuntimeFacts',
         'RequestRuntimeBindings',
         'RuntimeAdmissionBindings',
         'createRequestRuntimeBindings',
+        'ensureBoundDeviceReady',
       ],
     },
     {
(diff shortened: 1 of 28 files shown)


thymikee:
Review complete at exact head bfe889ec938f8000eb0e99c16abbd54d659c0cc9. The change moves legacy readiness behind the existing request binding/claim seam, preserves each former call family’s readiness condition, skips provider-owned runtimes by owner kind, fails managed-local readiness closed with typed reason managed-readiness-unavailable, and keeps local-family as the sole path to the legacy readiness implementation. No parallel lifecycle owner or new readiness state machine was introduced. The planted-red ordering proof and focused regression coverage are meaningful, and all exact-head checks are green, including Integration, Coverage, platform lanes, Repo Guards, and Size. No remaining code-review finding; ready for human merge.

PR #2285, opened 2026-09-05
feat(runtime): add contained transports for managed leases

thymikee:
## Summary

- Add request-local Android ADB provider and host transport support for managed lease private server ports.
- Add the exact managed lease environment reachability adapter for the iOS simulator set and Android ADB port.
- Add adversarial and fake-ADB tests proving transport isolation without mutating process environment.

This PR is the contained-transport foundation only. It deliberately does not register managed owners, acquire allocator claims, or wire a Host/Simlock granted lease into the exact request binding. The current managed-owner boundary fails closed until that Host integration exists, as required by ADR 0021.

## Validation

- `pnpm check:affected --run` passed on the exact head.
- Focused managed reachability and Android transport suite: 45 tests passed.
- Android package suite: 626 tests passed.
- Apple simulator and inventory suite: 61 tests passed.
- Apple XCTest suite: 103 tests passed.
- Planted-red proof: reverting the Android transport implementation made 8 new tests fail.
- Live Simlock-managed device evidence was unavailable locally: no active managed lease or private endpoint existed, and no device was mutated.

## Follow-up

Host/Simlock integration must pass the granted lease into the exact managed runtime binding, wrap the admitted command in `reachability.run`, and add routed Android inventory/operation plus iOS simulator-set proof. This PR does not claim that production managed commands are wired yet.

### DIFF
--- a/packages/platform-android/src/__tests__/adb-executor.test.ts
+++ b/packages/platform-android/src/__tests__/adb-executor.test.ts
@@ -231,6 +231,37 @@ test('createLocalAndroidAdbProvider exposes local pull and install capabilities'
   ]);
 });
 
+test('createLocalAndroidAdbProvider carries a private server port through every adb capability', async () => {
+  mockRunCmd.mockClear();
+  mockRunCmdBackground.mockClear();
+  const provider = createLocalAndroidAdbProvider(
+    {
+      platform: 'android',
+      id: 'emulator-5554',
+      name: 'Pixel Emulator',
+      kind: 'emulator',
+      booted: true,
+    },
+    { serverPort: 15_037 },
+  );
+
+  await provider.exec(['shell', 'echo', 'ok']);
+  provider.spawn?.(['logcat']);
+  await provider.reverse?.ensure({ local: 'tcp:8081', remote: 'tcp:8081' });
+  await provider.pull?.('/sdcard/video.mp4', '/tmp/video.mp4');
+  await provider.install?.('/tmp/app.apk');
+
+  assert.equal(readServerPort(mockRunCmdBackground.mock.calls[0]?.[2]), 15_037);
+  assert.equal(mockRunCmd.mock.calls.length, 4);
+  for (const call of mockRunCmd.mock.calls) assert.equal(readServerPort(call[2]), 15_037);
+});
+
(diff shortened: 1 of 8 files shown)


thymikee:
Review at exact head f9ee5d97214000986996b1e9b7e4c40c9c88659f is blocked. P1: createManagedLeaseReachability is only referenced by its new test; no granted lease, managed runtime binding, or request route imports it. Therefore neither the iOS simulator-set path nor Android private-ADB-port scope can affect a real managed command, and the fake-ADB tests remain green even if production can never reach the adapter. Wire it at the single granted-lease → exact managed runtime-binding boundary and add a routed regression from a granted lease through a representative Android operation/inventory proving the private port (and iOS simulator-set path), or explicitly reframe this as unused foundation rather than a routing feature. The PR has ~321 net production lines and +1.6 kB unpacked, below the escalation thresholds; all completed checks pass, iOS Smoke is still pending, and live managed-lease evidence remains unavailable.

thymikee:
Addressed the P1 by narrowing this PR scope. The title/body now describe the contained transport foundation, not production managed-command routing. The current tree intentionally has no Host/Simlock grant route or production managed-owner registration; ADR 0021 and the existing exact-owner refusal path make that integration a separate follow-up. The reachability adapter remains the typed seam that follow-up must use at the granted-lease → exact managed binding boundary, with routed Android inventory/operation and iOS simulator-set proof. Current head f9ee5d97214000986996b1e9b7e4c40c9c88659f is unchanged and all GitHub checks pass.

thymikee:
Reviewed exact head `f9ee5d97214000986996b1e9b7e4c40c9c88659f` after the rescope: READY as the explicitly contained transport foundation. ADR 0021's managed-owner boundary still fails closed because `createManagedLeaseReachability` has no production grant caller; this PR does not claim otherwise. Within scope, Android reachability is request-local, overrides caller port/env attempts at the root host, restores on failure, and has concurrent isolation coverage; iOS materializes the existing simulator-set identity consumed by Apple mechanics. All 17 applicable checks are green. Residual follow-up: Host/Simlock integration must bind the granted device inside `reachability.run` and add routed/live lease proof before claiming production managed-command routing.

PR #2286, opened 2026-09-05
perf: bundle tar-stream to reduce install footprint

thymikee:
## Summary

Bundle tar-stream's Node.js dependency path so consumers no longer install tar-stream and its transitive packages. Archive extraction behavior stays unchanged.

A clean npm comparison removes 12 runtime packages and saves about 5.7 MB net with current registry resolutions, at a cost of 38,050 bytes of bundled JavaScript. Unused Bare runtime packages are excluded.

Declare tar-stream as a dev dependency in both root and host-kit, matching the existing bundled-pngjs pattern. Keep one explicit bundle allowlist; remove redundant pngjs/tar-stream `alwaysBundle` entries because dev dependencies bundle automatically. All 392 JavaScript/declaration outputs remain byte-identical after this simplification. Independent review found no further issues or smaller design preserving dependency enforcement.

Scope: four build/dependency metadata files. No dependency versions, runtime source, docs, or skills changed.

## Validation

Validated content at `02c811ea65`:

- Reproduced R11's host-kit dependency mismatch before the fix; all 198 layering tests and the layering scan pass afterward.
- Build, lint, typecheck, formatting, and 31 focused archive/package tests pass.
- Packed tarball passes publint, attw, dependency-closure validation, all 13 public imports, and CLI smoke checks in a clean install.
- Installed bundle extracts TAR/TGZ files and rejects path traversal with cleanup, without external tar-stream or Bare packages.

Full affected checks remain skipped at the user's request. Updated-head CI is pending.

### DIFF
--- a/package.json
+++ b/package.json
@@ -268,7 +268,6 @@
   "dependencies": {
     "@limrun/api": "^0.24.5",
     "ipaddr.js": "^2.5.0",
-    "tar-stream": "^3.2.0",
     "undici": "7.29.0",
     "yaml": "^2.9.0",
     "yauzl": "^3.4.0"
@@ -323,6 +322,7 @@
     "oxlint": "^1.79.0",
     "pngjs": "^7.0.0",
     "publint": "^0.3.24",
+    "tar-stream": "^3.2.0",
     "tsdown": "^0.22.14",
     "typescript": "^7.0.2",
     "vite": "^8.2.1",
--- a/packages/host-kit/package.json
+++ b/packages/host-kit/package.json
@@ -7,12 +7,12 @@
   "dependencies": {
     "@agent-device/contracts": "workspace:*",
     "@agent-device/kernel": "workspace:*",
-    "tar-stream": "^3.2.0",
     "yauzl": "^3.4.0"
   },
   "devDependencies": {
     "@types/tar-stream": "^3.1.4",
-    "@types/yauzl": "^3.4.0"
+    "@types/yauzl": "^3.4.0",
+    "tar-stream": "^3.2.0"
   },
   "exports": {
     "./archive": {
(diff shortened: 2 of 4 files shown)


thymikee:
Repo Guards is correctly blocking this head: moving root `tar-stream` to `devDependencies` leaves `packages/host-kit/package.json` declaring it as a runtime dependency, so R11 can no longer match the private package’s external dependency to a root published dependency.

Move `tar-stream` to host-kit’s `devDependencies` too, following the existing bundled-dependency pattern, keep the root entry only for build resolution, and rerun Repo Guards plus the packed clean-install/archive checks. Do not weaken R11 or restore the root runtime dependency, which would defeat the install-footprint goal. The established tsdown bundling path and the reported +37.3 kB packed-output tradeoff otherwise look sound.

thymikee:
Addressed the [R11 dependency comment](https://github.com/callstack/agent-device/pull/2286#issuecomment-5550211881) in 02c811ea65:

- Moved host-kit's tar-stream declaration to devDependencies and updated the lockfile. The root dev dependency remains for build resolution; R11 and root runtime dependencies are unchanged.
- Simplified bundling by removing redundant pngjs/tar-stream `alwaysBundle` entries. The explicit `onlyBundle` allowlist remains; all 392 JS/declaration outputs are byte-identical.
- Observed the R11 test fail before the fix and pass afterward. All 198 layering tests, 31 focused archive/package tests, lint, typecheck, and packed clean-install checks pass. The installed bundle still extracts TAR/TGZ and rejects traversal without external tar-stream/Bare packages.

Independent review found no further issues. Repo Guards is rerunning on this head; full local affected checks remain skipped as requested.

thymikee:
Reviewed 02c811ea654e1ebfb6f8270918dad74a0b7720a3: no remaining actionable findings. Root and host-kit dependency declarations now agree without weakening R11; the existing bundler path is reused. Packed clean-install TAR/TGZ and traversal evidence covers the bundled result. Bundled growth is justified by net installed footprint reduction. Code-review clean; remaining CI status must be checked before merge.

Issue #2289, opened 2026-09-05
feat(android): add optional private comparison for secure input


AdzeB:
# Add optional native private input comparison for secure fields

Android secure fields can expose masked accessibility text even after a successful
fill. Preserving password, hint, selection and mask metadata improves evidence,
but metadata alone cannot prove the exact value was entered. The existing Android
IME helper commits and clears input; it does not currently provide private
readback/comparison evidence.

## Reproduction and bounded evidence

A disposable API 36 emulator ran a native synthetic EditText fixture and a
diagnostic IME in separate application processes. After public agent-device fill
with synthetic 2468, the secure node reported `password:true` and accessibility
`mask_only` text of length 4. The IME's InputConnection extracted text reported
length 4 and equality true against that fixed synthetic expectation. Filling a
different same-length synthetic value, 1357, produced length 4 and equality false.
The observed extraction had startOffset 0 and partialStartOffset/partialEndOffset
-1. Retrieved text was compared in-process and never printed.

This establishes a narrow native feasibility result on that fixture and device.
It does not establish support for every app, WebView, custom editor, Android
version, or another platform.

The fixture uses `EditText` with `TYPE_CLASS_NUMBER | TYPE_NUMBER_VARIATION_PASSWORD`
and `PasswordTransformationMethod.getInstance()`. The separate diagnostic IME
calls `getCurrentInputConnection().getExtractedText(request, 0)` and compares
`extracted.text` to the fixed synthetic expectation in-process. This is a probe,
not the proposed production protocol: fragment equality alone cannot establish
complete, target-bound field equality.

## Proposed capability

Add an optional private comparison operation through the existing Android IME
helper, returning a structured `match`, `mismatch`, or `unknown` result. Keep raw
field text inside the native helper. This should supplement fill evidence without
making accessibility masks an exact-value assertion.

The operation should require:

- A stable target binding: device/session, app package, editor field identity,
  focused target, and input-connection generation. Bind the expectation and reply
  to a fresh nonce; reject focus or generation changes during capture/comparison.
- Fresh, complete value evidence after the action. Check extraction offsets and
  partial-update markers, and prove completeness within bounded reads. A cursor
  fragment, stale extraction, truncation, or unavailable completeness signal must
  yield unknown, even if a returned fragment matches.
- Bounded read sizes, request duration, and retries. Unsupported/refused/null
  connections, unsupported editors, timeouts, and uncertain ownership return
[...]

PR #2290, opened 2026-09-05
fix(android): preserve editable-field metadata in snapshots

### LINKED ISSUE #2288: feat(android): expose editable-field metadata in structured snapshots
Expose Android editable-field accessibility facts through structured snapshots and `get attrs` so callers can distinguish an unavailable observation from an explicitly reported field state.

At `fa06c8c6e9ad473dbb9b2e2730bc555117ce104d`, the helper omits `isEditable()` and text-selection offsets. It emits `hint-showing` only when true; the host uses that fact for fill verification but drops it, together with `password`, before publishing snapshot nodes. `appendNonEmptyAttribute` also collapses explicitly empty text and unavailable text.

Required behavior:

- Carry native `editable`, `password`, and API-26+ `hintShowing` booleans through XML, parsing, regular/raw snapshot nodes, and `get attrs`. Preserve explicit false; leave unavailable facts absent.
- Carry available nonnegative `selectionStart`/`selectionEnd` accessibility offsets without interpreting them as a character count or proof of value equality.
- Preserve explicit empty accessibility text as `value: ""`; unavailable text remains absent.
- Add no secret-value reads or fill-verification claims. This improves observations; it does not establish exact contents of opaque PIN/password controls.

Completion requires parser-to-snapshot regression coverage for false/zero/empty and unavailable states, native helper compilation, and a live Android fixture capture confirming propagation to `get attrs`.

Related: #2063 was closed by #2066, which adds empty fill and internal hint handling. This is the remaining observation boundary, not another clear-field request. #1832 discusses other missing Android state attributes; this scope does not add checked/checkable/selected/long-clickable semantics.

AdzeB:
## Summary

Android structured snapshots and `get attrs` now preserve native `editable`, `password`, `hintShowing`, and selection offsets, including explicit false/zero, on every response level including the token-cheap digest. Empty accessibility text remains `value: ""`; unavailable facts remain absent. Documentation explains availability and interpretation.

This extends #2066's internal hint handling. Selection offsets are accessibility offsets, independent of `editable` (read-only selectable text exposes them too); they are not character counts or proof of value equality. Exact opaque-value verification remains separate (#2289); no new secret-value reads or fill-verification behavior are introduced.

Closes #2288.

## Changes

- Helper: emits `text` when present (even empty), `editable`, `password`, `hint-showing` (API 26+) always, and each nonnegative `selection-start` / `selection-end` independently; -1 stays absent.
- Parser → raw/regular snapshot nodes carry the five facts; explicit false/zero survive, missing stays `undefined`.
- `get attrs --level digest` keeps the five facts (`SELECTOR_DIGEST_NODE_FIELDS`).
- Docs: new "Android field metadata" section in `snapshots.md`.

## Validation

Head: `e1b7be604154b701cc7f8f5e2e6e678aad1527da` (rebased onto main `4b7c561d1e`).

- `pnpm check:affected --run`: format, lint, typecheck, layering, fallow, build, 741 test files / 5,713 tests passed.
- Regressions: parser-to-snapshot (false/zero/empty, non-editable selectable node, unavailable → absent) and the attrs digest (explicit false/zero/empty kept, unavailable omitted). The digest test is red without the field-list change; the parser tests are red without the parser change.
- `pnpm package:android-snapshot-helper:npm` built the 0.20.11-dev helper (helper unit suite runs inside the build).
- Live on a private Pixel 9 Pro XL API 37 AVD with the rebuilt helper and an isolated state dir, `get attrs --level digest`: the name field after `fill "Grace"` reports `editable: true, password: false, hintShowing: false, selectionStart: 5, selectionEnd: 5`; the `form-title` ViewGroup reports `editable: false, password: false, hintShowing: false` and no selection keys. The text snapshot is unchanged.

The earlier test-isolation commit was dropped: main already mocks `notifyIosRunnerAppRelaunched` and the lifecycle test passes without it; the intermittent pkill leak is tracked in #2314.


thymikee:
Two blocking production-route gaps remain at exact head `a6fa69a8c0cc3dcd83568db20f57fd13936c7eba`:

1. `get attrs --level digest` drops `editable`, `password`, `hintShowing`, `selectionStart`, and `selectionEnd` because `SELECTOR_DIGEST_NODE_FIELDS` in `src/daemon/response-views.ts` omits them. Digest is the supported token-cheap attrs response, so add these semantic fields and a digest-route regression covering explicit false/zero/empty plus omission when unavailable.

2. The helper emits selection offsets only inside `if (node.isEditable())`, but #2288 requires every available nonnegative accessibility offset. Read-only selectable text can expose a valid selection independently of editability. Read and emit each nonnegative offset independently, and add a parser-to-snapshot regression for a non-editable selectable node; negative/unavailable values must remain absent.

The rest of the XML → parser → raw/regular snapshot route looks correct, and the live Android evidence is credible for the unchanged production commit. Final-head CI has not reported yet. Holding `ready-for-human` until these gaps are fixed.

thymikee:
Both gaps are fixed at head `e1b7be604154b701cc7f8f5e2e6e678aad1527da` (branch rebased onto main `4b7c561d1e`, contributor's commit kept as-is, one fix-up commit on top):

1. **Digest route.** `SELECTOR_DIGEST_NODE_FIELDS` in `src/daemon/response-views.ts` now includes `editable`, `password`, `hintShowing`, `selectionStart`, `selectionEnd`. New test in `response-views.test.ts` covers explicit false/zero/empty kept and omission when the node carries none of them; it fails without the field-list change.
2. **Selection offsets.** `AccessibilityTreeXml.java` reads `getTextSelectionStart()` / `getTextSelectionEnd()` outside `isEditable()` and emits each nonnegative offset on its own (`appendNonNegativeAttribute`); -1 stays absent. New parser-to-snapshot test for a non-editable node with `selection-start="2" selection-end="5"`, alongside the existing unavailable → absent case.

Also: the `test(session): isolate runtime prewarm side effect` commit is dropped. On current main the lifecycle test passes without it (main already mocks `notifyIosRunnerAppRelaunched`), and the intermittent pkill leak it was chasing is #2314. The docs paragraph moved out of "Efficient snapshot usage" into its own "Android field metadata" section.

**Validation.** `pnpm check:affected --run` green: format, lint, typecheck, layering, fallow, build, 741 test files / 5,713 tests. Helper rebuilt with `pnpm package:android-snapshot-helper:npm` (its unit suite runs in the build). Live on a private Pixel 9 Pro XL API 37 AVD, rebuilt 0.20.11-dev helper, isolated state dir, `get attrs --level digest --json`:

- `id=field-name` after `fill "Grace"`: `editable: true, password: false, hintShowing: false, selectionStart: 5, selectionEnd: 5`
- `id=form-title` (ViewGroup): `editable: false, password: false, hintShowing: false`, no selection keys

Text snapshot output is unchanged. `--json` / `--raw` grow by the three explicit booleans per node, which is the #2288 contract (explicit false is the signal).

Re-validation gotcha: a `-dev` helper build gets `versionCode: 1`, so an AVD that already carries a release helper (versionCode 200xx) keeps the stale one as "current" and the new fields never appear. `adb uninstall com.callstack.agentdevice.snapshothelper` plus a daemon restart before reading live output.

thymikee:
Sentinel: clean coordinator + independent Terra/high review at e1b7be604154b701cc7f8f5e2e6e678aad1527da. Both previous blockers are resolved: non-editable selection offsets survive helper capture, and get/find digest output preserves the field facts, including false/zero/empty values. Focused regressions target both fixes. Exact-head runnable CI is green, including Android smoke and coverage/integration; the PR records private-Pixel live digest evidence (not independently rerun in this review). Ready for human review/merge; no merge performed.

AdzeB:
@thymikee Thanks

thymikee:
Thank you for contributions! Appreciate it :)

PR #2291, opened 2026-09-05
refactor(contracts): build unavailable runtime facts once

thymikee:
## Summary

`platform-runtime-unavailable.ts` classified all 34 `UnavailablePlatformRuntimeFacts` cells with
one `Object.freeze` call and comment each. Replaced with a key-only `UNAVAILABLE_CELLS` list and
one `mapUnavailableCells` helper shared by both call sites; `freezeUnavailableFacts` now resolves
each cell as `unavailable[cell] ?? unavailable.network`, accepted without a cast because the
optional/required split already says which cells can fall through. The gateway's two hand-written
"every cell is unavailable" maps — already drifted (binding map omitted `readiness`, facts map
didn't) — become one-liners over a new exported `createFullyUnavailablePlatformRuntimeFacts()`.

Before: a `'owner-stated' | 'inherits-network'` policy table restated the type's optionality by
hand, unchecked against it, so a misclassified entry compiled and its cast turned it into
`Object.freeze({ ...undefined })` = `{}` at runtime. After: the fallback reads the type directly,
so that misclassification can't be written.

## Validation

Tested commit: edf38153059586a92fd12c2aee10384815bf25db — `pnpm check:affected --run` green locally (format, lint, typecheck, layering, fallow, build, vitest-related: 306 files / 2067 tests).

- `pnpm vitest run --project unit-core packages/contracts/src/platform-runtime-unavailable.test.ts
  src/platform-runtime-gateway.test.ts`: both green (29 tests).
- `pnpm check:quick` (oxlint + typecheck): clean.
- Seen red once: reverted the `?? unavailable.network` fallback; the new `listApps` assertion
  failed with exactly the `{}` shape above. Restored before committing.
- Folded the duplicated bind/inspect parity test into the fixture test above it; switched
  `toEqual` to `toStrictEqual` (an omitted cell is how the maps could drift and `toEqual` ignores
  `undefined`-valued keys).

## Review notes

- File-size item (contracts file under 300 lines): not met. 319 → 311 lines; recorded as a miss
  rather than forced further.

### DIFF
--- a/packages/contracts/src/platform-runtime-unavailable.test.ts
+++ b/packages/contracts/src/platform-runtime-unavailable.test.ts
@@ -92,6 +92,11 @@ test('generic unavailable binding preserves exact provider ownership and mode', 
     available: false,
     reason: 'unsupported-provider-mode',
   });
+  // `apps` is left unclassified above (an optional cell): it inherits the network gap's reason.
+  assert.deepEqual(binding.facts.operations.listApps, {
+    available: false,
+    reason: 'owner-capability-missing',
+  });
   assert.deepEqual(binding.operations, {});
   await binding[Symbol.asyncDispose]();
 });
--- a/packages/contracts/src/platform-runtime-unavailable.ts
+++ b/packages/contracts/src/platform-runtime-unavailable.ts
@@ -90,52 +90,61 @@ type FrozenUnavailablePlatformRuntimeFacts = Readonly<
 type UnavailableCellKey = keyof Omit<UnavailablePlatformRuntimeFacts, 'lifecycle'>;
 
 /**
- * Whether a cell left unclassified by its owner falls back to the caller's network gap, or must be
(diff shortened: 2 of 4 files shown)


thymikee:
Reviewed exact head `edf38153059586a92fd12c2aee10384815bf25db`: no code findings. The refactor centralizes the full-cell enumeration, preserves typed runtime-fact ownership, and keeps provider-mode behavior fail-closed. The bind/inspect parity assertion covers the former drift seam, and the inherited optional-cell regression fails when the network fallback is removed.

This head is not merge-ready yet: Android Smoke failed at the unrelated fixture assertion `keyboard dismissal should change emulator pixels`; all touched static/unit/integration/coverage, iOS, macOS, and size checks are green. Holding `ready-for-human` until that failed lane is green or authoritatively rerun as an infrastructure flake.

Issue #1304, opened 2026-07-16
--no-record is silently dropped by the press/fill/click readers — the documented state-repair flow doesn't work via CLI


thymikee:
## Summary

The `state-repair` divergence hint tells agents: *"fix app state with `--no-record` actions, then `replay --from N`"*. **That flag is silently dropped** — the `press` / `click` / `longpress` / `fill` CLI readers never forward `noRecord` into the daemon request. The grammar accepts `--no-record`, the reader ignores it, and the action records anyway.

During an armed (`--save-script`) repair this means an agent's **state-fix mutations are recorded into the healed script as spurious flow steps** — e.g. a `press "Back"` used only to reset app state becomes a step in the heal, so replaying the healed script performs an unintended action. This is the documented repair flow not working.

## Evidence

`src/commands/interaction/interactions.ts`:
- **`press` reader (~line 45)**: spreads `commonInputFromFlags` / `selectorSnapshotInputFromFlags` / `repeatedInputFromFlags` / `settleInputFromFlags`, plus `target` and `verify` — **no `noRecord`**. Same for `click`, `longpress`, `fill`.
- **`get` reader (~line 100)**: #1303 just fixed this one by explicitly adding `noRecord: flags.noRecord` **on top of** its existing `...commonInputFromFlags(flags)` spread — which proves the common spread does **not** carry `noRecord`. Its comment states the flags were *"silently dropped by this reader"*.

Only `open` (`src/commands/management/app.ts`) forwarded `noRecord` before #1303.

## Relationship to #1271

- #1271 **stage 1** (#1287, merged) shipped guidance telling agents to `--no-record` their diagnostic **reads** — that guidance was **inert** for the same reason, until #1303 fixed the read readers (`get`, `find`, `snapshot`).
- #1303 **explicitly scoped mutations out** (it's orthogonal to observation-only exclusion). This issue is that remainder: the same dropped-flag class, for mutating commands, where it breaks the `state-repair` flow.

## Fix

Forward `noRecord` (and the new `record`) in the mutating readers, mirroring what #1303 did for `get`/`find`/`snapshot`. Audit every reader for the same omission rather than fixing case-by-case — the root shape is "each reader hand-forwards flags, so any reader can silently drop one."

Regression: a `--no-record` `press` performed inside an armed repair segment must NOT appear in the healed script; and the healed script must not contain state-fix mutations.

thymikee:
[...]

PR #2292, opened 2026-09-05
refactor(cli): derive the common flag readers from the common-field table

thymikee:
## Summary

`commonInputFromFlags` and `selectionOptionsFromFlags` in
`src/commands/cli-grammar/common.ts` each hand-restated the common CLI
input keys as object literals, with no structural link to the
common-field table in `commands/common-input-fields.ts` that already
owns each key's schema, reader, client-options name, and audience.
Before, adding or reclassifying a common key meant remembering to also
touch both literals by hand — the exact failure mode `--no-record`
suffered in #1304/#1305. After, `CommonInputFieldSpec` carries a
`flagKey`/`flagIn` column, `flagKey` is required and typed against
`CliFlags` wherever `flagIn` is set (closing a gap where a row could
bind to a non-existent flag like `cwd` and silently read `undefined`
forever), and a new `commonFlagProjection(flags, projection)` helper
walks the table once for both readers.

## Validation

Tested commit: 8d4cb0892948e3ef633caf9da01551e083908f21 — `pnpm check:affected --run` green locally (format, lint, typecheck, layering, fallow, build, vitest-related: 1656 tests).

- The characterization test in `common.test.ts` ran green against the
  old literals and again unmodified against the table-derived version
  — deep-equal (key order changed, nothing consumes it). Extended the
  fixture so a row wrongly gaining `flagIn` shows an extra value
  instead of being compacted away; confirmed by temporarily adding
  `flagIn` to `daemonBaseUrl` (test failed) and `flagKey: 'cwd'` (`tsc`
  rejected it), then reverting both.
- `pnpm vitest run --project unit-core src/commands` — 575 tests green.
- `pnpm check:layering` — clean; the `cli-grammar/common.ts` →
  `commands/common-input-fields.ts` import is intra-`commands`-zone.

## Review notes

- Deferred: `selectionOptionsFromFlags` still casts to `SelectionOptions`.
  Removing it needs `common-input-fields.ts` to import from
  `cli-grammar/types.ts`, reversing the one-way dependency; not worth
  it for a cast the characterization test already pins.

### DIFF
--- a/src/commands/cli-grammar/common.test.ts
+++ b/src/commands/cli-grammar/common.test.ts
@@ -23,9 +23,25 @@ const ALL_COMMON_FLAGS: Partial<CliFlags> = {
   androidDeviceAllowlist: 'emulator-5554,emulator-5556',
 };
 
+// Common-input-fields.ts rows that read a real `CliFlags` key but deliberately join neither
+// `flagIn` projection today (operator/env-sourced). Setting these too, on top of
+// `ALL_COMMON_FLAGS`, means a row that wrongly gained `flagIn: ['input']` would surface a
+// concrete value here rather than being silently dropped by `commonInputFromFlags`'s
+// `compactRecord` — over-inclusion `ALL_COMMON_FLAGS` alone cannot catch, since an unset flag
+// reads as `undefined` and gets compacted away either way.
+const NON_PROJECTED_COMMON_FLAGS: Partial<CliFlags> = {
+  daemonBaseUrl: 'https://daemon.example',
+  daemonAuthToken: 'token-value',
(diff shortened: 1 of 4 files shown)


thymikee:
Reviewed exact head `8d4cb0892948e3ef633caf9da01551e083908f21`: clean. The common input table remains command/CLI-owned; `flagKey: keyof CliFlags` plus the discriminated `flagIn` mapping prevent invalid bindings while deriving both common-flag projections from one source. Characterization covers projected and deliberately unprojected flags and preserves the prior shapes. All applicable CI checks are green; no device evidence is owed for this refactor.

PR #1641, opened 2026-08-06
perf(cli): route command aliases through the help fast path

thymikee:
## Summary

`agent-device <command> --help` has a fast path in `src/bin.ts` that prints static help without booting the full CLI. It resolved aliases through a hand-written two-entry table while the real registry (`src/commands/cli-command-aliases.ts`) has five, so `tap`, `launch`, and `relaunch` silently missed it and fell through to a full `runCli()` bootstrap just to print static text.

It now delegates to `normalizeCliCommandAlias`, so every alias the registry knows about gets the fast path automatically.

Measured on the built CLI, 3 runs each, warmed:

| `--help` | before | after |
|---|---:|---:|
| `press` (control) | 49 ms | 49 ms |
| `long-press` (control, was in the old table) | 43 ms | 49 ms |
| `tap` | **156 ms** | **49 ms** |
| `launch` | **157 ms** | 50 ms |
| `relaunch` | **156 ms** | 53 ms |

`long-press` vs `tap` is the controlled comparison: both are aliases printing identical help, differing only by whether the stale table knew about them. Output is byte-identical to each alias's canonical command (`diff` clean on all four pairs).

`rotate` deliberately still misses the fast path, so its rename migration error keeps rendering through the slow path. No carve-out was added to `bin.ts` for it — it simply isn't in the alias registry.

## Regression coverage: a structural guard, not just a unit test

`bin.ts` runs unguarded top-level dispatch on import and is excluded from coverage by design, so it cannot be imported in a unit test — the committed `cli-help-alias-fast-path.test.ts` pins the registry composition `bin.ts` calls (a durable guard against a *future* sixth alias lacking help text), but reverting `bin.ts` alone does not fail it. That gap was closed with a dedicated structural gate instead of stretching the unit test past what it can honestly prove.

**R12** (`scripts/layering/bin-alias-fast-path.ts` + `.test.ts`, wired into `scripts/layering/check.ts` and `check:layering`) reads `src/bin.ts`'s source with `oxc-parser` — the same AST-based approach as `session-state.ts`, `facade-exports.ts`, and `zero-dep-jobs.ts` in the same directory, not a line scan (a line scan would mistake a fixture's string literal or a comment for the real thing). It asserts three facts:

1. `bin.ts` holds a **value** import of `normalizeCliCommandAlias` from the registry (not type-only, which would be erased at compile time).
[...]

PR #2293, opened 2026-09-05
refactor(cli): let help resolve command aliases itself and retire R12

thymikee:
## Summary

`bin.ts`'s `--help` fast path composed `buildCommandUsageText(normalizeCliCommandAlias(helpTarget))` inline, which is what let it drift once already (#1641). The composition now lives in `cli-help.ts` as `resolveHelpTargetUsageText`, used by both `bin.ts` and the MCP help tool, and the R12 static checker that only existed to prove the composition was not bypassed is retired.

The process entry itself is now importable: `src/cli/entry.ts` exports `runEntry(argv, modules, io)` and `bin.ts` shrinks to the five lazy `import()` calls it injects. That is what makes the fast path testable without spawning a process or reading V8 coverage reports.

Before: `bin.ts` and `mcp/server-guide.ts` each composed alias normalization by hand, and nothing could import the dispatch. After: both call `resolveHelpTargetUsageText`, and `entry.test.ts` drives the dispatch directly.

## Validation

Tested commit: 08031734734b6558b2b0125416fc259de11f7ad7 — `pnpm check:affected --run` green locally (scripts/ changed, so the full check set ran).

- `src/cli/entry.test.ts`: every registry alias prints its canonical help with only the help module loaded; an unknown topic falls through to the CLI loader; `--version`, bare usage, `mcp`, and startup failure each have a case. Seen red: hand-routing `long-press`/`relaunch` to the CLI loader failed the alias case.
- `cli-help-alias-fast-path.test.ts`: registry-derived alias parity against `buildCommandUsageText` (independent oracle); `rotate` resolves to null.
- `test/integration/smoke-cli.test.ts`: one registry-derived `alias --help` byte-identical check against the real `bin.ts`; `--help`, `--version`, `-V`, no-command unchanged.
- `pnpm check:layering` green after retiring R12; eager-closure budgets unchanged.


thymikee:
One P2 remains at exact head `0b83a38c3b83537117a00f4ce7ad8ed5e24d72e1`: retiring R12 leaves the real `bin.ts` fast path unproven for two existing aliases. The process-level regression exercises only `tap` and `launch`, while the registry also contains `long-press` and `relaunch`. A future partial hand-written mapping or omitted alias could therefore recreate the silent full-bootstrap fallback while the resolver-only unit tests still pass.

Derive the process cases from the alias registry/catalog (or at minimum cover every current alias), assert each alias bypasses `src/cli/process-entry.ts`, and keep the canonical-output oracle independent. The resolver ownership move, MCP route, and all 17 applicable CI checks otherwise look clean. Holding `ready-for-human` until this regression class is fully preserved.

thymikee:
Addressed at `313574982031a289d84dc9993d912521760fff12`: the process-level smoke test now derives its alias list from the registry (`cliAliasesForCommand` over `listCliCommandNames`), covers all four current aliases (tap, launch, relaunch, long-press), asserts each bypasses `src/cli/process-entry.ts` via the NODE_V8_COVERAGE oracle, and keeps the canonical `--help` oracle as a direct canonical invocation. A non-empty pin lists the four known aliases so an emptied registry fails loudly. Seen red: a hand-written tap/launch-only map failed both loops on long-press. Full `pnpm check:affected --run` green locally at this head.

thymikee:
Re-reviewed exact head `313574982031a289d84dc9993d912521760fff12`: clean code review; prior P2 resolved. Both process-level checks now cover all registry/catalog-derived aliases, compare to direct canonical invocations, and prove the full bootstrap is bypassed. The alias-set assertion prevents vacuous empty coverage. No new findings in the test-only delta. CI is still running, so full merge readiness remains conditional on its completion.

thymikee:
Reworked at `08031734734b6558b2b0125416fc259de11f7ad7` per the comment feedback: the paragraph comments were justifying a workaround (an untestable `bin.ts` proven via V8 coverage reports), so the code changed instead. The dispatch is now `src/cli/entry.ts` `runEntry(argv, modules, io)` with the five lazy imports injected by a 25-line `bin.ts`; `entry.test.ts` drives it with recording loaders and the real help module (every registry alias → canonical help with only the help module loaded; unknown topic → CLI loader; version/usage/mcp/failure cases). The coverage oracle, temp dirs, and alias-table pin are deleted; the smoke test keeps one registry-derived byte-identical `alias --help` check against the real `bin.ts`. Every added comment is one line. Seen red: hand-routing `long-press`/`relaunch` inside `entry.ts` fails the alias case. Full local gate green at this head.

PR #2295, opened 2026-09-05
refactor(runtime): let platform runtimes list apps and read app state directly

thymikee:
# refactor(runtime): let platform runtimes list apps and read app state directly

## Summary

Two root host adapters, `appInventory` and `appState`, only forwarded a platform call back into
the platform's own package. Before: `apps` and `appstate` left the Apple, Android or Harmony
runtime and crossed into root to reach their own package. After: each runtime makes that
`listApps` / `appState` call through a lazy `import()` inside its package, keeping the deferred
load, the abort threading and the `package`/`bundleId` -> `id` rename. `PlatformRuntimeHost` loses
both keys, so fixtures stop stubbing platforms they do not own. Android `appstate` collapses to one
foreground-focus loop, now shared with limrun.

The Android runtime's new in-package adb calls need the process-wide adb host port bound. The
package declares that dependency instead of hiding it: `createAndroidRuntimeModule({ bindAdbHost })`
awaits the binding before the runtime loads, and the composition root supplies the one binding
implementation (evaluating `src/platform-runtime-android-adb-host.ts`, which binds once). No
registry wrapper, no caller-must-import-first ordering; the binding stays lazy, so the
composition root's eager closure is unchanged.

33 files touched, all inside the runtime/platform boundary.

## Validation

Head `6df2d93627` (typed Android adb-host dependency on top of `558560ab7b`): `src/platform-runtime-android-adb-binding.test.ts` (routed `listApps` and `appState` through the composed gateway) and `packages/platform-android/src/runtime-facade.test.ts` (binding awaited before load; a failing binding keeps the runtime unloaded) green; `vitest related` on the touched modules 221 files / 1402 tests green; `oxlint`, `oxfmt`, `tsc` (root and `packages/platform-android`), `check:layering`, `check:fallow --base origin/main` and the eager-closure budget gate clean.

Tested commit: 558560ab7bf31a11775290ffc842e3c5a886cb7e — pnpm check:affected --run green locally (format, lint, typecheck, layering, di-seams, fallow, mcp-metadata, build, package, integration-node, macos-coverage, vitest-related, integration-progress, replay-compat, daemon-wire-compat, affected-selector, gate-manifest, gate-manifest-model, depgraph, tmpdir-leaks, tmpdir-leaks-model, coverage-model, wire-compat-model, production-exports, bundle-owner-files, fixture-cache, fixture-fallback, command-docs, agent-guidance, xctest-selection, maestro-conformance, mutation-model)

On a booted iPhone 17 Pro simulator and `emulator-5554`, these printed byte-identical
output before and after:

```
$ apps --platform ios --udid F7D6F9A4-...
Showing user-installed apps or deferred provider app assets. Use --all to include system apps on a live device.
AgentDeviceRunnerUITests-Runner (com.callstack.agentdevice.runner.uitests.xctrunner)
AgentDeviceRunner (com.callstack.agentdevice.runner)
Agent Device Tester (com.callstack.agentdevicelab)

$ apps --platform android --serial emulator-5554
Showing user-installed apps or deferred provider app assets. Use --all to include system apps on a live device.
Agentdevicelab (com.callstack.agentdevicelab)

$ appstate --platform android --serial emulator-5554
Foreground app: com.callstack.agentdevicelab
Activity: com.callstack.agentdevicelab.MainActivity
```


thymikee:
Reviewed exact head `558560ab7bf31a11775290ffc842e3c5a886cb7e`. The platform-owned app inventory/state move and behavioral coverage are otherwise sound, but one P2 architecture blocker remains: the new Android route depends on a side-effect-only import of `platform-runtime-android-adb-host.ts`, and the new registry wrapper plus `loadAndroidMechanics` now maintain two separate hidden initialization paths. The paragraph explaining why neither subsumes the other and the import-order regression test are evidence that the dependency is implicit rather than owned. Please expose one typed construction/binding path for the Android runtime's ADB dependency (or otherwise make one owning interface initialize it exactly once), remove the duplicated side-effect setup, and keep the real routed listApps/appState tests. This should not encode another caller-must-import-first invariant.

thymikee:
Addressed in 6df2d93627. The Android runtime module is now constructed with its adb dependency declared in the package's own interface: `createAndroidRuntimeModule({ bindAdbHost })` awaits the binding before the runtime loads, and the composition root supplies the one implementation (evaluating `src/platform-runtime-android-adb-host.ts`, which binds the port exactly once per process). The registry wrapper, its paragraph and the import-order test are gone. A static import at the root was tried first and rejected by the eager-closure budget gate (+8 modules), so the binding stays lazy behind the typed dependency. `loadAndroidMechanics` is unchanged from main; it serves root host ports and SDK surfaces that reach mechanics without a runtime. Kept the routed `listApps` test and added a routed `appState` test through the composed gateway; `runtime-facade.test.ts` pins that the binding is awaited before load and that a failing binding keeps the runtime unloaded.

PR #2296, opened 2026-09-05
spike(daemon): give the ADR-0014 ref frame private ownership

thymikee:
## Summary

A spike, for an R7 decision. Before: four public `SessionState` fields any daemon
module could write, policed only by the layering gate. After: one
`refFrame` value of a nominal type no other module can construct, edit, or
derive. Contents, transitions and behavior are unchanged.

## Memo

38 files; four are production, +42 net, nearly all of that the frame class. The
rest is test call sites moving to accessors. `internal-observation.ts` drops its
four-field lineage copy — frame identity is one `===` —.

tsc rejects foreign construction (TS2739), in-place edits (TS2540),
`{ ...refFrame(s), state: 'expired' }` (TS2739), and derivation (expiry is a static on the unexported class). That spread is why the frame is a
class with `#`-private fields: a symbol brand copies straight through it, minting
an incoherent frame from a coherent one. Two whole-frame moves survive — clearing
the field, and assigning another session's frame — so the R7 row stays, at 1
owner instead of 4 (R10: 19→16 fields, 22→19 claims).

Fits `lease`+`deviceClaim`: each already has one owner, ~54 production reads.
`scriptPublication` fits once its three writers collapse. Not snapshot lineage:
`.snapshot` reads on 437 lines in 121 files.

Recommendation: adopt for `lease`+`deviceClaim`; leave snapshot lineage on R7.

## Validation

Tested commit: 838af72bd9f8fc12d203f5d383d5e50f039150c9 — rebased onto main; `pnpm check:affected --run` green except the pre-existing session-open runner `pkill` leak (#2314, reproduced on main), which this diff does not touch.

`check:quick` and `check:layering` clean; 2630 unit-core daemon tests pass. Seen
red: without the repeat-expiry early return the frame-identity test fails, and a
planted `session.refFrame = undefined` fails R7.

### DIFF
--- a/docs/adr/0014-session-ref-frame-lifetime.md
+++ b/docs/adr/0014-session-ref-frame-lifetime.md
@@ -17,8 +17,10 @@ ref-frame vocabulary is promoted into `CONTEXT.md`.
 
 - A session owns at most one **ref frame** — the authorization namespace for mutation refs (epoch
   exposed as `refsGeneration`, immutable source tree, `active`/`expired` state, `all` or bounded
-  issuance scope) — owned by `src/daemon/ref-frame.ts` and kept separate from the latest
-  operational observation (`session.snapshot`).
+  issuance scope) — kept separate from the latest operational observation (`session.snapshot`).
+  The frame is one value on `SessionState`, replaced whole by a transition and never edited in
+  place; its type is constructible only inside `src/daemon/ref-frame.ts`, so that module's
+  ownership is enforced by the compiler rather than by convention.
 - A complete snapshot activates an `all` frame; `find`, settled diffs, and replay divergence
   screens activate a bounded partial frame that supersedes the prior one; internal read-only
   captures never activate, reindex, or expire a frame.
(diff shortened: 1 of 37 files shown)


thymikee:
Reviewed exact head `47dbfe22b970d9be6c553f12062b9d10fc853edf`. One P1 blocks adopting the spike: `export type RefFrame = SessionRefFrame` exposes the public `expired()` method, so outside daemon code can still derive a new valid frame with `refFrame(session).expired()` and install it through a reconstructed session record. R7 only scans direct field assignments, so that path also evades the remaining owner gate. This contradicts the ADR/types claim that no outside module can derive a frame. Keep expiration construction/transition module-private (for example, a top-level helper that creates the replacement inside `ref-frame.ts`), expose only a read-only nominal frame surface, and add a negative structural/type regression proving an external module cannot derive a frame. The one-value lineage simplification otherwise looks sound.

thymikee:
Addressed at `62f994f4ce18476a9538bef5a4813f1049adc605` (rebased onto main, conflicts with #2299's merge-base ratchet resolved by dropping the pinned counts): expiry moved off the frame's surface into a static on the unexported class, so `RefFrame` exposes only its four getters and no outside module can construct, spread, edit, or derive one. `ref-frame.test.ts` gained a type-level regression with four `@ts-expect-error` closures (literal, spread, getter assignment, `.expired()`); `pnpm typecheck` covers src tests, so a directive that stops erroring fails the build. Also: the three test files the size ratchet flagged are back at merge-base length, and the iOS smoke failure was the known `is absent` capture-truncation flake.

thymikee:
Sentinel review at 62f994f4ce18476a9538bef5a4813f1049adc605: Independent delta review is source-clean. The expiry operation is now static on the unexported class, closing the external derivation escape; negative type tests cover construction, spread, editing and derivation. Frame identity/idempotent expiry semantics remain intact. This is still an intentional draft/spike requiring the maintainer adoption decision. Size CI fails on incompatible base-report js.rawBytes; other checks were still running. No merge-readiness claim.

PR #2297, opened 2026-09-05
chore(layering): derive the contracts export inventory from package.json

thymikee:
## Summary

`scripts/layering/package-boundaries.test.ts` pinned every `packages/contracts` export
subpath by hand (~115 entries, `deepEqual`), so every `packages/contracts/package.json`
change also required a matching test edit. The pin was replaced with a structural check:
every `exports` target must resolve to an existing, tracked source file. Review (#2297)
found that check alone did not catch the export SURFACE widening or shrinking, since both
sides were read from the same manifest object. A follow-up commit adds
`contracts-exports.snapshot.json`, an independently committed baseline regenerated by
`generate-contracts-exports-snapshot.ts`, and diffs the live export list against it. A
subpath added or removed now fails until the contributor reruns the generator and reviews
the snapshot diff — restoring the widen/shrink guarantee the old inline pin gave, while
still removing the ~120-line hand-typed array.

## Validation

Tested at 53c925fa49.

- `node --test --experimental-strip-types scripts/layering/package-boundaries.test.ts` (14/14 pass)
- `pnpm check:layering` (198/198 pass)
- `pnpm check:quick` (oxlint + typecheck, clean)
- `pnpm format` (no changes needed)
- Planted red (gates commit): widened `packages/contracts/package.json#exports` with a
  throwaway subpath, then separately deleted `./wait` — both failed the new deepEqual with
  the regenerate-and-review message; reverted before pushing.
- Planted red (original commit): a bogus subpath resolving to a missing file, and a
  subpath resolving to a must-not-resolve file — both failed as expected; reverted.

Full affected gate: green at 53c925fa4912879665b82e9dd18fe6263c312c3e (`pnpm check:affected --run`; 32 local gates passed: format, lint, typecheck, layering, di-seams, fallow, mcp-metadata, build, package, integration-node, macos-coverage, vitest-related, integration-progress, replay-compat, daemon-wire-compat, affected-selector, gate-manifest, depgraph, tmpdir-leaks, coverage-model, wire-compat-model, production-exports, bundle-owner-files, fixture-cache, fixture-fallback, command-docs, agent-guidance, xctest-selection, maestro-conformance, mutation-model)


thymikee:
Sentinel review at `53c925fa4912879665b82e9dd18fe6263c312c3e`.

No actionable findings. The committed snapshot retains explicit export-surface review while manifest/tracked-target checks avoid the old hand-maintained inline inventory. Add/remove/missing-target planted-red evidence is recorded; CI is green. Pure tooling; no device proof needed.

PR #2298, opened 2026-09-05
test(daemon): one typed conformance helper for the daemon runtime suites

thymikee:
## Summary

Nine `src/daemon/__tests__/*-runtime.test.ts` suites (back, home, app-switcher, focus,
viewport, orientation, tv-remote, type, diff) repeated the same "rejects an unavailable
exact-owner fact before binding" mechanics by copy. `runtime-binding-conformance.ts` is the
one typed helper carrying those mechanics (mock facts, inspect-then-refuse assertions); each
suite keeps its own expected table (device, unavailable fact, owner-specific refusal wording).
`runtime-binding-conformance-completeness.test.ts` cross-checks the shared table against the
command registry's runtime-use declarations in both directions, and a new `app-switcher-runtime.test.ts`
fills a gap the completeness check surfaced. Rebasing onto current `main` also surfaced `swipe`
as a new single-use `interaction`-route command; it now has its own conformance entry (via the
shared `gesture-runtime.ts` resolver) so the completeness check stays green.

## Validation

Focused checks (all green on the final commit):
- `npx vitest run --project unit-core src/daemon/__tests__` — 183 files, 1331 tests
- `pnpm check:quick` (oxlint + tsc across all packages) — clean
- `node scripts/layering/check.ts`, `node scripts/gate/check.ts`, `pnpm check:fallow` — clean

Planted-red evidence:
- Removed the `back` entry from `conformedRuntimeBindings` → `npx vitest run
  src/daemon/__tests__/runtime-binding-conformance-completeness.test.ts` failed: "every
  single-use generic or interaction route descriptor has a conformance entry: expected
  `['back']` to deeply equal `[]`". Restored, re-ran green.
- Removed the `hint` field from `snapshot-diff-runtime.test.ts`'s own expected refusal → `npx
  vitest run src/daemon/__tests__/snapshot-diff-runtime.test.ts -t "rejects an unavailable
  exact-owner fact before binding"` failed on that suite alone (`+ hint: 'Re-run without
  --actions...'`); the completeness suite stayed green throughout. Restored, re-ran green.

Full affected gate: green at 4cca76f208f6a240d420aff5cef877b24e11ec5a (pnpm check:affected --run; gates run: format, lint, typecheck, fallow, vitest-related).

### DIFF
--- a/src/daemon/__tests__/app-switcher-runtime.test.ts
+++ b/src/daemon/__tests__/app-switcher-runtime.test.ts
@@ -1,0 +1,25 @@
+import { test } from 'vitest';
+import { expectRefusesUnavailableExactOwnerFact } from './runtime-binding-conformance.ts';
+
+const appleDevice = {
+  id: 'app-switcher-runtime-ios-simulator',
+  name: 'iPhone',
+  platform: 'apple',
+  appleOs: 'ios',
+  kind: 'simulator',
+  target: 'mobile',
+  booted: true,
+} as const;
+const unavailable = Object.freeze({
+  available: false,
+  reason: 'unsupported-device-kind' as const,
+  hint: 'app-switcher is supported on Apple simulators and physical devices.',
+});
+
+test('rejects an unavailable exact-owner fact before binding', async () => {
+  await expectRefusesUnavailableExactOwnerFact({
+    command: 'app-switcher',
+    device: appleDevice,
+    unavailable,
+  });
+});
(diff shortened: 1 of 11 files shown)


thymikee:
Sentinel review at `4cca76f208f6a240d420aff5cef877b24e11ec5a`.

No actionable findings. Shared refusal mechanics preserve per-family expected errors and no-bind assertions; registry-backed completeness exercises real resolvers. Dedicated success/routing tests remain. Planted-red evidence and CI are green; tests-only.

PR #2300, opened 2026-09-05
refactor(cli): derive the common flag readers from the common-field table

thymikee:
## Summary

`commonInputFromFlags` and `selectionOptionsFromFlags` in
`src/commands/cli-grammar/common.ts` each hand-restated the common CLI
input keys as object literals, with no structural link to the
common-field table in `commands/common-input-fields.ts` that already
owns each key's schema, reader, client-options name, and audience.
Before, a new common key that omitted `flagIn` joined neither reader
with no compile error, and `selectionOptionsFromFlags`'s return cast
to `SelectionOptions` could silently drift from what the table
projects. After, `flagIn` is a required column on every row (a row
states `flagIn: []` or its projection list — there is no longer an
absent state), `flagKey` stays required wherever `flagIn` is
non-empty and typed against `CliFlags`, and `SelectionOptions`
(types.ts) is itself a mapped type derived from the table's
`selection`-joining rows, so the cast can no longer drift from the
table by construction.

## Validation

Tested commit: 1fcf8d9268bb1b46589f77ad1a046c6459b09265 — pnpm check:affected --run green locally (format, lint, typecheck, layering, fallow, build, vitest-related)

- `pnpm typecheck` and `pnpm lint` clean after every edit.
- Reintroducing the closed gap — `flagIn: ['input']` on `daemonBaseUrl`
  with no `flagKey` — now fails `tsc` at the table definition itself,
  not at a test; confirmed, then reverted.
- `pnpm vitest run --project unit-core src/commands/cli-grammar
  src/commands/__tests__/input-audience.test.ts
  src/commands/capture/index.test.ts
  src/commands/interaction/selectors.test.ts` — 29 tests green.
- `pnpm check:layering` — clean; the new `types.ts` →
  `common-input-fields.ts` edge stays intra-`commands`-zone.

## Review notes

- Dropped the stale `--no-record`
  (`COMMON_COMMAND_SUPPORTED_FLAG_KEYS`) parenthetical from the
  `commonInputFromFlags` doc comment — it attached the flag-groups
  parser list to a common-input-fields row; those are separate axes
  per `flag-groups.ts`.

### DIFF
--- a/src/commands/cli-grammar/common.test.ts
+++ b/src/commands/cli-grammar/common.test.ts
@@ -23,9 +23,25 @@ const ALL_COMMON_FLAGS: Partial<CliFlags> = {
   androidDeviceAllowlist: 'emulator-5554,emulator-5556',
 };
 
+// Common-input-fields.ts rows that read a real `CliFlags` key but deliberately join neither
+// `flagIn` projection today (operator/env-sourced). Setting these too, on top of
+// `ALL_COMMON_FLAGS`, means a row that wrongly gained `flagIn: ['input']` would surface a
+// concrete value here rather than being silently dropped by `commonInputFromFlags`'s
+// `compactRecord` — over-inclusion `ALL_COMMON_FLAGS` alone cannot catch, since an unset flag
+// reads as `undefined` and gets compacted away either way.
+const NON_PROJECTED_COMMON_FLAGS: Partial<CliFlags> = {
+  daemonBaseUrl: 'https://daemon.example',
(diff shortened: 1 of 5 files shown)


thymikee:
Closing: opened in error by a re-run of the automation after #2292 (the same branch) had already merged. The extra commits here (d0fdfb0b, 1fcf8d92) are an unrequested second pass over the merged change; if any of it is wanted it should come as a fresh, reviewed PR against main.

PR #2301, opened 2026-09-05
test(contracts): one typed conformance helper for the runtime-family suites

thymikee:
## Summary

Ten `packages/contracts/src/*-runtime.test.ts` files (alert, app-event,
app-switcher, back, clipboard, home, keyboard, orientation, settings,
tv-remote) repeated the same three binding-rule tests by copy for every
interactor-catalog operation. Those three rules now live once, in
`interactor-operation-conformance.test.ts`, driven by the catalog itself: each
rule walks the exported `INTERACTOR_OPERATIONS` and binds through the catalog
row's own `bind` and `label`, exactly as production does. The per-operation
expectations (interactor method, minimal input) sit in one table typed over
every catalog operation, so a missing, duplicated or unregistered row fails
typecheck and a table-vs-catalog test, and every row is executed by
construction — completeness is structural, not observed from source text.
Each facet's `*-runtime.test.ts` keeps only its dedicated tests.
`app-inventory-runtime.ts`/`app-state-runtime.ts`/
`platform-runtime-operations.ts` (#2295) and
`platform-runtime-unavailable.test.ts` (#2291) are untouched.

## Validation

Head `453d28e0e5`.

- Focused: `npx vitest run packages/contracts/src` — 60 files / 304 tests
  passed. `oxlint`, `oxfmt --check`, `tsc -b packages/contracts`, `check:layering`
  and `check:fallow --base origin/main` clean.
- Planted red (row commented out, spelling kept in the comment): commented
  the `keyboardEnter` row of the expectation table → `tsc -b packages/contracts`
  fails (`Property 'keyboardEnter' is missing`) and all four conformance tests
  fail. Reverted.
- Planted red (registry member removed): dropped `back` from
  `INTERACTOR_OPERATIONS` → typecheck fails on the now-excess `back` row and the
  table-vs-catalog test fails. Reverted.
- Earlier heads (`a87c6984d6` and before) carried the source-scanning gate
  that the review found; CI on those was green except a pre-existing iOS smoke
  flake on `d0b11ba2` that cleared on rerun.


thymikee:
Sentinel review at `a87c6984d60ef4e75ae1f46e15437ba8cee28398`.

P2: completeness reads raw test source and regex-matches operation literals (`packages/contracts/src/interactor-operation-conformance.test.ts`, lines 14–23), so commenting out the entire keyboardEnter row still counts it as conformed while no tests execute that row. Use typed row data shared by actual test registration and completeness rather than reconstructing coverage from source text. Add a planted-red case removing an executable row while its spelling remains in a comment; completeness must fail. Consolidation otherwise looks sound, but hold readiness until the gate proves executable coverage.

thymikee:
Addressed in 453d28e0e5. The source-scanning completeness gate is gone. The three binding rules now live in one test that walks `INTERACTOR_OPERATIONS` and binds through each catalog row's own `bind`/`label` (the production path), with per-operation expectations in a table typed over every catalog operation. So the row data executed by the tests *is* the completeness data: a row that is commented out (spelling kept in the comment) fails `tsc -b packages/contracts` (`Property 'keyboardEnter' is missing`) and all four conformance tests; a catalog member dropped fails typecheck on the excess row and the table-vs-catalog test. Both planted and reverted. The per-facet fixture and `conformInteractorOperations` blocks are removed; each facet test keeps only its dedicated tests.

thymikee:
Sentinel review at 453d28e0e5c7f54e1f13cbbf5361c03e55fee7a7: Independent delta review is source-clean. Source-text completeness scanning is gone: a catalog-keyed typed expectation table is executed through each production bind/label, so a commented-out row cannot count as conformance. Missing/excess rows have planted-red proof. Remaining blocker is exact-head CI: Size encountered the cross-version base-report js.rawBytes failure and rerun checks are pending. No additional device acceptance required for this test-only refactor.

PR #2302, opened 2026-09-05
refactor(daemon): inline the touch sub-switch into the interaction dispatcher

thymikee:
## Summary

`interaction-touch.ts` was a 31-line pass-through: a second command-name switch that only
re-dispatched `press`/`click`/`longpress`/`hover`/`fill` to the same handlers
`handleInteractionCommands` already pulled in. Inlined those five cases into the existing
switch in `src/daemon/interaction/internal/interaction.ts` and deleted `interaction-touch.ts`.
No client-, operator-, or command-author-visible behavior change — this only removes one
indirection hop on the press/click/longpress/hover/fill dispatch path (−1 hop).

Review follow-ups landed in the same branch:

- The ADR-0011 responseData guard scanned files by an `interaction-touch*.ts` name prefix,
  which stopped covering the switch once it moved into `interaction.ts`. The dispatcher is now
  held to a structural rule instead: the guard parses `interaction.ts` with oxc-parser and
  requires every `press`/`click`/`longpress`/`hover`/`fill` case to be exactly one
  `return await <handler>(...)` whose callee is imported from an `interaction-touch*.ts` module.
  Nothing can be constructed in the switch at all, so a nested switch, a local `responseData`,
  an inline literal or a foreign callee all fail. The guard's own tests plant each of those
  shapes in an inline dispatcher source and assert the exact violation.
- `docs/adr/0019-end-state-hop-trace.md` traced both files this PR merges by name and pinned
  itself as unaffected by this PR. Merged the corresponding table rows, renumbered the
  remaining hops, and updated every downstream count/range reference in that doc and in
  `docs/adr/0019-request-bound-platform-runtime.md`.

## Validation

Tested at `de5036093f` (structural guard), `0b982bfcc0` (oxfmt only) and `31af37752b` (folds the delegation check under fallow's complexity threshold, then merges main for the #2310 size-report schema) on top of `292c0edb220e50f5f34f9436c777fafad631eba0`.

- `npx vitest run src/daemon/interaction/internal/__tests__/` — 30 files, 197 passed
- `npx tsc -p tsconfig.json --noEmit` — clean
- `npx oxlint` on the changed test file — clean
- Planted red (`de5036093f`): gave `press` a nested `switch` followed by a hand-rolled
  `responseData` in the real `interaction.ts`; the guard failed with
  `case 'press': the case body is not exactly one `return await <handler>(...)``, then passed clean after
  reverting. The same shape, an inline response literal, a non-touch callee and a removed
  command are pinned as permanent inline planted-red cases in the guard test (7 tests green).
- Full affected gate: green at 292c0edb220e50f5f34f9436c777fafad631eba0 (check:affected; format, lint, typecheck, layering, fallow, build, vitest-related)


thymikee:
Sentinel review at 292c0edb220e50f5f34f9436c777fafad631eba0.

P2: the new case-body regex stops at the first nested case/default, so a hand-rolled responseData after a nested switch in press is invisible to the construction guard; the previous whole-file scan caught it. The delegation move itself is behavior-preserving, but the filename exception plus copied command list weakens the invariant. Keep construction enforcement at the owning typed/delegation boundary or reuse structural parsing instead of formatting-sensitive case extraction. Plant a nested switch followed by hand-rolled responseData and prove the gate rejects it before restoring. Hold readiness until this is resolved.

thymikee:
Addressed in de5036093f. The case-body regex is gone; the guard now parses `interaction.ts` with oxc-parser and requires each touch case to be exactly one `return await <handler>(...)` whose callee is imported from an `interaction-touch*.ts` module, so construction enforcement stays at the delegation boundary and nothing formatting-sensitive remains. Planted the nested-switch-then-hand-rolled-responseData shape in the real `press` case: the guard rejected it (`case 'press': the case body is not exactly one return statement`); reverted. That shape plus an inline literal, a foreign callee and a missing command are pinned as inline planted-red cases in the test itself.

thymikee:
Sentinel review at 0b982bfcc0d253f6f490467a37dcdeefc37be040: The nested-switch enforcement hole is fixed by the AST delegation check, independently reviewed. One actionable gate failure remains: delegationOf at interaction-response-construction-guard.test.ts:105 is 11 cyclomatic / 9 cognitive and fails Fallow in Compatibility & Provenance (job101366047914). Simplify/split the owning AST shape checks without raising the baseline, retain the planted negatives, and rerun the gate. Size also fails from incompatible base-report js.rawBytes; device CI must clear separately. No readiness label.

PR #2303, opened 2026-09-05
ci: avoid unrelated Apple runner cache invalidation

thymikee:
## Summary

Keep prebuilt Apple runners reusable across unrelated dependency and package-version changes. Hash the runner build commands instead of the whole root manifest and lockfile, and include architecture, bundle IDs, and the Swift build wrapper in cache identity. Report cache hits/misses in the job summary and save using the restore action's primary key.

The existing main-branch producer already serves the iOS workflows: run 33981982967 restored a 65 MB runner in 3.4 seconds and preflight reported `buildMs: 0`. No additional prebuild workflow is needed. The revised key requires one initial rebuild per variant.

Scope: one shared CI action; no runtime, docs, or skill changes.

## Validation

Head: `31b07ed5bf96ab837995a039eafdf8fcc329cf34`.

All runnable `pnpm check:affected --run` checks passed. The first sandboxed attempt stopped because the cleanup guard could not run `ps`; the rerun with process-list access passed.

Focused shell checks demonstrated the old key's invalidation gaps and passed against the revised action. Actionlint passed for all seven consuming workflows. Native/device CI and the first new-key cache restore remain to be observed on GitHub.

PR #2306, opened 2026-09-05
fix(apple-runner): never compare an unavailable toolchain probe; name the mismatching cache keys

thymikee:
## Summary

On a loaded host the Apple toolchain probes (`xcodebuild -version`, `xcrun --show-sdk-version`, `--show-sdk-build-version`) time out. The probe answered `unknown`, memoized it, and persisted it into the rebuilt cache, so the next healthy daemon mismatched and paid a full `build-for-testing` inside the first `open`.

Unavailability is now a distinct outcome with no comparable value: only a complete, parsed fingerprint is memoized (per SDK, after all three probes answered and parsed — a failed or malformed round keeps nothing, so the next request re-probes), and an unreadable toolchain fails the cache decision with a typed `COMMAND_FAILED` (`reason: apple_toolchain_probe_unavailable`, `retriable: true`) naming it. Skipping the field would rebuild anyway — the fingerprint names the derived-data directory — and risks another Xcode's cache; AGENTS.md keeps optional probes best-effort "only when the feature contract says so." `write-xcuitest-cache-metadata.mjs` refuses an unread probe too. `cache_metadata_mismatch` now names the differing keys.

## Validation

Head `4ffd542234` (`d6a89f5e33` parsed-fingerprint memo + oxfmt-only `4ffd542234`; `9f68e00641` for everything before it): `check:quick`, `format:check`, `check:layering`, fallow, gate-manifest, depgraph, the eager-closure and test-file-size gates, and `vitest --project apple-runner packages/platform-apple/src/runner/__tests__` (350 passed; one `runner-client` abort test timed out under host load 20+ and passed twice in isolation).

Planted red, same vitest command:

- `unknown` fallback restored → 6 failed (`expected [Function] to throw an error`).
- empty metadata diff → 4 failed (`+ [] - [{ key: 'runnerSandboxBuildArgs', ... }]`).
- shared source root dropped, then its ignored set emptied → 1 failed each in `runner-source.test.ts`.
- raw-output memo restored (`9f68e00641`'s `runToolchainProbe`) against `d6a89f5e33`'s tests → 2 failed: the malformed-to-healthy recovery in one process, and the partial-round-not-kept check.

Device evidence: the iOS lane was green on `cca16c55f0`, re-running on `9f68e00641`. Android smoke failed there in the alert-dismiss step; Apple-only diff, main green — flake pending rerun. Affected gate: pending.

## Tradeoffs

Invalidates the prebuilt runner cache once. The source fingerprint and its tests moved to `runner-source.ts`; the mismatch diagnostic sits in `runner-cache.ts`, keeping over-budget `runner-artifact.ts` behavior-free.


thymikee:
Sentinel review at 05c7ac0f568e37787946afbf085eb3d9c9eee7ae.

P2: Do not memoize Xcode output before validating its fingerprint. In runner-toolchain-fingerprint.ts, runToolchainProbe caches every nonempty zero-exit answer before parseXcodeVersionOutput can classify it as unparsable. A malformed answer therefore remains cached: after the host becomes healthy, the next request still throws apple_toolchain_probe_unavailable without rerunning xcodebuild. Exact-source fake-host reproduction: both attempts fail; the total probe-call count stays at three. Memoize only a successfully parsed result and add a malformed-output → healthy-output recovery test within the same process (without resetting the memo). The current recovery test only covers a thrown timeout. Separately, the body says #2303 already hashes runner-toolchain-fingerprint.ts, but its reviewed head does not; correct that dependency note.

thymikee:
Re-reviewed fa75e73e284b9f45e9d79636fa5025d8c91c1487 with independent review. The existing P2 remains after the move into runner-cache-metadata.ts: runToolchainProbe (201-211) memoizes nonempty zero-exit Xcode output before parseXcodeVersionOutput (236-251) validates it. A transient malformed answer stays cached even after the host recovers. Cache the parsed successful fingerprint, and prove malformed-output to healthy-output recovery in one process without resetting the memo; the current recovery case only covers a thrown timeout. No additional code finding. Hold readiness; exact-head native and remaining CI evidence is separate.

thymikee:
Addressed in d6a89f5e33. The memo no longer holds raw probe output: `requireRunnerToolchainFingerprint` caches the complete parsed `RunnerToolchainFingerprint` per SDK, written only after all three probes answered and parsed, so a malformed `xcodebuild -version` answer keeps nothing and the next request re-probes. Two tests pin it in one process without resetting the memo: malformed → healthy recovers and re-runs exactly `xcodebuild, xcrun, xcrun`; a partial round (healthy xcodebuild + empty sdk build) keeps nothing, and the first healthy round is then served from the memo with zero probe calls. Planted red: the previous `runToolchainProbe` against these tests fails both. The stale `#2303` dependency note is no longer in the body.

thymikee:
Sentinel review at 4ffd5422340cf83b03e4077e06fde8f3c9d6e6ab: Source review clean, independently confirmed. Complete parsed fingerprints are now memoized only after all probes succeed; same-process malformed-to-healthy and partial-round recovery tests close the previous finding. Size CI is blocked by the cross-version report schema (old reporter reads missing base js.rawBytes), not measured size growth. Exact-head iOS was still running at review. Clear those CI blockers before merge; do not suppress the size check.

PR #2245, opened 2026-09-02
feat: add strict native absence assertion

thymikee:
## Summary

Adds the strict native `is absent <selector>` assertion for zero matches in one readable, complete, unscoped, full-depth accessibility capture. It uses the existing `readAny` selector match domain, requests no rectangles, refuses `--scope` and `--depth`, and fails closed with typed evidence for sparse, truncated, or unreadable captures. Existing `is hidden` semantics are unchanged.

The predicate is available in the CLI, MCP metadata, SDK types, and help/docs, with both positional forms supported:

```sh
agent-device is absent 'label="Loading..."'
agent-device is 'label="Loading..."' absent
```

Present failures report only the match count and bounded first-match identity/text fields; multiple matches advise refining the selector. Authored `is absent` observations record without target-v1 annotation, and replay failures remain action-failures rather than identity mismatches.

Refs #2236.

PR A coordination:

- Base: `main`
- Exact base commit: `d04e213bcc841cd8d75361844aec871cbbd5a831`
- Branch: `feat/is-absent-2236`
- Exact PR A head: `97c555c3876e1ee4f2041302a820113db0c6cf43`
- PR B (`wait absent`) must stack on this exact head (`feat/is-absent-2236` at `97c555c3876e1ee4f2041302a820113db0c6cf43`), not on `main`.
- No merge requested; leave this PR for review.

## Validation

- Planted-red regressions were observed before the fix for strict absence positional/runtime behavior, cached-negative freshness, bounded first-match evidence, vocabulary/docs completeness, and the quality-less legacy iOS root-only sparse-capture case. Live iOS and Android scenarios now close a fixture control, assert it is absent, and verify the command path.
- Focused final unit suite: 10 files, 84 tests passed. Structural closure and file-size gates: 2 files, 410 tests passed. Typecheck and repository formatting checks passed.
- `pnpm check:affected --run` on this exact base/head: the first run was host-contended (4,328/4,345 tests passed); the retry passed 5,204/5,205 tests, with one unchanged provider-backed iOS lifecycle test timing out at its 5-second budget. That test passed in isolation (3/3), and all other runnable checks in the retry were green. GitHub remains authoritative for native/device/coverage lanes.
- Independent adversarial read-only review of the final diff reported no valid in-scope findings.
- Native/device, GitHub CI, and full coverage lanes were not run locally; GitHub remains authoritative for those lanes. This PR is published for review and is not merged.
[...]

PR #783, opened 2026-06-12
refactor(ios): snapshot capture plans with a structured quality verdict

thymikee:
Implements ADR 0004's explicit-strategies decision as architecture, following the snapshot-pipeline architecture review (candidates 1+2). Stacked follow-ups: #777 (annotations bundle) and #779 (verdict-driven find) target this branch and are review-approved; #780 reworks on top per its review.

## What

- **Snapshot backend seam**: three adapters — recursive tree, query sweep, private AX — behind one `captureWithBackend` dispatch. Each ADR-0004 strategy declares its chain as data (`regular: tree→queries→private-ax`, `compact: queries→private-ax`, `raw: tree→private-ax`); one plan runner walks the chain under a **20s umbrella budget** so chained recovery tiers can never stack past the runner's 30s main-thread watchdog (a real failure observed on a production app during validation). Terminal policy is per-plan: raw rethrows AX failures (diagnostics preserve errors), interactive fails closed with `runnerFatal` invalidation.
- **One quality classifier**: a single `sparsePayloadReason` predicate with machine reason codes plus one collapsed-leaf detector — replacing the three divergent sparse detectors (Swift structural / daemon `count===1` / CLI `count<=3`) that each got patched for a different failure shape.
- **Structured `snapshotQuality` verdict on the wire** (`state`, `backend`, `reason`, `reasonCode`, `effectiveDepth`, `collapsedLeafIndexes`): the daemon and CLI render warnings from it instead of re-deriving degradation from node shapes; it surfaces in `--json` for agents; budget starvation is worded as "busy app or simulator", not blamed on the app's accessibility. Legacy runner messages and daemon-side detectors remain behind a verdict-absent gate for mixed-version compat.
- **Review fixes** (external review): the query-sweep synthetic root uses the real viewport (off-screen candidates can't masquerade as on-screen in find's duplicate-label resolution), and private-AX `--scope` selects the matched **subtree** instead of text-filtering every node.

## Verification (live, four failure shapes + healthy)

| scenario | verdict | result |
|---|---|---|
| Settings full | `healthy / tree` | 109 nodes, zero warnings |
| Settings `-i -c` under load | `recovered / private-ax / budget` | 30 nodes, budget wording |
| Production RN login (modal overlay) | `sparse / queries` honest best-effort | structured warning |
| Collapse fixture | `healthy` + `collapsedLeafIndexes` | `@ref merges many labels…` warning |
[...]

PR #2315, opened 2026-09-05
fix: stop stamping recovered iOS captures truncated; confirm Android alert dismissal

thymikee:
## Summary

Two CI flake families hit main and several PRs from 2026-09-03 to 2026-09-05. Both are product defects that a test change from that week exposed, not environment noise.

| Lane | Failures | Signature |
|---|---|---|
| iOS Smoke | 7 of 13 | `is absent id="automation-close-sheet"` → `capture was truncated` |
| Android Smoke | 5 of 5 | `get text id="automation-alert-result"` → `Selector did not match` |

### iOS: `truncated` meant provenance, not completeness

`stampedSnapshotPayload` (`RunnerTests+SnapshotCapturePlan.swift`) set `truncated: true` on every non-healthy capture since #783. On a loaded macOS CI host the fixture app's XCTest tree/query captures time out or exceed the 3s slow threshold, the runner penalizes that channel for 120s, and every capture becomes `SNAPSHOT_RECOVERED backend=private-ax reason=deferred` — a whole tree reported as truncated. Nothing consumed the flag until the strict absence assertion (#2245, merged 09-03) refused truncated captures, which is when the flake started. The E2E conformance helper even asserted the conflation (`truncated === (state !== 'healthy')`).

Fix: `truncated = payload.truncated || state == "sparse" || effectiveDepth != nil`. New runner unit test `testStampedPayloadTruncationTracksCompletenessNotRecoveryProvenance` (added to the ios.yml targeted list); the conformance helper now asserts `truncated === false`.

### Android: `alert dismiss` returned before the dialog left

#2260 swapped a polling `wait text` for a one-shot `get text` right after `alert dismiss`. Android's `handleAndroidAlertAction` pressed the button and returned immediately, while the iOS runner already re-checks `alertStillVisible`. In the failing CI artifact the dismiss step took 4.9s and the following read 2.3s and still saw only the dialog window.

Fix: `alert accept|dismiss` poll until the same dialog (title + buttons) is gone, bounded by the existing `ALERT_ACTION_RETRY_MS` budget, else fail with `alert <action> did not dismiss the visible alert`. A different alert taking its place counts as dismissed.

## Validation

- **iOS, local:** on a private simulator, running `is absent` with the private-AX backend pinned through the daemon flag the conformance transport uses reproduced the CI error before the fix (`observation: "truncated"`, quality `recovered/deferred`, `truncated: true`) and passes after (`truncated: false`, quality unchanged). Runner unit tests (new + 4 neighbors) pass; the plain tree path passes before and after.
- **Android, local:** `smoke:automation-system` passes on a private emulator with the rebuilt dist; `alert dismiss` now 2.6s, the following read 0.8s. platform-android suite 656 tests green; the new tests are revert-sensitive (read count and the budget failure).
- Typecheck, oxlint, oxfmt clean on the changed files.

## Not addressed here

The remaining iOS Smoke failures (4–5 of 13) are `wait` timeouts on an overloaded host: `xcrun simctl terminate` timing out at 2s, tree captures over the 8s slice, the Settings replay failing attempt 1 with a capture stall even in green runs, and a hot runner reused from the replay step becoming unresponsive so a 15–20s restart consumes the wait's 10s budget. Whether a mid-wait runner restart should get its own budget, or the fixture E2E should stop reusing the replay step's runner, is a separate decision.


thymikee:
Independent Terra high sentinel review of aa8f5a7a5e3994aa30a81c365d2658831f6d13bd:

No confirmed production defect in the iOS completeness change or Android post-dismiss polling route. The reported local simulator/emulator evidence is consistent with the traced behavior.

Blocked on deterministic provider-integration fixture drift: [job 101372400747](https://github.com/callstack/agent-device/actions/runs/33990743898/job/101372400747) fails 4 Android alert cases because each static `snapshotXml` keeps serving the pre-action dialog after `input tap`/Back. The new re-check correctly waits and then reports the dialog still visible. Update those scenarios with the existing `snapshotXml` + `onAdbExec` world transition pattern (and explicitly restore a dialog between sequential accept/dismiss assertions), retain the focused polling regression, then rerun provider-integration and Coverage.

Remaining CI is still pending; this head is not merge-ready.

thymikee:
Superseding my prior sentinel note for updated head d4a4f833a8267b060f615f230ea1b3e54ced0672:

Clean. The provider scenarios now reuse the existing stateful world callbacks to remove a dialog after tap/Back, explicitly restore it before the sequential second action, and add the persistent-dialog negative case. That addresses the deterministic 4-case provider-integration failure without changing production behavior.

I found no confirmed production defect in the iOS completeness or Android dismissal-polling routes. The reported local simulator/emulator evidence is consistent with the unchanged production diff. Current-head CI remains pending, so this is not yet merge-ready.

thymikee:
Sentinel delta review at b783ab388c7cf511e08b212dbb321508b78ec595: clean coordinator + independent Terra/high review. Alert integration cases/fixtures remain intact after relocation; the E2E waits for the app callback result then pins it to the specific result element. Failure artifacts retain the original failed-step evidence and add best-effort screenshot/snapshot capture. Previous production review and reported local simulator/emulator proof remain applicable. Current Android, Coverage, Integration and core gates pass; iOS smoke is still pending, with no current failing checks or conflicts. Ready for human review, not yet confirmed merge-ready. Supersedes the old-head label/verdict.

PR #2325, opened 2026-09-06
fix(ios): honor the startup budget through a cold Simulator boot

### LINKED ISSUE #2324: Cold iPhone startup ignores the requested preparation deadline
## Description

A never-booted iPhone Simulator can require longer than the fixed internal boot wait to complete Apple's first-boot data migration. Increasing the public prepare timeout does not extend that internal wait. This prevents unattended cold-device startup even though the simulator eventually finishes booting successfully.

## Steps to reproduce

1. Create a new iPhone 17 Simulator with an installed iOS runtime, without booting it first.
2. In an isolated CLI state directory, run:
   `agent-device prepare ios-runner --platform ios --udid <fresh-iphone-udid> --timeout 240000 --json`
3. On a first boot whose migration exceeds 120 seconds, observe the internal simctl bootstatus timeout before the requested command deadline.
4. Observe the same target with read-only `xcrun simctl bootstatus <fresh-iphone-udid>` without -b. After initialization completes, runner preparation can succeed.

The duration is host-dependent; a fast first boot may not reproduce this.

## Expected behavior

An explicit startup budget should reach the simulator boot and runner preparation stages. Opening a cold simulator should retain its existing host-global claim continuously through startup into the established session, so a second workspace cannot acquire the same target while the first is initializing it.

## Actual behavior

With agent-device 0.20.10, the preparation path hit a 120-second internal bootstatus timeout despite --timeout 240000. A fresh iPhone's first-boot migration took approximately 4 minutes 42 seconds. Retrying preparation after boot completion succeeded in approximately 16 seconds. An earlier retry while startup was still in progress also failed, so blindly retrying is not a reliable provisioning workflow.

Ordinary open has a shorter request budget, while prepare owns only a transient claim. Splitting initialization across failed prepare requests and a later open therefore does not provide a continuous task-long claim.

## Environment

- agent-device: 0.20.10
- Target: newly created iPhone 17 Simulator, iOS 26.5 runtime
- Local macOS/Xcode simulator workflow, multiple independent workspace sessions
- Current follow-up host: macOS 26.6.2, Xcode 26.6, Node.js 24.15.0. The initial observation was made before this follow-up environment check.

No application installation or application-specific behavior is required to reproduce the boot wait.

## Proposed boundary and acceptance criteria

Proposed CLI shape, not an existing supported open option:
`agent-device open Settings --platform ios --udid <fresh-iphone-udid> --timeout 600000 --json`

- [ ] A caller-selected startup deadline propagates through request handling, simulator readiness, and required runner startup; an internal fixed 120-second cap must not preempt a longer explicit budget.
- [ ] Existing host-global claim ownership is retained from initial acquisition through successful open. No second ownership mechanism or release/reacquire gap is introduced.
- [ ] A competing workspace receives DEVICE_IN_USE throughout cold startup and the subsequent active session.
- [ ] Deadline expiry and cancellation are bounded, preserve structured diagnostics/progress, and finish owned cleanup before releasing the claim.
- [ ] Regression coverage proves deadline propagation, contention, cancellation, and cleanup; a real never-booted iPhone verifies the local CLI end to end.

## Related work

#2307 centralizes request/teardown budget calculation but describes preserving current arithmetic and behavior. #2308 and #2312 concern managed lease admission. This report concerns ordinary local iPhone startup and should reuse those existing ownership/budget boundaries where applicable, not duplicate managed allocation or workspace pinning.

Searched open issues and PRs for boot, bootstatus, first boot, migration, startup, and timeout; no matching cold-boot deadline report was found.

PrinceD96:
## Summary

Closes #2324.

Supersedes the original implementation by @PrinceD96 (head `8bdb85b7a3ea749ad9c631a8acb3fe573c831124`), who found the bug, the stale `close --shutdown` shortcut, and the cold-Simulator validation recipe; the commit carries a `Co-authored-by` trailer for them.

A never-booted iOS Simulator runs Apple's first-boot data migration, which can take several minutes. The boot wait in `ensureAppleReady` was capped at a fixed 120 seconds that neither `prepare --timeout` nor `open` could reach, so the cap preempted the caller's budget.

This head replaces the earlier implementation with the smallest change that makes the budget reach the boot wait:

- **The boot wait takes an absolute deadline.** `simctl boot` and `simctl bootstatus -b` share it; expiry fails with `error.details.reason: boot_timeout` and a hint. The Simulator is left booting (a deadline is not a cancellation), so a retry finds it further along or ready. Ctrl+C keeps its existing shutdown-on-cancel behaviour.
- **`prepare --timeout` is one budget for the boot and the runner.** The runner preparation gets what the boot left instead of the full budget again.
- **`open --timeout` is new** and bounds the boot for a fresh session. The device claim is held from `open` onward, so a competing workspace gets `DEVICE_IN_USE` throughout the cold start without any extra claim mechanism.
- **Client envelope.** `open` and `prepare` use the existing `budget + 30s margin, never below the base envelope` arithmetic (`envelope: 'margin'`, the same shape `wait` already uses), so the daemon's structured timeout wins the race against the client-side daemon reset. `prepare --timeout 240000` now yields a 270s envelope instead of 240s.
- **`close --shutdown` always asks simctl.** The session device carries `booted: false` from selection time for its whole life, so the old shortcut reported a successful shutdown on a Simulator that stayed Booted. Reproduced live before the fix, verified fixed after.

Dropped from the earlier head, on purpose: the daemon-side timer that cancelled the whole request (the budget is scoped to the boot, so `--foreground`'s snapshot is never cut by it), the typed cleanup-failure error and claim abandonment path, forcing runner readiness inside `open`, the `waitForSimulatorBoot` on an already-Booted device, and the `MAX_STARTUP_TIMEOUT_MS` plumbing. Timer overflow for absurd `--timeout` values is pre-existing for every `--timeout` command and is out of scope here.

## Validation

Unit: readiness (deadline shared by boot + bootstatus, expiry → `boot_timeout`, no shutdown; default 120s kept), lifecycle (prepare shares the budget; open forwards its deadline), timeout policy (envelope arithmetic + reviewed sets), open CLI/structured input, shutdown runtime. `pnpm check:affected --run` green.

Live, local Xcode 26 / iOS 26.2 runtime, three freshly created never-booted iPhone 17 Simulators, isolated `--state-dir`:

- `open com.apple.Preferences --timeout 5000` on a never-booted Simulator: failed after 8s total with `boot_timeout` + hint; request log shows `simctl boot` 1.8s then `bootstatus` cut at 3.1s (one shared budget). Simulator left Booted.
- `open com.apple.Preferences --timeout 600000` on a never-booted Simulator: success in 77s (boot 4s, bootstatus 52s). This Mac finishes the first boot under the old cap, so the unit test carries the cap-removal proof; the live run proves the budget propagates and the flow completes.
- `prepare ios-runner --timeout 600000` on a never-booted Simulator: success in 104s (bootstatus 40s, then runner connect + health check).
- `close --shutdown` after opening a Simulator selected while Shutdown: before the fix it returned `shutdown.success: true` with empty stdout/stderr and the Simulator stayed Booted; after the fix simctl is called and the Simulator reaches Shutdown.

### DIFF
--- a/packages/command-registry/src/registry.ts
+++ b/packages/command-registry/src/registry.ts
@@ -996,7 +996,8 @@ export const RAW_COMMAND_DESCRIPTORS = [
       allowSessionlessDefaultDevice: allowAnyDeviceSessionless,
       saveScriptFlagOwner: true,
     },
-    timeoutPolicy: DEFAULT_TIMEOUT_POLICY,
+    // --timeout is a startup budget: it reaches the Simulator boot wait (#2324).
+    timeoutPolicy: { ...DEFAULT_TIMEOUT_POLICY, budget: { source: 'flag', envelope: 'margin' } },
     batchable: true,
     platformExecution: { kind: 'device-runtime', uses: openApplicationRuntimePlanUses },
   },
@@ -1008,9 +1009,10 @@ export const RAW_COMMAND_DESCRIPTORS = [
     frameworkTier: 'extended',
     recordsSessionAction: false,
     daemon: { route: 'session', refFrameEffect: 'preserve' },
-    // Runner warm-up builds are the longest fixed envelope; --timeout overrides.
+    // Runner warm-up builds are the longest fixed envelope; --timeout is the
+    // daemon-side boot + runner budget, so the envelope keeps a margin over it.
     timeoutPolicy: {
-      budget: { source: 'flag' },
+      budget: { source: 'flag', envelope: 'margin' },
       envelopeMs: PREPARE_REQUEST_TIMEOUT_MS,
       onTimeout: 'reset-daemon',
     },
--- a/packages/command-registry/src/timeout-policy.ts
+++ b/packages/command-registry/src/timeout-policy.ts
@@ -80,6 +80,11 @@ function resolveFlagBudgetTimeoutMs(
   if (policy.budget.envelope === 'widen') {
     return resolveWideningFlagBudget(policy, policy.budget, flags);
   }
+  if (policy.budget.envelope === 'margin') {
+    return typeof flags?.timeoutMs === 'number'
+      ? widenToUserBudget(policy, flags.timeoutMs)
+      : policy.envelopeMs;
+  }
   return typeof flags?.timeoutMs === 'number' ? flags.timeoutMs : policy.envelopeMs;
 }
 
--- a/packages/command-registry/src/types.ts
+++ b/packages/command-registry/src/types.ts
@@ -36,9 +36,14 @@ export type DaemonCommandTraits = Omit<DaemonCommandDescriptor, 'command'>;
  *                           ever EXTENDS the envelope to envelopeMs + budget +
  *                           margin (interaction --settle semantics, #1101: the
  *                           flag bounds a post-action wait, so the request must
- *                           also cover selector/action overhead). `defaultBudgetMs`
- *                           is used when the feature flag is present but the
- *                           numeric timeout flag is omitted.
+ *                           also cover selector/action overhead). With
+ *                           `envelope: 'margin'` the budget is a daemon-side
+ *                           deadline (open/prepare startup): the envelope is
+ *                           budget + margin, never below `envelopeMs`, so the
+ *                           daemon's own structured timeout wins the race against
+ *                           the client envelope. `defaultBudgetMs` is used when
+ *                           the feature flag is present but the numeric timeout
+ *                           flag is omitted.
  *  - `'positional-parser'`— the budget travels inside the positionals; `parser`
  *                           extracts it (or returns null when none was given).
  *                           The client widens the envelope to
@@ -46,7 +51,7 @@ export type DaemonCommandTraits = Omit<DaemonCommandDescriptor, 'command'>;
  */
 export type CommandTimeoutBudget =
   | { source: 'none' }
-  | { source: 'flag'; envelope?: 'bound' | 'widen'; defaultBudgetMs?: number }
+  | { source: 'flag'; envelope?: 'bound' | 'widen' | 'margin'; defaultBudgetMs?: number }
   | { source: 'positional-parser'; parser: (positionals: string[]) => number | null };
 
 /**
--- a/packages/contracts/src/application-lifecycle-runtime.ts
+++ b/packages/contracts/src/application-lifecycle-runtime.ts
@@ -28,6 +28,11 @@ export function hasRuntimeTransportHintValues(values: RuntimeHintValues): boolea
 
 /** Request-scoped runner/diagnostic context, without daemon request types. */
 export type ApplicationLifecycleExecution = Readonly<{
+  /**
+   * Absolute time by which a cold Simulator's boot must finish, from `open --timeout`. Absent
+   * means the platform's default boot wait; `prepare` derives its own deadline from `timeoutMs`.
+   */
+  startupDeadlineAtMs?: number;
   requestId?: string;
   logPath?: string;
   traceLogPath?: string;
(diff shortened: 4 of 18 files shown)


thymikee:
One issue at `8bdb85b`: `open --foreground --timeout` keeps the startup timer running while taking the initial snapshot. If startup succeeds but the snapshot crosses that deadline, the snapshot gets canceled even though the timeout is documented as a startup budget.

Please stop the startup timer before taking the snapshot, and add a test where startup finishes within the budget but the snapshot finishes after it. The test should fail before the fix and pass after it.

The cold-start and cleanup runs are documented in the PR, but CI on this commit still requires action before it can run.

thymikee:
The startup-timeout finding still applies at `8bdb85b`. There are now conflicts with main as well. Please resolve those alongside the fix and run CI on the updated commit.

thymikee:
I replaced the branch head with a smaller implementation of the same fix (previous head `8bdb85b7a3ea749ad9c631a8acb3fe573c831124`, still reachable if you want to compare).

What stayed: the boot wait honors an absolute deadline, `prepare --timeout` is one budget for boot + runner, `open --timeout` is new, the client envelope keeps the 30s margin over the daemon budget, and the `close --shutdown` stale-`booted` shortcut is gone (reproduced live before the fix, verified after).

What went: the daemon-side timer that cancelled the whole request, the typed cleanup-failure error and claim-abandonment path, forcing runner readiness inside `open`, `waitForSimulatorBoot` on an already-Booted device, and the `MAX_STARTUP_TIMEOUT_MS` plumbing. The budget is scoped to the boot wait itself, so the earlier `--foreground` finding cannot recur: the initial snapshot never runs under the startup deadline. Expiry leaves the Simulator booting on purpose, so a retry with a larger `--timeout` finds it further along.

Live validation on three freshly created never-booted iPhone 17 Simulators is in the PR description; `pnpm check:affected --run` is green on this head.

thymikee:
The foreground-snapshot issue is fixed at 963e58b. One deadline gap remains: after bootstatus, waitForSimulatorBoot calls simulatorState with a fresh 15-second timeout and never checks the deadline on success. If bootstatus uses nearly all the budget, the final state query can cross it and still return success. Please pass the remaining budget to that query and add a regression where it finishes after the deadline; the result should be boot_timeout. CI is still unreported on this head.

thymikee:
Fixed in `c03332ee26`. The confirming `simctl list` after bootstatus now runs with `min(15s, remaining budget)`, and a confirmation that lands after the deadline is reported as `boot_timeout` (the budget is the contract, not the boot's outcome). Regression: boot at +2s, bootstatus at +22s, the listing returns Booted at +31s against a 30s deadline → `boot_timeout`, listing timeout asserted at 8s, no shutdown issued. It fails on the previous head and passes on this one.

`pnpm check:affected --run`: 3693/3694 unit tests green; the one failure is the unrelated Re.Pack `webpack-start` spawn test timing out at 5s under load, which passes in isolation. CI on this head still needs the workflow approval on the maintainer side.

thymikee:
The Coverage failure on this head is main-side: `no APPROVED_OVER_CEILING row is stale` fails on main at e7a5b8b622 and on every PR based on it, because the planned-operations row was consumed by #2329. #2373 removes the row; re-running Coverage here after it merges should go green without touching this branch.

thymikee:
Reviewed c03332e: the final boot-state query now shares the deadline, and the regression covers a successful response arriving too late. No remaining code findings; the reported cold-Simulator runs cover the startup path. Ready for human review. Coverage fails on a stale planned-operations.ts closure approval, which looks unrelated; one smoke lane is still running.

PR #2323, opened 2026-09-06
fix(ios): avoid replaying alert mutations

billsbooth:
## Summary

Prevent XCTest-backed `alert accept` and `alert dismiss` from being replayed after a lost response. Default/explicit `get` stays retryable. Six transport regressions fail before the fix and pass afterward.

## Scope

12 files: action-aware runner traits across client, session and recovery; transport/provider/startup regressions; shared TypeScript/Swift golden table; help and command docs. No native alert-identity change or QA driver version update.

## Validation

Commit `53d9f01f4ccde3665d73b76fd5441ddb64a00d97`: `pnpm check:affected --run` passed (2,915 related tests). All CI green. [macOS](https://github.com/callstack/agent-device/actions/runs/34011187893): 191 native tests passed, including the golden table. [iOS](https://github.com/callstack/agent-device/actions/runs/34011187890): 77 native tests plus live alert get/dismiss/accept outcomes and session cleanup passed. Local host bootstrap hung; separate Settings replay needed attempt 2 after a capture timeout.

## Verification

Observability audit: existing command IDs correlate dispatch and status recovery. Tests preserve typed lifecycle/recovery details and prove one mutation dispatch. No additional payload logging, secrets, or configuration.

## Automated Test Plan

- Run `pnpm exec vitest run --project apple-runner packages/platform-apple/src/runner/__tests__/runner-recovery-wiring.test.ts packages/platform-apple/src/runner/__tests__/runner-command-traits.test.ts`; expect no accept/dismiss replay for accepted, started, or completed lost-response states; queries still retry.
- Run the macOS host XCTest lane; require `testAlertReadOnlyClassificationMatchesGoldenTable` to pass.
- Build Agent Device Tester using `examples/test-app/README.md`; follow `docs/agents/device-verification.md` for fresh source/isolated simulator setup. Keep the same `--platform ios --udid <owned-udid> --session alert-retry` on every command. Open `com.callstack.agentdevicelab`, navigate Settings → Open automation lab, `click id="automation-open-alert"`, then `alert get` and `alert dismiss`; expect `Alert result: cancelled`. Reopen and `alert accept`; expect `Alert result: accepted`. Retain command/status evidence; `close` the original session. Executed in iOS CI.

PR #2326, opened 2026-09-06
fix(ios): confirm alerts without repeating activation

billsbooth:
## Summary

Prevent one iOS alert action from tapping a replacement. Native activation happens once; subsequent checks only observe. For example, `agent-device alert accept` must leave a following prompt untouched.

## Scope

Nine files: native confirmation, UIKit fixture, selected XCTest cases, help and docs. Independent of #2323's transport retry repair. No driver-version or permission-setup changes.

## Verification

Production Swift policy: five failures before repair, seven checks passing afterward. At `9a1da31e1`, frozen install, build and `pnpm check:affected --run` passed. An earlier cleanup failure reproduced on unchanged upstream; no retries/timeouts were added.

Automated QA evidence: [current-head iOS](https://github.com/callstack/agent-device/actions/runs/34015120372) passed 81 native tests, including all four replacement/deadline regressions. Hosted macOS passed 194 tests. [First-head iOS](https://github.com/callstack/agent-device/actions/runs/34013474448) also passed live ordinary alert get/wait/dismiss/accept and cleanup.

## Validation

At `9a1da31e1`, eleven checks passed; iOS failed its initial fixture-home text wait before alert actions. Settings replay needed three attempts. Capture/readiness diagnosis remains open; not merge-ready. First-head Android capture failure also remains recorded. Physical/system-popup parity is untested. Changed presentation is not unique identity or permission proof; identical alerts remain unconfirmed. UI is fixture-only.

## Automated Test Plan

Run this PR's **iOS / Smoke Tests** check. Four selected `RunnerTests+AlertObservationTests` launch a credential-free fixture: changed-title accept, same-title dismiss, identical-looking replacement, expired-before-action. Assert first/replacement counts 1/0 (0/0 before admission), inspect Cancel/OK independently, retain XCTest results. Tests terminate the fixture.

Require existing live fixture alert outcomes/cleanup and macOS host policy tests. Never replay live permission acceptance to reproduce failures.

### DIFF
--- a/.github/workflows/ios.yml
+++ b/.github/workflows/ios.yml
@@ -169,6 +169,10 @@ jobs:
             -only-testing:AgentDeviceRunnerUITests/RunnerTests/testCachedTargetInvalidationClearsProcessBoundState \
             -only-testing:AgentDeviceRunnerUITests/RunnerTests/testAlertResolutionCannotBypassRequestedDeadline \
             -only-testing:AgentDeviceRunnerUITests/RunnerTests/testAlertAcceptTreatsOpenAsAffirmative \
+            -only-testing:AgentDeviceRunnerUITests/RunnerTests/testAlertAcceptDoesNotActivateAReplacementWithASharedButton \
+            -only-testing:AgentDeviceRunnerUITests/RunnerTests/testAlertDismissDoesNotActivateAReplacementWithTheSameTitle \
+            -only-testing:AgentDeviceRunnerUITests/RunnerTests/testAlertCannotProveAnIdenticalReplacementAndDoesNotActivateIt \
+            -only-testing:AgentDeviceRunnerUITests/RunnerTests/testAlertDeadlineBeforeActivationLeavesTheOriginalUntouched \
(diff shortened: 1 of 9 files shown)


thymikee:
The macOS runner does not compile at `2413aeb`: `runningBackgroundSuspended` is not available on macOS, but `RunnerTests+AlertObservation.swift` uses it unconditionally. Please follow the platform-specific handling already used in `RunnerTests+Lifecycle.swift`, then rerun the macOS build and tests.

The iOS tests passed, including the replacement-alert cases. Android also hit a snapshot timeout while waiting for landscape; that needs a passing run, but there is no evidence here that the alert change caused it.

thymikee:
The macOS compile issue is fixed at `9a1da31e`; the build and all 194 native tests passed. The iOS native suite also passed, including all four new alert regressions. No new code findings. The remaining iOS smoke failure happens before alert actions, while waiting for the fixture home screen with no readable captures. Please resolve or establish the cause of that failure and get the required iOS check passing before merge.

PR #2327, opened 2026-09-06
fix(runtime): resolve each lazily imported platform module exactly once

### LINKED ISSUE #2314: session-open-runtime.test.ts leaks an Apple runner pkill into the next test
## Symptom

In `src/daemon/session-lifecycle/internal/__tests__/session-open-runtime.test.ts`, the test `open --metro-port alone stays host-ambiguous on a physical Android device` fails intermittently with the hermetic-signal guard:

```
Error: This test tried to signal 2 process(es) this vitest worker did not spawn.
spawn pkill -TERM -f xcodebuild.*test-without-building.*AgentDeviceRunner\.env\.session-sim-1-[0-9]
spawn pkill -KILL -f xcodebuild.*test-without-building.*AgentDeviceRunner\.env\.session-sim-1-[0-9]
    at packages/host-kit/src/internal/exec.ts:140 (runSpawnedCommand)
```

The pattern names device `sim-1`, i.e. the Apple simulator session opened by the *previous* test in the same file (`open --metro-port alone defaults the host to 127.0.0.1 on an iOS simulator`). That test's `open` reaches the real runner disposal (`packages/platform-apple/src/runner/runner-disposal.ts` `killRunnerXcodebuildProcesses` via `runner-session.ts`) and its two `pkill` spawns are not awaited by the test, so they land in whichever test runs next.

## Repro (on `main` @ 96727a0b42)

```
pnpm vitest run --project unit-core \
  src/daemon/session-lifecycle/internal/__tests__/session-open-runtime.test.ts \
  src/daemon/__tests__/wait-runtime.test.ts \
  src/daemon/handlers/__tests__/snapshot-alert.test.ts --maxWorkers=1
```

Failed 1 of 3 runs on main, 2 of 3 on a branch with unrelated daemon changes (#2296), and twice inside `pnpm check:affected --run`'s `vitest related` set for that branch. The file passes alone; the leak needs enough load for the disposal to spill past the test boundary.

## Fix direction

Mock the runner-session seam the iOS `open` reaches (the test already mocks `@agent-device/platform-apple/runner/operations`, but disposal goes through `runner-session.ts` / `platform-runtime-apple-application-tools.ts` `stopRunnerSession`), or make the iOS test await the disposal it triggers. Per the guard's own message: mock the seam, do not let a unit test spawn `pkill`.

thymikee:
Fixes #2314.

## What was happening

`session-open-runtime.test.ts > open --metro-port alone stays host-ambiguous on a physical Android device` failed the hermetic-signal guard, but only when `vitest related` scheduled it in a worker under load:

```
This test tried to signal 2 process(es) this vitest worker did not spawn.
spawn pkill -TERM -f xcodebuild.*test-without-building.*AgentDeviceRunner\.env\.session-sim-1-[0-9]
```

The Android test never touches an Apple runner. The `pkill` came from **two tests earlier in the same file**, and reached the host through a Vitest mocking race.

Each port on `createAppleApplicationTools()` opened its own `import(...)` of the same specifier — seven of them for `@agent-device/platform-apple/runner/operations` alone. In production those duplicates are equivalent, because the module loader caches. Under Vitest they are not: **while a `vi.mock` factory is still awaiting `importOriginal()`, a second dynamic import of that id resolves to the UNMOCKED module.**

The open path makes that overlap routine — `packages/platform-apple/src/lifecycle.ts` deliberately leaves the iOS runner prewarm unawaited (`recordUnawaited`). So:

1. `open applies launch-only flags…` (iOS simulator `sim-1`) schedules prewarm #1. Its dynamic import triggers the file's `vi.mock` factory, which awaits `importOriginal()` — the whole real Apple runner graph, slow under a loaded worker.
2. `open --metro-port alone defaults the host to 127.0.0.1 on an iOS simulator` schedules prewarm #2 while that factory is still in flight. It gets the **real** module.
3. The real `prewarmIosRunnerSession` → `prepareLocalIosRunner` → `ensureRunnerSession` → `prepareRunnerLeaseForStartup`, which finds an empty lease and spawns `pkill -f xcodebuild...session-sim-1-[0-9]` against the developer's own process table.
4. That lands during the next test, and the guard names it.

That is the guard doing exactly its job: a unit test really was reaching for foreign processes.

## The fix

Each specifier now has one memoized loader, and the ports reach their module only through it — so the second resolution the escape needs no longer exists. Same idiom already used in `src/platform-runtime.ts` and `packages/platform-android/src/index.ts`. Applied to the Apple tools, the Android tools, and `loadAndroidMechanics`, which had the same duplicated-lazy-import shape.

No production behavior changes: Node's loader already returned the same cached module for the duplicate imports.

## Regression test

`src/__tests__/platform-runtime-apple-application-tools.test.ts` drives two runner ports concurrently and asserts both reach the mock. It fails deterministically without the fix:

```
AssertionError: expected "vi.fn()" to be called 1 times, but got 0 times
```

It uses the two shutdown ports, whose real implementations are no-ops on an empty session map, so a regression run fails on the assertion without doing host process work.

## Validation

- Repro command from the issue, 6 consecutive runs: **0 refusals**, 231/231 files (run 6 had one unrelated `waitForFileContent` timeout in `android-lifecycle.test.ts` under contention; it passes on its own). Before the fix it reproduced in 1 of the first runs each time I tried.
- `pnpm test:unit` — 1203 files, 9011 passed.
- `pnpm lint`, `pnpm format:check`, `pnpm typecheck`, `pnpm check:layering`, `pnpm check:di-seams`, `pnpm depgraph:test`, `pnpm check:gate-manifest`, `pnpm check:fallow` — all green.

### DIFF
--- a/src/__tests__/platform-runtime-apple-application-tools.test.ts
+++ b/src/__tests__/platform-runtime-apple-application-tools.test.ts
@@ -1,0 +1,31 @@
+import { expect, test, vi } from 'vitest';
+import { createAppleApplicationTools } from '../platform-runtime-apple-application-tools.ts';
+
+const detachIosSimulatorRunnerSessionsForShutdown = vi.hoisted(() => vi.fn(async () => 0));
+const stopAllIosRunnerSessions = vi.hoisted(() => vi.fn(async () => {}));
+
+// The factory awaits the real Apple runner graph on purpose: that wait is what let a second,
+// concurrent dynamic import of this specifier overtake the still-unresolved mock and hand the
+// caller the UNMOCKED module (#2314).
+vi.mock('@agent-device/platform-apple/runner/operations', async (importOriginal) => {
+  const actual =
+    await importOriginal<typeof import('@agent-device/platform-apple/runner/operations')>();
+  return { ...actual, detachIosSimulatorRunnerSessionsForShutdown, stopAllIosRunnerSessions };
+});
+
+// Two ports of these tools run at once whenever the open path leaves its runner prewarm
+// unawaited, so every port must reach the runner module through the one memoized loader. A port
+// that opens its own `import(...)` resolves the specifier a second time, and a unit test that
+// loses the mock this way starts the real local XCTest runner — whose stale-process cleanup
+// `pkill`s xcodebuild on the developer's host, failing whichever test is running when it lands.
+test('concurrent runner ports share one module resolution, so the mock always applies', async () => {
+  const tools = createAppleApplicationTools();
+
+  await Promise.all([
+    tools.detachRunnerSessionsForShutdown(),
+    tools.finalizeRunnerSessionsForShutdown(),
+  ]);
+
+  expect(detachIosSimulatorRunnerSessionsForShutdown).toHaveBeenCalledTimes(1);
+  expect(stopAllIosRunnerSessions).toHaveBeenCalledTimes(1);
+});
--- a/src/platform-runtime-android-application-tools.ts
+++ b/src/platform-runtime-android-application-tools.ts
@@ -8,12 +8,24 @@ import { AppError } from '@agent-device/kernel/errors';
 import { emitDiagnostic } from '@agent-device/host-kit/diagnostics';
 import { loadAndroidMechanics } from './platform-runtime-android-mechanics.ts';
 
+/**
+ * One memoized loader per lazily imported module, and the only way the ports below reach one.
+ * See the same loaders in `platform-runtime-apple-application-tools.ts` for why a port never
+ * opens its own `import(...)` (#2314).
+ */
+let openTargetModule: Promise<typeof import('./platform-runtime-open-target.ts')> | undefined;
+const loadOpenTarget = () => (openTargetModule ??= import('./platform-runtime-open-target.ts'));
+
+let runtimeHintsModule: Promise<typeof import('./platform-runtime-runtime-hints.ts')> | undefined;
+const loadRuntimeHints = () =>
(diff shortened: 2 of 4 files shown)


thymikee:
Rebased onto `6e22e266d7`. The three red checks on the previous run were not from this branch — `main` is broken independently, and #2328 fixes it. Until that merges, this PR's Typecheck / Repo Guards / Integration Tests will keep failing on:

```
src/daemon/managed-device-allocation/lease-admission.ts(19,8): error TS2307:
  Cannot find module './record-validation.ts'
```

thymikee:
No code findings in the loader change reviewed at `807fbbc`. The shared lazy imports preserve runtime behavior, and the concurrency test covers the reported mock escape without risking host process cleanup. CI on that commit was blocked by the managed-allocation move and an iOS fixture startup timeout. The head changed to `52d65cc` during review, so this assessment does not yet cover the updated commit.

thymikee:
Stacked on #2328 so this branch's CI runs against a repaired tree — `main` itself is red (`Cannot find module './record-validation.ts'`, see #2328), which is where every failure on the previous run came from. Base retargets back to `main` automatically when #2328 merges.

The one failure that was *not* inherited — the Android smoke `wait text "Alert result: cancelled"` — passed on re-run; it was a live-emulator flake.

thymikee:
Looks good at 89e2f860. The shared lazy imports preserve initialization order and the regression covers concurrent callers. All checks are green. Please merge #2328 first; this PR is based on it.

PR #2320, opened 2026-09-05
refactor(capture-kit): move durable-capture resource mechanics out of the daemon

thymikee:
Closes #2317.

The daemon held two halves of one mechanism. `@agent-device/capture-kit` already owned the durable-resource envelope, JSON, and descriptor codec; the fence, transition, adoption, and recovery mechanics that operate on that envelope still lived in `src/daemon`.

## The four seams, and what replaced them

| Seam | Replacement |
| --- | --- |
| `SessionStore.set` / `.resolveSessionDir` | `DurableCaptureSessionStore<S>` — a two-member port |
| `SessionState` | a session type parameter `S`; only a definition's own `sessionSlot` looks inside it |
| `safeSessionName` | recovery takes `resolveSessionDir(sessionId)` from its caller |
| `DurableCaptureAdmissionLedger` | `DurableCaptureCleanupOutcome`, reported by the mechanics; the daemon decides admission |

`recoverFailedAdoption` computes `confirmed` exactly as before — `(!persisted && cleanupError === undefined) || transition.confirmed` — and reports it. `createDurableCaptureResource` maps `confirmed: true` to `clearUndurableCleanup` and `confirmed: false` to `blockUndurableCleanup(device, reason)` with the same reason string, still applied before the primary error is rethrown. The admission ledger is never named inside capture-kit.

Splitting `DurableCaptureRecordDefinition` out of `DurableCaptureResourceDefinition` says which half needs a session at all: recovery, finish-recovered, and start preflight terminalize a persisted record with no session in hand, so they name the record definition and carry no session type. That also drops the handle parameter from those signatures, which was only ever there to reach `sessionSlot`.

## What moved, and what stayed

Eight files (1,390 lines) left `src/daemon` for `packages/capture-kit/src/durable-capture/`, behind one new `@agent-device/capture-kit/durable-capture` subpath. It is deliberately not on the `.` index — that index is the eager closure every platform runtime imports.

The daemon keeps five `durable-capture-*` files, 297 lines: the admission ledger, start preflight, runtime binding, the closed kind set, and the composition root that wires the moved mechanics to the ledger and the session store's naming rule. All sixteen kind stamps stay.

`git show -M90%` reads every production move as a rename (90–100% similarity); `git show -M90% 50f9a5d` has the summary.

## Tests

[...]

Issue #2316, opened 2026-09-05
refactor(daemon): move managed device allocation into its own workspace package


thymikee:
Blocked by: none. Implements the packaging choice ADR 0021 §8 leaves open; related to #2284, #2285, #2278.

## Purpose

`src/daemon/managed-device-allocation/` (27 production files, 2,973 lines) imports nothing from `src/daemon`, and nothing in `src/` imports it. Its only dependencies are workspace packages: `@agent-device/contracts` (`managed-device-allocation` ×11, `platform-runtime` ×4, `client` ×3), `@agent-device/kernel/errors`, `@agent-device/host-kit/file` + `/process`, and `@agent-device/capture-kit`. It has no production consumer today: `src/managed-device-reachability.ts` uses only the contracts vocabulary, and only tests reach the implementation.

Measured on `main` at `f1d4efe059` (2026-09-05) with an import-graph walk over every non-test `.ts` under `src/`.

Move it into its own workspace package now, before the managed runtime binding (ADR 0021 §3) becomes its first consumer, so the package `exports` map defines the API instead of a directory convention. Ownership does not change: the daemon still composes the store, journal, and decision policy (ADR 0021 §3, "allocation-operation journal" stays a daemon authority). Only the module boundary moves.

## Scope — exact tracked files

Production (all of `src/daemon/managed-device-allocation/`):
`decision.ts`, `fence.ts`, `journal.ts`, `journal-action-binding.ts`, `journal-action-context.ts`, `journal-action-persistence.ts`, `journal-action-runtime.ts`, `journal-lane.ts`, `journal-projection.ts`, `journal-results.ts`, `journal-types.ts`, `record.ts`, `record-codec.ts`, `record-codec-state.ts`, `record-factory.ts`, `record-json.ts`, `record-validation.ts`, `schema.ts`, `status.ts`, `store.ts`, `store-creation.ts`, `store-filesystem.ts`, `store-lock.ts`, `store-results.ts`, `transition-outcome.ts`, `transition-support.ts`, `transitions.ts`.

Tests (move with the package):
`src/daemon/managed-device-allocation/__tests__/{decision,journal,record,store}.test.ts` and `__tests__/fixtures.ts`.

Not moved: `src/__tests__/managed-device-allocator-fake.test.ts` and `src/__tests__/test-utils/managed-device-allocator.fixtures.ts` import only `@agent-device/contracts/managed-device-allocation`; leave them where they are.

## Destination

[...]

PR #2328, opened 2026-09-06
fix: repair main after the managed-allocation move landed under stale PRs

thymikee:
`main` is red, and has been since #2308. Every open PR fails Typecheck, Repo Guards, and Integration Tests because of it — this unblocks them.

## What broke

#2316 moved managed-device allocation into `@agent-device/managed-allocation`. #2308 was authored before that move and merged after it, so the daemon files it added still import pre-move sibling paths:

```
src/daemon/managed-device-allocation/lease-admission.ts(19,8): error TS2307:
  Cannot find module './record-validation.ts'
src/daemon/managed-device-allocation/lease-admission.ts(70,5): error TS2322:
  Type 'string | undefined' is not assignable to type 'string'
src/daemon/managed-device-allocation/__tests__/lease-admission.fixtures.ts(10,61): error TS2307:
  Cannot find module './fixtures.ts'
test/integration/provider-scenarios/managed-runtime-automation.fixtures.ts(17,8): error TS2307:
  Cannot find module '.../managed-device-allocation/__tests__/fixtures.ts'
```

A semantic conflict, so neither PR's own CI could have caught it.

## Repairs

- **`lease-admission.ts`** reaches `freezeLease` / `isValidLease` / `isRequestGeneration` / `isVerbatimId` through the package's `./record` surface, which now re-exports them. `record-validation.ts` stays package-internal.
- **`TS2322` needed no separate fix.** It was a consequence of the unresolved import: with the module resolved, `isVerbatimId` narrows `grant.identityIncarnationId` to `string` again.
- **`src/daemon/managed-device-allocation/__tests__/fixtures.ts`** comes back, stated in contract terms only. Both trees keeping their own test data is the shape #2316 already chose for `managed-device-allocator.fixtures.ts`, which exists identically at both the root and package.
- **`.fallowrc.json`** drops its `@agent-device/managed-allocation` `ignoreDependencies` entry. Its comment read "no root consumer yet … it lands after the package's move" — the root consumer has now landed.

## Also failing on main, independently

`scripts/__tests__/eager-closure-budgets.test.ts > no APPROVED_OVER_CEILING row is stale` fails on pristine `origin/main`. The merge-base now carries `packages/capture-kit/src/durable-capture/index.ts` (added with #2320), so nothing can read its approval row, and the table's own staleness rule fails it — the intended outcome, just never actioned. Row removed.

## Validation

Verified failing on pristine `origin/main` first (typecheck and the closure gate, in a clean worktree), then green here:

- `pnpm typecheck` — clean
- `pnpm test:unit` — 1207 files, 9057 passed
- `managed-request-admission` + `managed-runtime-automation` provider suites — 19 passed
- `pnpm lint`, `pnpm format:check`, `pnpm check:layering`, `pnpm check:fallow`, `pnpm check:freerange` — all green

### DIFF
--- a/.fallowrc.json
+++ b/.fallowrc.json
@@ -56,9 +56,6 @@
     // @agent-device/provider-limrun owns the source import, while the published
     // root build externalizes @limrun/api and retains runtime imports in packed chunks.
     "@limrun/api",
-    // @agent-device/managed-allocation has no root consumer yet: the managed runtime binding
-    // (ADR 0021 §3) is its first, and it lands after the package's move out of src/daemon.
-    "@agent-device/managed-allocation",
     // Oxlint resolves the shared config's JS plugins dynamically from their package names.
     "@nkzw/eslint-plugin",
     "eslint-plugin-no-only-tests",
--- a/packages/managed-allocation/src/record.ts
+++ b/packages/managed-allocation/src/record.ts
@@ -4,6 +4,15 @@ export { ALLOCATION_OPERATION_SCHEMA_VERSION };
 export { newAllocationOperation } from './record-factory.ts';
 export { decodeAllocationOperationRecord } from './record-codec.ts';
 export { bindingFenceFor } from './record-fence.ts';
+// The daemon's managed lease admission (src/daemon/managed-device-allocation/lease-admission.ts)
+// validates a grant against the same record rules the package applies internally, so these leave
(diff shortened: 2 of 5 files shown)


thymikee:
Looks good at 1e65590c. The import repair uses the public package surface, the restored fixtures stay test-only, and the obsolete gate exception is removed. All checks are green.

PR #2279, opened 2026-09-04
feat(ios): route Simulator snapshots through AX bridge

thymikee:
## Summary

Closes #2197. Eligible local iOS Simulator snapshots use the packaged AX bridge and existing TypeScript presenter, with typed, generation-scoped XCTest fallback. Unknown generations remain incomparable; regular depth is presentation-only and raw depth bounds acquisition.

Target discovery is reused only while OS process-start identity matches. Native acquisition verifies the foreground owner before and after capture, preventing app-only trees beneath system dialogs. XCTest retains modal resolution; fallback disables the bridge until that app generation changes. Provider, physical-device, interaction, and prewarm routes are preserved.

Scope: 34 files, 1,353 gross lines; maintainer-approved exception for required correctness and regression coverage.

## Validation

Head: `ae26f7afc064c75c65c71e677d32d77cdf2f9709`; matched base: `7a2d48d160aaacb582c6f4cde98b6af7531bf7af`.

- [Published acceptance evidence](https://github.com/callstack/agent-device/tree/a0e64182142236fa71be1aea55ec1f7d98ccfd3b/pr-2279) is held separately on `evidence/ios-snapshot`, tagged `evidence/ios-snapshot/ae26f7afc0`: raw responses, traces, provenance, checksums, and verification script.
- Six screens × 20 warm samples per revision: 240 captures, zero failures or anchor mismatches. CLI median/p95 improved on all six; sequential runs under uncontrolled host load do not establish a general speed guarantee.
- Live foreground fallback, dismissal/relaunch circuit proof, and real native capture tests retained. Five native scenarios pass; deleting PRE produces three failures, deleting POST one.
- Final-head `pnpm check:affected --run` passed (746 files / 5,691 tests); all runnable CI checks green. Accepted Size growth: +35,845 unpacked bytes.

Acceptance evidence is complete. Runner-demand/open latency and proxy acceptance remain #2198 scope.

PR #2331, opened 2026-09-06
fix(ios): stop charging every capture for a slow Simulator app discovery

thymikee:
## Summary

Removes repeated concurrent app discovery on the iOS Simulator AX-bridge route and bounds how long each capture waits for it. Resolver and route regression tests pass; loaded-host smoke validation is what this PR's own CI run will show, and it was still pending when this was written.

## The failure it targets

Since #2279 landed (09-05 21:06), the iOS Smoke lane has failed four times with one signature, the first `wait text "Agent Device Tester"` after a cold `open --relaunch` reporting `wait_capture_stalled` with `readableCaptures: 0`:

- main `80997b6b`: https://github.com/callstack/agent-device/actions/runs/33992502696
- main `f30328d0`: https://github.com/callstack/agent-device/actions/runs/34014253127
- PR #2326: https://github.com/callstack/agent-device/actions/runs/34015120372
- branch `fix/2314-memoize-platform-runtime-la`: https://github.com/callstack/agent-device/actions/runs/34016148244

The request log of the third run (artifact `ios-artifacts`, `sessions/ios-e2e-smoke-14ax/requests/45ad58af8fd6970d.ndjson`) accounts for the 10s budget:

| phase | cost |
|---|---:|
| runner `findText` on the just-launched app | 5.8s |
| AX-bridge target discovery: `xcrun simctl spawn … launchctl list`, `xcrun timed out after 3000ms` | 3.4s |
| fallback XCTest snapshot | cancelled at the deadline |

The failed-step screenshot and snapshot the harness took right after (#2315) show the home screen rendered with a healthy tree. That is consistent with the timeline, not proof the app was readable throughout the wait.

`createSimulatorSnapshotTargetResolver` (`snapshot-target.ts`) ran that spawn with a fixed 3s timeout on every cache miss, on the capture's critical path, and kept nothing on timeout, so the next capture ran its own probe again while xcrun was slow. xcrun takes about 1s on an idle Mac with simulators booted; on these CI hosts it exceeded 3s in both captures of that run, and the same hosts time out `simctl terminate` at 2s in other artifacts.

## Change

- Discovery is single-flight per target and detached from the capture that starts it. A capture waits at most 1.5s for the in-flight discovery, then takes the XCTest fallback with reason `simulator-target-discovery-pending`. Captures that keep arriving while the probe runs each wait up to that slice; the probe is shared, not repeated.
- A discovery has one 15s deadline shared by both simctl probes and the `ps` identity read.
- A cancelled caller no longer aborts the probe. A failed discovery is forgotten, so the next capture starts a new one.
- The resolver error names its reason, so the `ios_snapshot_route_fallback` diagnostic says why the fallback ran.

A responsive host still reaches the bridge on the first capture (local probe ≈ 1s < 1.5s).

## What is verified

- Resolver tests (deferred `spawn` mock, fake timers): slow discovery yields to the fallback and finishes in the background; concurrent captures join one discovery; a cancelled caller leaves it running; one deadline is shared by the probes and the identity read. The first three fail on the old code (hang, second spawn, second spawn).
- Route test with the production resolver over a simctl that answers only when released: the first capture falls back within the slice with the `target-resolution-failed` warning, the released discovery then serves the bridge without a second spawn.
- Existing resolver and route tests pass (20 total); package typecheck, oxlint, oxfmt, fallow clean.
- Live fixture E2E on a private iOS 26.2 simulator with the rebuilt dist: cold launch, `wait for Agent Device Tester` 269ms, 23 captures served by the bridge, `smoke:automation-input` green. This proves the healthy route, not the loaded-host case. (A later `keyboard dismiss` failure in that run is my local Simulator.app hardware-keyboard environment.)

## Not verified here

- A loaded-host `wait` completing through the fallback: no local way to make xcrun slow; the lane's own runs are the evidence.
- The production poll path end to end under slow discovery; the route test covers one capture, not the `wait` loop.

## Relation to open work

#2329 changes `snapshot-route.ts` (launch grace on typed bridge failures) and makes `wait` observe runner-free; it does not touch the probe, and its design needs the probe to be cheap on the critical path. #2325 covers `open` startup budgets. This PR does not change the `wait` contract.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ## Unreleased
 
+- Fixed: the iOS Simulator AX snapshot route no longer charges every capture for a slow app
+  discovery. Discovery (`simctl launchctl list` through xcrun) takes seconds on a loaded host; a
+  capture now waits a bounded slice for it, takes the XCTest fallback, and the discovery keeps
+  running so the next capture joins or reuses it instead of paying the probe again. A `wait`
+  issued right after `open` could previously lose its whole budget to repeated 3s probe timeouts
+  and report `wait_capture_stalled` with the app already on screen.
 - Fixed: iOS snapshots no longer report `truncated: true` merely because a later backend produced
   them. The runner stamped every recovered capture as truncated — including a complete private-AX
   tree taken while the XCTest channel was penalized as slow — so a strict `is absent` / `wait absent`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,12 +2,13 @@
 
 ## Unreleased
 
-- Fixed: the iOS Simulator AX snapshot route no longer charges every capture for a slow app
-  discovery. Discovery (`simctl launchctl list` through xcrun) takes seconds on a loaded host; a
-  capture now waits a bounded slice for it, takes the XCTest fallback, and the discovery keeps
-  running so the next capture joins or reuses it instead of paying the probe again. A `wait`
-  issued right after `open` could previously lose its whole budget to repeated 3s probe timeouts
-  and report `wait_capture_stalled` with the app already on screen.
+- Fixed: the iOS Simulator AX snapshot route bounds how long a capture waits for app discovery
+  and stops starting a discovery per capture. Discovery (`simctl launchctl list` through xcrun)
+  takes seconds on a loaded host; a capture now waits at most 1.5s for the one in-flight
+  discovery, takes the XCTest fallback, and the discovery keeps running under its own 15s
+  deadline for the captures that follow. Previously each capture ran its own probe with a 3s
+  timeout on its critical path, so a `wait` issued right after `open` could spend its budget on
+  probe timeouts and report `wait_capture_stalled` with the app already on screen.
 - Fixed: iOS snapshots no longer report `truncated: true` merely because a later backend produced
   them. The runner stamped every recovered capture as truncated — including a complete private-AX
   tree taken while the XCTest channel was penalized as slow — so a strict `is absent` / `wait absent`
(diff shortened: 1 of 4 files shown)


thymikee:
One issue at 2f8824ae: Promise.all rejects as soon as either discovery probe fails, and finally removes the single-flight entry even if the other probe is still running. If list devices fails while launchctl list is slow, each following capture can start another launchctl probe. Three sequential resolver calls with that failure produced three pending probes.

Keep ownership until both probes settle, or cancel the sibling through a discovery-owned signal before removing the entry. Add that partial-failure case to the regression tests. The healthy-device evidence is useful; loaded-host polling evidence remains outstanding.

thymikee:
Addressed in 142265a7f9: both probes settle before the discovery does (`Promise.allSettled`, first rejection rethrown after both), so the single-flight slot is held until `launchctl list` returns even when `simctl list devices` fails first. Regression test `a failed runtime probe does not release the slot while the launch-job probe still runs`: three sequential captures against that partial failure spawn once; the settled discovery then reports `simulator-runtime-probe-failed` and frees the slot.

One more data point on loaded-host polling from #2344's iOS failure (a branch without this change, run 34021895012): the wait's single poll was `findText` 3.5s, then target resolution succeeded, then `ios.snapshot-source.prepare` 1.1s and `acquire` 3.5s cancelled at the deadline. So on that host the first bridge acquisition after a cold launch costs more than the probe this PR bounds; #2343 covers the accounting.

thymikee:
The single-flight fix is clean at 142265a7. Waiting for both probes to settle closes the early-rejection gap, and the new regression covers repeated callers while the other probe is still running.

The iOS simulator lane and static, integration, coverage and package checks pass. Android Smoke still fails waiting for the Alert result after readable captures. That is outside this Apple-only change, but it needs a clean rerun before merge.

Issue #2343, opened 2026-09-06
wait: account startup and readiness separately from observation time, and say when infrastructure prevented any observation


thymikee:
## What happens today

`wait <target> <timeoutMs>` charges its whole budget to whatever the polls cost, including work that is ours rather than the app's: the XCTest runner's first command on a freshly launched app, runner start or restart, and AX-bridge target discovery. The deadline also cancels the poll in flight, so a first poll that costs more than the budget ends with zero completed observations.

Four iOS Smoke runs since 2026-09-05 21:00 show it on the first `wait text "Agent Device Tester"` after a cold `open --relaunch` (`wait_capture_stalled`, `readableCaptures: 0`):

- main `80997b6b`: https://github.com/callstack/agent-device/actions/runs/33992502696
- main `f30328d0`: https://github.com/callstack/agent-device/actions/runs/34014253127
- PR #2326: https://github.com/callstack/agent-device/actions/runs/34015120372
- branch `fix/2314-memoize-platform-runtime-la`: https://github.com/callstack/agent-device/actions/runs/34016148244

Request log of the third run (`sessions/ios-e2e-smoke-14ax/requests/45ad58af8fd6970d.ndjson` in `ios-artifacts`):

| phase | cost |
|---|---:|
| runner `findText` on the just-launched app | 5.8s |
| AX-bridge target discovery, `xcrun timed out after 3000ms` | 3.4s |
| fallback XCTest snapshot | cancelled at the 10s deadline |

An earlier instance of the same accounting, before #2279, is https://github.com/callstack/agent-device/actions/runs/33984867542 (`sessions/ios-e2e-smoke-ohv/requests/6af0b46614602883.ndjson`): the runner stopped accepting connections mid-wait, the daemon invalidated and relaunched it, and the wait was cancelled 0.7s after `xcodebuild` started, reported as `wait_capture_stalled` rather than a runner restart because the cancelled capture carried no restart evidence.

#2331 takes the discovery probe off the per-capture critical path. #2325 gives `open` its own startup budget. Neither changes what a `wait` budget pays for.

## What to change

- Account startup and readiness time (runner start, runner restart, target discovery, first-command warm-up) separately from observation time in the wait loop, and surface both in the failure details.
- Keep one overall deadline and cancellation. A hung capture must still end at the deadline, so this is not a guarantee of "at least one completed observation".
[...]

PR #2344, opened 2026-09-06
feat(wait): carry a per-poll timeline in timeout failures

thymikee:
## Summary

Diagnostics only; no timeout semantics change. Follow-up 4 from the review of #2331, prerequisite for #2343.

A polling `wait` timeout carried `reason`, `readableCaptures`, and `waitedMs`, which cannot say where the budget went. The runs behind #2343 spent a 10s budget on a single poll (5.8s runner `findText` on a fresh app, 3.4s of target discovery, a fallback cancelled at the deadline) and reported the same `wait_capture_stalled` as a dead runner would.

## Change

The polling timeout paths (`wait <selector>`, `wait text`, `wait @ref`, and `wait absent` once a readable capture has been seen) now also carry in `error.details`:

- `captures`: every poll attempted, readable or not.
- `polls[]`: one entry per poll with `startedMs` on the wait's own clock, `durationMs`, and a typed `outcome`: `readable`, `unreadable` (content verdict the wait rode out), `deadline` (the deadline cancelled the capture in flight), or `runner-restart` (that cancellation carried restart evidence).

`wait --stable` timeouts use their own error builder and a never-readable strict absence preserves its predicate failure; neither carries the timeline. The top-level `reason` and the existing fields are unchanged; `logPath` still links the full request log. Long waits keep the first five and last twenty-five polls so the response stays compact. Whether a readable poll matched is the caller's verdict and is not recorded.

Example of the #2343 failure with this change:

```json
"details": {
  "reason": "wait_capture_stalled",
  "captureStalled": true,
  "timeoutMs": 10000,
  "readableCaptures": 0,
  "captures": 1,
  "polls": [{ "startedMs": 0, "durationMs": 10285, "outcome": "deadline" }],
  "waitedMs": 10285
}
```

## Validation

- New tests: three polls (readable, unreadable, cancelled at the deadline) produce the expected records on the wait clock; a 40-poll wait keeps its first five and last twenty-five.
- Existing wait suites pass (wait-polling, wait-text, wait-absent, wait-selector, wait-stable, daemon wait-runtime: 59 tests); root typecheck, oxlint, oxfmt, fallow clean.
- Docs: `website/docs/docs/commands.md` wait-failure bullet; CHANGELOG entry.

## CI on this branch

The first run's Android Smoke failed on `wait for Alert result: cancelled` with the new timeline in its output (17 polls, 16 readable, then a deadline at 10,261 ms). The failed-step evidence shows the device in landscape at that point, with the canary below the fold; every tap before the alert was at portrait coordinates. That is the open Android lane problem tracked separately, not a regression from this change. The iOS Smoke failure is the cold-launch wait that #2331 targets (this branch does not include it).


thymikee:
The diagnostic change looks sound at 96e1a801, with one wording correction: not every wait timeout carries the timeline. wait --stable uses a separate error builder, and never-readable strict absence preserves its predicate failure. Please narrow the docs, changelog and PR description to the polling timeout paths that actually emit these fields; no need to expand the implementation here.

Android CI failed waiting for Alert result: cancelled. Its output does demonstrate the new timeline: 17 polls, 16 readable, then a deadline at 10,261 ms. That does not establish a regression from this change, but the failed lane still needs resolution before merge.

thymikee:
Wording narrowed in cc1add6e2f: docs, changelog, and the description now name the polling timeout paths that emit the timeline (`wait <selector>`, `wait text`, `wait @ref`, and `wait absent` after a readable capture) and say that `wait --stable` and a never-readable strict absence keep their own diagnostics. Implementation unchanged.

On the Android lane: the failed-step evidence shows the device in landscape at the timeout with `Open automation alert` 2220px wide, and the wait's 16 readable polls never saw the canary because it sits below the fold in that layout. Every tap before the alert landed at x=540 (portrait), so the rotation happened after `orientation portrait` had been confirmed by the fixture, somewhere around the dismissal. The same off-screen layout explains the earlier `Selector did not match` failures on that step. I have not found the trigger in any log we keep; I am adding rotation-state and logcat rotation-decision evidence to the Android harness so the next failure names it.

thymikee:
The documentation exclusions are improved at cc1add6e. One case remains: a replayed selector wait with a recorded landmark mismatch throws directly from wait-selector.ts, without the new captures/polls evidence. If the final poll reports runner-restart, that branch also hides the restart outcome.

Please preserve the compact evidence on that refusal and cover landmark mismatch followed by a deadline or runner restart, or explicitly exclude replay destination-guard refusals from the documented promise. The current wording still includes them.

thymikee:
CI has now completed: the remaining red smoke job is the installed snapshot-bridge verification, where xcrun timed out after 10 seconds. Android smoke passed on this head. The replay-guard evidence finding above remains separate; the failed verification also needs a clean rerun.

thymikee:
This head now conflicts with main after the recent merges. Resolve the conflict and revalidate the wait changes; the existing replay-evidence finding is unchanged.

thymikee:
Addressed: the landmark-mismatch refusal in `wait-selector.ts` now carries the same failure evidence as a timeout (`captures`, `polls`, `readableCaptures`, `waitedMs`, and the runner-restart fields when the final poll ended in a restart) next to its mismatch details. Two regressions cover a mismatch followed by a deadline-cancelled capture (`['readable', 'deadline']`) and by a runner restart (`['readable', 'runner-restart']` with `runnerRestarted: true` on the refusal). Docs and changelog name the refusal with the polling timeout paths. Main merged in for the changelog conflict; the bridge-verification xcrun timeout is the CI host and will get its rerun with this push.

thymikee:
The prior landmark-evidence finding is fixed and the code is clean. Android Smoke still fails at the landscape wait, so this is not merge-ready yet.

PR #2345, opened 2026-09-06
feat(ios): support explicit iOS simulator keychain reset

### LINKED ISSUE #2282: Support explicit iOS simulator keychain reset for fresh-install testing
## Use case

A customer testing an app on an EAS-hosted remote iOS simulator reported that `settings clear-app-state` left the app authenticated through Firebase credentials stored in the keychain. Their in-app "Reset to Fresh Install" operation cleared the state successfully.

Agents need to repeat first-run and signed-out flows without depending on an app-specific reset button.

## Current behavior

The documentation describes iOS `settings clear-app-state` as clearing the app data container. Keychain removal is not part of that documented contract, so this is a capability request rather than a claim that the existing command regressed.

Version baseline for the customer feedback is 0.20.10; exact client/daemon versions and Firebase configuration still need confirmation. This is a customer report, not an independently reproduced test.

## Requested capability

Provide an explicit way to reset keychain-backed authentication state on an iOS simulator, alongside app data, for fresh-install testing. This could be an opt-in reset mode or a separate command; the API shape is open.

- Document what app data and keychain state the operation clears and what remains.
- Prefer targeting the selected app where technically possible. If only a broader simulator keychain reset is available, make that scope explicit and opt-in rather than silently clearing other apps' credentials.
- Make the operation usable through a remote daemon.
- Validate with a keychain-backed login: authenticate, reset, relaunch, and verify the app starts signed out.

Related: #103 implemented app-container clearing; this asks for the separate keychain-backed state needed for a complete reset.

thymikee:
## Summary

- Fixes #2282: `settings clear-app-state` clears the app's data container but never touched Keychain, so a customer's fresh-install reset left the app signed in via Firebase auth even though their in-app "Reset to Fresh Install" button cleared it.
- Adds `settings reset-keychain clear`, a new iOS-simulator-only settings action that runs `xcrun simctl keychain <device> reset`.
- Kept it a **separate, explicit** command rather than an option on `clear-app-state`: simctl has no per-app keychain reset, only a whole-simulator one, so folding it into the existing per-app command would silently widen its blast radius. `reset-keychain` documents and requires an explicit opt-in to that wider scope instead.
- Works through the daemon/remote path for free — `settings` already dispatches generically to the platform's `setSetting` operation, so no daemon wire changes were needed beyond the shared contracts/usage-string updates.
- Documented in `website/docs/docs/commands.md` and in the CLI's own `--help` text, including the "pair with `clear-app-state` for a full fresh-install reset" guidance the issue asked for.
- Split the pre-existing `apps.test.ts` and `snapshot-handler.test.ts` suites along the `app-settings.ts` / `snapshot-settings.ts` modules they actually test — both were already over the repo's test-file-size tripwire and could not grow to hold the new tests.

## Test plan

- [x] `pnpm typecheck`
- [x] `pnpm lint`
- [x] `pnpm test:unit` (full suite, 9061 passed)
- [x] Live validation: built the CLI, booted a disposable iOS simulator, ran `agent-device settings reset-keychain clear --udid <udid>` against it — succeeded and returned the expected `{ scope: 'simulator', cleared: true }` payload; confirmed an invalid state (`nope`) is rejected before touching the device. Disposable simulator deleted afterward.


thymikee:
Please reject app-scoped input before resetting anything. The CLI accepts settings reset-keychain clear com.example.app and drops the extra argument; the direct daemon parser also accepts it. That can turn an apparent per-app reset into a whole-Simulator keychain wipe. Require exactly reset-keychain clear, and add CLI and direct-daemon tests proving an extra app argument causes no settings mutation.

The remaining validation from #2282 is also needed: authenticate with keychain-backed credentials, reset, relaunch, and verify the app starts signed out. The reported disposable-Simulator command success proves invocation, not that login outcome. Reviewed at 32371284.

thymikee:
Addressed both points in 98ff05d8ae:

**Extra positional argument** — fixed in both places: the CLI reader (`readSettingsOptionsFromPositionals`) now requires exactly `reset-keychain clear` with no trailing positional, and the direct-daemon parser (`parseSettingsArgs`) now rejects a present `positionals[2]` for `reset-keychain` the same way. Added tests for both paths proving `com.example.app` as a third argument is rejected with `INVALID_ARGS` and causes zero settings mutations (`fixtureSettingsMutations` stays empty).

**Real login-outcome validation** — added a small "Keychain-backed auth" section to `examples/test-app`'s automation lab (`expo-secure-store`), giving this a repeatable regression surface, then ran the full cycle live against a disposable iOS simulator (fresh native build, not the cached one):

1. Sign in (writes a keychain item) → status: `signed-in`
2. Plain relaunch → still `signed-in` (baseline persistence)
3. `settings clear-app-state` → relaunch → still `signed-in` (confirms it does **not** touch keychain, matching the customer report)
4. `settings reset-keychain clear` → relaunch → `signed-out` (confirms the new command actually clears the credential)

Disposable simulator deleted afterward. Full unit suite (9063 tests), typecheck, and lint all green on the updated branch.

thymikee:
The head is unchanged at 32371284, so the extra-argument validation finding and signed-out-state evidence request still stand. Android Smoke has also finished red at the post-alert canary (wait for Alert result: cancelled). This is the same failure seen on other PRs; it does not change the keychain review, but the required lane still needs to pass.

thymikee:
The new head 98ff05d8 addresses both review points. Extra arguments now fail before mutation on the CLI and direct-daemon paths, with regressions for both. The reported SecureStore check covers persistence through relaunch and clear-app-state, then signed-out state after keychain reset.

Ready for human review. The previous comment referred to the superseded head; current CI is still running, with no failed checks at this review.

thymikee:
Update: Android Release has since failed at 98ff05d8 (Gradle reports its Node subprocess exited 1). I removed ready-for-human while that new failure is investigated. The code review remains clean; please establish the build failure cause and get the release build passing.

thymikee:
The iOS smoke lane also finished red: the depth-1 bridge assertion received XCTest fallback after target-resolution-failed. Keep this separate from the keychain behavior and Android release-build failure; establish the cause and rerun the affected lane before merge.

thymikee:
The code is clean. The simulator-wide keychain scope is explicit in the command, result, and documentation. Current Smoke failures still block merge.

Issue #2341, opened 2026-09-06
refactor(daemon): extract the session event journal into a workspace package


thymikee:
Part of #2335.

## Purpose

`src/daemon/session-event-*.ts` (6 files, 1,867 lines; ADR 0018) has zero value dependencies on other daemon files. Its root dependencies are `command-catalog.ts` (4 files), `core/command-descriptor/registry.ts` (1) and `core/keyboard-actions.ts` (7 lines, no imports). Its type dependencies are `DaemonRequest`, `DaemonResponse`, `DaemonResponseData` from `src/daemon/types.ts` in `session-event-request.ts`, `session-event-log.ts`, `session-event-action-presentation.ts`; the fields read are `req.command`, `req.flags`, `req.meta.requestId`, `req.meta.clientArtifactPaths`.

Relocating the type declaration is not enough: today's `DaemonRequest` carries `internal` (callbacks taking `SessionState`, an admitted `DeviceLease`, lifecycle hooks) and daemon-only `meta` fields.

## Required behavior

- New workspace package (`@agent-device/session-journal`, name open) with one subpath per moved module; consumers `request-generic-dispatch.ts`, `request-execution-scope.ts`, `session-store.ts`, `replay/internal/session-replay-action-runtime.ts`, `replay/internal/session-replay-runtime-failure-response.ts` switch specifier.
- Journal inputs are typed by shared request types only: the wire-only request shape C exports, or `@agent-device/kernel/contracts` public request/meta types. No reachability from any journal input type to `SessionState` or `DeviceLease`. Splitting files or lowering importer counts does not satisfy this.
- `keyboard-actions.ts` folds into the package; catalog and registry come from A's package.

## Completion conditions

- The package imports nothing under `src/` (R11) and compiles against A's and C's packages only.
- A type-level test in the package asserts its input type has no `internal` key and no property path typed as `SessionState` or `DeviceLease`.
- Event journal output byte-identical for the existing journal fixtures; ADR 0018 references updated.
- `check:layering`, eager-closure (subpaths at renamed files), package-boundaries, fallow green.

## Dependencies

Blocked by A (catalog, registry) and C (wire-only request shape).

Issue #2335, opened 2026-09-06
refactor: daemon and contracts decoupling, wave 3 (registry package, type split, leaf moves)


thymikee:
## Purpose

Third decoupling wave for `src/daemon` and `@agent-device/contracts`, following #2316, #2317 and #2318. Measured on `main` at `27a97ee619` with a production-only import graph (1,629 files, 8,805 edges):

- `src/daemon`: 350 production files, 56.8k lines; 197 flat top-level files (30k lines). The value graph is acyclic and the daemon has zero edges into platform packages. It has 154 value edges into 67 root `src/` files; the top hubs are `command-catalog.ts` (15 daemon files), `core/command-descriptor/registry.ts` (13), `core/dispatch-resolve.ts` (12), `provider-device-runtime.ts` (11).
- `@agent-device/contracts`: 190 modules, 18.2k lines, 118 subpaths. 106 modules export runtime code (13k lines). 146 modules are reachable from a platform/provider package; 44 are not; 19 are reachable from no package at all.

The command descriptor registry is the keystone: it is package-ready today (its only root dependency is `core/parse-timeout.ts`) and it is what forbids packaging the daemon client and the session event journal under R11.

## Children (dependency order)

- [x] A — #2336 → #2348, `2ec4e91`
- [x] B — #2337 → #2349, `6768a04`
- [x] C — #2338 → #2346, `dcd8b65`
- [x] D — #2339 (blocked by A) → #2357, `bd08e6e`
- [x] E — #2340 → #2347, `b4ebd77`
- [x] F1 — #2341 (blocked by A, C) → #2361, `233a34d`
- [x] F2 — #2342 (blocked by A, B, C, E) → #2360, `51ed621`

## Rules for every child

- Move-dominated PRs with rename-only proof; pure moves carry their tests unchanged; gate edits land in a final `chore(gates)` commit.
- No compatibility re-exports at the old path (`no-internal-package-reexports`); every consumer switches to the owning specifier.
- New package subpaths point directly at moved source files so the eager-closure gate carries each baseline through rename detection; no `APPROVED_OVER_CEILING` row without a measurement.
- Ownership classification (which daemon module owns which platform-aware responsibility) stays in #2278 and its children (#2332, #2333, #2334). No child here classifies ownership.

## Evaluated and not scheduled

[...]

PR #2346, opened 2026-09-06
refactor(daemon): split src/daemon/types.ts into request types and session state

### LINKED ISSUE #2338: refactor(daemon): split src/daemon/types.ts into request types and session state
Part of #2335.

## Purpose

`src/daemon/types.ts` (405 lines) has 197 type importers serving two audiences: the request shape (`DaemonRequest` 112 importers, `DaemonResponse` 111, `DaemonInvokeFn` 14) and live session state (`SessionState` 112). 79 files import only request types, 35 only session types, 83 both. It sits in one real type cycle: `types.ts` imports `RefFrame` from `ref-frame.ts`, which imports `SessionState` (documented at `types.ts` line 18). `session-script-publication-state.ts` has no imports and is not in a cycle.

## Required behavior

- Split into a request module (`DaemonRequest`, `DaemonRequestMeta`, `DaemonRequestInternal`, `DaemonOpenLifecycle`, `DaemonResponse`, `DaemonResponseData`, `DaemonInvokeFn`, ...) and a session-state module (`SessionState`, `SessionRef`, `SessionScope`, `InteractionSurfaceEntry`, `PendingInteractionOutcome`, `PostGestureStabilization`, ...). Names open.
- Export a wire-only request shape (`DaemonRequest` without `internal`, `meta` narrowed to public fields) that does not reference `SessionState` or `DeviceLease`. F1 consumes it. `DaemonRequestInternal` stays reachable only from the request module.
- Break the `ref-frame.ts` cycle by declaring the ref-frame slot type below both files. `ref-frame.ts` keeps construction authority: its R7 rows `refFrame` and `snapshotScopeSource` in `scripts/layering/session-state.ts` are unchanged.
- No behavior change; every importer switches to the module that owns the symbol it uses.

## Completion conditions

- Before/after inventory of the R9 largest type cycle (from `ratchet-reference.ts`) attached to the PR; the `types.ts` ↔ `ref-frame.ts` cycle is gone and no new cycle appears.
- Importer split reported (files importing request-only, state-only, both) after the change; the mixed count is a measurement, not a target.
- R7, R9, R10 ratchets green; `DAEMON_MODULARITY_BASELINE.externalDaemonTypesImporters` (`client/client-normalizers.ts`, `remote/daemon-artifacts.ts`) points at the right module.
- A type-level test proves the wire-only request shape has no `internal` key and no path to `SessionState`.

## Dependencies

None. F1 and F2 depend on this issue.

## Non-goals

The `SessionState` field-ownership program (R7 owners); moving any type into a package. An R9 pass alone is not proof of decoupling; the wire-only shape and the cycle inventory are.

thymikee:
## Summary

Closes #2338. `src/daemon/types.ts` served the dispatch request shape and the live session record from one file, and sat in the only daemon type cycle. Three modules replace it:

- `src/daemon/daemon-request-wire.ts` — **`DaemonWireRequest`**: no `internal` key, no property path to `SessionState` or `DeviceLease`, satisfying the wire-only requirement in #2338.
- `src/daemon/daemon-request.ts` — `DaemonRequest` = wire + `internal` (`DaemonRequestInternal` stays unexported), plus `DaemonResponse`/`DaemonResponseData`/`DaemonInvokeFn`/`DaemonArtifact`.
- `src/daemon/session-state.ts` — `SessionState`, `SessionRef`, `SessionScope`, `SessionRuntimeHints`, and the interaction-surface shapes.

For F1 (#2341): `DaemonWireRequest` lives under `src/`, so a journal **package** cannot import it under R11. The importable equivalent is `DaemonRequest`/`DaemonRequestMeta` from `@agent-device/kernel/contracts`, which already carry all four fields the journal reads (`command`, `flags`, `meta.requestId`, `meta.clientArtifactPaths`).

The cycle is cut by `ref-frame-slot.ts`, declared below both `ref-frame.ts` and `session-state.ts`. It owns the frame value (class still unexported, so the type stays nominal and unconstructible outside); `ref-frame.ts` keeps every transition and `session.refFrame` write — R7 rows `refFrame` and `snapshotScopeSource` are unchanged.

No behavior change; no re-export shim at the old path. **408 files touched, 386 of them import-line-only** — the rename/move exemption applies to the diff budget.

**R9 type cycles (production graph), measured on the current base `b4ebd77`:** 19 → 18. `src/daemon/ref-frame.ts ↔ src/daemon/types.ts` is gone; the other 18 are byte-identical, so no new cycle appeared. Largest cycle unchanged at 6 files (`packages/provider-webdriver/*`).

**Production importer split after the change:** 197 files — 82 request-only, 42 state-only, 73 both (issue baseline: 79/35/83).

Path-keyed gates follow the declarations, in one final `chore(gates)` commit: the daemon-modularity external-importer ratchet (membership unchanged — `client/client-normalizers.ts` and `remote/daemon-artifacts.ts`, both on `daemon-request.ts`) and the wire-compat closure waivers. R7 now finds `SessionState` by its declaration rather than a recorded path, because the same function measures the merge-base tree.

Three test files already over the 1,000-line ratchet gain one import line each and shed one redundant line to stay at base length — no assertion weakened. Splitting them along the modules they mirror is tracked in #2353.

## Validation

Rebased onto `b4ebd77`. One conflict, in `daemon/deferred-interaction-outcome.ts`: #2347 retargeted `runPostGestureStabilityLoop` to `@agent-device/capture-kit/post-gesture-stability` while this PR split the type import on the next line — resolved by taking both. (The earlier rebase onto `835af32` had one conflict in `handlers/session-app-source-deployment.ts`, resolved the same way.)

Tested at `6f2fe7b` (the pushed head) after `pnpm install --frozen-lockfile && pnpm build`.

`pnpm check:affected --run`: format, lint, typecheck, layering, di-seams, fallow, gate-manifest, build, package, integration and macos-coverage all pass. Layering guard green against merge-base `b4ebd77`: R7 (31 fields classified), R9 (largest cycle 6), R10 (R7 pressure at base, 2 external importers). Fallow: no issues in 408 changed files.

`vitest-related`: 3149 passed, 3 failed, none from this change. `app-log-session-resource.test.ts > unwritable tombstone…` and `durable-capture-resource.test.ts > an unconfirmed failed-adoption cleanup…` fail identically on `origin/main` in this container, which runs as root, so their `chmod`-based unwritable-tombstone setup cannot deny a write. `cli-config-trust.test.ts > project config rejects remote connection fields…` is untouched by this diff and passes in isolation here — a load artifact of the local run. CI is the authority on all three.

The type-level test is `src/daemon/__tests__/daemon-request-wire.test.ts`; I verified it fails `tsc` when an `internal`/`SessionState` path is reintroduced.

https://claude.ai/code/session_01Ujrc8LYmvM249WY8921J1Y

---
_Generated by [Claude Code](https://claude.ai/code/session_01Ujrc8LYmvM249WY8921J1Y)_


thymikee:
The type split and ref-frame ownership look sound at 688b1a5, but Coverage found two changes this PR still needs:

- Move the wire-closure declarations in test/wire-compat/closure-policy.ts off the deleted daemon/types.ts path. The gate now reports the request/response/artifact aliases at daemon-request.ts as unclassified. Preserve the existing wire guarantees rather than adding a broad waiver.
- Fix the test-size ratchet regressions in snapshot-handler.test.ts, find.test.ts and session-replay-repair-transaction.test.ts. Each grew by one import line while already over the limit; do not raise the limits.

Please rerun those focused gates and update the validation note. These failures are distinct from the reported root-user permission tests.

thymikee:
Both are fixed in `b954217`, pushed just before your comment landed — so this was written against `688b1a5`.

**Wire closure** (`chore(gates)`): the three existing keys move from `src/daemon/types.ts#…` to `src/daemon/daemon-request.ts#DaemonRequest`/`#DaemonResponse`/`#DaemonArtifact`. No new or widened waiver — same three declarations, same guarantees. `DaemonRequest`'s rationale now describes it as the server-side narrowing of the kernel declaration that fixes the wire shape, rather than calling it a re-export alias (it never was one).

**Test-size ratchet** (`test(daemon)`): no limit raised and no gate edit. Each of the three files sheds one line that carried nothing, putting it back at exactly its merge-base length (1656, 1199, 1202):

- `snapshot-handler.test.ts` and `find.test.ts` each drop a `toHaveLength` assertion that the adjacent `toEqual` on an explicit array literal already makes.
- `session-replay-repair-transaction.test.ts` names the filtered close actions instead of wrapping the expression across three lines inside `expect`.

No assertion weakened. Splitting these files along the modules they mirror is the standing remedy, but none of those source modules split in this PR, so it stays out.

Focused gates at `b954217`:

```
pnpm exec vitest run --project unit-core \
  test/wire-compat/wire-compat.test.ts \
  scripts/__tests__/test-file-size-ratchet.test.ts
→ Test Files 2 passed (2), Tests 9 passed (9)
```

Validation note in the PR body updated to `b954217`, and it now separates these two from the root-user permission tests.

---

thymikee:
The follow-up b954217b addresses both gate findings. Wire-closure entries follow the moved declarations, and the test files are back within their existing limits without losing assertions. The type split and ref-frame ownership remain clean.

Ready for human review; the new head still has CI running, so merge readiness awaits those results.

thymikee:
Update: Android Smoke has now failed on b954217b at wait for Alert result: cancelled, the same canary failing on other PRs. The two code/gate findings are resolved, but I removed ready-for-human until the required smoke lane passes.

thymikee (commented):
Adversarial review from the #2335 coordinator, at `b954217`. I checked the load-bearing claims against the tree rather than reading the body. The work is sound; one finding below changes #2341's plan and should be recorded before this merges.

**Verified independently:** `src/daemon/types.ts` is gone with no re-export shim; zero suppressions anywhere in the diff; `DaemonWireRequest` imports only `@agent-device/contracts/command`, `@agent-device/kernel/contracts` and `@agent-device/kernel/device` — there is genuinely no import path from it to `session-state.ts`, so the headline claim holds structurally and not just by the type test.

**I checked the three ratchet-balancing edits in `beaaef7` specifically, because "shed one line to stay at base length" is exactly where an assertion quietly dies.** It does not here. In `snapshot-handler.test.ts` and `find.test.ts` the dropped `toHaveLength(n)` is strictly implied by the adjacent `toEqual` on an explicit n-element array literal — including `find.test.ts`, where the cap of 5 is the point of the test and remains pinned by the literal. The third is a pure two-line reformat of the same `toHaveLength(0)`. The commit body's "no assertion is weakened" is accurate.

That said, this is ratchet-driven editing of tests the ratchet wants *split*, and you say so honestly. Please open a follow-up for splitting those three files along the modules they mirror, and reference it in the commit body, so the deferral is tracked rather than remembered.

**Finding — `DaemonWireRequest` cannot serve #2341 where it lives, and #2341 does not actually need it.**

#2341's completion conditions are that the session-journal package "imports nothing under `src/` (R11)" and that its input type has no path to `SessionState` or `DeviceLease`. `DaemonWireRequest` is declared at `src/daemon/daemon-request-wire.ts` — under `src/` — so a workspace package may not import it. As exported, it cannot be F1's input type.

It also does not need to be. `@agent-device/kernel/contracts` already exports `DaemonRequest` and `DaemonRequestMeta` carrying exactly the four fields the journal reads (`command`, `flags`, `meta.requestId`, `meta.clientArtifactPaths`), in a package F1 may depend on, with no `internal` key and no session-state path. Your own module is built by narrowing that very type. So F1 should type its inputs on `@agent-device/kernel/contracts` directly, and #2341's "or kernel/contracts public request/meta types" clause is the path that actually applies rather than the fallback it was written as.

Nothing to change here — this PR satisfies #2338 as written, and `DaemonWireRequest` remains the right shape for `src/`-side consumers. I am recording it so F1 is not built against an import it cannot legally make. One line in your Summary naming kernel/contracts as F1's actual source would carry that forward.

I am keeping F1 sequenced behind this PR regardless: it relocates `session-event-*.ts`, which your 408-file diff also touches, so the ordering still buys conflict avoidance even though the type dependency has dissolved.

Two `chore(gates)` commits with a test commit between them departs from the "implementation first, one final `chore(gates)`" convention — both later commits are CI fixes, and this repo squash-merges, so I would not rewrite history for it. Noting it only so the pattern is not taken as precedent.

---

thymikee:
This head now conflicts with main. The code review remains clean, but conflict resolution and a passing Android smoke lane are needed before merge.

thymikee:
**Smoke Tests failure on `6f2fe7b` is not this PR's.** Re-running it once; no code change from me.

Failing check: [Smoke Tests / iOS lane](https://github.com/callstack/agent-device/actions/runs/34026023647/job/101467087341) — `RunnerTests.testAlertCannotProveAnIdenticalReplacementAndDoesNotActivateIt()`, 2 failures out of 81, "0 unexpected", `TEST EXECUTE FAILED` (exit 65).

Why it is not this PR's:

- That test is a **Swift XCUITest** in `AgentDeviceRunnerUITests`. This PR is a TypeScript type split — it touches no Swift, no runner source, and changes no runtime behavior at all. There is no mechanism by which it moves an XCUITest result.
- The same lane **passed on `0fa63fe`**, the immediately preceding head of this branch, whose TypeScript is materially the same. Red → green → red on equivalent code is intermittency, not a regression introduced here.
- The subject area — proving an alert replacement is identical and not re-activating it — is the same one `27a97ee` (#2326, "confirm alerts without repeating activation") landed in on `main`.

Note this is a **different lane** from the earlier failure on this PR. That one was the Android emulator smoke (`wait text "Alert result: cancelled"`, 10s wait timeout on a live emulator) and it has since passed. Two different device lanes failing intermittently around alert handling, on a branch that cannot affect either, is worth someone looking at independently of this PR — I am not widening this one to chase it.

No fix exists to port: I checked the open PRs and none targets this XCUITest. If the re-run comes back red I will say so rather than re-running again.

---

thymikee:
Independent review at `6f2fe7b689ab11e5dcea3e579294872ca10e9bb8`: clean. The type split preserves the wire-only request boundary, and the nominal ref-frame value remains isolated while `ref-frame.ts` retains the transitions and session-slot writes. The corrected wire and size gates are sound; no remaining findings. All completed checks are green; iOS Smoke is still in progress, so merge readiness waits on that exact-head result.

Issue #2342, opened 2026-09-06
refactor(daemon): relocate the daemon client out of src/daemon


thymikee:
Part of #2335.

## Purpose

`src/daemon/client/` (9 files, 2,416 lines; zone `daemon-client`, rank 5) is the daemon's client, not the daemon. No daemon file imports it; its consumers are `src/cli.ts`, `src/agent-device-client.ts`, `src/cli/commands/proxy.ts`, `src/commands/management/output.ts` (via B) and `src/cli/injected-daemon-dispatch.ts` (type). Relocating it makes `src/daemon` server-only. This issue is a relocation; it does not promise a package.

## Remaining server and root seams (inventory to carry into the PR)

- `src/daemon/config.ts` (113, pure over host-kit/kernel), `src/daemon/daemon-process.ts` (122, host-kit), `src/daemon/request-progress-protocol.ts` (48, imports the request types C splits).
- `src/daemon/code-signature*.ts` — moved by E.
- `src/daemon/session-store.ts` → `findUnrecoveredRepairCommitFailure` from `daemon-client-lifecycle.ts:17`: a disk reader of session artifacts, to be extracted below both client and store.
- `src/core/lease-scope.ts` (258, kernel-only) from the RPC layer.
- `src/remote/daemon-artifacts.ts` (imports daemon types and `recording/output-path.ts`) and `src/remote/remote-request-diagnostics.ts` (imports `session-store.ts`).
- `src/command-catalog.ts`, `core/command-descriptor/registry.ts`, `timeout-policy.ts` — A's package.
- `src/daemon/types.ts` wire types — C's request module.
- `src/daemon/client/doctor-progress.ts` — deleted by B.

## Required behavior

- Move the directory to `src/daemon-client/` (zone name unchanged; `targetDagZone` in `scripts/layering/model.ts` updated from the `src/daemon/client/` prefix). Rank stays 5.
- The client's imports of `src/daemon/*` are reduced to the named shared kernel: `config.ts`, `daemon-process.ts`, `request-progress-protocol.ts`, the extracted session-artifact reader, and C's request module.
- No behavior change in daemon launch, takeover, progress streaming or code-signature checks; `test/integration/smoke-cli` and the daemon lifecycle tests pass unchanged.

## Completion conditions

- `src/daemon/client/` no longer exists; `src/daemon` contains only server code plus the shared kernel modules listed above.
[...]

PR #2347, opened 2026-09-06
refactor(daemon): move four pure leaves to their kits

### LINKED ISSUE #2340: refactor(daemon): move four pure leaves to their kits
Part of #2335.

## Purpose

Four `src/daemon` leaves have no value dependency on the daemon or root `src/` and a kit whose dependency set already covers theirs. Each move is a rename-only PR.

## Required behavior

| leaf | lines | dependencies | destination |
|---|---|---|---|
| `parameterized-recorded-fill.ts` | 472 | `contracts/replay` (type), `selectors` | `@agent-device/selectors` (not ad-script: selectors depends on ad-script, and this file imports selectors) |
| `code-signature.ts` + `code-signature-cache.ts` | 385 | `host-kit/version`, node `crypto`/`fs`/`path` | `@agent-device/host-kit` |
| `screenshot-overlay.ts` + `screenshot-overlay-draw.ts`, plus `src/snapshot/screenshot-overlay/{rects,android}.ts` (only these two import them) and `src/core/react-native-overlay.ts` (kernel + `contracts/react-native-overlay` only) | 558 + 172 + 304 | `kernel/snapshot`, `contracts/snapshot`, `capture-kit` png / png-worker-client / snapshot-rect-projection, node `fs` | `@agent-device/capture-kit`; `handlers/react-native.ts` and `commands/react-native/overlay.ts` switch specifier |
| `post-gesture-stability.ts` | 258 | `host-kit/diagnostics`, `host-kit/retry`; generic over `S`/`T`, no daemon types | `@agent-device/capture-kit` |

Excluded: `session-target-evidence.ts` (300 lines) needs `capture-kit/ios-snapshot-engine` and `ad-script`; no package depends on both, so it has no home without a new package edge.

Consumers to update: `interaction/internal/interaction-common.ts`, `session-action-recorder.ts` (fill); `client/daemon-launch-spec.ts`, `server/server-lifecycle.ts` (signature); `android-snapshot-timeout-evidence.ts`, `sparse-fallback-screenshot.ts`, `generic-runtime-execution.ts` (overlay); `deferred-interaction-outcome.ts` (stability).

## Completion conditions

- Each leaf is a git rename with tests carried unchanged; no re-export shim left in `src/daemon`, `src/snapshot` or `src/core`.
- Each new package subpath points directly at the moved file; eager-closure budgets green with no new exception row.
- `check:layering`, package-boundaries `exportTargets`, fallow (path-keyed baselines moved, not regenerated) green.
- None of these files carry an R7 owner row, so the session-state ratchet is unaffected; the PR confirms this.

## Dependencies

None. F2 depends on the code-signature move.

thymikee:
## Summary

Closes #2340. Four `src/daemon` leaves move to the kit whose dependency set already covered them; each is a git rename carrying its tests.

- `parameterized-recorded-fill.ts` → `@agent-device/selectors/parameterized-recorded-fill`
- `code-signature.ts` + `code-signature-cache.ts` → `@agent-device/host-kit/code-signature{,-cache}`
- `screenshot-overlay{,-draw}.ts`, `snapshot/screenshot-overlay/{rects,android}.ts`, `core/react-native-overlay.ts` → capture-kit, exposing `./screenshot-overlay` and `./react-native-overlay`; draw/rects/android stay package internals
- `post-gesture-stability.ts` → `@agent-device/capture-kit/post-gesture-stability`

`session-target-evidence.ts` is untouched and no re-export shim is left behind.

One behaviour fix rides along, because the moves exposed it (review thread): `walkDaemonCodeGraph` followed relative specifiers only, so moving daemon code behind workspace specifiers dropped it — and the walker itself — out of the source-checkout signature. It now resolves scoped specifiers through the owning package's `exports` map, stamping the manifest so a retarget invalidates. Graph from `src/daemon.ts`: 619 modules on `main`, 611 after the moves alone, 1459 now.

Issue correction: #2340 names `sparse-fallback-screenshot.ts` and `generic-runtime-execution.ts` as overlay consumers, but neither imports it — the real consumers are `screenshot-runtime.ts` and `android-snapshot-timeout-evidence.ts`.

Package `exports` additions sit in their move commits so each compiles and bisects; the R11 export-list pins are the `chore(gates)` commit.

## Validation

Full 58-check sweep at `41dec50`; re-run at `266a1ee`: format, lint, typecheck, layering (186), eager-closure (454), fallow, production-exports and the signature/daemon-client suites (89) all green.

`git diff -M90% --stat origin/main...HEAD`: every moved file a rename whose changed lines are specifier switches only.

`unit`/`vitest-related` fail on 3 durable-capture tests that fail identically at `origin/main` 27a97ee and predate this branch; no open PR fixes them.

Eager-closure green with no `APPROVED_OVER_CEILING` row — rename detection carries each baseline, closures unchanged (56/5/19/32/4/8).

R7 unaffected: `SESSION_STATE_FIELD_OWNERS` in `scripts/layering/session-state.ts` names none of the moved files.

Fallow green; no baseline entry is keyed to a moved path, so none needed moving.

https://claude.ai/code/session_01BPFjVXwbDrXPqp5W6K2iDK


thymikee:
The moves are otherwise clean at c4f5633f, but they break source-checkout daemon invalidation. walkDaemonCodeGraph follows only relative imports. The moved signature, overlay, fill and stability modules now sit behind workspace specifiers, so edits to them no longer change the running daemon signature. The cache also relies on the signature walker itself being in that graph, which is no longer true.

Please fix workspace-export resolution at the signature owner, reusing the existing resolution machinery where possible. Add a regression proving that a package-owned runtime edit changes the source signature, including invalidation when the walker changes. Do not restore cross-package relative imports to work around it.

thymikee (commented):
Adversarial review from the #2335 coordinator. I reviewed this branch at `41dec50` before the PR opened and raised one blocking finding directly with the author; recording the outcome here for the record, now confirmed at `c4f5633`.

**The finding was:** `packages/capture-kit/src/snapshot-state.fixtures.ts` duplicated `makeSnapshotState` byte-for-byte from `src/__tests__/test-utils/snapshot-builders.ts:12` and carried `// fallow-ignore-next-line code-duplication` to silence the finding. AGENTS.md is explicit that gate failures are diagnosed at their reported invariant, not suppressed.

**The resolution is right, and better than what I asked for.** Rather than assume the suppression was load-bearing, the author measured: `fallow dupes` reports 3 clone groups on this tree and the fixture is in none, so the suppression was never doing anything. It is gone at `c4f5633` and the gate passes without it. That is the correct answer — I had assumed the ignore was covering a real finding, and the measurement showed it was not.

The remaining duplication is a package-local test fixture, which is the shape this repo already uses (`tmp-dir.fixtures.ts` in capture-kit and host-kit) for the good reason that a package test cannot reach root `src/__tests__`. Pressing further would turn a rename-only move into a refactor of shared test utilities, against this issue's own non-goals. Settled.

**Verified independently at `c4f5633`:** all eight moved modules are true renames under `-M90%` whose content deltas are import-specifier rewrites only; four implementation commits plus one final `chore(gates)`; `session-target-evidence.ts` correctly untouched; `parameterized-recorded-fill.ts` landed in `@agent-device/selectors`, avoiding the ad-script trap the issue flags; no re-export shim in `src/daemon`, `src/snapshot` or `src/core`. I independently confirmed both completion claims I would otherwise have taken on trust: `SESSION_STATE_FIELD_OWNERS` in `scripts/layering/session-state.ts` names none of the moved files, and no `fallow-baselines/` entry is keyed to any moved path.

The consumer correction in the Summary is right and I have verified it: on `main`, `sparse-fallback-screenshot.ts` and `generic-runtime-execution.ts` do not import `screenshot-overlay`; `screenshot-runtime.ts` and `android-snapshot-timeout-evidence.ts` do. That is a defect in #2340's text, not in this PR. I will carry it to the umbrella, because #2342 hand-lists its seams the same way and will need the same treatment against the code rather than the issue.

Keeping `exports` additions with their move commits is the right call for bisectability and is now stated. CI is green across every completed check.

---

thymikee:
Confirmed and fixed in 266a1ee. Walking `src/daemon.ts` measures the regression exactly:

| tree | modules | walker stamped | overlay stamped |
|---|---:|---|---|
| `main` 27a97ee | 619 | yes | yes |
| c4f5633 (moves only) | 611 | **no** | **no** |
| 266a1ee | 1459 | yes | yes |

So the cache comment claiming "`code-signature.ts` is itself inside the graph it walks" was falsified by the moves, as you said.

`walkDaemonCodeGraph` now resolves a scoped specifier through the owning workspace package's `exports` map and walks into the file it names, reusing `resolveRelativeImportPath` for the target so candidate probing and `absentPaths` are unchanged. The manifest is stamped rather than merely probed, so an `exports` retarget invalidates without either endpoint changing; an uninstalled package becomes an absent path. Installed dependencies are still not followed — they change on install, not on edit — and the test is structural (a workspace link resolves out of `node_modules`) rather than a name pattern. No cross-package relative import was added.

Cost: the cold walk goes 619→1459 modules, ~30ms→~112ms, which is what the stat-validated cache absorbs; the warm path is unchanged.

Regression: `packages/host-kit/src/code-signature.test.ts` (6 tests) plus a package-owned-edit case in the cache suite. Five of the six fail against the previous walker; the sixth is the negative guard that installed dependencies are not followed. One test stamps the real daemon graph and asserts `packages/host-kit/src/code-signature.ts` is in it, which is the walker-invalidation property directly.

`walkDaemonCodeGraph` tripped fallow's complexity threshold after the change; the workspace branch is extracted into `followWorkspaceSubpaths`/`resolveWorkspaceOwner` rather than suppressed.

---

thymikee:
The workspace source-signature issue is resolved in 266a1ee. Package source edits and export retargets now invalidate the cached signature, and the regression tests cover both paths. No remaining code findings. The checks completed so far pass; remaining CI is still running.

thymikee (commented):
Second review pass from the #2335 coordinator, on `266a1ee` — the source-signature fix. This is the most consequential change in the whole wave and it deserves to be read as such rather than waved through as part of a move PR. Two things to settle before merge; the fix itself is right.

**What it actually found.** `walkDaemonCodeGraph` followed relative specifiers only, so the daemon's signature covered whatever the daemon still imported by relative path — 619 of its modules. That was already wrong on `main`, independent of this PR: more than half the daemon was invisible to the signature a client compares a running daemon against. The leaf moves pushed the walker, the overlay, the recorded-fill and the stability loop out of the relative graph, taking it to 611 and, worse, taking the walker itself out of the graph it walks, which is the exact property `code-signature-cache.ts`'s format guard rests on. Stamping the manifest rather than probing it, so an `exports` retarget invalidates without either endpoint changing, is the right call and closes the subtlest hole. Five of six new walker tests failing against the old walker is real regression proof.

**1. Say plainly that this exceeds the move's blast radius.** Restoring parity would mean getting back to ~619-module coverage. You went to 1459 — you fixed a pre-existing defect, not just the regression the moves caused. I think that is the correct choice and should stay in this PR: splitting it would knowingly ship a branch whose signature is *more* blind than `main`'s, and AGENTS.md says to repair at the owning construction path rather than route around it. But #2340 is a rename-only issue, and a reviewer scanning for renames will not expect a 202-line behavior change to `code-signature.ts`. State it in the Summary as a scope expansion with its reason, per the "note any expansion of scope" requirement — right now the body still reads as four pure moves.

**2. The startup numbers in the size report predate this commit.** The +0.2 ms `--version` / −1.8 ms `--help` figures were measured at `c4f5633`, before the walker changed. The walk is on the daemon-client hot path: every invocation stat-validates the cached graph, and that graph went from 619 to 1459 entries — roughly 2.4× the stat calls per command, plus whatever `absentPaths` grew to. The 112 ms figure you quote is the cold walk, which the cache absorbs; the number that matters for CLI latency is the **cached** revalidation cost at `266a1ee`. Please post the refreshed startup medians from the size report on this head, and the `absentPaths` count before and after. If `--version` is unchanged, that settles it in one line.

Minor, no action needed: `SCOPED_SPECIFIER_RE` will match `@scope/name` inside prose, comments and fixture strings, and unlike a stray relative specifier each miss costs a durable `absentPaths` entry the cache re-stats forever. Restricting to the scoped shape bounds this sensibly and the comment explains why — I mention it only so the trade-off is on the record.

**Sequencing consequence for the umbrella.** #2348 (A) creates `@agent-device/command-registry` and moves 12 modules into it — the same class of edge this walker could not follow. Its own thread already refers to a "shared source-signature prerequisite from #2347", so this fix is now load-bearing for A as well. That is a dependency #2335 does not declare (it lists A and E as independent). Either this lands before A, or A carries the same defect. I am recording it on the umbrella and sequencing E ahead of A accordingly.

---

PR #2350, opened 2026-09-06
test(android-e2e): record rotation state and logcat rotation decisions on a failed step

thymikee:
## Summary

Diagnostics for the Android Smoke failures on the post-alert canary (`get text id="automation-alert-result"` since 2026-09-03, `wait for Alert result: cancelled` on #2344).

The failed-step screenshot and snapshot from https://github.com/callstack/agent-device/actions/runs/34021894996 show the device in landscape at the timeout: `Open automation alert` is 2220px wide and the alert-result canary sits below the fold, so 16 readable polls never contained it. Every tap before the alert landed at x=540, so the device was portrait through `open alert`; `orientation portrait` had been confirmed by the fixture 12s earlier. The same layout explains the earlier `Selector did not match` failures on that step. Nothing in the artifacts says what rotated it: the emulator's Vulkan surface events happen ~95s after boot in passing runs too, and the guest settings alone never flip a local emulator.

## Change

A failed step now also writes `failed-step-N-device.txt` next to the screenshot and snapshot, through adb rather than agent-device so it stands even when the CLI path is what failed:

- `settings get system accelerometer_rotation` and `user_rotation` (the two values `orientation` writes)
- the rotation lines of `dumpsys display`
- WindowManager's rotation decisions from `logcat -d`, with the reason Android logs for each

The shared harness gains an optional `deviceEvidence` hook; iOS does not implement it yet.

## Validation

Probe commands verified against a local emulator (settings, display rotation lines, logcat rotation decisions all return output). Root typecheck, oxlint, oxfmt clean. The evidence itself only appears on a failed CI step, which the Android lane has been producing on roughly half its runs.


thymikee:
The diagnostics are useful, but 825faf09 needs a small follow-up before merge:

- Fallow fails on captureFailedStepEvidence. Extract the evidence collectors at the shared harness boundary rather than suppressing the gate.
- Add a deterministic failed-step test that verifies the device file contents and its path in failed-step.txt, plus a hook-failure case that still collects the screenshot/snapshot. A passing Android smoke run does not exercise this branch.
- Bound the optional probes as a group. Four sequential 20-second probes currently delay the existing screenshot/snapshot by up to 80 seconds, when the failure state may already have changed. Keep that collection from blocking the existing evidence and cap logcat line lengths as well as line count.

thymikee:
The iOS smoke lane also finished red at the depth-1 bridge assertion, after target-resolution-failed fell back to XCTest. This is outside the Android diagnostics diff; the existing collector/test findings remain, and the required lane still needs to pass.

thymikee:
Addressed:

- The collectors live in `test/integration/live-device-e2e/failed-step-evidence.ts`; the harness only builds the input (stem, bound CLI, optional hook). Fallow is clean on the changed files.
- `test/integration/live-device-e2e-failed-step-evidence.test.ts` (node:test, runs with the integration node suite): device facts land in `failed-step-N-device.txt` with the exact contents next to the screenshot and snapshot; a hook that throws or returns nothing leaves the CLI evidence in place; a hook that never answers is bounded (50ms in the test) and does not delay the CLI evidence; and through `createLiveDeviceHarness`, a failing step's message and `failed-step.txt` name all three files.
- The device hook runs alongside the screenshot and snapshot, not before them, bounded as a group at 15s. Each adb probe is bounded at 5s, logcat lines are capped at 60 and 240 chars, display lines at 8 and 240 chars.

Main merged in for the typecheck break that failed the iOS lane's build step here.

thymikee:
The prior findings are resolved: failed-step device evidence is collected independently under a bounded collector and the failure paths are covered by regression tests. CI is clean; no remaining code findings.

Issue #2334, opened 2026-09-06
refactor(daemon): separate open-target planning from platform mechanics


thymikee:
## Purpose

Child of #2278 (audit area: open-target planning). `src/platform-runtime-open-target.ts` mixes a neutral open plan/result policy with platform mechanics:

- neutral: `resolveRequestedOpenSurface`, `validateOpenRelaunchTarget`, `resolveSessionAppBundleIdForTarget`
- Apple mechanics: `resolveSoleForegroundIosApp` (sole-booted-simulator/foreground-app probe)
- Android mechanics: `resolveAndroidPackageForOpen`, `inferAndroidPackageAfterOpen`

Two classified daemon edges (R74 inventory) consume this file:

- `src/daemon/session-lifecycle/internal/session-open-prepare.ts` → `resolveRequestedOpenSurface`, `validateOpenRelaunchTarget`
- `src/daemon/handlers/session-selector-dispatch.ts` → `resolveAndroidPackageForOpen`, `resolveSessionAppBundleIdForTarget`

## Required behavior

- Separate the pure open plan/result from the platform implementation only when that creates a deeper interface and removes duplicated policy; preserve one construction path (one resolver producing the open plan — no second).
- The daemon `open` prepare and selector-dispatch code imports only the neutral plan/policy surface.
- Android package resolution moves behind the Android owning seam (package facts/runtime interface in `packages/platform-android`); the iOS probe is owned by #2332 (this issue depends on #2332 for the shared file `src/platform-runtime-open-target.ts`).
- If the neutral surface stays a root module, its remaining daemon consumers are classified `daemon-policy-essential` in the R74 inventory with the plan-construction rationale; if it moves, the move is a pure seam change with no behavior change.

## Observable completion conditions

- The two R74 inventory edges above are reclassified or removed.
- `resolveAndroidPackageForOpen` / `inferAndroidPackageAfterOpen` have no production importer outside their owning seam.
- `open` prepare and selector dispatch behave identically (existing behavior tests).
- Planted violation: a daemon import of the platform-mechanics symbols is red.
- `pnpm check:layering` and `pnpm check:affected --run` pass.

## Dependencies

Blocked by: #2332 (shared file `src/platform-runtime-open-target.ts` hosts the probe #2332 relocates).

Issue #2333, opened 2026-09-06
refactor(daemon): normalize lifecycle participation of platform resource owners


thymikee:
## Purpose

Child of #2278 (audit area: runtime lifecycle participation). The daemon startup/shutdown code in `src/daemon/server/daemon-runtime.ts` explicitly names platform resource owners through three classified edges (R74 inventory):

- `src/platform-runtime-apple-runner-owner.ts` — `configureAppleRunnerLeaseOwnerStateDir`, `configureAppleRunnerDeviceClaimAuthorityProbe` (daemon-owned inputs pushed into the Apple runner host at startup, cleared on lock release)
- `src/platform-runtime-resource-cleanup.ts` — `resetAndroidSnapshotHelperRuntime`, `cleanupManagedWebRuntimeOrphans` (named platform owners invoked at startup; the `platformResourceCleanup` symbol on the same edge is the neutral `PlatformResourceCleanup` contract capability and is the model to follow)
- `src/platform-runtime-operation-host.ts` — dynamic `import()` of `recoverLegacyAppLogMarkersAfterDaemonLock` at startup (app-log legacy marker recovery, a platform process-mechanic)

The daemon retains lifecycle sequencing, cancellation, failure policy, and budgets. Platform resource owners should participate through a typed owning interface; no generic hook bag, service locator, or dispatcher.

## Required behavior

- `daemon-runtime.ts` stops importing the named platform owners above. The daemon-owned values (lease-owner state dir, device-claim authority probe, cleanup inputs) are still supplied by the daemon — what changes is the hand-off: a typed lifecycle participation surface (startup/shutdown phases) that the root composition wires the platform owners into, or the existing `PlatformResourceCleanup`-style contract capability extended to cover snapshot-helper reset, Web orphan cleanup, and app-log legacy marker recovery.
- The daemon keeps: ordering (runner-owner configuration after daemon lock acquisition; web-orphan cleanup and marker recovery before servers open; resets on the shutdown path), best-effort failure policy with diagnostics, and the existing teardown budgets.
- One construction path for the participation surface; platform owners are registered by the root composition, not by daemon code naming platform modules.

## Observable completion conditions

- The three R74 inventory edges for `daemon-runtime.ts` (apple-runner-owner, resource-cleanup, operation-host dynamic import) are reclassified or removed.
[...]

Issue #2332, opened 2026-09-06
refactor(daemon): own Apple runner session observation behind a semantic port


thymikee:
## Purpose

Child of #2278 (audit area: Apple selector and session observations). The daemon consumes Apple runner session health and the sole-foreground-app probe directly from root composition modules — four classified edges in `scripts/layering/daemon-platform-runtime-inventory.ts` (R74):

- `src/daemon/request-recording-health.ts` → `src/platform-runtime-apple-resources.ts` (`inspectAppleRunnerSession`)
- `src/daemon/session-device-resolution.ts` → `src/platform-runtime-apple-resources.ts` (`inspectAppleRunnerSession`)
- `src/daemon/ios-app-session-hint.ts` → `src/platform-runtime-open-target.ts` (`resolveSoleForegroundIosApp`, import and re-export)

Define the smallest semantic observation/query interface that supports the daemon contract without exposing Apple runner mechanics.

## Required behavior

- The two runner-health consumers need one observation: for a UDID, whether a live runner session exists and its current session id (recording-health invalidation; simulator-boot evidence for device refresh). The daemon today compares `snapshot.sessionId` against the recorded `runnerSessionId` and treats `alive: false` as invalidation — the port must preserve those decisions.
- The open-hint consumer needs the sole-foreground-app probe with its never-guess contract: exactly one booted simulator in the device set with exactly one running app, else `undefined`; probe failures are inconclusive, control-flow/composition failures propagate. The hint policy (when to emit, `MAX_HINT_LENGTH`, command composition) stays daemon-owned.
- The owning seam is a neutral observation interface: declared in `@agent-device/contracts` or a root-composed capability named in `scripts/layering/architecture-ownership.ts`, with the Apple implementation in `packages/platform-apple`. Data shape sketch:

```ts
// contracts (sketch)
type RunnerSessionObservation = Readonly<{ alive: boolean; sessionId?: string }>;
type SoleForegroundApp = Readonly<{ device: DeviceInfo; app: Readonly<{ bundleId: string }> }>;
type AppleSessionObservation = Readonly<{
  observeRunnerSession(deviceId: string): Promise<RunnerSessionObservation | undefined>;
  resolveSoleForegroundApp(params: { simulatorSetPath?: string }): Promise<SoleForegroundApp | undefined>;
}>;
```

[...]

Issue #2278, opened 2026-09-04
refactor(daemon): audit and retire residual semantic platform coupling


thymikee:
## Purpose

The platform-package migration removed direct production dependencies from `src/daemon/**` to concrete platform code, but it did not prove that every remaining platform-aware responsibility belongs in the daemon.

Audit the residual semantic coupling, preserve daemon-owned policy and orchestration, and retire only platform mechanics that still leak through root `src/platform-runtime-*.ts` composition modules.

## Current baseline

Measured on `origin/main` at `658f822c40d14568959c483e83c4def24a6257f0`:

- R65 reports no direct production imports from `src/daemon/**` to `src/platforms/**` or `@agent-device/platform-*`.
- Nine daemon files contain 13 production imports from root `src/platform-runtime-*.ts` modules.
- The declared daemon modules cover replay, session lifecycle, interaction, and session observability. Other concerns are not classified with the same ownership metadata.
- `SessionState` write ownership is ratcheted, but broad state reads and `SessionStore` authority are not.
- The ADR 0019 hop trace records approximately 24 files on both the press/Android and snapshot/iOS entry-to-platform routes. There is no accepted hop budget or owning issue.
- The completion record for #2132 left 16 handler-owned `SessionState` shape edges and 23 `SessionStore` authority edges for a fresh audit. No active issue currently owns that follow-up.

The 13 imports are a review inventory, not 13 presumed violations. A neutral runtime interface may be the correct owning seam.

## Classification

Classify every daemon import of a root `src/platform-runtime-*.ts` module as exactly one of:

1. **Composition-essential** — assembling neutral runtime implementations at the process root.
2. **Daemon-policy-essential** — request admission, session ownership, conflicts, locking, cancellation, teardown ordering, artifact publication, or response/event semantics.
3. **Leaked platform mechanics** — platform-family operations or state that should be owned behind a deeper runtime interface or adapter.

Record the reason at the owning declaration or in a machine-readable architecture inventory. Only category 3 creates implementation work.

### Apple selector and session observations

Review daemon knowledge of Apple runner session health, direct selector queries, device refresh, recording health, and foreground app hints. Define the smallest semantic observation/query interface that supports the daemon contract without exposing Apple runner mechanics.

Coordinate any implementation with #2273 and #2274. Do not independently add a second selector producer or duplicate snapshot producer policy.

### Runtime lifecycle participation

[...]

PR #2354, opened 2026-09-06
chore(gates): classify #2278 daemon-platform-runtime edges and ratchet handler session authority

thymikee:
## Summary

Structural no-regrowth checks for the #2278 daemon-platform coupling audit (companion PR carries the audit record and ADR 0022):

- **R76 daemon-platform-runtime-inventory** (scripts/layering/daemon-platform-runtime-inventory.ts): classifies every production daemon import of root src/platform-runtime-*.ts modules — 14 edges across 9 files (13 entries; ios-app-session-hint.ts holds an import and a re-export of the same symbol) — as composition-essential, daemon-policy-essential, or leaked-platform-mechanics. The layering check fails on any unclassified edge, on symbol-set drift between a file and its inventory entry, on dynamic imports whose bindings the inventory cannot name (rest/computed destructure, namespace or side-effect form), and on stale entries.
- **R75 session-authority-overlay** (scripts/layering/session-authority-overlay.ts): symbol-level SessionState-shape / SessionStore-authority overlay, measured inside measureRatchets; a membership ratchet holds the handler-owned file sets at or under the merge base (currently 14 shape / 22 authority files).

Category-3 implementation work is owned by the child issues (#2332, #2333, #2334; the selector seam by #2273/#2274). Part of #2278. 12 files.

## Validation

Tested at 031f747c39: pnpm check:layering green (215 tests; R76 13 classified entries, R75 14/22 handler-owned files). Both rules observed red against planted violations before acceptance: an unclassified daemon→root platform-runtime import (R76), a rest binding and a namespace import beside the recorded binding (R76), and a new handler-owned SessionState import (R75; staged, since the scan reads tracked files only). pnpm check:affected --run passes at 031f747c39. No device-facing changes.

### DIFF
--- a/scripts/layering/check.ts
+++ b/scripts/layering/check.ts
@@ -102,6 +102,11 @@ import { recordRuntimeRegistryJoinViolations } from './record-runtime-registry-p
 import { recordRuntimeDaemonMechanicsViolations } from './record-runtime-mechanics-policy.ts';
 import { checkDaemonPlatformBoundary } from './daemon-platform-boundary.ts';
 import {
+  checkDaemonPlatformRuntimeInventory,
+  DAEMON_PLATFORM_RUNTIME_EDGES,
+} from './daemon-platform-runtime-inventory.ts';
+import { checkSessionAuthorityOverlay, handlerOwnedOverlay } from './session-authority-overlay.ts';
+import {
   listTrackedPlatformZoneFiles,
   listTrackedProductionSources,
   listTrackedSrcUtilsFiles,
@@ -352,6 +357,9 @@ function report(
       (sum, count) => sum + count,
       0,
     );
+    const measuredOverlay = handlerOwnedOverlay(ratchets.sessionAuthority);
+    const handlerOwnedShapeFiles = measuredOverlay.shapeFiles.length;
+    const handlerOwnedAuthorityFiles = measuredOverlay.authorityFiles.length;
     process.stdout.write(
       `Layering guard: OK — ${files.length} source files satisfy R2 and contain no ` +
         `value-import cycles (both checked globally); the ranked target spine contains no ` +
@@ -363,8 +371,12 @@ function report(
         `${ratchets.largestTypeCycle.length} files (R9); ${daemonModularitySummary(reference)}; ` +
         `${packageBoundariesSummary(repoRoot)}; ${platformPackagePolicySummary()}; ` +
         `runtime facts remain the only device-command admission authority and daemon code cannot ` +
-        `manufacture narrowed runtime proof (R66); and R65 keeps production src/daemon free of ` +
-        `concrete platform imports in every executable and type-only form.\n`,
+        `manufacture narrowed runtime proof (R66); R65 keeps production src/daemon free of ` +
+        `concrete platform imports in every executable and type-only form; ` +
+        `${DAEMON_PLATFORM_RUNTIME_EDGES.length} daemon-to-root platform-runtime edges hold ` +
+        `their #2278 classification (R74); and the handler-owned SessionState/SessionStore ` +
+        `authority overlay holds at or under the merge-base (R75, ` +
+        `${handlerOwnedShapeFiles} shape / ${handlerOwnedAuthorityFiles} authority files).\n`,
     );
     return 0;
   }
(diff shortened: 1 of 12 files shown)


thymikee:
R75 hardcodes src/daemon/types.ts for SessionState, but #2346 moves that declaration to src/daemon/session-state.ts. After that merge the shape overlay silently stops measuring the imports it is meant to ratchet. Reuse the owning declaration discovery and add a planted-red test against the moved declaration path.

thymikee:
The SessionState move is handled now. One gap remains in R74: dynamic imports are recorded with an empty symbol set. The existing `daemon-runtime.ts` import of `platform-runtime-operation-host.ts` therefore admits extra named bindings without triggering symbol drift. Please capture those bindings (or fail closed for this case) and add a planted-red test that expands this existing import. Android smoke also failed restoring the automation route after system UI; that looks unrelated to this tooling-only change.

thymikee:
Addressed in 4a7c2b771d. The layering scanner now captures named bindings from destructured dynamic imports: `const { x } = await import(...)` records `x` on the edge, so expanding an existing dynamic import's destructure trips R74's symbol-drift check. The `daemon-runtime.ts` → `platform-runtime-operation-host.ts` entry now records `recoverLegacyAppLogMarkersAfterDaemonLock`; a namespace-form import captures no bindings and therefore drifts against the recorded set (fail-closed when bindings are hidden). Planted-red test: "R74 rejects an expanded destructured dynamic import on a classified edge" extends the existing import with a second binding and asserts the drift violation; scanner semantics are pinned by a model-level test. `pnpm check:layering` and `pnpm check:affected --run` are green. Agreed the Android smoke route-restore failure is unrelated to this tooling-only change.

thymikee:
R74 still has a bypass at 4a7c2b7. Adding a rest binding alongside the approved name, such as { recoverLegacyAppLogMarkersAfterDaemonLock, ...operationHost }, leaves the recorded symbol set unchanged while exposing every other export. The scanner silently ignores RestElement and computed properties. Reject unclassified bindings instead of dropping them, and add a planted-red regression using the existing classified import. CI is green, but this leaves the no-regrowth guarantee incomplete.

thymikee:
Fixed in 6764951056. The scanner no longer drops rest or computed destructure bindings: `destructuredDynamicImportBindings` now reports the destructure's residue, and the edge carries it as `bindingResidue`. R74 rejects the pair with a dedicated "unnameable dynamic-import binding" violation (not a symbol drift), because a rest or computed binding exposes exports the inventory cannot name. Planted-red regression using the existing classified import: `{ recoverLegacyAppLogMarkersAfterDaemonLock, ...operationHost }` now fails R74, and the computed-key variant fails too. Scanner semantics are pinned in model.test (named / rest / computed / namespace cases). `pnpm check:layering` and `pnpm check:affected --run` are green; the branch was also rebased onto current main, so the prior 4a7c2b7 is now 2ab6596.

thymikee:
The rest/computed fix works at 6764951, but R74 still misses a namespace import added beside the approved named import. It unions symbols by file and target, so the approved name keeps the union unchanged while a second const operationHost = await import(...) exposes all exports; the existing test only replaces the named import. Validate each dynamic import before aggregation and add a planted-red test with both forms present. This leaves the no-regrowth guarantee incomplete. The branch also now conflicts with main; resolve that before rerunning the affected checks.

thymikee:
Fixed at 031f747c39. R76 now validates every dynamic import of a classified pair before the per-pair symbol union: a namespace or side-effect import() (one that names no binding) marks the pair open-ended and is rejected with a dedicated "open-ended dynamic import" violation, so a second const operationHost = await import(...) beside the approved destructure can no longer expose the whole module while the union stays unchanged. Planted-red regression with both forms present: { recoverLegacyAppLogMarkersAfterDaemonLock } = await import(...) plus const operationHost = await import(...) on the same classified pair now fails (the named form alone still passes); the replace-form test now asserts the open-ended violation instead of drift. pnpm check:layering and pnpm check:affected --run are green. Note: rebase onto current main surfaced a rule-number collision — the new snapshot-assembly rule took R74, so the inventory rule is renumbered to R76 (R75 unchanged); PR body and ADR references synced.

thymikee:
The namespace-import bypass is fixed at 031f747: each dynamic import is checked before symbol aggregation, and the paired named-plus-namespace regression covers the missed case. No remaining code findings; the R76 rename also avoids the new snapshot rule. Ready for human review, with all current checks passing.

PR #2356, opened 2026-09-06
fix(android): return from orientation once the display reports the rotation

thymikee:
## Summary

Android `orientation` wrote `accelerometer_rotation` and `user_rotation` and returned at once, while the display rotated some time later. On the loaded CI emulator that takes seconds, and accessibility reads hang meanwhile.

Evidence: the Android Smoke on #2344, https://github.com/callstack/agent-device/actions/runs/34025424834. `wait text landscape` right after `orientation landscape-left` reported `wait_capture_stalled` with zero readable captures: the helper session request timed out at 5s, the one-shot fallback ran into the wait's deadline, and the failed-step snapshot taken right after was already in landscape (nodes 2400px wide). The rotation had landed; the wait's whole budget went to the transition.

## Change

`setAndroidOrientation` now polls `dumpsys display` for `mCurrentOrientation` to match the requested rotation, every 500ms up to 15s, before returning.

- A display that reaches the rotation returns as soon as it does (about 1s on a local emulator with the fixture app foreground).
- A display that never gets there fails the command with the observed rotation and a hint, instead of reporting success. This is a behavior change worth a look: on the launcher, which pins portrait (`mCurrentAppOrientation=SCREEN_ORIENTATION_NOSENSOR`), `orientation landscape-left` previously "succeeded" without rotating and now fails after 15s with `observedRotation: 0`.
- A display that reports no `mCurrentOrientation` at all is left to the setting, as before.

## Validation

- Unit tests (fake adb, sleep mocked, fake clock for the budget): returns once the display reports the rotation after two stale reads; fails with `requestedRotation: 1, observedRotation: 0` when it never does; leaves a display that reports nothing to the setting. Package suite 641 tests green; typecheck, oxlint, oxfmt, fallow clean.
- Live on a local Pixel 9 Pro XL emulator with the fixture app foreground: `orientation landscape-left` 1s, `mCurrentOrientation=1`; `orientation portrait` 1s, `mCurrentOrientation=0`.

## What this does not explain

The other Android Smoke failure family (the post-alert canary missing while the device is in landscape although `orientation portrait` had been confirmed) is still open; #2350 adds the rotation evidence for it. This change removes the transition cost from the command that starts a rotation, which is also where a later flip would be visible if it is a delayed re-application.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,13 @@
   opening the request log. Long waits keep the first five and last twenty-five polls. The replay
   landmark-mismatch refusal carries the same poll evidence next to its mismatch details; `wait
   --stable` timeouts and a never-readable strict absence keep their existing diagnostics.
+- Fixed: Android `orientation` now returns once the display reports the requested rotation
+  (polling `dumpsys display`, up to 15s) instead of right after writing the settings. On a loaded
+  emulator the rotation takes seconds, during which accessibility reads hang, so the next command
+  paid for the transition; a `wait` issued right after `orientation` could spend its whole budget
+  there. A display that never reaches the requested rotation now fails the command with the
+  observed rotation instead of reporting success; a display that reports no rotation is left to
+  the setting as before.
 - Fixed: the iOS Simulator AX snapshot route bounds how long a capture waits for app discovery
   and stops starting a discovery per capture. Discovery (`simctl launchctl list` through xcrun)
   takes seconds on a loaded host; a capture now waits at most 1.5s for the one in-flight
--- a/packages/platform-android/src/__tests__/input-actions.test.ts
+++ b/packages/platform-android/src/__tests__/input-actions.test.ts
@@ -1,4 +1,4 @@
-import { test } from 'vitest';
+import { test, vi } from 'vitest';
 import assert from 'node:assert/strict';
 import {
   backAndroid,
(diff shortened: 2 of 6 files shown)


thymikee:
The new `dumpsys display` probe has no timeout derived from the remaining settle budget, so a stuck probe can bypass the intended 15-second loop. Add a hanging-probe regression and script the new ADB call in the provider scenario. Coverage and Integration currently fail on the unscripted `shell dumpsys display` call.

thymikee:
Addressed, rebased onto main (one commit):

- Each `dumpsys display` probe is bounded by what is left of the 15s settle budget (`timeoutMs: remainingMs(deadline)`), and the sleep between probes is bounded the same way, so a stuck probe ends the settle instead of bypassing it. A probe that times out fails the command with `orientation <x> could not confirm the display rotation: <adb error>`.
- Regressions: the never-rotates case now also asserts the first probe gets the full 15s and every later probe strictly less; a new hanging-probe case scripts one probe that consumes its whole timeout and throws, and expects one probe, the failure, and the clock at 15s.
- The provider world scripts `shell dumpsys display` to report the last `settings put system user_rotation`, so the Android Settings flow rotates the way the settle expects.

Local: platform-android 645 tests, provider Android scenarios 13, root typecheck, oxlint, oxfmt, fallow (against main) clean.

thymikee:
The bounded probe and provider scenario fixes are in place at 3104aa0. One failure path still returns success: dumpsys display runs with allowFailure, but its exit code is ignored. A failed ADB call with no orientation in stdout therefore looks like a successful probe that lacks the field. Reject failed probes through the existing ADB error handling and add a non-zero-exit regression; preserve the no-field case only for a successful probe. Android smoke also remains red at the later semantic-canary step.

thymikee:
The failed-display-probe handling is fixed at eb426b1: nonzero exits now fail the operation, with a dedicated regression. The live Android run passed both rotation assertions before failing at automation-press, so that failure is likely unrelated to this change. Ready for human review.

thymikee:
Addressed at cfb529b (rebased onto main, which now includes #2369):

- The `dumpsys display` probe no longer runs with `allowFailure`: a non-zero exit or a timeout goes through the existing ADB error handling and fails the settle as `orientation <x> could not confirm the display rotation: <adb error>`. Only a successful probe whose output lacks `mCurrentOrientation` is left to the setting.
- Regression: `a display probe that exits non-zero fails the settle instead of passing as no field` (one dumpsys call, rejects with the adb message).
- The earlier Android smoke failure on this branch (`press id="automation-press"` after `scroll down 0.7`) was the fixture growing below the canaries in #2345: the 0.7 drag alone reaches the list's maximum offset on the Pixel 7, which #2369 fixes in the scenario. Reproduced locally on a private Pixel 9 Pro XL emulator with the #2345 fixture (`failed-step-30.png`: list at max offset, press canary just above the fold) before #2369.

thymikee:
Reviewed cfb529b. No remaining code findings. Failed orientation probes now preserve the ADB error, and the single-probe regression remains covered. Android emulator smoke and the remaining checks pass on this head. Good for human review.

PR #2349, opened 2026-09-06
refactor(commands): move commands-side rendering out of src/daemon and retire the doctor progress flag

thymikee:
## Summary

Closes #2337.

`src/commands` and `src/mcp` reached into `src/daemon` for rendering they own — 11 `import()` sites, 5 file edges. Each moves to its owner:

- `daemon/result-serialization.ts` → `commands/output/result-serialization.ts`; the two deployment target resolvers its daemon handlers needed → `core/deploy-result-target.ts`.
- `daemon/handlers/doctor-output.ts` and `handlers/status-markers.ts` → `src/core/`, below the daemon handlers, the replay-test reporter and the doctor formatter.
- `daemon/client/doctor-progress.ts` (a process-global flag) is deleted, not deprecated. The CLI installs `createStderrCommandProgressSink` as the request's `RequestProgressSink` and hands its `CommandProgressState` to the output formatters; the daemon client only forwards events to the caller's sink.

The MCP edge becomes static. The `src/commands` edges stay function-scoped: static would take `src/cli.ts` from 380 to 384 eager modules, which the no-growth budget rejects. No budget raised, no compatibility re-export, no gate or baseline edit needed.

## Validation

Tested at `e4bc1ea`: `pnpm install --frozen-lockfile && pnpm build`, `pnpm typecheck`, `pnpm lint`, `pnpm check:affected --run` — all runnable checks passed (270 files / 2025 tests, plus daemon-wire-compat).

Graph query over `pnpm depgraph` output (non-type edges, `src/commands|src/mcp → src/daemon`): 5 at `27a97ee`, **0** at head — query, both outputs and the rename-only stat in the first comment.

`pnpm check:layering`: OK, 0 R5 back-edges, R2 clean. Eager-closure budgets: 442 pass; all 225 entries unchanged, `src/cli.ts` 380 → 380.

Doctor CLI output is byte-identical in all three modes (progress streamed, none, custom sink); captures diffed empty. The new CLI-level test was verified failing without the fix.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01Eddu5VC45qAXUkXNZJyruZ

---
_Generated by [Claude Code](https://claude.ai/code/session_01Eddu5VC45qAXUkXNZJyruZ)_

PR #2357, opened 2026-09-06
refactor(contracts): move single-owner modules out of @agent-device/contracts

### LINKED ISSUE #2339: refactor(contracts): move single-owner modules out of @agent-device/contracts

thymikee:
## Summary

Closes #2339. Stacked on #2348 (child A) — base is `claude/2336-command-registry-package`, not `main`.

Nine modules leave `@agent-device/contracts` as git renames, no re-export shim at any old path. Contracts production source drops **18,218 → 16,993 lines (−1,225)** across 190 → 180 files (`find packages/contracts/src -name '*.ts' ! -name '*.test.ts' ! -path '*__tests__*' | xargs wc -l`, base vs. HEAD). Exports map 118 → 109 subpaths, snapshot regenerated. 76 files changed; every moved module is `R091`–`R100` under `git diff -M90% --name-status origin/claude/2336-command-registry-package...HEAD`.

Corrections to the issue's tables, verified against the tree:

- **`wait-runtime-plan.ts` kept.** A's `packages/command-registry/src/__tests__/wait-runtime-execution.test.ts` imports it, so it is not root-only. Moving it would make a package import root `src/`.
- **`snapshot-presentation.ts` has 2 platform-android consumers**, not 6 (`snapshot-presentation.ts`, `ui-hierarchy-builder.ts`). Still single-package; moved.
- Two consequences of the moves: `android-touch-plan` lands as `touch-plan-lowering.ts` and retires the package-local `touch-plan.ts` re-export barrel it fed; `APPLE_OS_DISPLAY_NAMES` folds into `gesture-admission.ts`, its one remaining contracts caller, so both gesture refusals share one copy of the wording without a new subpath. `fast-check` follows the property tests from contracts to platform-android.
- `replay-divergence.ts`: implementation + its test to `src/core`, types stay on the `/divergence` subpath. `ReplayVarScrubEntry` follows the implementation — it is the sanitizer's parameter shape, not wire vocabulary. A split is not expressible as a rename.

**`interaction-guarantees.ts`: kept.** AGENTS.md names it a declaration site and points readers at that path; its non-daemon consumers are the interaction-contract test project and `scripts/depgraph/affected-run.ts`, i.e. repo tooling rather than a root zone. The ≥1,000-line target is met without it.

## Validation

Tested SHA `c83c1bbf4` — the rebased head, replayed onto A at `f887766` (which now carries #2346/C's `daemon/types.ts` split). Three conflicts against that split were resolved as this PR's move plus C's split types, in `snapshot-session.ts`, `session-replay-coordinator.ts` and `client/daemon-client-lifecycle.ts`; the divergence *types* stay on `@agent-device/contracts/divergence` there, as intended. The measurement above is re-run on the rebased pair and is unchanged.

After `pnpm install --frozen-lockfile && pnpm build`, `pnpm check:affected --run`: format, lint, typecheck, **layering**, di-seams, **fallow**, mcp-metadata, build, **package** (R11 package-boundaries), integration-node, macos-coverage all pass. `vitest-related`: 9728 passed, 3 failed — `app-log-session-resource`, `durable-capture-resource`, `capture-kit/durable-capture/adoption`, all asserting a `chmod 0500` directory is unwritable, which this container ignores because it runs as uid 0. All three are untouched by this branch and fail identically on the base.

The gates after that point were run individually and pass: replay-compat, daemon-wire-compat, affected-selector, gate-manifest, depgraph, tmpdir-leaks, coverage-model, wire-compat-model, production-exports, bundle-owner-files, command-docs, agent-guidance, xctest-selection, provider-integration, integration-progress, maestro-conformance, mutation-model, fixture-cache.

Fallow reports one pre-existing 19-line clone (`src/commands/output/error.ts` ↔ `src/mcp/tool-error.ts`), byte-identical on the base and only in scope because both files changed one import line. No gate suppressed.

The `platform-apple/src/snapshot-source/native-runtime.test.ts` failure seen twice on the pre-rebase head `6bf0b3d` is a native `xcrun clang` compile exceeding its 45 s budget on a loaded macOS runner; root cause and a proposed patch are in the comment thread. It is not this PR's and is deliberately not fixed here.

https://claude.ai/code/session_01SdXHhx9NkfH1PT4XjYe7tE


thymikee (commented):
Adversarial review from the #2335 coordinator, at `6bf0b3d`. I re-derived the load-bearing claims from the tree. Everything checks out, and the `wait-runtime-plan` catch is the most valuable thing in this PR.

**Verified independently against base `fc40090`:**
- Contracts production source **18,217 → 16,992 lines (−1,225)**, recomputed with my own count, not yours — matches your figure exactly and clears the ≥1,000 requirement.
- All nine moved modules are **absent** from `packages/contracts/src/`: no re-export shim survives at any old path.
- **Zero suppressions** anywhere in the diff — no `fallow-ignore`, no `APPROVED_OVER_CEILING`, no `ts-expect-error`, no disable comments, no skipped tests.
- Commit structure is exactly right: three implementation commits, then one final `chore(gates)`.

**The `wait-runtime-plan.ts` correction is right, and it is subtler than the PR body says.**

I confirmed it: `packages/command-registry/src/__tests__/wait-runtime-execution.test.ts:3` imports `@agent-device/contracts/wait-runtime-plan`. Moving that module to `src/daemon` as #2339's table instructs would make a workspace package import root `src/` — an R11 violation, and a gate failure you would then have been tempted to explain away.

What makes it worth calling out: **#2339's table was correct when it was written and became wrong underneath you.** It was measured on `main` at `27a97ee`, where that test lived at `src/core/command-descriptor/__tests__/` — genuinely root-only. A's packaging moved the test into a package and changed the fact. You verified against your actual base instead of the issue, which is the only reason this PR does not ship an R11 break.

That generalises, and I am carrying it forward: **every table in this umbrella was measured at `27a97ee`**, and `main` has since absorbed #2349 and #2347, with #2348 and #2346 close behind. #2341 and #2342 inherit tables with the same staleness risk, and #2342's seam list is hand-written in precisely this style. Re-verification against the real base is now a standing requirement for the rest of the wave, not a nicety.

Your `snapshot-presentation.ts` correction (2 platform-android consumers, not 6) is the third issue-table defect this wave after #2340's two wrong filenames. Worth noting the pattern rather than fixing each silently — which you did, correctly, in the body.

**Judgement calls I agree with:** keeping `interaction-guarantees.ts` (AGENTS.md names it a declaration site; its non-daemon consumers are repo tooling, not a root zone, and the line target is met without it); folding `APPLE_OS_DISPLAY_NAMES` into `gesture-admission.ts` rather than minting a subpath for one caller; and saying plainly that the `replay-divergence.ts` implementation/type split "is not expressible as a rename" instead of dressing it up as one. `ReplayVarScrubEntry` following the implementation is right — it is the sanitizer's parameter shape, not wire vocabulary.

The pre-existing 19-line `error.ts` ↔ `tool-error.ts` clone that fallow surfaces only because both files changed an import line is correctly left alone; adopting it here would be scope creep into a move PR.

Nothing blocking. CI is still settling — the `chmod 0500` trio is the known uid-0 artifact of this container and fails identically on the base, and the alert-related device lanes have been intermittently red across this whole wave independently of any of these branches.

---

thymikee:
**Smoke Tests failure on `6bf0b3d` is not this PR's.** Re-running it once; no code change from me.

Failing check: [Smoke Tests](https://github.com/callstack/agent-device/actions/runs/34027275657/job/101470381222) — `packages/platform-apple/src/snapshot-source/native-runtime.test.ts` > `native snapshot foreground ownership`: **`Error: Hook timed out in 60000ms`** at the `beforeAll` on line 9, which compiles a native binary from `path.resolve(import.meta.dirname, '../../../../…')`. 1 failed suite, 5 tests skipped, 61.6 s.

Why it is not this PR's — I checked the mechanism rather than assuming, because this PR *does* move modules into `packages/platform-apple` and a broken path there would be a plausible culprit:

- **A timeout is not a path break.** An unresolvable `nativeRoot` fails immediately with `ENOENT`; this burned the full 60 s doing work, which is a native compile that did not finish in budget.
- **This PR's entire `platform-apple` footprint is 3 files / +52 lines**: it adds the moved `multitouch-support.ts` and retargets two import lines in `gesture-facts.ts` and `interactions.ts`. It does **not** touch `packages/platform-apple/package.json`, `vitest.config.ts`, any vitest project definition, or anything under `src/snapshot-source/` — so it cannot have moved that `../../../../…` path or changed which project the test runs in or with what parallelism.
- **The same suite passed on this PR's base.** #2348 at `fc40090` — the exact base commit here — went fully green across the same `unit-core` project minutes earlier.

That leaves a slow native compile on a loaded macOS runner. Re-running once; if it fails again I will say so rather than re-running a second time.

For the record, this is the third distinct device/native lane failure across this wave (#2346 Android emulator wait-timeout, #2346 iOS alert XCUITest, this one a native-build hook timeout), on three branches that between them are a type split, a contracts move and a registry move. None share a mechanism with their failures. The macOS/device lanes look load-sensitive right now, which is worth a look independently of #2335.

---

thymikee:
Independent review at `6bf0b3d37c5c8541f6faffbfdb8fb5e659cab58e`: clean. The owner moves preserve the remaining contracts vocabulary and remove the superseded contract paths; the replay implementation/type split and platform-local moves are sound. No findings. All completed checks are green; iOS Smoke is still in progress. Land after #2348 once that exact-head gate passes.

thymikee:
**Second failure, so I am treating it as real rather than re-running again.** My re-run budget on this PR is spent. Root cause below, with a proposed patch — I am deliberately **not** applying it here, because it is unrelated to this PR's change and would widen a move PR.

Same suite both times, different timeout each time:

| run | error | suite duration | preceding build step |
|---|---|---|---|
| [101470381222](https://github.com/callstack/agent-device/actions/runs/34027275657/job/101470381222) | `Hook timed out in 60000ms` (outer vitest budget) | 61.6 s | 13.0 s |
| [101471711264](https://github.com/callstack/agent-device/actions/runs/34027275657/job/101471711264) | `xcrun timed out after 45000ms` (inner `runCmd` budget) | 47.9 s | 28.5 s |

Both are the same thing: `packages/platform-apple/src/snapshot-source/native-runtime.test.ts`'s `beforeAll` shells out to `xcrun --sdk macosx clang` to compile `SnapshotBridgeRuntime.m` + `fixtures/foreground-owner.m`, under `timeoutMs: 45_000` with a `60_000` outer hook budget. On a loaded macOS runner that compile does not finish in budget. The build step immediately before it took **2.2× longer** in the second run than the first, which is direct evidence the runner itself was slower, not that anything changed in the code.

**Why it is not this PR's** — the mechanism, not just correlation:
- This PR's entire `packages/platform-apple` footprint is 3 files / +52 lines: it adds the moved `multitouch-support.ts` and retargets two import lines. It touches no `.m`/`.swift` source, nothing under `src/snapshot-source/`, no `package.json`, and no vitest project definition. TypeScript module moves cannot slow an ObjC compile.
- The same `unit-core` project passed on this PR's exact base (#2348 at `fc40090`) shortly before.

I should be straight about the limit of that second point: I have not run this lane on `main` myself, so I cannot claim it is independently red there. The mechanism argument is what carries the weight.

**Proposed patch (for a separate PR, not this one):** the inner 45 s budget is the binding constraint and is too tight for a cold `xcrun` SDK resolution plus a two-file clang compile on a contended runner. Either raise both budgets together so the inner one stays the informative failure (e.g. `timeoutMs: 120_000` with a `150_000` hook budget), or — better, since the fixture sources rarely change — cache the compiled binary keyed on a hash of `SnapshotBridgeRuntime.m` + `foreground-owner.m` so only the first run in a job pays the compile. The former is a one-line change and would have made both of these runs pass.

This is the third distinct device/native lane failure in the #2335 wave (Android emulator wait-timeout and an iOS alert XCUITest on #2346, this one twice), across branches that are respectively a type split, a registry move and a contracts move. Two of the three are timeouts. The lanes look load-sensitive independently of anything in this umbrella.

---

thymikee:
The iOS smoke check has now finished red. I removed ready-for-human while that failure is investigated. The code review remains clean; land after #2348 once the stack and required checks are ready.

PR #2359, opened 2026-09-06
test(android-e2e): read the logcat tail for rotation evidence

thymikee:
The first CI failure carrying the rotation evidence (PR #2356, https://github.com/callstack/agent-device/actions/runs/34029660070) lost its logcat section: `(failed: Command failed: adb -s emulator-5554 logcat -d -v time)`, the 5s per-probe bound killing a full 2MB buffer dump on the loaded host. The probe now reads `logcat -d -v time -t 4000`, which holds the rotation decisions of the last minutes and returns well inside the bound (a local 2MB dump takes under a second; the tail is 290KB).

The settings and display sections of that file were intact and show the device in portrait at that failure, which is a different failure from the alert-canary one and is being handled on #2356.

### DIFF
--- a/test/integration/android-emulator-e2e/live-harness.ts
+++ b/test/integration/android-emulator-e2e/live-harness.ts
@@ -42,6 +42,8 @@ export function createContext(): LiveContext {
 const execFileAsync = promisify(execFile);
 
 const ROTATION_PROBE_TIMEOUT_MS = 5_000;
+/** Lines read back from logcat; the rotation decisions of the last few minutes fit comfortably. */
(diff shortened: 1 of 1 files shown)


thymikee:
The logcat-tail change looks good at dcb60a5. The reported artifact contains the rotation decisions, and the existing timeout and output caps remain intact. The remaining merge blocker is the failing Android smoke semantic-canary step.

thymikee:
The logcat-tail change remains clean at dd2d4554. The rebase leaves the reviewed code unchanged, and all checks are green. Ready for human review.

PR #2368, opened 2026-09-06
test(e2e): capture scenario assertion failure evidence

billsbooth:
## Summary
Capture failure artifacts when a live scenario assertion fails outside a CLI step. The iOS long-press search exhausted allowed probes, then threw without preserving the failed screen or accessibility tree.

## Scope
Two files: shared live-scenario harness and its deterministic tests. Capture before caller teardown, preserve the original exception, avoid duplicate capture for already-reported command failures, and leave successful scenarios unchanged. No gesture, driver, timeout, recovery or product behavior change.

## Validation
At `6ef99b084`, `pnpm check:affected --run` passes all runnable gates: format, lint, types, Fallow and 101 Node integration cases (9 native skips). Six new cases cover allowed misses, pre-command assertions, duplicate suppression, unavailable capture, artifact I/O failure and success; three failed before implementation. CI pending. This does not fix the original iOS visibility failure.

## Verification
[iOS failure](https://github.com/callstack/agent-device/actions/runs/34057724859/job/101552567368) contains four unsuccessful visibility probes and cleanup, but its uploaded fixture artifacts lack failure screenshot/snapshot. Source confirms only unexpected command failures previously invoked collection. The new tests assert original-error identity, owned-session flags, artifact output and coverage bookkeeping.

## Automated Test Plan
Run `node scripts/node-test-tmpdir.ts --test test/integration/live-device-e2e-runtime.test.ts test/integration/live-device-e2e-failed-step-evidence.test.ts`. All ten tests must pass with no device or account. Automatic native fixture lanes remain selected: on any scenario assertion, inspect failure report and capture artifacts before session cleanup; successful runs must create no failure captures. No production data, funds or external delivery required.

### DIFF
--- a/test/integration/live-device-e2e-runtime.test.ts
+++ b/test/integration/live-device-e2e-runtime.test.ts
@@ -1,0 +1,155 @@
+import assert from 'node:assert/strict';
+import fs from 'node:fs';
+import path from 'node:path';
+import test from 'node:test';
+
+import { mkdtempForTestSync } from '../../src/__tests__/test-utils/tmp-dir.ts';
+import type { CliJsonResult } from './cli-json.ts';
+import { createLiveDeviceContext, createLiveDeviceHarness } from './live-device-e2e/runtime.ts';
+
+function fixture(options: { captureThrows?: boolean } = {}) {
+  const context = createLiveDeviceContext<string>({
+    artifactRoot: mkdtempForTestSync('scenario-failure-evidence-'),
+    session: 'owned-fixture',
+  });
+  const calls: string[][] = [];
+  let reports = 0;
+  const harness = createLiveDeviceHarness<typeof context, string>({
+    behaviorsForScenario: () => [],
+    commandsForScenario: () => [],
+    commonFlags: (current, args) => [...args, '--session', current.session, '--json'],
+    runCli: async (args): Promise<CliJsonResult> => {
+      calls.push(args);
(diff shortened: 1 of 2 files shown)


thymikee:
The change looks good at 6ef99b084: scenario assertions now reuse failure capture before cleanup, preserve the original error, and avoid capturing command failures twice. The deterministic regressions cover those boundaries. CI remains the blocker: iOS failed in the alert XCTest before fixture E2E, and Android failed at automation-press. Neither run verifies the new scenario-assertion capture path; confirm the uploaded screenshot and snapshot on a deliberate assertion failure when that lane can run.

thymikee:
I refreshed this change on current `main` in [#2921](https://github.com/callstack/agent-device/pull/2921). The gap remains: scenario assertions after allowed CLI misses still lost their screenshot and snapshot. The replacement keeps the original implementation commit authored by @billsbooth, shares the command/scenario failure report formatter, and covers the newer platform-owned device evidence hook.

The two-file replacement passed 11 focused tests and `pnpm check:affected --run`. A deliberate local iOS 26.2 simulator assertion after an allowed miss preserved the original error and captured a valid PNG and Settings snapshot before cleanup. New CI is pending on #2921.

I tried updating this fork branch with an exact-head lease, but GitHub rejected the rebased push because the maintainer's OAuth token lacks `workflow` scope for workflow commits now on `main`. #2921 carries the same work without changing the contributor's fork, so I am closing this PR as superseded.

PR #2369, opened 2026-09-06
test(android): reveal smoke canaries by visibility

thiagobrez:
## Summary

The Android automation smoke scrolls a fixed `0.7` after rotation and overshoots `automation-press`, then fails before dispatching input. The [repeated failure on #2362](https://github.com/callstack/agent-device/actions/runs/34058548247/attempts/2) retains `failed-step-34.png` showing the button above the viewport.

Reset the route to its top after rotation, then check semantic visibility between bounded small scrolls. Use the same positioning for input outcomes, the alert, and the diff baseline. Keep all existing outcome assertions. Two test files change; no runtime or retry-policy changes.

## Validation

At `6211bfc3f15789fbb8cf5fd5907d399a7dd58d54`, `pnpm check:affected --run` passes (format, lint, typecheck; no related Vitest tests selected). The [hosted Android smoke passes](https://github.com/callstack/agent-device/actions/runs/34061497846), including the automation-system scenario and Settings replay. CI/coverage, integration, Linux/macOS smoke and all other checks are green. Prepared separately to preserve the iOS cold-start comparison at #2362's current SHA.

### DIFF
--- a/test/integration/android-emulator-e2e/live-assertions.ts
+++ b/test/integration/android-emulator-e2e/live-assertions.ts
@@ -22,6 +22,37 @@ export const { assertElementText, assertWaitSelector, assertWaitText, capturePng
     PUBLIC_COMMANDS.wait,
   );
 
+export async function scrollToVisibleSelector(
+  context: LiveContext,
+  selector: string,
+): Promise<void> {
+  for (let attempt = 0; attempt < 6; attempt += 1) {
+    const result = await runStep(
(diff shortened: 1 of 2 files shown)


thymikee:
Looks good at 6211bfc3. The smoke test now confirms visibility before acting, keeps the existing outcome assertions, and stops on unexpected errors. Hosted Android smoke and the remaining checks pass.

Issue #2371, opened 2026-09-07
Android scroll ignores releaseBehavior: every controlled scroll flings ~60% past its drag


thymikee:
## Problem

`scroll <direction> [amount]` is a controlled scroll by contract (`ScrollReleaseBehavior` in `packages/contracts/src/scroll-command.ts`: `controlled` by default, `inertial` only for the `scroll top` / `scroll bottom` edge passes). iOS honours it (`scrollDragProfile` → `controlledScroll`, a cubic ease-out that reaches near-zero velocity before lift). Android ignores it: `scrollAndroid` (`packages/platform-android/src/input-actions.ts`) builds a two-sample pan, the lowering expands it to 16 ms MOVEs, and the helper injects UP at the last MOVE's timestamp. The list receives a fling and keeps moving after the finger lifts, so `scroll down 0.3` does not scroll by 0.3 of the viewport, and the reported `pixels` (planned finger travel) says nothing about where the content ended up.

## Measurements

Pixel 9 Pro XL emulator, Android 17 (API 37), 1344x2992 @ 480 dpi. Displacement = y of a fixed node before vs. after `scroll down 0.3` (898 px drag over 300 ms). Snapshots taken ≥1 s after the gesture (an immediate snapshot can still show the pre-gesture tree).

| gesture | Settings (RecyclerView) | RN 0.85 `ScrollView` (fixture app) |
|---|---|---|
| 898 px / 300 ms, UP at the last MOVE (today) | 1430 px | 1423–1446 px |
| + 150 ms stationary tail (MOVEs at the end point, then UP) | – | 1423 px |
| + 400 ms stationary tail | 822 px (no fling) | 1427 px |
| + 400 ms tail drifting 40 px back toward the origin | – | 786–807 px (no fling) |
| + 160 ms tail alternating x by ±1 px, y fixed | – | 843–847 px (no fling) |
| 898 px / 1000 ms | – | 944 px |
| 898 px / 2000 ms | – | 897 px (1:1) |

So today's release is a fling on every list, the overshoot scales with pan velocity, and a purely stationary tail is not enough for an Android `ScrollView`.

## Why a stationary tail does not work

- `InputConsumer::rewriteMessage` keeps identical incoming coordinates as the last *resampled* sample, and `VelocityTracker` skips resampled samples (both in AOSP `frameworks/native/libs/input`). A tail whose samples repeat the previous coordinates never reaches the tracker, so the pan's velocity stays in place.
[...]

PR #2372, opened 2026-09-07
fix(android): honor scroll releaseBehavior to stop the fling overshoot

thymikee:
## Summary
- `scrollAndroid` ignored `ScrollReleaseBehavior` and always released the pan like a fling, so a "controlled" `scroll` overshot its requested distance by ~60% on both a plain `RecyclerView` and an RN `ScrollView` — measured and root-caused in #2371.
- For the default `'controlled'` release, append a short (~160ms) tail after the pan's endpoint that holds the scroll axis exactly fixed at the endpoint (zero velocity there by construction) while nudging the orthogonal axis every frame, so no two consecutive samples repeat. That dodges an AOSP quirk where `InputConsumer::rewriteMessage` treats a truly stationary tail as a "resampled" duplicate that `VelocityTracker` ignores, so the list still flings even with a hold. `'inertial'` (the `scroll top`/`scroll bottom` edge passes) is unchanged.
- Direction-aware: for vertical scrolls the tail jitters x while holding y; for horizontal scrolls it jitters y while holding x, so the scroll axis itself always carries exactly zero release velocity rather than relying on the jitter axis happening to be orthogonal to it.

## Known trade-off (not addressed here)
`scrollAndroid`'s touch plan extends `durationMs` by the tail length before Android's shared densification (`lowerAndroidTouchPlan`) re-grids the whole gesture at ~16ms cadence. Because that re-grid is computed from the total (move + tail) duration rather than per-segment, it can interleave a few extra sub-pixel interpolated samples near the tail's own 16ms-spaced points (this already happens today for `buildDragGesturePlan`'s hold segments — not something this PR introduces). It doesn't affect correctness (verified live below), just adds a handful of extra touch events. Fixing it properly means making the shared lowering segment-aware, which is a larger, separate change.

## Test plan
- [x] `packages/platform-android/src/__tests__/input-actions.test.ts` — 16 tests, including new coverage for: the default controlled-release tail (zigzag, never-repeating, duration composition), the duration floor composing correctly with the tail, a horizontal scroll jittering the correct (orthogonal) axis, and the `'inertial'` edge-pass path being unchanged.
- [x] `pnpm tsc --noEmit` (repo-wide) and full affected-package `vitest` runs (680 tests) — all green.
- [x] Live-measured on a Pixel 9 Pro XL / API 37 emulator: `scroll down --pixels 900 --duration-ms 300` on the Settings app (matching the maintainer's original repro) now lands within touch-slop of the requested drag (828–837px measured, vs. the old fling's ~1350–1440px overshoot) across two independent runs; `scroll bottom` (`'inertial'`) still completes cleanly and lands on the list's actual last row.


thymikee:
At fce45a3, the release tail is added after gesture duration validation. A valid 10000ms scroll therefore produces a 10160ms plan, beyond the shared total-duration limit. Please account for the tail at the owning duration boundary and add a maximum-duration regression without silently shortening the requested move.

The reported Settings result is encouraging, but #2371 also requires RN ScrollView and CI Pixel 7 displacement evidence. Please add those measurements before marking this ready.

The provider integration failure is related: it still expects a 350ms plan instead of the new 510ms controlled gesture. Update it to assert controlled versus inertial release. Coverage also reports a stale eager-closure approval, which looks unrelated to this change.

thymikee:
At 4a67a37, the duration cap is respected, but controlled scrolls near the maximum lose the required braking tail; at 10000ms there is no tail at all. Please preserve controlled-release behavior across the accepted duration range, or make an explicit validated duration constraint rather than silently dropping the braking step. Add a boundary regression and the RN ScrollView/Pixel 7 displacement evidence requested earlier.

Coverage also fails because the new explanatory comment grows the over-budget android-lifecycle.test.ts. Keep the corrected expectation without growing that legacy test. The IME recovery persistence failure and stale closure approval look unrelated.

thymikee:
Addressed both rounds of feedback.

**Duration boundary (fixed properly this time):** the first fix I pushed shrank the tail near the ceiling, which — as you caught — silently dropped the braking step exactly where a long scroll needs it. Reworked at 9a3ea26: a controlled scroll's own accepted `durationMs` range now stops 160ms short of `GESTURE_DURATION_MAX_MS` (max 9840ms), and anything past that is rejected with a clear `INVALID_ARGS` error naming the constraint, rather than truncating the move or the tail. Every accepted controlled scroll now runs the full, unshortened tail. `'inertial'` keeps the full 10000ms range (no tail to make room for). Added regression coverage for: the largest duration that still gets the full tail (lands exactly at 10000ms), the rejection just past that boundary, and `'inertial'` still accepting the full range.

**Provider integration + file size:** fixed the stale `350`→`510` expectation, this time as a same-line comment instead of a multi-line block — `android-lifecycle.test.ts` is back to 1260 lines (same as the merge-base), so it no longer trips the size ratchet.

**RN ScrollView + Pixel 7 evidence**, gathered live on a fresh `pixel_7`/API 36/`google_apis_playstore` AVD (matching `android.yml`'s emulator-runner config exactly, not the reserved `Pixel_7_CI`):

| Surface | Drag | Displacement | 
|---|---|---|
| RN `ScrollView` (examples/test-app, gesture-lab screen) | 900px | 850px, then 828px (2 runs) |
| Settings `RecyclerView`, same Pixel 7/API 36 device | 900px | 828px |

All three land at 92–94% of the requested drag (touch-slop territory), consistent with `controlled` release — not the ~1.5–1.6x overshoot the fling bug produced. Also re-confirmed `scroll top`/`scroll bottom` (`'inertial'`) still completes cleanly on this same device/API level.

The IME recovery persistence failure and the `planned-operations.ts` stale-approval row are still there on the latest push and, as you noted, look unrelated to this change (no IME or command-registry code touched) — I re-ran the Integration Tests job once and it passed clean on retry, consistent with flakiness rather than a real regression.

thymikee:
The duration fix at 9a3ea26 is sound: every accepted controlled scroll retains the full braking tail, and inertial scrolling keeps its existing range. All checks pass, and the reported Pixel 7 measurements cover both Settings and RN ScrollView.

One small acceptance item remains from #2371: update the scroll documentation to describe reduced momentum, not exact displacement, and explain the new 9840ms Android controlled-scroll limit. The current docs only contain the generic app-physics caveat.

thymikee:
Updated the scroll docs at 0214cc4 — `website/docs/docs/commands.md` (scroll section) now states plainly that the gesture releases with reduced momentum rather than an exact stop, and explains Android's controlled-release `durationMs` ceiling (9840ms, headroom for the braking tail) with what happens past it (rejected, not silently truncated). Also tightened the `scroll` CLI `--help` summary in `src/commands/interaction/metadata.ts` with the same "reduced momentum" framing. All CI checks green on this commit.

thymikee:
The help and duration-limit docs at 0214cc4 resolve the remaining review finding; no further code findings. Android smoke failed because the Back diff did not add the expected Open automation lab line: the connection to this scroll change is unclear and needs a reproduction or a justified rerun before merging. The iOS xcrun timeout appears unrelated. Ready for human review, not yet CI-green.

thymikee:
Rebased onto main (01bb115, 12 commits — includes #2373/#2375, which fixed the stale `planned-operations.ts` eager-closure approval, so that's now clean without any action from me).

**Android Smoke Tests**: reran and it passed clean — the "Back diff did not add the expected Open automation lab line" failure didn't reproduce. Consistent with flakiness rather than a regression; nothing in this diff touches back-navigation or the automation-lab route, and `scrollToVisibleSelector`'s convergence is accessibility-polling based (`is visible`), not sensitive to exact scroll distance.

**iOS Smoke Tests**: failed again on rerun, but I checked main's own tip commit (be62249, no relation to this branch) and it fails the identical job the same way — "Wait for ... to idle" stalling repeatedly during the XCTest run, unrelated to app navigation. This diff touches zero iOS files (`packages/platform-android/*`, one Android integration test, a shared CLI description string, and docs), so this is a pre-existing, repo-wide iOS CI infra issue, not something this PR introduced or can fix.

All checks now green except that pre-existing iOS flake.

thymikee:
No new code findings at 01bb115: the rebase preserves the Android implementation and tests, and Android smoke now passes. Ready for human review; the remaining iOS smoke failure occurs during initial fixture observation and appears unrelated to Android scroll. It is not the same failure as main’s text-input failure, so that comparison should not be treated as a confirmed matching flake; CI is still red.

PR #2373, opened 2026-09-07
chore: drop the stale planned-operations closure approval

thymikee:
## Summary

`scripts/__tests__/eager-closure-budgets.test.ts › no APPROVED_OVER_CEILING row is stale` fails on main at `e7a5b8b622` and on every PR based on it (for example #2325, #2372, #2363).

The `APPROVED_OVER_CEILING` row for `packages/command-registry/src/planned-operations.ts` was consumed by #2329. Now that main carries that entry, the no-growth rule governs it and nothing reads the row again, so the probe reports it stale and asks for its removal. This does exactly that: one row deleted, no budget or ceiling changed.

## Validation

`scripts/__tests__/eager-closure-budgets.test.ts`: 478/478 green locally against `origin/main`; `tsc -p tsconfig.json` clean.

### DIFF
--- a/scripts/__tests__/eager-closure-budgets.ts
+++ b/scripts/__tests__/eager-closure-budgets.ts
@@ -126,16 +126,7 @@ export const NEW_ENTRY_CEILINGS: Readonly<Record<EntryCategory, number>> = Objec
  */
 export const APPROVED_OVER_CEILING: Readonly<
   Record<string, { issue: string; reason: string; owner: string }>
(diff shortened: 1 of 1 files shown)


thymikee:
Reviewed ae41671. Removing the stale approval preserves the existing no-growth enforcement. Ready for human review; the iOS smoke failure is an unrelated snapshot fallback. #2375 includes this removal and also fixes the main-branch failure, so I would land that instead.

PR #2375, opened 2026-09-07
fix(gates): stop an eager-closure approval from turning main red on merge

thymikee:
## The failure

Coverage is red on `main` ([run 34099687663](https://github.com/callstack/agent-device/actions/runs/34099687663)), one test:

```
FAIL  unit-core  scripts/__tests__/eager-closure-budgets.test.ts > no APPROVED_OVER_CEILING row is stale
AssertionError: These approvals name an entry that no longer exists, that the merge-base now
carries, or that now fits its ceiling: remove the rows.
+ [ "packages/command-registry/src/planned-operations.ts" ]
```

## Why it fires, and why it was guaranteed to

The staleness rule reads `introduced`, which is derived from `git merge-base origin/main HEAD`. **On a push to `main` the merge-base IS the head**, so no entry is first-introduced — and every `APPROVED_OVER_CEILING` row reads as stale whatever its real state.

That is exactly the shape of the approving PR's own merge commit. #2329 had to add the `planned-operations.ts` row to merge (its closure is 74, over the domain-facade ceiling of 20), and the merge that followed it was the run that called the row dead. `main` has been red since, and every branch cut from `main` afterwards inherits the same failure — so this is not one bad row, it is a rule that cannot be satisfied by any approval.

## The fix

- **`staleApprovalRows`** makes the verdict a named rule in `eager-closure-budgets.ts` and defers it when the merge-base is the head, where no row is readable at all. Enforcement is not lost: a row that outlives its PR is still reported on the first branch whose merge-base could have read it.
- **The `planned-operations.ts` row goes**, which is what the rule asks for now that `main` carries the entry — its closure is governed by the no-growth rule from here on, not by the ceiling.
- `headCommit` joins `mergeBaseWithMain` in `committed-source-tree.ts`, the one owner of that git question.

## Validation

Both directions proven against the real tree, not just the unit rule:

| | merge-base ≠ head (branch) | merge-base = head (`main`) |
|---|---|---|
| planted dead row `packages/does-not-exist/src/gone.ts` | **fails** — reported stale | **passes** — deferred |

- `scripts/__tests__/eager-closure-budgets.test.ts`: 479 passed (was 1 failed / 478).
- New failing-direction test pins all four cases (live / now-fits / entry-gone / deferred) against the real `NEW_ENTRY_CEILINGS`, so the deferral cannot silently swallow a genuinely stale row.
- Whole coverage lane run locally: `test:coverage:ci`, `check:coverage-changed:test`, `check:tmpdir-leaks`, plus `format:check`, `lint`, `typecheck`, `check:fallow`.
- Three `durable-capture` adoption tests fail in my container both with and without this change, and passed on CI at `e7a5b8b` — they chmod a path unwritable and this sandbox runs as root, which ignores permission bits. Not touched by this PR.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>

https://claude.ai/code/session_01SfQqXj7JKQVgBA8eg9SMVB

---
_Generated by [Claude Code](https://claude.ai/code/session_01SfQqXj7JKQVgBA8eg9SMVB)_

### DIFF
--- a/scripts/__tests__/committed-source-tree.ts
+++ b/scripts/__tests__/committed-source-tree.ts
@@ -32,6 +32,16 @@ export function mergeBaseWithMain(repoRoot: string): string {
   }
 }
 
+/**
+ * The commit under measurement. Compared against `mergeBaseWithMain`, this answers whether the
+ * head carries work of its own: they are equal exactly when `main` already carries this commit --
+ * a push to `main`, or a branch that has not committed anything yet -- and then no entry is
+ * first-introduced, whatever the tree contains.
+ */
+export function headCommit(repoRoot: string): string {
+  return git(repoRoot, ['rev-parse', 'HEAD']).toString('utf8').trim();
+}
+
 /** Files renamed since `base`, current path -> path at `base`, so a rename is not a new entry. */
 export function renamedSince(repoRoot: string, base: string): ReadonlyMap<string, string> {
   const renamed = new Map<string, string>();
--- a/scripts/__tests__/eager-closure-budgets.test.ts
+++ b/scripts/__tests__/eager-closure-budgets.test.ts
@@ -6,6 +6,7 @@ import path from 'node:path';
 import { eagerClosureGraphOf } from '../../src/__tests__/eager-import-closure.fixtures.ts';
 import {
   createCommittedSourceTree,
+  headCommit,
   mergeBaseWithMain,
   renamedSince,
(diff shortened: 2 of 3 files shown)


thymikee:
Reviewed d920f2. The main-branch deferral fixes the false stale-approval failure while subsequent branches still reject obsolete approvals. The regression covers both directions, and all checks pass. Ready for human review; this also includes the removal in #2373.

Issue #2374, opened 2026-09-07
refactor(cli): cut src/cli.ts eager closure by taking vocabulary off runtime barrels


thymikee:
> **Status: closed. A and B shipped; C is withdrawn because this issue's premise was wrong.**
>
> - [x] **A** — landed as #2378 (`65ff270`), by @dvd233, byte-identical to the fix described below.
> - [x] **B** — landed as #2379 (`6371192`).
> - [ ] **C** — **withdrawn, do not start.** See "The premise was wrong" below.
>
> `src/cli.ts` eager closure on `527a56a`: **295 modules**, from 365. The prediction for A+B was 294; the extra module is `main` moving underneath, not a modelling error.
>
> **Measured startup effect: none.** Details below. A side-finding — a stale `APPROVED_OVER_CEILING` row for `command-registry/src/planned-operations.ts` — was real and landed separately as #2373.

## The premise was wrong

This issue opened on the sentence *"every `agent-device` invocation — `--version` included — eagerly evaluates 365 modules through `src/cli.ts`."* That is true of the **source** graph and false of the **shipped artifact**, and the difference is the whole point.

`tsdown` rolls the ~1,670 source modules into ~371 `dist/src/*.js` chunks. `dist/src/cli.js` statically imports **19** of them. Building the base and the branch and diffing those two chunk sets:

```
chunks removed:  (none)
chunks added:    (none)
```

The 70 source modules A and B took off the eager path were already merged into chunks the CLI imports regardless. Nothing stopped loading. The size report agreed across two independent runs:

| | run 1 | run 2 (rebased, smaller diff) |
|---|---|---|
| `--version` | 20.6 → 20.6 ms | 27.3 → 27.5 ms |
| `--help` | 55.9 → 58.3 ms (+2.4) | 76.8 → 75.4 ms (−1.4) |

The `--help` delta flipped sign on a smaller diff, and the **baseline** moved 55.9 → 76.8 ms on unchanged code — 37% runner-to-runner. Both readings are noise; there is no effect to measure in either direction.

**So the eager-closure gate measures the source graph, not what the shipped CLI evaluates.** It remains a real invariant — layering, test-suite import cost, its own no-growth rule — but it does not predict CLI startup, and this issue presented it as though it did.

## Why C is withdrawn rather than deferred

C proposed splitting `commands/family/registry.ts` (dominates 82 modules on the current head) so twelve command families stop loading their runtime for a metadata read. Two things kill it as written:

[...]

PR #2378, opened 2026-09-07
perf(cli): keep scroll vocabulary off gesture runtime

dvd233:
## Summary

- Import `SCROLL_INPUT_DIRECTIONS` from its owning contracts vocabulary module instead of the gesture runtime re-export.
- Keep command metadata behavior unchanged while removing the runtime edge from `src/cli.ts`.
- Touch one file and reduce the eager closure from 365 to 325 modules (-40), implementing Step A of #2374.

## Validation

Tested commit `ab05815b31c7aa5258a11c72f3ff429f52070c91`.

- focused `src/cli.ts` eager-closure test
- `pnpm exec vitest run src/commands/interaction/metadata.test.ts --maxWorkers=2`
- `pnpm run format:check`
- `pnpm run lint`
- `pnpm run typecheck`
- `pnpm run check:layering`
- `pnpm run check:fallow --base origin/main`
- `pnpm run build`

`pnpm check:affected --run` cannot complete on this Windows host: its related suites include existing POSIX/device assumptions. The target suite and structural gates above pass; GitHub CI remains authoritative for Linux, provider integration, and coverage.

Refs #2374

PR #2379, opened 2026-09-07
refactor(cli): take the CLI's vocabulary off runtime barrels

thymikee:
## Summary

Steps A and B of #2374. `src/cli.ts` eager closure **365 → 295 modules (−19%)**, measured with the eager-closure gate's own walker.

Both are the same defect: a module that needs *vocabulary* reached it through a *runtime* barrel and paid for the runtime.

**A — `interaction/metadata.ts` (−40, 365 → 325).** Every import in that metadata module is vocabulary except one, which took `SCROLL_INPUT_DIRECTIONS` from `./runtime/gestures.ts` and dragged 40 modules of gesture runtime into every CLI invocation to read a six-string constant. `gestures.ts` does not own it — it re-exports it from `@agent-device/contracts/scroll-gesture`, with a comment saying the vocabulary lives there precisely so the public API needn't depend on the command runtime. This file already imported from that module five lines earlier.

**B — `provider-policy.ts` (−30, 325 → 295).** It asks three questions of `@agent-device/provider-webdriver`: the known-provider map, the predicate over it, one type. All three live in `providers.ts`, a leaf with **zero imports**. The package predates the subpath-per-file rule and published only `"."`, so answering "is this string a known provider name" loaded 31 modules of WebDriver session handling, capture and XML. It now also publishes `./providers`. `provider-policy.ts` was the only in-closure consumer of the barrel, so the whole 30 comes back.

**C is not here, deliberately.** #2374 scoped it as "a design task, not a mechanical one" and said to split it out if it grew past one PR. It does. `commands/family/registry.ts` still dominates 82 modules and remains the largest single prize, but only 2 of the 12 families expose metadata separately; the other 10 declare `metadata` and `run` together per command (`family/types.ts:31`), so separating them changes how commands are authored across ten families. That belongs in its own PR against #2374 rather than bolted onto two one-line fixes.

## One change here is not mine

`APPROVED_OVER_CEILING` still named `packages/command-registry/src/planned-operations.ts` from #2198. The merge-base carries that entry now, so the row is stale and the gate's own staleness check fails on it. **Verified red on a clean `8299d5b` checkout with none of this branch applied.** Removing the row is exactly what the gate's message asks for, and it is one line in this PR's own subject area — but it is unrelated to the closure work and is trivially droppable if you would rather it went separately.

## Validation

At `775035d`: `pnpm check:tooling` exits 0 (format, lint, typecheck, layering, depgraph, gate-manifest, production-exports, tmpdir-leaks, xctest-selection, mcp-metadata, build, bundle-owner-files, package). `pnpm check:fallow` — no issues across the 5 changed files. `pnpm check:affected --run` passed. `eager-closure-budgets.test.ts` 480/480 (477/478 on main before the stale-row removal). Interaction and cli-connection unit tests: 29 files, 258 tests green.

Closure verified by the gate's own `eagerClosureGraphOf`: 365 at the merge-base, 295 here.

**Wall clock is not claimed.** Module count is a proxy for parse-and-evaluate cost, not a linear one; today's size report reads `--version` 27 ms and `--help` ~79 ms. Whatever the size report says on this PR is the real number.

Refs #2374.

https://claude.ai/code/session_01Jqfa11D8QsCMuL17SsLvDz

---
_Generated by [Claude Code](https://claude.ai/code/session_01Jqfa11D8QsCMuL17SsLvDz)_

### DIFF
--- a/packages/provider-webdriver/package.json
+++ b/packages/provider-webdriver/package.json
@@ -14,6 +14,10 @@
     ".": {
       "types": "./src/index.ts",
       "default": "./src/index.ts"
+    },
+    "./providers": {
+      "types": "./src/providers.ts",
+      "default": "./src/providers.ts"
     }
   }
 }
--- a/scripts/layering/package-boundaries.test.ts
+++ b/scripts/layering/package-boundaries.test.ts
@@ -656,10 +656,10 @@ test('the real tree parses, declares, and passes R11', () => {
     (pkg) => pkg.name === '@agent-device/provider-webdriver',
   );
   assert.ok(providerWebDriverPackage, 'provider-webdriver package must exist');
-  assert.deepEqual(
-    [...providerWebDriverPackage.exportTargets.keys()],
-    ['@agent-device/provider-webdriver'],
-  );
+  assert.deepEqual([...providerWebDriverPackage.exportTargets.keys()].sort(), [
+    '@agent-device/provider-webdriver',
+    '@agent-device/provider-webdriver/providers',
+  ]);
   assert.deepEqual([...providerWebDriverPackage.workspaceDependencies].sort(), [
     '@agent-device/capture-kit',
     '@agent-device/contracts',
--- a/src/cli/connection/provider-policy.ts
+++ b/src/cli/connection/provider-policy.ts
@@ -2,7 +2,7 @@ import {
   CLOUD_WEBDRIVER_PROVIDERS,
   isCloudWebDriverProviderName,
   type CloudWebDriverKnownProviderName,
-} from '@agent-device/provider-webdriver';
+} from '@agent-device/provider-webdriver/providers';
 
 export type DirectDeviceConnectProvider = CloudWebDriverKnownProviderName | 'limrun';
 export type ConnectProvider = 'cloud' | 'proxy' | DirectDeviceConnectProvider;


thymikee:
## The size report came back null, and I can explain why. It invalidates this PR's stated motivation.

| | base | current | diff |
|---|---:|---:|---:|
| CLI `--version` | 20.6 ms | 20.6 ms | +0.1 ms |
| CLI `--help` | 55.9 ms | 58.3 ms | **+2.4 ms** |
| Package (download) | 1.34 MB | 1.34 MB | +105 B |

Not noise, and not a measurement I can wave away. **The shipped CLI loads the same modules either way.**

`tsdown` rolls 1,667 source modules into 371 `dist/src/*.js` chunks. `dist/src/cli.js` statically imports **19** of them. I built `8299d5b` and this branch and diffed the two chunk sets:

```
chunks removed by this branch:  (none)
chunks added by this branch:    (none)
```

Identical. The 70 source modules this PR takes off the eager path were already merged by the bundler into chunks `cli.js` imports anyway, so nothing stops being loaded at runtime. The +105 B is `providers.ts` becoming its own small chunk now that it is a published subpath.

**So the eager-closure gate measures the source graph, not what the shipped CLI evaluates.** That is still a real invariant — it governs layering, the test suite's import cost, and the gate's own no-growth rule — but it does not predict CLI startup, and I framed #2374 as though it did. That was my error, and it is the interesting part of this result.

### What I think should happen

**These two changes still stand on their own merits, just not the ones in the title.** `interaction/metadata.ts` should not reach into `runtime/gestures.ts` for a constant that `contracts/scroll-gesture.ts` owns — `gestures.ts` only re-exports it, and its own comment says the vocabulary lives in contracts precisely to avoid this. `provider-policy.ts` should not import a runtime barrel to answer a string-membership question when the leaf has zero imports. Both are the repo's own rules; both are two lines. If you want them, they are ready. **If you would rather not carry a +105 B, zero-benefit diff, closing this is a completely reasonable call** and I will not argue it.

**Step C should not be started on the current premise.** Its 82 source modules would only matter if they map onto whole chunks that `cli.js` would stop importing, and nothing so far suggests they do. A ten-family authoring redesign for an unmeasurable win is a bad trade. I have recorded this on #2374; the right next step there is to check the chunk graph first and re-scope from what actually loads.

---

thymikee:
## Rebased — two of the three commits were landed independently, so only step B is left

`main` moved to `65ff270` while this was open. #2378 landed step A **byte-identical** to my commit (same blobs, `b4b7f2b..5c00400`), and the stale `APPROVED_OVER_CEILING` row is gone from `main` too. Rebasing dropped both automatically; git skipped `de41aa8` as already applied. No credit claimed for either — #2378 got there first.

What remains is one idea and three files:

```
packages/provider-webdriver/package.json      | 4 ++++
scripts/layering/package-boundaries.test.ts   | 8 ++++----
src/cli/connection/provider-policy.ts         | 2 +-
```

`provider-policy.ts` stops importing the WebDriver runtime barrel to answer whether a string names a known cloud provider, and reads the zero-import `providers.ts` leaf through a published `./providers` subpath instead. `src/cli.ts` eager closure is 295 on this base.

**My recommendation is unchanged from the comment above: this is layering, not performance.** The chunk-diff evidence says the shipped CLI loads the same 19 chunks either way, and the size report showed +105 B for it. It is correct by the repo's subpath rule and it is three lines, so it is ready if you want it — but **closing this is the call I would make** if you would rather not carry a small size regression for no runtime change. I have no attachment to it either way.

Re-validated at `79d8134` on the new base: typecheck 0, lint 0, format clean, layering OK, eager-closure + provider-policy tests 482/482.

---

thymikee:
Size report re-ran on the rebased head. Worth recording, because it retracts the one thing that looked like a regression:

| | first run (775035d) | rebased (79d8134) |
|---|---|---|
| `--version` base → current | 20.6 → 20.6 ms | 27.3 → 27.5 ms |
| `--help` base → current | 55.9 → **58.3** ms (+2.4) | 76.8 → **75.4** ms (−1.4) |

The `--help` delta flipped sign, on a diff that is now *smaller* than the one that produced +2.4 ms. And the **baselines** moved 55.9 → 76.8 ms — 21 ms, 37%, on unchanged code, purely between runners.

So the +2.4 ms in my earlier comment was noise, and I should not have presented it in the same table as a finding without saying so. Both readings are inside the runner spread; neither is evidence of anything.

That leaves the conclusion cleaner than before: this PR is **not** a startup regression, it is a no-op, exactly as the chunk diff predicted (19 chunks imported by `dist/src/cli.js`, unchanged). The only real cost is +111 B of download for `providers.ts` becoming its own chunk.

Merge-or-close still reads the same to me, minus the regression worry: correct layering, measurably zero runtime effect, small size cost.

---

thymikee:
Reviewed 79d8134. The CLI now reads the same provider vocabulary through its existing leaf module, without loading the provider runtime. No findings; all checks pass. Ready for human review.

Issue #2199, opened 2026-08-31
refactor(ios): prune converged snapshot paths and close release gates


thymikee:
Parent: #2188

Blocked by: #2195, #2198

## Outcome

Delete the compatibility and backend-name policy that remains solely for the migrated iOS snapshot
implementation, enforce no-regrowth at the owning seams, and run the final package-size,
conformance, provider, Simulator, proxy, and affected release evidence.

This is a deletion and enforcement issue, not a place to finish incomplete behavior from blockers.
Each owning migration must already have removed its obvious local duplication; this issue audits and
closes the cross-cutting remainder.

## Prune audit

Delete when equivalent engine-interface coverage is already proven:

- `shouldPresentIosInteractiveSnapshot` from `src/core/snapshot-state.ts`;
- `backendScopesAfterWire` and remaining backend-name scope planning;
- direct `presentIosInteractiveSnapshot` orchestration from `buildSnapshotState`;
- remaining provider-specific post-wire scope/option branches;
- remaining synthesized WebDriver hittability;
- old iOS semantic-compaction locations after `capture-kit` owns the implementation;
- runner-local Swift geometric-presentation copies after the runner and harness use the shared source;
- simulator observation dependencies on runner readiness and repeated bridge retries after the
  generation circuit opens;
- tests whose only purpose was to assert removed producer-name branches, after equivalent behavior is
  covered through the engine interface.

Move matching path-keyed fallow baselines rather than regenerating unrelated baselines. Stage new
structural modules before trusting the layering scan.

## Do not delete

- XCTest tree, query-sweep, or private-AX capture backends.
- Apple runner manager/cache, `.xctestrun` preparation, or interaction protocol.
- Swift geometric presentation or TypeScript iOS semantic compaction.
- Generic scope/normalization, refs, quality, occlusion, Android policy, provider acquisition
  adapters, golden fixtures, fuzz harnesses, live smokes, or physical-device code.
- Proxy routes, device leases, authentication, artifact transfer, or RPC compatibility machinery.

If deletion appears possible only by weakening another consumer, stop and report the remaining owner
rather than adding a compatibility re-export or fallback.

## No-regrowth enforcement

- Add or update structural ownership tests so producer adapters cannot import presentation behavior,
  contracts cannot own algorithms, and root orchestration cannot reconstruct engine policy.
- Reject old path recreation, backend-name scope/presentation branching, or direct iOS presenter
  invocation outside the engine.
[...]

PR #2382, opened 2026-09-07
fix(remote): let a plain-session client read its own failure diagnostics record

thymikee:
**This is an additional improvement, not retroactive acceptance of #2198.** #2198 is referenced only as context for where this route's authorization landed; nothing here re-opens or endorses that review.

## The defect

`GET /sessions/<session>/requests/<requestId>/diagnostics` (#1801) refused a caller its own failure record whenever the caller carried a tenant identity but its command had run in a plain, un-prefixed session.

The route applied the `<tenant>:` prefix rule to every caller with a tenant:

```ts
// src/daemon/request-diagnostics-http.ts (before)
if (auth.tenantId && !isTenantOwnedSessionName(auth.tenantId, ref.session)) -> UNAUTHORIZED
```

But the session-naming owner only writes that prefix under tenant isolation:

```ts
// src/daemon/request-admission.ts:35
if (isolation !== 'tenant') return req;   // session stays plain
```

and the daemon forces tenant isolation for exactly one class of caller — one whose tenant the auth hook **attested** (`http-server.ts`, the `/rpc` handler). A client that merely **declares** a tenant (`x-agent-device-tenant` on a daemon with no auth hook, which `resolveTrustedTenant` trusts at face value) runs in a plain session — `default`, or the `cwd:<hash>:default` that `resolveEffectiveSessionName` produces for any implicit session — gets a `diagnosticsRecord.session` naming that plain session, and is then answered 401 fetching it. Directly, and through `agent-device proxy`, which forwards the tenant header verbatim.

The read was stricter than the write. On that same daemon the same caller can already run **any** command in **any** session over `/rpc` — the two "no hook configured keeps a client-declared tenant unchanged" regressions in `http-server-tenant-trust.test.ts` prove it — so the prefix refusal denied an attacker nothing and denied honest callers their own records.

## The seam, and why it belongs to the authorization owner

`src/daemon/session-tenant-scope.ts` exists so the naming side and the record-reading side cannot disagree (its header comment says so). The disagreement was that the naming rule has a **precondition** the reading side did not carry: the prefix applies only where the namespace is actually partitioned. So the precondition now lives in that module, next to the rule:

```ts
export type TenantSessionNamespace = { tenant: string; partitioned: boolean };

export function isTenantAddressableSessionName(
  namespace: TenantSessionNamespace,
  sessionName: string,
): boolean {
  if (!namespace.partitioned) return true;
  return isTenantOwnedSessionName(namespace.tenant, sessionName);
}
```

The signal is threaded from the place that already computes it. `resolveTrustedTenant` (`server/tenant-trust.ts`) is the one function that distinguishes an attested tenant from a declared one, so its decision now reports `attested`, and `authorizeAuxiliaryHttpRequest` — the shared token/auth-hook gate for every non-RPC route — hands the route a `sessionNamespace` built from it. The `/rpc` handler's own `authResult.tenantId ? … : …` isolation default now reads `tenantTrust.attested`: same value, named for what it means. No new wire field, header, route, or public flag.

**The attested case is untouched.** An attested tenant is still refused every session outside its `<tenant>:` prefix, with the same typed `UNAUTHORIZED` / `Session is outside this tenant` error.

### One thing that differs from the reported diagnosis

The sketch suggested keeping the prefix refusal for tenant-prefixed names in the unpartitioned mode and relaxing only "plain" names. That is not decidable, and trying it would have left the main real case broken. A plain implicit session is named `cwd:<hash>:default`, so "plain" cannot mean "has no `<segment>:` prefix" — a leading-segment heuristic refuses exactly the session a CLI client actually runs in, and `cwd` is itself a reserved prefix indistinguishable from a tenant id. The route has no tenant registry and records outlive their sessions, so there is no other oracle.

So the implemented rule is the fully general one: **the prefix rule is applied only where the daemon partitioned the namespace.** Consequence, stated plainly: on a daemon with no auth hook, a caller declaring tenant B can now read a record from `tenant-a:default`. That is not a weakening — in that mode the same caller can already `POST /rpc` a command into `tenant-a:default` — but it does flip two existing assertions, so both are updated with the reason, and the cross-tenant refusal is now asserted where it is actually enforceable (under an attesting hook), which it previously was not.

### Adjacent finding, deliberately not fixed here

Even with an attested tenant, a client that sends `sessionIsolation: 'none'` keeps isolation `none` (the `/rpc` handler defers to the client's value), so it writes to a plain session and is then refused reading that record. The right fix is on the *write* side — attestation should force partitioning rather than defer — and it is a cross-tenant write-visibility question, not a read one. Left alone rather than weakening the attested read rule.

## Tests

1. **Direct retrieval** — `request-diagnostics-http.test.ts`, "serves a plain session to the tenant that ran the command": a caller with a declared tenant reads the record for session `default`.
2. **Proxy retrieval** — new `test/integration/provider-scenarios/remote-proxy-request-diagnostics.test.ts`: real daemon + `createDaemonProxyServer` + the real client (`localizeRemoteDaemonError` with `endpoint.tenantId`), asserting the localized record through the proxy and byte-identical parity with the direct fetch.
3. **Owning-client access** — `request-diagnostics-http.test.ts`, "serves the cwd-scoped session its owner ran in": the `cwd:9f1:default` shape a real implicit session actually gets.
4. **Cross-tenant refusal intact** — `http-server-tenant-trust.test.ts`, "an attested tenant is still refused a record outside its partition": under an attesting hook, `tenant-real` is refused both `tenant-other:default` and the plain `cwd:9f1:default`, each with `code: UNAUTHORIZED` / `Session is outside this tenant`, while its own `tenant-real:default` still serves 200.

Two existing assertions changed, both the unattested cross-tenant half described above: `request-diagnostics-http.test.ts`'s tenant test (renamed to say what it now proves) and the `otherTenant` half of `http-server-tenant-trust.test.ts`'s "no hook configured keeps the header-declared tenant unchanged".

## Red-then-green evidence

Cases 1, 2 and 3 were written first and run against unmodified `main`.

Cases 1 and 3, `vitest --project unit-core src/daemon/__tests__/request-diagnostics-http.test.ts`:

```
FAIL  |unit-core| ... > request diagnostics route serves a plain session to the tenant that ran the command
401 !== 200
- 200
+ 401
 ❯ src/daemon/__tests__/request-diagnostics-http.test.ts:172:12

FAIL  |unit-core| ... > request diagnostics route serves the cwd-scoped session its owner ran in
401 !== 200
 ❯ src/daemon/__tests__/request-diagnostics-http.test.ts:188:12

 Test Files  1 failed (1)
      Tests  2 failed | 5 passed (7)
```

Case 2, `vitest --project provider-integration test/integration/provider-scenarios/remote-proxy-request-diagnostics.test.ts`:

```
FAIL  |provider-integration| ... > Provider-backed integration local proxy serves a plain-session diagnostics record
AssertionError: the proxy must serve the record
+ 'remote daemon http://127.0.0.1:57696/agent-device, request abc123: HTTP 401'
- undefined
 ❯ test/integration/provider-scenarios/remote-proxy-request-diagnostics.test.ts:84:12

 Test Files  1 failed (1)
      Tests  1 failed (1)
```

Case 4 was planted red twice against the finished fix — once per side of the seam — to prove it watches the rule and the threading, not just the route.

Plant A, the rule (`isTenantAddressableSessionName` short-circuited to `return true`):

```
FAIL  |unit-core| ... > aux route: an attested tenant is still refused a record outside its partition
AssertionError: expected 401 for session tenant-other:default
200 !== 401
 ❯ src/daemon/__tests__/http-server-tenant-trust.test.ts:432:16
      Tests  1 failed | 14 skipped (15)
```

Plant B, the threading (`http-server.ts` hard-coding `partitioned: false`):

```
AssertionError: expected 401 for session tenant-other:default
      Tests  1 failed | 14 skipped (15)
```

Both plants reverted; all four cases green.

## Wire ledger

`isTenantOwnedSessionName` was already listed in `test/wire-compat/surface.ts`; `TenantSessionNamespace` and `isTenantAddressableSessionName` join it (the closure walk demands the type, since the listed authorizer signatures reach it). Five digests pasted into `ledger.json`, and `compatibleChanges` acks added for the three declarations that moved — `authorizeAuxiliaryHttpRequest` (its stale #2095 ack dropped, per the "an ack expires with its digest" rule), `RequestDiagnosticsHttpAuthorizer`, `handleRequestDiagnostics`. All three are additive: the new field is an in-process value, never serialized, and the only observable difference is a 401 that becomes the 200 + ndjson the caller already parses. No `wire-mutations.test.ts` `from:` snippet touched; no new break class, so no new mutation case.

## Gates

Run from the worktree:

- `pnpm format` (oxfmt) — clean
- `pnpm lint` — exit 0
- `pnpm typecheck` — exit 0
- `pnpm check:layering` — OK
- `pnpm check:daemon-wire-compat` — `checked against v0.20.10 (protocol unchanged): 176 declarations, 18 changed, 0 removed, 14 added, 18 moved`
- `pnpm check:affected --run` — **all runnable checks passed**: format, lint, typecheck, layering, fallow (`✓ No issues in 9 changed files`), build, vitest-related (204 files / 1259 tests passed), daemon-wire-compat

`fallow` initially flagged one new finding: `authorizeAuxiliaryHttpRequest` crossed the 60-LOC "very high" threshold. Fixed by extracting `sendAuxiliaryAuthHookRejection` — the one place these routes render a hook rejection as a flat REST error rather than the JSON-RPC envelope the decision carries — rather than by suppressing it.

Two runs died to local memory pressure before the clean one (~13.4G of 14.3G swap in use): a vitest teardown race in the unrelated `session-devices-batch-runtime.test.ts` (passes in isolation; the apple/xml import-after-teardown signature), and one silent worker death. Both cleared on re-run; the recorded results above are from clean runs. No pre-existing host-kit code-signature failures were hit.

### DIFF
--- a/src/daemon/__tests__/http-server-tenant-trust.test.ts
+++ b/src/daemon/__tests__/http-server-tenant-trust.test.ts
@@ -401,14 +401,48 @@ test('aux route: no hook configured keeps the header-declared tenant unchanged (
       },
     });
     assert.equal(owner.status, 200);
+    // Nothing attested either header, so `scopeRequestSession` partitioned no
+    // session namespace and the `<tenant>:` prefix carries no ownership to check
+    // — the two RPC regressions above show the same caller may already run any
+    // command in `tenant-a:default`. The token is the gate in this mode.
     const otherTenant = await fetch(diagnosticsUrl(baseUrl, 'tenant-a:default', 'abc123'), {
       headers: {
         authorization: `Bearer ${DAEMON_TOKEN}`,
         [DAEMON_HTTP_TENANT_HEADER]: 'tenant-b',
       },
     });
-    assert.equal(otherTenant.status, 401);
+    assert.equal(otherTenant.status, 200);
   });
+});
+
+test('aux route: an attested tenant is still refused a record outside its partition', async (t) => {
+  if (await skipWhenLoopbackUnavailable(t)) return;
+  const root = mkdtempForTestSync('agent-device-tenant-trust-aux-cross-');
+  try {
+    const hookPath = writeAttestingAuthHook(root);
+    await withDiagnosticsHookServer(hookPath, async ({ baseUrl, sessionsDir }) => {
+      // The hook attests `tenant-real`, so every session that caller can reach is
+      // named `tenant-real:...`. Another tenant's record, and the plain sessions
+      // outside the partition, both stay outside what it may address.
+      writeDiagnosticsRecord(sessionsDir, 'tenant-other:default', 'abc123');
+      writeDiagnosticsRecord(sessionsDir, 'cwd:9f1:default', 'abc123');
+      const auth = { authorization: `Bearer ${DAEMON_TOKEN}` };
+      for (const session of ['tenant-other:default', 'cwd:9f1:default']) {
+        const refused = await fetch(diagnosticsUrl(baseUrl, session, 'abc123'), { headers: auth });
+        assert.equal(refused.status, 401, `expected 401 for session ${session}`);
+        const body = (await refused.json()) as { error?: string; code?: string };
+        assert.equal(body.code, 'UNAUTHORIZED');
+        assert.equal(body.error, 'Session is outside this tenant');
+      }
+      writeDiagnosticsRecord(sessionsDir, `${ATTESTED_TENANT_ID}:default`, 'abc123');
+      const own = await fetch(diagnosticsUrl(baseUrl, `${ATTESTED_TENANT_ID}:default`, 'abc123'), {
+        headers: auth,
+      });
+      assert.equal(own.status, 200);
+    });
+  } finally {
+    fs.rmSync(root, { recursive: true, force: true });
+  }
 });
 
 test('aux route (upload): a hook configured but silent on tenant refuses a client-declared header', async (t) => {
--- a/src/daemon/__tests__/request-diagnostics-http.test.ts
+++ b/src/daemon/__tests__/request-diagnostics-http.test.ts
@@ -140,7 +140,15 @@ test('request diagnostics route rejects ids that do not name one record', async 
   });
 });
 
-test('request diagnostics route keeps a tenant inside its own session namespace', async (t) => {
+/**
+ * No auth hook is configured on this server, so every tenant here is a
+ * client-DECLARED label. `scopeRequestSession` partitions nothing for such a
+ * caller, and the same caller can already run any command in any session over
+ * `/rpc` (`http-server-tenant-trust.test.ts`, "no hook configured keeps a
+ * client-declared flags.tenant unchanged"). The prefix rule that does bite is
+ * proven against an ATTESTING hook in that file.
+ */
+test('request diagnostics route reads an unattested tenant as no session partition', async (t) => {
   if (await skipWhenLoopbackUnavailable(t)) return;
 
   await withDiagnosticsServer(async ({ baseUrl, sessionsDir }) => {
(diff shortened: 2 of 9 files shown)


thymikee:
The adjacent finding in this PR's "deliberately not fixed here" section is now fixed in **#2386**, stacked directly on this branch — and the fix is the one this body predicted (attestation forces partitioning rather than defers to the client).

I verified it rather than taking it on report: an attested tenant sending `sessionIsolation: 'none'` kept it, so `scopeRequestSession` returned early, the session never got its `<tenant>:` prefix, and `resolveSessionScope` dropped the request into the client-named `cwd:<hash>:` namespace that another attested tenant can name identically. Red against this branch: `Expected: "tenant" / Received: "none"`.

I also checked the load-bearing claim here — that a hookless daemon has no cross-session boundary to weaken — against the code rather than the cited regressions alone. It holds: with no hook, `sessionIsolation` is entirely client-supplied, so `scopeRequestSession` uses the client's `session` string verbatim; the token is the only gate in that mode. Relaxing the read there removes an inconsistency rather than a boundary.

Suggested merge order: this PR → #2386 → #2383 (the #2199 prune). All three merge into `main` with no conflicts, and typecheck, lint and `check:layering` are green on the integrated head.

thymikee:
Reviewed 3b5e8e8. Plain-session diagnostics now follow the existing authorization boundary, while attested tenants remain restricted to their own prefix. The direct and proxy regressions cover both cases. Ready for human review; Coverage failed in the unchanged iOS prewarm test and looks unrelated. Land #2386 afterward to close the existing isolation-downgrade hole.

Issue #2381, opened 2026-09-07
perf(ios): first-interaction stabilization over-sleeps; the 790 ms daemon cold load is a source-mode artifact


thymikee:
Follow-up from #2198. Traced **open → first readable snapshot → first successful interaction** on a local Simulator to find where the time goes, with the explicit constraint that a "win" must reduce the total, not move work into the next command.

## The ~790 ms daemon cold module load is refuted as stated

It is not the shipped path. On the bundled `dist` the daemon's process-start → listening-socket is **182–366 ms** (median ~250), of which module read + compile + resolve is **~87 ms**; the whole eager closure is 134 chunks / 1.25 MB and evaluates in **51–70 ms**, and the lazy Apple tail (`request-providers` → … → `interactor` → `mechanics`) adds **36–42 ms**.

The ~790 ms matches a different arm: booting from TypeScript source under `--experimental-strip-types`, measured at **671–934 ms** (median ~740). CPU-profiled it is amaro/SWC WASM type-stripping (~324 ms self) plus module-resolution stat storms (~202 ms self) over ~1,472 source modules.

Which arm runs is decided at [`src/daemon-client/daemon-launch-spec.ts:44`](https://github.com/callstack/agent-device/blob/main/src/daemon-client/daemon-launch-spec.ts#L44): `useSrc` is true only when the client itself runs under `--experimental-strip-types`, or when no `dist` exists. **A published npm install never pays it.** The existing comment on `resolveLocalDaemonCodeSignature` records the same ratio independently ("a dist entry … walks in ~5ms"; "a source checkout's graph — ~1,500 modules — costs tens of milliseconds").

So this is a dev-tooling cost on `pnpm test:replay:ios` and friends, not architecture, and not something users see. ADR 0019 laziness is genuinely in force: `packages/platform-apple/src/index.ts` compiles to a 661-byte chunk that eagerly evaluates exactly one module.

Caveat: absolute milliseconds were measured while two full test suites shared the host, so they are biased high. The ratios and the A/B comparisons are the trustworthy part.

## The first-interaction cost is confirmed, and 250 ms of it is a literal sleep

`open` deliberately leaves the runner un-warmed: it proves liveness with `uptime`, which is `isLifecycle: true` and therefore skips the app-activation preflight entirely; a relaunch's `targetReset` actively de-warms it.

[...]

PR #2387, opened 2026-09-07
test(remote): assert the proxied snapshot's whole wire conversation

thymikee:
Closes the one #2198 acceptance line whose proof was inference rather than an artifact.

## Why

#2198 required:

> The regular-snapshot trace must prove the bridge adds no network round trip and transfers only the published response.

What actually existed was a derivation from the controlled-RTT benchmark — identical `responseBytes` base vs head, and wall-minus-daemon at 20 ms RTT working out to ≈2.85 RTT — plus the fact that `isSupportedDaemonRoute` is a closed allowlist. That is good circumstantial evidence, but no test asserted it and the `ios-snapshot-convergence.v1` raw-result schema carries no round-trip count. I flagged it as derived-not-asserted when closing #2198; this is the promised follow-up.

## What it asserts

In the existing parity suite, a second `request` listener on the upstream daemon records every request the proxy forwards while one `snapshot -i` runs (registering another listener records without displacing the server's own handler, so the daemon behaves exactly as it does untraced).

It then pins the **whole wire conversation**, not just the RPC:

```
['GET /health', 'POST /rpc']
```

Pinning the whole list is the point — any new call of any kind breaks it, which is the regression #2198 forbids. Plus a byte check that what crossed is the published response and not a second payload (an acquired tree, lineage, a runner quality blob would not fit).

## A finding worth your attention

I expected `['POST /rpc']` and got `['GET /health', 'POST /rpc']`. The extra call is the client's own ADR 0006 protocol-compatibility probe in `ensureRemoteDaemon` (`daemon-client-lifecycle.ts:175`), not anything the bridge does — so the #2198 claim holds, and the test documents that distinction rather than hiding it.

But it fires on **every** `sendToDaemon` in remote mode, not once per client. For a fresh-process CLI that is one probe per process and unavoidable; for a **persistent** client it is a full extra RTT on every command. That is consistent with the ~2.5 round trips per snapshot the RTT benchmark measured, and at 80 ms RTT it is roughly 40 % of the proxied persistent-client latency.

I have not changed it here — this PR is test-only, and a fix needs to keep the compat guarantee (a short-TTL memo keyed on base URL and token is the obvious shape, but it needs its own measurement and its own review). Recorded on #2381.

## Red-then-green

Planted red by making the proxy issue one extra upstream call per forwarded request:

```
  [
    'GET /health',
+   'GET /artifacts',
    'POST /rpc'
  ]
```

Plant reverted; suite green.

## Validation

- `test/integration/provider-scenarios/remote-proxy-parity.test.ts` — 5 passed
- `src/__tests__/daemon-proxy.test.ts` — 13 passed
- `pnpm typecheck`, `pnpm format` — clean

Test-only; no production code changed.

### DIFF
--- a/test/integration/provider-scenarios/remote-proxy-parity.test.ts
+++ b/test/integration/provider-scenarios/remote-proxy-parity.test.ts
@@ -1,5 +1,6 @@
 import assert from 'node:assert/strict';
 import fs from 'node:fs';
+import http from 'node:http';
 import os from 'node:os';
 import path from 'node:path';
 import { test } from 'vitest';
@@ -228,7 +229,7 @@ async function runLeg(leg: ScenarioLeg, legName: string): Promise<DaemonResponse
 }
 
 async function withProxiedWorld<T>(
-  run: (context: { world: ParityWorld; proxied: ScenarioLeg }) => Promise<T>,
+  run: (context: { world: ParityWorld; proxied: ScenarioLeg; upstream: http.Server }) => Promise<T>,
   options: ParityWorldOptions = {},
 ): Promise<T> {
   const world = await createParityWorld(options);
@@ -259,7 +260,7 @@ async function withProxiedWorld<T>(
         return responseFromClientError(error);
       }
     };
-    return await run({ world, proxied });
+    return await run({ world, proxied, upstream });
   } finally {
     await closeLoopbackServer(proxy);
     await closeLoopbackServer(upstream);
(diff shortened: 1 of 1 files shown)


thymikee:
The request-list assertion at 6cd0f29 correctly catches extra round trips, but the byte check does not prove that only the published response crossed. A small extra envelope payload still fits below twice the response size, and extra fields inside result can grow both sides together. Capture the upstream RPC body and assert the expected envelope and published result directly; plant a small extra payload to prove rejection. CI is green.

thymikee:
You were right, and the byte check was worse than you said — I proved it before replacing it.

Planting a small extra field **inside `result`** passed the old assertion outright. Not because of the ratio, but because the check was circular: the client publishes whatever `result` holds, so `deepEqual(result, response)` moves both sides together and can never fail. The size bound was loose; the comparison next to it was vacuous.

Replaced at f89b906 with a read of what actually crossed, pinned at every level of the payload against its declared key set — envelope, `result`, and the published `data` — since an extra field is small at any level and a size bound cannot see it.

Three plants, each caught by the level that owns it:

| Plant | Caught by |
|---|---|
| extra envelope field `lineage` | `the upstream RPC body carried something beyond the JSON-RPC envelope` |
| extra `result` field `acquiredTreeRef` | `the RPC result carried something beyond the published response` |
| extra payload field `runnerQuality` | `the published snapshot payload carried a field it does not publish` |

The last two are the cases #2198 actually names — "the internal acquired tree, lineage, target-generation facts, and runner quality payload are not separate/intermediate wire payloads" — so the guard now matches the requirement rather than approximating it. All plants reverted; suite green (5 tests), `daemon-proxy.test.ts` green (13), typecheck clean.

Also rebased onto main after #2382 merged.

thymikee:
The wire-payload regression is fixed in f89b906: the test now checks actual RPC envelope, result, and data keys, rather than a circular byte comparison. No remaining code findings; ready for human review. The clean-install devices timeout appears unrelated to this test-only change; CI is not fully green.

PR #2390, opened 2026-09-08
fix(conformance): isolate post-tap settling from app launch

thymikee:
## Summary

Fix the `settle-after-tap` scenario from [failed run 34086351141](https://github.com/callstack/agent-device/actions/runs/34086351141). Both engines passed every flow, but the Settings tap recorded one settle latch and one timeout: per-command metrics also include the deferred launch boundary.

Add an explicit visual wait before the measured tap so launch stabilization cannot contaminate its inline-settle counters. Keep the zero-timeout and proof-of-life invariants unchanged. Three files changed: the flow, its semantic guard, and a runtime regression executing that checked-in flow with slow startup snapshots.

## Validation

Tested commit: `f80dc2c3e1`.

- Regression observed red before the fix with the CI signature (3 captures, 1 latch, 1 timeout), then green (2 captures, 1 latch, 0 timeouts).
- Focused runtime suite: 7 passed; differential invariant suite: 26 passed.
- `pnpm check:affected --run`: all runnable checks passed, including typecheck, lint, layering, build, packaging, and Node integration.
- CI and the live iOS differential rerun are pending. No production runtime behavior changed.

### DIFF
--- a/packages/maestro/test/conformance/differential/flows/settle-after-tap.yaml
+++ b/packages/maestro/test/conformance/differential/flows/settle-after-tap.yaml
@@ -7,6 +7,8 @@ appId: com.callstack.agentdevicelab
 - launchApp:
     clearState: true
 - assertVisible: Agent Device Tester
+# Discharge launch stabilization before measuring the inline post-tap settle.
+- waitForAnimationToEnd
 - tapOn:
     text: Settings
     retryTapIfNoChange: true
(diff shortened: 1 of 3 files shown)


thymikee:
The visual wait at f80dc2c separates launch stabilization from the measured tap without weakening the settle invariants, and the regression exercises the checked-in flow. No code findings. Exact-head iOS differential and smoke runs are now green; ready for human review.

PR #2396, opened 2026-09-08
refactor(daemon): move the device selection cluster into @agent-device/device-selection

thymikee:
## Summary

Moves the device-selection cluster (`dispatch-resolve`, `device-selection-resolver`, `device-inventory-context` + their tests/fixtures) into a new private workspace package `@agent-device/device-selection` — one subpath per module, no index, no re-export shims at the old paths.

Device selection no longer imports platform-apple app-resolution directly: the composition root (`src/platform-runtime-device-inventory.ts`) attaches an optional `findInstalledApp` probe to the composed inventory gateways (deferred Apple import), and the package reads the probe from the request context — absent probes fall back to the ordinary inventory rules. All ~60 consumers switch to the package subpaths; the layering gates gain the new leaf zone (the `request` zone is retired).

Dependency deltas (value edges, depgraph recipe: edges from the daemon-server zone to root files in other zones): 936 → 919 (−17, the three moved files). Replay root value deps: 6 → 4 (−2).

## Validation

- `pnpm check:affected --run` on the rebased head: format/lint/typecheck/layering/di-seams/fallow/mcp-metadata/build/package/integration-node/macos-coverage all green; vitest-related 3856/3857 — one transient 5s hard-timeout in `trace-runtime.test.ts` (not in this diff) passes in ~1.3s in isolation.
- `pnpm test:unit`: remaining failures are wall-clock budget flakes reproducible on baseline under the same host load (load avg ~45: concurrent emulators + other worktree suites); `test:smoke` green.
- `check:fallow`: no new findings; test-file-size ratchet and eager-closure budgets green; `check:production-exports` unchanged (all 65 pre-existing, none in touched files).


thymikee:
The move looks sound at ee11cdb, but the new installed-app probe needs one production-wiring regression. The moved selection tests supply the probe themselves, so removing it from createComposedDeviceInventoryGateways would leave those tests green while app-based simulator selection silently falls back. Add a test through that factory and the request context with two booted simulators and one installed-app match; show it fails when the factory omits the probe. Current CI passes.

thymikee:
Added in b93bfe1e4a: 'narrows app-based simulator selection through the factory-installed probe' in src/platform-runtime-device-inventory.test.ts (the factory's co-located test).

It runs two booted simulators through createComposedDeviceInventoryGateways + withDeviceInventoryContext with com.example.demo installed on exactly one, and asserts device.id, reason: single-app-installed-local, and that the probe consulted both simulators. Only the leaf xcrun spawn is faked (runXcrun in platform-apple's core tool-provider), so the real findIosSimulatorInstalledApp → listapps chain executes behind the factory-installed probe.

Red proof: with findInstalledApp removed from the factory's return, the same test fails with AMBIGUOUS_MATCH ("2 devices match this request equally") — the silent fallback to generic local rules you described.

thymikee:
The new regression test closes the wiring gap: it exercises the real inventory factory and selection route, with only the tool invocation mocked. Code review is clean at b93bfe1. The iOS alert-replacement test is still failing, but looks unrelated to this device-selection move; the relevant package, coverage and integration checks pass.

PR #2397, opened 2026-09-08
refactor(move): move the selector pipeline and interaction targeting into @agent-device/selectors

thymikee:
## Summary

Moves the selector pipeline and its interaction-targeting chain (selector-pipeline, selector-pipeline-policy, interaction-targeting, touch-semantics, interaction-positionals, press-retarget, interaction-touch-point, absence-observation and its error/resolution companions, interaction-error) out of `src/core` into `@agent-device/selectors`, the package that owns the engine the pipeline drives. Twelve subpaths are added to the package manifest; the R19 ownership rule now pins `packages/selectors/src/selector-pipeline.ts` and catches in-package relative imports of the engine as well as the specifier route.

Destination was measured, not assumed: the eager static closure of every selectors-importing entry point (maestro 154, ad-replay 105, commands 255, interaction-find 717 modules) is byte-identical whether the chain lands in selectors or capture-kit, so ownership decides: the pipeline is the selector pipeline, and landing here adds one forward edge (selectors → capture-kit) instead of two. `press-retarget.test.ts` stays in core because its oracle imports a daemon module; its import now points at the new subpath. Two fallow-surfaced dead exports were dropped: the unused `buildNodes` in the package-local snapshot-builder copy and a stale `SCROLL_INPUT_DIRECTIONS` re-export.

Deferred follow-ups: hoisting the shared test fixtures (#2402); moving `capture-kit`'s snapshot-occlusion module into contracts.

## Validation

- Rebased onto latest main (`06773095c6`); conflicts in `snapshot-state.test.ts`, `selector-action-resolution.ts`, `screenshot-crop.ts` resolved by keeping main's capture-kit imports alongside the relocated selectors imports.
- `62caaebd6c`: `pnpm check:affected --run` passed — 544/544 related test files (4124/4124 tests), layering 187/187, fallow clean.
- Rename-only: `git diff -M90% --stat origin/main...HEAD` shows 17 files as pure renames; the remainder is import re-pointing, the 12 subpath entries, the R19 pin, and one fallow baseline key move.
- One earlier gate run flaked on two untouched, timing-sensitive main tests (`daemon-entrypoint`, `interaction-verify`); both pass in isolation and on the full re-run.
- No CLI, wire, or device-visible behavior change.


thymikee:
No code findings at d8f0ff3. The selector move preserves behavior and the import gate covers relative-path bypasses. Resolve the snapshot-state test conflict by retaining the current test changes and the relocated import, then revalidate. The iOS failure is an xcrun timeout in an unchanged snapshot test and looks unrelated; the conflict still blocks readiness.

thymikee:
Adversarial review (pre-merge): structure is sound — the R19 ownership rule is correctly hardened so an in-package relative import of `engine.ts` is refused the same way as the specifier route, and the exhaustive export pin keeps the 12-subpath surface honest.

One follow-up to track (not a blocker for this rename-only PR): the moved tests now carry package-local copies of the shared fixtures — `packages/selectors/src/internal/__tests__/snapshot-builders.ts` mirrors the root `src/__tests__/test-utils/snapshot-builders.ts` (the duplication is forced by the packages-can't-import-root-src layering rule, host-kit's fixture subpaths are the precedent), and `property-arbitraries.ts` grew ~118 lines of touch-point/viewport arbitraries beside the root's replay/gesture ones. Two copies of `makeSnapshotState` will drift; a post-merge follow-up should hoist the shared snapshot/selector test fixtures to a single canonical location (selectors package subpath with the root re-importing, or a shared test-utils package).

thymikee:
Addressed both review threads:

1. Conflict (thymikee): rebased onto latest main (06773095c6) and resolved the snapshot-state test conflict by retaining main's capture-kit import alongside the relocated selectors import — `buildSnapshotState` now comes from '@agent-device/capture-kit/snapshot-state' and `resolveActionableTouchResolution` from '@agent-device/selectors/interaction-targeting'. Revalidated with 'pnpm check:affected --run' at 2409f70162: all runnable checks passed (547/547 test files, 7342/7342 tests, layering 187/187, fallow clean).

thymikee:
Fixture-duplication follow-up is tracked in #2402 (hoist shared snapshot/selector test fixtures to a single canonical location post-merge).

thymikee:
Looks good at 2409f70. The conflict resolution preserves main’s snapshot-state tests and both owning-package imports; the moved implementation bodies are unchanged. Ready for human review, with fixture cleanup tracked in #2402 and CI still running.

thymikee:
The clean code verdict also covers 62caaeb: its only change is a comment path. iOS passed all 82 native tests, then the fixture flow timed out waiting for Agent Device Tester without a completed readable snapshot. This looks like an unrelated capture/readiness failure rather than a selector assertion failure, though the root cause is unconfirmed; ready-for-human remains, but CI is not green.

PR #2398, opened 2026-09-08
fix(ios): avoid duplicate alert routing queries

thiagobrez:
## Summary

On our cold hosted iOS 26.5 simulator, `alert get` times out in a four-second routing probe and returns `RUNNER_BUSY` before its handler runs. The handler already resolves system, remote-hosted and app alerts itself.

Remove that duplicate probe for alert commands so resolution, activation and verification use the existing alert deadline. Coordinate taps retain their bounded routing probe and busy recovery. Move the deadline test beside a new actual-dispatch regression and run both in the iOS PR lane. Three files change.

This addresses a setup blocker while validating #2362; it does not fix delayed gesture delivery. [Original failure](https://github.com/thiagobrez/react-native-reorderable/actions/runs/34213575805).

## Validation

Head: `64de91773793d5a4cfdbd70c625e3fe4f3f46685`.

- `pnpm check:affected --run` and all active PR checks pass, including iOS smoke, integration and coverage.
- Native regression fails on the original code. With the fix, that regression, the requested-deadline test and coordinate-tap busy/drain regression pass (3/3).
- Fresh package/native builds pass. Actual Scenario Lab preflight passes locally on erased and previously confirmed iOS 26.5 setups.
- [Hosted diagnostic](https://github.com/thiagobrez/react-native-reorderable/actions/runs/34219362348): alerts are accepted in boots 2 and 3 without the duplicate probe. Boot 1 still times out inside alert handling; later viewport and lifecycle errors also keep the overall workflow red. This establishes the changed route, not general reliability under host load.

### DIFF
--- a/.github/workflows/ios.yml
+++ b/.github/workflows/ios.yml
@@ -153,6 +153,8 @@ jobs:
             -xctestrun "$XCTESTRUN_PATH" \
             -destination "platform=iOS Simulator,id=${{ steps.ios-simulator.outputs.simulator-udid }}" \
             -only-testing:AgentDeviceRunnerUITests/RunnerTests/testSinglePointerFlingFallsBackToXCTestCoordinateDragWhenPrivateSynthesisFails \
+            -only-testing:AgentDeviceRunnerUITests/RunnerTests/testAlertDispatchResolvesItsOwnModalWithoutCoordinateTapRoutingProbe \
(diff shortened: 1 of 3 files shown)


thymikee:
Ready for human review at 64de917. Alerts now use their existing resolver and deadline without the duplicate routing probe; coordinate-tap recovery is preserved, and the dispatch regression covers the removed probe. Current CI passes, including iOS smoke. The reported hosted successes support this narrow fix, not a claim that all cold-host alert or gesture failures are resolved.

PR #2366, opened 2026-09-06
feat: support standalone Maestro clearState command

Rohit3523:
## Summary

Maestro YAML flows using standalone `clearState` (`- clearState` / `- clearState: <appId>`) no longer fail with `Maestro command "clearState" is not supported`. Previously only `launchApp.clearState` (clear-then-open) existed.

Before: any flow with a bare `clearState` step (e.g. a login helper doing `stopApp` → conditional `clearState: ${APP_ID}` → `setPermissions` → open deeplink) was rejected at parse time.
After: the step parses, resolves `${VAR}`/config appId like `stopApp`, and clears app state without relaunching — the daemon projects it to `settings clear-app-state <app>` (session app when the id is omitted).

Examples a reviewer can run:

```yaml
appId: com.example.app
---
- clearState: com.example.app
- clearState
```

```sh
agent-device replay ./flow.yaml --maestro --platform android
```

Scope stayed inside the Maestro command family (`packages/maestro` + its daemon adapter + the generated-help/user-docs strings the sync test enforces): 16 files, +201/−20, no new packages or registries.

## Validation

Evidence gathered at `ea4b179a9`:

- New regression tests fail without the fix (implementation stashed: exactly the 2 new tests fail, 26 pass) and pass with it.
- `pnpm typecheck` clean; `pnpm test:maestro-compat` 331 pass; `pnpm maestro:conformance` 57 pass (coverage gate satisfied via `UNVERIFIED_COMMANDS`, unit-tested by other means); fuzz-arbitraries, help/docs-sync, export, and daemon-adapter suites green.
- `pnpm check:affected --run`: all runnable checks passed on a solo run. One full-gate run under concurrent load showed 2 subprocess-backed flakes (`session-open-runtime` metro-port) that pass in isolation; unrelated files, untouched by this diff.

Claim: published and reported — branch is pushed, CI on the head is still to come.

PR #2400, opened 2026-09-08
docs(eval): measure ripwire as a context tool for agents on this repo

thymikee:
## Summary

Measures whether [ripwire](https://github.com/redhat-et/ripwire) — an offline tool that hands an agent a ranked call graph instead of files to grep — makes agents better on this repo.

**Recommendation: do not adopt as a default; keep it a one-line opt-in. The token result is for default output and is not yet settled — see below.**

Method: six merged PRs (#2366, #2290, #2356, #2382, #2344, #2331) replayed as change-set localization — intent in prose, no file names, agent working in a clone pinned at the commit's parent so the fix is unreachable; ground truth is the commit's own file list. Two arms, identical briefs except one tooling paragraph, two replicates, 24 runs, plus two deterministic benches.

Result: **8 of 12 pairs returned identical file sets**; F1 0.807 → 0.830, within run-to-run noise. ripwire read **30% fewer source bytes** but spent **10% more tokens** on default-legend output. The documented `--legend=compact` costs no recall and saves 55% on `--callers`, 10% on `--affected`, 1% on `--for`; the arms' calls were mostly `--for`-shaped, so it should narrow that gap rather than close it. Re-running the A/B under compact is the open question this does not answer.

`--recall` finds all five CLI-flag declaration sites this repo documents by hand, where `--for`/`--pack-task` reach one to three — the document verbs are the strongest thing measured. For upstream: the preamble is 15% of a `--for` response and 62% of a `--callers` one, and compact clears it from the latter but not the former.

39 files; ~3.1k lines are the 24 raw runs, kept so every number is re-derivable. No runtime code, nothing ships.

## Validation

Tested at `f868f0c`. Docs and standalone tooling only — nothing under `src/` or `packages/`, no export, gate or command-surface change, so no device run applies.

`pnpm format:check`, `pnpm lint`, `pnpm typecheck` clean. `pnpm check:fallow --base origin/main`: no issues in 39 files — it caught three real harness defects at `614ec10` (duplicated flag/exec code, a 14-cyclomatic scorer arrow, undeclared entry points), fixed at source with gate entries in `chore(gates)` commits. Reproduced from a clean directory: `setup-clones.sh` → `make-briefs.mjs` → both benches → `score.mjs`, matching the checked-in results apart from measured `ms`.

Review follow-ups narrowed three claims to the invocations tested ([detail](https://github.com/callstack/agent-device/pull/2400#issuecomment-5597012075)): the terse-mode recommendation (compact already ships), the CLI-flag conclusion (withdrawn — wrong verb), and an `--affected` precision figure scored against a list that cannot score it (removed; it reports transitive reach). Recall and every A/B number are unchanged.

CI at `0e86a2d` was 16/18 green; both failures pre-existing, [evidence](https://github.com/callstack/agent-device/pull/2400#issuecomment-5588331973). iOS `testAlertCannotProveAnIdenticalReplacementAndDoesNotActivateIt()` — the iOS lane is red on `main` including at this PR's base `78cfc45` ([run](https://github.com/callstack/agent-device/actions/runs/34226333904)); no open PR fixes it. macOS — artifact `FinalizeArtifact 403` after the lane's only test passed. The one allowed re-run was spent at `6ed849f`.

`pnpm check:affected` fails open to all 58 checks for any `scripts/` change (`workflow-tooling`) — a classification, not reach. Risk: the harness has no test, matching `scripts/ios-snapshot-benchmark`.

https://claude.ai/code/session_01F7vMn8ehPq3NTPYMfxs7ro


thymikee:
The evaluation needs a few corrections at 614ec10 before its results are reusable.

The two benchmark commands do not rerun the agent A/B or its token measurement; they only rerun retrieval and test selection. Preserve the exact arm briefs and execution configuration, or describe the A/B as an archived observation rather than a reproducible experiment. Also fix the setup commands: after cloning and entering ripwire, the task clones use that directory as their origin instead of agent-device. The scoring example needs --worktrees to reproduce source-byte totals.

The test-selection score counts runtime-port-fixtures.ts and fake-adb.ts as expected tests because they sit under __tests__. Exclude helpers or score them separately; the test-only result is 9/12, not 9/14. Label retrieval recall as existing-files-only, since it excludes added_files while the overview includes them.

The failing quality check is related to the new harness: its standalone scripts are not declared as entry points, and it reports duplicated parsing/execution code and scorer complexity. Address those locally without weakening the repository-wide gate, then regenerate the results and check the documented commands from a clean directory.

thymikee:
All five addressed in 8498a76 (+ 56d10df for the gate entry). Results regenerated from freshly cut clones via the documented commands.

**Test-selection denominator.** You're right — `isTest()` matched anything under `__tests__/`, so `runtime-port-fixtures.ts` and `fake-adb.ts` counted as harnesses `--affected` should have named. Now only `*.test.*` files score; a file that merely lives in a test location (those two, plus the provider-scenario worlds and `test/wire-compat/surface.ts`) is a helper — neither a seed nor an expected hit — listed per task under `helpers_not_scored`. **9/12, not 9/14.** Per task that moves T3 to 1/1 and T4 to 2/2; the two remaining misses are real (two `packages/maestro` harnesses, and `snapshot-route.test.ts`, which covers the direct caller of the changed module).

**Retrieval recall basis.** Labelled existing-files-only in the doc, the bench header, and as `ground_truth_basis` on every row, with the reason stated: a retrieval verb ranks what the tree contains, so a file the commit created is not a hit it could have scored. The A/B scores the whole change set, and §1 now says why the denominators differ.

**Setup commands.** They could not have worked — `"$PWD"` was the ripwire checkout, and git refuses to fetch an abbreviated SHA, so every fetch would have failed. Replaced with `scripts/ripwire-eval/setup-clones.sh`, which resolves `origin` from the checkout it runs in, fetches the full parent SHA out of `tasks.json` (now storing full SHAs), and *fails* if the fix commit turns out reachable rather than silently handing an agent the answer. The scoring example carries `--worktrees`.

**A/B reproducibility.** Correct, and the fix is to stop overclaiming: it is now described as an archived observation in both the README and §3, and the caveats say the two deterministic benches reproduce while this one does not. What is preserved is what made it fair and auditable — `arms/baseline.md` and `arms/ripwire.md` (the only difference between the arms), `make-briefs.mjs`, which renders the 12 briefs and reproduces the ones the recorded runs were given byte-for-byte, the 24 answers in `runs/`, and the execution configuration written down. `score.mjs` re-derives every published number from those files.

**Quality check.** Fixed at the source, no gate weakened: extracted `bench-cli.mjs` for the duplicated flag reader and ripwire invocation, split the scorer into `accuracyOf`/`costOf` and trimmed the `??`-chains that pushed CRAP over threshold, and declared the four operator CLIs as `.fallowrc.json` entry points alongside `scripts/runner-request-count/run.ts` — in separate `chore(gates)` commits per `docs/agents/pull-requests.md`.

Verified from a clean directory: `setup-clones.sh` → `make-briefs.mjs` → both benches → `score.mjs`. Retrieval and affected outputs are identical to the checked-in ones except measured `ms`; the scorer's `by_arm` and `by_task` are identical. `fallow audit --base origin/main` → no issues in 39 changed files; `oxfmt --check` and `oxlint --deny-warnings` clean.

---

thymikee:
One measurement gap remains at 56d10df: affected-bench.mjs only accepts JS/TS source seeds, so T2 omits AccessibilityTreeXml.java and reports five seeds instead of six. That does not measure all non-test sources as the report claims. Include supported Java sources and regenerate the results, or explicitly limit and relabel this benchmark as JS/TS-only.

thymikee:
Confirmed and fixed in 6ed849f. The seed predicate required a `.ts`/`.mjs` extension, so T2 started its walk from five of the change's six sources while the prose claimed all non-test sources — the report was describing something the code did not do.

Fixed by including rather than relabelling: ripwire does build a call graph for Java here (`--for` on the helper's concepts returns `AccessibilityTreeXml.java`), so limiting the benchmark to JS/TS would have thrown away a measurement it can make. Seeds are now the code languages ripwire parses, named in one `SEED_EXTENSIONS` constant; data files (`.json`, `.yaml`) stay excluded because they carry no call edges. §2 says so explicitly now.

**T2 goes 5 → 6 seeds. Nothing else moves**: its selected set stays at 185 (19.7 KB, up 105 B), every task's hit count is unchanged, and the total is still 9/12. Worth stating plainly — the correction makes the method match its description without changing the result, so the 185-selected-files finding stands as reported.

Verified at 6ed849f: `affected-bench.mjs` re-run against freshly cut clones reproduces the checked-in results field-for-field except measured `ms`; `fallow audit --base origin/main` → no issues in 39 changed files; `oxfmt --check` and `oxlint --deny-warnings` clean.

---

thymikee:
**Standing down on the iOS Smoke Tests failure at 6ed849f — it is not this PR's.** Re-run triggered once ([run 34242118941](https://github.com/callstack/agent-device/actions/runs/34242118941)); flagging it here rather than silently, and I'll keep watching until the head is green.

**What failed:** `RunnerTests.testBareTypeUsesTappedInputWhenSoftwareKeyboardIsHidden()` — 4 failures of that one test, 0 unexpected, 82 executed, `TEST EXECUTE FAILED` (exit 65). It lives in `apple/runner/AgentDeviceRunner/AgentDeviceRunnerUITests/RunnerTests+CommandExecution.swift:363`, last touched on `main` by #2395.

**Why it is not this PR's:**

- The previous head **56d10df was green on all 18 checks, this same Smoke Tests job included**. The entire diff from there to 6ed849f is `git diff --stat 56d10df 6ed849f` → one Markdown file, `scripts/ripwire-eval/affected-bench.mjs`, and a JSON result file. No Swift, no `src/`, no `packages/`, nothing in the runner's build inputs. There is no path by which it reaches an XCUITest.
- The failure shape is the simulator's software keyboard, not the assertion: the log is `Checking existence of \`Keyboard (First Match)\`` retrying through a 12.3 s window, with `AGENT_DEVICE_RUNNER_TEXT_ENTRY_PHASE … phase=focus durationMs=8076.5 … mode=none`. That is host/simulator timing on a keyboard-hidden text-entry path.

**No fix to port:** there is no open PR or revert addressing this test that I could pull in — the failure is in a path this branch does not touch and does not modify. If the re-run comes back red on the same test, that is a real pre-existing failure in the iOS runner suite rather than something this PR introduced or can fix within its scope, and I'll say so rather than widen the PR.

---

thymikee:
The Java seed correction is sound at 6ed849f: T2 now includes all six source files, and the published recall is unchanged. No blocking findings remain, and CI is green. Small follow-up: remove Kotlin from the seed allowlist to match the documented parser gap; none of the current six tasks uses it, so this does not affect these results.

thymikee:
Done in 0e86a2d — and it was a real inconsistency, not just tidying: the allowlist claimed Kotlin as a language ripwire builds a call graph for while §"Fit with this repository" names it as one of the two parser gaps in this tree. Seeding from a file ripwire cannot parse would have measured an empty walk as a real one. The comment above `SEED_EXTENSIONS` now names the gap and points at the report, so the two cannot drift apart again silently.

As you said, no task in the set carries a Kotlin source, so every published number is unchanged — I re-ran `affected-bench.mjs` against the pinned clones and the output is identical to the checked-in results field-for-field apart from measured `ms`; the file is regenerated anyway so it matches the script that produced it. Still 9/12.

`fallow audit --base origin/main` → no issues in 39 changed files; `oxfmt --check` and `oxlint --deny-warnings` clean.

---

thymikee:
**Second CI failure at 0e86a2d — different job, different cause, and no test failed.** Flagging it separately since it is not the iOS one I commented on earlier.

**What failed:** the **macOS** Smoke Tests lane ([job 102137903363](https://github.com/callstack/agent-device/actions/runs/34248891505/job/102137903363)) — in the *artifact upload* step, not in a test:

```
✓ 01-system-settings.ad 32.3s
Test summary: 1 passed (1) in 32.3s
…
[error] Failed to FinalizeArtifact: Received non-retryable error:
        Failed request: (403) Forbidden: Error from intermediary with HTTP status code 403 "Forbidden"
```

The `replay-macos` gate step ended `outcome=success`; the job then went red finalizing `macos-artifacts` against GitHub's artifact storage. That is infrastructure returning 403 after the lane's only test had already passed — nothing in this repository, and nothing in this branch, which adds no files under `test/artifacts/**` or `test/screenshots/**`.

**No fix to port, and I am not re-running.** There is no change in this repo that stops GitHub's artifact intermediary from 403-ing, so there is nothing to port. My one permitted re-run was already spent on the earlier iOS failure, so I am leaving this one rather than quietly re-running past the budget: a maintainer re-run, or the next push to this branch, will clear it.

**If it recurs, one small workflow change would stop evidence collection from failing a green lane** — `.github/actions/upload-agent-device-artifacts` uploads diagnostics, not a gate, so `continue-on-error: true` on its upload step would keep a 403 from turning a passing lane red. That is a workflow decision outside this PR's scope, so I am proposing it rather than pushing it here.

Everything else on 0e86a2d is green (Lint & Format, Typecheck & Package, Compatibility & Provenance, Integration Tests, Bundle Size, CodeQL, the other Smoke lanes); Coverage and Repo Guards were still running at the time of writing. I'll keep watching until the head is green.

---

thymikee:
**The iOS lane is red on `main` itself, including at this PR's exact base.** That settles the earlier two failures better than my first comment could, so recording it here.

| iOS workflow run | `main` commit | Result |
| --- | --- | --- |
| [34226333904](https://github.com/callstack/agent-device/actions/runs/34226333904) | `78cfc45` `fix(ios): avoid duplicate alert routing queries (#2398)` — **this PR's base** | failure |
| [34220266299](https://github.com/callstack/agent-device/actions/runs/34220266299) | `31b01a2` | failure |
| [34217996652](https://github.com/callstack/agent-device/actions/runs/34217996652) | `07c4e40` | failure |

So this is the "red on the base branch too" case, established rather than assumed.

**The failure on `0e86a2d` is a *different* iOS test again:** `RunnerTests.testAlertCannotProveAnIdenticalReplacementAndDoesNotActivateIt()` (2 failures, 82 executed, exit 65) — not the keyboard test that failed on `6ed849f` and passed on re-run. The iOS suite is failing a different test on each run, and the one that failed here is about **alerts**, in the area the base commit #2398 changed. I'm naming that as a lead, not a conclusion — I have not bisected it.

**Nothing to port:** I searched the open PRs; #2362 is an unrelated draft on synthesized input (and reports its own trouble with this lane), and no open PR fixes the alert test. There is no change in this branch's scope that would fix a suite that is already red on its base.

**Not re-running:** my one re-run was spent on the `6ed849f` failure. Re-running until green would be papering over a real, pre-existing failure on `main`.

**Summary of the three CI failures on this PR, none of which are its own:**
1. `6ed849f` iOS — `testBareTypeUsesTappedInputWhenSoftwareKeyboardIsHidden` — passed on re-run.
2. `0e86a2d` macOS — artifact upload `403` *after* the lane's only test passed (infrastructure).
3. `0e86a2d` iOS — `testAlertCannotProveAnIdenticalReplacementAndDoesNotActivateIt` — same suite that is red on `main`.

The diff this branch adds is Markdown, a `.mjs` harness, a shell script and JSON, imported by nothing and shipping nothing; every non-device check is green. I'll keep watching, but the iOS lane going green here depends on `main`, not on this PR.

---

thymikee:
Code review is clean at 0e86a2d; the Kotlin allowlist nit is fixed and all non-timing benchmark results are unchanged. The two red checks look unrelated: macOS passed its test before artifact upload failed with HTTP 403, and iOS failed an alert-replacement assertion in unchanged Swift code. Ready for human review; CI is not green.

quaterniondrift:
noticed several issues w your measurements 
If you do not include the routing it is easy to get crazy results, it isn’t calling the thing designed for 

Several noted which change what the numbers mean.
Kotlin. The allowlist fix in 0e86a2d is right but the reason attached to it isn't. It calls Kotlin "one of the two documented parser gaps in this tree," which reads as though ripwire parses Kotlin generally and misses it here. It doesn't parse Kotlin anywhere — the README's Languages section enumerates all sixteen vendored grammars and Kotlin isn't among them. Worth rewording so the report doesn't leave readers with a wrong picture of coverage.
The CLI-flag task. The conclusion is that docs/agents/cli-flags.md beat the graph because those sites share a convention rather than call edges. Markdown is indexed as a document, and --recall / --mentions are the doc↔code verbs. The graph didn't lose to a doc file; the run never used the verb that reads doc files.
--affected naming 185 test files. --affected is blast radius, and a large reach set is the expected output. "Which tests must run" is --test-gate, which is scoped to that question and exits nonzero while obligations remain. Scoring --affected's raw set as test selection measures a different verb than the one the finding is about.
The preamble cost — this one stands, and it's mine to fix. 1.6–3.1 KB fixed preamble reaching 62% of a --callers response is a per-call invocation pattern. --pack-task returns ranking, bodies, callers and tests in one budgeted bundle for exactly that reason, and skills/ (seventeen task-shaped skills) is what tells an agent which verb the moment calls for — the arms here differed by one tooling paragraph, so the ripwire arm ran without it.

thymikee:
The benchmark conclusions need narrowing before this is ready. At the tested ripwire revision, docs/COMMANDS.md already documents --legend=compact, yet the report recommends waiting for a terse mode. Keep the measured default-output cost, but compare compact output before treating that cost as a reason to defer adoption.

The CLI-flag comparison also tests --for/--pack-task rather than the documented --recall/--mentions document paths, and the test-selection score counts tests untouched by the historical commit as excess even though --affected reports transitive reach. Describe those as limits of the tested invocation and ground truth, or add matched controls before claiming a tool limitation. The archived numbers can remain; the broader recommendation needs to reflect what was actually tested.

thymikee:
All three are right, and the second one reverses a conclusion rather than softening it. Fixed in f868f0c.

**`--legend=compact` exists — I recommended waiting for something already shipped.** Measured on this repo, it costs no recall and saves:

| verb | default | compact | |
| --- | --- | --- | --- |
| `--callers=attachRefs` | 5,092 B | 2,292 B | **−55%** |
| `--affected=…snapshot.ts` | 12,478 B | 11,186 B | −10% |
| `--for` (mean, 6 tasks) | 10,145 B | 10,043 B | −1% |

So it strips the whole preamble from the narrow verbs and only part of it from `--for`, whose bulk is ranked rows. The arms ran on default output and their calls were mostly `--for`-shaped, so compact should **narrow** the +10% token gap, not close it — but that's an inference from the deterministic benches, not a measurement of the agent arm. The report now says exactly that and names re-running the A/B under compact as the open question. `retrieval-bench.mjs` gained a `for-compact` variant so it's reproducible.

**The CLI-flag comparison tested the wrong verb, and the finding is withdrawn.** `--recall` finds **all five** documented declaration sites in 15.7 KB, where `--for`/`--pack-task` reach one to three:

| site | `--for` prose | `--for` idents | `--recall` |
| --- | --- | --- | --- |
| `packages/contracts/src/cli-flags.ts` | — | yes | **yes** |
| `src/commands/cli-grammar/*` | yes | yes | **yes** |
| `src/commands/command-projection.ts` | — | — | **yes** |
| `src/cli-schema/command-overrides.ts` | — | — | **yes** |
| `src/cli-schema/cli-config.ts` | — | yes | **yes** |

"A ranked call graph does not recover a convention" was an artifact of my verb choice. What's left is narrower and less interesting: graph verbs are the wrong tool for a documented convention, and ripwire's own routing sends that question to `--recall`. This also upgrades the document verbs in the verdict — they're now the strongest thing measured here.

**`--affected` precision was scored against a list that can't score it.** It answers transitive reach; a test it names that the commit left alone is not a false positive. The `precision` field is removed, recall stands at 9/12, and `selected` is now reported as read cost with the 185-file case described as a correct-but-expensive answer rather than a defect.

Recall and every A/B number are unchanged — the corrections are to framing, one withdrawn conclusion, and one removed metric. A caveat now states plainly that results are bounded by the invocations tested, citing §4 as the case where the verb changed the answer.

Gates at f868f0c: `fallow audit --base origin/main` no issues in 39 files; `oxfmt --check` and `oxlint --deny-warnings` clean.

---

quaterniondrift:
seems like should update X/twitter post?

PR #2414, opened 2026-09-09
perf(ios): speed up deep snapshots and keep first taps reliable

thymikee:
## Summary

Recover rejected deep iOS snapshots with bounded lower-depth requests and complete withheld branches. Preserve zero-based depth limits, reuse unchanged native trees, and fall back with non-launch failure codes when recovery cannot finish.

Contain optional coordinate-tap probe issues without penalizing later taps or granting an invalid typing witness. Required queries and mutation failures remain visible.

26 files across Apple capture, runner, tests and packaging/CI enforcement. No app-specific logic or public API change. Size comparison uses the measured revision's asset policy.

## Validation

Tested head: `7c84acfe684fbb45191460766ac8e9f6d739aa95`.

- Rebased onto #2391; Capture/launch/circuit tests passed, including 31 compiled native cases. Planted failures prove recovery and non-retry classification.
- Before the rebase, eight native iOS tests and iOS/macOS builds passed; native sources are unchanged. Immediate tap-then-type recovery is covered.
- Live Settings depths 0/1/4/64 and instrumented Bluesky frontiers verified. Missing/invalid boundary evidence fails closed. Independent review clear; follow-up #2424.
- Earlier benchmark at `8e402a0ac4`: warm startup to a verified Bluesky post averaged 11.14s vs 16.13s on 0.21, two warm samples each. This fixture-specific benchmark has not been repeated after review adjustments.
- Updated GitHub CI pending.


apple/snapshot-bridge/SnapshotBridgeCapture.m
```diff
@@ -0,0 +1,113 @@
...
+NSDictionary *captureSnapshotTree(id element, NSUInteger maxDepth, NSUInteger maxNodes,
+                                  SnapshotElementReader reader, BOOL *truncated, NSError **error)
+{
+  SnapshotTreeCapture *capture = [SnapshotTreeCapture new];
+  capture.reader = reader;
+  capture.acceptedDepth = maxDepth;
+  capture.remainingNodes = maxNodes;
+  capture.maximumNodes = maxNodes;
+  NSDictionary *tree = [capture read:element depth:maxDepth error:error];
+  NSDictionary *result = tree ? [capture materialize:tree depth:MAX(1, maxDepth) error:error] : nil;
```
thymikee: **Off-by-one: the materializer delivers one level less than the request's `maxDepth`.**

`materialize:depth:` treats `depth` as the total number of levels *including* the root (`depth <= 1` strips the root's children), but the rest of the pipeline uses "levels below the root": `nodeFromSnapshot` in `SnapshotBridgeRuntime.m` still accepts nodes down to `depth == maxDepth` with the root at 0, and the pre-PR TS residue threshold (`maxTraversalDepth >= maxDepth`) expected trees to reach traversal depth `maxDepth`. The two bounds in the same runtime now disagree by one, and the TS side had to be shifted to `maxDepth - 1` to match.

Concrete scenario: `snapshot --raw --depth 1` → `rawTraversalDepth: 1` → native `maxDepth: 1` → `materialize:… depth:MAX(1, 1)` → the Application root's direct children are dropped and `truncated: true` is reported, where the previous release returned root + children. Every depth-bounded raw capture loses its deepest permitted level.

Suggested fix: `[capture materialize:tree depth:maxDepth + 1 error:error]` (then the `adapter.ts` threshold can go back to `>= maxDepth`). The `depth-bound` fixture expectation (`count == 4` for `maxDepth:4`) encodes the same off-by-one and would need to become 5.

---

apple/snapshot-bridge/SnapshotBridgeCapture.m
```diff
@@ -0,0 +1,113 @@
...
+- (nullable NSDictionary *)read:(id)element depth:(NSUInteger)depth error:(NSError **)error;
+- (nullable NSDictionary *)materialize:(NSDictionary *)tree depth:(NSUInteger)depth error:(NSError **)error;
+@end
+
+@implementation SnapshotTreeCapture
+- (nullable NSDictionary *)read:(id)element depth:(NSUInteger)depth error:(NSError **)error
+{
+  NSUInteger attemptDepth = MIN(depth, self.acceptedDepth);
+  for (;;) {
+    if (self.requests >= maximumRequests) {
```
thymikee, line 25: **Budget exhaustion throws away an otherwise valid partial tree.**

Once 32 native requests are spent, `read:` returns nil with error code 1 and `materialize:` propagates the nil all the way up, so `captureSnapshotTree` returns nothing even though every node materialized so far is complete and the truncation accounting already exists.

Scenario: root rejected at depth 64 → accepted at 8 → the frontier at depth 8 is a list with 40 withheld cells. Continuations 2..32 succeed for 31 cells, the 32nd hits the budget, and the whole capture fails with `application-server-unavailable`. Previously this screen failed once, fast; now it costs 32 round trips and *then* fails, and the XCTest fallback starts later.

Suggested: when the budget is exhausted at a continuation, treat it like the node-budget case a few lines below (`self.truncated = YES; result[childrenKey] = @[]; return result;`) so the caller gets a truncated tree with the disclosure it already understands, and reserve the hard failure for the root request.

---

apple/runner/AgentDeviceRunner/AgentDeviceRunnerUITests/RunnerTests+TextInputProbe.swift
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+  func coordinateTapTextInputAt(app: XCUIApplication, x: Double, y: Double) -> XCUIElement? {
+    switch probeTextInputs(app: app, point: CGPoint(x: x, y: y)) {
+    case .matches(let elements):
+      return elements.first
+    case .absent:
+      return nil
+    case .unavailable(let failure):
+      penalizeSnapshotXCTestChannel(bundleId: currentBundleId, reason: failure.rawValue)
```
thymikee: **A single contained probe issue now disables typing-after-tap for the bundle for 120 s.**

`.unavailable` (both `.recordedIssue` and `.exception`) calls `penalizeSnapshotXCTestChannel`, whose duration is `snapshotXCTestChannelPenaltyDuration = 120`. While the penalty is active, `shouldProbeCoordinateTapTextInput` returns false, so every later coordinate tap on that bundle sets `textInput = nil` and no `textEntryTapWitness` is established.

Scenario: tap on field A → the optional probe records one issue (before this PR `safely` just swallowed exceptions and the AX-suppression path muted the fast `kAXErrorIllegalArgument` signature without any penalty) → tap on field B, then bare `type` → `TEXT_INPUT_NOT_FOCUSED`, for the next two minutes, and every `snapshot` on that bundle also loses the XCTest tier. The penalty's own comment describes it as memory for *slow, timed-out, or abandoned* XCTest captures; a fast probe hiccup does not meet that bar.

Suggested: keep the probe's outcome local (return nil without penalizing) unless the probe actually exceeded a time budget, or track a probe-specific, shorter penalty rather than reusing the snapshot-channel one.

---

apple/snapshot-bridge/SnapshotBridgeCapture.m
```diff
@@ -0,0 +1,113 @@
...
+      if (error) *error = [NSError errorWithDomain:@"agent-device.snapshot" code:1
+          userInfo:@{NSLocalizedDescriptionKey: @"snapshot continuation request budget exhausted"}];
+      return nil;
+    }
+    self.requests++;
+    NSError *failure = nil;
+    id tree = self.reader(element, attemptDepth, MIN(self.maximumNodes, self.remainingNodes + 1), &failure);
+    if (tree) return tree;
+    NSNumber *nativeCode = failure.userInfo[@"accessibility-error"];
+    BOOL rejected = ([nativeCode isKindOfClass:NSNumber.class] && nativeCode.integerValue == -25201) ||
```
thymikee, line 35: **The retry key (`-25201` = `kAXErrorIllegalArgument`, or any automation-support code 5) is the generic AX rejection signature, not a depth-specific one, so non-depth failures are retried up to 7 extra times.**

`kAXErrorIllegalArgument` is the same code the runner already recognises as the #758 "AX-broken screen" fast-fail, and it is also what the AX server returns for a stale/invalid element. Nothing here distinguishes "tree too deep" from those.

Scenario: an element the AX server rejects outright at any depth → attempts at 128, 64, 32, 16, 8, 4, 2, 1 before the original error surfaces (8 round trips instead of 1); if each rejection is a slow grind rather than a fast fail, the request approaches `maxDurationMs` and the watchdog SIGKILLs the bridge, which the host sees as a socket close rather than the typed failure it got before.

Suggested: bound the retry to, say, two halvings unless the halved request actually succeeds, or stop retrying once a halved attempt fails with the same code (a depth problem should clear after one or two halvings; a generic rejection never will).

---

packages/platform-apple/src/snapshot-source/adapter.ts
```diff
@@ -247,7 +247,7 @@ function truncationResidue(
   if (nodeCount >= limits.maxNodes) {
     return { kind: 'truncated', dimension: 'nodes', limit: limits.maxNodes };
   }
-  if (maxTraversalDepth >= maxDepth) {
+  if (maxTraversalDepth >= Math.max(0, maxDepth - 1)) {
```
thymikee: **This threshold shift patches the symptom of the native off-by-one and widens misclassification.**

The only reason `>= maxDepth` no longer matched is that `captureSnapshotTree` now materializes `maxDepth` levels *including* the root (see the comment on `SnapshotBridgeCapture.m:110`). Compensating here means the TS side now claims `dimension: 'depth'` for any truncated tree that reaches `maxDepth - 1`, including trees the bridge cut for payload size that happen to be that deep, and it bakes the reduced delivered depth into the contract.

Suggested: fix the native bound (`materialize … depth:maxDepth + 1`) and restore `maxTraversalDepth >= maxDepth`; the adapter fixture's `truncated: request.maxDepth === 1` then also needs a child node to stay meaningful.

---

apple/snapshot-bridge/SnapshotBridgeCapture.m
```diff
@@ -0,0 +1,113 @@
...
+  }
+  if (self.remainingNodes == 0) {
+    self.truncated = YES;
+    return nil;
+  }
+  self.remainingNodes--;
+  NSArray *children = tree[childrenKey];
+  NSNumber *childCount = tree[childCountKey];
+  BOOL withheld = [childCount isKindOfClass:NSNumber.class] && childCount.unsignedIntegerValue > children.count;
+  NSMutableDictionary *result = [tree mutableCopy];
```
thymikee: **Efficiency (healthy path): every node is `mutableCopy`'d even when nothing changes.**

The PR description says a healthy capture still uses one request, but it now also pays a full extra pass over the tree that copies every snapshot dictionary (and rebuilds every children array) before `nodeFromSnapshot` walks it again. At the 10 000-node limit that is 10 000 dictionary copies per snapshot on the hot path the PR is trying to speed up.

Suggested: only copy when a child was actually replaced or stripped (return `tree` unchanged when `!withheld`, `depth > 1`, and the node budget was not hit), or fold the withheld-continuation into `nodeFromSnapshot`'s existing walk so there is one pass.

---

apple/snapshot-bridge/SnapshotBridgeRuntime.m
```diff
@@ -311,13 +312,26 @@ - (nullable NSDictionary *)snapshotForProcess:(pid_t)pid
...
+            if (captureError) *captureError = [NSError errorWithDomain:@"agent-device.snapshot" code:5
+                userInfo:@{NSLocalizedDescriptionKey: @"foreground owner changed during continuation"}];
+            return nil;
+          }
+          NSMutableDictionary *bounded = [options mutableCopy];
+          bounded[@"maxDepth"] = @(depth);
+          bounded[@"maxChildren"] = @(nodes);
+          bounded[@"maxArrayCount"] = @(nodes);
+          return [_framework userTestingSnapshotForElement:element options:bounded error:captureError];
+        }, &acquisitionTruncated, &runtimeError);
```
thymikee, line 334: **Capture-internal failures are misreported as `application_unavailable` / `application-server-unavailable`, which the host treats as a transient launch state.** (Anchored here because the mapping at L348 is outside the diff.)

The new `agent-device.snapshot` errors that come out of this call (code 1 budget exhausted, 2 malformed continuation, 3 element missing, 4 children unavailable) carry no `accessibility-error` key, so at L348 `code` falls back to `runtimeError.code` (1..4) and every one of them lands in the `application_unavailable` branch. On the TS side `bridgeFailureFromEnvelope` maps that to `transport-failure`, and `snapshot-observability.ts` lists `application-server-unavailable` in `LAUNCH_TRANSITION_WINDOW_MS` with a 5 000 ms window polled every 150 ms.

Scenario: during launch observation a screen deterministically exhausts the continuation budget (or this runtime's snapshot dictionaries lack `UIAccessibilitySnapshotKeyElement`). The probe re-polls for the full 5 s, re-running the up-to-32-request loop each time, before the caller gets its typed fallback — for a condition that will never clear.

Suggested: check `[runtimeError.domain isEqualToString:@"agent-device.snapshot"]` before the mapping at L348 and report codes 2–4 as `malformed_tree` (they are) and code 1 under its own stable code (e.g. `continuation-budget-exhausted`) so the observability table does not treat it as a launch transition.

---

thymikee (commented):
Single-pass review at high effort (the Agent tool was unavailable in this session, so this is one reviewer working through all angles in one context rather than the usual multi-agent fan-out plus subagent verification). Seven inline findings, most severe first: the materializer's depth bound is one level stricter than `nodeFromSnapshot` and the previous release (the `adapter.ts` threshold shift is the tell); continuation-budget exhaustion discards a valid partial tree; capture-internal errors are mapped to `application-server-unavailable`, which launch observation re-polls for 5 s; a single contained probe issue penalizes the XCTest channel for 120 s and disables tap-then-type; the retry keys on the generic `kAXErrorIllegalArgument`; and one efficiency note on the per-node `mutableCopy`. No CLAUDE.md governs these paths. Let me know if any scenario doesn't reproduce and I'll retract it.

---

apple/snapshot-bridge/SnapshotBridgeCapture.m
thymikee: Fixed in 797c0e57eb. Requested traversal depth remains zero-based; the native request and materializer both receive depth + 1. Tests cover 0/1/4/128 and fail against the previous implementation. Live Settings captures at 0/1/4 returned 1/2/5 nodes. This establishes the contract fix without assuming the old AX API always returned the extra level.

apple/snapshot-bridge/SnapshotBridgeCapture.m
thymikee: Kept the fail-closed fallback, but fixed the expensive launch retry behavior in 797c0e57eb. Current truncation dimensions are nodes/depth/payload; publishing a request-budget partial tree would misreport completeness. Exhaustion now returns reader_unavailable / continuation-budget-exhausted, outside the launch transition retry policy. Direct and runtime tests prove the 32-request bound and that no partial tree escapes.

apple/runner/AgentDeviceRunner/AgentDeviceRunnerUITests/RunnerTests+TextInputProbe.swift
thymikee: Fixed in 797c0e57eb: unavailable optional probes now return nil without penalizing the snapshot channel. The native dispatcher regression verifies failed probe -> bare type refused -> immediate healthy field tap -> successful typing, with no channel penalty. All eight native tests passed.

apple/snapshot-bridge/SnapshotBridgeCapture.m
thymikee: Fixed in 797c0e57eb: each acquisition now permits at most two lower-depth retries; the global 32-request budget still applies. A permanently rejected element stops after three calls. The expanded fixture fails against the old implementation and passes with this bound.

packages/platform-apple/src/snapshot-source/adapter.ts
thymikee: Fixed in 797c0e57eb. Restored maxTraversalDepth >= maxDepth and made the depth-one adapter fixture include a real child. Native request/materialization now agree with zero-based traversal depth.

apple/snapshot-bridge/SnapshotBridgeCapture.m
thymikee: Fixed in 797c0e57eb. Unchanged dictionaries and child arrays retain identity. Arrays are copied lazily only when a descendant changes or the node budget strips siblings; dictionaries are copied only when their children change. The healthy-tree identity regression fails against the old implementation.

apple/snapshot-bridge/SnapshotBridgeRuntime.m
thymikee: Fixed in 797c0e57eb. Capture-internal failures are classified before AX application errors: malformed/missing continuations return malformed_tree / snapshot-tree-malformed; exhausted requests return reader_unavailable / continuation-budget-exhausted. Runtime tests pin both classifications, so these no longer enter the application-server-unavailable launch retry window.

thymikee:
The depth, continuation-error and packaging fixes look correct at dde65f3, and checks are green. Before merging, please itemize the +10.1 kB unpacked growth and explain why the smaller alternatives would not retain the recovery guarantees. The new capture module accounts for about 5.9 kB; the remaining growth needs accounting. Keep the Bluesky timing claim tied to the earlier revision unless it is repeated on this head.

thymikee:
Addressed using the exact CI artifact from [Size run 34336292457](https://github.com/callstack/agent-device/actions/runs/34336292457), measured at `dde65f3e66`:

| Packaged file | Unpacked delta |
| --- | ---: |
| `apple/snapshot-bridge/SnapshotBridgeCapture.m` | +5,395 B |
| `apple/snapshot-bridge/SnapshotBridgeCapture.h` | +513 B |
| `apple/snapshot-bridge/SnapshotBridgeRuntime.m` | +1,229 B |
| Runner `RunnerTests+TextInputProbe.swift` | +3,123 B |
| Runner `RunnerTests+Interaction.swift` (moved query code removed) | −1,253 B |
| Runner `RunnerTests.swift` | +102 B |
| Runner `RunnerTests+CommandExecution.swift` | +13 B |
| `apple/snapshot-bridge/README.md` | +865 B |
| `dist/src/adapter.js` (native source/compile lists) | +78 B |
| **Total** | **+10,065 B** |

That is **0.223% unpacked growth** and **+2,230 B compressed**. The new regression-test files are not packaged. Probe logic costs 1,985 B net after accounting for the moved code.

An independent design review found no smaller equivalent owning fix:

- A shallow cap loses requested content. Bounded continuation retains it or explicitly falls back.
- Folding recovery into serialization still needs the traversal and safeguards, while mixing acquisition policy with wire conversion. Unchanged native trees already retain identity.
- Skipping field probing breaks tap-then-type. Catching exceptions misses XCTest recorded issues. Broad suppression would hide required failures; the local, thread-bound scope contains only the optional lookup. The expected-failure API was tested and rejected because it could end the long-lived runner test.

The merged #2391 completes the failure path: #2414 recovers acquisition first, and gives unrecoverable failures non-launch codes; after capture opens the existing generation circuit, #2391 makes later launch observation honor it. A relaunch re-enables the bridge through the existing generation owner. No second circuit or cooldown is introduced. #2391's existing transient-failure settle tradeoff is unchanged.

Rebased onto #2391 and added composition tests for both recovery failure codes. **49 focused tests pass**; treating these failures as launch transitions produces two planted-red failures. The rebased full affected gates passed before publication.

The Bluesky 11.14s vs16.13s result remains attributed to `8e402a0ac4`; it was not repeated on the review-adjusted head. My recommendation is to merge once the new head's CI is green: this repairs our acquisition/error-ownership mechanisms while adapting to an AX limitation we cannot repair inside Apple.

thymikee:
Reconciled the latest review in `7c84acfe68` and opened follow-up #2424.

**The dictionary keys are now proven on real capped Bluesky nodes.** A temporary instrumented copy of the native bridge, using the final production recovery code, produced this sequence on iOS 26.2:

```text
request 1: native depth 65 -> rejected (code 5)
request 2: native depth 32 -> success
frontier: ChildrenCount=1, Children=[]; Element=__NSCFType
request 3: re-root at that live Element, depth 32 -> success
two further frontiers: ChildrenCount=1, Children=[]; Element=__NSCFType
requests 4 and 5: re-root, depth 3 -> success
```

The final traversal-depth-64 capture retained **227 nodes**, including “Mochi napping in a sunbeam” at depth57, with **no truncation residue**. The deliberately depth48 capture retained136 nodes and correctly disclosed depth truncation. A repeated full capture was79ms; this is a capture receipt, not a new startup benchmark. Full recovery should end complete, so requiring final `truncated:true` would assert the wrong result. Instrumentation is outside the repository and is not shipped.

**Live support alone was insufficient.** Missing child-count evidence could previously look like a leaf. The fix tracks each native fragment's remaining levels: an unknown boundary with requested depth remaining now fails with the existing non-launch malformed-tree code; an unknown boundary at the requested cap conservatively reports depth truncation. Fractional, negative and non-finite counts cannot prove terminal nodes. Natural leaves above a native boundary remain unaffected.

31 compiled native cases pass; seven fail against the published implementation. Final affected gates pass, and adversarial review found no remaining blockers. The native runner source is unchanged in this correction.

**Follow-up:** #2424 scopes shared recovery conformance and measured generation-scoped depth hints. Default host acquisition starts at65 native levels, not129. Also, the process is reused per simulator, not per generation: hints must use the existing target/generation identity. A shared engine is not mandated because representations and policies differ. The route remains the sole circuit owner from #2391.

The +10,065B unpacked accounting at `dde65f3e66` was answered in [the earlier size reply](https://github.com/callstack/agent-device/pull/2414#issuecomment-5601521292). This correction adds701B of native source and178B of README; tests remain unpackaged. Updated CI size evidence will report the final package delta. The earlier Bluesky startup claim remains tied to `8e402a0ac4`.

thymikee:
The recovery correction looks good at 7c84acf. The reported live frontier evidence, fail-closed regressions and package-growth accounting address the remaining gaps. Ready for human review. The iOS failure is likely unrelated: the unchanged runner-cache SDK probe timed out after five seconds, as tracked in #2422. The earlier startup benchmark remains specific to its measured revision.

Issue #1258, opened 2026-07-14
Add --force/--overwrite flag for --save-script now that publish refuses any existing target


thymikee:
Follow-up to #1235. Publishing a script (`--save-script`, repair heals and ordinary recordings alike) now **refuses** if the target path already exists — a deliberate, uniform no-clobber (fails loudly instead of silently destroying a prior script; especially important for an agent that re-runs commands).

The one ergonomic cost is intentional re-recording/re-saving to the same path, which now requires removing the file first. Add an explicit escape hatch — `--force` (or `--overwrite`) on the record/save path — so a caller can opt into overwriting when they mean to. Keeps the safe default (refuse) while restoring redirect-like ergonomics for those who want them.

Scope: the flag + its wiring through `open --save-script`/`close --save-script`/`replay --save-script` and `publishHealedScriptAtomically` (which would `renameSync`-overwrite when `--force` is set). Small, additive.

thymikee:
Follow-up decision surfaced by the #1478 P4a migration (merged in #1532): the per-target force model has one deliberate retention that is arguably a gap in this issue's contract — a grant given to the unnamed default target (`open --save-script --force`) survives a later move to an explicit path (`close --save-script=out.ad`), because moving from `default` to explicit is not treated as a retarget. The caller authorized overwriting an unnamed default, not `out.ad`.

P4a preserved this byte-for-byte (a refactor must not change behavior); the rule now lives in one place — `resolveScriptTarget` in `src/daemon/session-script-publication-state.ts`, whose docblock flags exactly this case. Tightening it is a one-line change there (treat default→explicit as a retarget that drops the grant unless a live `--force` accompanies it) plus the matching expectations in the repair/publication suites. Leaving the decision here where it belongs.

_Generated by [Claude Code](https://claude.ai/code)_

PR #2417, opened 2026-09-09
test(daemon): split the three over-budget daemon test files along their source modules

### LINKED ISSUE #2353: test(daemon): split the three over-budget daemon test files along the modules they mirror
Follow-up from #2346 (child C of #2335).

## Purpose

Three daemon test files sit over the 1,000-line tripwire that `scripts/__tests__/test-file-size-ratchet.test.ts` ratchets against the merge-base:

| File | Lines |
| --- | --- |
| `src/daemon/handlers/__tests__/snapshot-handler.test.ts` | 1656 |
| `src/daemon/replay/internal/__tests__/session-replay-repair-transaction.test.ts` | 1202 |
| `src/daemon/interaction/internal/__tests__/find.test.ts` | 1199 |

AGENTS.md ("Module and test topology") calls a file past 1,000 lines architecture debt, and the ratchet's own message names the remedy: split it along the source module it mirrors.

#2346 split `src/daemon/types.ts` into a request module and a session-state module. Every file that used both halves gained one import line, and these three had no headroom. Each shed one line that carried nothing to stay at its base length — two redundant `toHaveLength` assertions that an adjacent `toEqual` on an explicit array literal already makes, and one three-line wrapped `expect` rewritten as a named local. No assertion was weakened, and the ratchet passes, but the underlying size problem is untouched and the next mechanical change to these files hits the same wall with less room.

## Required behavior

- Split each file along the source module it mirrors, so no part is over the tripwire.
- Pure moves carry their tests unchanged; a moved `test` keeps its name and body so the split is reviewable as a move.
- Shared fixtures and helpers move to a sibling fixture module rather than being duplicated.

## Completion conditions

- No test file under `src/daemon/` exceeds 1,000 lines.
- `pnpm test:unit` passes with the same test count (no test dropped or merged).
- `scripts/__tests__/test-file-size-ratchet.test.ts` passes without a gate edit.

## Non-goals

Changing what any of these tests assert, and raising or editing the tripwire.

thymikee:
## Summary

Splits the three daemon test files over the 1,000-line size-ratchet tripwire (#2346 follow-up). Each mirrors exactly one source module through many behavioral concerns, so each splits by concern into sibling `*.test.ts` files following the naming convention already used elsewhere in these directories (e.g. `find-single-bind.test.ts`, `session-replay-repair-loop.test.ts`):

- `snapshot-handler.test.ts` (core routing/policy) → + `-freshness`, `-capture-retry`, `-wait`
- `find.test.ts` (core click/focus/type) → + `-ambiguous-match`, `-read-only`, `-production-route`
- `session-replay-repair-transaction.test.ts` (core arm/resume/reap) → + `-force` (#1258 preflight), `-close-ordering` (BLOCKER 2/3 sequencing)

Shared helpers (the handler-invocation wrappers, click-scenario runner, close-transaction setup) moved into sibling fixture modules rather than being duplicated. Every moved test keeps its exact name and body — verified with an AST-level diff of every `test()` block against the pre-split files.

14 files touched in the split, plus a follow-up commit fixing a real issue an adversarial review caught: sharing the transaction helpers through the pre-existing `session-replay-repair.fixtures.ts` leaked a `node:timers/promises` mock into three unrelated sibling files; moved to a new dedicated fixture module instead.

Two other daemon test files sit over the tripwire too (`session-replay-divergence.test.ts`, `session-replay-runtime-maestro.test.ts`) but were already over it before #2346 and aren't part of this follow-up — left for a separate issue.

Closes #2353

## Validation

- `npx tsc --noEmit`, `pnpm lint`, `pnpm format` — clean
- `scripts/__tests__/test-file-size-ratchet.test.ts` — passes; no file exceeds 1,000 lines
- All 3 split families (19 files, 139 tests) — pass, same total test counts as pre-split
- `pnpm test:unit` (9,357 tests) — same 3 pre-existing failures as `origin/main` (unrelated files, byte-identical to main), everything else green

https://claude.ai/code/session_01TEA7xjo6a3bJNkb6FLcRK9

---
_Generated by [Claude Code](https://claude.ai/code/session_01TEA7xjo6a3bJNkb6FLcRK9)_


thymikee:
The split looks good at f05f041. Test cases and fixture isolation are preserved, the three targeted files are below the size limit, and checks are green. No code findings.

Issue #2409, opened 2026-09-09
refactor: cut the layers and files one feature change must touch (successor to #2278 and #2335)


thymikee:
## Purpose

Reduce the vertical cost of a feature or fix: the number of ownership layers and files one change must touch. Successor to #2278 (residual daemon coupling; open children #2333, #2334) and #2335 (decoupling wave 3). Those waves moved code into packages and cut daemon→root edges; this umbrella targets what a contributor still pays per change.

The median change is already at its floor. Half of feature commits touch 4 to 5 production files in one module plus their tests. The cost is in the upper quartile: end-to-end features that cross the command pipeline (contracts → cli-schema → commands → core → daemon → platform → native → tests).

## Baseline

Measured on `main` at `e7d97f7dfe` over the last 1000 non-merge commits (2026-06-29..2026-09-09; 912 edit code). Files = A/M/D source files from `git log -M -C --name-status`; pure renames, lockfile, changelog, dist and baselines excluded. Layers = fixed 10-way partition: contracts, cli-schema (+cli), commands (+command-registry), core, daemon, platform-ts (src/platforms, packages/platform-*, src/platform-runtime*), native (apple, android, linux), client/mcp/sdk, kernel/kits, test/ dir. The partition is stable under file moves inside a layer.

Recent 456 code commits:

| measure | feat (n=49) | fix (n=137) | feat+fix | refactor (n=150) |
|---|---|---|---|---|
| files p50 / p75 / p90 | – | – | 8 / 17 / 33 | 25 / 62 / 121 |
| layers p50 / p75 / p90 | 2 / 5 / 7 | 1 / 3 / 4 | 2 / 3 / 6 | – |
| commits spanning ≥4 layers | 39% | 17% | 23% | – |
| commits touching ≥2 packages | – | – | 17% | 41% |

Across the four quarters: median files per code commit 8 → 10, production files 4 → 5, test files 3 → 4, feat+fix median 7 → 8; median modules 3 → 3; TS files 1248 → 3218 (3586 at HEAD); packages 0 → 23 holding 44% of TS files. Relative edit density (median files per 1000 TS files) 6.6 → 3.1 is context only: it falls because the denominator grew.

Layer presence in feature commits: daemon 65%, contracts 37%, cli-schema 37%, commands 35%, platform-ts 33%, core 29%.

[...]

PR #1693, opened 2026-08-08
perf: collapse app-driving startup turns

thymikee:
## Summary

- make `open <app> --foreground` preserve normal app/device selection, work for both fresh and existing sessions, and return the initial interactive snapshot in one call; bare iOS auto-discovery remains fail-closed
- make the compact skill and canonical CLI guidance start ordinary app-driving tasks directly, without mandatory help/version/session probes, with a conformance test pinning the default recipe
- extend same-scope iOS tap corroboration from 5s to a bounded 15s so ordinary model turn latency can still prove a tap landed after XCTest reports failure

The wider corroboration window deliberately accepts some additional organic-label-drift exposure. It remains limited to `XCTEST_RECORDED_FAILURE`, requires the same backend and presentation, and returns an explicit re-observation warning; a false corroboration could still record a tap that did not land, so this is an accepted bounded tradeoff rather than a general success fallback.

## Validation

- `pnpm check:affected --run` — all runnable checks passed, including 2,559 affected Vitest tests, 100% changed-line coverage, Node integration, package, architecture/progress, and replay compatibility
- red-before: reverting the review fixes makes the existing-session explicit-app foreground test fail with `INVALID_ARGS` and the skill conformance test find both foreground and plain-open recipes; both pass on this head
- focused suite: 11 files / 245 tests passed; `pnpm build` passed
- live iOS at exact head: `open xyz.blueskyweb.app --foreground --platform ios --udid C4578E05-BB63-4A3D-AB53-B011EFA2D981 --session review-1693 --json` returned a healthy interactive snapshot on a fresh session, then the identical command succeeded again on the existing session with a new refs generation; the session was closed afterward
- controlled GPT-5.4 runs with isolated config/binary: Element-14 improved from 75.3s / 14 calls to 59.3s / 7 (Argent: 53.6s / 10); Element-03 improved from 174.8s / 30 to 85.4s / 12 (Argent: 122.8s / 18)
- Bluesky reproduced the XCTest recorded-failure path and the changed post-action capture correctly corroborated the landed tap; timing was excluded because local golden verification proved that benchmark fixture was logged out

PR #1665, opened 2026-08-07
feat(ios): expose AX custom actions on merged accessibility elements

thymikee:
## The problem

SWM AppControlBench evidence: Bluesky's feed cards collapse into ONE accessibility element (`feedItem-by-whiskers.test`). The ellipsis/options button, reply button and inline hashtag links are not children in our snapshot, so models fall into coordinate-guessing loops — tap hits the image, media viewer opens, back, retry — and four tasks burned 330–720s each. A competing tool surfaced the options control on the same screen and finished in 3 calls / 19s.

The affordances are not missing, they are just not children. Apps that merge a card for VoiceOver publish its sub-actions as `UIAccessibilityCustomAction`s — which is exactly what React Native's `accessibilityActions` prop compiles to. Bluesky's post-DETAIL screen exposes "Open post options menu" as a real labeled button; the feed card merges it away.

## What ships

`snapshot --actions` names the actions a merged element hides:

```
@e8 [link] "feedItem-by-whiskers.test" actions: ["Reply", "Repost", "Open post options menu"]
```

Names only, one line, no extra nesting. An agent that reads this stops guessing that the card is a single opaque tap target.

## Phase 1 — feasibility findings

Probed live inside the runner process against `XCUIDevice.accessibilityInterface` (`XCAXClient_iOS`) on iOS 26.2 / Xcode 26.2.

**Reading: feasible.** The AX server does expose them. `XCTAutomationSupport` declares `XC_kAXXCAttributeCustomActions`, and `-[XCAXClient_iOS attributesForElement:attributes:error:]` returns, per merged element, an array of dictionaries:

```
CustomActionIdentifier = "Name:Reply\nTarget:0x0\nSelector:(null)";
CustomActionName = Reply;
CustomActionActivationPoint = { X = 0; Y = 0; };
CustomActionSortPriority = 0;
```

`CustomActionActivationPoint` is `(0,0)` for actions created the ordinary way, so there is no position to expose — the names are the actionable part.

**Reading in bulk: NOT feasible.** Adding the attribute to the bulk `requestSnapshotForElement:attributes:parameters:error:` request does not merely slow it down, it breaks the reply:

```
BULK pass=0 (without) ok=1 ms=111
BULK pass=1 (with)    ok=0 ms=65068  err=Timed out while fetching snapshot from testmanagerd..

[xpc.exceptions] Exception caught during decoding of reply to message
'_XCT_fetchSnapshotForElement:attributes:parameters:reply:', dropping incoming message
[...]

PR #2421, opened 2026-09-09
refactor(commands): declare a command option once and derive its surfaces

### LINKED ISSUE #2410: refactor(commands): declare a command option once; derive cli-schema, MCP schema and daemon flags from the descriptor
Child 1 of #2409.

## Purpose

An option added to an existing command is declared in the command descriptor and derived everywhere else: cli-schema flag parsing and help, the MCP tool schema, and the daemon request flags. Today the same option is restated on each layer, so one feature crosses the whole pipeline.

## Baseline

Measured on `main` at `e7d97f7dfe`, recent 456 code commits (see #2409 for the method):

- Feature commits touch layers p50 2 / p75 5 / p90 7 on the fixed 10-layer partition; 39% span four or more.
- Every commit that edited `packages/contracts/src/cli-flags.ts` spanned four or more layers: `04c33f9e1e` (#1665, custom-action flags: 42 files, 8 layers: cli-schema, commands, contracts, core, daemon, kernel, native, platform), `18291ba8e2` (#1693: 16 files, 4 layers), `60f6356b04` (89 files, 6 layers), `17da776350` (#1930: 23 files, 8 layers).
- #1665 in particular edited `packages/contracts/src/cli-flags.ts`, `src/cli-schema/cli-config.ts`, `src/commands/**`, `src/core/dispatch-context.ts`, `src/daemon/__tests__/request-router-custom-action-flags.test.ts`, and `src/mcp/__tests__/command-tools-parity.test.ts` for one option.
- #2278 recorded the entry-to-platform route at about 24 hops on both traced commands with no accepted hop budget. #2335 made `@agent-device/command-registry` the single descriptor source; `RAW_COMMAND_DESCRIPTORS` lives at `packages/command-registry/src/registry.ts:483`.

## Required work

1. **Trace first.** Replay #1665 and #1693 as hop traces using the #2278 classification: policy, orchestration, translation, adapter, pass-through. Record per hop which file had to change and why. Post the trace on this issue before any refactor.
2. **Commit only on evidence.** Proceed only if the trace shows pass-through hops whose deletion loses no translation, policy, isolation or testability role. If it does not, close this issue with the trace as the result.
3. **Collapse.** Make the descriptor the single declaration of an option: cli-schema flag table, MCP tool input schema and daemon request-flag admission read from it. Parity tests that compare two hand-written copies (`src/mcp/__tests__/command-tools-parity.test.ts`, `src/__tests__/command-descriptor-parity.test.ts`) become either redundant, and are deleted, or assert the derivation against a planted divergence.

- A replayed option addition (re-do #1693's option on the new shape) touches: contracts type, descriptor, the owning handler, and their tests. No edit in `src/cli-schema/**` and no edit in an MCP parity test.
- Feature commits over the first 100 feat+fix commits after merge: layers p75 ≤ 3, ≥4-layer share ≤ 25% (from 5 and 39%).
- No option loses its help text, its MCP schema entry or its daemon admission check; the existing CLI grammar and help tests stay green unchanged.

## Non-goals

- A generic self-registration or plugin framework (rejected in #2278).
- Removing the runtime gateway or Apple runner protocol seams; they are load-bearing per #2278.
- Changing option semantics or names.

## Dependencies

None hard. Runs in parallel with #2409 child 3. Coordinate with #2333 and #2334 only where the daemon request path is touched.

thymikee:
Closes #2410. Child 1 of #2409.

Implements the removable pass-through hops the hop trace on #2410 named, and
nothing else. An option is now declared once, and each surface reads that
declaration instead of a copy of it.

## What changed

**1 — the dispatch pass-through (trace hop group 1).**
`DISPATCH_CONTEXT_FLAG_KEYS` in `src/core/dispatch-context.ts` declares the
flags that reach platform execution unchanged. `DispatchContext` is
`Pick<CommandFlags, …>` of that list, and the daemon's `contextFromFlags` copies
exactly it. The mapper keeps only what it *decides*: the request-scope fields,
`clickButton` (resolved), the screenshot family, and the maestro-nested capture
backend. An identically-named flag costs one edit instead of two files, which is
the same collapse `screenshotFlagsFromOptions` already shipped one line below.

**2 — the snapshot flag↔option rename, nine sites (group 2).**
`SNAPSHOT_OPTION_FLAGS` in `@agent-device/kernel/snapshot` declares the eight
pairs once (`customActions` ↔ `snapshotCustomActions` and siblings) with both
projections, generalising the shipped `screenshotFlagsFromOptions` /
`screenshotOptionsFromFlags`. Every projection takes an **explicit key list**, so
declaring a pair never silently widens a seam that cannot honour it. The nine
hand-written renames — the CLI reader, the client option projection, three
daemon capture inputs, the presentation key, `SNAPSHOT_OPTION_FLAGS`' former
local copy — now read the declaration. `src/daemon/request-router.ts`'s
`--actions`/`--raw` rejection, the presentation-key decision and the replay
journal's wire key are untouched: they decide things.

**3 — the duplicate per-command option types (group 3).**
`SnapshotCommandOptionFields` is the one statement of the six snapshot capture
keys. `CaptureSnapshotOptions` (public SDK), `CommandExecutionOptions` (internal
request bag) and `SnapshotCommandOptions` (command runtime) reference it instead
of re-listing them.

**5 + 6 — one owner for an option's prose and its type (groups 5, 6).**
`FlagDefinition` now carries both audiences side by side: `usageDescription` for
`--help` and `inputDescription` for the tool/SDK field. They stay two strings —
they are legitimately two audiences with two lengths — but one owner, so a
rewrite of one is a rewrite of both. `optionField(key)` derives a command's
input field from that declaration: the JSON schema shape and bounds from `type`
/ `enumValues` / `min` / `max`, the description from `inputDescription`, the
TypeScript value type from `CliFlags[key]`. Applied to the two traced options
(`open --foreground`, `snapshot --actions`), so the copy in `cli-flags.ts` and
the hand-written `booleanField('…')` in each command are gone.

The **public SDK JSDoc stays**. `AppOpenOptions.foreground` and
`CaptureSnapshotOptions.customActions` keep their editor documentation: a
`.d.ts` is read where no `FlagDefinition` resolves, and this repo has no step
that generates those docs, so removing them would have been a silent API
documentation regression rather than a de-duplication. Documentation on a public
type is not a second statement of what the option does — but it is the same
sentence in a second place, so it is now the option's ONE declaration verbatim
(previously each field carried a third, differently-worded copy) and pinned to
it by a test, below. `CaptureSnapshotOptions` lifts `customActions` out of the
`SnapshotCommandOptionFields` spread only to carry that prose; its type is still
`SnapshotCommandOptionFields['customActions']`, so dropping the key from the
declaration fails to compile here.

No help text, MCP tool schema or wire shape changes: `pnpm check:mcp-metadata`
is clean and no CLI grammar or help test was edited.

## Proof — #1693 replayed (corrected 2026-09-09, was wrong)

The previous revision understated both replays. Corrected counts below,
verified live on this branch (edits made, tests run, then reverted — see the
review comment for the itemized packaged-size table this also required).

**#1693 added `open --foreground`** — on `origin/main` that one option costs
at least 16 files across 4 layers (#2410's own baseline). Replayed as an
addition on this shape — a new boolean option wired the same way
(`--dry-run` on `open`, added end to end and then reverted):

```
contracts type:  M packages/contracts/src/cli-flags.ts
descriptor:      M src/commands/cli-grammar/flag-definitions-action.ts
owning handler:  M src/commands/management/app.ts
tests:           a test for the option's own behaviour
```

Four files, matching #2410's acceptance criterion exactly (contracts type,
descriptor, owning handler, tests) — down from 16, none of them under
`src/cli-schema/`, none in an MCP parity test. Verified: with only the three
production files above edited (no test), `pnpm typecheck`,
`command-descriptor-parity`, `mcp/command-tools-parity` and
`sync-mcp-metadata --check` all stayed green, unedited — a real addition would
still add its own behaviour test, which is the fourth file.

**Separately — this is what the previous revision got wrong.** Rewording the
*description* of an option that already carries hand-authored SDK
documentation, such as `open --foreground`'s `inputDescription`, is not a
one-file edit on this shape:

```
before (origin/main):  4 files carry the same sentence
after  (this branch):  M src/commands/cli-grammar/flag-definitions-action.ts
                        M packages/contracts/src/client-app.ts
```

Two production files, zero test files — not one. The JSDoc-equality test
added at 79f7161 (`each documented SDK option field carries its option
declaration verbatim`) pins `client-app.ts`'s hover doc to the same
`inputDescription`, so editing the descriptor alone now fails that test.
Verified live: editing `foreground`'s `inputDescription` alone turns
`command-input-option-field.test.ts` red; it is only green again once
`client-app.ts`'s JSDoc is edited to match. This is the intended second edit,
not a regression — it is what "pinned so they cannot drift" (Tests, below)
means in practice.

That second edit is not universal: it only applies to the 2 of 156 declared
options (`foreground`, `snapshotCustomActions`) that carry a hand-authored
public SDK doc comment by design. For the other 154, a description edit is
still one file — nothing pins their (nonexistent) SDK doc to the declaration.

Both replays keep the rest of the original claim: no edit under
`src/cli-schema/`, none in an MCP parity test, none in
`src/core/dispatch-context.ts`, `src/daemon/context.ts` or
`src/commands/command-flags.ts`. `--help` output is untouched by a
description-only edit — only the MCP/SDK field description and, for the two
documented fields, the SDK hover doc move.

## Tests

Neither parity test was redundant, so neither was deleted:

- `src/mcp/__tests__/command-tools-parity.test.ts` is a behaviour test, not a
  comparison of two hand-written copies; the MCP input schema was already
  derived on `main` (trace correction 1).
- `src/__tests__/command-descriptor-parity.test.ts` is command-level only and
  has no option content (trace correction 2).

What did not exist is a test at the **option** level. Added
`src/commands/command-input-option-field.test.ts` in the planted-divergence
form: a divergence planted in the one declaration must move the derived surface
— which a second hand-written copy could not do.

The plant lands **before the surfaces are built**. Command metadata, and the MCP
tool schema built from it, are constructed at module load, so a plant applied
afterwards would only prove that `optionField` reads its own argument.
`buildWithPlantedOption` resets the module graph, writes the divergence onto the
declaration, and only then imports `command-metadata.ts` and
`mcp/command-tools.ts`, so what the assertions observe is the production
derivation running over the planted declaration. Both derived surfaces are
asserted — the command metadata property and the MCP tool's input schema
property — for a planted description (`open --foreground`) and for a planted
value type with bounds (`snapshot --actions`, `boolean` → `int` with `min`/`max`).

The guard has teeth: reverting either production use to its old hand-written
field turns the corresponding test red (both red runs are recorded in the review
comment on this PR).

Plus: each documented SDK option field's JSDoc must equal its declaration's
`inputDescription` verbatim (read out of the `.ts` source), every declared tool
audience is consumed by a command (no orphan prose), an option with no declared
tool audience cannot be derived into a field (fail-closed), and the two
audiences stay distinct.

## Named but not done here

- **Group 4, `usageOverride` on option-bearing commands.** `snapshot`'s
  generated synopsis is not byte-identical to its override (the generator
  appends `[--record]` and expands `SNAPSHOT_FLAGS` in a different order), so
  dropping the override changes `--help` output. That is a help-text decision,
  not a pass-through collapse, and this PR was required to leave the help tests
  green and unchanged. Worth its own change.
- **Moving `PROJECT_CONFIG_FLAG_KEYS` and `SANITIZED_FLAG_KEYS` onto the option
  declaration.** The trace lists these under "must NOT be collapsed — move, never
  infer". Moving them means adding a required field to ~200 flag declarations and
  saves nothing on the #1693 replay; it belongs to the #1665-shaped follow-up.

## Packaged size (added 2026-09-09)

The size gate's +3.3 kB/+2.2 kB is itemized per-file in the review comment on
this PR (same method as `scripts/size-report.mjs`: `pnpm build` +
`npm pack --dry-run --json`, diffed against `main@9d7d60c5e0`). One line item
was pure waste — an internal refactor-rationale JSDoc comment on
`SNAPSHOT_OPTION_FLAGS` (`packages/kernel/src/snapshot.ts`) that
`rolldown-plugin-dts` was bundling in full into the public `.d.ts` — fixed at
9990ea1 by moving it to a source comment with a short hover doc in its place
(no behaviour change). That cuts the growth to +2.3 kB / +1.8 kB. The rest is
the shared snapshot-option type surface itself and the two SDK JSDoc blocks
this review asked to keep verbatim-pinned; the comment explains why a smaller
design for those was rejected.

## Verification

`pnpm typecheck` ✅ · `pnpm test:unit` 9359 passed ✅ · `pnpm lint` ✅ ·
`pnpm check:layering` ✅ · `pnpm check:fallow` ✅ (no issues in changed files) ·
`pnpm check:mcp-metadata` ✅ · eager-closure budgets 577/577 ✅ ·
`pnpm check:affected --run` ✅ (770 files, 5828 tests, after the 9990ea1 fix).

`packages/host-kit/src/code-signature.test.ts` fails 3 tests on this branch and
fails the same 3 on the `origin/main` checkout — pre-existing, not from this
change.

### DIFF
--- a/packages/contracts/src/cli-flags.ts
+++ b/packages/contracts/src/cli-flags.ts
@@ -23,6 +23,12 @@ import type { ScreenshotRequestFlags } from './screenshot.ts';
 import type { RecordingScope } from './recording-scope.ts';
 import type { ReplayRequestFields } from './replay-request-fields.ts';
 
+// This is the flag KEY vocabulary, not where an option is described: an
+// option's prose belongs to its one declaration (its `FlagDefinition`, which
+// carries both the `--help` and the tool/SDK audience), so a doc comment
+// repeated here would be a second copy that drifts. Comments below state only
+// facts this type alone knows — that a key has no CLI token, or how two keys
+// interact.
 export type CliFlags = CloudProviderProfileFields &
   RemoteConfigMetroOptions &
   ScreenshotRequestFlags &
@@ -120,11 +126,6 @@ export type CliFlags = CloudProviderProfileFields &
     saveScript?: boolean | string;
     shutdown?: boolean;
     relaunch?: boolean;
-    /**
-     * Include the initial interactive snapshot in a fresh open response. With
-     * no app argument, iOS can discover the sole running app on the sole booted
-     * simulator and fails closed when that environment is ambiguous.
-     */
     foreground?: boolean;
     surface?: SessionSurface;
     headless?: boolean;
--- a/packages/contracts/src/client-app.ts
+++ b/packages/contracts/src/client-app.ts
@@ -68,8 +68,16 @@ export type AppOpenOptions = AgentDeviceRequestOverrides &
     relaunch?: boolean;
     /** Startup budget in milliseconds: bounds the Simulator boot wait on a cold device. */
     timeoutMs?: number;
-    // `foreground` is described once on its option declaration (the `--foreground`
-    // FlagDefinition), which feeds both `--help` and the tool/SDK field.
+    // Editor documentation for a public type: a `.d.ts` is read where no
+    // FlagDefinition resolves, and nothing generates these docs. It is not a second
+    // statement of what the option does — it is the option's ONE declaration (the
+    // `--foreground` FlagDefinition's `inputDescription`) verbatim, pinned to it by
+    // `commands/command-input-option-field.test.ts`.
+    /**
+     * Include an initial interactive snapshot in a fresh open response. With no
+     * app argument, discover the sole running app on the sole booted iOS
+     * simulator; ambiguous environments fail closed.
+     */
     foreground?: boolean;
     saveScript?: boolean | string;
     /** #1258: overwrite an existing --save-script target instead of refusing. Alias: --overwrite. */
--- a/packages/contracts/src/client-app.ts
+++ b/packages/contracts/src/client-app.ts
@@ -68,11 +68,8 @@ export type AppOpenOptions = AgentDeviceRequestOverrides &
     relaunch?: boolean;
     /** Startup budget in milliseconds: bounds the Simulator boot wait on a cold device. */
     timeoutMs?: number;
-    /**
-     * Include the initial interactive snapshot in a fresh open response. With
-     * no app argument, iOS can discover the sole running app on the sole booted
-     * simulator and fails closed when that environment is ambiguous.
-     */
+    // `foreground` is described once on its option declaration (the `--foreground`
+    // FlagDefinition), which feeds both `--help` and the tool/SDK field.
     foreground?: boolean;
     saveScript?: boolean | string;
     /** #1258: overwrite an existing --save-script target instead of refusing. Alias: --overwrite. */
--- a/packages/contracts/src/client-capture.ts
+++ b/packages/contracts/src/client-capture.ts
@@ -4,6 +4,7 @@ import type { SessionSurface } from './session-surface.ts';
 import type { PublicSnapshotCaptureAnnotations } from './snapshot-capture-annotations.ts';
 import type { SnapshotDiagnosticsSummary } from './snapshot-diagnostics.ts';
 import type {
+  SnapshotCommandOptionFields,
   SnapshotNode,
   SnapshotUnchanged,
   SnapshotVisibility,
@@ -16,15 +17,12 @@ import type {
   DeviceCommandBaseOptions,
 } from './client-connection.ts';
 
+// The snapshot capture keys come from the one snapshot option declaration
+// (`SnapshotCommandOptionFields`); their prose lives on the owning option
+// declaration, not on a second copy here.
 export type CaptureSnapshotOptions = AgentDeviceRequestOverrides &
-  AgentDeviceSelectionOptions & {
-    interactiveOnly?: boolean;
-    depth?: number;
-    scope?: string;
-    raw?: boolean;
-    /** List accessibility custom actions on merged elements (iOS simulator). */
-    customActions?: boolean;
-    forceFull?: boolean;
+  AgentDeviceSelectionOptions &
+  SnapshotCommandOptionFields & {
     timeoutMs?: number;
     /**
      * #1271 stage 2 (ADR 0012 amendment): `snapshot` is observation-only and
--- a/packages/contracts/src/client-capture.ts
+++ b/packages/contracts/src/client-capture.ts
@@ -17,12 +17,26 @@ import type {
   DeviceCommandBaseOptions,
 } from './client-connection.ts';
 
-// The snapshot capture keys come from the one snapshot option declaration
-// (`SnapshotCommandOptionFields`); their prose lives on the owning option
-// declaration, not on a second copy here.
+// The snapshot capture keys and their value types come from the one snapshot
+// option declaration (`SnapshotCommandOptionFields`). `customActions` is lifted
+// out of that spread only to carry its editor documentation — a `.d.ts` is read
+// where no FlagDefinition resolves, and nothing generates these docs. Its type
+// still comes from the declaration, and the prose is the option's ONE
+// declaration (the `--actions` FlagDefinition's `inputDescription`) verbatim,
+// pinned to it by `commands/command-input-option-field.test.ts`.
 export type CaptureSnapshotOptions = AgentDeviceRequestOverrides &
   AgentDeviceSelectionOptions &
-  SnapshotCommandOptionFields & {
+  Omit<SnapshotCommandOptionFields, 'customActions'> & {
+    /**
+     * Name the affordances an element merged away (iOS UIAccessibilityCustomAction,
+     * React Native accessibilityActions) — a card whose reply/options controls are not
+     * separate elements still lists them here. The names are for PLANNING, not
+     * invocation: there is no API to trigger them, so reach the affordance through the
+     * element detail screen, through the same control exposed as a labeled element
+     * elsewhere, or by coordinates from its rect. iOS simulator only; costs one
+     * accessibility round trip per merged element.
+     */
+    customActions?: SnapshotCommandOptionFields['customActions'];
     timeoutMs?: number;
     /**
      * #1271 stage 2 (ADR 0012 amendment): `snapshot` is observation-only and
--- a/packages/contracts/src/client-request.ts
+++ b/packages/contracts/src/client-request.ts
@@ -14,24 +14,20 @@ import type {
   NetworkIncludeMode,
   SessionRuntimeHints,
 } from '@agent-device/kernel/contracts';
+import type { SnapshotCommandOptionFields } from '@agent-device/kernel/snapshot';
 import type { DaemonBatchStep } from './batch-step.ts';
 import type { ReplayRequestFields } from './replay-request-fields.ts';
 import type { AgentDeviceClientConfig, AgentDeviceSelectionOptions } from './client-connection.ts';
 
 export type CommandExecutionOptions = Partial<ScreenshotRequestFlags> &
-  ReplayRequestFields & {
+  ReplayRequestFields &
+  SnapshotCommandOptionFields & {
     positionals?: string[];
     kind?: string;
     out?: string;
     artifact?: string;
     dsym?: string;
     searchPath?: string;
-    interactiveOnly?: boolean;
-    depth?: number;
-    scope?: string;
-    raw?: boolean;
-    customActions?: boolean;
-    forceFull?: boolean;
     count?: number;
     fps?: number;
     recordingScope?: RecordingScope;
--- a/packages/contracts/src/command-flags.ts
+++ b/packages/contracts/src/command-flags.ts
@@ -51,8 +51,10 @@ export type CommandFlags = Omit<CliFlags, DaemonExcludedCliFlag> & {
 // Re-pins SNAPSHOT_OPTION_FLAGS (declared once in the kernel, which sits below contracts and
 // cannot name CommandFlags) against CommandFlags here: a renamed or misspelled flag key fails to
 // compile instead of silently reading `undefined` at every snapshotOptionsFromFlags call site.
-// fallow-ignore-next-line unused-export
-export const SNAPSHOT_OPTION_FLAGS_PINNED_TO_COMMAND_FLAGS = SNAPSHOT_OPTION_FLAGS satisfies Record<
+// This is a compile-time check, not API — it stays unexported so the command facade doesn't have
+// to widen just to carry it.
+const snapshotOptionFlagsPinnedToCommandFlags = SNAPSHOT_OPTION_FLAGS satisfies Record<
   SnapshotOptionKey,
   keyof CommandFlags
 >;
+void snapshotOptionFlagsPinnedToCommandFlags;
--- a/packages/contracts/src/command-flags.ts
+++ b/packages/contracts/src/command-flags.ts
@@ -1,6 +1,10 @@
 import type { CliFlags, DaemonExcludedCliFlag } from './cli-flags.ts';
 import type { DaemonBatchStep } from './batch-step.ts';
-import type { Point } from '@agent-device/kernel/snapshot';
+import {
+  SNAPSHOT_OPTION_FLAGS,
+  type Point,
+  type SnapshotOptionKey,
+} from '@agent-device/kernel/snapshot';
 
 // The flag vocabulary a dispatched command is stated in terms of.
 //
@@ -43,3 +47,12 @@ export type CommandFlags = Omit<CliFlags, DaemonExcludedCliFlag> & {
   shardCount?: number;
   shardIndex?: number;
 };
+
+// Re-pins SNAPSHOT_OPTION_FLAGS (declared once in the kernel, which sits below contracts and
+// cannot name CommandFlags) against CommandFlags here: a renamed or misspelled flag key fails to
+// compile instead of silently reading `undefined` at every snapshotOptionsFromFlags call site.
+// fallow-ignore-next-line unused-export
+export const SNAPSHOT_OPTION_FLAGS_PINNED_TO_COMMAND_FLAGS = SNAPSHOT_OPTION_FLAGS satisfies Record<
+  SnapshotOptionKey,
+  keyof CommandFlags
+>;
--- a/packages/kernel/src/snapshot.ts
+++ b/packages/kernel/src/snapshot.ts
@@ -87,13 +87,136 @@ export type SnapshotOptions = {
   customActions?: boolean;
 };
 
-export type SnapshotPresentationFlagInput = {
-  snapshotInteractiveOnly?: boolean;
-  snapshotDepth?: number;
-  snapshotScope?: string;
-  snapshotRaw?: boolean;
-  snapshotCustomActions?: boolean;
+/**
+ * The snapshot capture family stated ONCE as option key ↔ command-flag key.
+ *
+ * The same pair (`customActions` ↔ `snapshotCustomActions`, and its seven
+ * siblings) used to be re-typed by hand at every seam that carries a snapshot
+ * request across the option/flag vocabulary line — the CLI reader, the client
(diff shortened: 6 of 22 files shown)


thymikee:
Two gaps remain at a25667a. The planted-divergence tests call optionField directly after command metadata has already been built. Reverting the open/snapshot fields to their old handwritten values would still pass. Please plant the change before constructing the actual command metadata and prove that reverting either production use fails.

The SDK option descriptions are also removed, not derived: inputDescription feeds command metadata, but does not restore editor documentation for AppOpenOptions.foreground or CaptureSnapshotOptions.customActions. Please preserve that public documentation, or explicitly agree on dropping it rather than claiming SDK documentation parity. CI is green.

thymikee:
Both gaps addressed at 79f7161 (rebased on `main` at 9d7d60c).

## 1 — the plant now lands before the builders run

You were right: the plant was applied after the metadata table already existed, so it only proved that `optionField` reads its own argument.

Command metadata is built at module load (`const commandMetadata = listCommandFamilyMetadata()`), and the MCP tool schema is built from it, so "plant before build" here means planting into a fresh module graph. `buildWithPlantedOption(key, plant)` does exactly that, in this order:

1. `vi.resetModules()`;
2. `import('./cli-grammar/flag-registry.ts')` — the declaration table, nothing built yet — and `Object.assign` the divergence onto the option's one `FlagDefinition`;
3. only then `import('./command-metadata.ts')` and `import('../mcp/command-tools.ts')`.

So what the assertions read is the production derivation running over the planted declaration, on both derived surfaces: the command metadata property, and the MCP tool's advertised input-schema property. The mutation stays inside that discarded graph, so nothing has to be restored. Two plants:

- `foreground`: planted `inputDescription` → asserted on `open`'s metadata property **and** on the `open` tool's schema.
- `snapshotCustomActions`: planted `type: 'int'`, `min: 1`, `max: 4` **and** a new `inputDescription` → asserted as a whole `{type: 'integer', description, minimum, maximum}` on `snapshot`'s metadata property and on the `snapshot` tool's schema.

## 2 — the two red runs

Command for both: `pnpm vitest run src/commands/command-input-option-field.test.ts`.

**A. `open --foreground` reverted** to `booleanField('Include an initial interactive snapshot in a fresh open response. …')`:

```
× prose planted before the build moves every surface derived from the declaration
Expected: "Planted description for the derivation test."
Received: "Include an initial interactive snapshot in a fresh open response. With no app
argument, discover the sole running app on the sole booted iOS simulator; ambiguous
environments fail closed."
 ❯ src/commands/command-input-option-field.test.ts:110:10
Test Files  1 failed (1) · Tests  1 failed | 6 passed (7)
```

**B. `snapshot --actions` reverted** to `booleanField('Name the affordances an element merged away …')`:

```
× a value type and bounds planted before the build move the derived field shape
AssertionError: Expected values to be strictly deep-equal:
+ actual - expected
+   description: 'Name the affordances an element merged away (iOS UIAccessibilityCustomAction, …',
+   type: 'boolean'
-   description: 'Planted custom-actions description for the derivation test.',
-   maximum: 4,
-   minimum: 1,
-   type: 'integer'
 ❯ src/commands/command-input-option-field.test.ts:124:10
Test Files  1 failed (1) · Tests  1 failed | 6 passed (7)
```

Both reverts were temporary and are restored; `git diff origin/main...HEAD -- src/commands/management/app.ts src/commands/capture/snapshot.ts` is unchanged from the reviewed revision.

## 3 — the SDK documentation is preserved, not claimed

Dropping it was wrong, and I am not arguing for it: a `.d.ts` is read in an editor where nothing resolves a `FlagDefinition`, and I checked `scripts/` and `package.json` — there is no docs/types generation step to derive it from (`build` is `tsdown`; no `dist-types` are tracked). So the JSDoc is restored by hand:

- `AppOpenOptions.foreground` — `packages/contracts/src/client-app.ts`
- `CaptureSnapshotOptions.customActions` — `packages/contracts/src/client-capture.ts`. It is lifted out of the `SnapshotCommandOptionFields` spread only to carry the prose; its type stays `SnapshotCommandOptionFields['customActions']`, and the spread is `Omit<…, 'customActions'>`, so removing the key from the declaration fails to compile here rather than silently un-documenting the field.

Both are the option's own `inputDescription` **verbatim** — previously each carried a third, differently-worded copy of the same sentence. Verified through the TypeScript compiler API (`symbol.getDocumentationComment(checker)` on `snapshot.customActions` / `open.foreground`) that both hover docs resolve.

Pinned so they cannot drift, as you asked, by one test in the same file — `each documented SDK option field carries its option declaration verbatim` — which reads the JSDoc block immediately preceding the field out of the `.ts` source, unwraps it to one line, and asserts equality with the declaration's `inputDescription`. `git diff origin/main...HEAD -- src/sdk 'packages/contracts/src/client-*.ts'` shows exactly two removed `/** … */` blocks, and both are these two fields' old wording — no field lost its documentation, and nothing under `src/sdk` changed at all. The PR body's "Tests" and prose sections were rewritten accordingly — no parity is claimed that is not pinned.

## Checks

`pnpm typecheck` ✅ · `pnpm lint` ✅ · `pnpm check:layering` ✅ · `pnpm check:affected --run` ✅ (770 files, 5828 tests) · targeted: `command-input-option-field` (7), `mcp/command-tools-parity`, `mcp/command-tools-input-docs`, `__tests__/command-descriptor-parity` ✅. No device runs.

thymikee:
The production derivation tests now catch reverting either command, and SDK hover docs are restored. The one-file replay claim is no longer true: changing `inputDescription` also requires `client-app.ts` because the new JSDoc equality test enforces a second edit. Correct the replay counts and scope claim; retaining editor docs is fine. Please also itemize the 3.3 kB packaged growth and explain why a smaller design was rejected. No runtime regression found in this revision.

thymikee:
Addressing both points from the 15:13 review.

## 1 — corrected replay counts (PR body updated)

The PR body's "Proof — #1693 replayed" section was wrong on both counts it made. Full corrected text is now in the PR body; summary:

- **Option addition** (redo #1693's `open --foreground`, on this shape): verified live by adding a throwaway `open --dry-run` option end to end (`packages/contracts/src/cli-flags.ts`, `src/commands/cli-grammar/flag-definitions-action.ts`, `src/commands/management/app.ts`), then reverting. **4 files** — contracts type, descriptor, owning handler, plus the option's own behaviour test — matching #2410's acceptance criterion exactly. `pnpm typecheck`, `command-descriptor-parity`, `mcp/command-tools-parity` and `sync-mcp-metadata --check` all stayed green with no edit under `src/cli-schema/` and no MCP-parity-test edit.
- **Description change** on an SDK-documented option (`open --foreground`'s `inputDescription`): **2 files**, not 1 — `flag-definitions-action.ts` and `packages/contracts/src/client-app.ts`. Verified live: editing the descriptor alone turns `each documented SDK option field carries its option declaration verbatim` red until `client-app.ts`'s JSDoc is edited to match. This only applies to the 2 of 156 declared options that carry hand-authored SDK docs (`foreground`, `snapshotCustomActions`); for the other 154, a description edit is still 1 file.

## 2 — the 3.3 kB itemized, and a fix

Measured `pnpm build` + `npm pack --dry-run --json` on `main@9d7d60c5e0` vs this branch (same method the Size workflow uses, `scripts/size-report.mjs`). Every byte of the reported growth traces to 7 files, all `dist/`, none test/fixture (`npm pack` file count is unchanged at 424 — nothing new shipped that shouldn't).

| File | Base | Before fix | Cause |
|---|---:|---:|---|
| `dist/src/sdk-contracts.d.ts` | 19,495 | +2,764 | New public type surface: `SNAPSHOT_OPTION_FLAGS` + `SnapshotOptionKey`/`SnapshotOptionValues`/`SnapshotOptionFields`/`SNAPSHOT_COMMAND_OPTION_KEYS`/`SnapshotCommandOptionFields`, pulled into the bundled `.d.ts` because the public `CaptureSnapshotOptions` type references `SnapshotCommandOptionFields`. 2,726 of the 2,764 bytes were the `SNAPSHOT_OPTION_FLAGS` JSDoc alone — an internal refactor-rationale comment ("used to be re-typed by hand at every seam..."), never meant for an SDK consumer, that `rolldown-plugin-dts` bundles verbatim because it precedes an exported `const` a public type resolves through. |
| `dist/src/session-snapshot.js` | 28,017 | −778 | The `foreground`/`customActions` description text moved out of `booleanField('…')` calls in `app.ts`/`snapshot.ts` (this chunk) into the shared descriptor. |
| `dist/src/sdk-contracts.js` | 11,794 | +597 | Runtime counterpart of the same `SNAPSHOT_OPTION_FLAGS` object + its two projection functions, now shared instead of duplicated per seam. |
| `dist/src/registry.js` | 170,798 | +559 | The two options' description text, now declared once in the flag registry instead of only in the command module. |
| `dist/src/client-types.d.ts` | 56,410 | +477 | SDK hover docs: `customActions`'s JSDoc grew from a 69-char hand-written one-liner to the 508-char `inputDescription` verbatim (+439); `foreground`'s reworded JSDoc is 21 chars *shorter*; the rest is `Omit<SnapshotCommandOptionFields, 'customActions'> &` replacing 5 inlined field lines. |
| `dist/src/backend-snapshot-options.js` | 392 | −177 | Same move as `session-snapshot.js`, smaller chunk. |
| `dist/src/snapshot-runtime.js` | 8,558 | −151 | Same move, smaller chunk. |
| **Total (unpacked)** | | **+3,291 B** (reported: +3.3 kB) | |
| **Total (download, gzip)** | | **+2,167 B** (reported: +2.2 kB) | |

**Root cause of the single biggest line item:** `packages/kernel/src/snapshot.ts`'s `SNAPSHOT_OPTION_FLAGS` already carries a comment noting "Exported only because the exported types below say `typeof` it" — the author knew the *value* leaks into the public surface as an unavoidable consequence of TypeScript needing it for the derived types. What wasn't caught is that its 900-byte prose *comment* leaks too, in full, into every consumer's `node_modules/agent-device/dist/src/*.d.ts` — dead weight nobody reads there (a real SDK consumer sees it only as bundled text, not as an attributed source comment).

**Fix implemented** (comment-only, no behaviour change, `pnpm typecheck`/`lint`/`check:affected --run` all green at 79f7161's own numbers — 770 files, 5828 tests): converted that JSDoc block to a plain `//` source comment (kept in full, still readable by the next person reading `snapshot.ts`) plus one short `/** */` line for the SDK's actual hover doc, and removed the `fallow-ignore-next-line unused-export` suppression `fallow` flagged as stale once the export's own doc shrank. Re-measured with the same script:

```
Installed (including dependencies) | 4.56 MB | 4.56 MB | +2.3 kB  (was +3.3 kB)
Package (unpacked)                 | 4.56 MB | 4.56 MB | +2.3 kB  (was +3.3 kB)
Package (download)                 | 1.35 MB | 1.35 MB | +1.8 kB  (was +2.2 kB)
```

**Why the remaining ~2.3 kB / 1.8 kB isn't further cut:**

- The rest of `sdk-contracts.d.ts`'s growth (now ~1.8 kB) is the actual *type shape* — the const map and its derived aliases — not prose. Removing it means not deriving `CaptureSnapshotOptions` from the shared declaration, i.e. reintroducing the 9-site hand-typed duplication #2410 asked to collapse. A smaller design here is available only by giving up the PR's stated purpose, so it's rejected.
- `customActions`'s +439-byte SDK doc growth is the direct, intended cost of "retaining editor docs is fine" from this review: the JSDoc-equality test pins it to the long, tool-facing `inputDescription` verbatim so the two copies can't drift. A smaller design — a short hand-written SDK doc distinct from the long MCP/tool description — was the *pre-PR* state and is exactly the "third, differently-worded copy" #2410 asked to remove; reintroducing it un-pins the two audiences and was rejected for that reason, not overlooked.
- The `registry.js`/session-chunk movement nets out close to zero by construction (text moved, not duplicated) and isn't a design choice to revisit.

No test or fixture file is present in either tarball (`npm pack` file count: 424 on both sides).

thymikee:
The remaining review points are addressed at 9990ea1. The replay counts now distinguish option additions from SDK description edits, and the package growth is itemized and reduced to 2.3 kB without changing runtime behavior. No remaining findings; all checks are green.

Issue #2424, opened 2026-09-09
perf(ios): align AX recovery contracts and evaluate generation-scoped depth hints


thymikee:
## Purpose

Reduce repeated AX rejection work without weakening snapshot completeness, and prevent the host bridge and XCTest runner recovery paths from drifting.

Follow-up to #2414 and #2391; implement after #2414 lands. This is performance/design work, not a release blocker. Unknown frontier completeness must already fail closed in #2414.

## Current data flow and ownership

`open` -> launch observation -> Simulator source -> native capture/recovery. Capture failure opens the snapshot route's generation circuit; #2391 makes subsequent launch observation consult that same owner. Relaunch rebaselines the circuit. Optional coordinate-tap field observation has separate, local failure containment.

The host source defaults to traversal depth 64 (65 native levels), allows two lower-depth retries per acquisition and 32 total requests. It forgets accepted depth between captures. The runner has a 56/40/24/12 ladder, 8 frontier-extension calls and expiring accepted-depth memory. These are different private native representations and currently different failure/completeness policies, not interchangeable implementations.

Crucially, `SnapshotBridgeManager.ensureSession` reuses a process by simulator UDID. The process is not scoped to one app generation. Never use bridge lifetime as a depth-hint validity key.

## Work

First attribute the full path: launch observation currently asks for regular presentation depth1 (not raw traversal depth1), acquires a tree, and discards it before a later capture. Measure resolve/start/connect, launch observation, acquisition/continuation, presentation, and first interaction separately. Do not assume accepted-depth memory is the largest remaining cost. A cheaper readiness contract or request-local handoff requires separate evidence of equivalent readiness/freshness; do not quietly reinterpret presentation depth or add a cross-command stale-tree cache in this issue.

1. Add shared, executable recovery conformance fixtures consumed through each producer's adapter. Cover zero-based requested depth versus native levels, node budgets, complete/withheld/unknown frontier evidence, missing live elements, owner change, request exhaustion, deadline/cancellation, and explicit raw-depth requests. Document intentional backend differences. Start with shared contracts and fixtures; extract common executable policy only where equivalence is proven. Do not combine private object decoding with recovery or wire serialization.
[...]

PR #2425, opened 2026-09-09
test(ios): add a shared AX recovery conformance fixture with a host adapter

thymikee:
## Summary

Adds `contracts/fixtures/ios-ax-recovery-conformance.json`, a shared executable recovery contract for the host AX bridge and the XCTest runner's private AX bridge. Fifteen recovery cases (569 fixture lines; hint cases arrive with the runner adapter in #2428) cover zero-based depth versus native levels, node budgets, complete/withheld/unknown frontier evidence, missing live elements, owner change, request exhaustion, deadline, explicit raw depth, hinted first requests, and multi-branch recovery. Every expectation names the outcome, the native request accounting, the deepest level, and the delivered tree as a canonical preorder signature with its retained node count, so a producer that drops, duplicates, reorders, or re-parents nodes cannot pass as complete. Expectations are per producer, and the fixture documents the intentional differences instead of averaging them away.

This PR carries the host adapter: an Objective-C driver over `captureSnapshotTree`, compiled and run per case by `native-runtime.test.ts`. The runner adapter and the accepted-depth memory characterization follow in #2428, which executes the fixture's `runner` column and adds the hint cases; README and ADR 0004 mark the runner adapter as pending until then.

First step of #2424, stacked on #2414; no runtime behaviour change. Five files, 903 lines (569 fixture).

## Validation

Tested head: `d9a3c2afa9`.

- `pnpm check:affected --run --base origin/fix/ios-ax-depth-recovery` passed.
- Native: 46 compiled cases (31 existing + 15 fixture cases) pass via `native-runtime.test.ts`.
- Planted red in the host materializer, each restored afterwards: keep only the first sibling → `multi-branch-recovery`, `runner-extension-budget` red; duplicate the last sibling → 8 cases red (signature and node count); reverse siblings → 2 red. Also: missing boundary child count accepted → 7 red; request budget 33 → 3 red; native levels off by one → 8 red.
- Review follow-ups: structural signatures and the multi-branch case added (P1); hint characterization moved to #2428 (P2/scope), diff reduced from 1,551 to 903 lines; docs mark the runner adapter as pending; the fixture no longer restates code constants (ladder, budgets, retries) that the cases pin behaviourally.

### DIFF
--- a/apple/snapshot-bridge/README.md
+++ b/apple/snapshot-bridge/README.md
@@ -63,3 +63,19 @@ use non-launch failure codes, so the route falls back without launch re-polling.
 At each native fragment boundary, an absent or invalid child count means unknown
 completeness and fails closed. Natural leaves above that boundary need no
 continuation evidence. Unchanged native dictionaries and child arrays are reused.
+
+## Recovery conformance
+
+`contracts/fixtures/ios-ax-recovery-conformance.json` is the shared, executable
+recovery contract for this bridge and the XCTest runner's private AX bridge.
+`packages/platform-apple/src/snapshot-source/fixtures/recovery-conformance.m`
+replays each case through `captureSnapshotTree`; the runner adapter that
+replays the same cases through the XCTest runner's private AX bridge is
+pending in [#2428](https://github.com/callstack/agent-device/pull/2428), so at
+this revision only the fixture's `host-bridge` column is executed. Each
+expectation names the
+outcome, the native request accounting, and the delivered tree as a canonical
+preorder signature with its retained node count, so a producer that drops,
+duplicates, reorders, or re-parents nodes cannot pass as complete. The fixture
+records each producer's expectation and documents the intentional differences
+(depth vocabulary, ladders, completeness evidence, budgets, hint lifetime).
--- a/contracts/fixtures/ios-ax-recovery-conformance.json
+++ b/contracts/fixtures/ios-ax-recovery-conformance.json
@@ -1,0 +1,567 @@
+{
+  "version": 1,
+  "description": "Shared, executable recovery contract for the two iOS Simulator accessibility producers: the host AX bridge (apple/snapshot-bridge) and the XCTest runner's private AX bridge (apple/runner). Each producer runs every case through its own adapter and interprets the synthetic native world in its own model; expectations are recorded per producer so intentional differences stay visible instead of being averaged away.",
+  "nativeModel": {
+    "tree": "A chain of `chain` nodes at levels 0..chain-1 (labels are the level number). An optional `fan` gives the last chain node `count` children at level `at + 1`, each heading a further chain of `chain` nodes; fan labels are `<level>.<branch>`.",
+    "request": "A native request rooted at level r for L node levels returns levels r..r+L-1 that exist. `rejectLevelsAbove` rejects any request asking for more node levels than that with the producer's depth-rejection code.",
+    "frontierEvidence": "`counted` gives every returned node its true child count (host only; the runner has no counts). `unknown` omits the count on the deepest level of every returned fragment.",
+    "vanishAtFrontier": "The deepest level of every returned fragment has no live accessibility element, so no continuation can be rooted there.",
+    "ownerChangesAfterRequests": "The primary foreground owner changes after that many native requests (host only).",
+    "deadlineAfterRequests": "The capture deadline is spent after that many native requests (runner only; the host deadline is the guest watchdog and the host request timeout).",
+    "signature": "`tree` is the delivered tree in preorder: every node's identity, with `<parent` appended when its parent is not the canonical one (the previous level on the same branch, or the fan's last chain node for a branch head); runs of consecutive levels on one branch collapse to `first-last` (`0-44`, `30.1-49.1`). `nodes` is the retained node count. Dropped, duplicated, reordered, or re-parented nodes change the signature."
+  },
+  "producers": {
+    "host-bridge": {
+      "recoveryOwner": "apple/snapshot-bridge/SnapshotBridgeCapture.m",
+      "hintOwner": "not yet: the host learns no accepted-depth hint (follow-up to #2424 step 1)",
+      "adapters": ["packages/platform-apple/src/snapshot-source/fixtures/recovery-conformance.m"],
+      "nativeLevelsForTraversalDepth": "depth + 1 (delivers exactly `depth` edges below the root)",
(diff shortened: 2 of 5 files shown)


thymikee:
Reviewed **509bc1ca6e7e0351d511b497d2bdc57e2adab672** against #2424 and both capture producers. The shared fixture is the right direction, and I found no runtime regression in the production extraction. I would address these before merging:

### [P1] Assert the recovered tree, not only its maximum depth

The [host assertions](https://github.com/callstack/agent-device/blob/509bc1ca6e7e0351d511b497d2bdc57e2adab672/packages/platform-apple/src/snapshot-source/fixtures/recovery-conformance.m#L174-L176) and [runner assertions](https://github.com/callstack/agent-device/blob/509bc1ca6e7e0351d511b497d2bdc57e2adab672/apple/runner/AgentDeviceRunner/AgentDeviceRunnerUITests/UnitTests/RunnerTests%2BAXRecoveryConformanceTests.swift#L288-L293) check outcome, counts and deepest level, but not which nodes survived or their structure.

**Verified counterexample:** in a temporary copy of the host materializer, I discarded every sibling except the first after recovery. Every applicable shared host case still passed (11 executed; 3 marked not applicable). A capture can therefore lose branches and still satisfy this conformance suite as “complete.”

Please assert a compact canonical tree signature covering ordered node identities and parent relationships, plus retained node count. Include successful multi-branch recovery and the expected retained tree for bounded captures, with producer-specific expectations where necessary. Prove dropped, duplicated and reordered siblings fail. This is central to #2424's completeness contract.

### [P2] Remove or exercise the unused hint-outcome fields

The [hint loop](https://github.com/callstack/agent-device/blob/509bc1ca6e7e0351d511b497d2bdc57e2adab672/apple/runner/AgentDeviceRunner/AgentDeviceRunnerUITests/UnitTests/RunnerTests%2BAXRecoveryConformanceTests.swift#L327-L333) reads `acceptedLevels` and directly invokes the memory helper. Its decoded `outcome.complete` and `outcome.rejected` fields are never consumed. Consequently, the complete/bounded learning cases never exercise those capture outcomes; changing either field cannot change the result.

Either drive the real recovery-to-learning transition with these inputs and assert its observations, or delete the unused fields and narrow the case names/claims to direct memory-state tests. Preserve the runner's current behavior of learning a lower accepted rung even from bounded recovery; this characterization PR need not change that policy.

### Scope cleanup

The actual diff is **1,551 gross lines** (1,512 additions + 39 deletions), above the repository's 1,000-line PR budget. Please reduce unnecessary fixture/schema scaffolding or split the memory characterization into a separate cohesive PR. Keep meaningful recovery coverage rather than compressing JSON just to lower the count.

Validation: compiled and ran the standalone host fixture, then repeated it with the sibling-loss mutation. I did not run the native Swift suite locally. No repository source files were changed.

thymikee:
Re-reviewed `2ac55a1b369cdbb0e5456e51e0bfb64c54b396d5`: **the code-level findings are addressed for this host-only slice.**

- Independently compiled and ran all 12 applicable host cases: green (the other 3 cases are runner-only).
- Repeated the sibling-loss mutation, plus sibling duplication and reversal: each now fails `multi-branch-recovery` and `runner-extension-budget`.
- The misleading hint tests/fields and runner production extraction are removed from this PR. The runner follow-up needs its own review; this does not yet complete two-producer conformance.
- Actual diff: 5 files, 904 additions, within the repository budget. No new code blocker found.

One small cleanup from the split: `apple/snapshot-bridge/README.md` says the runner already replays the cases in `RunnerTests+AXRecoveryConformanceTests.swift`, and ADR 0004 says each producer replays every case. That adapter is absent at this head. Mark runner execution as pending the stacked follow-up and link it when available. The PR body also still mentions hint cases and a 720-line fixture; this head has no `hintCases` and the fixture is 571 lines.

CI on this exact head is still running, with no reported failures so far. Code looks ready once CI is green; the above is documentation cleanup, not a new runtime blocker.

thymikee:
The host-only correction at 2ac55a1 addresses the tree-completeness findings; the smaller scope also reduces packaged growth to 831 B. Please update the README and ADR to mark the runner adapter as pending, as noted above. Merge #2414 first; this slice does not yet complete two-producer conformance. CI on the new head is still running.

thymikee:
Addressed on `2ac55a1b36`; the stack is now three PRs.

**P1 (tree structure).** Every recovery expectation now carries `tree`, a canonical preorder signature (identities, `<parent` for a non-canonical parent, `first-last` runs per branch), and `nodes`, the retained count; `nativeModel.signature` documents the format. Added `multi-branch-recovery` (three branches recovered through three continuations / three extension calls) and the bounded trees of `node-budget`, `deeper-than-requested`, `explicit-depth-honored`, and `runner-extension-budget` (eight of twelve frontiers extended) are spelled out per producer. Planted in the host materializer and restored: keep only the first sibling → 2 cases red; duplicate the last sibling → 8 red; reverse siblings → 2 red. Planted in the runner serializer: reverse children → 3 cases red (`multi-branch-recovery`, `host-request-budget`, `runner-extension-budget`). The signature walk is bounded by the bridge's node limit, so a shared or cyclic subtree reads as an oversized signature rather than a hang (the duplicate mutation found that).

**P2 (unused hint fields).** The runner memory test now drives the real transition: each hint step builds a chain that fits the accepted rung (`complete: true`) or a longer chain under a node budget of that rung (`complete: false`), runs the real ladder against a client that rejects above the accepted rung, asserts the observed rejections (`rejected.runner`), the accepted rung, and boundedness, and only then hands `effectiveDepth` to the runner's own recording step. `rejected` became per-producer because the ladders differ. The runner's policy of learning a lower rung from a bounded recovery is preserved and asserted.

**Scope.** Split by producer: this PR keeps the fixture and the host adapter (904 lines, 720 of them fixture); #2428 carries the runner seam, the Swift adapter, and the memory characterization; #2427 rebases on top and adds the host hint column (its hint cases now run through the source adapter rather than the unit).

thymikee:
The d9a3c2a cleanup resolves the documentation scope gap without weakening the behavioral fixture assertions. This host-only slice looks clean; merge #2414 first. #2428 supplies the runner adapter later. CI is still running.

thymikee:
Docs and body reconciled at `d9a3c2afa9`: README and ADR 0004 now say the runner adapter is pending in #2428 (which restores the present-tense wording when it lands), and the body reflects 15 recovery cases, 569 fixture lines, no hint cases. The fixture also stopped restating code constants (ladder, budgets, retry count) that the cases pin behaviourally. The failed iOS smoke job was re-run; the diff has no runtime change.

PR #2428, opened 2026-09-09
test(ios): run the shared AX recovery fixture through the runner and characterize its depth memory

thymikee:
## Summary

Runs `contracts/fixtures/ios-ax-recovery-conformance.json` (#2425) through the XCTest runner's private AX bridge. A Swift unit test replays every recovery case through a fake AX client against the real bridge capture, depth ladder, frontier extension, and completeness verdict, asserting outcome, request accounting, deepest level, and the delivered tree's canonical signature and node count. To make that possible `RunnerAXSnapshotBridge` gains a resolved client/target capture seam and the ladder loop becomes a platform-neutral function; behaviour is unchanged.

The fixture gains its eight hint cases, executed here as the runner's accepted-depth memory characterization: each step is a real ladder capture (a chain that fits the accepted rung, a longer chain under a node budget of that rung when the step is bounded, or a world that rejects every depth when the step fails) against a client that rejects above the accepted rung; the observed rejections, accepted rung, and boundedness are asserted, and the runner's own learning step runs only after a successful ladder, exactly as in production. The runner keeps learning a lower rung from bounded recoveries and keeps its memory untouched across a failed capture.

Second step of #2424, stacked on #2425; restores the README/ADR wording that the runner adapter exists. Five files, 970 lines.

## Validation

Tested head: `0f44265a58`.

- `pnpm check:affected --run --base <#2425 head>`: every check passed except three `packages/host-kit/src/code-signature.test.ts` cases, which fail identically on a clean `origin/main` checkout on this Mac (the fixture root is not realpath'd while the walker realpaths manifests under the `/tmp → /private/tmp` symlink); unrelated to this diff, CI's Linux lane is authoritative. `pnpm check:xctest-selection` passes (both tests reach the host lane).
- `testPrivateAXRecoveryMatchesSharedFixture` and `testPrivateAXAcceptedDepthMemoryMatchesSharedFixture` pass on an iOS 26.2 simulator through `xcodebuild test-without-building` on the unit-test build; the macOS host lane is CI-verified (it cannot run on this machine).
- Planted red in the runner, each restored: serialize children in reverse order → `multi-branch-recovery`, `host-request-budget`, `runner-extension-budget` red; stop counting a vanished frontier element as missed → `missing-live-element` red. The adapter also caught one mis-simulated expectation on its first run (`host-request-budget`: four rejected rungs before the 12-level rung).

### DIFF
--- a/apple/runner/AgentDeviceRunner/AgentDeviceRunnerUITests/RunnerAXSnapshotBridge.h
+++ b/apple/runner/AgentDeviceRunner/AgentDeviceRunnerUITests/RunnerAXSnapshotBridge.h
@@ -49,6 +49,18 @@ FOUNDATION_EXPORT NSString *const RunnerAXSnapshotCustomActionsBlockedKey;
                                            customActionLimit:(NSInteger)customActionLimit
                                                     deadline:(nullable NSDate *)deadline;
 
+/// The capture behind `snapshotTreeForApplication:`, for an AX client and target
+/// application element that are already resolved. Every request, extension, and
+/// custom-action decision happens here; the application entry point only resolves
+/// the two identities.
++ (NSDictionary<NSString *, id> *)snapshotTreeWithClient:(id)axClient
+                                                  target:(id)target
+                                                maxDepth:(NSInteger)maxDepth
+                                                maxNodes:(NSInteger)maxNodes
+                                  deepExtensionCallLimit:(NSInteger)deepExtensionCallLimit
+                                       customActionLimit:(NSInteger)customActionLimit
+                                                deadline:(nullable NSDate *)deadline;
+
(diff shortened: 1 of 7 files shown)


thymikee:
The production extraction at `bf0be60` preserves the existing ladder and gives the shared fixture a real runner capture path; no new runtime regression found. The runner memory case characterizes existing bounded-recovery learning rather than changing that policy.

Resolve the conflict with the updated #2425 and rerun the fixture tests before merging. Merge #2414, then #2425 first. iOS smoke is still running.

thymikee:
The conflict resolution at 0f44265 leaves the production capture path unchanged. The shared runner fixture now covers failed recovery without teaching a new depth; no remaining code finding. Merge #2414 and #2425 first. iOS smoke and coverage are still running.

PR #2429, opened 2026-09-09
fix(host-kit): resolve the code-signature root so a symlinked checkout stamps its own files

thymikee:
## Summary

`walkDaemonCodeGraph` resolved a workspace manifest through `realpathSync.native` but took the root and the entry as given. Under a symlinked prefix the two then sat on either side of the link, and two things went wrong at once:

- Every workspace package was labelled by the route out of the repository and back in (`../../../private/tmp/.../packages/kit/src/owned.ts`) instead of `packages/kit/src/owned.ts`.
- `isInstalledDependencyPath` tests `node_modules` containment, so it read **every installed dependency as a workspace package** and walked its whole closure.

macOS hits this on every run: `os.tmpdir()` is `/tmp`, a link to `/private/tmp`. A real checkout hits it under any symlinked parent directory.

The fix routes the root, the entry, and the manifest through one `resolveDaemonCodePath`, and hands the cache the same resolved pair so a stored document's entry label still matches the walk that wrote it. The cache's private copy of that resolver is deleted.

The fixtures now build a resolved root, which is the shape a production caller passes — `findProjectRoot` derives it from an already-resolved `import.meta.url`. Naming a root through a link is covered on purpose instead, by two new tests that fail without this change on any platform.

5 files, all in `packages/host-kit/src`.

## Validation

Tested at `e0cab5dd26`, on macOS 25.6.0 (darwin).

- `npx vitest run packages/host-kit/src/code-signature.test.ts packages/host-kit/src/code-signature-cache.test.ts` — 29 passed. The three cases named in the report (`a workspace subpath is walked...`, `retargeting the exports map...`, `an installed dependency is not followed into`) failed on macOS before this change and pass now.
- `pnpm check:affected --run` — all runnable checks passed (124 test files, 826 tests, plus typecheck, layering, fallow, build).
- Reverting only the two `resolveDaemonCodePath` calls in the walker fails exactly the two new symlink tests, so neither is vacuous.

Not device-facing: the code signature is computed from the filesystem, and no device lane exercises a path this change alters.

### DIFF
--- a/packages/host-kit/src/code-signature-cache.test.ts
+++ b/packages/host-kit/src/code-signature-cache.test.ts
@@ -5,8 +5,7 @@ import path from 'node:path';
 import { afterEach, test, vi } from 'vitest';
 import { computeDaemonCodeSignature } from './code-signature.ts';
 import { resolveCachedDaemonCodeSignature } from './code-signature-cache.ts';
-import { writeWorkspaceFixture } from './code-signature.fixtures.ts';
-import { mkdtempForTestSync } from './internal/tmp-dir.fixtures.ts';
+import { createCheckoutRootForTest, writeWorkspaceFixture } from './code-signature.fixtures.ts';
 
 afterEach(() => {
   vi.restoreAllMocks();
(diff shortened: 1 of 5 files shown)


thymikee:
The symlink fix at e0cab5d consistently resolves the root, entry and manifest, and the new regressions cover both workspace labels and installed-dependency exclusion. No findings; all checks are green.

PR #2431, opened 2026-09-09
fix(ios): serve regular --depth from every snapshot backend

### LINKED ISSUE #2403: 0.21.0: --depth on the iOS Simulator returns COMMAND_FAILED "regular iOS snapshot presentation requires a valid viewport"
On 0.21.0, `is <selector> visible --depth <n>` returns `COMMAND_FAILED` with the message
`regular iOS snapshot presentation requires a valid viewport` on every screen we tried. The same
command on 0.20.10 answered normally on some of those screens.

Found while confirming the fix for #2273, which is confirmed and working. Filing separately because
this is a different path.

## Version

`agent-device` 0.21.0 on an `ubuntu-latest` GitHub Actions runner. The daemon is a remote EAS
Simulator session and also reports 0.21.0. The target is an iOS simulator. The app is a React Native
app built with Expo.

## What we ran

Ten `is <selector> visible` calls per variant per screen, on six screens. The variants are unscoped,
`--scope <identifier>`, and `--depth 20`.

The depth argument is a plain integer and our invocation is unchanged from the 0.20.10 measurement.
The screen numbering matches #2273 and #2274.

## `--depth` on 0.21.0 against 0.20.10

| screen   | 0.20.10 depth    | 0.21.0 depth | 0.21.0 error                                                |
| -------- | ---------------- | ------------ | ----------------------------------------------------------- |
| screen 1 | 0 of 10, correct | 0 of 10      | regular iOS snapshot presentation requires a valid viewport |
| screen 2 | 10 of 10         | 0 of 10      | regular iOS snapshot presentation requires a valid viewport |
| screen 3 | 10 of 10         | 0 of 10      | regular iOS snapshot presentation requires a valid viewport |
| screen 5 | 0 of 10          | 0 of 10      | regular iOS snapshot presentation requires a valid viewport |
| screen 6 | 0 of 10          | 0 of 10      | regular iOS snapshot presentation requires a valid viewport |
| screen 7 | 0 of 10          | 0 of 10      | regular iOS snapshot presentation requires a valid viewport |

All 60 calls returned `COMMAND_FAILED` with that one message. The message does not appear anywhere
in our 0.20.10 logs.

Two rows are the clearest evidence. Screens 2 and 3 answered 10 of 10 on 0.20.10 and answer 0 of 10
now.

Screen 1 is worth naming too. On 0.20.10 its depth-capped calls returned `COMMAND_FAILED` with
`Selector did not match`, which is the correct answer, because its landmark sits deeper than a cap of 20. That correct negative is now replaced by the viewport error, so no screen gets a real answer from
`--depth` any more.

## The unscoped path is fixed

For completeness, since it is the same command without the flag. The three screens that answered 0
of 30 on 0.20.10 now answer 10 of 10 unscoped, and `XCTEST_RECORDED_FAILURE` is gone from the whole
run. Median latency fell from 1.2 to 5.2 seconds down to 0.5 to 1.1 seconds. That part is #2273 and
it works.

## An observation about `--scope`, which we are not calling a defect

We do not have enough to claim a regression here, and the error code is a correct-negative code
rather than an infrastructure one, so this may be intended.

| screen   | 0.20.10 scoped | 0.21.0 scoped |
| -------- | -------------- | ------------- |
| screen 1 | 10 of 10       | 10 of 10      |
| screen 2 | 10 of 10       | 0 of 10       |
| screen 3 | 10 of 10       | 0 of 10       |
| screen 5 | 0 of 10        | 0 of 10       |
| screen 6 | 0 of 10        | 10 of 10      |
| screen 7 | 0 of 10        | 0 of 10       |

Every 0.21.0 failure above is `COMMAND_FAILED` with `Selector did not match`. We scope to the same
identifier we then query for, so a scoped query for an element that the unscoped query finds should
match. If scope resolution changed deliberately, this is us needing to change how we call it, and a
note to that effect would be enough.

## One correction to what we told you in #2273

We said there that a truncated capture reports a node count which is a lower bound, and reasoned that
the affected screens were therefore probably the largest trees.

On 0.21.0 screen 5 reports the same 178 nodes it reported when it was truncated, and it is no longer
truncated. So that count was complete all along and our inference was wrong. Recording it here
because we put the claim in front of you.

thymikee:
## Summary

Closes #2403.

On 0.21.0 every `is <selector> visible --depth <n>` on the reporter's iOS Simulator failed with `regular iOS snapshot presentation requires a valid viewport`. The cause is in the runner, not the daemon: `SnapshotBackendKind.canServeRegularPresentedDepth` (from #1947) refused a regular depth-capped request on private AX at any depth and on the query sweep past depth 1. Whenever the plan is pinned or deferred to private AX (custom actions, a private AX verdict already on the session, the XCTest channel penalty), no backend could serve, the plan fell through to its synthetic sparse root (one `Application` node, zero rect), and the daemon rejected that root as a missing viewport.

Presentation already applies the presented-depth cut to whatever hierarchy a backend acquired, and a depth-capped regular capture is a subset of the unscoped one from the same backend, so the refusal protected nothing that `truncated`/`effectiveDepth` did not already disclose. This PR deletes the gate, declares private AX as `regular-depth=presentation-cut` (fixture, TS registry, Swift enum, all parity-tested), rewrites the ADR 0004 paragraph that made the old claim, and adds a presentation test for the cut on a hierarchy acquired past the frontier.

```
agent-device snapshot --depth 20 --actions   # private AX pinned: 58 nodes, was COMMAND_FAILED
agent-device is visible "role=button label=General" --depth 20
```

8 files touched. Not changed here: a genuine sparse runner verdict (no backend could read the screen) still surfaces as the viewport-invariant error instead of the runner's own "use screenshot" message; that is a separate daemon-side seam.

## Validation

Tested commit: `bb3dfdc8a2b8627d5f16adb423c4a18ddef1af50`.

- Repro before the fix on a local iOS 26.2 Simulator (Settings): `snapshot --depth 20 --actions` returned the runner's sparse payload (`state: sparse`, one zero-rect `Application` node) and the daemon failed with the reported message; plain `--actions` worked.
- After the fix, same device: `snapshot --depth 20 --actions` returns 58 nodes; `--depth 1 --actions` returns 4. Through the test-owned `snapshotPreferredBackend: 'private-ax'` transport seam: `snapshot --depth 20` is a 58-node `recovered` private AX capture, `is visible role=button label=General` passes at depth 20 and 3, and depth 0 returns the correct `Selector did not match`.
- New runner-path regression `testPrivateAXPinnedRegularDepthReachesAcquisitionAndPresentation` (PR iOS lane): green at 1db684c239 in 4.9 s; red with the gate restored (5 assertions, runner logs `SNAPSHOT_BACKEND_DEPTH_UNSUPPORTED backend=private-ax depth=1`). Evidence in the review reply.
- Runner unit tests (`AGENT_DEVICE_XCUITEST_INCLUDE_UNIT_TESTS=1`, iOS 26.2 sim): `testSnapshotBackendDeclarationsMatchCapabilityFixture`, `testSparsePayloadReasonMatrix`, `testLegacyQualityMessageStatesFallbackMeaning` pass (3 executed).
- `swift test --package-path apple/snapshot-presentation`: 3 tests pass, including the new `RegularDepthTests`.
- `pnpm check:affected --run`: green on bb3dfdc8a2 (387 files, 2827 tests). On 1db684c239 the vitest-related stage hit 7 contention timeouts on a Mac at load 125; those five files pass in isolation (68/68) and the delta is Swift test code plus one workflow line; `check:xctest-selection` passes.

Risk: the physical-device plan now also lets the query sweep answer a `--depth ≥ 2` request with its flat sweep (same payload it already returned unscoped) instead of no answer.

### DIFF
--- a/.github/workflows/ios.yml
+++ b/.github/workflows/ios.yml
@@ -241,7 +241,8 @@ jobs:
             -only-testing:AgentDeviceRunnerUITests/RunnerTests/testCustomActionCoverageParsesOnlyCompletePairs \
             -only-testing:AgentDeviceRunnerUITests/RunnerTests/testPartialCustomActionPassIsDisclosedAndCompleteOneIsNot \
             -only-testing:AgentDeviceRunnerUITests/RunnerTests/testActionNamesAreCappedPerElementAndReported \
-            -only-testing:AgentDeviceRunnerUITests/RunnerTests/testHungCustomActionReadIsContainedAndRecovers 2>&1 | tee /tmp/agent-device-runner-regressions.log
+            -only-testing:AgentDeviceRunnerUITests/RunnerTests/testHungCustomActionReadIsContainedAndRecovers \
+            -only-testing:AgentDeviceRunnerUITests/RunnerTests/testPrivateAXPinnedRegularDepthReachesAcquisitionAndPresentation 2>&1 | tee /tmp/agent-device-runner-regressions.log
           node --input-type=module -e '
             import { readFileSync } from "node:fs";
             const log = readFileSync("/tmp/agent-device-runner-regressions.log", "utf8");
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,15 @@
 
 ## Unreleased
 
+- Fixed: iOS `--depth` on `snapshot`, `is`, `wait`, `get`, and `find` no longer fails with
+  `regular iOS snapshot presentation requires a valid viewport` when the runner plan is pinned or
+  deferred to the private AX backend (custom actions, a private AX verdict on the session, or the
+  XCTest channel penalty). The runner refused a regular depth-capped request on every backend but
+  the recursive tree, fell through to its synthetic sparse root, and the daemon rejected that root
+  as a missing viewport. Presentation applies the presented-depth cut to whatever hierarchy a
+  backend acquired, so every backend serves the request; the private AX declaration is now
+  `regular-depth=presentation-cut` and an acquisition that stopped short of the cut keeps
+  disclosing that through `truncated`/`effectiveDepth` as it does unscoped.
 - Added: `replay export` supports flows that switch apps and return, preserving each
   `open <appId>` target as an explicit Maestro `launchApp.appId`.
 - Added: `replay export` converts recorded `home` actions to Maestro `pressKey: Home`, allowing
--- a/apple/runner/AgentDeviceRunner/AgentDeviceRunnerUITests/RunnerTests+SnapshotBackendCapabilities.swift
+++ b/apple/runner/AgentDeviceRunner/AgentDeviceRunnerUITests/RunnerTests+SnapshotBackendCapabilities.swift
@@ -5,14 +5,19 @@ enum SnapshotBackendEnvironment {
   case physicalDevice
 }
 
+/// How a backend's acquisition relates to a regular `--depth` request. Every backend serves the
+/// request: `SnapshotPresentation` applies the presented-depth cut to whatever hierarchy was
+/// acquired, and an acquisition that stopped short of the cut discloses that through its own
+/// truncation verdict. The capability only says how much acquisition work the request bounds.
 enum SnapshotRegularDepthCapability: String {
-  /// The backend can stop acquisition at the requested regular presented-depth frontier.
+  /// Acquisition stops at the requested regular presented-depth frontier.
   case presentedFrontier = "presented-frontier"
-  /// The backend is flat; it can answer the root and one presented level, but has no hierarchy
-  /// from which to prove deeper regular depth.
+  /// The backend is flat: it acquires the root and one presented level, so a cut past depth 1
+  /// returns the sweep unchanged.
   case flat
-  /// The backend can return raw traversal depth, but cannot prove regular presented depth.
-  case rawOnly = "raw-only"
+  /// Acquisition walks its raw-depth ladder regardless of the request; the presented cut happens
+  /// in presentation and the ladder's cap is disclosed as `effectiveDepth`.
+  case presentationCut = "presentation-cut"
 }
 
 enum SnapshotBackendKind: String, CaseIterable {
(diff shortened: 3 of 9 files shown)


thymikee:
The fix looks sound at bb3dfdc8, and the reported simulator reproduction plus CI are green. One regression test is still needed: RegularDepthTests calls presentation directly, so it passes even with the old runner backend refusal restored. Add a runner-path test for private AX pinned with regular --depth that reaches acquisition and presentation rather than the sparse fallback, and show that restoring the deleted gate makes it fail.

thymikee:
Added in 1db684c239: `testPrivateAXPinnedRegularDepthReachesAcquisitionAndPresentation` (runner bundle, `os(iOS)`, registered in the ios.yml PR lane; `check:xctest-selection` now reads 237 declared / PR list 89 / 0 unreachable).

It launches the host app with `--agent-device-selector-read-regression`, pins `preferredBackend: private-ax`, runs `runSnapshotCapturePlan(regularVisiblePlan, depth: 1)`, and requires a private-ax verdict that is not `sparse`, more than one node, max presented depth 1, a non-zero root rect, the `Readable target` button, and a payload no larger than the unscoped capture from the same backend.

**Green at 1db684c239** (iOS 26.2 Simulator, `AGENT_DEVICE_XCUITEST_INCLUDE_UNIT_TESTS=1`): passed in 4.9 s; runner log shows `PRIVATE_AX_SNAPSHOT_USED nodes=9 depth=64` followed by `SNAPSHOT_RECOVERED backend=private-ax` for both captures.

**Red with the gate restored** (`git apply -R` of the bb3dfdc8a2 gate hunks, rebuilt, same test): failed in 4.0 s on five assertions — runner log `SNAPSHOT_BACKEND_DEPTH_UNSUPPORTED backend=private-ax depth=1`, then

```
XCTAssertNotEqual failed: ("sparse") is equal to ("sparse")
XCTAssertGreaterThan failed: ("1") is not greater than ("1")
XCTAssertEqual failed: ("Optional(0)") is not equal to ("Optional(1)")   // max depth
XCTAssertNotEqual failed: ("SnapshotRect(x: 0.0, y: 0.0, width: 0.0, height: 0.0)") is equal to (...)
XCTAssertTrue failed                                                       // "Readable target" absent
```

i.e. exactly the synthetic sparse root the daemon then rejects as a missing viewport.

`pnpm check:affected --run` on 1db684c239: the vitest-related stage hit 7 timeouts (`/bin/ps timed out after 1000ms`, `Test timed out in 5000ms`) while this Mac was at load 125 with two other worktrees building; the same five files pass in isolation (68/68), and the full gate was green on bb3dfdc8a2, whose only delta is this Swift test plus the workflow line.

thymikee:
The runner-path regression at 1db684c addresses the remaining review gap. It exercises private AX acquisition and presentation, and the reported simulator red/green run shows that restoring the old gate produces the sparse-root failure. No remaining code findings; ready for human review while CI finishes.

PR #2440, opened 2026-09-10
fix(wait): retire native selector bypass and recover text observations

### LINKED ISSUE #2404: 0.21.0: a wait step inside replay still fails with querySelector, while the is command answers the same selector
On 0.21.0, a `wait <selector>` step inside `agent-device replay` still fails with
`XCTest recorded a failure while executing querySelector`, while the standalone
`agent-device is <selector> visible` answers the same selector ten times out of ten in the same
session.

This is the read path from #2273, which is fixed for the `is` command and still broken inside
`replay`. Filing separately because it looks like a different call site rather than a regression of
that fix.

## Version

`agent-device` 0.21.0 on an `ubuntu-latest` GitHub Actions runner. The daemon is a remote EAS
Simulator session and also reports 0.21.0. The target is an iOS simulator. The app is a React Native
app built with Expo.

## One session, two ways of asking

A single run drove the app to a menu screen, then ran two steps back to back against the same screens.

The probe step called `is 'id="..."' visible` ten times per screen. The replay step ran five `.ad`
flows whose steps are only `wait id="..."` and `press id="..."`, with no `--scope` and no `--depth`.

| identifier        | `is <id> visible`, 10 calls | `wait <id>` inside `replay`                                     |
| ----------------- | --------------------------- | --------------------------------------------------------------- |
| `id="LANDMARK_A"` | 10 of 10, median 377ms      | failed, XCTest recorded a failure while executing querySelector |
| `id="LANDMARK_B"` | 10 of 10, median 835ms      | failed, XCTest recorded a failure while executing querySelector |

`LANDMARK_A` is the menu container the flows start from. `LANDMARK_B` is the arrival landmark on one
of the screens the flow opens.

The full replay errors:

```
Error (REPLAY_DIVERGENCE): Replay failed at step 3 (wait "id=\"LANDMARK_B\"" 15000):
XCTest recorded a failure while executing querySelector; the action may not have been performed.

Error (REPLAY_DIVERGENCE): Replay failed at step 1 (wait "id=\"LANDMARK_A\"" 10000):
XCTest recorded a failure while executing querySelector; the action may not have been performed.
```

## Why we think this is a separate call site

`LANDMARK_A` is on a screen that never had the original problem. Its capture does not truncate and
does not fall back to the private-AX backend, and it answered 10 of 10 on 0.20.10 as well. So this is
not the degraded-tree condition from #2273 reappearing. A `wait` inside `replay` fails there while the
same selector answers immediately outside it.

Per #2273 the `get`, `is` and `wait` paths no longer use the direct `querySelector` shortcut. The
message above says `replay` still reaches it, so we suspect `replay` resolves its `wait` steps through
a different path than the `wait` command.

## What this blocks

`wait <selector>` is the only assertion the `.ad` format has. So a flow cannot check that it arrived
anywhere, which makes `replay` unusable for us on this app even though every selector it uses answers
when queried directly.

We are otherwise ready to use `.ad` flows as the portable unit of our device checks, and this is the
one thing in the way.

## Not the cascade

Three of our five flows also failed at step 1 with a plain `wait timed out`, and that part is ours,
not yours. The first flow failed partway and left the app on the wrong screen, so the later flows'
opening `wait` had nothing to match. We mention it so the log excerpt above is not read as five
independent failures.

thymikee:
## Summary

Selector waits could enter XCTest `querySelector` before canonical capture and fail while `is` succeeded. Replay inherited the normal wait route.

Delete the unused `findSelector` runtime operation end to end: contracts, owner facts, binding plumbing, dispatch, and unreachable tests. Selector waits now use canonical capture across platforms. The runner's interaction-facing selector command remains intact.

Native text-observation failures now defer to capture at the shared wait binding; cancellation and canonical capture failures remain visible. Correct `findText` runner-demand classification without adding conditional observations to required execution plans.

33 files changed, primarily removal of dead plumbing. Closes #2404.

## Validation

- Head `985701a340`: `pnpm check:affected --run` passed, including 3,753 related tests, 12 documentation tests, format, lint, typecheck, layering, DI, Fallow, and build.
- Standalone and replay regressions proved red before their fixes for selector and native-text failures. Tests also preserve cancellation, capture errors, fresh polling, and observation-only startup.
- Live iOS 26.2 simulator, Settings: selector and text waits passed standalone and together in a two-step replay on this head. Session closed; isolated daemon and runner cleaned.
- Reporter’s EAS app unavailable locally. GitHub CI pending on this head.


packages/platform-apple/src/runtime.ts
```diff
@@ -303,7 +299,7 @@ export function createApplePlatformRuntime(host: PlatformRuntimeHost): PlatformR
         ...screenshotRuntimeOperationFacts({ capture: appleScreenshotFact(device) }),
         ...selectorObservationRuntimeOperationFacts({
           findText: appleSnapshotFact(device),
-          findSelector: appleFindSelectorFact(device),
+          findSelector: nativeSelectorUnavailable,
```
thymikee: This is the last producer of `findSelector` in the repo. Android, HarmonyOS, web and Limrun already declare it unavailable unconditionally, so flipping this cell to `false` makes the operation globally dead rather than Apple-specific.

Concretely: `Interactor.findSelector?` now has zero implementers, `PlatformRuntimeOperations.findSelector` has zero binders, `waitOperations?.findSelector` is permanently `undefined`, and `dispatchConditionalWaitIfNeeded` in `wait-runtime.ts` is a permanent no-op guarding an unreachable `wait-conditional-selector.ts`.

Don't disable the operation, delete it. That removes this cell, the four sibling cells, the daemon module and its dispatch, the two binding-thread lines, the contracts types and plan-use entries, `Interactor.findSelector?`, and the `runner-demand.ts` row. The typechecker will walk you through it, and the result is strictly smaller than what this diff leaves behind.

Precedent is in the repo: the note in `ios-world.ts` records that R37 retired the same `querySelector` shortcut for `is`. Finish the same job for `wait`.

(To be clear, the runner's `querySelector` command and `queryAppleRunnerSelector` both stay — the direct-iOS tap path and the offscreen target probe still use them.)

src/daemon/__tests__/wait-conditional-selector.test.ts
```diff
@@ -139,6 +139,21 @@ test('a negative native observation falls through to the same bound canonical ca
   expect(runtime.bindDevice).toHaveBeenCalledTimes(1);
 });
 
+test('an unavailable native selector query cannot fail a capture-backed wait', async () => {
```
thymikee: This assertion is now true by construction. No owner can produce a binding with `findSelector` bound, so `findSelector` can never be called for any input and this case cannot fail.

The wider problem is the 201-line harness above it: it hand-forges a `DeviceBinding` with `findSelector` available, a shape the product no longer has. That's what keeps unreachable code alive — the module is dead, but the suite proves it works, so nothing flags it.

Delete this file together with `wait-conditional-selector.ts`. The real coverage for the fix is the new provider scenario in `ios-wait-selector.test.ts`, which exercises the actual bound path.

website/docs/docs/commands.md
```diff
@@ -413,7 +413,7 @@ agent-device alert dismiss
 ```
 
 - `wait` accepts a millisecond duration, `text <value>`, a snapshot ref (`@eN`), a selector, or strict `absent <selector>`.
-- `wait <selector> [timeoutMs]` polls until the selector resolves or the timeout expires.
+- `wait <selector> [timeoutMs]` polls until the selector resolves or the timeout expires. Apple selector waits use snapshot capture, including steps inside `replay`, so they share the capture backend and recovery used by `is`.
```
thymikee: Singling out Apple implies other platforms behave differently. After this PR no platform has a native selector wait, so the qualifier is stale on landing.

With the operation deleted, capture-backed is simply what `wait <selector>` means and there is nothing to contrast it with. Either state it platform-neutrally or drop the sentence.

thymikee (commented):

The behavioral fix is right and the live validation is convincing. But the PR retires an operation by **disabling** it rather than **deleting** it, and it leaves the identical defect live on the sibling port. Requesting changes on two structural items.

---

### 1. `findSelector` is now globally dead. Delete the operation, don't flip a fact to `false`.

After this diff there is **not one producer of `findSelector` left in the repo**. Every owner declares it unavailable:

| owner | declaration |
| --- | --- |
| apple | `nativeSelectorUnavailable` (this PR) |
| android | `snapshotKindUnavailable` |
| harmonyos | `snapshotKindUnavailable` |
| web | `openTargetKindUnavailable` |
| limrun | `customSnapshotUnavailable` / `liveSessionUnavailable` |

All five are unconditional constants, not device-dependent cells. `Interactor.findSelector?` has zero implementers. `PlatformRuntimeOperations.findSelector` has zero binders. So `waitOperations?.findSelector` is now permanently `undefined`, which makes `dispatchConditionalWaitIfNeeded` in `wait-runtime.ts` a permanent no-op and `wait-conditional-selector.ts` unreachable in its entirety.

The PR description says "Other runtime owners retain their observation contracts." They retain a declaration that the operation does not exist. That is not a contract, it is a row.

What survives, alive and unreachable:

- `src/daemon/wait-conditional-selector.ts` — the whole 64-line module
- `src/daemon/wait-runtime.ts` — `dispatchConditionalWaitIfNeeded` plus its import and call site
- `src/daemon/selector-operation-binding.ts` — `BoundNativeSelectorRead` and the `selectedSelector` half of `selectWaitObservationOperations`, which then collapses to a one-key projection shaped exactly like `selectElementTextOperation` right above it
- `src/daemon/selector-capture-binding.ts` and `src/daemon/snapshot-runtime-binding.ts` — three threading lines each
- `packages/contracts` — `FindSelectorInput`, `FindSelectorResult`, `FindSelectorRuntimeOperations`, the facts key, the operation name, both `conditional: ['findText', 'findSelector']` plan uses, and the unavailable-table row
- `packages/contracts/src/interactor-types.ts` — the optional `findSelector?` method
- `packages/platform-apple/src/runner-demand.ts` — the demand classification
- six owner fact declarations

That is roughly 200 lines of production plumbing kept alive to express "nobody does this," plus a 201-line test file (below) that will guard it indefinitely.

The code-judo move is to delete the operation end to end. This repo has already done exactly that once: the note in `ios-world.ts` records that R37 retired the same `querySelector` shortcut for `is`, and the scripted transcript is the proof the bypass is gone. Finish the same job for `wait`. The deletion is mechanical, the type checker walks you through it, and it is strictly smaller than the state this PR leaves behind.

To be precise about scope, this is **not** an argument to delete the runner's `querySelector` command or `queryAppleRunnerSelector`. Both stay: the direct-iOS tap path and the offscreen target probe still use them legitimately. Only the runtime *operation* is dead.

---

### 2. `findText` is the surviving twin of this exact bug, and the PR does not close the class.

`interactor.findText` calls `runAppleRunnerCommand({ command: 'findText' })` — the XCTest runner, same as the retired `querySelector` did. Nothing between `bindAppleFindTextRuntime` and `wait-text.ts` catches a throw:

- `runtime-snapshot.ts` awaits `interactor.findText(...)` bare
- `selector-runtime-backend.ts` forwards it bare
- `wait-text.ts` awaits `runtime.backend.findText(...)` bare, inside the poll loop

So on a live-but-poisoned runner — precisely the state in #2404, where the error was `XCTest recorded a failure while executing querySelector` — a `wait text <value>` still fails outright while a capture would have satisfied it. `hasLiveRunnerSession` gates *startup*, not *health*, so the admission check does not cover this.

The doc block on `bindAppleFindTextRuntime` states the invariant explicitly: "All report `found: false` — 'not proven here' — never an error, so the caller's canonical tree remains the complete path (ADR 0019 section 2)." Today that invariant holds only for the admission conditions the author remembered to enumerate. A runner-level throw walks straight past it.

Right now the invariant is enforced by taste. Make it structural: the one place that forwards a native observation should be **total** — any failure degrades to `{ found: false }` and the poll consults the canonical tree. One `try/catch` at the single forwarding seam, and the defect becomes unrepresentable for every current and future native find port, instead of being re-fixed port by port.

Related, and probably the actual mechanism worth putting in the PR description: `runner-demand.ts` classifies both `findText` and `findSelector` as `'simulator'`, meaning "served by simctl or the host AX bridge and never waits for runner readiness." Both implementations call the runner. `readTextAtPoint`, which also calls the runner, is correctly `'runner'`. That misclassification is a plausible root cause for the wait plan never guaranteeing a healthy runner in the first place. If it is, the same question applies to the rest of the table.

---

### 3. `wait-conditional-selector.test.ts` now tests a shape the product cannot produce.

The 201-line harness hand-forges a `DeviceBinding` with `findSelector` available and bound. No owner can produce that binding any more. The new case added here, "an unavailable native selector query cannot fail a capture-backed wait," asserts `findSelector` was not called — which is now true by construction for every input, so the test can no longer fail.

This is what keeps dead code alive: the module is unreachable, but the suite proves it works, so nothing flags it. Delete the file along with the module in item 1.

---

### 4. The docs line documents a transient state.

"Apple selector waits use snapshot capture, including steps inside `replay`" singles out Apple and implies other platforms differ. After this PR no platform has a native selector wait, so the Apple qualifier is wrong the moment it lands. Either state the behavior platform-neutrally, or drop the sentence — with the operation deleted, capture-backed is simply what `wait <selector>` means, and there is nothing left to contrast it with.

---

### Verdict

The behavior change is correct and well validated. But per the approval bar: there is a visible path that deletes a whole category of complexity instead of rearranging it, and it is not taken; and the fix addresses one instance of a defect whose sibling is still live and reachable. Items 1 and 2 are blocking. Items 3 and 4 fall out of item 1.

packages/platform-apple/src/runtime.ts
thymikee: Addressed in 985701a340. Deleted findSelector end to end: contracts and operation vocabulary, every owner fact, binding projections, Interactor method, wait dispatch/module, and runner-demand entry. No runtime-operation references remain. querySelector and queryAppleRunnerSelector remain for interaction paths. The actual daemon selector/replay regressions pass.

src/daemon/__tests__/wait-conditional-selector.test.ts
thymikee: Addressed in 985701a340. Deleted wait-conditional-selector.test.ts together with the unreachable production module. Coverage now runs through the real bound daemon routes in ios-wait-selector.test.ts, including standalone and replayed selector waits and native-text failures.

website/docs/docs/commands.md
thymikee: Addressed in 985701a340. The documentation now states selector-wait capture behavior platform-neutrally, including replay.

thymikee:
Addressed all four review items in 985701a340; the three inline threads have detailed replies.

For item 2, the sibling bug is now reproduced and fixed: standalone and replayed `wait text` failed when a live runner's `findText` threw despite a matching capture. The shared wait-observation binding now returns `found: false` on native failure so capture runs. Explicit tests preserve cancellation and ensure a subsequent capture failure remains visible.

Also corrected `findText` to `runner` in the demand table. This classification was misleading, but it was not the reproduced cause: `resolvePlannedRuntimeOperations` includes required operations only, so conditional `findText` never entered these open plans. The regression reproduces with a live runner and no preceding open. Planner and lifecycle tests retain observation-only startup without a runner.

Validation: `pnpm check:affected --run` passed on 985701a340 (3,753 related tests plus 12 documentation tests). Live iOS 26.2 Settings selector/text waits and two-step replay passed; session and daemon/runner cleanup completed. Updated-head GitHub CI is pending.

thymikee:
No remaining findings at 985701a. Selector waits and replay now use canonical capture; native text failures defer to capture without hiding cancellation or capture errors. The reported iOS verification covers the changed route, and CI is green. Ready for human review.

PR #2453, opened 2026-09-10
refactor(commands): declare project-config admission and recorder sanitization on the flag declaration

### LINKED ISSUE #2445: refactor(commands): declare project-config admission and recorder sanitization on the flag declaration
Wave 2 child of #2409. The "#1665-shaped follow-up" named in PR #2421; hops A11 and A18 in the #2410 hop trace.

## Purpose

A flag that may be set from `agent-device.json`, or that the session recorder may copy into `SessionAction.flags`, is declared once, on its flag declaration. Today each property is a separate hand-maintained allowlist, fail-closed by omission, so adding a flag that needs either behaviour edits the allowlist file as well.

## Baseline

On `main` at `41e2633f10`:

- `PROJECT_CONFIG_FLAG_KEYS`: module-private `Set<FlagKey>` of 85 keys at `src/cli-schema/cli-config.ts:37`. One consumer at `:219` rejects any project-config key outside the set with `INVALID_ARGS`.
- `SANITIZED_FLAG_KEYS`: module-private tuple of 39 keys (33 literals plus `SCREENSHOT_ACTION_FLAG_KEYS` from `packages/contracts/src/screenshot.ts:81`) at `src/daemon/session-action-recorder.ts:360`. One consumer, `sanitizeFlags` at `:397`, copies only these keys into the recorded action.
- Flag declarations: 156 across `src/commands/cli-grammar/flag-definitions-{action,connection,target,workflow}.ts` (39 / 38 / 44 / 35), typed by `FlagDefinition` at `src/commands/cli-grammar/flag-types.ts:19`. #2421 added `inputDescription` there and derived the command input field from it.
- No test pins membership of either set. A flag omitted from either set is silently rejected or dropped.
- The #2410 hop trace classifies both sets as "fail-closed by omission, so they must move onto the option declaration as explicit fields, never be inferred from a flag's existence".
- Replay of #1665 (`snapshotCustomActions`): the change edited both sets in addition to the declaration.

## Required work

1. Add two required fields to `FlagDefinition`: `projectConfig: boolean` (may be set from `agent-device.json`) and `recorded: boolean` (copied into `SessionAction.flags`). Required, not optional. Omission must be a type error, so the fail-closed property is held by the compiler instead of by a list.
2. Populate the 156 declarations from the current sets exactly (85 true / 71 false; 39 true / 117 false). Mechanical.
3. Derive: `cli-config.ts:219` and `sanitizeFlags` read the sets derived from the declarations. Delete the two literal sets. `SCREENSHOT_ACTION_FLAG_KEYS` stays while `screenshot-options.test.ts` still consumes it.
4. Planted divergence test, per the #2421 pattern (`src/commands/command-input-option-field.test.ts`): flipping `recorded` or `projectConfig` on one declaration changes the corresponding admission or sanitization outcome, observed through the production derivation.

- Re-doing #1665's flag addition edits the declaration (with both fields) and no allowlist. `src/cli-schema/cli-config.ts` and `src/daemon/session-action-recorder.ts` are untouched by the replay.
- The derived sets equal the deleted sets: the PR records the one-off diff (empty) and the existing project-config and recorder tests stay green with no edits.
- A flag declaration without both fields does not compile.

## Non-goals

- Inferring either property from a flag's existence, type or command. Rejected in the trace.
- `CONFIG_EXCLUDED_FLAG_KEYS` (6 keys) and `ENV_EXCLUDED_FLAG_KEYS` (2 keys) at `src/cli-schema/option-schema.ts:21` and `:30`. These are fail-open exclusions with a different safety property. Fold them in only if the same PR shows the same fail-closed guarantee for them; otherwise leave them.
- Per-command `allowedFlags` (36 sites): per-command admission, not collapsible (trace A3).

## Dependencies

None. Builds on #2421 (`ad79461c72`).

thymikee:
## What

A flag's two fail-closed properties — *may it be set from a project `agent-device.json`* and *does the session recorder copy it into `SessionAction.flags`* — move off their hand-maintained allowlists and onto each `FlagDefinition` as required `projectConfig` / `recorded` fields. Omitting either is now a type error, so the compiler holds the fail-closed property a list held by omission.

Closes #2445 (Wave 2 child of #2409; the "#1665-shaped follow-up" named in #2421).

## Changes

- `FlagDefinition` gains required `projectConfig: boolean` and `recorded: boolean`. All 156 declarations plus the 6 `ScreenshotSpecificFlagDefinition` entries carry both.
- **`recorded` is constrained to `CommandFlags` keys by the type**: `FlagDefinition` is a union that pins `recorded` to `false` for a `NonRecordableFlagKey = Exclude<FlagKey, keyof CommandFlags>` (json / help / version / replayMaestro / daemonAuthToken). This preserves the `satisfies readonly (keyof CommandFlags)[]` guard the old list carried, so `recorded: true` on a CLI-only key does not compile and cannot leak into a recorded action.
- `cli-config.ts` and `session-action-recorder.ts` derive their sets from the registry and no longer list keys (`PROJECT_CONFIG_FLAG_KEYS` / `SANITIZED_FLAG_KEYS` literals deleted). `SCREENSHOT_ACTION_FLAG_KEYS` stays (`screenshot-options.test.ts` still consumes it).
- Derivations (`projectConfigFlagKeys()` / `recordedFlagKeys()` in `flag-registry.ts`) recompute per call so each consumer freezes its set at its own module load — which is what lets a planted declaration move the surface. `recordedFlagKeys()` returns `ReadonlySet<RecordableFlagKey>` via a narrowing predicate, so `sanitizeFlags` needs no cast.
- The recorder reaches the derivation through the `cli-schema/command-schema.ts` seam: R2 `commands-floor` forbids `daemon/ -> commands/`.
- Planted-divergence tests per the #2421 pattern, plus two `@ts-expect-error` compile-time pins (missing fields; a CLI-only key cannot opt into recording).
- `docs/agents/cli-flags.md` points at the declaration fields, not the deleted allowlist.

## Acceptance (from the issue)

- **Re-doing #1665 edits the declaration, no allowlist**: both consumer files now derive; re-adding a flag touches only the declaration.
- **Derived sets equal the deleted sets** — one-off diff is **empty** (verified against `HEAD`):
  - PROJECT_CONFIG: derived 85 = original 85; only-derived `[]`, only-orig `[]`
  - RECORDED: derived 39 = original 39; only-derived `[]`, only-orig `[]`
  - `cli-config-trust.test.ts` (19) and `session-action-recorder.test.ts` (27) stay green **with no edits**.
- **A declaration without both fields does not compile**: enforced by the required fields + `@ts-expect-error` pins; full `pnpm typecheck` green.

## Size & design tradeoff

The size bot reports roughly **+3 kB unpacked** (the exact figure drifts run to run — it has shown +3.0 to +3.3 kB; download is slightly negative; startup within noise). The growth is the two required booleans repeated across all 162 declarations in the emitted bundles — inherent to stating the property per declaration rather than in one list.

The smaller alternative is to keep the two sets as standalone allowlists and not touch the declarations: near-zero bundle cost, but it is exactly the "fail-closed by omission, so a new flag edits a second file anyway" shape #2410/#2445 retire, and it cannot express the `CommandFlags` guard on a per-declaration basis. Retaining **required** declarations (so omission is a type error, not a silent drop) over another allowlist is the deliberate tradeoff here: +3 kB unpacked, one file per flag instead of two, and the guard moves onto the declaration.

## Verification

`pnpm build`, `pnpm typecheck` (all packages + examples), `pnpm lint`, `pnpm format:check`, `pnpm check:layering`, `check:production-exports`, `check:mcp-metadata`, `check:gate-manifest`, `check:bundle-owner-files`, `check:package` — all green. Targeted suites: `src/daemon` (2639), `packages/contracts` + `src/commands/capture` + `src/cli/parser` (447), plus cli-schema/commands/cli-grammar and the new test. Rebased onto `main` (`3bbeb61917`); no upstream change touched these files.

## Notes

- `docs/adr/0012-interactive-replay.md` still names `SANITIZED_FLAG_KEYS` in a point-in-time record; left unchanged (ADRs are historical and the described behavior — `--settle`/`--verify` dropped from recordings — is unchanged).

### DIFF
--- a/docs/agents/cli-flags.md
+++ b/docs/agents/cli-flags.md
@@ -8,9 +8,12 @@ Thread a flag only through the layers that consume it:
    flag; find the owner with
    `rg -n "<command>|supportedFlags|allowedFlags" src/commands src/cli-schema src/cli/parser`. For
    schema-only CLI commands, the owner is `SCHEMA_ONLY_CLI_COMMAND_SCHEMAS` in
-   `src/cli-schema/command-overrides.ts`. New flags are operator-only by default. Add a flag to `PROJECT_CONFIG_FLAG_KEYS` in
-   `src/cli-schema/cli-config.ts` only when repository control is safe; this positive allowlist is
-   the completeness gate.
+   `src/cli-schema/command-overrides.ts`. Every flag declaration states `projectConfig`
+   (may be set from a project `agent-device.json`) and `recorded` (the session recorder
+   copies it into `SessionAction.flags`). Both are required, so a new declaration that
+   omits either does not compile — that, not an allowlist, is the completeness gate. Set
+   `projectConfig: true` only when repository control is safe; a new flag is otherwise
+   operator-only. Set `recorded: true` only when a `.ad` recording must carry the flag.
 2. `src/commands/cli-grammar/*`: read the CLI flag into command input.
 3. `src/commands/command-projection.ts` and command-family projection helpers: write the input into
    the daemon request only if the flag affects daemon execution.
--- a/packages/contracts/src/screenshot.ts
+++ b/packages/contracts/src/screenshot.ts
@@ -97,6 +97,8 @@ type ScreenshotSpecificFlagDefinition = {
   max?: number;
   usageLabel: string;
   usageDescription: string;
+  projectConfig: boolean;
+  recorded: boolean;
 };
 
 export const SCREENSHOT_SPECIFIC_FLAG_DEFINITIONS: readonly ScreenshotSpecificFlagDefinition[] = [
@@ -107,6 +109,8 @@ export const SCREENSHOT_SPECIFIC_FLAG_DEFINITIONS: readonly ScreenshotSpecificFl
     usageLabel: '--crop-on <selector-expression>',
     usageDescription:
       'Screenshot: crop the capture to the frame of the selector resolved on the same screen',
+    projectConfig: false,
+    recorded: true,
   },
   {
     key: 'screenshotPixelDensity',
@@ -116,6 +120,8 @@ export const SCREENSHOT_SPECIFIC_FLAG_DEFINITIONS: readonly ScreenshotSpecificFl
     usageLabel: '--pixel-density <n>',
     usageDescription:
       'Screenshot: output PNG pixel density in pixels per logical point (currently supported on iOS simulators)',
+    projectConfig: true,
+    recorded: true,
   },
   {
     key: 'screenshotFullscreen',
@@ -124,6 +130,8 @@ export const SCREENSHOT_SPECIFIC_FLAG_DEFINITIONS: readonly ScreenshotSpecificFl
     usageLabel: '--fullscreen, --full, -f',
     usageDescription:
       'Screenshot: on web capture the full page; on macOS app sessions capture the full desktop instead of the app window',
+    projectConfig: true,
+    recorded: true,
   },
   {
     key: 'screenshotScale',
@@ -134,6 +142,8 @@ export const SCREENSHOT_SPECIFIC_FLAG_DEFINITIONS: readonly ScreenshotSpecificFl
     usageLabel: '--scale <0.01-1>',
     usageDescription:
       'Screenshot: resize both dimensions by this factor (or use AGENT_DEVICE_SCREENSHOT_SCALE)',
+    projectConfig: true,
+    recorded: true,
   },
   {
     key: 'screenshotNoStabilize',
@@ -142,6 +152,8 @@ export const SCREENSHOT_SPECIFIC_FLAG_DEFINITIONS: readonly ScreenshotSpecificFl
     usageLabel: '--no-stabilize',
     usageDescription:
       'Screenshot: skip Android demo-mode/status-bar stabilization and settle delay for low-latency capture loops',
+    projectConfig: true,
+    recorded: true,
   },
   {
     key: 'screenshotNormalizeStatusBar',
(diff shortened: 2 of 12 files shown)


thymikee:
The current admitted and recorded sets are preserved, but the recorder loses a compile-time guard. The old list had to satisfy keyof CommandFlags; recordedFlagKeys now returns any FlagKey and sanitizeFlags casts it, so recorded: true on a CLI-only key such as help or daemonAuthToken compiles. Constrain recorded declarations to CommandFlags keys, remove the cast, and add a compile-time rejection test.

Please also briefly account for the +3.0 kB unpacked growth and the smaller design considered. Explicit required booleans appear to explain most of it; retaining required declarations rather than another allowlist is a reasonable tradeoff to record.

thymikee:
Addressed both points:

**Recorder guard restored on the declaration.** `FlagDefinition` is now a union that pins `recorded` to `false` for a `NonRecordableFlagKey = Exclude<FlagKey, keyof CommandFlags>`, so `recorded: true` on `help` / `json` / `version` / `replayMaestro` / `daemonAuthToken` is a compile error — the same guarantee the old `satisfies readonly (keyof CommandFlags)[]` held, but stated on the declaration. `recordedFlagKeys()` returns `ReadonlySet<RecordableFlagKey>` via a narrowing predicate, and `sanitizeFlags` drops its `as keyof CommandFlags` cast. A `@ts-expect-error` test asserts a CLI-only key cannot opt into recording. Full `tsc`/lint/format/layering and the daemon + config + screenshot suites are green.

**Size.** +3.0 kB unpacked is the two required booleans repeated across 162 declarations; download is `-386 B`. Accounted for in the description along with the smaller alternative (keep the standalone allowlists) and why required-per-declaration wins: one file per flag instead of two, and the `CommandFlags` guard can only live on the declaration.

Rebased onto `main` (`3bbeb61917`).

thymikee:
No remaining code findings at c56f347. The declaration type now prevents CLI-only flags from being recorded, and the derived sets preserve all 85 project-config and 39 recorded keys. Current checks are green; the +3.3 kB unpacked growth is itemized and justified.

PR #2454, opened 2026-09-10
fix(ios): budget snapshot bridge test compile from the build ceiling

### LINKED ISSUE #2439: ios(ci): native-runtime bridge compile test uses a fixed 45 s budget that trips on a cold runner
Split out of #2422 (its third site; PR #2423 covers the two identity-probe sites only).

## Symptom

iOS `Smoke Tests`, step `Verify clean-installed Simulator snapshot bridge preparation`, fails inside the unit test `packages/platform-apple/src/snapshot-source/native-runtime.test.ts` (lines 13-32 at `9d7d60c5e0`): it compiles the snapshot bridge with `xcrun --sdk iphonesimulator clang …` under a fixed `timeoutMs: 45_000`, and on a cold runner reports `xcrun timed out after 45000ms`. Seen on PR #2421 (run 34366211699, 2026-09-09 14:54) and PR #2420 (2026-09-09 10:49). Every other check green; the branches did not touch the bridge.

## Cause

The first `xcrun` on a fresh macOS host pays the signature-scan stall (#2422) and then the clang compile itself; a fixed 45 s budget sized for a warm host trips. The production path budgets the same compile with `BUILD_TIMEOUT_MS` bounded by the 120 s snapshot-source deadline (`packages/platform-apple/src/snapshot-source/cache.ts`), so the test's budget is stricter than production's.

## Required behavior

Pick one:

- Budget the test compile from a deadline the way production does (`createSnapshotSourceDeadline(120_000, …)` and `remainingSnapshotSourceMs`), so the unit test is never stricter than the code it tests; or
- Skip the compile in the unit lane when the CI step has already verified it against the clean-installed package moments earlier (the step runs `verify-installed-snapshot-bridge.ts` right after), keeping one compile per job.

Either way: no rerun needed on that signature afterwards; record the run URL if it fires again.

## Non-goals

Widening the identity-probe budgets (done in #2423) or the production `BUILD_TIMEOUT_MS`.

thymikee:
## Summary

The host bridge unit test (`native-runtime.test.ts`) compiled the snapshot bridge under a fixed `timeoutMs: 45_000` with a `60_000` hook budget. On a cold macOS runner the first `xcrun` pays the signature-scan stall (#2422) and then clang, so the step reports `xcrun timed out after 45000ms` while every other check is green.

Both `beforeAll` compiles now budget from a deadline sized to the production build ceiling — `createSnapshotSourceDeadline(BUILD_TIMEOUT_MS)` + `remainingSnapshotSourceMs` — so the unit lane is never stricter than the preparation path it sits beside. The budget is a true `Math.min` of the same value production clamps to, not a fresh constant: `BUILD_TIMEOUT_MS` is now `@internal`-exported as the single source of truth for the compile ceiling. The hook budget leaves headroom beyond it so `runCmd`'s own timeout (not the Vitest hook) surfaces on a real hang.

Two files touched, both in `snapshot-source`. No production behavior or `BUILD_TIMEOUT_MS` value changes.

## Validation

Tested at `c394a46e59b03dc7af8a147969a48562678c4649`.

Local gates for the change pass: format, `oxlint`, `tsc`, `check:layering`, `check:fallow` (2 changed files, no issues), `check:production-exports`, `build`, `check:package`, and the full `snapshot-source` vitest family (native-runtime 46 tests green).

A broad `check:affected --run` fan-out surfaced unrelated daemon/provider failures (`daemon-entrypoint` 5 s timeout flakes that pass in isolation, a provider-integration CI lane, an xml teardown race, and two `request-router-lock-policy` real-time asserts). The lock-policy pair was reproduced failing identically on `origin/main` with this change reverted, so none are regressions here. Provider-integration, coverage, and device lanes remain GitHub-authoritative.

Cannot produce a local failing repro: the failure is cold-runner-only and this host compiles the bridge in seconds. The definitive signal is the CI step staying green; per #2439, reopen with the run URL if it fires again.

Closes #2439

### DIFF
--- a/packages/platform-apple/src/snapshot-source/cache.ts
+++ b/packages/platform-apple/src/snapshot-source/cache.ts
@@ -29,7 +29,13 @@ type SnapshotBridgeCacheManifest = Readonly<{
 const CACHE_SCHEMA_VERSION = 1 as const;
 const BRIDGE_FILENAME = 'snapshot-bridge';
 const MANIFEST_FILENAME = 'manifest.json';
-const BUILD_TIMEOUT_MS = 120_000;
+
+/**
+ * @internal Upper bound on a single snapshot-bridge clang invocation, exposed for the host bridge
+ * tests so they budget their own compile from the same ceiling instead of a stricter constant. The
+ * live build also stays under the caller's snapshot-source deadline, which can bind tighter.
+ */
+export const BUILD_TIMEOUT_MS = 120_000;
 
 export async function ensureSnapshotBridgeBinary(
   input: Readonly<{
--- a/packages/platform-apple/src/snapshot-source/native-runtime.test.ts
+++ b/packages/platform-apple/src/snapshot-source/native-runtime.test.ts
@@ -4,37 +4,48 @@ import path from 'node:path';
 import { beforeAll, describe, test } from 'vitest';
 import { runCmd } from '@agent-device/host-kit/command';
 import { mkdtempForTest } from '../__tests__/tmp-dir.ts';
+import { BUILD_TIMEOUT_MS } from './cache.ts';
+import { createSnapshotSourceDeadline, remainingSnapshotSourceMs } from './deadline.ts';
+
+// The host bridge compile is budgeted from a deadline sized to production's build ceiling, not a
+// stricter warm-host constant, so a cold runner's first `xcrun` (signature-scan stall plus clang)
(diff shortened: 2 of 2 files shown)


thymikee:
No code findings at c394a46. Both test compiles use the production build ceiling without widening the production budget, and all 46 native bridge tests passed in CI. The remaining iOS failure is in the unchanged identical-alert replacement test, not bridge compilation; it looks unrelated and needs a green rerun before merge.

Issue #2446, opened 2026-09-10
test(daemon): split the two replay test files still over the size tripwire


thymikee:
Follow-up to #2353 (child 7 of #2409). Zero metric credit; maintainability only.

## Purpose

No daemon test file sits over the 1,000-line tripwire, so the size ratchet (`scripts/__tests__/test-file-size-ratchet.test.ts:27`; rule at `docs/agents/testing.md:180`) can only fail on growth in `src/daemon/**`, never on legacy size. PR #2417 split three files and left these two.

## Baseline

On `main` at `41e2633f10` (#2417 merged as `29a4a3c64d`):

- `src/daemon/replay/internal/__tests__/session-replay-divergence.test.ts`: 1,100 lines, 13 flat `test(...)` cases, no `describe`. 12 exercise `buildReplayFailureDivergence` (dedup, keyboard and status-bar chrome filtering, the #1264 overlay/occlusion cap, freshness-retry routing, clean-flags policy; lines 53–1057). The last covers `captureDivergenceObservation` retry deadlines (1058 to end). Mirrors `session-replay-divergence.ts` (657 lines).
- `src/daemon/replay/internal/__tests__/session-replay-runtime-maestro.test.ts`: 1,963 lines, 43 flat cases. Its header says it is the Maestro-heavy remainder of the old `session-replay-vars.test.ts`, kept apart from `session-replay-runtime.test.ts` only because that file needs a different `vi.mock` of `@agent-device/device-selection/dispatch-resolve`. Clusters: `runReplayCommand` and `.ad` dispatch with env overrides (164–433); Maestro `runScript`, http and json (434–670); `tapOn`, `assertVisible`, `assertNotVisible`, `extendedWaitUntil` targeting (670–1223); swipe and text entry (1223–1493); enter, animation, retry (1493–1627); `runFlow.when` nesting and variables (1627–1863); shell env and trace/artifact plumbing (1863–1963). Closest source: `session-replay-maestro-runtime.ts` (423 lines).
- The ratchet has no baseline file. The limit for an over-tripwire file is its line count at the merge-base with `origin/main`.
- 17 other test files over 1,000 lines exist outside `src/daemon/` (largest: `src/__tests__/remote-connection.test.ts` 2,413; `src/commands/interaction/runtime/settle.test.ts` 2,359). Not in scope here.

## Required work

[...]

PR #2455, opened 2026-09-10
test(daemon): split the two replay test files over the size tripwire

thymikee:
## Summary

Follow-up to #2353. No `*.test.ts` under `src/daemon/**` should sit over the 1,000-line tripwire, so the size ratchet can fail on new size, not only on growth (#2417 left these two). Split into sibling `<base>-<concern>.test.ts` files:

- `session-replay-divergence.test.ts` (1,100) -> suggestions, chrome-filter, overlay, capture-policy, observation — 13 cases.
- `session-replay-runtime-maestro.test.ts` (1,963) -> dispatch, run-script, targeting, swipe, keyboard, run-flow, shell-artifacts — 43 cases.

Divergence siblings mock inline (Vitest hoists a file's own `vi.mock` above its own imports, so the SUT loads under them), matching the existing `session-replay-divergence-android-occlusion.test.ts` sibling. The freshness-retry `sleep` stub is declared only in the two siblings that exercise a retry branch, so it stays local. Maestro keeps `session-replay-runtime-maestro.fixtures.ts` (24-line `resolveTargetDevice` + the `runReplayFixture` harness): Vitest allows one `vi.mock` per module per file, so the Maestro family cannot reuse `session-replay-runtime.test.ts`'s differently-resolved device, and copying that mock + harness per file is what #2446 asked to avoid. 17 files touched, test-only.

**Isolation tradeoff vs #2446.** #2446 asked for a shared divergence fixtures module so a mock can't leak into unrelated siblings. A file's own inline mock cannot leak at all, and the divergence shared setup was only the 6-line dispatch+interactor mock plus a `resetLegacySnapshotCapture` call that already exists — extracting a fixtures module there forced a frozen-object SUT hand-off and a load-bearing import-order invariant, costing more than the leak risk it prevented. Maestro genuinely needs fixtures for the reasons above, so the two families diverge on purpose.

## Validation

Tested at `ba061f5` (final revision).
- 56 test cases preserved (13 + 43) under `src/daemon/replay/internal/__tests__`; test titles and assertion multisets diffed identical vs `main`.
- `pnpm check:affected --run` green (vitest-related, layering, mutation-registry, size ratchet — largest new file 605 lines).
- `pnpm lint`, `pnpm typecheck` green.
- Test-only change: no runtime/coverage/device behavior touched, so provider/coverage lanes are unaffected.


thymikee:
No test-behavior findings at ba061f5: all 56 test bodies are preserved, and the split keeps retry mocks isolated. Please refresh the summary and validation revision: the final diff has 17 files and file-local divergence mocks, not the shared divergence fixtures currently described; explain that isolation tradeoff against #2446. Completed checks are green, with iOS still running.

thymikee:
Refreshed the summary/validation to the final revision (`ba061f5`, 17 files): divergence siblings now carry file-local mocks (Vitest hoists a file's own `vi.mock` above its imports) with the retry `sleep` stub confined to the two retry siblings, and only Maestro keeps a fixtures module. Added an **Isolation tradeoff vs #2446** paragraph: inline mocks can't leak, and divergence's shared setup was too small to justify the frozen-object hand-off + import-order invariant the fixtures module required — Maestro keeps fixtures only because the one-mock-per-module rule and shared harness size make them load-bearing.

PR #2456, opened 2026-09-10
refactor(cli-schema): derive the flag tail of usage synopses

### LINKED ISSUE #2444: refactor(cli-schema): derive the flag tail of usage synopses so an option addition never edits usageOverride
Wave 2 child of #2409. Named but not done in PR #2421.

## Purpose

An option added to a command whose synopsis is hand-written must be restated in that synopsis by hand. #2421 made the descriptor the single declaration of an option for cli-schema flags, the MCP schema and daemon flags; the `usageOverride` string is the last restatement left on the help surface.

## Baseline

On `main` at `41e2633f10`:

- `buildCommandUsage` returns `schema.usageOverride` verbatim when it is set (`src/cli-schema/usage.ts:19`). Otherwise it renders the positionals plus one `[label]` per key in `allowedFlags`, in that order, using the flag declaration's `usageLabel`.
- 39 command descriptors set `usageOverride` (`grep -rn "usageOverride:" src/commands`). 38 encode shapes the generator cannot express: alternation (`clipboard read | clipboard write <text>`), subcommand forms (`session list | session state-dir | …`), value enumerations (`orientation <portrait|…>`), multi-line families (`perf`, `metro`).
- `snapshot` is the one override whose content is flags-only (`src/commands/capture/snapshot.ts:54`). It differs from the generated line in two ways. The generator renders `--depth, -d <depth>` and `--scope, -s <scope>` (the declared `usageLabel`, `src/commands/cli-grammar/flag-definitions-workflow.ts:185`) where the override shows `-d <depth>` and `-s <scope>`. The generator appends `[--record]` because `record` is in `allowedFlags` (`snapshot.ts:56`), while no synopsis in `website/docs/docs/commands.md` shows `[--record]` today (0 occurrences).
- Replay of #1665 (`--actions` on `snapshot`): the change edited the override string and the docs line `website/docs/docs/commands.md:353`, which `src/__tests__/command-doc-coverage.test.ts:78` pins byte-for-byte against `buildCommandUsage('snapshot', …)`.
- Help tests are literal `assert.match` asserts, not golden files (`src/cli-schema/cli-help-command-usage.test.ts:270` and siblings). None pins the full `snapshot` line.

## Required work

1. Decide the two rendering rules on the flag declaration, not per command: (a) how a flag with a short alias renders in a synopsis (full `--depth, -d <depth>` as generated lines do today, or short-only as the `snapshot` override does), and (b) which cross-cutting flags stay out of synopses. Encode (b) as a field on the flag declaration (for example `usageHidden: true` on the `record` definition), never as an override.
2. Drop `usageOverride` from `snapshot`. Its `--help` line is generated; the docs line at `commands.md:353` moves to the generated form and the byte-for-byte doc test stays green.
3. Classify the remaining 38 overrides in the PR body: "positional grammar the generator cannot express, keep" or "flag tail derivable". Where an override is a hand-written positional part followed by flags, split it so only the positional part stays hand-written and the flag tail is generated from `allowedFlags`. Pure alternation and subcommand overrides stay as they are.
4. Guard: a test fails when a command has `usageOverride` set and a key in its `allowedFlags` that is neither marked hidden nor mentioned in the override. An override can no longer silently lag its flags.

- Re-doing #1665's `--actions` addition on the new shape edits no `usageOverride` string. The docs line still changes; that is the documented coverage rule, not a restatement.
- `agent-device snapshot --help` output is either byte-identical to today or changed exactly per rules 1(a) and 1(b), with the help tests updated in the same PR and the rule pinned by one test.
- Every remaining `usageOverride` mentions every non-hidden flag in its command's `allowedFlags` (the step-4 guard is green), or the command is listed in the PR as an intentional exception with the reason.

## Non-goals

- A usage grammar for alternation or subcommands; the non-derivable overrides stay hand-written.
- Changing option names or semantics. `listUsageOverride` is untouched.

## Dependencies

None. Builds on #2421 (`ad79461c72`).

thymikee:
## Summary

A `--help` synopsis restated the options its command already declared. #2421 made the flag declaration own the option; this removes the last copy.

A synopsis is now grammar plus a generated `[label]` tail, and the two rendering rules sit on the declaration: the tail names an option with its declared `usageLabel`, alias included — the token `Command flags:` already shows; and `usageHidden: true` keeps a cross-cutting opt-in out of every synopsis (`--record` today, still documented under `Command flags:`).

`usageFlags` is the one addition a command may state: the options its synopsis names, defaulting to `allowedFlags`, `[]` when the synopsis is pure grammar or writes its own mutually-exclusive brackets. So adding an option updates `--help` on its own unless that command said its synopsis stays short — the curation is now a declaration instead of an unstated property of a prose string.

```
$ agent-device snapshot --help
Usage:
  agent-device snapshot [--diff] [-i] [--depth, -d <depth>] [--scope, -s <scope>] [--raw] [--actions] [--force-full] [--timeout <ms>]
```

`snapshot` and `proxy` drop their `usageOverride`; five more drop only the flag brackets. `is` loses `[--record]`; the other 67 synopses are byte-identical, verified by generating all of them before and after. `commands.md` moves to the generated line and its byte-for-byte doc test stays green. 27 files. Closes #2444.

## Validation

`pnpm check:affected --run` green on `7240ef1db9` (format, lint, typecheck, layering, fallow, build, related Vitest, command-docs, agent-guidance); `check:fallow --base origin/main` clean. Replay, web-smoke and device lanes stay GitHub-authoritative: only synopsis strings changed.

Guards verified by planting: a forgotten `usageFlags: []` renders `[--settle] [--settle]` and fails; a tail naming an unaccepted option fails; an added option reaches `snapshot` and `prepare` with no override edit; a terse synopsis keeps it in `Command flags:` only.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ## Unreleased
 
+- Changed: a command whose synopsis is generated names each option with the label its declaration
+  carries, so `snapshot` now shows `--depth, -d <depth>` and `--scope, -s <scope>` where it used to
+  show the short aliases, and `--record` is documented under `Command flags:` instead of inside the
+  `snapshot` and `is` synopsis lines. `snapshot`, `proxy`, `daemon`, `device`, `doctor`, `prepare`
+  and `tv-remote` no longer restate their option list in a hand-written usage string, so adding an
+  option to those commands updates `--help` on its own (#2444).
 - Fixed: iOS `--depth` on `snapshot`, `is`, `wait`, `get`, and `find` no longer fails with
   `regular iOS snapshot presentation requires a valid viewport` when the runner plan is pinned or
   deferred to the private AX backend (custom actions, a private AX verdict on the session, or the
--- a/docs/agents/cli-flags.md
+++ b/docs/agents/cli-flags.md
@@ -40,6 +40,13 @@ steps 1-3, plus step 9.
 ## Where CLI help and schema live
 
 - Long help prose: `src/cli-schema/cli-help.ts`. Flag definitions: `src/commands/cli-grammar/`.
+- Synopsis: `src/cli-schema/usage.ts` generates the `[label]` flag tail from `allowedFlags`, so a
+  new option reaches `--help` without any synopsis edit. Declare `usageFlags` on the command only
+  when its synopsis names fewer options: `[]` for a synopsis that is pure grammar (or writes its own
+  mutually-exclusive brackets), otherwise the subset it names. `Command flags:` always lists
+  everything in `allowedFlags`. Keep a cross-cutting opt-in out of every synopsis with
+  `usageHidden: true` on its flag definition. `src/cli-schema/usage.test.ts` fails a tail that names
+  an option the command does not accept, or one the hand-written grammar already wrote.
 - Command-specific usage/flag metadata lives with the command family metadata that owns the command.
 - Parser/help *rendering* stays in `src/cli/parser/`; command schema metadata is derived from command
   metadata, family declarations, and the schema-only merge path in
--- a/src/cli-schema/cli-help-command-usage.test.ts
+++ b/src/cli-schema/cli-help-command-usage.test.ts
@@ -277,6 +277,13 @@ test('snapshot command usage documents diff alias', async () => {
   assert.match(help, /verify with diff snapshot -i or snapshot --diff/);
 });
 
+test('snapshot documents the synopsis-hidden record flag', async () => {
+  const help = await usageForCommand('snapshot');
+  if (help === null) throw new Error('Expected command help text');
+  assert.doesNotMatch(help, /agent-device snapshot \[[^\n]*--record/);
+  assert.match(help, /--record\s+Force-record this action/);
+});
+
 test('network command usage documents include flag', async () => {
   const help = await usageForCommand('network');
   if (help === null) throw new Error('Expected command help text');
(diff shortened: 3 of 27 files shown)


thymikee:
Rebased onto `main` (`805ffb4690`) and simplified the design in `7240ef1db9`. No review comments had landed, so this is self-inflicted.

The first shape added `deriveUsageFlagTail` and kept the 19 hand-written synopses that show only some options honest through a guard table listing ~90 flag keys per command. That table was itself a restatement of the option lists — the thing this change exists to delete — and it reconstructed per-command facts inside a test.

The new shape states them where the parser and help already read them: `usageFlags` (default `allowedFlags`) is the option set a synopsis names, so the tail is always generated and the ~23 commands that keep a terse synopsis say `usageFlags: []` at their own descriptor. Removed: `deriveUsageFlagTail`, the waiver table, and the prose-matching that had to guess whether a synopsis "mentioned" a flag. Two guards remain, both structural: a tail may not name an option outside `allowedFlags`, and a generated tail may not repeat a bracket the grammar already wrote. Diff is 224 added lines against 492.

Behaviour is unchanged from the previous head: all 69 synopses regenerate byte-identically, still only `snapshot` and `is` moving.

thymikee:
No code findings at 7240ef1. The shared renderer derives the flag tail, keeps intentional short forms explicit, and preserves the documented help output apart from the declared alias and record changes. Current checks are green; package growth is 100 B.

PR #2447, opened 2026-09-10
fix(ios): recover simulator recorder startup failures

PLASMA-FR:
## Summary

Closes #2170.

Recover iOS Simulator recording startup without leaving CoreSimulator's host-wide recording slot stuck. Unpublished recorder rollback now sends `SIGINT`, waits a bounded five seconds for graceful detach, and escalates only if the process remains alive. A native `simctl recordVideo` exit 16 is returned as non-retriable `DEVICE_IN_USE` with `details.reason: apple_simulator_recording_busy` and actionable recovery guidance; other exits retain their existing classification.

Startup now observes exits that settle before PID discovery or on the final process-identity poll. The provider scenario proves a failed start adopts no recorder and a later start/stop succeeds. Five files, +195/−26.

## Validation

Tested `06e98c7e6a7bf0517c31339fb2a4814ae055d9cc` with Node 24.13.1 and pnpm 11.17.0.

- The focused regressions failed before the fix; 19 simulator-host tests and the provider contention/recovery scenario pass.
- `AGENT_DEVICE_VITEST_MAX_WORKERS=1 pnpm check:affected --run` passes: 301 files / 2,144 tests, 12 command-doc tests, format, lint, types, layering, Fallow, and build.
- GitHub Coverage, Integration, and every platform smoke job pass on this head. This Linux host cannot run CoreSimulator, and ordinary iOS smoke does not poison/recover the host recording slot; that native contention scenario remains a disclosed review risk.

PR #2457, opened 2026-09-10
fix(ios): reap orphaned simulator recorders with a graceful finalize window

### LINKED ISSUE #2170: iOS sim: ungraceful recorder kill leaves CoreSimulator host recording lock dangling (code 16 for all later recordings)

thymikee:
## Summary

Narrows #2170 to its **non-overlapping** piece. #2447 already owns the simulator-recording *start* path — the exit-16 busy classifier, the fast-exit observation, and the graceful SIGINT rollback — under the `DEVICE_IN_USE` contract. To land a single start-path error contract, this PR keeps **only** the daemon-startup fix that #2447 does **not** touch.

When the daemon starts, it reaps orphaned `simctl recordVideo` children (`purposes: ['simctl-screen-recording']`). It previously used the generic 2s reaper grace. An orphaned recorder releases CoreSimulator's host-wide recording lock only after finalizing on `SIGINT`; force-killing it after 2s fixes the one device but leaves the lock dangling so every later recording on the host fails with `EBUSY`. This passes the recorder the same finalize grace the live stop path allows (5s). Bounded to the recorder purpose, so it never widens the web or generic reap budgets.

Adds one startup-wiring test proving the recording purpose is reaped with a `>= 5000ms` term timeout.

Closes #2170 (recorder-recovery half); start-path recovery and the busy error contract land in #2447.

## Why 5s

- `simctl recordVideo` finalizes the mp4 on `SIGINT`; the host recording slot frees only after that.
- Orphan reaping is sequential, but only a *live* wedged recorder pays the grace (dead records clear
  instantly), and the host-wide lock permits at most one live recorder — so startup stays under the
  15s client timeout.

## Validation

- Added `daemon-runtime-recording-reaper.test.ts`: drives real `startDaemonRuntime` (reaper spied) and
  asserts it invokes `reapOwnedProcessRecordsAtStartup` for the `simctl-screen-recording` purpose with
  a graceful finalize window. Fails if the grace drops to the generic default.
- `pnpm check:affected --run`, `tsc`, and `oxlint` pass on head `99a379f9f0`.
- Net change: `src/daemon/server/daemon-runtime.ts` +5 and one new test; host start path unchanged.
- **Scope note:** the busy `EBUSY` failure path and in-session rollback are validated in #2447, not
  here; a normal start/stop run does not exercise this startup reap (it only matters after an
  ungraceful daemon exit leaves a recorder orphan).

Related: #2170, #2447.


thymikee:
The startup failure is still missed at 1fb63cb: an immediate exit 16 can happen before process identity is readable, so waitForManagedProcessIdentity returns no marker and the new classifier is never reached. The test keeps identity metadata available for an already-exited child, masking that path. Observe process exit during identity acquisition and add a regression with missing identity plus exit 16.\n\nCanceled acquisition and identity-failure cleanup still use immediate SIGKILL. A recorder may already hold the stream slot there; discarding the video does not remove the need to detach cleanly. Reuse graceful rollback and verify recording works afterward. #2447 already addresses these same paths with a different error contract, so consolidate the overlapping changes before either lands.\n\nCI is green, but the reported normal start/stop run does not validate orphan recovery or the busy failure path.

thymikee:
@thymikee consolidating per your comment so #2447 and this don't land two start-path error contracts:

- **Fast exit before identity / busy classifier:** you're right that at `1fb63cb` an immediate exit 16 could settle before `waitForManagedProcessIdentity` reads a marker, so the classifier in `startError` was never reached, and the busy test kept identity metadata for an already-exited child and masked that path. Rather than duplicate the fix, I've **dropped the exit-16 classifier and the graceful start rollback here** — #2447 already observes exits that settle before PID discovery and on the final identity poll, and does it with the `DEVICE_IN_USE` / `apple_simulator_recording_busy` contract. Point 1 and point 2 (fast-exit regression, graceful rollback) are #2447's to own and validate.
- **Consolidation:** this PR is now reduced to the one non-overlapping piece #2447 does not touch — the daemon-startup reaper grace in `daemon-runtime.ts`. It reaps orphaned `simctl recordVideo` children with the recorder's 5s finalize window instead of the generic 2s, so the host-wide recording lock is released rather than left dangling. Net diff is now `daemon-runtime.ts` +5 and one test; the simulator-host start path is untouched.
- **Validation (point 3):** the normal start/stop claim was for the reduced-scope change and did not cover orphan recovery or the busy path — corrected in the body. Orphan recovery is now covered by `daemon-runtime-recording-reaper.test.ts` (drives real `startDaemonRuntime`, asserts the recording purpose is reaped with a >=5000ms term timeout); the busy path and in-session rollback are validated in #2447's provider scenario.

Gate on head `99a379f9f0`: `check:affected`, `tsc`, `oxlint` green.

thymikee:
The overlap is resolved at 99a379f: this now only gives the existing recorder-only startup reaper five seconds to finalize after SIGINT. No code findings. Before marking it ready, please show a CoreSimulator run that leaves a live recorder orphaned after an ungraceful daemon exit, restarts the daemon, and then successfully starts and stops a new recording. The wiring test and normal recording runs do not prove that recovery. The current iOS failure concerns alert replacement and appears unrelated.

thymikee:
@thymikee CoreSimulator recovery run done (iPhone 16, isolated daemon state dir, host restored after). The orphaned-daemon scenario you asked for:

**1. Contention proves the host-wide lock / exit 16 (raw `simctl io recordVideo`):**
```
rec2 exit code=16
Error starting video recorder: Error Domain=NSPOSIXErrorDomain Code=16 "Resource busy"
  UserInfo={NSLocalizedFailureReason=Host recording is already in progress}.
An error was encountered processing the command (domain=NSPOSIXErrorDomain, code=16):
The operation couldn’t be completed. Host recording is already in progress
```

**2. Orphan after ungraceful daemon exit, then daemon-restart recovery (`agent-device record start --scope device`):**
- `record start` → recorder pid 86202 live; raw contender → `exit=16` (lock held).
- `kill -9` the daemon (83157): `daemon=dead recorder=ALIVE(orphaned)`; raw contender → `probe2 exit=16` (orphan still holds the host lock — the exact #2170 stuck state).
- Restart daemon (`session list`) → the startup reaper reaps the orphan: `session-list returned after 0.35s; recorder alive = False` (finalized inside the 5s SIGINT grace). Raw contender then `probe3 recording OK` → lock freed.
- Fresh session after restart: `record start` **succeeded (no EBUSY)**, `record stop` succeeded → `recovered.mp4` = 52,241 bytes.

**3. Host restored:** graceful `record stop`, graceful `daemon stop`, sim shut down, no leftover `recordVideo` — recording is healthy again for other sessions.

This exercises the exact path this PR changes: the orphaned `simctl-screen-recording` record is now finalized with the 5s grace (`purposes:['simctl-screen-recording'],termTimeoutMs:5e3`, SIGINT) instead of the generic 2s, so the host-wide lock is released on restart instead of dangling.

CI: the only red check is an unrelated iOS-runner XCTest alert-replacement timing flake (`testAlertAcceptDoesNotActivateAReplacementWithASharedButton`), as you noted — nothing in this diff touches the runner.

thymikee:
The reported CoreSimulator run now exercises the requested orphan-recovery path: daemon death leaves the recording lock held, startup reaping releases it, and a new recording starts and stops. No remaining code or device-evidence concern at 99a379f; ready for human review. The failing iOS check is testAlertCannotProveAnIdenticalReplacementAndDoesNotActivateIt, which appears unrelated to this recorder-only grace change. Keep #2170 open for #2447’s start-path work.

PR #2459, opened 2026-09-10
refactor(contracts): additive capability facts on the keyboard family

thymikee:
## Summary

`keyboardRuntimeOperationFacts` required one cell per operation, so an operation only one owner
serves still cost a hand-written denial in every other owner. The builder now takes the family's
denial once, as a required `unsupported` cell; `status`/`dismiss`/`enter` are optional and an
omitted cell reports that denial verbatim.

The shared unavailable-facts record collapses `keyboardStatus`/`keyboardDismiss`/`keyboardEnter`
into one required `keyboard` cell, so the owners that wrote raw keys — webdriver, vega, linux,
limrun's no-session binding — answer the family through the builder, its one entry point. Facts are
unchanged for every owner: no test asserting a non-supporting platform's unavailable reason was
edited. Keyboard is the first family of #2443, on the shape #2420 merged.

## Replay (#2443 item 3)

Replayed addition: `keyboardCommitIme`, one operation named by exactly one owner. Files that must
change until `pnpm typecheck` is green again:

| shape | files | packages | unsupported-owner edits |
| --- | ---: | ---: | ---: |
| old (`origin/main`) | 25 | 9 | 17 |
| new (this branch) | 8 | 2 | 0 |

The 8: 6 in `packages/contracts` (the family, the interactor method, the operation-name list, the
binding catalog, the conformance row, the family's own exhaustive test) + 2 in
`packages/platform-apple` (the implementing cell, plus `runner-demand.ts`, a facts-free total
operation→host map). Those operation-surface files are forced on both shapes. The old shape
additionally forced 9 owner/builder fact cells (apple navigation, harmonyos, linux, vega, web,
limrun ×2, webdriver, the shared record) and 8 cell-enumerating test doubles.

Stop rule: 17 avoided ≥ 2 → keyboard merges. Measurement also posted on #2443.

## Validation

`880c9f4367`: `pnpm check:affected --run` green — format, lint, typecheck, layering, fallow, build,
and 350 related test files / 2339 tests. Provider integration and coverage stay
GitHub-authoritative. Parity is proven by the untouched admission tests
(`packages/platform-*/src/runtime.test.ts`, provider runtime tests); two owner tests now also
assert the keyboard *reason* (WebDriver inactive session, stale Limrun identity), because one family
cell answers three operations.

Refs #2443

Issue #2443, opened 2026-09-10
refactor(contracts): expand additive capability facts to the keyboard, gesture and clipboard families


thymikee:
Wave 2 child of #2409. Continues #2412 (pilot merged as #2420, `edbd96ea30`). The pilot's stop rule passed with seven unsupported-owner edits avoided against a floor of two, so expansion is authorized on the same terms, one family per PR.

## Purpose

Adding a capability to one platform stops forcing edits in the platforms that do not support it, for the next families where owners still write a denial out cell by cell.

## Baseline

On `main` at `41e2633f10`:

- The touch family is additive. `TouchRuntimeOperationFactsInput.unsupported` is a required family-level denial (`packages/contracts/src/touch-runtime.ts:88`), every operation input is optional, and `declared(fact) = fact ?? input.unsupported` (`:103`). An owner that leaves an operation blank gets the family's typed denial with the same reason and hint it would have written by hand.
- Every other multi-operation family is required-per-operation: `keyboardRuntimeOperationFacts` (`packages/contracts/src/keyboard-runtime.ts:46`, 3 operations), `gestureRuntimeOperationFacts` (`gesture-runtime.ts:74`, 5), `clipboardRuntimeOperationFacts` (`clipboard-runtime.ts:48`, 2), plus alert (4), snapshot (3), selector-observation (2), perf (7), audio-probe (4), application-lifecycle (9).
- Owners that write a whole family as unavailable today, cell by cell:
  - keyboard: web (`packages/platform-web/src/runtime.ts:416`), webdriver (`packages/provider-webdriver/src/platform-runtime.ts:568`, raw keys without the builder), vega (`packages/platform-vega/src/runtime.ts:203`, raw keys). harmonyos implements 2 of 3.
  - gesture: vega (`packages/platform-vega/src/runtime.ts:236`), web (`packages/platform-web/src/runtime.ts:401`).
  - clipboard: harmonyos, web, vega.
- PR #2420's replay of an Apple-only touch operation: 16 files / 8 packages before, 5 files / 2 packages after.

## Required work

One PR per family, in this order: keyboard, gesture, clipboard.

1. Give the family builder the touch shape: a required `unsupported` cell on the input, optional per-operation inputs, `declared(fact) = fact ?? input.unsupported`. Owners that write raw keys (webdriver and vega for keyboard) move to the builder in the same PR, so the family has one entry point.
[...]

PR #2460, opened 2026-09-10
refactor(contracts): additive capability facts on the gesture family

thymikee:
## What

Additive capability facts on the gesture family, the second family in #2443. Stacked on #2459.

`gestureRuntimeOperationFacts` took one fact per tier, and it had five: `plan`, `directionalFling`, `multiTouch`, `targetAuthoredDrag`, `viewport`. Gesture tiers split exactly where an owner's mechanics split — no tier is one every owner serves, and several owners serve none — so every tier is optional and `unsupported` names the denial an omitted tier reports. A call naming no denial is still refused: omission is a classified refusal, never an unclassified tier and never an implied success.

An owner states a tier only to say something the family denial does not. Web refuses every tier with one cell and names only the drag tier, whose wording came from a retired closure; Vega names the two tiers whose refusals came from separate closures; Limrun's session-less and iOS branches name one denial each where each wrote five. The owners that serve gestures keep naming what they serve — the win is the next tier, which an owner can now ignore.

## Measured

Replayed addition: `performEdgeSwipePlan`, one tier named by exactly one owner (Apple), applied identically to this PR's base and to this head.

| shape | files | packages | unsupported-owner edits |
| --- | ---: | ---: | ---: |
| old (required cell per tier) | 15 | 9 | 10 |
| new (`unsupported` + optional tiers) | 5 | 2 | 0 |

The five the new shape forces are the addition and the implementing owner. The ten that disappear are the seven non-implementing owner cells, the shared unavailable record, and two cell-enumerating test doubles. No test asserting a non-supporting owner's unavailable reason was edited.

**Stop rule: 10 avoided >= 2 -> gesture merges.** Clipboard follows.

Refs #2443

## Validation

`pnpm check:affected --run` green on the base and on head; provider integration and coverage are GitHub's to report.


thymikee:
No code findings at 3b4b615. The updated Apple family denial preserves leaf-specific refusals, and existing gesture facts and bindings stay unchanged. Merge after #2459; several checks are still running, so this is ready for human review rather than a CI-green merge verdict.

PR #2464, opened 2026-09-10
refactor(contracts): additive capability facts on the clipboard family

thymikee:
## What

Additive capability facts on the clipboard family, the third family in #2443. Stacked on #2460.

`clipboardRuntimeOperationFacts` took exactly two cells, read and write, and both were required. The halves stay separate claims — a WebDriver provider whose Appium clipboard extension exposes only a getter is a real owner with one half and not the other, and `clipboard read` must not be refused because the write half is missing — so both are optional now and `unsupported` names the denial an unnamed half reports. A call naming no denial is still refused: omission is a classified refusal, never an unclassified half and never an implied success.

Six of the eight owners drive both halves from one shell command set or one leaf gate, so they state that once: web, HarmonyOS and Vega name a single denial, and the shared unavailable record carries one `clipboard` cell where it carried two. The owners that serve clipboard keep naming both halves, and the one install-source double that enumerated the keys by hand now goes through the family builder like every other construction site.

## Measured

Replayed addition: `readClipboardHtml`, one clipboard operation named by exactly one owner (the Linux desktop), applied identically to this PR's base and to this head.

| shape | files | packages | unsupported-owner edits |
| --- | ---: | ---: | ---: |
| old (required cell per half) | 23 | 9 | 15 |
| new (`unsupported` + optional halves) | 8 | 3 | 0 |

The eight the new shape forces are the operation itself (`contracts/src/{clipboard-runtime,interactor-types,interactor-operation-catalog,runtime-operation-names}.ts`, the family's conformance table, its own test) plus the implementing owner and Apple's facts-free operation-to-host map. The fifteen that disappear are the shared unavailable record and its test, six owner fact cells (`android`, `apple`, `harmonyos`, `vega`, `web`, `webdriver`), both Limrun cells, and five cell-enumerating test doubles including the shared `runtime-operation-facts` helper.

**Stop rule: 15 avoided >= 2 -> clipboard merges.** That is the third and last family.

Refs #2443

## Validation

`pnpm check:affected --run` green on the base and on head; provider integration and coverage are GitHub's to report.


thymikee:
Looks good at d5185f8. Read/write admission and existing refusal reasons are preserved, including partial WebDriver support; the new tests cover omitted and explicitly declared facts. All completed checks pass; iOS smoke is still running, and this follows #2460.

One nonblocking correction: the replay table should say 3 packages, not 2, since the listed files span contracts, Linux, and Apple.

thymikee:
Corrected: the new-shape row now reads 3 packages (contracts, `platform-linux`, `platform-apple`). Verified the same count on the other two families: keyboard 2 (`contracts`, `platform-apple`), gesture 2 (`contracts`, `platform-apple`); the PR body and the #2443 measurement comment are both updated.

PR #2467, opened 2026-09-10
perf(package): strip comments from the Apple runner source the npm package ships

### LINKED ISSUE #2461: perf(package): strip comments when packaging apple/runner source into dist
**Purpose.** `apple/runner/**` ships to npm as plain Swift **source**: `scripts/prepare-publish-assets.mjs` → `scripts/package-apple-runner-source.mjs` copies it into `dist/apple/runner/`, stripping only `#if AGENT_DEVICE_RUNNER_UNIT_TESTS` blocks. Comments are shipped verbatim.

**Finding.** Measured while itemizing the size growth on #2448: comments are **16.6% of the 446 kB of shipped runner Swift — ~74 kB across 45 files**. For context, that is roughly 9x the entire +8.1 kB unpacked growth of #2448, and it is paid on every install.

**Required behavior.** The packager already rewrites these files, so extend it to strip comments (line and block) while preserving semantics — string literals containing `//`, `#if` directives, and any comment the Swift build depends on must survive. Verify the packaged runner still builds and its tests pass.

**Observable completion.** `dist/apple/runner/**` is materially smaller with an unchanged runner build; the Bundle Size job shows the one-time drop.

**Notes.** Independent of #2448, which only surfaced the measurement. A BSD-`awk` gotcha already exists in that script (`\b` regexes silently fail) — prefer Node for the transform.

thymikee:
## Summary

`apple/runner/**` ships to npm as Swift source, and the packager rewrote it only to remove
`#if AGENT_DEVICE_RUNNER_UNIT_TESTS` blocks — so every doc comment and design note was downloaded
on every install. `scripts/package-apple-runner-source.mjs` now also strips comments, through a new
`scripts/strip-swift-comments.mjs`.

A regex cannot do this safely: `//` and `/*` open a comment only in code position, a raw literal
(`#"…"#`) moves its own closing delimiter and interpolation opener with the `#` count, interpolation
segments hold code including further literals, extended regex literals (`#/…/#`) are a second
delimiter family that also opens on a `#` run, and Swift block comments nest. The module is a
lexical scanner over those states; a construct it cannot account for throws at packaging time
rather than shipping Swift that does not compile. Blank lines and every byte inside a literal
survive. The unit-test block strip is unchanged in what it removes and still runs first, so which
blocks it removes does not depend on comment removal.

Bare `/…/` regex literals are the one construct no scanner can resolve — the same `/` opens a
comment, divides, and starts a regex literal, and only the parse separates them. Where one could
begin (an expression position whose `/` is not followed by a space, a tab or `)`), packaging throws
by file and line instead of rewriting bytes it cannot prove are code. Divisions (`width/2`,
`Double(3)/Double(4)`), the recording scripts' `#!` shebang and `(/)` flow through untouched.

## Line numbering is preserved

`dist/apple/runner/**` is the source a user's `xcodebuild` and the runner name a file and line in
(it lands in `runner.log`), so those numbers are only worth reading if they point at the same line
of `apple/runner/**`. Every removed line is emitted **empty** rather than deleted, in both passes:

- comment removal: 889 comment-only lines, 889 B;
- the pre-existing unit-test `#if` block strip, which was already moving everything below a block
  by up to 883 lines: 3,737 lines, 3,737 B.

Proved over all 44 shipped files: `wc -l` is identical for 44/44, and all 613 declarations the
packaged files still carry sit on their checkout line (`RunnerTests.swift` `func testCommand` at
245 in both; `recording-overlay.swift`'s first `struct` at 14 in both).

## A gate that reads the packaged output

Nothing in this repo compiles, imports or reads `dist/apple/runner/**` — the first consumer is a
user's `xcodebuild` — so a future mis-lex that failed to throw would ship Swift that does not build
with no signal. `pnpm check:packaged-runner-swift` (new, registered in `CHECK_CATALOG`, wired into
`pnpm check:tooling` next to `check:xctest-selection`, and declared on the macOS lane) packages into
a throwaway root and holds both properties: line parity per file, then `swiftc -parse` over each
packaged `.swift`. Seconds, no simulator. Parity needs no toolchain; the parse reports itself
skipped where no Swift exists, which is why the macOS lane owns the gate. `apple/runner/**` and
both packaging scripts select it in `check:affected`.

## Size

`dist/apple/runner/` 555,907 B -> 488,635 B (-67,272 B, -12.1%); its Swift alone 441,196 B ->
373,924 B (-67,272 B, -15.2%). Line parity costs 4,626 B of the 71,898 B the first pass saved
(-12.9% before parity). Closes #2461.

## Validation

Tested commit e0e4246da7.

- `pnpm format`, `pnpm lint`, `pnpm typecheck`, `pnpm check:affected --run` (which runs
  `check:fallow --base origin/main`), `pnpm check:xctest-selection`,
  `pnpm check:packaged-runner-swift`, `pnpm check:layering`, `pnpm check:gate-manifest`: all pass.
- Packaged runner builds: `xcodebuild build-for-testing` on
  `dist/apple/runner/.../AgentDeviceRunner.xcodeproj` reports `** TEST BUILD SUCCEEDED **` for both
  `generic/platform=iOS Simulator` and `platform=macOS,arch=arm64`. Build only; no `xcodebuild test`
  and no simulator or device run.
- All 44 packaged Swift files pass `xcrun swiftc -parse` — now as the permanent gate above, not a
  one-off.
- Both new gate properties were proved by planting a violation: reverting comment-only lines to
  being deleted fails the gate on the shifted files by name, and a scanner that leaves a stray `/`
  where a comment was fails the parse on 35 of 44 files with `file:line`.
- Re-running the scanner over its own output removes 0 further comments in all 44 files.
- Every Swift snippet in the regex-literal regressions was checked against `xcrun swiftc -parse`
  (Swift 6.2.3) both as written and as the scanner leaves it, so the fixtures are valid Swift
  rather than a guess at the grammar.
- No device run: packaging-only change, and the packaged Swift is unchanged apart from comments and
  blanked lines. Confirm the Bundle Size job for the one-time drop.


thymikee:
One packaging safety gap at d95e6ab: valid Swift extended regex literals are silently rewritten. `let pattern = #/foo//bar/#` becomes `let pattern = #/foo`, because the scanner treats the regex contents as a line comment. Please preserve regex literals or reject unsupported regex syntax before rewriting, and add this regression. The current runner corpus has no such literals, so the existing compile evidence is consistent with this gap; the claimed fail-closed behavior still needs fixing.

thymikee:
You were right, and the characterisation was exact: at d95e6ab the scanner only knew the `#"`
raw-string family, so `#/foo//bar/#` lost everything from its first inner `//` and shipped
`let pattern = #/foo`. Fixed in 9770a12e15.

**What changed in `scripts/strip-swift-comments.mjs`**

- Extended regex literals are now their own frame kind, opened by `#+` followed by `/`. This is the
  `#`-run discrimination you flagged: `#` + quote is a raw string literal, `#` + `/` is a regex
  literal, and `#` + anything else stays a directive (`#if`, `#available`, the recording scripts'
  `#!`). The `#` count sets the terminator, so `##/…/##` closes only on `/##`.
- Both forms are handled. A newline straight after the opener selects Swift's multi-line form,
  whose closing delimiter has to start its own line; the single-line form throws
  `Unterminated regex literal` at a newline instead of scanning on. `\/` is copied as a pair, so an
  escaped slash cannot close the literal early.
- Bare `/…/` takes your second option. A `/` in code position that could open one — an expression
  position, and not followed by a space, a tab, a newline or `)`, which Swift itself bars a regex
  literal from opening on — now throws
  `Ambiguous bare regex literal or division in <file>:<line>`, naming the file and line and
  suggesting `#/…/#` or a spaced operator. Nothing is guessed at. Expression position is decided
  from the emitted tail: a `/` after an operand (identifier, `)`, `]`, `"`, backtick, `?`, `!`)
  divides, except after a keyword that can precede an expression, so `return /x/` is refused while
  `width/2` is not.
- `scanCodeCharacter`'s slash handling moved into `consumeSlash` to stay under the fallow
  complexity threshold.

**What the new regressions prove**

`scripts/__tests__/strip-swift-comments.test.ts` (18 -> 25 tests) and
`src/__tests__/apple-runner-package-source.test.ts` (+1):

| Regression | Before this commit |
| --- | --- |
| `#/foo//bar/#` survives byte for byte | shipped `let pattern = #/foo` |
| `##/a//b/#c/##` survives (the `#`-count terminator) | shipped `let pounded = ##/a` |
| multi-line `#/ … /#` holding `//`, `/*e*/`, `a\/#b` and a blank line survives | shipped the literal with two lines gutted |
| `#/x/*y/#` survives | threw `Unterminated block comment` — a false packaging failure |
| `#/a//b/#  // trailing` keeps the literal, drops only the real comment | shipped `let trailing = #/a` |
| `/foo//bar/` and `return /x\/y/` throw, naming file and line | silently shipped `let pattern = /foo` |
| `width/2`, `Double(3)/Double(4)`, `width / 2`, `#!/usr/bin/env swift`, `(/)` pass through | (unchanged — these are the no-false-fail-close side) |
| unterminated `#/…` and a multi-line literal closing mid-line throw | n/a |

Every Swift snippet in those tests was checked against `xcrun swiftc -parse` (Swift 6.2.3) both as
written and as the scanner leaves it, so the grammar is measured rather than assumed. Two things
that came out of that and are encoded in the scanner: Swift honours `\/` inside extended
delimiters (`#/a\/#b/#` parses), and it closes a multi-line literal at the first unescaped
delimiter but then requires it to start its own line — so a mid-line `/#` is a file that does not
compile either way, and packaging refuses it rather than stripping around it.

**Re-validated on 9770a12e15**

- `pnpm format`, `pnpm lint`, `pnpm typecheck`, `pnpm check:affected --run` (including
  `check:fallow --base origin/main`), `pnpm check:xctest-selection`: all pass.
- `xcodebuild build-for-testing` on the packaged project: `** TEST BUILD SUCCEEDED **` for both
  `generic/platform=iOS Simulator` and `platform=macOS,arch=arm64`. Build only — no `xcodebuild test`
  and no device or simulator run.
- All 44 packaged Swift files pass `xcrun swiftc -parse`; both runtime-compiled recording scripts
  compile with `xcrun swiftc` (0 errors); re-running the scanner over its own output removes 0
  further comments.
- You were also right that the corpus has no such literals, so the size numbers are unchanged and
  the PR body still holds: `dist/apple/runner/` 555,907 B -> 484,009 B (-71,898 B, -12.9%), Swift
  441,196 B -> 369,298 B (-16.3%), 891 comments stripped. The compile evidence now rests on a
  scanner that handles the construct rather than on the corpus not containing it.

thymikee:
The regex corruption is fixed at 9770a12: the original failing literal now survives unchanged, and the new tests cover delimiters, escapes and unsupported bare-regex syntax. No remaining code findings. The reported packaged iOS/macOS builds pass; marking ready for human review while the remaining smoke check finishes.

thymikee:
Before this merges — I want to challenge the size of it, with measurements, because I think ~95% of the machinery is buying ~0.6% of the bytes.

I measured the shipped corpus (`apple/runner/**/*.swift`, 44 files):

| construct the scanner handles | occurrences in the corpus |
|---|---:|
| whole-line `//` comments | **1,351 lines / 109,378 B** |
| trailing `//` comments | ~652 B (crude upper bound) |
| block comments `/* */` | **0** |
| extended regex literals `#/…/#` | **0** |
| multi-line string literals `"""` | 4, in 2 files |
| multi-line *raw* literals `#"""` | **0** |

So **99.4% of the comment bytes are whole-line `//`**, and the constructs that motivated most of the 425-line scanner — nested block comments, extended regex literals, the bare-`/` expression-position heuristic with its 128-character lookbehind — cover cases that do not occur at all here.

A much smaller rule gets essentially the same saving: *delete lines whose trimmed text starts with `//`, while tracking only `"""` literals so a line inside one is never deleted.* That is one construct instead of six, roughly 30 lines instead of 425, and it never has to decide whether a `/` is a comment, a division or a regex — which is exactly where the corruption you caught lives. I checked the one hazard it would still have: zero whole-line `//` currently sit inside a multi-line literal.

What that trades away is real but small: the ~652 B of trailing comments stay, and doc comments on the same line as code stay. Call it 108.7 kB recovered instead of 109.4 kB.

The reason I think this matters more than the byte count: the current design can **fail a publish** on valid Swift. The bare-slash heuristic decides expression position from a 128-char lookbehind, and if it guesses wrong on some future file it throws rather than shipping. A rule that only ever deletes a line that is entirely a comment cannot corrupt code and cannot block a release — worst case it leaves a comment in.

Two smaller points, whichever way you go:
- If you keep the full scanner, consider making the bare-`/` case a **per-file bail-out** (ship that file verbatim, comments intact) instead of a hard packaging failure. Same safety, no publish-blocking false positive.
- The 30.9 kB download / 4.2 ms `--help` win is real and I am not arguing against doing this at all — only against hand-rolling six constructs' worth of Swift grammar at publish time to get the last 0.6%.

Happy to cut the smaller version if you agree; I did not want to rewrite an approved PR on my own initiative.

thymikee:
The current code verdict is unchanged, and checks are green. The smaller approach is worth considering, but tracking only triple-quoted strings is not enough to guarantee safety: a line starting with // can also be content inside a multiline regex literal. Before replacing the scanner, define the supported input and prove that other files are preserved unchanged. This is a design choice before merging, not a new failure in the reviewed head.

thymikee:
Withdrawing the risk half of my challenge above — the review's evidence beats my argument, and I checked the two load-bearing facts myself rather than taking them on trust.

**"It can fail a publish on valid Swift" was wrong.** `scripts/check-xctest-selection.ts` invokes the packager, and `check:xctest-selection` sits inside `check:tooling`, so the fail-closed throw lands at **PR time**, not at publish. A false positive is a red check on a branch, not a blocked release. That was the strongest point I had and it does not hold.

**The AST-equality check is better evidence than anything I offered.** Comparing normalised `-dump-parse` output across all 44 packaged files proves the transform removed only comments — no literal corruption, no eaten code, no statement-separator damage. My argument was that hand-rolled Swift lexing is risky in principle; that is an argument about priors, and a byte-identical AST diff over the whole corpus is a measurement. The measurement wins.

What survives is only the size/complexity observation — 425 lines near the 500-line ceiling, covering constructs (block comments, regex literals) that do not occur in this corpus. That is a real but weak argument against a design whose correctness is now empirically demonstrated, and not worth rewriting an approved PR over. I'd drop it.

One genuine interaction worth noting: blanking comment-only lines instead of deleting them, which the parity point requires, is also what would make a much simpler line-oriented rule viable, since it removes the need to reason about `//` in expression position at all. If this scanner ever does need to shrink under the ceiling, that is the direction — but as a later simplification against a green baseline, not as a change to this PR.

Both requested changes are in progress: blank-line parity with objective per-file line-count proof, and a permanent `swiftc -parse` gate over `dist/apple/runner/**` (I confirmed nothing in the repo consumes that directory today, so the silent-mis-lex gap is real).

thymikee:
Both requests are in `e0e4246da7`. Thanks for the AST diff — the scanner is untouched; only where a
removed line goes changed.

**1. Line parity — and it was worse than the comment pass.** A comment-only line is now emitted
empty, and a trailing comment takes no newline with it, so output line N is input line N. While
verifying that, the `#if AGENT_DEVICE_RUNNER_UNIT_TESTS` strip turned out to be the bigger offender,
and it predates this PR: it deleted 3,737 lines, moving everything below a block by up to 883 lines
in `RunnerTests+CommandExecution.swift`. Comment-only lines were 889, not ~1,356 — the other 467 sit
inside unit-test blocks that are removed first, so they never reach the comment pass. Parity for all
44 files is only reachable with both passes emptying instead of deleting, so both now do. Cost is
4,626 B (889 + 3,737), not ~1,356 B.

What it guarantees: for every one of the 44 shipped Swift files, line N of
`dist/apple/runner/<path>` is line N of `apple/runner/<path>`. A `runner.log` `file:line` is now
directly openable in the checkout.

Measured proof: `wc -l` identical for 44/44 files, and all 613 declarations the packaged files still
carry sit on their checkout line — `RunnerTests.swift` `func testCommand` at 245 in both,
`recording-overlay.swift`'s first `struct` at 14 in both. Regression tests cover a comment-only line
becoming blank with the count held (`scripts/__tests__/strip-swift-comments.test.ts`) and the
blanked unit-test block, line for line, in `src/__tests__/apple-runner-package-source.test.ts`.

**2. Permanent gate over the packaged output.** New `pnpm check:packaged-runner-swift`
(`scripts/check-packaged-runner-swift.ts`): packages into a throwaway root over a symlinked
`apple/`, asserts line parity per file, then runs `swiftc -parse` over each packaged `.swift`.
Registered in `CHECK_CATALOG`, wired into `pnpm check:tooling` right after `check:xctest-selection`,
selected by `check:affected` for `apple/runner/**` and both packaging scripts, and declared on the
macOS lane (`.github/workflows/macos.yml`) with a `run-gate` step — that is the only lane with a
Swift toolchain. Parity needs none and runs anywhere; the parse prints `parse skipped, no Swift
toolchain on this host` instead of failing, which is why the macOS lane owns it rather than the
ubuntu guards job. Seconds, no simulator. `pnpm check:gate-manifest` is green with it (53 checks).

I planted a violation for each half rather than trusting a green run: putting the comment-only-line
deletion back fails the gate on the shifted files by name, and a scanner that leaves a stray `/`
where a comment was fails `swiftc -parse` on 35 of 44 files with `file:line`.

**Size.** `dist/apple/runner/` 555,907 B -> 488,635 B (-67,272 B, -12.1%), Swift alone 441,196 B ->
373,924 B (-15.2%). Parity gives back 4,626 B of the 71,898 B the previous head saved.

`scripts/strip-swift-comments.mjs` is 431 lines, +6 — well inside the 500-line ceiling, so I left it
whole rather than adding a split to a file you had just verified byte-for-byte.

Packaged runner still builds on `e0e4246da7`: `** TEST BUILD SUCCEEDED **` for both
`generic/platform=iOS Simulator` and `platform=macOS,arch=arm64`. Build only, no `xcodebuild test`.

PR #2470, opened 2026-09-10
chore(layering): runner modules reach host-kit only through the runner host port

thymikee:
## Rule

`packages/platform-apple/src/runner/**` may not VALUE-import `@agent-device/host-kit/*` directly. The host-capability port at `runner/host.ts` is the only door: a host-kit symbol the runner needs is added to `AppleRunnerHost` there and bound to the real implementation in `core/runner-host.ts`. A type-only import (`import type … from '@agent-device/host-kit/…'`) is exempt everywhere, since it evaluates nothing. A pure constant that both the runner and another package need does not go in host-kit either — it belongs in a runner module already inside every façade closure (e.g. `runner/apple-runner-platform.ts`), imported directly from there.

This lands as:
- One paragraph added to the header doc of `runner/host.ts` stating the rule and its reason.
- A new layering rule, `scripts/layering/apple-runner-host-port-policy.ts` (R77 apple-runner-host-port), registered in `scripts/layering/check.ts`. Violation message: *"imports '\<spec>' directly. Reach host-kit through the runner host port (runner/host.ts, bound in core/runner-host.ts); a direct import grows the Apple façade eager closures (eager-closure-budgets)."*
- `scripts/layering/apple-runner-host-port-policy.test.ts`: a planted violation (every import form — named, namespace, dynamic, re-export, side-effect) goes red; a type-only import (in the runner subtree and on `host.ts` itself) stays green; imports outside the runner subtree and non-host-kit runner imports are untouched.

## Reason

`runner/**` sits in the eager import closure of seven Apple façade entries (`app-lifecycle-facade.ts`, `app-resolution-facade.ts`, `doctor-facade.ts`, `perf-facade.ts`, `physical-device-facade.ts`, `runner-operations-facade.ts`, `runner/index.ts`), each held at a fixed size (no growth against the merge-base) by `scripts/__tests__/eager-closure-budgets.ts`. A direct `@agent-device/host-kit/*` value import from a runner module adds every module on its own import path to all seven closures at once.

[...]

PR #2474, opened 2026-09-10
revert: carry one phase Deadline through the runner interfaces (#2473)

thymikee:
Reverts #2473 (`f4c8f3ddda`) to restore the iOS live lanes on main.

## Evidence

- main iOS: green at `3e89f821d6` (16:00), red at `f4c8f3ddda` (16:44, first run containing #2473) on `Run iOS Settings replay smoke test` (replay step 6 `wait` timed out after 3 attempts).
- #2473's own post-rebase run at 16:15 failed `Run fixture-backed iOS simulator E2E smoke` (`wait text "Agent Device Tester" 10000` timed out). #2470 and #2471, run at the same minute on the same base without #2473, passed.
- #2471 rebased onto `f4c8f3ddda` fails the same fixture smoke at 17:00; `fix/ios-record-video-host-busy-2170` fails at 16:52.
- Session log of the failing wait (artifacts of run 34503978665 / #2471 run at 17:00): `open` finished in 4.3 s; the wait's `ios.snapshot-source.prepare` reported `request canceled` at 5.2 s, the runner fallback logged `ios_runner_session_reuse`, `ios_runner_readiness_preflight`, `ios_runner_connect` ×2 and `retry` ×3, and the wait failed at 10.4 s. A later plain `snapshot` in the same session succeeded (6.2 s), so the runner itself was alive.

#2473 changed only `packages/platform-apple/src/runner/*` (the startup, adoption, reuse and build phases now spend one `RunnerPhaseBudget`). The live reproduction is not yet isolated to a line; reverting first, fixing forward with a reproduction on a booted simulator second.

No test changes are lost that matter to main: the cancellation matrix and the phase-budget interface come back with the fix-forward PR.

### DIFF
--- a/packages/platform-apple/src/runner/__tests__/runner-adoption.test.ts
+++ b/packages/platform-apple/src/runner/__tests__/runner-adoption.test.ts
@@ -13,10 +13,7 @@ import {
 } from '../runner-lease.ts';
 import { isIosRunnerDetachEnabled, tryAdoptRunnerSessionFromLease } from '../runner-adoption.ts';
 import { sendRunnerCommandOnce } from '../runner-transport.ts';
-import {
-  createRunnerPhaseBudget,
-  resolveExpectedRunnerCacheMetadata,
-} from '../runner-xctestrun.ts';
+import { resolveExpectedRunnerCacheMetadata } from '../runner-xctestrun.ts';
 import { appleRunnerTestHost } from '../test-host.ts';
 import { appleToolchainProbeResult } from './apple-toolchain-fixtures.ts';
 import { mkdtempForTestSync } from './tmp-dir.ts';
@@ -164,9 +161,7 @@ test('a request canceled during the fingerprint probe fails adoption instead of 
   });
 
   await expect(
-    tryAdoptRunnerSessionFromLease(simulator, {
-      budget: createRunnerPhaseBudget(undefined, request.signal),
-    }),
+    tryAdoptRunnerSessionFromLease(simulator, { signal: request.signal }),
   ).rejects.toSatisfy(isRequestCanceledError);
   expect(mockSendRunnerCommandOnce).not.toHaveBeenCalled();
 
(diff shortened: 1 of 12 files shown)


thymikee:
Ready at 4353e76. This restores the exact tree before #2473, and the iOS simulator replay and smoke lanes now pass along with the other checks. Please keep the phase-budget refactor out until a fix-forward change reproduces and resolves the live failure.

thymikee:
Withdrawing this revert. A local reproduction on a booted simulator did not flip between `f4c8f3ddda` (#2473) and its parent `fea7ca8a43`: the CI settings replay (`test … 01-settings.ad`, `--retries 0`) passed 4/4 on both, and an instrumented `open → click → wait` produced byte-identical request logs on both, served by a warm AX bridge in ~107 ms.

The CI evidence, re-read against the artifacts, does not support #2473 either:

- Every failing wait shows the same shape: `ios_runner_session_reuse` → connect ok in 7–59 ms → `ios_runner_readiness_preflight` (1 s) → a second `ios_runner_connect` that stalls 4.5–5.1 s → the wait's own 5 s / 10 s deadline cancels it (`captures: 1, readableCaptures: 0`). Every budget number in those logs (preflight 1000 ms, connect ~45 000 ms, command ~44 9xx ms) is byte-identical to a green run without #2473, and the green run shows the same second connect at 11 ms to 2.3 s, i.e. the same distribution inside the margin. The difference is runner latency on a cold host, not a budget or cancellation semantic.
- Two of the four "correlated" red runs (`fix/replay-deep-link-classifier-b`, `fix/ios-record-video-host-busy-2170`) failed a pure Swift XCTest step on a different image (Xcode 26.6 / iphonesimulator 26.5) with `ALERT_DEADLINE_EXCEEDED`; they cannot involve #2473 and I should not have counted them.
- #2473's own run failed the settings replay on attempt 1 and passed on attempt 2. This revert's green iOS run is one sample.

Code reading found no defect: the only semantic delta in #2473 is `requireRunnerPhaseRemainingMs` losing a numeric fallback that was unreachable at all three call sites.

The real problem is the readiness preflight abandoning a runner command that keeps the runner busy for ~5 s; filed separately with the artifact evidence. Closing without merging; #2473 stays on main.

PR #2483, opened 2026-09-11
fix(orientation): disclose an unconfirmed rotation instead of asserting it

### LINKED ISSUE #2481: orientation can assert a rotation the owner never reported
# `orientation` can assert a rotation the owner never reported

agent-device 0.21.0. Code-level report, with a device check.

## What the code does

`src/daemon/orientation-runtime.ts` (`executeSetOrientation`):

```ts
const orientation = result?.orientation ?? requestedRotation;
return { action: 'orientation', orientation, ...successText(`Rotated to ${orientation}`) };
```

When the owner reports no resulting orientation, the success claim comes from the request alone. A pinned test ("falls back to the requested rotation when the owner reports nothing") asserts exactly that.

## Device check (iPhone 15 sim, Xcode 26.6)

The claim is DEVICE-level and truthful on iOS: `orientation landscape-left` really rotates the device (`simctl` screenshot 2556x1179; back to `portrait` -> 1179x2556), and `packages/platform-apple/src/interactor.ts` verifies the runner-observed rotation and throws on a mismatch. A screenshot that stays portrait shows a portrait-only APP (`UISupportedInterfaceOrientations_iPhone`), not a failed rotation. iOS simulators never take the no-report branch, so this is defensive, not a live failure.

## Fix

Branch `fix/orientation-unconfirmed-claim` (fork `prateekranka/agent-device`): keep `orientation: requestedRotation` for compatibility, return `confirmed: false` plus a warning, and name the request in the message ("Rotation requested: portrait (unconfirmed)"). The owner-confirmed path is unchanged. Gate 130 files / 646 tests.

## Question

Prefer a hard failure (typed error) when the owner reports nothing, or the disclosed-warning shape?

prateekranka:
# Disclose an unconfirmed rotation instead of asserting it

Fixes #2481

## What changes

`executeSetOrientation` did:

```ts
const orientation = result?.orientation ?? requestedRotation;
return { action: 'orientation', orientation, ...successText(`Rotated to ${orientation}`) };
```

When the owner reported no resulting orientation, the response asserted `Rotated to <request>` from the request alone. A pinned test ("falls back to the requested rotation when the owner reports nothing") encoded that.

Now the requested rotation is kept for compatibility, but the claim is disclosed:

- `confirmed: false`
- `warning: "Requested portrait; the device owner reported no resulting orientation, so the rotation is unconfirmed."`
- `message: "Rotation requested: portrait (unconfirmed)"`

The owner-confirmed path is unchanged.

## Tests

`src/daemon/__tests__/orientation-runtime.test.ts`: the updated test fails on the base with `Received: "Rotated to portrait"` and passes after.

## Gates

`pnpm check:affected --run`: **130 files / 646 tests, all runnable checks passed.**

## Device-lane scope (honest note)

On an iPhone 15 simulator the Apple owner either reports the observed rotation or throws (`packages/platform-apple/src/interactor.ts` verifies the runner-reported orientation), so `result?.orientation` is never empty there and this branch is defensive. On device I confirmed that the existing claim is truthful for iOS: `orientation landscape-left` really rotates the device (`simctl` screenshot 2556x1179; back to `portrait` -> 1179x2556). An app that stays portrait is a portrait-only app, not a failed rotation. So: robustness for owners that report nothing, with unit coverage only.

## Open question

Prefer a hard failure (typed error) when the owner reports nothing, or this disclosed-warning shape?

### DIFF
--- a/packages/contracts/src/navigation.ts
+++ b/packages/contracts/src/navigation.ts
@@ -31,11 +31,20 @@ export type BackCommandResult = {
   settle?: SettleObservation;
 };
 
-/** `orientation` — `{ action: 'orientation', orientation, message: 'Rotated to <orientation>' }`. */
+/**
+ * `orientation` — `{ action: 'orientation', orientation, message: 'Rotated to <orientation>' }`.
+ *
+ * An owner that reports no resulting rotation is not evidence that the device
+ * rotated. That case keeps the requested `orientation` for compatibility, but
+ * discloses `confirmed: false` plus a `warning`, and names the request in
+ * `message` (`Rotation requested: <orientation> (unconfirmed)`).
+ */
 export type OrientationCommandResult = {
   action: 'orientation';
   orientation: DeviceRotation;
   message: string;
+  confirmed?: boolean;
+  warning?: string;
 };
 
 /** `app-switcher` — `{ action: 'app-switcher', message: 'Opened app switcher' }`. */
--- a/packages/session-journal/src/__tests__/session-event-action.test.ts
+++ b/packages/session-journal/src/__tests__/session-event-action.test.ts
@@ -192,6 +192,18 @@ test('structural action events preserve navigation, viewport, keyboard, and gest
 
   assert.equal(buildActionSummary(back), 'Went back using system navigation');
   assert.equal(buildActionSummary(orientation), 'Rotated to landscape-left');
+
+  // An owner that reported no resulting rotation keeps the requested value, so
+  // the recorded action must not claim the device rotated.
+  const unconfirmedOrientation = action('orientation', {
+    action: 'orientation',
+    orientation: 'portrait',
+    confirmed: false,
+    warning:
+      'Requested portrait; the device owner reported no resulting orientation, so the rotation is unconfirmed.',
+    message: 'Rotation requested: portrait (unconfirmed)',
+  });
+  assert.equal(buildActionSummary(unconfirmedOrientation), 'Requested portrait (unconfirmed)');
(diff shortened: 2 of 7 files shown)


thymikee:
The direct response is more honest, but the action journal still reconstructs “Rotated to portrait” from orientation alone, ignoring confirmed: false. Please carry the unconfirmed state through buildOrientationActionSummary in packages/session-journal and add a regression for the recorded action.

The public OrientationCommandResult and MCP output description also need the new confirmation and warning fields so clients can consume the distinction. A disclosed warning is reasonable for an owner whose contract permits no reported orientation; this does not need to become a hard failure.

prateekranka:
Addressed in `c80caba99`.

- `buildOrientationActionSummary` no longer reconstructs "Rotated to <orientation>" from the orientation field alone. With `confirmed: false` it records "Requested <orientation> (unconfirmed)"; the journal regression fails on the previous commit and passes here.
- `OrientationCommandResult` declares the optional `confirmed` and `warning` fields, with the doc comment stating the contract, and the MCP output schema advertises both. `command-tools-navigation-schemas.test.ts` keeps the schema and the dispatch shape in lockstep.
- The disclosed-warning shape stays, as you suggested: no hard failure.

Gate: `pnpm check:affected --run` — 332 files / 2158 tests, all runnable checks passed.

thymikee:
The journal and public contract gaps are addressed at c80caba. The journal regression fails without the fix, and all 72 focused tests passed. Ready for human review; GitHub currently reports no checks for this head, so CI readiness is still unverified.

PR #2486, opened 2026-09-11
fix(ios): refuse Simulator bridge trees that end at a web view's remote content (#2484)

### LINKED ISSUE #2484: iOS simulator: Safari web content is absent from snapshots (empty webview nodes)
## Summary

Snapshots of `com.apple.mobilesafari` on the iOS simulator contain Safari's chrome (address bar, Back, Page Menu, More) and three nested empty `webview` nodes, but nothing from the loaded page: no links, static text, text fields, or buttons. `--raw`, `-i`, `--scope WebView`, and `-d 20` all return the same empty web view.

## Versions

| Component | Version |
| --- | --- |
| agent-device | 0.21.0 (latest on npm) |
| Xcode | 26.2 (17C52) |
| macOS | 26.5.2 |
| Simulator | iPhone 17 Pro, iOS 26.2 |

## Repro

```bash
agent-device open com.apple.mobilesafari --platform ios --session safari-repro
xcrun simctl openurl booted https://example.com   # one heading, one paragraph, one link
sleep 4
agent-device snapshot --session safari-repro
agent-device snapshot --raw -d 20 --session safari-repro | grep -ciE 'Link|Example Domain|More information'   # 0
agent-device snapshot --scope WebView --session safari-repro
```

## Observed

```
Snapshot: 20 visible nodes
@e1 [application] "Safari"
  @e2 [other] "SafariWindow?View=Narrow&...&IsSecure=true&IsTabBarAvailable=false"
    @e3 [scroll-area] [scrollable]
      @e4 [other] "TabDocument?...&IsPageLoaded=true&WebViewProcessID=21704"
        @e5 [webview] "WebView"
          @e6 [webview] [scrollable]
            @e7 [webview]
            @e8 [other] "Vertical scroll bar, 1 page" [scrollable]
            @e9 [other] "Horizontal scroll bar, 1 page" [scrollable]
            @e10 [other] "Vertical scroll bar, 1 page" [scrollable]
            @e11 [other] "Horizontal scroll bar, 1 page" [scrollable]
      @e12 [other] "Horizontal scroll bar, 1 page" [scrollable]
      @e13 [other] "Horizontal scroll bar, 1 page" [scrollable]
    @e14 [other] "CapsuleViewController"
      @e15 [button] "Back"
      @e16 [other] "CapsuleNavigationBar?isSelected=true&isDistractionControlOverlayUp=false"
        @e17 [button] "Page Menu"
        @e18 [other] "Address"
        @e19 [button] "refresh"
      @e20 [button] "More"
```

`snapshot --raw -d 20` yields 70 nodes, none of them page content (the only `StaticText` is an empty `UIButtonLabel` in Safari's chrome). `snapshot --scope WebView` yields the same three empty `webview` nodes plus scroll bars. The simulator screen shows the page fully rendered, and the `TabDocument` identifier reports `IsPageLoaded=true`, so the page is there; only the accessibility tree is missing it. Same result on a page with a form (text fields and a submit button), and whether the URL was opened with `simctl openurl` or typed into the address bar.

## Expected

The page's accessibility elements under the web view (links, static text, text fields, buttons), as XCUITest exposes for WKWebView content, so refs can target them and `press` / `type` can act on the page.

## Notes

`snapshot --help` prints the iOS capability contract as `tree: ... deep-extension=no, known gap deep-extension; private-ax: ... deep-extension=yes`, so this may be the known gap. If so, is there a way to opt into the private-ax path for snapshots, or a documented flag? I only checked Safari, so I cannot say whether a WKWebView inside a regular app behaves the same.

Context: hit this while driving the simulator from a coding agent over MCP, where the agent had a screenshot of the page but no ref to act on.

thymikee:
Fixes #2484.

## What broke

0.21.0 made the host AX bridge (`apple/snapshot-bridge`) the snapshot source for local iOS Simulators (#2279). The bridge snapshots one process by pid. WebKit content — a Safari page, a `WKWebView` — lives in a separate WebContent process and reaches UIKit's tree as an `AXRemoteElement` whose children are in that other process. The bridge delivered that element as a leaf, and the route published the tree as if it were the screen: Safari chrome plus three empty `webview` nodes, no links, text, fields, or buttons, so no ref could reach the page. The in-guest XCTest runner, which served these captures before 0.21.0, resolves remote elements.

Raw bridge output of the reporter's screen ends at:

```
{"type":"WebView","role":"WKContentView", ...}
  {"type":"Other","role":"AXRemoteElement","subrole":"NSObject","rect":{"x":0,"y":0,"width":402,"height":874}}   ← no children; page lives in WebViewProcessID=9153
```

## The fix

The source refuses a tree that ends at content another process owns, the way it refuses a missing automation mode, and the existing route fallback does the rest:

- `snapshot-source/tree.ts` counts `AXRemoteElement` leaves that sit under a `WebView`-typed ancestor and whose frame reaches the reported viewport (`opaqueRemoteElements`). A zero-area or off-screen leaf hosts nothing the capture can miss and is published; a leaf that reports no frame is refused, because nothing proves it empty; a remote element with children means the reader crossed the boundary. Remote elements outside a web view are not classified (no capture has shown one), and a web view truncated away by the node or depth cap stays disclosed as truncation.
- `snapshot-source/adapter.ts` throws `unsupported` / `remote-content-boundary` (`details.remoteElements`) after the protocol checks and before the acquisition is built, so no depth hint is learned from the refused tree.
- `snapshot-route.ts` is unchanged: `fallbackAfterFailure` serves XCTest, discloses it in the snapshot warning (`Simulator AX snapshot unavailable (remote-content-boundary); used XCTest for this app generation.`), and disables the bridge for that app generation. Later captures in the generation go straight to XCTest (`circuit-disabled`), as 0.20.x did; a relaunch re-enables the bridge.

The guest ObjC reader is untouched: the class name it already delivers is the fact, and the host owns the decision (ADR 0004).

## Evidence

Live, iPhone 17 Pro simulator (iOS 26.2), Safari on a local page with a sign-in form, CLI from this branch:

| step | before (0.21.0) | after |
| --- | --- | --- |
| `snapshot -i` | chrome + 3 empty `webview` nodes | heading `Sign in`, `text-field "Email"`, `securetextfield "Password"`, `button "Log in"`; warning `remote-content-boundary`; `snapshotQuality.backend: tree` |
| second `snapshot -i` | same | same content; warning `circuit-disabled` |
| `fill @e8 "probe@example.com"` | no ref to fill | `Filled 17 chars`; next snapshot shows the value |

The fixture app's WebView lab (`agent-device-test-app:///webview`) shows the same shape: page heading, `Link "Jump to form"`, `Email address` label present, `backend: tree`.

Observed but not in scope: the first `fill` after the fallback once failed with `no text input found at the provided coordinates to clear` and succeeded on retry; that is the runner's text-entry path on a web field, unchanged by this PR.

## Gates

- Unit: `tree.test.ts` (counting rules: on-screen, frameless, zero-area, off-screen, no viewport, outside a web view, crossed), `adapter.test.ts` (refusal for both acquisition intents, no hint learned, recovers when content is native). The route's fallback + circuit are already pinned by the existing `snapshot-route.test.ts` cases; this change adds no route branch.
- Live: new smoke scenario `smoke:webview-remote-content` in the fixture-backed iOS E2E lane (behavior `webview-remote-content`) — opens the WebView lab, `wait`s 20 s for a page-only landmark (the `Jump to form` link; the heading equals the document title, which the web view node carries before the body exists), asserts the semantic `Link` node and the `Email address` label, and that the snapshot discloses the XCTest fallback (`remote-content-boundary` or, once the circuit is open, `circuit-disabled`). No WebView went through the bridge route in any gate before.
- `snapshotNodes` moved to `live-assertions.ts` and shared with the depth-frontier scenario.
- ADR 0004 amended; bridge README gains a "Remote content" section; CHANGELOG entry.

## Review

Two adversarial review passes (core mechanics; gates and docs). Applied: scope the detector to viewport-visible, web-hosted leaves instead of any remote leaf in the tree (an off-screen prewarm or analytics `WKWebView` would otherwise have disabled the bridge for the run); state the exact frameless rule and drop an unverified "extension-hosted view" claim from ADR/README; write the per-generation circuit trade-off into ADR 0004; gate the live scenario on page-only content with a 20 s budget; drop a route test that only re-stated an existing one; reuse the kernel rect helpers. Declined: re-asking the bridge per capture instead of opening the circuit (it would charge a refused round trip to every `wait` poll on a web screen).

## Follow-ups (not here)

Scoped and ranked in #2492 after a design consultation (content-class gate, probe opens the circuit, help text, bounded WebContent spike, truncation policy).

- Bridge-native remote content: read the remote element's pid and graft a second `snapshotForProcess:` for WebContent. Needs a spike; unknown whether WebContent serves AX to a host remote-access client.
- The launch-observation probe reads the full tree, so on a web-first app it answers `unobservable` and `open` pays the fixed 300 ms settle plus one refused bridge capture before the first fallback; the probe does not open the circuit itself.
- `open <http url>` on a simulator with no app opens the URL but does not bind Safari to the session (`platform-runtime-open-target.ts` gates that fallback to physical devices), so a following `snapshot` answers `SESSION_NOT_FOUND`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,15 @@
 
 ## Unreleased
 
+- Fixed: iOS Simulator snapshots of Safari and of apps with a `WKWebView` stopped showing the
+  page in 0.21.0 — chrome plus empty `[webview]` nodes, no links, text, or form fields, so no ref
+  could reach the page (#2484). The host AX bridge that 0.21.0 made the Simulator's snapshot source
+  reads one process, and WebKit content lives in another; the bridge delivered the boundary as
+  an `AXRemoteElement` leaf and the tree was published as if that were the screen. The source now
+  refuses a tree whose on-screen web view ends at that leaf (`remote-content-boundary`) and the
+  route serves XCTest, which resolves remote elements, for the rest of that app generation — the
+  same path 0.20.x used. The snapshot discloses the switch through its warning, and a relaunch
+  re-enables the bridge.
 - Fixed: JUnit reports remain readable when replay results contain characters forbidden by XML 1.0,
   replacing them with U+FFFD while preserving legal Unicode and whitespace. Original suite values
   remain available in JSON and other reporters.
--- a/apple/snapshot-bridge/README.md
+++ b/apple/snapshot-bridge/README.md
@@ -48,6 +48,16 @@ resolution. Secondary owners such as the return-to-app status-bar control do
 not replace the native primary owner. The route's generation circuit remains
 disabled after fallback until that app relaunches.
 
+## Remote content
+
+The reader snapshots one process. A WebKit page — Safari's or a `WKWebView`'s —
+lives in a WebContent process and appears in that tree as an `AXRemoteElement`
+leaf under the web view. The guest returns the leaf as delivered; the host
+refuses a tree in which such a leaf sits under a web view and reaches the
+viewport, or reports no frame (`remote-content-boundary`), and routes that app
+generation to XCTest, which resolves remote elements (#2484). A zero-area or
+off-screen leaf is published: it hosts nothing the capture can miss.
+
 ## Bounded depth recovery
 
 A healthy capture uses one native request. If native acquisition rejects it,
--- a/docs/adr/0004-ios-snapshot-backend-strategy.md
+++ b/docs/adr/0004-ios-snapshot-backend-strategy.md
@@ -42,6 +42,22 @@ fallback when bridge acquisition or presentation fails. Disable the bridge for t
 after fallback; a new app generation re-enables it. Physical devices, providers, custom-action
 captures, and interactions remain on their existing owners.
 
+A WebKit page — Safari's, or a `WKWebView`'s — lives in a WebContent process and reaches UIKit's
+tree as an `AXRemoteElement` under the web view, with its children in that other process. The
+bridge reads one process, so it delivers the element as a leaf. The source refuses a tree in which
+such a leaf sits under a `WebView`-typed ancestor and reaches the viewport (`remote-content-boundary`)
+instead of publishing a screen without its page: refs issued from it would target the host views
+around the page rather than the page. A leaf whose frame is zero-area or off screen hosts nothing
+the capture can miss and is published; one that reports no frame is refused, because nothing proves
+it empty. Remote elements outside a web view are not classified — no capture has shown one — and a
+web view truncated away by the node or depth cap stays disclosed as truncation. XCTest resolves
+remote elements, so the fallback serves the page (#2484).
+
+The refusal opens the generation circuit like any other bridge failure, so a hybrid app that showed
+one web screen takes XCTest for its remaining native screens until it relaunches — the 0.20.x path
+for every screen. Re-asking the bridge per capture would instead charge a refused bridge round trip
+to every `wait` poll on the web screen; the circuit keeps that cost to one capture per generation.
+
 Keep the two public snapshot strategies explicit:
 
 - **Regular visible strategy**: use recursive XCTest snapshots, emit the effective user-visible
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "agent-device",
-  "version": "0.21.0",
+  "version": "0.21.1",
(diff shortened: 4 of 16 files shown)


thymikee:
No findings at 0e87812. The refusal uses the existing source and fallback path, and the reported Safari and WebView runs cover the missing page content and field interaction. Ready for human review; the iOS smoke check is still running, with the other checks passing.

PR #2487, opened 2026-09-11
fix(android): retire completed recording evidence after pid reuse

### LINKED ISSUE #2476: Android recording start refuses completed manifest after PID reuse
## Problem
With agent-device0.20.10 on an Android emulator reused across app workspaces, `record start` fails with `Android screenrecord completed evidence cannot be safely retired`.

## Evidence
A recording completed normally and its remote video was removed. The emulator retained `/sdcard/agent-device-recording-active.json` with completion metadata and a chunk naming pid4004/startTime3766. After emulator reuse, `/proc/4004/stat` identifies a different binder process with startTime5432; the recorded remote artifact no longer exists. The previous local recording file is valid and saved. Subsequent record-start calls all fail; no recorder is alive for the old identity.

## Cause
The Android host transport inspect correctly returns ownership-lost for the reused PID. Completed-manifest retirement in runtime2 accepts only missing, so it treats proven identity replacement like an uncertain live recorder. The new unrelated process must not be killed.

## Suggested direction
Distinguish proven original-process termination/PID reuse from unreadable or uncertain identity. Retire only completed metadata whose recorder identity is proven gone, without signaling the new process. Keep conservative behavior for live, uncertain and incomplete recordings. Include an emulator-reuse regression test.

thymikee:
## Summary

On a reused emulator the pid a completed Android recording named can be reassigned. The transport
proves that with `ownership-lost`, but completed-evidence retirement and reattach accepted only
`missing`, so they read proven termination as an uncertain live recorder: every later `record start`
died on `Android screenrecord completed evidence cannot be safely retired`, and `record stop` could
not return an already-finalized recording. Closes #2476.

The four ownership observations are now classified once, in the contract that declares them
(`provesAndroidScreenRecordTermination`), and the completed-evidence decisions ask that question
instead of comparing to `missing`. Retirement still refuses a live or unreadable recorder; it removes
device-side files only and never signals a pid it proved is not its own. A recovered recording whose
recorder was already proven gone now discloses that its MP4 may be truncated. 11 files, two commits.

Deferred, matching the issue's conservatism instruction for live and incomplete recordings:
`stopChunk` and `waitForStopped` still refuse to read `ownership-lost` as stopped, so an open
recording whose recorder died and whose pid was reused needs its own decision — as does a typed
reason for this refusal, which surfaces as `UNKNOWN` today.

## Validation

Tested `204e447fec4` on Node 26.8.1.

- Mapping `ownership-lost` back to "not gone" fails the new start test, the new provider scenario and
  the truncation-warning test; the alive-recorder test passes with and without the fix.
- The transport test attributes each producer: reassigned executable, foreign remote path, exited
  task with no command line.
- 72 unit tests and 7 provider scenarios pass; the reuse case runs end to end against a fake `adb`.
- `pnpm check:affected --run` passes every runnable gate: 1,551 tests / 253 files plus format, lint,
  types, layering, Fallow and build. GitHub CI is pending.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,14 @@
 
 ## Unreleased
 
+- Fixed: Android `record start` no longer refuses to begin after a reused emulator reassigned the
+  previous recorder's pid. A completed recording's native marker is retired only once its recorder is
+  proven gone, but only an absent pid counted as proof — a pid that now names an unrelated process,
+  or a recorder that exited and waits to be reaped, did not. `record start` then failed every later
+  attempt with `Android screenrecord completed evidence cannot be safely retired`, and `record stop`
+  could not return an already-finalized recording. Proven termination now retires the marker and
+  returns the stored completion; a live or unreadable recorder still blocks, and the unrelated
+  process is never signalled (#2476).
 - Fixed: iOS Simulator snapshots of Safari and of apps with a `WKWebView` stopped showing the
   page in 0.21.0 — chrome plus empty `[webview]` nodes, no links, text, or form fields, so no ref
   could reach the page (#2484). The host AX bridge that 0.21.0 made the Simulator's snapshot source
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,9 @@
   attempt with `Android screenrecord completed evidence cannot be safely retired`, and `record stop`
   could not return an already-finalized recording. Proven termination now retires the marker and
   returns the stored completion; a live or unreadable recorder still blocks, and the unrelated
-  process is never signalled (#2476).
+  process is never signalled. A reused pid that runs a replacement `screenrecord` on the same
+  remote path proves the old recorder gone but not that the path is free, so that marker and
+  artifact are retained until the replacement ends, and neither is signalled (#2476).
 - Fixed: iOS Simulator snapshots of Safari and of apps with a `WKWebView` stopped showing the
   page in 0.21.0 — chrome plus empty `[webview]` nodes, no links, text, or form fields, so no ref
   could reach the page (#2484). The host AX bridge that 0.21.0 made the Simulator's snapshot source
--- a/packages/contracts/src/screen-recording-runtime-host.test.ts
+++ b/packages/contracts/src/screen-recording-runtime-host.test.ts
@@ -1,0 +1,19 @@
+import { expect, test } from 'vitest';
+import {
+  type AndroidScreenRecordingProcessOwnership,
+  provesAndroidScreenRecordTermination,
+} from './screen-recording-runtime-host.ts';
+
+const TERMINATION_PROOF: readonly (readonly [AndroidScreenRecordingProcessOwnership, boolean])[] = [
+  ['missing', true],
+  ['ownership-lost', true],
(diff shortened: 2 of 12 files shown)


thymikee:
Completed-evidence retirement still has an unsafe case at 204e447: ownership-lost also covers a replacement screenrecord using the same remote path with a different start time. Treating that as permission to call cleanupChunks deletes the new recording’s active MP4. Please distinguish that foreign-writer case and retain its file and marker; add a regression using the same executable/path with a changed start time, alongside the unrelated-process reuse case. Proving the old recorder ended is not proof that the path is unused.

The provider tests use fake adb, so a real emulator retirement/recovery run is still needed. The branch also now conflicts with main; its existing CI checks are green.

thymikee:
The rebase at 0bf1b84 resolves the conflict but leaves the reviewed production changes unchanged, so the foreign-writer cleanup finding and emulator validation requirement remain open. The failing iOS smoke check reports RUNNER_BUSY at the WebView page wait, also seen on unrelated PRs; it appears unrelated to this Android change.

thymikee:
Addressed in 9c573650a0.

**Foreign-writer case.** The transport now classifies a pid that runs `screenrecord` on the recorded remote path with a different start time as `foreign-writer`, distinct from `ownership-lost`. The contract answers two questions: `provesAndroidScreenRecordTermination` (true for `missing`, `ownership-lost`, `foreign-writer`) keeps recovery and the truncation warning reading proven termination, and the new `provesAndroidScreenRecordPathUnclaimed` (true only for `missing`, `ownership-lost`) is what completed-evidence retirement asks. A foreign writer therefore retains the marker and the artifact and is never signalled; `record start` fails with `…names an artifact another recorder is writing; it is retained until that recorder ends` until it does. `stop`/`waitForStopped` refuse it the same way as `ownership-lost` (open-recording handling stays deferred as before).

Regressions added alongside the unrelated-process case: contract proof table, transport classification (`same executable/path, changed start time`), start reconciliation (nothing removed, nothing launched), completed-evidence reattach, and a provider scenario with a `replacement-recorder` reused pid.

**Emulator validation** (Pixel_9_Pro_XL, API 36, real adb, this build):
- start/stop, then a second start retiring the completed marker on an absent pid: ok.
- marker rewritten to pid 1 (`/system/bin/init`) with a foreign start time: start retired the artifact and marker, launched, pid 1 untouched.
- a live `screenrecord` started on the recorded path, marker rewritten to its pid with a different start time: start refused with the new message; marker, artifact (3232 B, still growing) and pid all retained. After `kill -2`, the next start retired both and proceeded.
- crash recovery: `kill -9` the daemon mid-recording; `record stop` reattached the manifest-owned recorder, stopped it, and pulled a playable MP4.

The iOS smoke failure is the `smoke:webview-remote-content` watchdog on main since #2486; the job was re-run.

thymikee:
The foreign-writer distinction at 9c57365 resolves the cleanup finding: a replacement recorder keeps its artifact and marker and is never signalled. The reported emulator runs cover unrelated-PID retirement, replacement-writer refusal and recovery after daemon death. Ready for human review; the remaining iOS smoke failure is RUNNER_BUSY at the WebView wait and appears unrelated to this Android change.

PR #2488, opened 2026-09-11
chore(react-devtools): pin agent-react-devtools 0.5.0 for React Native 0.87+

thymikee:
## Summary

Pins `agent-react-devtools` to 0.5.0, which restores React DevTools attachment on React Native 0.87+ ([callstackincubator/agent-react-devtools#58](https://github.com/callstackincubator/agent-react-devtools/pull/58)). React Native 0.87 removed the built-in DevTools websocket ([facebook/react-native#56897](https://github.com/facebook/react-native/pull/56897)), so `react-devtools wait --connected` could never succeed on those apps.

The `help react-devtools` topic now states the one-time `agent-react-devtools init` requirement (Metro wrapper plus an entry import, then rebundle) and that an empty observation with 0 apps connected is not a pass.

## Verification

End to end on a bare `react-native@0.87.1` app (community CLI template, Metro on 8081), driven through `agent-device react-devtools` with this pin:

| scenario | 0.5.0 (this PR) | upstream main (#59, unreleased) |
| --- | --- | --- |
| before `init`: `wait --connected --timeout 15` | times out, exit 1 | same |
| before `init`: `errors` | `No components with errors or warnings`, exit 0 | fails, exit 1 |
| after `init` + reload: `wait --connected` | attached, 149 components | same |
| `count`, `errors`, `get tree` | real data | same |
| React Native DevTools attaches alongside | count 298, `find App --exact` returns 2 | count stays 149, 1 hit |
| app killed, then `errors` | exit 0 | `No React app is attached ... app disconnected 3s ago`, exit 1 |
| `uninit` | reverts `index.js` and `metro.config.js` | same |

The help text in this PR tells agents to run `init` and `uninit`, and not to read the first-column empty results as a pass. The right-hand column arrives with the next pin bump.

- `cli-react-devtools`, `cli-help`, `cli-help-topics`, and `command-doc-coverage` tests pass; lint and typecheck clean.

## Not in this PR

The vacuous pass from #2430 (`errors` exits 0 with nothing attached) is a daemon-side defect, fixed upstream in [callstackincubator/agent-react-devtools#59](https://github.com/callstackincubator/agent-react-devtools/pull/59), pending release. Another pin bump follows once it ships.

Refs #2430

## Expo upgrade attempted and dropped

Upgrading the test-app to a React Native 0.87 Expo SDK was tried in this branch so the scenarios could run on our own app. Latest stable SDK 57 ships 0.86.3, which still has the legacy DevTools websocket. SDK 58 preview ships 0.87.1 but does not build on Xcode 26.2 / Swift 6.2.3: `expo-modules-jsi` fails first on the constructor annotations tracked in expo/expo#49667, then on Swift 6 `sending` data-race errors in `JavaScriptRuntime.swift` under `NonisolatedNonsendingByDefault`. The file is identical to Expo main, so this is a toolchain-versus-preview mismatch. The test-app stays on SDK 56 for now.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "agent-device",
-  "version": "0.21.0",
+  "version": "0.21.1",
   "description": "Mobile app automation and verification for AI coding agents. CLI, MCP server, and typed Node.js API for iOS, Android, HarmonyOS, TV, web, macOS, and Linux.",
   "mcpName": "io.github.callstack/agent-device",
   "license": "MIT",
--- a/server.json
+++ b/server.json
@@ -7,12 +7,12 @@
     "url": "https://github.com/callstack/agent-device",
     "source": "github"
   },
-  "version": "0.21.0",
+  "version": "0.21.1",
   "packages": [
     {
       "registryType": "npm",
       "identifier": "agent-device",
-      "version": "0.21.0",
+      "version": "0.21.1",
       "transport": {
         "type": "stdio"
       },
--- a/src/__tests__/cli-react-devtools.test.ts
+++ b/src/__tests__/cli-react-devtools.test.ts
@@ -43,12 +43,12 @@ afterEach(() => {
 });
 
 test('react-devtools passthrough pins agent-react-devtools package version', () => {
-  assert.equal(AGENT_REACT_DEVTOOLS_PACKAGE, 'agent-react-devtools@0.4.0');
+  assert.equal(AGENT_REACT_DEVTOOLS_PACKAGE, 'agent-react-devtools@0.5.0');
   assert.deepEqual(buildReactDevtoolsNpmExecArgs(['get', 'tree', '--depth', '3']), [
     'exec',
     '--yes',
     '--package',
-    'agent-react-devtools@0.4.0',
+    'agent-react-devtools@0.5.0',
     '--',
     'agent-react-devtools',
     'get',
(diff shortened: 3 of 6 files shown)


thymikee:
The pin is appropriate, but the new help needs two corrections. React Native setup also requires agent-react-devtools as an app dependency: init writes Metro and entry imports, while this wrapper’s npm exec package is temporary and does not install it into the app. Document that prerequisite before init and rebundling.

The claim that observation commands fail with no attached app is not true of the pinned 0.5.0 release; that fix is in the later upstream #59. Until it is released and pinned, tell users to verify an attached app rather than relying on errors or an empty result.

The iOS smoke failure is RUNNER_BUSY at the WebView page wait, also failing on unrelated PRs. It appears unrelated to this pin.

thymikee:
Both corrections applied in the help topic: the app must have agent-react-devtools as a dev dependency before init, since the npm exec package here is temporary; and agents are told to verify an attached app with status or wait --connected first, because on 0.5.0 count, errors, and get tree return empty results with 0 apps connected.

thymikee:
The corrected help at d14189a now covers the app dependency, init/rebundle and attachment check required by the pinned release. No code findings; the reported React Native 0.87 run covers the attachment change. The 57 kB size increase compares against newer main packaging that strips Apple-runner comments, rather than showing growth from this pin. Refresh the branch onto that packaging change and rerun the affected and Size checks before merge; ready for human review.

thymikee:
The iOS smoke job has now completed: it fails with RUNNER_BUSY while waiting for the WebView page link, matching the separately reported main failure. This appears unrelated to the DevTools pin; the code verdict and ready-for-human label are unchanged.

PR #2490, opened 2026-09-11
fix(test): keep the working directory literal during glob discovery

PLASMA-FR:
## Summary

Completes caller-side suite discovery established in #1810. From a checkout named `[workspace]`, `agent-device test './*.ad'` currently sends no sources; when the directory name matches a sibling, it can select that sibling's scripts instead. Missing literal files also lose their `INVALID_ARGS` error.

Pass the authored glob and literal working directory separately to Node's glob API. Keep home expansion, explicit-file precedence, directory DFS, and YAML-first glob ordering. Six files: discovery, behavioral and CLI regressions, command help, user docs, and changelog.

## Validation

Tested `cdf39a5bd79683c922362ae2ded18f2946347ea9` with Node 24.13.1 and pnpm 11.17.0, using `AGENT_DEVICE_VITEST_MAX_WORKERS=1`.

- Six regression cases failed before the fix; all 68 focused discovery/source-bundle/writer/CLI/startup tests pass.
- `pnpm build`, `pnpm test:maestro-compat` (363 tests), and `pnpm check:command-docs` (12 tests) pass.
- Built CLI against an owned HTTP fixture: all nine native/Maestro/missing-file cases pass across plain, bracketed, and braced working directories. Valid requests contain the exact source/include bundles; missing files fail before transport. The upstream build fails six of these cases.
- `pnpm check:affected --run` passes every runnable gate: 2,823 tests across 385 files, plus formatting, lint, types, layering, Fallow, and build. CI is pending.

No device execution was used: the changed boundary is caller filesystem discovery and request preparation. Upstream CI remains authoritative.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+- Fixed: `test` expands relative globs from the caller's literal working directory, so directory
+  names containing glob characters no longer cause missing suites or select a different directory.
+  Missing non-glob inputs also retain their not-found error in these directories.
 - Fixed: JUnit reports remain readable when replay results contain characters forbidden by XML 1.0,
   replacing them with U+FFFD while preserving legal Unicode and whitespace. Original suite values
   remain available in JSON and other reporters.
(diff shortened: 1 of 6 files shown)


thymikee:
No findings at cdf39a5. Passing the authored pattern and literal cwd separately fixes the bracketed-directory case without changing explicit-file precedence or suite ordering, and the CLI regressions cover the actual source bundles. Ready for human review; the failing iOS check reports RUNNER_BUSY at the WebView page wait, also seen on unrelated PRs, rather than in source discovery.

PR #2493, opened 2026-09-11
fix(wait): poll through a retriable runner refusal instead of surrendering the budget

thymikee:
## Why

The GitHub **iOS** workflow's Smoke Tests job has failed on `main` since the merge of #2486, and on any PR rebased onto it. The failing step is the new `smoke:webview-remote-content` scenario:

```
✖ live iOS simulator fixture E2E
  AssertionError: step: wait for the WebView page to expose its link
  code: RUNNER_BUSY
```

The wait was given a 20 000 ms budget and failed after 288 ms.

## What actually happened

Reading the run's request logs and step history (main run 34592935964) gives the full chain:

| Step | Result |
| --- | --- |
| `wait for deep-link destination before inspecting system UI` | failed after 4 065 ms |
| `inspect delayed deep-link system alert` (`alert get`) | failed after 10 616 ms, `main thread execution timed out` |
| `wait for the WebView page to expose its link` | failed after 1 613 ms, `RUNNER_BUSY` |

Three separate defects stack up.

**The scenario asked the runner a question it should never have asked.** It reuses `acceptDeepLinkConfirmationIfPresent`, whose readiness landmark was hard-coded to the Automation lab's text. Off that route the landmark can never match, so the helper always fell through to its `alert get` probe. An XCTest alert query against a live `WKWebView` screen exceeds the runner's 30 s main-thread execution watchdog, which abandons the main-thread work and leaves the runner refusing every later command as `RUNNER_BUSY`. I reproduced this exactly on a local simulator: the same `alert get` takes 10.1 s and fails with `runner_main_thread_execution_timeout`. The `smoke:regular-visible-depth-frontier` scenario carried the same landmark mismatch and paid the same doomed probe.

**A budgeted `wait` surrendered its budget to a transient condition.** `RUNNER_BUSY` means an earlier command's abandoned work is still draining; the runner's own hint says to wait a few seconds and retry. A polling wait is that retry loop, and it quit on the first refusal.

**The retriability classification was missing on the path the failure took.** A runner error recovered from the lifecycle journal after a lost transport response was built with a bare `toAppErrorCode`, so `RUNNER_BUSY` reached callers as a `RUNNER_BUSY` wire code with no `retriable` flag. The live-response path publishes the identical condition as `COMMAND_FAILED` with `details.runnerErrorCode` and `retriable: true`. The repo's own policy test (`retriability cannot be claimed for a code outside the enumeration`) says `RUNNER_BUSY` must not be a wire code at all.

## What changed

- **`acceptDeepLinkConfirmationIfPresent` takes the route's own landmark.** Each caller passes the `wait` predicate for the screen it deep-linked into, so the alert probe runs only when the destination genuinely did not arrive. The WebView lab uses its native `close-webview-lab` chrome, which renders with the route and before the page.
- **`wait` rides out a retriable poll failure.** A poll whose failure its producer marked retriable is treated like an unreadable capture: the wait keeps polling to its deadline, records the poll as `retriable` in its timeout evidence, and surfaces the refusal only when no readable capture ever completed. `RUNNER_WEDGED` is not retriable and still ends the wait at once.
- **One classifier reads a runner-reported error code.** `classifyRunnerReportedError` in `runner-contract.ts` now backs both the live-response path and the journal-recovery path, so the same runner condition cannot surface under two codes with only one of them retriable.

## Live validation

On a booted iPhone 17 Pro simulator with the Release fixture app, driving the scenario's exact CLI steps against this branch's build:

| Step | Before | After |
| --- | --- | --- |
| deep-link destination landmark | 2 670 ms, failed | 389 ms, resolved |
| `alert get` probe | 10 100 ms, wedged the runner | not issued |
| `wait text "Jump to form" 20000` | RUNNER_BUSY | 81 ms, resolved |

The snapshot still carries `Link "Jump to form"`, the `Email address` field label, and the `Simulator AX snapshot unavailable (remote-content-boundary); used XCTest for this app generation.` warning, so #2484's behavior is unchanged.

Driving the old sequence first reproduces the wedge, and the fixed wait then polls its full 20 s in `retriable` polls instead of failing instantly, which is the ride-out working on a real refusal.

## Checks

`pnpm lint`, `pnpm check:fallow`, `tsc --noEmit`, and the touched Vitest suites pass. `packages/platform-apple/src/runner/__tests__/runner-client.test.ts` times out on one unrelated spawn test under parallel load and passes when run alone; that is the known contention flake.

## Note on a separate local failure

Running the whole smoke suite locally fails earlier, in `smoke:automation-input`, at `get text id="automation-event-name"`. I confirmed this reproduces on pristine `origin/main` in the same environment, so it is not from this change. The `get` resolves against the session's cached snapshot in 5 ms without capturing, and locally that snapshot predates the event render. CI passes that scenario. Worth a separate issue.

Closes the iOS Smoke regression from #2486.


thymikee:
A wait that spends its entire budget on retriable refusals loses the new timeout evidence at 76e5101. waitTimeoutError calls rethrowIfNeverReadable, which throws the last runner error without attaching captures, waitedMs or polls. The common all-RUNNER_BUSY case therefore cannot show where its budget went, contrary to the documented behavior. Preserve the runner code and retry details while adding the bounded wait evidence, and assert that an exhausted all-retriable wait reports its retriable polls. The shared classifier and cancellation handling otherwise look sound; iOS smoke is still running.

thymikee:
The merge from main at aeebb3d leaves the reviewed wait code unchanged, so the missing poll-evidence finding remains open. The latest iOS smoke still fails at the WebView page wait. Its error is now correctly classified as COMMAND_FAILED with runnerErrorCode RUNNER_BUSY and retriable: true, but it still carries no wait timeline. This is the scenario the PR aims to repair, so the failure cannot be dismissed as unrelated: preserve the exhaustion evidence and diagnose why this run still cannot observe the page.

thymikee:
Fixed in 056d014 — the finding was right, and taking it holistically turned up a mirror gap next to it.

A wait can exhaust its budget two ways, and both dropped half the story:

- **Polled away between refusals** (no deadline): threw the last refusal raw, so no `captures`, `waitedMs` or `polls`. This is the case you flagged.
- **Deadline cancelled the final poll**: reported a generic `wait_capture_stalled` and dropped the runner code and retry details instead.

Both now raise one error that keeps the producer's code, message, hint and retry details **and** carries the wait's evidence, with `reason: wait_capture_stalled` and the original as its `cause`. A content verdict is still preserved exactly as its producer wrote it, since it already describes the capture it came from; whether it outranks the stall verdict stays the caller's policy, so `wait absent` is unchanged and its pinned expectations still hold.

Live-verified against a genuinely wedged simulator runner (`alert get` on a loaded WKWebView page, then the wait):

```
code: COMMAND_FAILED | retriable: true
message: The iOS runner is still finishing a previous command that exceeded its exec…
runnerErrorCode: RUNNER_BUSY | reason: wait_capture_stalled
captures: 6  readable: 0  waitedMs: 8041  timeoutMs: 8000
polls: retriable,retriable,retriable,retriable,retriable,deadline
```

Two unit tests assert each exhaustion shape, plus one that a content verdict is still returned untouched.

On the smoke failure you were right not to let me wave it off. I diagnosed it rather than retrying: the WebView lab landmark had a 2.5s budget inherited from the Automation lab, the lab rendered fine but was not in the tree that fast (the failure screenshot shows it fully rendered), so the helper fell through to its `alert get` probe, which is what wedges the runner. Measured on a freshly created simulator: with the confirmation alert up the probe is correct and costs 1.6s, because the alert blocks the route and there is no web view to query; with no alert the landmark resolves in 0.1-1.7s. So overshooting the budget is free and undershooting costs the scenario — hence 15s in 5276cbf. Whole scenario on a cold simulator: landmark 400ms, page wait 613ms, snapshot keeps the link, the field label and the fallback warning.

thymikee:
No remaining findings at 056d014. All-retriable wait exhaustion now preserves the original refusal and includes poll/capture counts and elapsed time, including when the last capture reaches the deadline. The regression tests and recorded simulator run cover the prior blocker. Ready for human review; current-head CI is still running.

PR #2495, opened 2026-09-11
fix(browserstack): carry the full provider-allocation config over the lease_allocate envelope (#2494)

### LINKED ISSUE #2494: BrowserStack: --provider-project, --provider-build, --provider-session-name not forwarded to daemon in lease_allocate RPC
## Summary

The `--provider-project`, `--provider-build`, and `--provider-session-name` CLI flags are parsed and stored in the local connection profile, but are **never forwarded to the daemon** in the `lease_allocate` JSON-RPC request. The daemon's BrowserStack `prepareSession` callback reads these from `req.flags.providerProject` etc., but the fields are absent — so all BrowserStack sessions land in "Untitled Project" / "Untitled Build" with an empty session name.

Affects **v0.20.10** and **v0.21.0** (confirmed by source reading).

## Root Cause

Two functions drop the provider metadata fields:

### 1. CLI → daemon RPC builder (`daemon-client-lifecycle.js`, `qt()` function)

When building the `lease_allocate` RPC params, only `providerApp` is forwarded:

```js
// qt() in daemon-client-lifecycle.js
{
  ...Xt(e, e.command, t),
  ...e.command === 'lease_allocate' && typeof e.flags?.providerApp === 'string'
    ? { providerApp: e.flags.providerApp }
    : {}
}
```

`providerProject`, `providerBuild`, and `providerSessionName` are **not included** in the RPC params sent to the daemon.

### 2. Daemon request constructor (`internal/daemon.js`, `Pi()` function)

When reconstructing the request object from the RPC params, only `providerApp` is placed into `flags`:

```js
// Pi() in internal/daemon.js
flags: e === 'lease_allocate'
  ? { providerApp: q(t, 'providerApp') }
  : undefined
```

So even if the CLI sent them, the daemon would drop them on the receiving end.

## Reproduction

```bash
export BROWSERSTACK_USERNAME="..."
export BROWSERSTACK_ACCESS_KEY="..."

  --platform ios --device "iPhone 15" --provider-os-version 17 \
  --provider-app "bs://..." \
  --provider-project "MyProject" \
  --provider-build "Build-2026-09-11" \
  --provider-session-name "smoke — iOS" \
  --session test-s1

agent-device open <app-id> --relaunch
# → Session created on BrowserStack with "Untitled Project" / "Untitled Build" / empty name
```

## Evidence

**Local connection profile is correct** (`.agent-device/remote-connections/generated/browserstack-*.json`):
```json
{
  "providerProject": "Galapagos",
  "providerBuild": "Smoke — 2026-09-11 — 25906e0f",
  "providerSessionName": "conversations-2 — iOS"
}
```

**BrowserStack API response** for the same session:
```json
{
  "name": "",
  "build_name": "Untitled Build",
  "project_name": "Untitled Project"
}
```

## Suggested Fix

Forward the three missing fields alongside `providerApp` in both locations:

### `daemon-client-lifecycle.js` — `qt()` function
```diff
  ...e.command === 'lease_allocate' && typeof e.flags?.providerApp === 'string'
    ? {
        providerApp: e.flags.providerApp,
+       providerProject: e.flags.providerProject,
+       providerBuild: e.flags.providerBuild,
+       providerSessionName: e.flags.providerSessionName,
      }
    : {}
```

### `internal/daemon.js` — `Pi()` function
```diff
  flags: e === 'lease_allocate'
    ? {
        providerApp: q(t, 'providerApp'),
+       providerProject: q(t, 'providerProject'),
+       providerBuild: q(t, 'providerBuild'),
+       providerSessionName: q(t, 'providerSessionName'),
      }
    : undefined
```

## Environment

- agent-device: 0.20.10 and 0.21.0 (both affected)
- Node: 22.23.2, Linux x64
- BrowserStack App Automate
- iOS (iPhone 15@17, iPhone 16@18)

## Notes

- PR #1544 (merged Aug 2) fixed a related `bstack:options` overwrite issue but did not address this RPC forwarding gap.
- The `connection.js` `N()` function correctly captures all three fields from CLI flags.
- The `provider-webdriver.js` `prepareSession` callback correctly reads them via `Q(req, 'providerProject')` and maps to `bstack:options.projectName` — the downstream code is fine, the fields just never arrive.
- BrowserStack REST API allows updating session name and status after creation, but **project and build name are immutable** — there is no post-creation workaround for those fields.

thymikee:
## Summary

`--provider-project`, `--provider-build`, and `--provider-session-name` were captured in the connection profile and reach the daemon on the line transport (which forwards the whole request), but the compact JSON-RPC `lease_allocate` envelope forwarded **only `providerApp`**. Over an HTTP/remote daemon the daemon's lease-lifecycle provider received no session metadata, so BrowserStack sessions landed in "Untitled Project"/"Untitled Build" with an empty name. Project and build names are immutable after creation, so there is no post-hoc workaround.

The envelope is a hand-written projection, so it also dropped the fields `prepareSession` needs to create the session at all — device selection (`platform`, `device`), `providerOsVersion`, and the configured device-feature/AWS knobs — meaning a fresh remote allocation still failed before the names could take effect. Fix at the construction seam: one shared projection, `readLeaseAllocateProviderFlags`, now carries the full provider-allocation set that both the client's producer and the daemon's consumer read, so the two transports agree and a sibling cannot be dropped again (the drift #2110 left when it forwarded `providerApp` alone). The key list is pinned exhaustive against `CloudProviderProfileFields`. Additive on the wire: each field rides only when supplied; a request carrying none is byte-identical.

```
connect browserstack --provider-app bs://… --provider-project MyProject \
  --provider-build Build-2026-09-11 --provider-session-name "smoke — iOS"
```

Closes #2494

## Validation

Tested at commit d99226b. `pnpm check:affected --run` green (unit-core 3172 tests, provider scenarios, `check:daemon-wire-compat`: 2 changed, protocol unchanged — acked additive in `ledger.json`).

Regression coverage at each seam: a contract unit test for the shared projection, a producer test that the flags ride the `lease_allocate` payload, an HTTP reconstruct check (`daemon-http-lease-allocate.test.ts`), and `cloud-webdriver-lease-http.test.ts`, which drives the exact client envelope through a real daemon HTTP server into the real BrowserStack `prepareSession` and asserts the `POST /wd/hub/session` capabilities (`deviceName`/`os_version`/`app` + `bstack:options.{projectName,buildName,sessionName,…}`). Narrowing the projection back to the label trio makes that end-to-end test fail, so it is a real guard. `tsc` (contracts + root), `oxlint`, `oxfmt --check`, `check:layering`, and `check:production-exports` pass. Not device-verified against live BrowserStack.

### DIFF
--- a/packages/contracts/src/__tests__/lease-scope.test.ts
+++ b/packages/contracts/src/__tests__/lease-scope.test.ts
@@ -8,7 +8,7 @@ import {
   leaseScopeToConnectionMetadata,
   leaseScopeToLeaseRpcParams,
   leaseScopeToRequestMeta,
-  readLeaseAllocateProviderMetadata,
+  readLeaseAllocateProviderFlags,
 } from '../lease-scope.ts';
 
 test('leaseScopeFromOptions normalizes public aliases and projects request meta', () => {
@@ -168,9 +168,15 @@ test('findMissingProxyLeaseFields enforces complete proxy ownership scope', () =
   );
 });
 
-test('readLeaseAllocateProviderMetadata carries the session-naming flags and drops the rest', () => {
+test('readLeaseAllocateProviderFlags carries the provider-allocation flags and drops the rest', () => {
   assert.deepEqual(
-    readLeaseAllocateProviderMetadata({
+    readLeaseAllocateProviderFlags({
+      session: 'default',
+      token: 'secret',
+      runId: 'run-a',
+      deviceKey: 'dk-1',
+      provider: 'browserstack',
+      ttlMs: 1000,
       platform: 'ios',
       device: 'iPhone 15',
       providerApp: 'bs://abc',
@@ -178,17 +184,34 @@ test('readLeaseAllocateProviderMetadata carries the session-naming flags and dro
       providerProject: 'MyProject',
       providerBuild: 'Build-1',
       providerSessionName: 'smoke',
+      providerDeviceOrientation: 'landscape',
+      providerGeoLocation: '52.5,13.4',
+      providerLanguage: 'en',
+      providerNoResignApp: true,
+      awsRegion: 'us-west-2',
+      awsProjectArn: 'arn:aws:devicefarm:0',
+      awsInteractionMode: 'NO_VIDEO',
     }),
     {
+      platform: 'ios',
+      device: 'iPhone 15',
       providerApp: 'bs://abc',
+      providerOsVersion: '17',
       providerProject: 'MyProject',
       providerBuild: 'Build-1',
       providerSessionName: 'smoke',
+      providerDeviceOrientation: 'landscape',
+      providerGeoLocation: '52.5,13.4',
+      providerLanguage: 'en',
+      providerNoResignApp: true,
+      awsRegion: 'us-west-2',
+      awsProjectArn: 'arn:aws:devicefarm:0',
+      awsInteractionMode: 'NO_VIDEO',
     },
   );
   assert.deepEqual(
-    readLeaseAllocateProviderMetadata({ providerApp: 'bs://abc', providerBuild: undefined }),
+    readLeaseAllocateProviderFlags({ providerApp: 'bs://abc', providerBuild: undefined }),
     { providerApp: 'bs://abc' },
   );
-  assert.deepEqual(readLeaseAllocateProviderMetadata(undefined), {});
+  assert.deepEqual(readLeaseAllocateProviderFlags(undefined), {});
 });
--- a/packages/contracts/src/__tests__/lease-scope.test.ts
+++ b/packages/contracts/src/__tests__/lease-scope.test.ts
@@ -8,6 +8,7 @@ import {
   leaseScopeToConnectionMetadata,
   leaseScopeToLeaseRpcParams,
   leaseScopeToRequestMeta,
+  readLeaseAllocateProviderMetadata,
 } from '../lease-scope.ts';
 
 test('leaseScopeFromOptions normalizes public aliases and projects request meta', () => {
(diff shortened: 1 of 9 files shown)


thymikee:
The shared naming-field projection is correct, but the claimed fresh HTTP BrowserStack path is still incomplete at 7e36094. readLeaseAllocateProviderMetadata drops platform, device and providerOsVersion; toLeaseDaemonRequest reconstructs no other flags, and leaseLifecycleContext passes that bag directly to the provider. BrowserStack prepareSession requires those fields before creating the session, so a fresh remote allocation still fails before the names can take effect. Configured BrowserStack device-feature flags are also dropped by this projection. This omission predates the PR, but it is at the same allocation boundary and blocks the advertised fix. Carry the required provider configuration through the shared contract and extend the HTTP regression through the real lease/provider preparation path to assert the outgoing session capabilities. The current test substitutes handleRequest and only proves the intermediate flags bag. The iOS smoke failure is RUNNER_BUSY at the WebView wait and appears unrelated.

thymikee:
Addressed in 176e8ebe.

The shared projection is now `readLeaseAllocateProviderFlags` and carries the full set `prepareSession` reads, not just the labels: device selection (`platform`, `device`), `providerOsVersion`, the app, the session-naming fields, the eight BrowserStack device-feature knobs, and the AWS knobs. Producer and consumer already shared the one reader, so both transports move together. The key list is pinned exhaustive against `CloudProviderProfileFields`, so a new provider field that skips the projection fails to compile instead of silently breaking a fresh remote allocation.

The reviewer's "carries the required provider configuration" and "assert the outgoing session capabilities" points are both in. New `cloud-webdriver-lease-http.test.ts` drives the exact client envelope (`buildHttpRpcPayload`) over a real `node:http` request into a real `createDaemonHttpServer` + `createRequestHandler`, through the real BrowserStack `prepareSession`, and asserts the `POST /wd/hub/session` body — `deviceName`/`os_version`/`app` and `bstack:options.{projectName,buildName,sessionName,deviceOrientation}`. The provider fake hijacks `globalThis.fetch`, so the daemon RPC is spoken over `node:http` to avoid the interception.

Proof it is a real guard, not a green-by-construction test: narrowing the projection back to the label trio makes that end-to-end test fail (the daemon returns not-ok because `prepareSession` rejects the missing `--device`/`--platform`). The `daemon-http-lease-allocate.test.ts` reconstruct check is broadened in step, now that a boolean (`providerNoResignApp`) and device-feature/AWS keys ride too. Per the tripwire note, `daemon-http-server.test.ts` was left at your version and not grown.

Wire digests re-pinned; ack rationale updated. `check:affected --run` green (3172 unit tests, daemon-wire-compat "2 changed, protocol unchanged"). The iOS `RUNNER_BUSY` at the WebView wait looks unrelated, as you said.

thymikee:
No remaining findings at d99226b. The shared projection carries the full allocation configuration, and the HTTP regression now reaches the real BrowserStack preparation path and checks the outgoing capabilities. Ready for human review; live BrowserStack acceptance was not tested, and the iOS WebView RUNNER_BUSY failure appears unrelated to this transport change. Revalidate iOS smoke before merge.

PR #1505, opened 2026-07-30
test: characterize replay-test reporter contract and extend R10

thymikee:
Preparatory slice of **P3** from #1478. This is **not** the extraction — no production code moves, `packages/replay-test` is not created, and no façade, manifest type, or neutral outcome type is added. The only non-test file is the layering gate.

P3's brief calls for exactly this first: *"Before moving production code, characterize the exact reporter start/step/result values (especially `session`) and the late-timeout finalization/double-cleanup behavior"* and *"Extend R10 so replay-test also rejects imports from `src/request/**` and engine internals."* Doing it now keeps P3 progressing while P2 owns the Maestro façade and shared discovery/format routing, which the real extraction depends on.

All characterization tests pass against current `main` **unchanged** — they document today's behavior, they do not propose new behavior.

## Reporter value projection (`toReplayTestReporterProgressEvent`)

`command` progress never reaches a hook; `replay-test-suite` → `onSuiteStart`; `start` → `onTestStart`; `progress` → `onTestStep`; `pass|fail|skip` → `onTestResult` with `status` verbatim.

**Key presence** is pinned: the projection rebuilds an object literal, so all 12 case keys (`file,title,index,total,attempt,maxAttempts,session,artifactsDir,shardIndex,shardCount,deviceId,deviceName`) exist with `undefined` values even when the wire event omitted them, so `in` checks keep working. Step adds `stepIndex,stepTotal,stepCommand,stepValue`; result adds `status,durationMs,retrying,message,hint`. Step values carry no `status` key at all. Suite-start is exactly `{total,runnable,skipped,artifactsDir,shardMode,shardCount}`.

## `session` — exact shape and provenance

Format: `` `${base}:test:${suiteInvocationId}:${runnableOrdinal}${-slug}:attempt-${n}` `` (`session-test-discovery.ts:93`, `attemptIndex = 0` default at `:85`).

Four distinct provenances that deliberately disagree:

| hook | value | source |
|---|---|---|
| `onTestStart` | **always `:attempt-1`** — built before any attempt runs, with the defaulted `attemptIndex` | `session-test-attempt.ts:194` |
| `onTestStep` | the **running** attempt (`attempt-1`, then `attempt-2`, …) | ALS context set at `session-test-attempt.ts:160`, read at `session-replay-runtime.ts:449` / `session-replay-maestro-observer.ts:81` |
| `onTestResult` (retry) | that failed attempt's session | `session-test-attempt.ts:248` |
[...]

PR #959, opened 2026-06-30
feat: support live replay test reporters

thymikee:
## Summary

Enable custom replay test reporters to receive live progress events while `agent-device test` is running, with stream-aware stdout/stderr reporter context.

Update the default reporter to render through the same live reporter path and document a custom emoji reporter example.

Touched 16 files; scope stayed within replay test reporting, daemon client progress forwarding, CLI command wiring, tests, and docs.

## Validation

`pnpm exec vitest run src/__tests__/cli-network.test.ts src/utils/__tests__/daemon-client.test.ts` passed: 2 files, 46 tests.

`pnpm typecheck` passed.

PR #2497, opened 2026-09-11
fix(test): prevent custom reporter exit-code wraparound

PLASMA-FR:
## Summary

Completes the reporter exit-code guarantee from #959, characterized in #1505: a custom reporter must not mask a failing suite. On current `main`, a reporter returning `256` or `512` makes this command exit successfully on Linux even when a test fails:

```sh
agent-device test ./workflows --reporter ./reporter.mjs
```

Validate `getExitCode` results as integers from `0` to `255`, or `undefined`, before aggregation and JSON success output. Invalid values fail with `INVALID_ARGS`; valid reporters retain the highest-code rule and failed-suite floor. Eight files, including regression tests, public type documentation, help, user docs, and changelog.

## Validation

Tested `a745e45c6aa5d05afdb0e4659608ef6e93c40193` using Node 24.13.1, pnpm 11.17.0, and `AGENT_DEVICE_VITEST_MAX_WORKERS=1`.

- Eleven regression cases fail before the fix. The JSON case also fails with only the range check applied.
- Build, 138 focused reporter/CLI/command/startup tests, and 12 command-doc tests pass.
- Built CLI against an owned local HTTP fixture: all 20 text/JSON cases pass, covering successful/failed suites and codes `0`, `3`, `255`, `256`, and `512`. Upstream gives eight incorrect exit statuses.
- `pnpm check:affected --run` passes all runnable checks: 2,805 tests across 383 files, formatting, lint, types, layering, Fallow, and build.

No device run: the changed boundary is local CLI reporting. Upstream CI is pending.


thymikee:
No findings at a745e45. Reporter codes are validated before aggregation and JSON success output, and the CLI regression covers the false-success case. Ready for human review. The remaining iOS smoke failure is an alert-replacement XCTest assertion and appears unrelated to this CLI-only change.

PR #2499, opened 2026-09-11
fix(scroll): pace edge passes to rest and stop a stuck surface

thymikee:
## Summary
`scroll bottom` flung up to 40 times without net progress on a container that reports hidden content but does not move (focused `TextInput`, nested scroller, or a bottom that rubber-bands). The loop captured mid-bounce, so each pass looked new and another fling stacked on the bounce.

- The edge loop now waits for the scoped scroll surface to hold between passes before deciding or flinging again, and stops with a typed `scroll_edge_no_progress` when descendants never shift. The end backstop keeps its own `scroll_edge_pass_limit`.
- `scroll --until` shares the stuck-signature window and raises `scroll_until_no_progress`.
- The surface fingerprint and stuck-window bookkeeping are shared from capture-kit so both loops read one definition of "stuck".

7 files, capture-kit + daemon scroll family.

```
$ agent-device scroll bottom   # stuck checkout form, keyboard focused
Error: scroll bottom moved nothing across 3 passes: the container still reports hidden content but its contents never shifted  (reason: scroll_edge_no_progress)
```

## Validation
Commit `93de350fec`.
- `pnpm check:affected --run` → all runnable checks passed (722 tests, incl. fallow complexity on the changed files).
- `pnpm typecheck` / `pnpm build` clean.
- Live on iPhone 17 Pro (iOS 26.2), `com.callstack.agentdevicelab`:
  - Stuck Form with keyboard focused: `scroll bottom` stopped in **3 passes (~4s)** with `scroll_edge_no_progress` (was ~34s / 40 passes).
  - Home list (genuinely scrollable): `scroll bottom` reached the edge in **2 passes**, no false stop; `scroll down --until` returned on an already-visible target.

### DIFF
--- a/packages/capture-kit/src/snapshot/__tests__/scroll-edge-state-fingerprint.test.ts
+++ b/packages/capture-kit/src/snapshot/__tests__/scroll-edge-state-fingerprint.test.ts
@@ -1,0 +1,17 @@
+import { test } from 'vitest';
+import assert from 'node:assert/strict';
+import { capture, recycledCellSnapshot } from './scroll-edge-state-fixtures.ts';
+
+test('surface signature changes when a recycled cell keeps its slot but changes text', async () => {
+  const before = await capture(recycledCellSnapshot('First story'));
+  const after = await capture(recycledCellSnapshot('Second story'));
+  assert.ok(before.fingerprint && after.fingerprint);
+  assert.notEqual(before.fingerprint, after.fingerprint);
+});
+
+test('surface signature is stable when nothing on the container moves', async () => {
+  const first = await capture(recycledCellSnapshot('First story'));
+  const second = await capture(recycledCellSnapshot('First story'));
+  assert.ok(first.fingerprint);
+  assert.equal(first.fingerprint, second.fingerprint);
+});
--- a/packages/capture-kit/src/snapshot/__tests__/scroll-edge-state-fixtures.ts
+++ b/packages/capture-kit/src/snapshot/__tests__/scroll-edge-state-fixtures.ts
@@ -66,6 +66,32 @@ export function scrollSnapshot(hiddenContentBelow: boolean): SnapshotNode[] {
   ];
 }
 
+/**
+ * A scrolled container whose one visible cell keeps a stable identifier and slot while only its text
+ * varies — the recycled-cell shape where identity alone would hide real progress.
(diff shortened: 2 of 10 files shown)


thymikee:
Two false-stop cases remain at 93de350:

- `selection.ts` builds the surface signature from `identifier ?? label ?? value`. Recycled rows can keep the same identifier and coordinates while their labels or values change, so a list making real progress is reported as stuck after three passes. Include changing content in the signature and cover recycled rows with stable identifiers and positions.
- `scrollSurfaceIsStuck` accepts any four captures with at most two distinct signatures. That includes `A,A,A,B`, where the latest pass just made progress; both loops immediately throw after adding B. Require evidence of no new progress, and add a regression that lets this sequence continue.

The reported simulator runs cover a stuck form and a short healthy list, but do not cover these cases. The iOS smoke failure is SESSION_NOT_FOUND during typing after replacement of an unreachable daemon; it appears unrelated to the scroll change.

thymikee:
Both false stops are addressed in `e43287c2`.

- **Recycled cells** — `surfaceIdentity` no longer coalesces to `identifier ?? label ?? value`; the signature now keys on identifier, label, and value together, so a recycled cell that keeps its identifier and slot while its text changes reads as progress. Covered by `scroll-edge-state-fingerprint.test.ts` (recycled cell, stable id+slot, changing label → signatures differ; static surface → stable).
- **`A,A,A,B` stopping after real progress** — `scrollSurfaceIsStuck` now requires the newest signature to be one the window already showed. `A,B,A,B` rubber-bands onto a seen signature and trips; `A,A,A,B` ends on unseen `B` and continues. Covered by the `scroll-edge-state-stuck.test.ts` table and a loop regression in the pass-orchestration suite (`a fresh surface signature continues after identical passes`).

`pnpm check:affected --run` green at this head. Agreed the SESSION_NOT_FOUND iOS smoke failure is unrelated to the scroll change.

thymikee:
Both false-stop findings are resolved at e43287c. Recycled rows now include their changing text in the fingerprint, and a newly observed signature lets the loop continue; the focused regressions cover both cases. Ready for human review. The current iOS smoke fails before scrolling, while waiting for the initial app screen with no readable capture, so it appears unrelated; revalidate that check before merge.

PR #2501, opened 2026-09-11
fix(selectors): resolve a wrapper chain's control for uniqueness reads

### LINKED ISSUE #2498: wait still refuses an unverified-hittability wrapper chain (native runner path)
# `wait` still refuses an unverified-hittability wrapper chain

PR #2482 (merged as `c94e66e`) collapses an unverified-hittability wrapper chain for the **action-target resolution** only: `press`/`click`/`fill`/`longpress`, the `is` target verification, and the post-action settle observation. Issue #2480 was closed by that merge, but its second half is not fixed, so I am opening this to keep it tracked rather than lost.

## Symptom (0.21.0 and current `main`)

A SwiftUI toolbar item that carries an accessibility identifier yields two nodes: the wrapper `Other` and the `Button` under it, with no hittability evidence and rects ~0.7-1 pt apart.

```
$ agent-device press 'id="scoring_home_button"' --settle
Error (AMBIGUOUS_MATCH): Selector matched 2 distinct actionable elements: id="scoring_home_button"
Candidates: @e3 [other] "Home"  @e4 [button] "Home"        <- fixed by #2482

$ agent-device wait 'id="scoring_home_button"' 3000
Error (AMBIGUOUS_MATCH): selector matched multiple elements
Diagnostics Log: .../runner.log                            <- still fails
```

Observed screen state in both cases: `scoring_home_button` -> `Other [20,63,36,36]` + `Button [21,63,35,36]`.

## Why it is a separate path

`wait` does not run the action-target resolution. `src/daemon/wait-runtime.ts` builds its own selector runtime:

```
src/daemon/wait-runtime.ts:46   await bindWaitOperations(params, parsed, session, device)
src/daemon/wait-runtime.ts:156  const runtime = createSelectorRuntimeForDevice({ ... })
```

`classifyActionableTouchCandidates` is reached only from `src/commands/interaction/runtime/selector-action-resolution.ts` (called by `resolution.ts`), which `wait` never imports. On iOS the refusal surfaces from the native runner, which does not use the shared structural-equivalence rules:

- `apple/runner/AgentDeviceRunner/AgentDeviceRunnerUITests/RunnerTests+CommandExecution.swift:1683`
- `apple/runner/AgentDeviceRunner/AgentDeviceRunnerUITests/RunnerTests+Interaction.swift:238`

## Options

1. **Delegate**: have the runner hand its multi-match decision to the shared equivalence rules, so the selector layer and the runner cannot drift.
2. **Mirror**: re-implement the wrapper-chain rule inside the runner.

Option 1 is preferable: the two layers disagree today, and mirroring guarantees a future divergence.

## Verification

A device lane is required: an iOS simulator run of `press` + `wait` against one toolbar control, with the same screen state recorded on both sides. I can prepare the patch and run that lane if you pick a shape.

thymikee:
## Summary

`wait` already succeeds on `main`; what still fails on the reported screen is every read that insists on one match. A SwiftUI toolbar wrapper and its button share one identifier, a regular iOS snapshot omits unverified hittability, so the ladder that relates a wrapper to its control cannot fire: `press` taps the button while `is visible` and `get attrs` report `Selector did not match` and `screenshot --crop-on` refuses with "matched 2 nodes". #2482 collapsed that chain for mutating resolution only.

This exports the collapse (`resolveUnverifiedWrapperControl`) and applies it at the read door: `readUnique` and `cropTarget` resolve the control instead of refusing. Rows that resolve before a refusal are untouched, and a candidate set the rule does not recognise as one control - a cell and the button inside it, matches in distinct subtrees - still refuses.

Replay was the second door. Verification re-resolves a recorded selector under the same row's refusal rules, so a step whose screen had not changed verified as `IDENTITY_MISMATCH` on its first replay; it now names the collapsed control, the node dispatch acted on.

ADR 0011 is amended: its 2026-08-07 clause still described `get`/`is`/`wait` as runner `querySelector` reads. 11 files, +431/-31.

Closes #2498

## Validation

- `pnpm check:affected --run` at `bab4435` (rebased onto `origin/main` d80fb35): 563 files / 4084 tests pass. An earlier run at the pre-rebase commit had one failure under concurrent load from other worktrees on this host and passed on rerun.
- Live iOS Simulator (iOS 26.2) with a SwiftUI fixture whose toolbar wrapper shares its button's identifier. Before: `is visible` and `get attrs` selector-not-matched, `screenshot --crop-on` matched 2 nodes. After: `is visible` passes, `get attrs`/`get text` resolve the Button, crop is 35x36, `wait` and `press` unchanged, `is hidden` fails with `actual={"visible":true}`, `find … list` still lists both nodes, and a distinct-subtree duplicate still refuses.
- Mutation-checked: disabling the collapse or its ancestry-chain guard fails the new tests.
- Unresolved scope: on XCTest-backed captures where the wrapper reports `hittable: true`, reads and acts both refuse. That is #2482's existing boundary and is unchanged here.

### DIFF
--- a/docs/adr/0011-interaction-guarantee-contract.md
+++ b/docs/adr/0011-interaction-guarantee-contract.md
@@ -260,6 +260,36 @@ classifier. Reads (`querySelector`, and so `get`/`is`/`wait`) keep the prior rul
 decorative duplicate into an error. Maestro's explicit expected-point /
 non-hittable compatibility path remains intentionally separate.
 
+### 2026-09-11 amendment: one collapse for one control, at both doors
+
+Two clauses of the amendment above went stale as the read paths moved, and the
+structural rule never reached the read door.
+
+- `querySelector` no longer backs `get`/`is`/`wait`. Those reads resolve against a
+  capture, through the `readUnique`/`readAny`/`readText` rows of
+  `packages/selectors/src/selector-pipeline.ts`, so the runner's hittable-preference
+  rule governs only the direct XCTest paths that still query it: the direct-iOS
+  touch fast path and the off-screen target probe.
+- A control reported through its own accessibility wrapper answers a selector
+  twice. A regular iOS snapshot omits unverified hittability, so the ladder that
+  relates a wrapper to its control cannot fire, and #2482 collapsed that chain for
+  mutating resolution only: `is visible` and `get attrs` refused the same screen
+  as ambiguous while `press` tapped it. The collapse now sits beside the
+  classification that asks for it (`resolveUnverifiedWrapperControl`) and applies
+  where a refusal was the answer: the uniqueness rows — `is <predicate>`,
+  `get attrs`, `screenshot --crop-on` — resolve the control instead of reporting
+  no match. A row that resolves before any refusal is untouched by it, and a
+  wrapper chain always resolves before one: depth separates a wrapper from its
+  control, so `get text` ranks onto the control and `wait`/`is exists` answer from
+  the document-order head without asking which element was meant. A candidate set
+  the rule does not recognize as one control — a cell and the button inside it, or
+  matches in distinct subtrees — still refuses.
+- Replay verifies a recorded target by resolving its recorded selector again under
+  the same row's refusal rules, so the screen dispatch had resolved read as an
+  identity mismatch on the step's first replay. Verification names the collapsed
+  control too, which is the node dispatch acted on and the node the recorded
+  identity carries.
+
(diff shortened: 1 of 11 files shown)


thymikee:
No findings at bab4435. Uniqueness reads and replay now resolve the same wrapper control while distinct subtrees and genuinely separate controls still refuse; the reported live simulator run covers is/get/crop and the regression tests cover replay. Ready for human review and merge; current-head checks pass.

PR #2503, opened 2026-09-11
fix(scroll): keep the swipe above the keyboard, refuse when it cannot

### LINKED ISSUE #2500: scroll flings into the keyboard; keep the swipe in the visible surface and fail closed
## Purpose

`scroll down` / `scroll bottom` synthesizes a viewport-center-symmetric swipe (`centerY +/- travel/2`). When a text input is focused, the keyboard covers the lower screen, so the swipe endpoints land on the keys and the surface never moves. The agent sees a stuck scroll.

#2499 turned this into a fast typed stop (`scroll_edge_no_progress`) that hints "dismiss the keyboard first", but nothing yet makes the scroll *succeed* when it can. This issue is the fix: make directional scroll keep its swipe above the keyboard, and refuse with a typed reason when it cannot, instead of swiping under the keys.

## Reproduce

Test app (`com.callstack.agentdevicelab`) Form screen: tap the editable "Delivery notes" field (keyboard up), then `scroll bottom`. Today it stops with `scroll_edge_no_progress` because the fling hit the keyboard.

## Verify the actual failing mode first

The iOS keyboard clip is gated by `keyboardPolicy: .whenAccessibilityHealthy` (`RunnerTests+SynthesizedGesturePolicy.swift:91`), and health flips to `.healthy` after any successful snapshot. An agent that snapshots before scrolling is usually already past the gate — so the gate is **not** the common cause.

Read `axHealth=` from the `AGENT_DEVICE_RUNNER_SYNTHESIZED_GESTURE_POLICY` log line of a failing run:
- `unknown`/`unavailable` -> gate is closed; the fix is the policy change (slice 1a).
- `healthy` -> the gate is open and the clip is **failing open**: `frameAvoidingKeyboard` returns the full frame when remaining safe height < 25% (`RunnerTests+Interaction.swift:780`), which is exactly the under-keyboard swipe. The fix is to fail closed (slice 1b).

## Required behavior

- **Geometry avoidance is the default; never auto-dismiss.** Dismiss drops focus (a `type`/`scroll`/`type` loop breaks), is not idempotent across platforms (Android's ESC loop can throw `UNSUPPORTED_OPERATION`), and mutates state that session-action provenance does not record. `keyboard dismiss` stays an explicit command. No "keep keyboard" flag is needed because avoidance never mutates.
- **Owners read their own live keyboard frame.** Do not reuse the daemon's last snapshot (for single-pass scroll it predates the keyboard and is stale in the harmful direction) and do not thread a frame across two coordinate spaces (iOS screenshot points vs Android absolute pixels) — that is a parity trap.
- **No new neutral-intent field and no change to the geometry planners.** The clip only reduces `referenceHeight` before the planner runs, so `buildScrollGesturePlan` (TS) and `runnerScrollGesturePlan` (Swift) stay byte-identical and the `contracts/fixtures/scroll-gesture.json` parity fixture does not change. Report honest travel through the existing resolved-frame recomputation.
- **Add evidence, not just a behavior change.** Surface `keyboardAvoided: boolean` and `keyboardMinY: number` on the scroll result.
- **Fail closed with a typed reason.** When the keyboard leaves too little visible surface to swipe (floor ~15% of the viewport) or the clip would cross the app top, return `scroll_keyboard_occludes_surface` with the "dismiss the keyboard" hint instead of swiping under the keys or reporting a generic `scroll_edge_no_progress`. A tiny clipped swipe otherwise reads as "stuck" to the #2499 fingerprint.
- **Account for the input accessory bar.** `keyboard.frame` excludes the accessory/composer bar, so keep a fixed allowance above `keyboardMinY`.
- **Codify in the runner policy table, not the ADR 0011 matrix.** Scroll has no rows in `packages/contracts/src/interaction-guarantees.ts` and the runner policy is deliberately separate; extend the policy table + its tests + `scroll` help.

## First slice — iOS

1a. Change the scroll `keyboardPolicy` to `.requiredWhenAvailable` and update the pinned policy tests (`RunnerTests+SynthesizedGesturePolicy.swift:187`). Note the probe is not free: `visibleKeyboardFrame` resolves `app.keyboards.firstMatch` via a live AX fetch.
1b. Make `frameAvoidingKeyboard` fail closed: return an occlusion outcome; map it in the scroll case to the typed `scroll_keyboard_occludes_surface` stop.
1c. Add `keyboardAvoided` / `keyboardMinY` to the scroll response.

**Completion:** on the simulator, focus the Form field and run `scroll bottom` as the *first* command after a runner relaunch: the edge loop makes progress and ends at the edge, the response reports `referenceHeight` smaller than the screen height, and the runner log shows `axHealth=unknown keyboardPolicy=requiredWhenAvailable`. A landscape run where the keyboard leaves <15% ends with `scroll_keyboard_occludes_surface`, not `scroll_edge_no_progress`.

## Second slice — Android (free keyboard frame)

`GestureViewportReader` already calls `UiAutomation.getWindows()` and keeps `TYPE_APPLICATION` (`GestureViewportReader.java:26`); the same list has the `TYPE_INPUT_METHOD` window with screen bounds. Return those bounds, clip the scroll viewport before `buildScrollGesturePlan` (`packages/platform-android/src/input-actions.ts:181`), same floor + typed reason. This also covers `adjustPan`/`adjustNothing`, which the current window-bounds frame misses. Validate live on an `adjustPan` activity.

## Explicit non-goals

- No auto-dismiss of the keyboard inside scroll.
- No container-rect-based swipe origin: a full-height list still extends under the keyboard, so it does not solve this and would change the golden geometry for every small container. Threading the selected container rect (and closing the selection-center vs swipe-center split) is a separate, later change.

## References

- Depends on the no-progress stop: #2499.
- Precedent for dismiss-then-act (kept out of default here): `webdriver-interactor.ts` `hideKeyboard()`; dismiss-await race pattern: #2260.

thymikee:
## Summary

iOS half of #2500, stacked on #2537, where the shared rule, the golden table, the typed refusal and the response keys are stated once for every platform.

A directional `scroll` planned a viewport-center-symmetric swipe, so a focused field put the swipe's lower endpoint under the keyboard: the gesture hit keys, the surface never moved, and the edge loop reported a stuck container. The runner now reads its own live keyboard frame, clips the swipe band against the shared rule, reports `keyboardAvoided` / `keyboardMinY`, refuses with `scroll_keyboard_occludes_surface` when the keys leave under 15% of the band, and never dismisses the keyboard. The superseded scroll keyboard path was deleted, not guarded.

Closes #2500 once #2537 and #2514 land. Android goes on top of this in #2514. main now carries #2499, so both halves of the issue are in: the clip stops the wasted gesture, and #2499's no-progress stop ends the loop.

## Review pass

The width-fraction gate is gone: the clip asks whether the keyboard reaches the line the swipe runs along, not what share of the viewport's width it covers, because a vertical swipe travels the centre line and a floating keyboard straddling that line blocks it however narrow it is. The row that proves it (`centred floating keyboard under half the viewport width sits in the swipe's path`) is in the shared table landed by #2537, and the Swift lane asserts that same JSON row here.

Two things found while checking for reuse, both left alone deliberately:

- `frameAvoidingKeyboard` (`RunnerTests+Interaction.swift:789`) still gates on `intersection.width / frame.width >= 0.5`, i.e. the same defect for tap/drag reference frames. Fixing it changes the tap family, so it belongs in its own change; flagged rather than bundled.
- `keyboardAvoidingDragPoints` (same file, `:702`) keeps a local `padding: Double = 12` that names the same fact as `accessoryAllowance`, and fails open where scroll refuses. Sharing the constant means editing the drag rule, and the two failure policies differ on purpose: a drag that cannot fit still wants to happen, a scroll entirely under the keys does not.

Reused `visibleKeyboardFrame` and the iOS lane's existing `-only-testing` list.

A second pass found the clip leaking into the coordinate rotation. `resolvedScrollViewport` returned one frame and the command used it for two jobs: planning the swipe, and re-basing the touch context. `nativeSynthesizedPoint` derives a `landscapeRight` native x from that frame's HEIGHT, so clipping an 834pt landscape viewport to 576pt moved the dispatched gesture 258pt sideways off the lane the plan had just been built for. The viewport now carries `planFrame` and `coordinateFrame` separately and the gesture comes from one `gestureDispatch` decision, so the two cannot be swapped; `testScrollViewportDispatchKeepsTheUnclippedFrameAsItsCoordinateRotationBasis` runs the landscape case through that decision and fails on the swap.

A third pass found that fix had moved the reported frame too. Splitting the frames left the response naming the synthesis frame: `executeSynthesizedDragGesture` builds its `DragVisualizationFrame` from the frame the coordinates rotate in, so a swipe planned inside a 571pt band answered with `referenceHeight: 874`, and `pixels` — which is a fraction of the band — was measured against an axis the caller was never told about. The keyboard evidence now travels with the dispatch instead of being bolted on afterwards: `ScrollGestureDispatch.attachingEvidence(to:)` names `planFrame` in `referenceWidth` / `referenceHeight` and keeps `coordinateFrame` for synthesis, and the old `attachingScrollViewportEvidence` helper is gone. The regression above now asserts the assembled response as well, so reporting the synthesis frame is a red test rather than a reviewer's job.

## Validation

`pnpm check` green at `8a67c377cf` (exit 0).

Reproducing the runner tests locally needs `AGENT_DEVICE_XCUITEST_INCLUDE_UNIT_TESTS=1` on the build: `xcodebuild test-without-building -only-testing:` executes **0 tests and still reports the suite as passed** without it, because that env var is what defines `AGENT_DEVICE_RUNNER_UNIT_TESTS`. The iOS lane always sets it; my own earlier local runs did not and were therefore empty. With `AGENT_DEVICE_XCUITEST_INCLUDE_UNIT_TESTS=1 pnpm build:xcuitest:ios` the scroll and policy tests genuinely execute (`Executed 4 tests, with 0 failures`), and each guard bites when neutered: handing the clipped band to the rotation basis fails on a `0.0` vs `258.0` shift, and dropping the two evidence lines fails with `("Optional(834.0)") is not equal to ("Optional(576.0)")` — the symptom reported in review, now pinned.

iOS, iPhone 17 Pro, Form screen, field focused:
- `scroll down --json` -> `referenceWidth: 402`, `referenceHeight: 571`, `keyboardAvoided: true`, `keyboardMinY: 583`, swipe `y1: 472 -> y2: 100`, `pixels: 371`, and the runner logged `decision=avoided keyboardMinY=583.0 swipeHeight=571.0`. Same command with no keyboard up reported `referenceHeight: 874`, `y1: 721 -> y2: 153`, `pixels: 568`, so the two answers differ exactly by what the keys took.
- `scroll top --json` on the notes field earlier in the same build returned `referenceHeight: 571`, `keyboardMinY: 583`, `y2=472`.
- First scroll after runner relaunch logs `axHealth=unknown keyboardPolicy=requiredWhenAvailable decision=avoided`; the old `whenAccessibilityHealthy` gate skipped the probe there.
- No keyboard -> no evidence keys, `referenceHeight: 874`, `decision=noKeyboard`.

iOS, iPadOS 26 windowed Safari (address bar focused): `referenceWidth: 834`, `referenceHeight: 913`, `keyboardAvoided: true`, `keyboardMinY: 925`. The window is 913pt tall inside a 1219pt display, so this is the clip reading the app window rather than the display. That simulator was created for the attempt and deleted afterwards, and the host's `ConnectHardwareKeyboard` was returned to its original value after the soft-keyboard run.

Unreproduced live: the `occluded` refusal. The app-side route was prototyped and ruled out on a device, so this is now a measured ceiling rather than an untested hypothesis.

- **iPhone 17 Pro, full screen.** The system keyboard keeps one height whatever the text size: with "Larger Accessibility Sizes" on and the slider at 100% (`UIPreferredContentSizeCategoryName = UICTContentSizeCategoryAccessibilityXXXL`, verified in the device's `com.apple.UIKit` domain, and the app layout visibly at AX5), `scroll down --json` still reported `keyboardMinY: 582.9999999999999`, `referenceHeight: 570.9999999999999`, swipe `471 -> 99`, `pixels: 371` — the same clip as the default size. Dynamic Type does not grow the keyboard, so the band above it is 571 of 874 = 65% and cannot fall under 15%.
- **iPadOS 26 windowed Safari.** Measured window 834x913 inside a 1219pt display with keyboard top 925. Even placing that window's bottom edge on the display's bottom edge leaves a 607pt band = 66%; getting under 15% needs a window no taller than ~360pt, and the window controls available on this simulator did not produce one. Dragging the window from the host desktop also timed out at the daemon budget before the drag landed.
- **App-sized custom `inputView`.** A temporary `examples/test-app` route attached an 800pt `UITextField.inputView`. It was visibly attached and the app reported the custom surface, but `snapshot --raw` exposed no `Keyboard` element and `scroll down` still returned `referenceHeight: 874` with no keyboard evidence. That is structural: `visibleKeyboardFrame` asks `app.keyboards.firstMatch`, and XCUITest models system keyboards there — not arbitrary app-provided input views. Reaching a real XCUITest keyboard with an app-controlled height would require a full custom keyboard app extension and its enablement flow, so the prototype was removed rather than merged.
- **Accessibility Keyboard.** Not reached. Settings navigation at AX sizes did not get past the Accessibility list, and the panel would have to exceed ~743pt of 874 to change the answer anyway.

No live iOS response is claimed for `decision=occluded`. The refusal remains exercised by the shared golden row, the Swift policy test, and the command's return before gesture synthesis; the measured iOS ceiling means those remain the available evidence on this platform. Device state was reverted: the accessibility category key was deleted, the temporary iPad simulators were deleted, the custom-input-view prototype and its pod changes were removed, the stim supervisor and its simulator were stopped, and host `ConnectHardwareKeyboard` was restored to `YES`.

## Size

6 files, 536 gross lines, one command family, 4 commits: the runner clip, its lane wiring, and the response evidence. The shared rule and golden table are #2537 (615 gross) and the Android helper read and clip are #2514 (612 gross), which is what the budget needed and also where those facts belong.

### DIFF
--- a/.github/workflows/ios.yml
+++ b/.github/workflows/ios.yml
@@ -165,6 +165,9 @@ jobs:
             -only-testing:AgentDeviceRunnerUITests/RunnerTests/testEmptyReplacementWithoutResolvableTargetFailsClosed \
             -only-testing:AgentDeviceRunnerUITests/RunnerTests/testTextEntryTapWitnessIsBoundToTargetIdentity \
             -only-testing:AgentDeviceRunnerUITests/RunnerTests/testCoordinateTapTextInputProbeSkipsPenalizedXCTestChannel \
+            -only-testing:AgentDeviceRunnerUITests/RunnerTests/testScrollViewportKeyboardClipMatchesGoldenParityTable \
+            -only-testing:AgentDeviceRunnerUITests/RunnerTests/testScrollViewportPolicyUsesParityTableConstants \
+            -only-testing:AgentDeviceRunnerUITests/RunnerTests/testSynthesizedGesturePoliciesMatchCommandContracts \
             -only-testing:AgentDeviceRunnerUITests/RunnerTests/testFreshCoordinateTapContainsUnavailableTextInputProbe \
             -only-testing:AgentDeviceRunnerUITests/RunnerTests/testTextInputProbeIssueScopeIsThreadBound \
             -only-testing:AgentDeviceRunnerUITests/RunnerTests/testTextInputProbePreservesEnclosingRunnerWait \
--- a/.github/workflows/ios.yml
+++ b/.github/workflows/ios.yml
@@ -167,6 +167,7 @@ jobs:
             -only-testing:AgentDeviceRunnerUITests/RunnerTests/testCoordinateTapTextInputProbeSkipsPenalizedXCTestChannel \
             -only-testing:AgentDeviceRunnerUITests/RunnerTests/testScrollViewportKeyboardClipMatchesGoldenParityTable \
             -only-testing:AgentDeviceRunnerUITests/RunnerTests/testScrollViewportPolicyUsesParityTableConstants \
+            -only-testing:AgentDeviceRunnerUITests/RunnerTests/testScrollViewportDispatchKeepsTheUnclippedFrameAsItsCoordinateRotationBasis \
             -only-testing:AgentDeviceRunnerUITests/RunnerTests/testSynthesizedGesturePoliciesMatchCommandContracts \
             -only-testing:AgentDeviceRunnerUITests/RunnerTests/testFreshCoordinateTapContainsUnavailableTextInputProbe \
             -only-testing:AgentDeviceRunnerUITests/RunnerTests/testTextInputProbeIssueScopeIsThreadBound \
--- a/apple/runner/AgentDeviceRunner/AgentDeviceRunnerUITests/RunnerScrollViewportPolicy.swift
+++ b/apple/runner/AgentDeviceRunner/AgentDeviceRunnerUITests/RunnerScrollViewportPolicy.swift
@@ -1,0 +1,256 @@
+import XCTest
+
+// The scroll viewport rule the runner shares with the TS runtime (#2500).
+//
+// RULE: a directional scroll centres its swipe, so a focused field puts the swipe's lower endpoint
+// under the keyboard — the gesture lands on keys, the surface never moves, and the edge loop reads a
+// stuck container (#2499) rather than a refusal. Clipping the viewport to the band above the
+// keyboard BEFORE the gesture planner runs keeps the swipe in what is visible, and when that band is
+// too thin to hold one the rule REFUSES instead of handing back the full frame: a keyboard-struck
+// swipe and a tiny clipped swipe both read as "stuck", so failing open is what hides the failure.
+//
+// The pure rule below is geometry on purpose — no XCUIApplication — so its exact decision is proven
+// against the golden table shared with its TS twin, `clipScrollViewportAboveKeyboard` in
+// packages/contracts/src/scroll-gesture.ts, asserted in that file's test beside this one. The table
+// carries only frames representable in both languages: `CGRect` standardizes a negative extent into a
+// positive height at a moved origin, so a negative `height` is tested on the TS side alone.
+//
+// The `extension RunnerTests` below is the one impure caller: it reads the runner's own live keyboard
+// frame, because a frame threaded from the daemon would predate the keyboard.
+
+/** What an on-screen keyboard leaves of a scroll viewport. */
+enum RunnerScrollKeyboardClip: Equatable {
+  /** No keyboard, or one that does not own this surface: swipe the whole viewport. */
+  case unobstructed
+  /** The viewport trimmed above the keyboard. Report the reduced reference height honestly. */
+  case avoided(frame: CGRect, keyboardMinY: Double)
+  /** Too little surface left to swipe. The caller refuses; it never swipes under the keys. */
+  case occluded(keyboardMinY: Double, visibleHeight: Double)
+}
+
+/** Where one directional scroll may place its swipe, once the keyboard has taken its share. */
+enum RunnerScrollViewport {
+  /** The frame to plan inside, plus the keyboard top when the swipe was clipped for one. */
+  case swipe(frame: CGRect, keyboardMinY: Double?)
+  /** Nothing to swipe. The caller answers `occlusionRunnerCode` and performs no gesture. */
+  case occluded(keyboardMinY: Double, visibleHeight: Double)
+}
+
+enum ScrollViewportPolicy {
+  /** Below this fraction of the viewport, the clipped band cannot hold a reliable swipe. */
+  static let minVisibleFraction: Double = 0.15
+  /**
+   * A fixed allowance kept above the keyboard's top edge, in points. `keyboard.frame` reports the
+   * key plane, not the input accessory or composer bar riding above it, so a swipe ending exactly
+   * at the reported edge can still land on a bar.
+   */
+  static let accessoryAllowance: Double = 12
+
+  /// The runner's own wire vocabulary, not a shared policy constant: the host keeps it
+  /// `COMMAND_FAILED` and reads it back from `details.runnerErrorCode`.
+  static let occlusionRunnerCode = "SCROLL_KEYBOARD_OCCLUDES_SURFACE"
+
+  /// Clips a scroll viewport to the band above an occluding keyboard, failing open on a frame the
+  /// runner cannot measure: a missing keyboard query is not evidence that the surface is blocked.
+  static func clip(viewport: CGRect, keyboard: CGRect) -> RunnerScrollKeyboardClip {
+    guard isUsable(viewport), isUsable(keyboard) else {
+      return .unobstructed
+    }
+    // A vertical swipe runs along the viewport's centre line, which is the only part of the width
+    // the keyboard has to reach to be struck: a 320pt keyboard centred in an 834pt viewport is 38%
+    // of the width and sits exactly in the path.
+    let swipeCenterX = viewport.minX + viewport.width / 2
+    if swipeCenterX < keyboard.minX || swipeCenterX >= keyboard.maxX {
+      return .unobstructed
+    }
+    let keyboardMinY = keyboard.minY
+    if keyboardMinY >= viewport.maxY || keyboard.maxY <= viewport.minY {
+      return .unobstructed
+    }
+    let visibleHeight = max(0, keyboardMinY - accessoryAllowance - viewport.minY)
+    if visibleHeight < minVisibleFraction * viewport.height {
+      return .occluded(keyboardMinY: keyboardMinY, visibleHeight: visibleHeight)
+    }
+    return .avoided(
+      frame: CGRect(
+        x: viewport.minX,
+        y: viewport.minY,
+        width: viewport.width,
+        height: visibleHeight
+      ),
+      keyboardMinY: keyboardMinY
+    )
+  }
+
+  private static func isUsable(_ rect: CGRect) -> Bool {
+    return [rect.minX, rect.minY, rect.width, rect.height].allSatisfy(\.isFinite)
+      && rect.width > 0 && rect.height > 0
+  }
+}
+
+extension RunnerTests {
+  /// Resolves the frame one directional scroll places its swipe in, and what the keyboard leaves of
+  /// it. Never dismisses: a dismiss drops focus, breaks a `type`/`scroll`/`type` loop, and mutates
+  /// state session-action provenance does not record, so `keyboard dismiss` stays an explicit
+  /// command and this path only ever reduces the space it swipes in.
+  func resolvedScrollViewport(
+    app: XCUIApplication,
+    context: SynthesizedCoordinateContext
+  ) -> RunnerScrollViewport {
+#if os(iOS)
+    // Every scroll reports its decision, including the two ways it avoids reading the keyboard at
+    // all: a policy that forbids the probe, and a probe that finds no keyboard.
+    guard context.allowsKeyboardProbe else {
+      logScrollViewport(decision: "probeSkipped", keyboardMinY: nil, swipeHeight: context.referenceFrame.height, context: context)
+      return .swipe(frame: context.referenceFrame, keyboardMinY: nil)
+    }
+    guard let keyboardFrame = visibleKeyboardFrame(app: app) else {
+      logScrollViewport(decision: "noKeyboard", keyboardMinY: nil, swipeHeight: context.referenceFrame.height, context: context)
+      return .swipe(frame: context.referenceFrame, keyboardMinY: nil)
+    }
+    switch ScrollViewportPolicy.clip(viewport: context.referenceFrame, keyboard: keyboardFrame) {
+    case .unobstructed:
+      logScrollViewport(decision: "unobstructed", keyboardMinY: nil, swipeHeight: context.referenceFrame.height, context: context)
+      return .swipe(frame: context.referenceFrame, keyboardMinY: nil)
+    case .avoided(let frame, let keyboardMinY):
+      logScrollViewport(
+        decision: "avoided",
+        keyboardMinY: keyboardMinY,
+        swipeHeight: frame.height,
+        context: context
+      )
+      return .swipe(frame: frame, keyboardMinY: keyboardMinY)
+    case .occluded(let keyboardMinY, let visibleHeight):
+      logScrollViewport(
+        decision: "occluded",
+        keyboardMinY: keyboardMinY,
+        swipeHeight: visibleHeight,
+        context: context
+      )
+      return .occluded(keyboardMinY: keyboardMinY, visibleHeight: visibleHeight)
+    }
+#else
+    return .swipe(frame: resolvedTouchReferenceFrame(app: app, appFrame: app.frame), keyboardMinY: nil)
+#endif
+  }
+
+#if os(iOS)
+  /// The #2500 diagnostic for a scroll that reports no travel: whether the swipe was clipped, and
+  /// whether the keyboard probe was even permitted. `axHealth` is the first thing to read, because a
+  /// policy that skipped the probe looks exactly like a keyboard that was never found.
+  private func logScrollViewport(
+    decision: String,
+    keyboardMinY: Double?,
+    swipeHeight: Double,
+    context: SynthesizedCoordinateContext
+  ) {
+    NSLog(
+      "AGENT_DEVICE_RUNNER_SCROLL_VIEWPORT kind=scroll axHealth=%@ keyboardPolicy=%@ decision=%@ keyboardMinY=%@ swipeHeight=%.1f",
+      context.accessibilityHealth.rawValue,
+      context.keyboardPolicy.rawValue,
+      decision,
+      keyboardMinY.map { String(format: "%.1f", $0) } ?? "none",
+      swipeHeight
+    )
+  }
+#endif
+}
+
+#if AGENT_DEVICE_RUNNER_UNIT_TESTS
+private struct ScrollViewportPolicyFixture: Decodable {
+  struct Frame: Decodable {
+    let x: Double
+    let y: Double
+    let width: Double
+    let height: Double
+
+    var cgRect: CGRect {
+      CGRect(x: x, y: y, width: width, height: height)
+    }
+  }
+
+  struct Constants: Decodable {
+    let minVisibleFraction: Double
+    let accessoryAllowance: Double
+  }
+
+  struct Expected: Decodable {
+    let kind: String
+    let viewport: Frame?
+    let keyboardMinY: Double?
+    let visibleHeight: Double?
+  }
+
+  struct TestCase: Decodable {
+    let name: String
+    let viewport: Frame
+    let keyboard: Frame
+    let expected: Expected
+  }
+
+  let constants: Constants
+  let cases: [TestCase]
+}
+
+extension RunnerTests {
+  /// Golden parity table (#2500): every case in contracts/fixtures/scroll-keyboard-policy.json must
+  /// agree with the vitest twin. Add cases there, never fork the rule.
+  func testScrollViewportKeyboardClipMatchesGoldenParityTable() throws {
+    let fixture = try loadScrollViewportPolicyFixture()
+    XCTAssertFalse(fixture.cases.isEmpty, "parity table must not be empty")
+    for testCase in fixture.cases {
+      let clip = ScrollViewportPolicy.clip(
+        viewport: testCase.viewport.cgRect,
+        keyboard: testCase.keyboard.cgRect
+      )
+      switch testCase.expected.kind {
+      case "unobstructed":
+        XCTAssertEqual(clip, .unobstructed, testCase.name)
+      case "avoided":
+        let expectedFrame = try XCTUnwrap(testCase.expected.viewport, testCase.name).cgRect
+        let expectedMinY = try XCTUnwrap(testCase.expected.keyboardMinY, testCase.name)
+        XCTAssertEqual(
+          clip,
+          .avoided(frame: expectedFrame, keyboardMinY: expectedMinY),
+          testCase.name
+        )
+      case "occluded":
+        let expectedMinY = try XCTUnwrap(testCase.expected.keyboardMinY, testCase.name)
+        let expectedVisibleHeight = try XCTUnwrap(testCase.expected.visibleHeight, testCase.name)
+        XCTAssertEqual(
+          clip,
+          .occluded(keyboardMinY: expectedMinY, visibleHeight: expectedVisibleHeight),
+          testCase.name
+        )
+      default:
+        XCTFail("unknown expected kind `\(testCase.expected.kind)` in \(testCase.name)")
+      }
+    }
+  }
+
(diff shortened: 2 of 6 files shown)


thymikee:
The clipping rule still allows swipes into a floating keyboard at 0fb3f39. Both `clipScrollViewportAboveKeyboard` and its Swift counterpart ignore keyboards covering less than half the viewport width, but vertical scrolling uses the viewport's center line. A centered 320px keyboard in an 834px viewport covers that line despite being only 38% wide. Check whether the planned swipe intersects the keyboard, and add that centered case to the shared fixture.

Required live evidence is also incomplete: please exercise Android with an adjustPan activity and the iOS `scroll_keyboard_occludes_surface` refusal, including confirmation that no gesture was dispatched. The reported iOS clipping run and unit parity table do not cover those paths.

The change has 1,561 gross lines and adds 8.2 kB unpacked. Please split it below the repository's 1,000-line PR budget and explain why the smaller owning change is insufficient, or obtain an explicit budget exception. The iOS smoke failed while waiting for the Automation lab screen, before exercising keyboard clipping; that appears unrelated. The separately reported mutation-model failure was reproduced at the base and also appears unrelated.

thymikee:
Addressed at `c7f4bd2b7b`.

**Centre-line gate.** The width fraction is gone from `clipScrollViewportAboveKeyboard` and its Swift twin; the gate is now `swipeCenterX` inside the keyboard's x-range, which is what a vertical swipe actually travels. Your case is a shared fixture row, `centred floating keyboard under half the viewport width sits in the swipe's path` (834x1112 viewport, 320x332 keyboard centred at 417, expected `keyboardMinY: 780`), asserted by vitest and by `testScrollViewportKeyboardClipMatchesGoldenParityTable` against the same JSON. Two near-neighbours kept out of scope and named in the body: `frameAvoidingKeyboard` (`RunnerTests+Interaction.swift:789`) still uses `intersection.width / frame.width >= 0.5` for tap/drag reference frames, and `keyboardAvoidingDragPoints` (`:702`) hard-codes the same 12pt allowance and fails open where scroll refuses.

**Android adjustPan, live.** Pixel 7 emulator, Contacts editor (`sim={adjust=pan}` in `dumpsys window windows`), LatinIME bound: helper reports the IME window `[0,1517][1080,2400]` against a 1080x2400 app window, and `scroll down --json` returns `referenceHeight: 1505`, `keyboardAvoided: true`, `keyboardMinY: 1517`, swipe `1205 -> 301` — it used to swipe `1920 -> 480` into the keys. Two things only came out on device: `UiAutomation.getWindows()` returns an empty list until interactive window retrieval is on (`c7f4bd2b7b`), and `open` re-activates agent-device's headless `imehelper`, which draws no keyboard, so a real IME has to be bound *after* `open` or there is no IME window to avoid. Also added an iPadOS 26 windowed-Safari reading (`834x913` window in a `1219pt` display, `keyboardAvoided: true`, `keyboardMinY: 925`) to show the clip tracks the app window and not the display.

**Still open: the live `occluded` refusal.** Not reachable where I can reach: iPhone leaves 583 of 874 above the keys (65%), and iPadOS 26 keeps a window clear of the keyboard (window bottom 913 vs keyboard top 925), so it needs a window dragged down to about the keyboard's height from the host desktop — the macOS desktop session timed out on this host before I could drag. Table coverage on both sides for now; say the word and I'll keep pushing on the host-side resize.

**Size.** 1,443 gross (817 production) after the trims listed in the body, so still over 1,000. The clean split is this PR minus Android (~260 lines: helper window read, `input-actions` clip, their tests) as a follow-up. Prefer the split or an exception?

thymikee:
Split per the size ask. This branch is now the contract plus the Apple runner path: 17 files, **980 gross lines**, under the 1,000 budget. The Android half (helper IME read, `input-actions` clip, their tests, 463 lines) is #2514, stacked on this branch.

`pnpm check` green at `faa75e7860` on this branch and on the Android branch. History was rebuilt for the split, so the earlier head is gone; nothing else changed in content — the two branches together differ from the previous head only by the split itself and by one `commands.md` sentence that now claims iOS only until #2514 lands.

One thing the rebuild surfaced: the previous head also deleted a nine-line comment in `interaction-guarantees.ts` about the offscreen rescue override's separate-request read (#2448). That was collateral damage in my own edit, not intent, and it is not carried into either branch — main's comment is intact here.

Still open and unchanged from my last note: the live `scroll_keyboard_occludes_surface` refusal needs an iOS window dragged down to about the keyboard's height from the host desktop (iPhone leaves 65% above the keys, iPadOS 26 holds the window clear at 913 vs 925), and the macOS desktop session timed out here. Table coverage on both sides meanwhile.

thymikee:
The floating-keyboard width check and size split are resolved at faa75e7, but the clipped frame now also changes the native coordinate-rotation basis. RunnerTests+CommandExecution.swift:2004 passes the shortened frame into scrollContext.withReferenceFrame; synthesizedDragAt uses its height for landscapeRight rotation. Clipping an 834-point frame to 588 points shifts native x by 246 points, so the gesture misses its planned path. Preserve the original coordinate frame for rotation and use the clipped frame only for swipe planning; cover a landscape keyboard case through dispatch. Current-head checks pass and there are no conflicts. The reported live iOS refusal/no-dispatch validation is still missing.

thymikee:
Fixed at `b6ea4e6e5d`. You had the mechanism exactly: `resolvedScrollViewport` returned one frame and dispatch used it for both jobs, so the clip reached the rotation basis.

The resolved viewport now names both frames — `.swipe(planFrame:coordinateFrame:keyboardMinY:)` — and the gesture comes out of one `gestureDispatch(direction:amount:pixels:)` decision, so the band the plan is built inside and the frame the coordinates rotate against are different values the dispatch cannot swap. `ScrollViewportPolicy.frames(referenceFrame:clip:)` is where they are split, and the plan keeps running against the band.

`testScrollViewportDispatchKeepsTheUnclippedFrameAsItsCoordinateRotationBasis` runs a landscape viewport through that dispatch: 1210x834 with a 246pt key plane, band `576`, `coordinateFrame` still `834`, planned swipe ends at or above `keyboardMinY - 12`, and the native x mapped through the coordinate frame differs from the band-basis one by exactly the 258pt the keyboard took. Wired into the iOS lane's `-only-testing` list.

Proved the test bites rather than asserting my own shape: handing the band to `coordinateFrame` makes it fail with `("(0,0,1210,576)") is not equal to ("(0,0,1210,834)")` and a `0.0` vs `258.0` shift.

One verification correction worth recording, because it wasted my earlier confidence: `xcodebuild test-without-building -only-testing:...` **executes 0 tests and still reports the suite as passed** unless the build was made with `AGENT_DEVICE_XCUITEST_INCLUDE_UNIT_TESTS=1`, which is what defines `AGENT_DEVICE_RUNNER_UNIT_TESTS`. My earlier local "Swift tests passed" lines were those empty runs; the iOS lane was always real, since `.github/workflows/ios.yml` sets the flag. Rebuilt with `AGENT_DEVICE_XCUITEST_INCLUDE_UNIT_TESTS=1 pnpm build:xcuitest:ios` and these 4 tests genuinely executed and passed:

```
Executed 4 tests, with 0 failures (0 unexpected) in 0.051 seconds
```

`pnpm check` green at `b6ea4e6e5d`. Android has no analogue: its swipe goes to the helper as absolute screen pixels with no orientation rotation on the path.

thymikee:
Both halves of your comment are in: the rotation-basis fix is `b6ea4e6e5d` (detail in my previous comment) and the Android multi-window selection is `67c6ba6bbb` on #2514.

**Size needs your call, and I would rather ask than degrade the fix.** The split head you cleared was 980 gross. The rotation-basis fix lands **1,125** (+145): ~55 lines of the landscape dispatch test you asked for, and the rest the `planFrame`/`coordinateFrame` pair plus the single `gestureDispatch` decision that makes them unswappable. #2514 is 612, so the pair is 1,737 against a 1,000-per-PR budget.

Three ways through, priced:

1. **Exception for the +125.** The test is what makes this defect impossible to reintroduce silently, and the typed pair is where the reviewer-visible invariant lives. This is my recommendation.
2. **Minimal variant, ~1,003 gross.** Stop re-basing the context at all (delete the `withReferenceFrame(frame)` call, ~4 lines) and keep only a plan-geometry test. Under budget, and the bug is gone — but nothing guards the basis, so the next person who threads a clipped frame into the context reintroduces the 258pt shift with green tests. That is the failure mode your comment caught, so I do not recommend trading the guard for the line count.
3. **A third PR for the contract alone.** I built and measured this: it is not independently green. `check:production-exports` reports `clipScrollViewportAboveKeyboard` and `scrollKeyboardOccludesSurfaceError` as unused production exports, because a shared rule with no runtime consumer has no consumer. Suppressing that is suppressing the signal, so the contract has to land beside at least one runtime.

Say 1 and I leave it as is; say 2 and I will push the reduced variant in one commit.

thymikee:
The iOS Smoke lane failed at `b6ea4e6e5d` on one alert test, retried three times:

```
RunnerTests.testAlertAcceptDoesNotActivateAReplacementWithASharedButton()
Executed 94 tests, with 3 failures (0 unexpected) in 295.754 seconds
```

Not this diff:

- #2514 carries the identical Swift tree (this branch is its base) and its three Smoke lanes all passed, same `-only-testing` list, which includes that alert test.
- `main`'s own most recent `ios.yml` run fails the same alert-replacement family: `RunnerTests.testAlertCannotProveAnIdenticalReplacementAndDoesNotActivateIt`.
- Nothing here touches alert resolution; the scroll dispatch and the rotation-basis change are the only Swift edits, and the four scroll/policy tests in this lane are green.

Reran the failed job for the record.

thymikee:
The rotation basis is fixed at b6ea4e6, but the response now reports the coordinate frame instead of the clipped planning frame. executeSynthesizedDragGesture builds its DragVisualizationFrame from plan.referenceFrame (RunnerTests+CommandExecution.swift:2492-2513), while attachingScrollViewportEvidence only adds keyboard fields. A swipe planned in a 571-point band therefore reports the full 874-point referenceHeight. Keep the full frame for synthesis and report the clipped plan dimensions; assert the returned scroll response in the regression. Current checks pass and there are no conflicts. Live iOS refusal/no-dispatch evidence remains missing, and the diff is now 1,125 gross lines, so the size budget needs a split or explicit exception again.

thymikee:
All three points addressed.

**Reported frame** — fixed in `8a67c377cf`. The keyboard evidence travels with the dispatch instead of being bolted onto a response that had already chosen its frame: `ScrollGestureDispatch.attachingEvidence(to:)` names `planFrame` in `referenceWidth`/`referenceHeight` and leaves `coordinateFrame` to synthesis, and `attachingScrollViewportEvidence` is deleted, so there is one place that decides what a scroll's axis is. Synthesis still rotates in the full frame — the landscape assertion for that is untouched.

The regression now asserts the assembled response, not just the plan: `testScrollViewportDispatchKeepsTheUnclippedFrameAsItsCoordinateRotationBasis` builds the landscape dispatch (1210x834 viewport, keyboard top 588, band 576) and checks `referenceHeight == 576`, `referenceWidth`, `keyboardMinY`, `keyboardAvoided`. Deleting the two assignment lines reproduces your symptom as a red test — `XCTAssertEqual failed: ("Optional(834.0)") is not equal to ("Optional(576.0)") - the payload names the band the plan ran inside, not the synthesis frame`.

Live iPhone 17 Pro, Form screen, field focused, `scroll down --json`: `referenceWidth: 402`, `referenceHeight: 571`, `keyboardAvoided: true`, `keyboardMinY: 583`, swipe `y1: 472 -> y2: 100`, `pixels: 371`, and the runner logged `decision=avoided keyboardMinY=583.0 swipeHeight=571.0`. The same command with no keyboard up: `referenceHeight: 874`, `y1: 721 -> y2: 153`, `pixels: 568`. Coordinates and reference frame are the same space in both answers.

**Size** — resolved by splitting rather than by asking you for an exception I could have avoided. My earlier claim that a contract-only PR fails `check:production-exports` was a false failure: I read a stale `dist` from another branch, and that gate also prints a pre-existing 63-finding report while exiting 0. Measured properly, the seam is the language boundary:

- #2537 — shared rule, golden table, typed refusal, response keys, CLI/MCP/docs surface: 11 files, 615 gross, `pnpm check` green.
- #2503 (this PR, now based on #2537) — the runner clip, its lane wiring, the response evidence: 6 files, 536 gross, `pnpm check` green.
- #2514 — Android: 13 files, 612 gross, unchanged.

The split is tree-exact: the two branches reassemble to the same tree the 1,151-line branch had (`git diff` between them is empty), and each half was gated on its own branch after a rebuild. Also rebased onto current main (#2511) while I was in there.

**Live occlusion refusal** — still not reproduced, and I want to be explicit that this one remains open rather than implied. The band above the keys cannot fall under 15% on an iPhone (583 of 874 = 65%), and iPadOS 26 keeps an app window clear of the keyboard (window bottom 913, keyboard top 925), so reaching it needs an iPad app window dragged down to roughly the keyboard's own height from the host desktop; the macOS desktop target timed out at the daemon budget here before the drag landed. The command that would close it is `agent-device scroll down --json` against such a window, on a host session that answers desktop snapshots. Risk left open is narrow: the refusal returns from the same clip decision the runs above exercised, it returns before any gesture is dispatched, and the table row `adjustPan window with only 88px of 900px left above the IME is refused` plus the platform tests pin that no swipe is planned.

thymikee:
The reported-frame correction is clean on 8a67c377: successful responses now report the planned clipped frame while rotation retains the full coordinate frame. The split also brings this layer to 536 changed lines. The reported live iPhone keyboard-clipping run supports the successful path. Readiness still needs live iOS keyboard-occlusion refusal evidence showing no gesture dispatch; contract tests do not establish that device path. This also depends on #2537.

thymikee:
I went after the live occlusion refusal three ways today and the answer is structural, not a harness gap: no iOS system keyboard can put 85% of an app frame under the keys.

- **iPhone 17 Pro, full screen.** Forcing the largest accessibility size ("Larger Accessibility Sizes" on, slider at 100%, `UIPreferredContentSizeCategoryName = UICTContentSizeCategoryAccessibilityXXXL` read back from the device's `com.apple.UIKit` domain, app layout visibly at AX5) left the keyboard exactly where it already was: `keyboardMinY: 582.9999999999999`, `referenceHeight: 570.9999999999999`, swipe `471 -> 99`, `pixels: 371`. Same clip as at the default text size, so Dynamic Type does not drive keyboard height and the band stays at 65% of 874.
- **iPadOS 26 windowed Safari.** Measured 834x913 window in a 1219pt display, keyboard top 925. With the window's bottom edge on the display's bottom edge the band is 607 of 913 = 66%. Crossing 15% needs a window no taller than ~360pt and iPadOS does not offer one.
- **Accessibility Keyboard.** Not reached — Settings navigation at AX sizes did not get past the Accessibility list, and the panel would need to exceed ~743pt of 874 to change the arithmetic.

That makes the refusal reachable on device only through an input surface the app sizes itself. So it's a call rather than more effort on my side:

1. **Land a fixture**: a tall custom `inputView` on one text field in `examples/test-app`, then `scroll down --json` against it on a booted iPhone answers `scroll_keyboard_occludes_surface` and the runner log shows `decision=occluded` with no synthesized-gesture line for that command — the live no-dispatch proof you asked for. I would add it as its own small PR stacked under this one, because the test app is a different module group than the scroll family, and re-run the lane.
2. **Accept it as defensive for iOS**: the branch stays covered by the parity table row `adjustPan window with only 88px of 900px left above the IME is refused` and the runner/TS tests, and the PR body records the ceiling above so nobody spends a day rediscovering it. Android does reach it for real and is proven live on the emulator (#2514).

Say which and I will execute it. The PR body's "Unreproduced live" section now carries today's numbers, and the device state from the attempts is reverted: accessibility category key deleted, `ConnectHardwareKeyboard` back to its original value, the iPad simulator deleted.

thymikee:
Evidence hunt is closed after testing the app-side route on device. The temporary 800pt custom `inputView` visibly attached, but XCUITest exposed no `Keyboard` element and the runner saw no keyboard: `visibleKeyboardFrame` queries `app.keyboards.firstMatch`, which covers system keyboards only. The prototype was removed. iPhone/iPad measurements put every available system-keyboard path above the 15% floor, so the iOS body now states `decision=occluded` as contract/unit/synthesis-path evidence and explicitly does not claim a live refusal response. Device state and test-app changes were reverted.

thymikee:
The additional device attempts document the limit clearly: no live refusal was reproduced, and the custom input view did not expose a Keyboard element to the current query. The successful clipping path remains validated, but those measurements do not establish that refusal is impossible on every supported iOS configuration. Code review remains clean on 8a67c377. Readiness still needs either live refusal/no-dispatch proof or an explicit maintainer decision to accept the defensive branch with contract and unit evidence only.

thymikee:
Rebased on the updated #2537 (`b5a83ff682`); no content change in this layer.

PR #2506, opened 2026-09-11
fix(ios): gate alert activation on a fresh hittable read

thymikee:
## Summary
Hardens the single, never-repeated alert activation. A snapshot can expose an alert button a beat before the owning app has made it hittable, and a starved CI host widens that window. The one activation tapped into that gap, so the press was dropped, the presentation never changed, and the whole budget rode an unchanged alert to `ALERT_DEADLINE_EXCEEDED` with `First actions: 0` — the signature of the flaky `alert`-replacement runner regressions on `main`.

Now `handleAlert` waits for a fresh `exists && isHittable` read before the activation, still activating at most once (preserves #2326's no-retry guarantee).

One file, runner-test target only. No production, CLI, or cross-platform surface change.

## Validation
Commit `0d89661f18`.
- iOS runner XCTest: full targeted PR list (89 tests) green, `0 failures`; the three `alert`-replacement regressions pass.
- macOS runner build: `TEST BUILD SUCCEEDED` (shared alert source, guards the #2326 macOS-compile pitfall).
- `pnpm check:affected --run`: `xctest-selection` and `packaged-runner-swift` pass; Swift builds + device lanes are GitHub-authoritative (running on this PR).

Residual risk: the flake is CI-contention-specific (~10% of recent `main` runs, ~13x slowdown) and did not reproduce on warm machines, so the green device lanes confirm no regression rather than a reproduced-before/after. The `webview-remote-content` `RUNNER_BUSY` failures were the separate dominant cause and are addressed by the already-merged #2493.

### DIFF
--- a/.github/workflows/ios.yml
+++ b/.github/workflows/ios.yml
@@ -182,6 +182,7 @@ jobs:
             -only-testing:AgentDeviceRunnerUITests/RunnerTests/testAlertDismissDoesNotActivateAReplacementWithTheSameTitle \
             -only-testing:AgentDeviceRunnerUITests/RunnerTests/testAlertCannotProveAnIdenticalReplacementAndDoesNotActivateIt \
             -only-testing:AgentDeviceRunnerUITests/RunnerTests/testAlertDeadlineBeforeActivationLeavesTheOriginalUntouched \
+            -only-testing:AgentDeviceRunnerUITests/RunnerTests/testAlertHittableProbeCompletingAfterDeadlineLeavesTheOriginalUntouched \
             -only-testing:AgentDeviceRunnerUITests/RunnerTests/testSystemModalProbeSliceSharesAndClampsToPlanDeadline \
             -only-testing:AgentDeviceRunnerUITests/RunnerTests/testDispatchRecoverySkipsBookkeepingWhileXCTestChannelOccupied \
             -only-testing:AgentDeviceRunnerUITests/RunnerTests/testBoundedSystemModalProbeTimeoutRecoversThenReleasesOnDrain \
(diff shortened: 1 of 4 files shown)


thymikee:
The new readiness wait can activate an alert after the command deadline at 0d89661. It checks the deadline before the synchronous `exists && isHittable` query, but returns true even if that query finishes late; `handleAlert` then taps without checking time again. Recheck the deadline after the probe and add a regression where the probe completes late, asserting that no alert action occurs.

The earlier iOS smoke failure was an `automation-longpress` visibility check after scrolling and appears unrelated to this alert change; the rerun is pending. The reported green alert tests do not cover expiry during the new readiness query.

thymikee:
Good catch — fixed in 7bbb1488. The hittable read is itself a synchronous query, so it now forfeits instead of buying back the single activation: `waitUntilAlertButtonHittable` returns `Date() < deadline` after the probe rather than unconditional `true`, so a read that lands late never reaches `activateElement`.

Regression `testAlertHittableProbeCompletingAfterDeadlineLeavesTheOriginalUntouched` drives `handleAlert` with a probe overridden to complete past the deadline and asserts the fixture's own `First actions: 0; replacement actions: 0` label (typed activation count, no hint-text sniffing). Red/green proved locally: with the pre-fix `return true` it fails at `First actions: 1`; with the recheck it passes. Full targeted list green (90 tests, 0 failures); macOS runner build green; new method selected in the iOS PR lane (check:xctest-selection: 90 selected, 0 reachable by no lane).

thymikee:
The post-probe deadline check fixes the late activation at d1e1d1e. The selected regression forces a late successful probe and verifies the deadline error with zero original or replacement actions; current-head checks pass, including iOS and macOS, and there are no conflicts. Ready for human review and merge. Please also refresh the description’s old validation SHA, file count and test count to match this revision.

PR #2507, opened 2026-09-11
fix(test): preserve colliding diagnostic artifacts

PLASMA-FR:
## Summary

`agent-device test ./flows --artifacts-dir ./out` could silently lose diagnostic files: copying two `device.log` files generated `device-2.log`, which a later artifact with that name overwrote. Diagnostics could also overwrite `replay.ad` or the timing trace, or disappear under the final manifests.

This completes the existing attempt-artifact retention contract from #266 and the replay documentation. The artifact writer now chooses unused numbered names, reserves manifest names, preserves files already in the attempt directory, and deduplicates failure logs. `result.txt` lists the retained filenames. Five files change, including regressions, help, docs, and changelog.

## Validation

Tested `8f6acb16e39cea991df6f1611a43aef5edcbc8a2` with Node 24.13.1 and pnpm 11.17.0 on Linux:

- Nine new regression cases fail without the fix; all 15 artifact tests pass with it.
- `pnpm exec vitest run` over the focused replay/suite/startup files: 87 tests pass.
- Remote artifact materialization: two provider integration tests pass.
- Public scheduler probes preserve all seven diagnostics, source, timing trace, manifests, and suite outcome for passing and failing attempts.
- `pnpm build`, `pnpm check:command-docs`, and built `test --help` pass.
- `pnpm check:affected --run`: all runnable checks pass, including 2,801 tests across 384 files.

CI pending; no local device run. The changed boundary copies files after an attempt completes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Fixed: Replay test artifacts with colliding filenames retain distinct copies without overwriting
+  other diagnostics, replay sources, timing traces, or attempt manifests.
 - Fixed: Custom test reporters reject invalid exit codes, including values such as `256` that
   could wrap to success and hide a failing suite. `getExitCode` accepts integers from `0` to `255`
   or `undefined`; JSON output reports an invalid code as one `INVALID_ARGS` error.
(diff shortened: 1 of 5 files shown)


thymikee:
No code findings at 8f6acb1. Artifact allocation preserves colliding diagnostics, existing attempt files and manifests, with regression coverage for ordering and duplicate logs. There are no conflicts; ready for human review. The iOS smoke check still fails because a depth-1 snapshot falls back to XCTest and carries unexpected tree-quality metadata; this appears unrelated to artifact copying, but the check needs to pass before merge.

PR #214, opened 2026-03-13
feat: add `diff screenshot` command for pixel-level image comparison

dannyhw:
Uses pngjs to decode pngs and then compares each pixel with euclidean distance.
Mirrors the agent-browser CLI interface.

Usage:
  agent-device diff screenshot --baseline before.png [--out diff.png] [--threshold 0.2]

try it out with

use scroll down for diff
```
node bin/agent-device.mjs open Settings --platform ios 
sleep 1
node bin/agent-device.mjs screenshot before.png 
node bin/agent-device.mjs scroll down 0.5
node bin/agent-device.mjs diff screenshot --baseline before.png --out diff.png
```

take two equal screenshots

```
node bin/agent-device.mjs open Settings --platform ios 
node bin/agent-device.mjs screenshot before.png 
node bin/agent-device.mjs diff screenshot --baseline before.png --out diff.png
```

image

Example diff
diff

PR #2508, opened 2026-09-11
fix(diff): include maximum RGB distance at threshold one

PLASMA-FR:
## Summary

`agent-device diff screenshot --baseline black.png white.png --threshold 1` reported 100% different pixels instead of a match. The maximum RGB distance rounded one floating-point step below the per-pixel calculation.

Use the same square-root calculation for the maximum. This completes the existing Euclidean comparison and saved-image workflow from #214 and #404; thresholds below 1 still detect maximum contrasts. Five files change: the calculation, regression tests, command help, and documentation.

## Validation

Tested `cc92eb97d6c07314b87641aa19ff10ae516def59` with Node 24.13.1 and pnpm 11.17.0 on Linux:

- Six new regression cases fail without the fix.
- Focused screenshot, runtime, CLI, PNG-worker and startup suites: 82 tests pass.
- `pnpm build` and `pnpm check:command-docs` pass (12 documentation tests).
- Fourteen built-CLI text/JSON cases pass, including the nearest threshold below 1, dimension mismatches, and stale artifact cleanup.
- `pnpm check:affected --run`: all runnable checks pass, including 3,062 tests across 414 files.

CI pending. Saved-image comparisons require no device; no local native run.


thymikee:
No code findings at cc92eb9. The maximum RGB distance now follows the same rounding as the pixel comparison, and the regressions cover threshold 1 and the boundary immediately below it. Current-head checks pass and there are no conflicts; ready for human review and merge.

Issue #2492, opened 2026-09-11
iOS Simulator bridge and web content: follow-ups to the #2484 fence (content-class gate, probe circuit, WebContent spike)


thymikee:
Follow-up to #2484 / #2486. That PR is a fence: the Simulator AX bridge refuses a tree that ends at a web view's `AXRemoteElement` leaf (`remote-content-boundary`) and the route serves XCTest for the rest of the app generation. This issue scopes what makes the class of defect impossible or cheaper, in the order the design consultation ranked them (two adversarial reviews plus a `codex exec` architecture pass; verdicts below are the agreed ones).

## Why we regressed

- #2279 judged the bridge on node/depth parity, ownership, and latency against native apps. The one fact that mattered — the reader snapshots one process (`apple/snapshot-bridge/README.md`) — was never a declared capability or a test.
- The failure was silent: a well-formed tree with a viewport and no truncation passes every protocol check, and the quality verdict knows sparse roots, not an on-screen web view with nothing inside.
- No gate drove web content through the default route. The WebView lab (#1397) was a manual lab. `http(s)` URLs open Safari by design, but Safari was never in the acquisition validation set.

### 1. Content-class gate (high value, moderate cost)

Extend the fixture-backed iOS E2E lane with a fixed screen set captured through the **default** route, asserting coarse content presence, backend-independent: native, WKWebView (exists: `smoke:webview-remote-content`), Safari on a harness-served local page (Node http server in the scenario; no network), a system sheet. Each row opens a **fresh app generation** so an earlier fallback cannot mask a broken bridge path. Add one deterministic page action with an observable result (a page ref `press`/`fill`) to the WebView row — today it checks content and warnings only (`test/integration/ios-simulator-e2e/live-webview-remote-content.ts`). Keep fallback-policy assertions (`remote-content-boundary` / `circuit-disabled`) separate from content assertions.

### 2. Launch-observation probe opens the circuit (high value, low cost)

[...]

PR #2510, opened 2026-09-12
feat(snapshot): disclose a cut capture on every platform; raise the iOS bridge node cap to 5000

thymikee:
Part of #2492 (task 5). Two changes, both measured first.

## One disclosure for a cut capture, every platform

Every backend already reports the same fact — `truncated: true` on the snapshot — but only in JSON. The text path said nothing: an agent read a screen missing its footer, tab bar, or the items after a long list as complete. The backends walk the tree in document order, so what falls off is what comes last, on screen or not.

`truncatedCaptureWarning` in capture-kit renders one warning from that flag, and the single cross-platform warnings assembly in the snapshot runtime calls it. Android helper (5000-node cap), iOS Simulator bridge, XCTest runner, and the web provider all flow through it. The limit and dimension stay backend-side; the copy names neither, and tells the agent what to do: navigate or scroll so fewer elements render, re-run, use `screenshot` for the rest.

Producers that cannot observe truncation keep `truncated` unknown (#2188 invariant 5) and get no warning.

## iOS Simulator bridge node cap 1500 → 5000

Measured on iPhone 17 Pro with a synthetic unvirtualized list in the fixture dev client (8 regular snapshots per cell, bridge `acquire` duration from `--debug` diagnostics):

| tree | cap 1500 | 3000 | 5000 | 10000 |
| --- | --- | --- | --- | --- |
| 600 rows (~3.5k raw nodes) | 969 ms, truncated | 976 ms, truncated | 962 ms, complete | 970 ms, complete |
| 1500 rows (~8k raw nodes) | 2378 ms, truncated | – | 2344 ms, truncated | 2404 ms, complete |

The native read fetches the whole tree regardless of the cap; the cap only stops conversion. Cost is the tree itself (~0.25 ms per raw node). At 1500 the screen's on-screen footer (drawn last) was absent; at 5000 present. 5000 matches the Android helper. The 4 MB response bound did not bite at 8k nodes.

## Not here

- Falling back to XCTest for a capture the bridge still cuts at 5000: a policy decision (#2492 task 5), not a performance one — such a screen already costs over a second per capture.
- Acquisition-rooted `--scope` on the bridge as a real override; today `--scope` is presentation-only on iOS, so the warning does not promise it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,16 @@
 
 ## Unreleased
 
+- Changed: a capture that a backend cut at one of its limits now says so in the snapshot's
+  warnings, on every platform, instead of only setting `truncated: true` in JSON. The text path
+  had no disclosure at all, so an agent read a screen missing its footer, tab bar, or the items
+  after a long list as complete — the backends walk the tree in document order, so what falls
+  off is what comes last, on screen or not. One shared warning renders from the shared flag; the
+  limit and dimension stay backend-side.
+- Changed: the iOS Simulator AX bridge caps a capture at 5000 nodes, up from 1500, the Android
+  helper's bound. Measured on a synthetic 600-row screen, acquisition time did not move with the
+  cap (the native read fetches the whole tree; the cap only stops conversion) while the 1500 cut
+  dropped the screen's on-screen footer.
 - Fixed: Android snapshots carry the accessibility `selected` state an app sets on a control, so
   `is selected`, a `selected=true` selector, and a Maestro `selected:` qualifier work on Android
   (#2462). The helper never serialized the attribute, and the host reads only the helper's XML, so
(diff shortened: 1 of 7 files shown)


thymikee:
The new warning misclassifies incomplete captures at a552b16. quality-warnings.ts:43-47 treats every truncated flag as a backend-limit cut, but Android presentation failures and iOS accessibility-unavailable responses also set that flag. Those failures now get misleading advice about missing later elements and scrolling. Use wording that describes incomplete capture without inventing its cause, or select limit-specific advice from typed evidence; cover a real sparse/presentation-failure result. Current-head checks pass and there are no conflicts. The separate bridge publish-versus-fallback policy remains deferred, as the description states.

thymikee:
This head now conflicts with main. Resolve the conflict alongside the already-reported warning correction before merging; the code finding at a552b16 is unchanged.

PR #2511, opened 2026-09-12
refactor(snapshot): drop the bridge truncation dimension; stop promising --scope for depth caps

thymikee:
Stacked on #2510 (merge that first; this diff is the top commit). Second-order cleanup after the shared truncation disclosure.

## Drop the truncation dimension inference

The bridge adapter guessed a truncation `dimension` (`nodes` / `depth` / `payload`) and `limit` from counts and put them in the acquisition residue. Nothing rendered them: residue never reaches the CLI, and the only consumer was the comparison-identity string in `ios-snapshot-planning.ts`, where `{ kind: 'truncated' }` yields the same comparability (two cut captures compare equal; a cut and a complete one do not). The runner never produced the field. `IosTruncationDimension` and the inference function are gone; `createAcquisition` and `createAcquisitionResidue` lose the parameters that only fed it.

## Depth-cap warning stops promising `--scope`

The copy said "Re-run with `--depth N --scope <container>` only if you need deeper content." On iOS, scope narrows the presented view; acquisition is scope-blind, and the private AX ladder already extends depth on its own. The warning now says the backend capped the tree and what to do instead (navigate so the content sits higher; screenshot for the rest), the same shape as the new truncation disclosure.

Tests: residue fixtures updated; the three depth-copy assertions updated. Lint, fallow, typecheck green.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,8 @@
   had no disclosure at all, so an agent read a screen missing its footer, tab bar, or the items
   after a long list as complete — the backends walk the tree in document order, so what falls
   off is what comes last, on screen or not. One shared warning renders from the shared flag; the
-  limit and dimension stay backend-side.
+  limit and dimension stay backend-side. The depth-cap warning no longer suggests `--scope` as a
+  way to read deeper: on iOS, scope narrows the presented view and acquisition stays scope-blind.
 - Changed: the iOS Simulator AX bridge caps a capture at 5000 nodes, up from 1500, the Android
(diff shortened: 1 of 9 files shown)


thymikee:
No code findings in the cleanup layer at 2e9dedb. Removing inferred truncation dimensions preserves the complete-versus-truncated comparison, and the depth warning no longer promises an acquisition change from --scope. There are no conflicts, but merge depends on the warning correction in #2510. The iOS smoke failure is automation-longpress not becoming visible after scrolling; it appears unrelated to this layer’s residue/copy changes and still needs resolution before merge.

PR #2514, opened 2026-09-12
fix(android): keep a scroll's swipe out of the IME window

thymikee:
## Summary

Third layer of #2500, stacked on #2503 (the shared rule and golden table land in #2537, the iOS runner in #2503). Android had no keyboard input into a scroll's plan at all: an `adjustPan` or `adjustNothing` activity reports window bounds that already run under the IME, so the swipe was planned into the keys, the surface never moved, and the edge loop reported a stuck container.

The helper now reports the input method window edges beside the application window, in the same absolute screen pixels, and `scroll` clips its band with the shared `contracts/scroll-gesture` rule or refuses with `scroll_keyboard_occludes_surface` when the keys own the surface. 13 files, 612 gross lines.

Depends on #2537 and #2503. Do not merge before them.

## Behaviour

- An older helper returns no keyboard keys and a provider-supplied viewport has no IME channel. Both read as "nothing to avoid", which is what the shared rule already does with a missing frame; neither becomes a refusal.
- `keyboardAvoided` and `keyboardMinY` appear only when the clip ran, matching the Apple side, so the two platforms answer with the same shape.
- The keyboard is never dismissed: that drops focus and breaks a `fill`/`scroll`/`fill` loop. `keyboard dismiss` stays explicit.
- Multiple input method windows are unioned, not narrowed to one. A composer bar and its key plane can arrive as separate `TYPE_INPUT_METHOD` windows, and the key plane is the larger rectangle, so keeping the largest one discarded the composer's earlier top edge and the clipped band still ended inside the composer. The read unions the input method windows that cross the swipe's centre line and overlap the application window, which is the same line the shared clip rule tests; an input method window beside that line, such as a floating candidate strip at the right edge, no longer shortens the band either.
- The selection runs on plain window edges because `android.graphics.Rect` is a stub off-device (`new Rect(...)` throws `Stub!`), which is what lets `GestureViewportReaderTest` cover the two-window case in both orderings as part of the build's test run.

## Validation

`pnpm check` green at `5d55192f3e`; `pnpm build:android` compiles and runs the helper's Java suite.

Live on a Pixel 7 emulator, Contacts editor with `sim={adjust=pan}` (from `dumpsys window windows`), LatinIME bound — and re-run identically on the `WindowEdges` build:
- helper reports the IME window `[0,1517][1080,2400]` while the application window keeps its full `1080x2400`;
- `scroll down --json` -> `referenceHeight: 1505`, `keyboardAvoided: true`, `keyboardMinY: 1517`, swipe `1205 -> 301`;
- before this change the same command swiped `1920 -> 480`, starting under the keys.

`UiAutomation.getWindows()` answers with an empty list until the service asks for interactive windows, so the viewport read applies the seam the tree capture already uses instead of trusting that a snapshot capture ran earlier in the same instrumentation. Without it the read sees no windows at all, fails open, and looks identical to "no keyboard on screen".

Regression proofs: neutering the clip fails 2 scroll tests and restores green; restoring largest-window selection fails `GestureViewportReaderTest` with `Expected [0,1400][1080,2400], got [0,1517][1080,2400]`.

## Harness note for the next agent

`open` activates agent-device's headless `imehelper`, which draws no keyboard, and the emulator defaults to no soft keyboard while a hardware keyboard is attached. Bind a real IME **after** `open`:

```
adb -s <serial> shell settings put secure show_ime_with_hard_keyboard 1
adb -s <serial> shell ime set com.google.android.inputmethod.latin/com.android.inputmethod.latin.LatinIME
```

Restore `com.callstack.agentdevice.imehelper/.TestInputMethodService` afterwards, or typing determinism in later sessions is silently different.

## Size

13 files, 612 gross lines, one command family. The Android-shaped rows it satisfies (`adjustPan window with only 88px of 900px left above the IME is refused`) are already in the shared table landed by #2537.


thymikee:
The IME window selection can still leave a scroll inside the keyboard at f63e72f. GestureViewportReader.java:54-61 keeps only the largest input-method rectangle, even though the composer and key plane may be separate windows. A larger lower key plane discards a taller composer’s earlier top edge, so the computed swipe band can overlap that composer. Preserve the earliest top edge of the IME windows intersecting the swipe path and add a two-window regression. Current-head checks pass and the reported Pixel 7 adjustPan run covers the single-window case. This layer also depends on #2503.

thymikee:
Fixed at `67c6ba6bbb`. Largest-window selection is gone: the read copies every `TYPE_INPUT_METHOD` window and unions the ones that cross the swipe's centre line and overlap the application window, so the earliest top edge survives — the same centre line the shared clip rule tests, which keeps the two sides asking the same question.

The pre-filter has a second effect worth naming: an input method window beside that line, e.g. a floating candidate strip at the right edge, no longer shortens the band even though its top edge is higher, because a vertical swipe can never reach it.

The rule moved onto plain window edges (`GestureViewportReader.WindowEdges`) rather than `Rect`, because `android.graphics.Rect` is a stub in `android.jar` off-device — `new Rect(...)` throws `RuntimeException: Stub!`, which is why the existing helper tests avoid it. That is what makes the two-window case a unit test instead of a simulator-only path:

- `GestureViewportReaderTest` (registered in `SnapshotHelperTestSuite`, which `scripts/build-android-helper.sh` compiles and runs as part of `pnpm build:android`) covers composer `[0,1400][1080,1517]` over key plane `[0,1517][1080,2400]` in both orderings → `[0,1400][1080,2400]`, the side strip ignored, and no-IME null.
- Proved it bites: restoring largest-window selection fails with `Expected [0,1400][1080,2400] for composer above its key plane, got [0,1517][1080,2400]` — the exact band your comment describes.

Two limits I want to be explicit about:

- `pnpm check` green at `67c6ba6bbb`, but I could not re-run the device case: `open` refuses because `Pixel 7 CI` is held by a live foreign claim (`session=cwd:bf47053517e86bd1:default`, workspace `t3code-45b6bb98`), and I did not take it over. The `adjustPan` reading in the description came from the pre-refactor build; `pnpm build:android` recompiles the read and runs the suite, and the application-window path is unchanged by this commit.
- A composer reported as its own window is not something this emulator produced — LatinIME gave one window at `[0,1517][1080,2400]` — so multi-window coverage here is the unit test, not a device trace. If you have a device/IME pair that splits them, I'd rather add that reading than trust the union.

thymikee:
Device claim cleared, so the read path is re-verified end to end on the `WindowEdges` build (`67c6ba6bbb`), same Pixel 7 / Contacts `adjustPan` / LatinIME setup:

```
"referenceHeight": 1505, "keyboardAvoided": true, "keyboardMinY": 1517, "pixels": 903, swipe 1205 -> 301
```

Identical to the reading in the description, so the refactor and the multi-window union change nothing on a single-window IME — which is the only shape this emulator can produce. IME restored to `com.callstack.agentdevice.imehelper/.TestInputMethodService` and `show_ime_with_hard_keyboard` removed afterwards.

thymikee:
The multi-window repair is clean at 67c6ba6: the helper unions relevant IME windows instead of discarding the composer, with ordering and side-window regressions. The reported current-head Pixel 7 run verifies the single-window path, and checks pass with no conflicts. This layer still depends on resolving #2503. Please update the description’s old largest-window wording and file/line counts to match the union implementation.

thymikee:
The rebase onto #2503 preserves the previously reviewed Android changes byte for byte. No new code finding on 5d55192f, and current checks pass without conflicts. This layer remains dependent on #2503 closing its live iOS refusal-evidence gap before the stack is ready.

thymikee:
Rebased on the updated #2537 and pushed `cef825eb77`: the refusal names its three numbers explicitly instead of spreading the clip variant (so `kind` never leaks into error details), and the duplicate `isMeasurableRect` is gone. The helper reader no longer validates the IME frame itself; the shared clip already fails open on a frame it cannot measure, and the Java side already drops empty bounds. Tests unchanged, `pnpm check` green.

thymikee:
No actionable findings in cef825eb. The shared frame guard preserves malformed-IME fail-open behavior, and the refusal now passes only its measured fields. The earlier Android changes and live evidence remain applicable; current checks pass. Stack readiness still depends on #2503 resolving its live iOS refusal-evidence decision.

PR #2515, opened 2026-09-12
fix(android): carry accessibility selected state into snapshots

### LINKED ISSUE #2462: Android: the `selected` state is dropped from snapshots, so the `selected` Maestro selector never matches
## Summary

On Android, `agent-device` does not surface the accessibility `selected` state. The
underlying platform exposes it — `uiautomator dump` reports `selected="true"` on the
very same node — but it appears in neither `get attrs` nor `snapshot --force-full`.

As a consequence, a Maestro selector qualified with `selected` can never match on
Android, and `assertVisible` on it fails with the generic "visible condition did not
match". #558 lists `selected` as supported (both as a selector and for `tapOn.selected`),
so this reads as a silent gap rather than an unimplemented feature.

Version: `agent-device` 0.21.0 (latest on npm at the time of writing).
Environment: local Android emulator, Android 16 / API 36, macOS 26.6.2, release build of
a React Native app (Expo Router bottom tabs). Reproduced on a freshly booted AVD.

## Reproduction

Any app with a bottom tab bar whose tab buttons carry a testID works. Below,
`home-tab` is the currently selected tab and `local-tab` is not.

```bash
agent-device open com.example.app --platform android --device "<name>" --foreground
agent-device get attrs 'id="home-tab"'
```

```json
{
  "index": 107,
  "type": "android.view.View",
  "label": "Home",
  "identifier": "home-tab",
  "enabled": true,
  "editable": false,
  "password": false,
  "hintShowing": false,
  "visibleToUser": true,
  "hittable": true,
  "depth": 15,
  "parentIndex": 106,
  "ref": "e108"
}
```

No `selected` field. The full tree does not contain the string either:

```bash
agent-device snapshot --force-full | grep -ic selected
# => 0
```

The platform does expose it, on the same screen, seconds apart:

```bash
adb shell uiautomator dump /sdcard/r.xml
adb shell cat /sdcard/r.xml | tr '<' '
' | grep 'home-tab'
# resource-id="home-tab" class="android.view.View" content-desc="Home" selected="true"

adb shell cat /sdcard/r.xml | tr '<' '
' | grep 'local-tab'
# resource-id="local-tab" selected="false"
```

So the information is present in the source `uiautomator` XML and lost on the way to the
agent-facing snapshot.

## Impact on `--maestro`

This flow step passes under Maestro on the same device and fails under
`agent-device test --maestro`:

```yaml
- assertVisible:
    id: 'home-tab'
    selected: true
```

```
failed at: Replay failed at step 22 (assertVisible "home-tab"):
Maestro visible condition did not match.
```

The unqualified assertion passes on the same screen, on the same element, in the same
run — which locates the failure squarely on the `selected` qualifier:

```yaml
# step 12 — passes
- assertVisible:
    id: 'home-tab'
# step 13 — fails
- assertVisible:
    id: 'home-tab'
    selected: true
```

```
failed at: Replay failed at step 13 (assertVisible "home-tab"):
Maestro visible condition did not match.
```

The diagnostic is misleading in a costly way: the message names the element, and the
element *is* visible. Time goes into the selector and into timing before the missing
attribute is suspected. A selector qualifier that cannot be evaluated would be better
rejected loudly — the way relationship selectors are, per #558 — than silently treated as
unmatched.

Note the asymmetry: the same flows assert the same thing successfully on iOS, so a
cross-platform suite migrated from Maestro fails only on Android.

## Impact

Any suite that asserts tab-bar or segmented-control selection. In our case 13 assertions
across 5 flows, all of which pass under Maestro and under `agent-device` on iOS. We have
had to restrict them to iOS to get the Android suite moving.

## What would help

1. Carry `selected` through the Android snapshot pipeline into `get attrs` and the
   structured tree, mirroring what `uiautomator` already reports. This is the fix that
   makes the documented `selected` support in #558 true on Android.
2. Failing that: reject a `selected`-qualified selector on Android with a distinct reason
   rather than reporting an unmatched condition, so the cause is visible from the failure
   message alone.

## Related

- #558 — Maestro flow compatibility tracker; lists `selected` as supported
- #2288 — expose editable-field metadata in Android structured snapshots (same class:
  platform data present, absent from the agent-facing snapshot)
- #2273 / #2274 — iOS selector queries: false negatives and no fallback producer

thymikee:
## Summary

Android never carried the accessibility `selected` state. The snapshot helper did not serialize it, and the host reads only the helper's XML, so no later layer could recover it: `get attrs` had no field, no snapshot node was marked, `is selected` could not match, and `assertVisible {id, selected: true}` failed with "Maestro visible condition did not match" on a visible element — while `selected: false` matched every Android node.

The helper now emits both answers, like `enabled` and `password`: an unselected control answers `false`, a helper older than the attribute answers nothing. Parser, Android hierarchy node, and published node carry it to `get attrs`, the `[selected]` marker, selectors, and Maestro qualifiers.

Smaller second fix: snapshot lines render `[selected]` on the default formatter path too. `--settle`/`diff` already compared selection, so a tab tap produced a changed pair whose two lines looked identical.

13 files. Deliberately not here: `checked`/`checkable`/`long-clickable` (still #1832 residue), Android `focused`'s omitted-false encoding, and Apple's true-or-absent `selected` — on Apple a `selected: false` assertion still passes for a selected control.

## Validation

Commit `602b731`. `pnpm check:affected --run` passed (3715 selected tests + command-docs); `pnpm test:unit` 9753 passed; lint and typecheck clean. Regression evidence: the 5 platform selection tests, the selector-read test, and the new diff test fail with the mapping or the line marker rendered inert.

Manual device evidence, Android 16 / Pixel 7 CI AVD (not an automated lane): on the Dialer tab bar, `get attrs @e31` reports `"selected": true`, and `press 'id=…tab_call_history' --settle` prints `- @e40 [group] "Keypad" [selected]` / `+ @e31 [group] "Home" [selected]`. A flow asserting `assertVisible {id, selected: true}` passes; with a helper APK built from `HEAD` the same flow fails with the issue's exact message and no `selected` field. Provider integration and coverage remain with GitHub CI.

Closes #2462


thymikee:
No code findings at 602b731. Android selected state now reaches attrs, selectors and Maestro while preserving false versus unknown, and selection-only snapshot diffs have visible markers. The reported Pixel 7 run covers the changed path, current-head checks pass and there are no conflicts; ready for human review and merge.

PR #2516, opened 2026-09-12
fix(cloud): cancel a screen read that its caller gave up on

### LINKED ISSUE #2509: snapshot -i hangs on AWS Device Farm Android when the screen plays looping video, then the whole session freezes
Hi, I am driving tests on real Android phones rented through AWS Device Farm using agent-device 0.20.10, and I hit a wall I hope you can help with.

What happens:
Our app opens fine and the first photo of the screen always arrives in about a second. But asking for the screen list (snapshot -i) on our onboarding screen never answers. The request runs into the 30 second limit, twice, then fails. After that, even photos stop working and the session starts reporting lease not active, so every later command fails too. One stuck request freezes the whole session.

What is on the screen:
A Flutter app onboarding page with a looping background video and one big Continue button. Small simple screens (like the splash) answer in about a second. So it looks like the list writer waits for the screen to go still, and a looping video means it never does.

What I already tried:
- snapshot -i with longer timeouts (up to 65s), same hang
- snapshot -i -d 8, same hang (I later saw depth trims after arrival, so this was never going to help a hang, my mistake)
- screenshot --no-stabilize, same freeze pattern
- settings animations off, which answers not supported on the Device Farm setup
- two phones (Pixel 8a and Pixel 8 Pro, Android 14), same result on both, across 7 separate sessions

What would help me:
- Is there any way to get helper style capture on a rented phone, or any flag that skips the wait for stillness?
- Is there an option like skipping unimportant items to make the list smaller before it is built?
- If not, a clear error naming the video problem would save a lot of rented minutes. I saw a newer release mentions a looping animation hint, does 0.21.x cover this case?

Happy to share logs. I have the Appium server log showing a fast small answer (1.3s) followed by the request that never returns, plus the phone log showing video decoding at that exact second. One failure reference: mtxi6le4-87a50325.

Thanks for the tool, and happy to test anything on my side.

thymikee:
## Summary

A hosted page-source read that outran its request was dropped by the client while the driver kept
walking the UI tree. Nothing at the wire said the answer was no longer wanted, so the read stayed in
flight and the session stayed busy serving a capture nobody was waiting for. On a screen that never
goes still — looping video, live ticker, continuous animation — that is what turns one stuck
`snapshot -i` into a frozen session.

The read is now bound to the request that asked for it, on Android and iOS, and a source read that
runs out of budget keeps its `webdriver_request_timeout` reason and names the cause.

```sh
agent-device snapshot -i --session df
# reason webdriver_request_timeout; the hint names a never-idle screen and says --timeout cannot widen this read
```

Closes #2509. Touched files: 10. The lease half of the investigation is stacked separately in #2517.

## Validation

- `pnpm check:affected --run` passed at `513df9839` (all runnable checks; provider integration and
  coverage are CI-owned jobs on this head).
- `npx vitest run packages/provider-webdriver/src test/integration/provider-scenarios/cloud-webdriver-runtime.test.ts`
  — green. Red without the fix: `the abandoned capture was still running after its client hung up`.
- Live on AWS Device Farm, Google Pixel 8a (Android 14): `connect aws-device-farm`, `snapshot -i`
  (66.4s cold, 25 nodes), `screenshot` (6.3s, 1080x2400), warm `snapshot -i` (0.37s). Session closed
  and disconnected.
- Review round: `--timeout` is gone from the hint and the docs. It widens the command around the read,
  not the read itself — which is why the report's 65-second attempt changed nothing. A real source-read
  budget is unimplemented and deserves its own issue.
- Docs placement: the never-idle rule now also sits on the Snapshots page, where shared hosted
  WebDriver behaviour belongs; the provider page keeps the metered-device part.
- Remaining risk: aborting at the wire cannot cancel the driver's own tree walk. The reporter's
  `lease not active` is a distinct mechanism and is not fixed here.


thymikee:
The cancellation wiring is sound at 1adbc57, but the new recovery advice cannot work: webdriver-interactor.ts:611-614 and the AWS docs suggest a larger --timeout for webdriver_request_timeout, while that flag only widens the daemon envelope. The source request still uses the transport policy timeout (30 seconds by default), with no timeout override from snapshot. Remove that advice or connect it to an actual source-read budget. Current checks pass and there are no conflicts; the reported AWS run covers ordinary reads, not cancellation and recovery on the provider.

thymikee:
Fixed in `c740e5ca6`. Confirmed your reading first: `WebDriverRequestOverrides.timeoutMs` is never
passed for `/source`, so every read runs on the transport's 30s policy with one retry no matter what
`--timeout` says — which is also why the report's 65-second attempt changed nothing.

The hint now names whose budget it is and drops the advice, the AWS docs say the same, and two added
assertions pin the hint text so the advice cannot come back. A source-read budget the caller can
actually set is still missing; that needs a surface decision of its own rather than riding in here.

thymikee:
The timeout advice is corrected on 2d7d7413. One documentation issue remains in website/docs/docs/snapshots.md:63: cancelling the client read does not establish that later provider commands stop waiting behind the abandoned capture. The Appium tree walk can continue, as the PR description acknowledges. Please bound that paragraph to stopping client-side waiting and acknowledge that provider work may continue. The aborting fixture proves local cleanup, not provider queue recovery. Current checks pass and there are no conflicts.

thymikee:
Bound in `513df9839`. The paragraph now claims only what the fixture proves: agent-device stops
waiting and stops holding the session, and the driver's own walk can keep running on the provider and
still occupy that session's queue.

Stack was also rebased onto `main` `973b74cc14` in the same pass, so both heads carry a fresh
`pnpm check:affected --run` on the new merge-base: green at `513df9839` here and `25c563d02` on #2517.

thymikee:
The cancellation wording is now accurate on 513df9839: client waiting stops, while the provider walk may continue and occupy its queue. No actionable findings remain. The runtime changes are preserved by the rebase and the reported live AWS evidence still applies; current CI is running, so merge readiness awaits those checks.

PR #2517, opened 2026-09-12
fix(daemon): let admitted work keep the lease it is working on

thymikee:
## Summary

A lease renewed only at admission, so a command slower than its own inactivity TTL expired the lease
that was paying for the device it was using. Expiry then tore the provider session down underneath a
client still waiting for that same command's result, and every later command reported a lease that was
no longer active.

Admitted work now preserves its lease the way a human-control hold does: while the request is still
wanted it defers expiry, and finishing while still wanted renews the lease for its existing TTL from
the moment the work ended. Work whose client hung up preserves nothing — it neither defers expiry past
that cancellation nor renews the lease when it lands. ADR 0007 records the rule.

Found while investigating #2509, and **not** its mechanism: a cloud WebDriver profile asks for a
ten-minute lease, so a one-minute hang cannot starve it. This reaches the daemon's one-minute default.
Depends on #2516 through shared fixtures.

## Validation

- `pnpm check:affected --run` passed at `25c563d02` (all runnable checks).
- `npx vitest run src/daemon/__tests__/{lease-registry,lease-in-flight-work,request-lease-work,request-execution-scope}.test.ts`
  — 67 passed. Red without the layer: `an admitted request that outlives the lease TTL keeps its lease
  and session`, `work nobody waited for renews nothing once its lease fell due`.
- Review round: a completed pass emptied its set but left the map key, and only the expiry sweep
  removed keys — which never reads a released lease again, so every connect-and-close leaked one.
  Releasing the last pass now removes the entry, and releasing a lease drops its claims and disarms the
  passes still running on them. All three new tests fail at the previous head and pass here.
- Live check that set this scope: on AWS Device Farm a 66-second command kept its lease on the pre-fix
  build, which is what ruled this layer out as the #2509 cause.
- Remaining risk: a handler that ignores cancellation while holding the session lock now keeps the
  lease open instead of expiring it mid-command.

- Second round: `25c563d02` is comment-only — it restates the invariant without borrowing #2509's cause, which a ten-minute cloud WebDriver lease rules out for that report.

### DIFF
--- a/docs/adr/0007-remote-device-leases.md
+++ b/docs/adr/0007-remote-device-leases.md
@@ -51,6 +51,22 @@ owning lease expiry.
 Backend-only leases remain valid for older remote clients, while provider-aware
 clients get device-level contention and clearer recovery.
 
+## Admitted request work
+
+A lease renews when a request is admitted and never again while that request runs, so an admitted
+command slower than the inactivity TTL used to expire the lease that was paying for its own device
+and tear the session down underneath the client still waiting for its result. Admitted work
+therefore preserves its lease the way a human-control hold does: while the request is still wanted
+it defers expiry, and finishing while still wanted renews the lease for its existing inactivity TTL
+from the moment the work ended. A request whose client hung up preserves nothing — it neither defers
+expiry past that cancellation nor renews the lease when it finally lands — so a handler that ignores
+its cancellation cannot hold a rented device open.
+
+Which leases this reaches depends on the inactivity TTL the client asked for: the daemon default is
+one minute, while a cloud WebDriver connection profile asks for ten. A single command that runs
+longer than its own lease is therefore ordinary on the default and only reachable through a profile
+on the longer one.
+
 ## Human control
 
 Human-control holds coexist with an open remote session. They belong to `LeaseRegistry` and use
--- a/src/daemon/__tests__/lease-in-flight-work.test.ts
+++ b/src/daemon/__tests__/lease-in-flight-work.test.ts
@@ -1,0 +1,55 @@
+import { test, expect } from 'vitest';
+import { LeaseInFlightWorkRegistry } from '../lease-in-flight-work.ts';
+
+test('a pass defers its lease until it is released', () => {
+  const work = new LeaseInFlightWorkRegistry();
+  const pass = work.retain('lease-a', () => true);
+
+  expect(work.isDeferred('lease-a')).toBe(true);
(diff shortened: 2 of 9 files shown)


thymikee:
The new in-flight registry leaks entries under lease churn at f6f7b05. lease-in-flight-work.ts:60 removes a completed pass from its set but leaves the empty map entry; explicit lease release then removes the lease before expiry can call isDeferred to clean that entry. Each completed, explicitly released lease leaves another permanent key in the daemon. Delete empty entries when releasing work and cover the normal lease-release path. Current checks pass and there are no conflicts; this layer also depends on #2516.

thymikee:
Fixed in `b0cdafc2d`. `releasePass` removes the lease's entry once its set empties, and
`LeaseRegistry.releaseLease` calls `inFlightWork.forget(leaseId)` — which also marks passes still
running as released, so work that outlives its own lease renews nothing.

Covered at both ends: releasing the last pass leaves no key recorded, `forget` disarms live passes,
and a registry test walks the ordinary `releaseLease` path. All three fail at `f6f7b05` and pass here.
The two expiry removals already cleaned, since both reach `isDeferred` through `isLeaseProtected`
before deleting.

thymikee:
The lease cleanup fix is clean on f00fe88e: releasing the last pass removes the empty map entry, and forgetting a lease disarms retained passes. The reported regression tests cover those cases. This layer still depends on the documentation correction in #2516. Current CI also contains cancelled Smoke Tests and Bundle Size runs, so it is not fully green.

thymikee:
The dependency is closed: #2516 carries the bounded paragraph at `513df9839`, and this branch is
rebased onto it. Both were rebased onto `main` `973b74cc14` (a config-scope merge landed) and re-run:
`pnpm check:affected --run` green at `513df9839` and `25c563d02`.

On the cancelled runs — those are this stack's own force-pushes cancelling their predecessors, not
failures. The current heads report no failures; Coverage, Integration Tests, Repo Guards, Typecheck &
Package and three of four Smoke legs were green on `f00fe88e9`, and CI is re-running on `25c563d02`.

thymikee:
No actionable findings on 25c563d02. This is a byte-identical rebase of the reviewed lease changes, and #2516 has corrected its documentation. The reported focused regressions still apply. Current CI is running; the cancelled predecessor runs do not establish a failure on this head. Merge readiness awaits current checks and the prerequisite.

PR #2532, opened 2026-09-12
fix(daemon): keep a discarded rejection from shutting the daemon down

### LINKED ISSUE #2519: fix(daemon): a discarded promise rejection must not shut the daemon down
## Defect

A rejected promise that nobody handles reaches the daemon's only `unhandledRejection` handler (`src/daemon/server/daemon-runtime.ts:644`), which calls `shutdown({ exitCode: 1 })`. Every open session dies.

Four sites, verified at `eefe37b51e`:

1. `src/platform-runtime-screen-recording-apple-simulator-host.ts:131` and `:146` — `void started.then(rollback)` has no rejection handler. `localTransport.start` awaits two dynamic imports and then calls `signal?.throwIfAborted()`; an abort during the imports rejects `started`.
2. `src/platform-runtime-screen-recording-apple-simulator-host.ts:165` — inside `rollbackAcquiredSimulatorProcess`, `delay(UNPUBLISHED_STOP_GRACE_MS, grace.signal).then(() => false)` is a `Promise.race` member. When `process.wait` wins, `.finally(() => grace.abort())` rejects the losing `delay` (it rejects on abort, `:280-294`), and nothing handles it. Same bug class, no `void` prefix, so a `void x.then(` grep misses it.
3. `src/daemon/managed-device-allocation/lease-admission.ts:211` — `void ….then(` without a rejection handler.
4. `packages/provider-limrun/src/request-cancellation.ts:63` — deferred-only, benign today; fix for uniformity.

## Fix

Inline `.then(ok, fail)` (or `.catch`) at each site. For the race loser, attach a no-op rejection handler to the `delay` branch before racing.

Do not add a helper module, a new host-kit subpath, or a regex guard: the guard proposed in the draft plan (`void\s+…\.then\(`) does not catch site 2.

## Done when

- A unit test aborts a recording start during the transport's import phase and asserts the daemon process handler is never invoked.
- A unit test for `rollbackAcquiredSimulatorProcess` where `process.wait` resolves first asserts no unhandled rejection (`process.on('unhandledRejection')` spy).

thymikee:
## Summary

Closes #2519.

`acquireSimulatorProcess` abandoned an aborted acquisition with `void started.then(rollback)`.
`localTransport.start` awaits two dynamic imports and then rejects with the abort reason, so every
aborted `record start` delivered that rejection to the daemon's only `unhandledRejection` handler
(`src/daemon/server/daemon-runtime.ts:644`), which exits 1 and kills every open session. Both abandon
sites now end the rollback chain in a handler, which also covers a rollback that itself fails.

Swept all 24 `void … .then(` sites rather than adding the proposed regex guard. Three rows changed
shape:

- **Site 2 refuted.** The losing `delay` branch is a `Promise.race` member, and `Promise.race`
  attaches a rejection handler to every member, so the post-settlement abort rejection is already
  handled. Verified in Node and by a test that passes at `origin/main`. No no-op handler added.
- **Site 3 unreachable, and the proposed shape was unsafe.** `work` is `renew(...)`, which cannot
  reject: every path ends in a return or `stop()`. Resolving a rejection as "renewal settled" would
  make `run` re-arm `pending = renew(...)` with no backoff, so a broken `renew` becomes a renewal
  spin instead of a daemon exit. A rejection now routes through the closure's own `stop` as
  `teardown-required`/`authority-unconfirmed`, which is what `renew`'s catch already reports. No
  failing test exists because no reachable rejection does.
- **Site 4** is a resolve-only deferred. Deleted the discarded continuation at its construction path —
  `settled` now removes itself from `pending` — instead of adding a dead handler.

Adds `packages/provider-limrun/src/request-cancellation.test.ts`, which pins the drain's
abort-while-in-flight ownership contract that site 4 edits; those three tests pass before and after,
so they are contract coverage, not a regression test. 5 files.

Two review findings were rejected as unreachable rather than guarded: `child.kill('SIGINT')` inside
`rollbackAcquiredSimulatorProcess` cannot throw for a valid signal name (`ChildProcess#kill` returns
`false` for a dead child), so the SIGKILL escalation is not reachable through a thrown `kill`; and a
rejected `start()` cannot have spawned anything, because the local transport spawns synchronously
after its last `throwIfAborted`.

## Validation

Tested at `1326c85c71`. `pnpm check:affected --run` passed (14 checks: format, lint, typecheck,
layering, fallow; 1,684 unit tests). Both new abort tests fail at
`origin/main`, capturing `cancel simulator transport start` as an `unhandledRejection`, and pass at
HEAD; the grace-race assertions pass both ways, which is the site-2 refutation. The unhandled-
rejection observer now flushes with two `setImmediate` hops instead of a 10 ms timer, so it is immune
to fake timers. The transport is injected and no device-facing path changed, so no live run applies.
Coverage and provider integration stay CI-owned.

### DIFF
--- a/packages/provider-limrun/src/request-cancellation.test.ts
+++ b/packages/provider-limrun/src/request-cancellation.test.ts
@@ -1,0 +1,41 @@
+import { expect, test } from 'vitest';
+import { createLimrunRequestOperationDrain } from './request-cancellation.ts';
+
+test('releases a settled operation from the drain', async () => {
+  const drain = createLimrunRequestOperationDrain();
+  await expect(drain.wait(Promise.resolve('provider line'))).resolves.toBe('provider line');
+  await expect(drain[Symbol.asyncDispose]()).resolves.toBeUndefined();
+});
+
+test('propagates a provider rejection and still releases the drain', async () => {
+  const drain = createLimrunRequestOperationDrain();
+  await expect(drain.wait(Promise.reject(new Error('socket closed')))).rejects.toThrow(
+    'socket closed',
+  );
+  await expect(drain[Symbol.asyncDispose]()).resolves.toBeUndefined();
+});
+
+test('keeps an aborted operation owned until its provider source settles', async () => {
+  const drain = createLimrunRequestOperationDrain();
+  let settleSource!: (value: string) => void;
+  const source = new Promise<string>((resolve) => {
+    settleSource = resolve;
+  });
+  const controller = new AbortController();
+  const waiting = drain.wait(source, controller.signal, 'limrun operation aborted');
+  const reason = new Error('release the instance');
+
+  controller.abort(reason);
+  await expect(waiting).rejects.toBe(reason);
+
+  let released = false;
+  const disposing = drain[Symbol.asyncDispose]().then(() => {
+    released = true;
+  });
+  await new Promise((resolve) => setTimeout(resolve, 5));
+  expect(released).toBe(false);
+
+  settleSource('late provider line');
+  await expect(disposing).resolves.toBeUndefined();
+  expect(released).toBe(true);
+});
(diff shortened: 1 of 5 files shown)


thymikee:
No actionable findings on 1326c85c. The renewal rejection path stops admission before finishing the wait, preserving normalized failure handling without a retry loop. The focused race regressions cover the changed behavior. Current checks pass and there are no conflicts. Ready for human review and merge.

Issue #2281, opened 2026-09-04
Remote iOS recording: recover completed export after record stop times out


brentvatne:
## Problem

A customer running agent-device against an EAS-hosted remote iOS simulator reported that `record stop` timed out client-side after 90 seconds while the daemon continued exporting. The recording eventually became available, but they could not find a CLI command to retrieve it and fetched it manually through the daemon's `/artifacts` endpoint.

This leaves an agent unable to finish retrieving its recording through the normal CLI workflow even though the export succeeds.

## Environment and evidence

- Remote iOS simulator on EAS Simulator.
- Version baseline for this report: 0.20.10; exact client/daemon versions and recording options still need confirmation from the reporter.
- Customer-reported observation, not independently reproduced by us. The exact timeout transcript, session ID, export duration, and a minimal reproduction are not yet available. The customer has offered transcripts and diagnostics.
- We have not established which layer owns the 90-second timeout.

## Requested behavior

- Distinguish a failed export from an export that is still running when the client stops waiting.
- Provide a supported CLI recovery path to inspect export status and retrieve the completed daemon-managed recording after the original request times out, without starting another recording or manually calling HTTP endpoints.
- Give actionable recovery information in the timeout response, including a stable recording/export identifier if needed.
- Preserve normal session/tenant ownership checks on retrieval.

The exact command shape is open. The existing documented `artifacts` command lists provider-hosted artifacts (for example BrowserStack/AWS Device Farm); this request concerns a recording exported by the remote agent-device daemon.

## Related work

- #197 covered normal remote artifact retrieval; this is specifically recovery after the original export request times out.
- #2170 concerns a host recording lock after an ungraceful recorder kill. Here the daemon reportedly kept exporting and produced an artifact, so this is not known to be the same mechanism.

thymikee:
This is a valid recovery gap, but we need the customer evidence before locking the owner and API. Please attach the exact client and daemon versions, full `record start` / `record stop` options, the timeout response including diagnostic ID and log path, session ID and timestamps, the daemon request log around export completion, and the resulting artifact metadata/path. Please also confirm whether the client and daemon were on the same version. With that, we can distinguish timeout/progress handling from a durable export-operation/recovery contract.

PR #2534, opened 2026-09-12
fix(record): replay the finished export from a retried record stop

thymikee:
## Summary

`record stop` against a remote daemon can run out of its 90 s client window while the daemon is still exporting. The manifest then reaches `completed`, but a repeated `record stop` read it as `no active recording`, and the stored completion carried no caller-side output path — the recording existed on the daemon host with no CLI way to collect it (#2281).

A repeated `record stop` now serves that export from the durable manifest: session and device guards still apply, the file must still exist, and the caller gets the ordinary stop response, including the caller-side paths that make it downloadable. The manifest stores the finished response as the one object `record stop` returned, under a single key, so a replay cannot lose a field, and recovery refuses a response whose served or caller-side paths are not whole. The replay records no second session stop action, and a remote timeout now names that retry.

```sh
agent-device record stop --session demo --out ./capture.mp4  # client window expired
agent-device record stop --session demo --out ./capture.mp4  # returns the completed recording
```

16 files touched. No new command, flag, or timeout change; the 90 s envelope is untouched. Addresses #2281 — its `needs-info` request for customer evidence is still open.

## Validation

`pnpm check:affected --run` passed at `0cb272ae553a9b48c2189a8b8ce2dda8d63e9bcd` (format, lint, typecheck, layering, fallow, build, related vitest, daemon wire compat, command docs), and all PR checks including Coverage pass on that head.

Regression proof: removing the completed-manifest replay fails `record stop returns the export whose response never reached the caller` and its downloadable-path sibling in `src/daemon/handlers/__tests__/record-runtime-stop-recovery.test.ts`.

Two gates shaped the code: session teardown's eager-closure budget keeps the completion writers inside the resource definition, and the test-size ratchet moved hint assertions to `src/daemon-client/__tests__/daemon-client-timeout.test.ts`.

Live evidence for the readiness case is in a comment on this PR, captured at `8139aeb1ea`; the reuse commits after it declare no behavior change. On an iPhone 16 Pro simulator behind `agent-device proxy`, `record stop` timed out at 90 s, the manifest turned `completed` 97 s later, and the retried `record stop` returned in under a second and downloaded a recording whose SHA-256 matches the host file, with one export in the host log.


thymikee:
No actionable code findings on 8139aeb1. The simplified completion envelope validates stored metadata and the matching session/device before recovering an existing artifact. Current checks pass, including coverage, and there are no conflicts. Readiness still needs the live remote-export case: an export lasting beyond 90 seconds, a completed manifest after client timeout, and a retry that downloads the completed recording without starting another export.

thymikee:
Live remote-export evidence on `8139aeb1ea6ca6e5b60e8b113113e1abf4908d8a` — iPhone 16 Pro simulator (iOS 26.2), daemon behind `agent-device proxy` so the client drove the remote HTTP path.

1. `record start` (backend `simctl recordVideo`), then `record stop`: the client failed at its 90 s envelope exactly as reported — `COMMAND_FAILED: Daemon request timed out` with `details.timeoutMs: 90000` and the new hint naming the retry and session.
2. The daemon kept exporting. 94 s after the client gave up the session manifest still read `lifecycle: open, phase: completing`; it turned `completed` 97 s after the timeout, storing the caller-side path — `"clientOutPath": "/private/tmp/ad-2281-out/capture.mp4"`. That caller directory was still empty.
3. Re-ran `record stop`: returned in under a second with `recording: stopped`, `durationMs: 135888`, and artifacts carrying `localPath` for the video and its telemetry. `capture.mp4` then existed caller-side, 58,703 bytes, SHA-256 `906ec2d6…de016563` matching the host file `/tmp/agent-device-recording-1789280208874-hd9vfq.mp4`.
4. No second export: the host daemon log holds exactly one export pair (`… start` / `done`) for the whole run, and the retry resolved from the completed manifest without binding a runtime.

Method note: to make the export deterministically exceed the 90 s window I ran this verification build with a temporary 120 s delay in the recording finalizer host adapter, gated on an environment variable. That patch is not part of this PR — `src/platform-runtime-screen-recording-host.ts` is unchanged at this head. Daemon, proxy, client, manifest, and artifact download are all this branch.

Cleanup: session closed, proxy and host daemon stopped with `--clean`, scratch state dirs removed, verification patch reverted and rebuilt.

thymikee:
No actionable findings on 811e4a35. The reported live remote-HTTP run at 8139aeb proves timeout, completed manifest, retry and SHA-matched download without a second export; the temporary finalizer delay tests recovery, not ordinary export latency. The later refactors preserve that behavior. Required live evidence is satisfied; merge readiness follows the current checks.

thymikee:
No actionable findings on 0cb272ae. The shared timeout-context helper preserves the same fields for socket and HTTP requests, so the earlier live timeout/retry/download proof still applies. Current checks pass and there are no conflicts; ready for human review and merge.

PR #2535, opened 2026-09-12
chore(config): scope the qs override and drop dead fallow, worktree and prettier entries

### LINKED ISSUE #2521: chore(config): scope dependency overrides and delete dead fallow, worktree and prettier entries
## Drift

All verified at `eefe37b51e`.

- `pnpm-workspace.yaml:19` pins `qs: 6.15.2`, the exact version behind GHSA-x5fp-wj9c-mxmx and GHSA-4mjr-xmp4-gh2g, unscoped by major. `pnpm audit --prod` is clean; `qs` reaches the tree only via `@stryker-mutator/core → typed-rest-client`. Rewrite to `qs@6: ^6.16.0`, which clears both. Also `undici@7: '^7.29.0'` (`:16`) is a caret override while `package.json:328` pins `7.29.0` exactly and `tsdown.config.ts` bundles undici (`onlyBundle`). Renovate's `lockFileMaintenance` can move the bundled build under the caret without the visible pin changing. Align the override to the exact pin.
- `.fallowrc.json:22` (`scripts/runner-request-count/run.ts`, deleted in `26ac865c63`) and `:34` (`test/scripts/metro-prepare-packaged-smoke.mjs`, removed in `8889a17cf1`) are dead entry roots. Fallow drops them silently. Delete them; `entry` already accepts globs (`:35`), so prefer a glob where a directory of runners exists.
- `.worktreeinclude:13-14` lists `android/multitouch-helper/` paths that no longer exist.
- `.prettierignore` duplicates `.oxfmtrc.json:8-15`; nothing references prettier. Delete it.

## Fix

Config edits only. Do not add an `existsSync` validator script; the dead-root class is closed by globs plus review.

## Done when

`pnpm audit` reports zero advisories, `pnpm check:affected --run` is green, and no `entry` path in `.fallowrc.json` is absent on disk.

thymikee:
## Summary

Closes #2521. Config and lockfile only.

`qs: 6.15.2` overrode every major and pinned the exact version behind GHSA-x5fp-wj9c-mxmx and
GHSA-4mjr-xmp4-gh2g. `qs@6: ^6.16.0` keeps the scoped major and clears both. The caret is deliberate
here and not for undici: `qs` reaches the tree only through `@stryker-mutator/core →
typed-rest-client` and is never bundled, so a caret keeps patching itself; `undici` is bundled into
the published build by `tsdown.config.ts:109`, so its override is now the exact `7.29.0` that
`package.json:328` and `packages/provision-kit/package.json:52` already declare.

The caret had already done damage: `pnpm-lock.yaml` recorded `specifier: ^7.29.0` for **both** undici
importers, so the lockfile disagreed with their own manifests. The new lockfile records `7.29.0`,
matching them.

Deleted: two `.fallowrc.json` entry roots pointing at files removed in `26ac865c63` and `8889a17cf1`
(Fallow drops absent roots silently), two `.worktreeinclude` paths under a nonexistent
`android/multitouch-helper/`, and `.prettierignore` — its six patterns are identical to
`.oxfmtrc.json`'s `ignorePatterns` and nothing in the repo reads prettier.

The issue suggested a glob "where a directory of runners exists"; neither root sits in one.
`scripts/runner-request-count/` is gone, and `test/scripts/` holds one runner that
`test/ci/trusted-fixture-artifact.test.mjs` imports statically, so it is already in the graph. All 29
remaining `entry` paths exist with brace and glob expansion applied.

## Validation

`pnpm audit` — dev tree included — reports **no known vulnerabilities** (was 2 moderate);
`pnpm audit --prod` was already clean. `pnpm check:affected --run` at `5f48e47163` ran the full
59-check plan (a lockfile change fails open) and passed every check except `mutation-model`, which
fails identically on a clean `origin/main` tree in this worktree: its claim that
`src/commands/interaction/runtime/gestures.test.ts` reaches the `scroll-edge-state` kernel went stale
when the scroll path moved to `src/daemon/scroll-runtime.ts`. Pre-existing, advisory in CI, unrelated
to this diff.

### DIFF
--- a/.fallowrc.json
+++ b/.fallowrc.json
@@ -19,7 +19,6 @@
     "src/daemon.ts",
     "packages/capture-kit/src/png-worker.ts",
     "scripts/patch-xcuitest-runner-icon.ts",
-    "scripts/runner-request-count/run.ts",
     "packages/capture-kit/src/ios-snapshot-engine/replay.ts",
     // #1596 regression fixture: runs as a real `node --experimental-strip-types`
     // subprocess (test/integration/daemon-replace-exit-flush.test.ts), so
@@ -31,7 +30,6 @@
     "examples/sdk/contracts-result.ts",
     "examples/sdk/batch-orchestration.ts",
     "examples/sdk/ai-sdk-tools.ts",
-    "test/scripts/metro-prepare-packaged-smoke.mjs",
     "test/integration/*.test.ts",
     "website/docs/404.mdx",
     "website/rspress.config.ts"
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,6 +1,0 @@
-.claude/**
-dist/**
-node_modules/**
-**/*.md
-scripts/maestro-conformance/corpus/**
-fallow-baselines/**
--- a/.worktreeinclude
+++ b/.worktreeinclude
@@ -10,8 +10,6 @@ examples/test-app/.expo/
 DerivedData/
 android/snapshot-helper/build/
 android/snapshot-helper/dist/
-android/multitouch-helper/build/
-android/multitouch-helper/dist/
 
 # Tool caches
 .fallow/
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -5,11 +5,11 @@ settings:
   excludeLinksFromLockfile: false
 
 overrides:
-  undici@7: ^7.29.0
+  undici@7: 7.29.0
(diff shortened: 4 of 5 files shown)


thymikee:
No code findings at 5f48e47. The scoped qs override and exact undici pin match the lockfile, and the removed entries have no active consumers. Current-head checks pass and there are no conflicts; ready for human review and merge.

PR #2537, opened 2026-09-12
fix(contracts): state the scroll keyboard clip once for every platform

thymikee:
## Summary

First layer of #2500: the rule that keeps a scroll's swipe out of the keyboard, stated once in TypeScript, with the golden table both native runners read. iOS implements it in #2503 and Android in #2514, so nothing here moves a gesture yet.

`clipScrollViewportAboveKeyboard` is the whole decision: a keyboard that reaches the swipe's centre line takes the band below it, and when what remains cannot carry a swipe the command refuses with `scrollKeyboardOccludesSurfaceError` rather than scrolling a surface the keys own. `contracts/fixtures/scroll-keyboard-policy.json` is the cross-language table, so the Swift and Java lanes assert against the same rows instead of each platform inventing its own thresholds.

## Behaviour stated here

- A keyboard is struck when it reaches the swipe's centre line, and a floating keyboard sitting past that line is not — its width fraction no longer decides anything.
- `keyboardAvoided` and `keyboardMinY` join a scroll response only when the clip ran, and the CLI help, MCP output schema and command docs name them here, so both platforms answer with one shape.
- The keyboard is never dismissed. Dismissing drops focus and breaks a `fill`/`scroll`/`fill` loop; `keyboard dismiss` stays an explicit command.
- The refusal code lands before any runner emits it, which is deliberate: the typed reason and its evidence keys are defined once here, and #2503 and #2514 are what produce them.

## Validation

`pnpm check` green at `1b8548af6d`, which covers the golden parity table (18 rows, including the Android-shaped `adjustPan window with only 88px of 900px left above the IME is refused` and the iOS-shaped `landscape keyboard leaving 28pt of a 402pt axis is refused instead of swiping under the keys`), the Apple-side mapping tests, MCP metadata and the production-exports lane.

No device run belongs to this layer — it dispatches nothing. The live simulator and emulator evidence for the paths it states is posted on #2503 and #2514.

## Size

11 files, 615 gross lines: one shared module, its golden table, and the surface that names its two response keys.

Part of the #2500 stack: this PR -> #2503 (iOS runner) -> #2514 (Android). Do not merge out of order.


thymikee:
No actionable findings on 1b8548af. The shared contract and Apple surface extraction preserves the previously reviewed source, with the dispatch changes correctly left in the dependent PRs. Typed errors, response fields, MCP schema, fixtures, help, and docs are aligned for this layer. Current checks pass and there are no conflicts. Ready for human review and merge.

thymikee:
Reviewed the whole stack and pushed a simplification at `673ea455d6` (layers 2 and 3 rebased on top, content of #2503 unchanged).

**What changed here**
- `scrollKeyboardOccludesSurfaceError` takes a plain measured occlusion (`keyboardMinY`, `visibleHeight`, `viewportHeight`), which only Android produces in-process. The unmeasured variant, the `Extract<…> & {…}` type and the test that pinned the fabricated message are gone.
- The reason and hint are one frozen `SCROLL_KEYBOARD_OCCLUDES_SURFACE_DETAILS` in `scroll-gesture`. The Apple owner now adds it to the runner's own error and keeps the runner's measured message and transport details, instead of rebuilding the error with a message that had no number in it.
- The `scroll` help text is one short clause; the error hint carries the recovery when it happens.

Net −41 lines on this layer, same reason and hint on every path.

**Tried and reverted:** hosting the details in `classifyRunnerReportedError` so the lifecycle-journal recovery route gets them for free. Any value import from contracts into `runner-contract.ts` (or `scroll-gesture → scroll-command`) grows the eager closures and fails `eager-closure-budgets`, so the per-command wrapper stays. The journal route is still covered because the wrapper matches on `details.runnerErrorCode`, which recovery sets.

`pnpm check` green on the stack head `cef825eb77`.

thymikee:
No actionable findings on 673ea455. The Apple wrapper preserves the runner message and transport details while adding the shared reason and hint, including after journal recovery.

Coverage failed in the unchanged snapshot-source lifecycle fixture: "a crashed helper is removed and the next request starts a fresh helper" exceeded its dispatch deadline and raised bridge-connect-deadline. This looks unrelated to the scroll delta, but CI needs to pass before merge. The ready-for-human label remains valid. Failure: https://github.com/callstack/agent-device/actions/runs/34748827127/job/103701514870

Issue #2540, opened 2026-09-13
refactor(daemon): stop importing the commands barrel through src/runtime.ts


thymikee:
## Purpose

Five daemon files import `src/runtime.ts`, which imports the command barrel with a value import:

```ts
// src/runtime.ts
import { bindCommands, type BoundAgentDeviceCommands } from './commands/index.ts';
```

The value closure of `src/runtime.ts` is 122 files, 48 of them under `src/commands/`. The daemon's total value closure is 717 files and carries **121 `src/commands/**` files** — roughly the entire commands zone (140 files, 21k LOC) — plus 17 root flat files. The daemon loads a command surface it does not dispatch through, and every change in commands is potentially a daemon-restart-cost event.

Umbrella: #2545. Chosen first because it is the only child that changes what the daemon actually loads, and because #2469 (eager-closure budget tolerance) is unblocked by cutting closure, not by adding budget.

## Required behavior

1. `src/runtime.ts` keeps its public shape: `AgentDevice`, `createAgentDevice`, `createMemorySessionStore`, `localCommandPolicy`, `restrictedCommandPolicy`, and the re-exported command types. No consumer-side API change is part of this issue.
2. Daemon consumers of `src/runtime.ts` take the specific declarations they use through their owning module or contract, not through the runtime facade. Determine the used symbol set per importer before choosing the seam; do not invent an interface wider than the symbols in use.
3. The barrel `src/commands/index.ts` either stops being the value entry for runtime binding, or the binding moves to the composition root that needs it. One construction path, no second registry.

## Completion conditions

- Value-only transitive closure from `src/daemon/**` contains **zero** files under `src/commands/`. Measured with a probe over `scripts/layering/model.ts` as described in the parent, and reported in the PR.
- `src/runtime.ts` has no value edge into `src/commands/**`, or is no longer reachable from `src/daemon/**` at all.
- No new eager work: the eager-import-closure budgets for the daemon entry surfaces do not rise.
- R4 stays 0, R5 stays 0, R6 does not grow past the ratchet.

## Out of scope

- Repackaging `src/commands` as a workspace package. That decision needs the `cli-schema` orientation settled first (sibling child).
- Any change to the command descriptor registry or the `command-registry` package.

## Dependencies

- None. Wave A, independent branch.

thymikee:
## Closure measurement: the runtime.ts cut alone does not reach the completion condition

[...]

Issue #2545, opened 2026-09-13
refactor: cut the daemon's commands-closure leak and its last root hubs (successor to #2278 and #2335)


thymikee:
## Purpose

Remove the last structural coupling that makes the daemon expensive to change and impossible to package, and stop it paying for the whole command pipeline it never calls. Successor to #2278 (residual semantic coupling, closed) and #2335 (decoupling wave 3, closed). Those waves moved code into packages and classified what remained; this one cuts the two things measurement shows are still load-bearing.

Measured at `9aa6465768` with the gate's own model (`pnpm depgraph` plus a throwaway probe against `scripts/layering/model.ts`): 1,684 production files, 8,689 edges, 38 zones, R4 = 0, R5 = 0, R6 = 4.

Two findings drive the sequence:

1. **The daemon's eager value closure contains 121 `src/commands/**` files** (717 files total in the closure). One value edge is responsible for 48 of them: `src/runtime.ts` → `./commands/index.ts`. Five daemon files import `src/runtime.ts`.
2. **Daemon → `src/` outside the daemon is 51 file pairs** (42 root flat files, 7 `src/core/**`, 2 `src/cli-schema/**`), down from 154 edges into 67 files at the wave-3 baseline. Twelve of those pairs are one file, `src/provider-device-runtime.ts`, and ten of its twelve importers need exactly one symbol, `isActiveProviderDevice`.

Fan-in is not cost. Churn over the last 400 commits: `src/backend.ts` 53, `src/provider-device-runtime.ts` 19, `src/runtime.ts` 6, `src/runtime-contract.ts` 5, `src/io.ts` 4. `runtime-contract.ts` has 27 importers and is nearly frozen; it is not a target.

## Children

Wave A (independent):

- #2540 — `refactor(daemon): stop importing the commands barrel through src/runtime.ts`. The closure leak; the only child that changes what the daemon loads.
- #2541 — `refactor(runtime): own provider-device admission behind a typed capability`. 12 pairs → 0.
- #2542 — `chore(gates): classify root hubs, dynamic edges, and the root-to-daemon entry`. Makes the residue in this issue visible and non-regrowing.

Wave B (after Wave A):

- #2543 — `refactor(cli-schema): orient the schema layer below commands`. 5 value back-edges, 7 type-only.
- #2544 — `refactor(daemon): extract the maestro adapter and durable-capture admission subtrees`. Blocked on where `daemon-request` lives.

[...]

Issue #2543, opened 2026-09-13
refactor(cli-schema): orient the schema layer below commands


thymikee:
## Purpose

`src/commands` (140 files, 21k LOC) and `src/cli-schema` (9 files, 2k LOC) point at each other. Forward: `commands → cli-schema` 24 file pairs. Back: `cli-schema → commands` 12 pairs, and the kind split says the direction is nearly settled already:

- **value (5):** `cli-help.ts → commands/cli-command-aliases.ts`, `cli-help.ts → commands/command-text.ts`, `command-overrides.ts → commands/family/registry.ts`, `command-overrides.ts → commands/cli-grammar/flag-groups.ts`, `command-schema.ts → commands/command-metadata.ts`, `command-schema.ts → commands/cli-grammar/flag-registry.ts`, `command-schema.ts → commands/cli-grammar/flag-groups.ts`
- **type-only (7):** `cli-config.ts`, `command-overrides.ts`, `command-schema.ts`, `types.ts` reaching `cli-grammar/flag-types.ts`, `command-metadata.ts`, `command-text.ts`

`command-schema.ts` is also one of the 51 daemon → root pairs, imported by 2 daemon files, and it is the daemon's **second** path into the commands zone: `command-schema.ts → command-metadata.ts → family/registry.ts → every command family`. Measured at `e6f288012b`: cutting only the `src/runtime.ts → commands/index.ts` edge (#2540) leaves 75 `src/commands/**` files in the daemon value closure; also cutting the daemon's two `command-schema.ts` imports takes it to 0. This issue is therefore on the umbrella's critical path, not only a layering nicety.

Nothing can package `src/commands` — the largest packaging decision left in the repo — until the schema layer's position is stated.

Umbrella: #2545.

## What the back edges actually carry

Two kinds, and they go opposite ways:

1. **Vocabulary commands owns but nothing in commands' runtime needs:** `cli-grammar/flag-types.ts`, `flag-definitions-*.ts`, `flag-registry.ts`, `flag-groups.ts`, `command-text.ts`, `cli-command-aliases.ts`. Their closures are package-only (`@agent-device/contracts/command`, `@agent-device/command-registry`). These sit **below** both layers.
[...]

Issue #2547, opened 2026-09-13
mutation-model: indirect-ownership self-test pins a test that stopped reaching scroll-edge-state


thymikee:
## Purpose

`mutation-model` — the mutation harness self-test inside `pnpm check:affected` — fails on a clean `origin/main` checkout, so any contributor whose diff selects that check sees a red local gate for a reason unrelated to their work. It hit #2533 and #2521 during wave 1 of #2531.

## Reproduce

```sh
pnpm install --frozen-lockfile
pnpm mutation:test
# or, as check:affected selects it:
pnpm check:affected --run   # → check:affected: mutation-model failed.
```

```text
ℹ pass 39
ℹ fail 1
✖ a kernel is owned by tests that reach it indirectly
  AssertionError [ERR_ASSERTION]: The expression evaluated to a falsy value:
  at scripts/mutation/ownership.test.ts:40:10
```

The other 39 assertions in that file pass, including `ownership is complete: every test reaching a kernel owns it`.

## Root cause

`scripts/mutation/ownership.test.ts:40-44` pins `src/commands/interaction/runtime/gestures.test.ts` as an *indirect* owner of the `scroll-edge-state` kernel. That kernel is `packages/capture-kit/src/snapshot/scroll-edge-state.ts` plus `packages/capture-kit/src/snapshot/scroll-edge-state/selection.ts`.

`gestures.test.ts` no longer imports anything on the scroll path: its resolved import graph is 257 modules and contains neither kernel source, so `ownersOf` returns `kernel-errors`, `interaction-settle`, `selectors`, `target-annotation-serde`, `snapshot-occlusion` and no `scroll-edge-state`. Scroll moved out of the interaction command family — `src/daemon/scroll-runtime.ts` was added by `bcca714a07` ("refactor: move gesture family to platform runtime", #1952) and the snapshot zone relocation landed in `1f9d940bff` (#2385) — and the ownership assertion was never moved with it.

The kernel itself is not uncovered: it has 82 owning test files, including the mirrored `packages/capture-kit/src/snapshot/__tests__/scroll-edge-state-*.test.ts` set, and the `ownership is complete: every test reaching a kernel owns it` test passes. Only this hand-pinned pair is stale.

## Required behavior

- The indirect-reachability assertion keeps its purpose — proving the deriver credits a test that exercises a kernel without importing it by name — while naming a pair that is true today.
[...]

PR #2548, opened 2026-09-13
refactor(daemon): bind only the command families the daemon dispatches through

thymikee:
## Summary

The daemon built its in-process `AgentDevice` through `src/runtime.ts`, which value-imports the all-family command barrel. Five daemon files therefore evaluated the management, recording, observability and system families they never call.

Splits the runtime assembly (backend, artifacts, sessions, policy, cancellation) into `src/runtime-factory.ts`, which carries no command surface, and adds `src/runtime-command-surface.ts` for the three families an in-process executor actually dispatches through: `capture`, `selectors`, `interactions`. `createAgentDevice` keeps its exact exported shape — including the `CommandPolicy` and `CommandSessionRecord` type re-exports — and now composes both, so there is one construction path and `bindCaptureCommands` is the single capture-family binding definition reused by `bindCommands`.

Part of #2540, **not** closing it: see the completion-contract question below.

**Completion-contract mismatch.** #2540 asks for zero `src/commands/**` files in the daemon value closure. Cutting `runtime.ts → commands/index.ts` and replacing nothing leaves 75; once the scoped binder the daemon still needs is added back it is **108**, which is what this PR measures (closure 717 → 705 files, commands 121 → 108, 19,659 → 17,434 LOC). The floor is **34**: the dispatch surface itself, plus the two `cli-schema/command-schema.ts` edges owned by #2543. Zero is only reachable if the daemon stops executing commands in-process. Posted on #2540 and #2545 for decision.

## Validation

Commit `75e5142a89`, rebased on `origin/main` `3394d5b8`. `pnpm check:affected --run`: unit/fallow/layering/tooling pass; `mutation-model` fails on `scripts/mutation/ownership.test.ts` "a kernel is owned by tests that reach it indirectly" — reproduced identically on unmodified `origin/main`, inherited, not caused here. `pnpm test:unit` 1,275 files / 9,815 tests (pre-rebase commit). `check:layering` OK: R76 12 edges, R6 4, R4/R5 zero. No device-facing change.


thymikee:
Two corrections remain on 3ee002cb. src/runtime.ts:5 drops the CommandPolicy and CommandSessionRecord type exports, despite #2540 requiring the existing exported surface to stay intact. Restore those type exports; no current in-repo consumer break was found.

The scoped runtime binding otherwise preserves the reviewed behavior, but this PR reports 108 command files still in the daemon closure and a future floor of 34, while #2540 requires zero. Resolve that completion-contract mismatch with the maintainer and #2543 before claiming to close #2540; the measured reduction is useful but does not meet the current issue condition.

thymikee:
Both corrections addressed at `75e5142a89` (rebased on `origin/main` `3394d5b8`):

1. `CommandPolicy` and `CommandSessionRecord` are re-exported from `src/runtime.ts` again, so the module's exported surface is unchanged. Fallow's unused-export finding is carried as an explicit `ignoreExports` entry naming the file and the two symbols, with the reason (#2540 keeps the surface; no in-repo consumer), rather than a bulk baseline regeneration.
2. The close claim is withdrawn: the commit says "Part of #2540", the PR body states the measured floor, and the completion-contract question is posted on #2540 and #2545 for your decision.

One inherited failure to flag: `mutation-model` fails on `scripts/mutation/ownership.test.ts` ("a kernel is owned by tests that reach it indirectly"). Reproduced identically on unmodified `origin/main`, so it is not caused by this branch.

thymikee:
Both findings are resolved on 75e5142a: the two type exports are restored, the Fallow exception is limited to those symbols, and the PR now describes a partial runtime split without closing #2540. No actionable delta findings remain. The closure-target decision stays on the issues; current CI is still running, and the reported mutation-model failure is tracked separately in #2547. Ready for human review; merge readiness awaits CI.

Issue #2555, opened 2026-09-13
Resolve device interactors behind a composed capability


thymikee:
## Problem

`src/daemon/snapshot-interactor-capture.ts` reaches interactor mechanics through a dynamic
`import('../core/interactors.ts')` and calls `getInteractor(device)`. That is the only production
edge leaving the daemon zone through a dynamic import, so the ranked-spine rules (R4 cycles, R5
back-edges, R6 type inversions) never see its direction, and R76's target predicate never named it
either: the file it lands on is not in the `src/platform-runtime*` family, even though that module
reaches the platform runtime through `src/provider-device-runtime.ts`.

`src/core/interactors.ts` also reads `isActiveProviderDevice` and `getProviderDeviceInteractor` from
`src/provider-device-runtime.ts`, so it is the reason `#2541` cannot reach zero root edges from
`src/` alone: the daemon and `src/core` share one ambient interactor lookup.

## Desired behaviour

Interactor resolution arrives as a composed capability the daemon is handed, the way the platform
gateway, host diagnostics, and resource cleanup already are (ADR 0019 section 1/2). The daemon calls
the capability with a device; root composition names `src/core/interactors.ts`. The daemon zone then
holds no interactor lookup and no dynamic import of a module that reaches platform mechanics, and
`src/core/interactors.ts` keeps a root edge only from the composition module that owns it.

## Acceptance

- The dynamic `src/daemon/snapshot-interactor-capture.ts -> src/core/interactors.ts` edge is gone.
- R76's inventory records no `leaked-platform-mechanics` edge whose target is reached dynamically.
- The interactor capability is bound at the same composition site that installs the provider-device
  admission, so one change names both.

Issue #2541, opened 2026-09-13
refactor(runtime): own provider-device admission behind a typed capability


thymikee:
## Purpose

`src/provider-device-runtime.ts` (303 LOC) is the largest remaining daemon → root hub: 12 file pairs (11 daemon files plus `src/core/interactors.ts`).

The symbol sets say the port is already narrow. Ten of the twelve importers take **only** `isActiveProviderDevice`:

- `android-foreground-surface.ts`, `android-system-dialog.ts`, `device-ready.ts`, `direct-ios-selector.ts`, `handlers/session-doctor.ts`, `interaction/internal/interaction-gesture.ts`, `interaction/internal/interaction-touch-reference-frame.ts`, `request-generic-dispatch.ts`, `session-device-resolution.ts`, `snapshot-session.ts`
- `server/daemon-runtime.ts` takes `createProviderDeviceRuntimeRequestProviders`
- `src/core/interactors.ts` takes `getProviderDeviceInteractor` and `isActiveProviderDevice`

This is therefore a small typed capability, not an interface with a method per importer. Do not design an eleven-method surface; the data refuses it.

Umbrella: #2545.

## Required behavior

1. The daemon consumes provider-device admission through a typed contract it can name, composed at the process root by one module that is the sole importer of provider/platform mechanics — the shape already accepted for lifecycle participation in #2333 (`src/daemon/platform-owner-lifecycle.ts` + `src/platform-runtime-daemon-lifecycle.ts`).
2. `isActiveProviderDevice` keeps its exact truth table. Characterize it against unmodified code first: it is read on 10 call sites spanning Android foreground detection, alert handling, iOS selector fast paths, doctor, device refresh, and gesture reference frames, and each reads a boolean to decide a fast-path or a refusal.
3. `src/core/interactors.ts` consumes the same contract rather than the root module, so `src/core` loses its root dependency on `provider-device-runtime.ts`.
4. `createProviderDeviceRuntimeRequestProviders` stays root-composed. If the request-provider composition genuinely belongs to the daemon runtime, classify that edge in the R76 inventory rather than deepening it.

## Completion conditions

- No daemon file imports `src/provider-device-runtime.ts`.
- `src/core/interactors.ts` does not import it either, or the remaining edge is recorded in `DAEMON_PLATFORM_RUNTIME_EDGES` with a rationale — after the sibling gates child makes root hubs classifiable.
- Key behavior is on typed reasons, not on error text.
[...]

PR #2556, opened 2026-09-13
refactor(runtime): own provider-device admission behind a typed capability

thymikee:
## What

Gives the daemon its own typed capability for the one provider-runtime fact it decides on, so the daemon zone stops importing `src/provider-device-runtime.ts` from ten sites (#2541).

`src/daemon/provider-device-admission.ts` declares `ProviderDeviceAdmission` — one method, `isActive(device)` — defaults to the no-provider state every un-composed process already sees, and is installed by root composition at the site that already composes the provider request providers (`src/daemon/server/daemon-runtime.ts`). The predicate keeps its name, its per-call read, and the request-scoped ALS behaviour underneath it; the ten leaf call sites change only their import specifier.

## Why a capability and not a re-export

A re-export would keep the same edge under a different filename. The seam is the point: the daemon names the fact it needs, root names where it comes from, and R76 now has exactly one daemon edge to that hub to classify (see the stacked gate PR).

## Measured

| | before | after |
| --- | --- | --- |
| daemon production sites importing `src/provider-device-runtime.ts` | 11 (10 leaves + the runtime composition) | 1 (the composition only) |
| `src/**` production edges into that hub | 12 pairs | 2 pairs |

The two remaining pairs are `src/daemon/server/daemon-runtime.ts -> src/provider-device-runtime.ts` (root composition, now classified in the R76 inventory) and `src/core/interactors.ts -> src/provider-device-runtime.ts`. `src/core/interactors.ts` keeps its edge: it also needs `getProviderDeviceInteractor` and sits below the daemon, so it cannot consume the daemon's seam. It is reached from the daemon only through the dynamic interactor lookup, which is now visible and classified with its own deepening issue (#2555).

## Test seams

Nine daemon test files and the provider-scenario integration harness drive provider ownership through the root ambient scope. They now compose the admission the way `daemon-runtime.ts` does — one `installProviderDeviceAdmission({ isActive: isActiveProviderDevice })` — instead of relying on the daemon reading root internals. Two `vi.mock` call sites retarget to the daemon module.

## Validation

- `pnpm test:unit` — 1,282 files, 9,872 tests passed
- `pnpm check:layering` — OK (R76 inventory with the widened predicate on the stacked tip)
- `npx vitest run test/integration/provider-scenarios` — 66 files, 209 tests passed
- `pnpm check:affected --run` — all slices pass except `mutation-model`, whose `scripts/mutation/ownership.test.ts` "a kernel is owned by tests that reach it indirectly" failure reproduces identically on unmodified `origin/main` (3394d5b89c)

Part of #2541 (the `src/core/interactors.ts` edge and the interactor seam remain).


thymikee:
No actionable findings on 32266c47. The composed delegate preserves per-call request-scoped ownership, and the router-level provider test covers the affected behavior. The remaining interactor seam is explicitly deferred to #2555. Current checks pass and there are no conflicts; ready for human review, with #2548 first in the merge order.

Issue #2542, opened 2026-09-13
chore(gates): classify root hubs, dynamic edges, and the root-to-daemon entry


thymikee:
## Purpose

Three blind spots let daemon/root coupling reappear without failing a gate. Each is a measured hole, not a hypothetical.

1. **Root hubs are not classifiable.** R76 (`scripts/layering/daemon-platform-runtime-inventory.ts`) matches `src/daemon/**` importing `src/platform-runtime(?:\.ts|-[a-z0-9-]+\.ts)`. A daemon file importing a differently-named root hub that itself reaches platform mechanics is invisible: `src/provider-device-runtime.ts` accounts for 12 daemon file pairs and imports `src/platform-runtime.ts` and `src/platform-runtime-screen-recording-apple-runner-transport.ts` on its own behalf, neither of which is under any inventory. ADR 0022's "one leaked edge remains" is true only for direct edges to gated names.
2. **Dynamic imports are unwatched.** R4/R5/R6 see static edges. Today's one instance: `src/daemon/snapshot-interactor-capture.ts` → `src/core/interactors.ts` (dynamic; `src/core/interactors.ts` imports `src/provider-device-runtime.ts`). `docs/dependency-graph-findings.md` already records dynamic direction as "outside every rule".
3. **Nothing owns `root → daemon`.** 37 file pairs reach into daemon internals from outside. `src/daemon-client/**` — which wave 3's #2360 moved out of `src/daemon/` — still holds 13 pairs / 8 value edges into `daemon-request.ts`, `config.ts`, `daemon-process.ts`, `session-repair-tombstone.ts`, `request-progress-protocol.ts`. The move changed the client's location, not its imports.

Umbrella: #2545.

## Required behavior

1. Extend the R76 inventory so a daemon edge to a **root module that reaches platform mechanics** is classified even when the module name is not `platform-runtime-*`. Keep the existing three categories and the same no-drift/no-stale rules; widen the target predicate rather than adding a second table.
2. Make dynamic import direction observable in the same pass: a dynamic daemon edge into a root or lower zone must be classifiable or reported. Do not add a separate gate job.
[...]

PR #2557, opened 2026-09-13
test(layering): classify daemon edges that reach platform mechanics through a hub

thymikee:
## What

Makes R76 classify daemon edges by what they reach instead of by filename (#2542, first half).

The inventory keyed on the target filename `src/platform-runtime*.ts`. A daemon import of a root module that imports that family itself — `src/provider-device-runtime.ts` is the real case — carried the same platform mechanics and was invisible to the gate. Dynamic edges were invisible too, and the ranked spine cannot see a dynamic import's direction at all (R4 cycles, R5 back-edges, R6 inversions all skip it).

`computePlatformMechanicsHubs` now derives the target set from the tree: the platform-runtime family plus every module **outside** the daemon zone that reaches it, over static and dynamic edges alike. Intra-daemon hops stay out, so the rule does not flag every file in the zone (the measured alternative was 298 pairs; the real answer is 3).

## Newly visible, and classified rather than allowlisted

| edge | classification |
| --- | --- |
| `daemon/server/daemon-runtime.ts -> src/provider-device-runtimes.ts` | composition-essential |
| `daemon/server/daemon-runtime.ts -> src/provider-device-runtime.ts` | composition-essential |
| `daemon/snapshot-interactor-capture.ts -> src/core/interactors.ts` (dynamic) | leaked-platform-mechanics, deepened by #2555 |

The dynamic interactor lookup is the interesting one: a filename pattern never would have asked for a rationale there, and it is the only production edge that leaves the daemon zone dynamically.

## Verification

Planted-violation cases in `scripts/layering/daemon-platform-runtime-inventory.test.ts` (24 tests): a new hub import is red, a symbol added to a classified hub edge is red, a binding added to the dynamic lookup is red, an intra-daemon hop is not a hub, and the accepted paths hold on the real shapes. `pnpm check:layering` passes on this tip with the three entries; a recorded edge that disappears is still stale, so the inventory cannot drift in either direction.

## Not in this PR

The rest of #2542 is the entry half, and it is a different shape of work: declaring the root→daemon entry surface and repairing the eight `daemon-client -> src/daemon/**` value edges (`config.ts`, `daemon-process.ts`, `session-repair-tombstone.ts`, `request-progress-protocol.ts`) rather than classifying them. The cut sets are small for three of those four (`config.ts` has 42 inbound files and needs a subset decision), so it belongs in its own change. I will file it as a child of #2542 rather than grow this PR across both halves.

Part of #2542


thymikee:
No actionable findings on 696a9a93. Hub derivation includes static and dynamic edges while retaining stale-entry and symbol-drift enforcement; the daemon-entry work is explicitly left to #2559. Current checks pass and there are no conflicts; ready for human review after its prerequisites.

PR #2561, opened 2026-09-13
refactor(command-registry): move CLI flag grammar, text and aliases down (#2543)

thymikee:
## What

Pure relocation of the CLI flag grammar, text vocabulary and command aliases **down** into `packages/command-registry`, the layer below both the CLI parser and `src/commands`. No compat re-export at the old paths — every consumer switches to the owning `@agent-device/command-registry/*` subpath.

Moved (content unchanged, new export subpaths added):
- `flag-types`, `flag-registry`, `flag-groups`, `flag-definitions-{action,connection,target,workflow}`
- `command-text`
- `cli-command-aliases`

Colocated `flag-registry.test.ts` moves with the source; the cross-layer `flag-declaration-admission.test.ts` stays at `src/commands/cli-grammar/` (it observes the daemon + `cli-schema` consumers) and now imports the package — its `resetModules` + dynamic-import divergence contract survives the move (verified).

## Why this is the safe half of #2543

The per-command defaults move and the daemon's two `cli-schema/command-schema.ts` imports are the harder cut (the `apps` default is load-bearing — `assertResolvedAppsFilter` throws on `undefined` — and the constant sits behind the heavy `contracts/device` façade, which blows the flag-grammar entry's eager budget). Those belong with the daemon-closure work, not this move. This PR only re-locates declarations so #2543-B can point the daemon at the package.

## Gates

- `pnpm check:layering` — OK. R2/R5 clean, **R6 = 4 (unchanged)**, **R76 = 15 (unchanged)**, R11 now 295 exported subpaths (+5).
- `pnpm typecheck` — clean.
- `pnpm vitest run scripts/__tests__/eager-closure-budgets.test.ts` — 600/600; no entry closure grows (renames tracked).
- `pnpm check:production-exports` — **parity with merge-base** (68), no new unused export.
- `pnpm check:mcp-metadata`, `pnpm check:command-docs` — pass.
- `oxlint` + `oxfmt` — clean.
- Affected `unit-core` suites (command-registry, cli-schema, cli, cli-grammar, command-input/explain/flags, mcp, capture, interaction, metro, replay, family, session-action-recorder): ~1000 tests pass.

Part of #2545 (Wave B) / implements the "flag grammar goes down" step of #2543.

PR #2562, opened 2026-09-13
refactor(daemon): cut the last edges into the CLI schema layer (#2543)

thymikee:
## Summary
Stacked on #2561. Cuts the daemon's last two `src/cli-schema/` edges and, per review, makes command defaults a **single source**.

- `registry.ts` now owns a `COMMAND_DEFAULTS` table typed on `CommandFlags` plus `applyCommandDefaults`. The CLI parser and the daemon request scope both import it from `@agent-device/command-registry/registry` — the one module already in both callers' eager closure. The table adds a literal (registry already declares every other default as a literal), not a new import, so the registry entry and `cli.ts` closure are unchanged.
- Removed the apps facet default, `CommandSchema.defaults`, and the earlier `command-defaults` module + parity guard. `DEFAULT_APPS_FILTER` collapses into `resolveAppsFilter`'s provider fallback and stops being a public export.
- `session-action-recorder` reads `recordedFlagKeys` from `flag-registry`; `command-schema`'s now-unconsumed re-export is removed.
- R10 now rejects any daemon import of `src/cli-schema/` (value or type).

## Validation
Head `9ff38c2351` (stacked on #2561). Green: `check:layering` (234/234, incl. the new R10 boundary), `typecheck`, `eager-closure-budgets` (600/600), `check:production-exports` (parity 68), `check:mcp-metadata`, `check:command-docs`, `lint`, `format`, and the affected unit suites (command-registry, cli-schema, cli parser, daemon request/management, MCP parity).

`check:affected`'s only failure is `scripts/mutation/ownership.test.ts` (`gestures.test.ts` → `scroll-edge-state` reachability), which fails **identically on `origin/main`** — pre-existing and unrelated to this change.


thymikee:
Current defaulting behavior is preserved on e0e093cd, but the ownership requirement in #2543 is not: src/commands/management/app.ts:109 still declares the apps-filter default, and packages/command-registry/src/command-defaults.ts:9 declares it again. The parity test detects drift but leaves two declarations to maintain.

Please make one defaults table typed on CommandFlags serve both callers and remove the facet copy. The CLI already loads the registry and the facet that imports DEFAULT_APPS_FILTER, so placing the table/application function in an already-loaded registry module is a concrete alternative to adding a new module to its eager closure. Verify the closure budget and CLI/daemon default parity after that change. The new daemon boundary check is otherwise sound; no current runtime divergence was found. CI was still finishing its last smoke check at review time.

thymikee:
Addressed in `9ff38c2351`.

One `COMMAND_DEFAULTS` table typed on `CommandFlags` now lives in `registry.ts` and serves both callers; the facet copy is gone. Chose `registry.ts` over a new module per your hint: the registry is already in both the CLI parser's and the daemon's eager closure, and it already declares every other command default as a literal (the `defaultValue` descriptors), so the table adds a literal rather than a new import — the registry entry stays 72 modules and `cli.ts` stays 295 (verified: `eager-closure-budgets` 600/600, and a probe importing the `contracts/device` facade into `registry.ts` grew 8 entries, so the literal is required, not just convenient).

`DEFAULT_APPS_FILTER` was only ever consumed by the facet + the fallback, so it collapses into `resolveAppsFilter` as a private const and drops out of the `contracts/device` facade rather than lingering as an unused export. Parity guard deleted — with one source there is nothing to keep in lockstep. CLI/daemon default parity now rests on both calling the same function (covered by the parser test, the daemon request-scope test, and a new registry unit test).

thymikee:
The duplicate defaults are resolved in 9ff38c2: CLI and daemon now use the same registry-owned table. No remaining findings in this follow-up; checks are green and this is ready for human review.

PR #2564, opened 2026-09-13
fix(android): retire recording evidence stranded by a re-adopted device id

### LINKED ISSUE #2550: Android record start refuses with "native recovery evidence already exists" on an emulator re-adopted from a previous session
## What happened

agent-device 0.21.1, Android emulators managed by an external tool that stops, parks and later re-adopts them under a new serial. On the first `record start` of a new session against a re-adopted emulator:

```
Android screenrecord native recovery evidence already exists
```

(internal error, "retry with --debug"). `record stop` then says "no active recording"; closing and reopening the session does not clear it. Cause: a stale `/sdcard/agent-device-recording-active.json` left by an earlier session on the same AVD, whose `deviceId` (the old serial) no longer matches, so the recovery check throws instead of retiring the manifest. Deleting the file by hand (`adb shell rm /sdcard/agent-device-recording-active.json`) fixes it.

Seen on three separate runs in two days, whenever an emulator changed serial between sessions. The failure is also easy to miss: with default output the last line is `Session state: ...` and the error is only in the request ndjson, so the run continues unrecorded.

## Proposal

When the manifest's `deviceId` does not match the current serial, or its session is no longer registered, retire the manifest and start the recording; and print the refusal on stdout so a caller sees it.

thymikee:
## Summary

Android `record start` refused forever once an emulator was re-adopted under a new serial. The
device-side marker names the identity that wrote it, and reconciliation retired evidence only when
that identity matched, so a marker from an earlier session on the same AVD blocked every later
recording with `native recovery evidence already exists` — surfaced as `UNKNOWN` with a hint asking
for a bug report — while `record stop` owned nothing that could clear it.

Reconciliation now retires evidence that is terminal or names an unreachable identity, after proving
every artifact it names is released: committed recorders through process inspect, an uncommitted
pending artifact against the recorders running on the device. Nothing is deleted under a live
recorder: the refusal is `DEVICE_IN_USE` with `details.writer`. Unreadable evidence, the other
transport mode, and an open recording this identity owns stay refused, now as typed errors naming the
marker path and the command that clears it. Reconciliation moved to `start-reconciliation.ts`, marker
persistence to `manifest-store.ts`. 18 files, 911 lines. Closes #2550.

Residual risk: retiring evidence under an unreachable identity removes artifacts nobody ever pulled,
since no reachable session can recover them; marker removal is not compare-and-delete.

## Validation

Heads `44682608bc` → `0bf8b96c6f`, each passing `pnpm check:affected --run` (format, lint, typecheck,
layering, fallow, build, unit, provider-integration). The follow-up commits answer review findings: a
writer probe that could not read a candidate read as "that path is free", then a scan that named one
recorder while hiding the candidates it could not read. Each regression was confirmed to fail when its
guard is reverted to doubt-as-proof.

Live on `emulator-5554`: a terminal marker naming `emulator-9999` was retired and the recording
started; an open marker naming `emulator-5554` refused with `native_recovery_evidence_open`,
non-retriable, marker retained; a marker naming a path an unmanaged `screenrecord` was writing refused
with `native_recording_artifact_claimed`, retained, and the retry after that recorder ended succeeded.
Sessions closed and device markers cleaned. Coverage stays CI-authoritative.


thymikee:
The pending-artifact cleanup at start-reconciliation.ts:108 can delete a recording that is still being written. findRunning drops uncertain process inspections, so a failed /proc read for a live recorder can produce an empty list; reconciliation then removes its artifact and recovery marker. Preserve that uncertainty and refuse cleanup when ownership cannot be established, with a regression covering an inconclusive inspection. The reported device scenarios and green checks do not cover this case. Reviewed 4468260.

thymikee:
Fixed in 200ad13429. `findRunning` filtered candidates down to proven recorders, so an empty list was read as "path is free" while a live recorder's `/proc` reads were failing; an unreadable process table threw a bare `Error`.

The transport probe now answers `clear | found | uncertain` (`probeRunningWriters`), and only `clear` proves an artifact free: start retirement refuses with `native_recording_recorder_unproven` and owned cleanup returns the evidence pending, so the marker and artifact stay put until the answer is conclusive.

Regressions: `src/platform-runtime-screen-recording-android-host.test.ts` covers an unreadable candidate and an unreadable process table; `start-reconciliation.test.ts` refuses with zero side effects; `recovery-cleanup.test.ts` retains evidence. All three fail when the `uncertain` branch is reverted to doubt-as-proof.

thymikee:
The all-uncertain case is fixed at 200ad13, but a mixed scan still loses uncertainty. In src/platform-runtime-screen-recording-android-host.ts:87, any identified writer returns `found` before unreadable candidates are considered. Owned cleanup then stops only those identified writers and removes the pending artifact and marker, even if an unreadable recorder is still writing that path. Preserve uncertainty in mixed results and retain the evidence; cover a readable writer plus an unreadable candidate through cleanup, asserting neither artifact nor marker is removed. Start reconciliation already refuses `found`. Current checks are green, but this cleanup path remains blocked.

thymikee:
Fixed in 8cf4552d08. The writer search no longer short-circuits on the first identified recorder: it returns the writers it proved together with `conclusive` — whether every candidate process was readable.

Owned cleanup now requires both halves before it acts: an inconclusive scan retains the artifact and marker and signals nothing, so an identified recorder can no longer be stopped while an unreadable one still holds the same path. Start reconciliation keeps naming the writer when one was found (`native_recording_artifact_claimed`) and refuses with `native_recording_recorder_unproven` when none could be proven.

Coverage: `keeps a mixed scan inconclusive when one candidate cannot be read` (transport), `retains evidence when only some pending artifact writers can be identified` (cleanup — asserts no signal, no artifact removal, marker retained). Both fail when the guard is reverted to `found`-wins.

thymikee:
The mixed-scan cleanup issue is fixed at 8cf4552: incomplete scans retain the artifact and marker without signalling any recorder, and the regressions cover both the transport result and cleanup. No remaining correctness findings; current checks are green and this is ready for human review. Before merge, reduce or split the aggregate diff: 915 additions plus 224 deletions is 1,139 lines, above the repository’s 1,000-line PR budget.

thymikee:
Size ask done in 0bf8b96c6f: aggregate diff is now 911 lines (739 additions, 172 deletions) across 18 files, inside the 1,000 budget.

The reduction came from not spending the budget on restyling. The reconciliation scenarios that already existed keep their original transport setup, and the stale table row that expected evidence naming another device identity to be kept was dropped, since retiring it is the point of this PR. New scenarios share one `evidenceRig` that holds the marker and the ordered side-effect log, the owned-cleanup uncertainty cases run from one table over both incomplete scans, and the transport's inconclusive writer-scan cases run from one table over the same scoped ADB fake.

Coverage is unchanged in substance: 21 reconciliation scenarios, 12 transport cases, 9 recovery-cleanup cases.

Issue #2549, opened 2026-09-13
record stop fails under host load because owner-identity ps probes have a fixed 1s timeout, leaving the recording in cleanup-pending


janicduplessis:
## What happened

agent-device 0.21.1 (also 0.20.10), macOS, iOS simulator. During an Xcode build on the same host (load average around 350), `agent-device record stop` failed with:

```
/bin/ps timed out after 1000ms
```

exit 1, empty output. The same `ps -ax -o pid=,ppid=,command=` takes 0.09 s from a shell when the host is idle. After the failure the recording resource stayed in `cleanup-pending`, `record start` refused with "recording already in progress" for the rest of the session, and the only way out was to SIGINT the `simctl recordVideo` process by hand and record with `simctl io recordVideo` directly.

## Where

`dist/src/owner-identity.js`: every `/bin/ps -p <pid> -o lstart=|command=|state=` probe runs with `timeoutMs: 1000` (the `m=1e3` constant). The daemon uses those probes to confirm the recorder's identity (pid + start time) before signalling it; when `ps` exceeds a second the stop refuses instead of signalling.

## Proposal

Scale the probe timeout with load, or retry once with a longer budget before refusing; and when identity cannot be confirmed within the budget, say so and leave the recording in a state `record stop` can retry, rather than `cleanup-pending`.

PR #2565, opened 2026-09-13
fix(record): end the recorder a loaded host could not confirm

thymikee:
## Summary

`record stop` refused to signal a recorder whose identity probe it could not afford, and the recording left in `cleanup-pending` then blocked `record start` until the daemon restarted (#2549).

- `readProcessIdentityFacts` (host-kit) reads start time, command and state concurrently, off the event loop, with a budget the caller names. The stop path names 5s; identity polling keeps the 1s default.
- Ownership evidence separates `unprovable` (host would not answer) from `not-ours` (host named a different process). Only the latter refuses. The recorder this owner spawned is ended through its own `child.kill` handle while the host is silent, and a marker the owner cannot account for blocks a `terminated` claim rather than being waited out.
- A process table that cannot be read degrades to "no extra descendants" with a warn diagnostic, since the markers persisted at start already cover the recorder.
- A termination or a finish the host refused is re-driven by the next `record stop` instead of replaying one rejection forever.
- Drops a dead outcome branch in `createAppleSimulatorProcess`.

10 files. No CLI, help, or docs change: the documented `record stop` behavior is what the code now delivers.

## Validation

- `pnpm check:affected --run` green at `c594622bb7` (5053 tests). One pre-existing apple-runner test (`runner-client.test.ts :: ensureXctestrunArtifact aborts only the disconnected request build…`) intermittently trips the slow-test gate under full-suite load; it passes standalone in 58 ms and is untouched here.
- Live iPhone 17 (iOS 26.2) simulator: `record start --scope device` then `record stop`, twice. Both exported an mp4, no leftover `recordVideo` process, and the second `record start` was accepted.
- Re-run of the same head: the iOS **Smoke Tests** job (`fixture-backed iOS simulator E2E`, scroll-to-visible on `automation-longpress`) passed on retry, so that failure is load-sensitive and not caused by this diff. Every check is green at `c594622bb7`.


thymikee:
No code findings at c594622. The reported simulator start/stop runs cover the normal path, but do not exercise the ownership-probe failure this fixes. Please record a live simulator run with delayed or unavailable ps identity probes, showing record stop exports the video, leaves no recorder process, and a subsequent record start succeeds. Checks are now green; that changed-path evidence is the remaining readiness gap.

thymikee:
Changed-path evidence, as requested in review. iPhone 17 simulator (iOS 26.2), head `c594622b`, isolated `--state-dir` under `/private/tmp` with a fresh daemon per run.

Fault injection was a temporary local patch to `packages/host-kit/src/internal/host-process.ts` only: a mode file made `readProcessIdentityFacts` throw (or answer 3 s late) and `listHostProcesses` throw, and every firing appended a line to a log, so the runs prove the fault was live rather than bypassed. The patch is reverted — `git status` clean, `git diff --stat origin/main...HEAD` back to 10 files / +489 −66, and `dist` carries no experiment code.

**Unavailable probes** — armed after `record start`, so only the stop saw it:

- fault log: 3 × `scope=identity` + 1 × `scope=table` fired during that one stop
- `record stop` → `"recording": "stopped"`; export 120 267 bytes, `isPlayableVideo` true (repo helper `packages/capture-kit/src/recording/video.ts`)
- `pgrep -fl recordVideo` → 0
- the next `record start`, with the fault still armed, was accepted; its stop exported a second playable mp4 (120 283 bytes)

**Delayed probes** — every identity probe answered 3 s late, past the 1 s budget this replaces: 4 delayed probes in one stop, `"recording": "stopped"`, playable 1.6 MB export, 0 leftover recorder processes.

So the ownership decision no longer decides the recording's fate: identity the host will not name still ends the spawned recorder through its own handle, an unreadable process table costs a descendant instead of the export, and `record start` works again without restarting the daemon. Cleanup: `daemon stop --state-dir … --clean`, then state dir, artifacts and the injected fault removed and the simulator shut down.

thymikee:
The reported live run with identity probes delayed by 3 seconds closes the requested changed-path evidence gap: stop exported a playable video, left no recorder process, and recording could start again without a daemon restart. No code change since c594622; checks are green and this is ready for human review.

PR #2566, opened 2026-09-13
fix(android): report the clip an Android recording really captured

### LINKED ISSUE #2551: Android screen recording stops capturing about one second after the last input, regardless of how long record stop waits
## What happened

agent-device 0.21.1, Android emulator, `record start <path> --quality high`, an action, then 6 to 11 s of waiting, then `record stop`. Across three runs and about ten clips, every clip ends roughly one second after the last input event: clips of 2.1 s, 2.7 s and 3.8 s for windows of 8 to 11 s. No error is reported. A result that appears on screen more than a second after the last tap (a toast, a list re-render after a network round trip) is not in the clip. iOS recordings in the same runs ran to `record stop`.

## Proposal

Keep capturing until `record stop`, or, if the capture is intentionally bounded by input activity, say so in the `record stop` output with the captured duration so the caller can re-record.

thymikee:
## Summary

Android `screenrecord` encodes a frame only when the screen changes, so a window ending on an unchanged screen returns a video far shorter than the requested duration while `record stop` reported only host wall-clock `durationMs`. `record stop` now also reports `capturedDurationMs`, measured from the pulled MP4 timelines, and warns with the clip length against the window the host bracketed around the recorder: `Date.now()` immediately before the recorder launches and immediately before the stop signal. A host that slept mid-recording reports a shorter window, which costs the warning rather than inventing one. iOS is unchanged.

```console
$ agent-device record stop --json
{ "durationMs": 15402, "capturedDurationMs": 6736,
  "warning": "…it covers 6.7s of the 13.0s recording window." }
```

Closes #2551. 30 files, inside the recording family plus its contracts, MCP schema, help and docs. Measuring the timeline needed an ISO-BMFF box walk, which replaces the private atom scan MP4 container detection used to do.

## Validation

Rebased onto `04052fdcd2`; conflicts were keep-both merges over #2564's `probeRunningWriters` op and its manifest-retirement help. Human review findings 1 and 2 are in `d4c0ad2adb`: the device clock is gone and the window module owns the clock arithmetic.

- Live on `emulator-5554` at `d4c0ad2adb`: swipe then 10s idle → `capturedDurationMs: 6736` against `ffprobe duration=6.736144`, warning above, recorder and session gone. Earlier heads measured `13102`/`13.101478` and `9733`/`9.732689`.
- `pnpm check:affected --run` at `d4c0ad2adb`: every gate passed except `mutation-model`, which fails identically at `04052fdcd2` (*a kernel is owned by tests that reach it indirectly*) and is not this PR's. Validation is therefore not fully green.
- **Coverage** and **Integration Tests** passed at `625dc00bab`; both rerun on this head.


thymikee:
No blocking findings at e590e85. The measured duration reaches the response and recovery metadata, and the short-clip warning reaches normal CLI output. The reported Android run matches ffprobe and current checks are green. The shared MP4 reader replaces the existing scan and reasonably accounts for the size increase; ready for human review.

thymikee:
Human review at 625dc00. The MP4 side is fine: the walker is a fair replacement for the old top-level scan since it has to descend into `moov`, and the sum-over-chunks policy and warning text are right. The weight of this PR is elsewhere, and two of the three parts are avoidable.

**1. The device clock is not needed; use host time.**
`elapsedUptimeMs` on the transport, the 1.5 s probe budget, `device-clock.ts` + test, and two extra adb round trips per recording (one on the start path) all exist to defend against host-vs-encoder clock drift. Quartz drift is tens of ppm; over a 30-minute chunked recording that is under 100 ms against a 2 s threshold. `Date.now()` taken at the same two points (before `startChunkAt`, before the stop signal) gives the same window with zero device I/O and no contract change. The one case where the clocks really diverge is a host that sleeps mid-recording, and there the host window comes out *shorter*, so the warning is missed rather than invented. That is the safe direction. This removes roughly a third of the production diff.

**2. Merge `device-clock.ts` into `captured-window.ts`.**
Whatever clock code survives is two tiny functions about the concern the window module already owns.

**3. The function-scoped `await import('./mp4-atoms.ts')` is a gate artifact.**
The old scan was inline in `video.ts` (closure = 1 module). Splitting it out made it 2, and the eager-closure rule for existing entries is no-growth with no approval path, so the gate punishes splitting a file and rewards a fake-lazy import of a 99-line dependency-free module. Keep the dynamic import for this PR so it lands, but I will file a follow-up to let an existing mechanics-surface entry grow within its category ceiling. Once that is in, `video.ts` imports the walker statically and `readMp4DurationMs` can live on the existing `recording-video` subpath instead of a new one.

**4. Not for this PR:** `capturedDurationMs` touched seven sites (two contracts, MCP schema, response builder, metadata encoder, and two hand-written validators in different packages checking the same shape). A single codec for `ScreenRecordingCompletion` in contracts, shared by the daemon store and the Android manifest, would make the next field a two-site change. Separate PR.

Requested changes: 1 and 2.

thymikee:
Addressed in `d4c0ad2adb`.

**1. Device clock removed.** `elapsedUptimeMs` is off the transport contract and out of the adb host, the 1.5s probe budget and both round trips went with it, and `AndroidScreenRecordingTransport` is back to the shape it had at the merge-base. The window is now `Date.now()` taken immediately before `startChunkAt` and immediately before the stop signal, which keeps the one property the read existed for — export latency is not attributed to an unchanged screen — at zero device I/O. A stop recovered through daemon recovery now passes the manifest's own `startedAt`, the first host timestamp the recording ever had, so a recovered stop gets the comparison it did not have before.

**2. `device-clock.ts` is gone.** What survived is one subtraction, and it lives in `captured-window.ts`, which already owned the comparison; `measureCapturedWindow` takes `startedAtMs`/`stoppedAtMs` and the callers take no part in the arithmetic.

Live at `d4c0ad2adb` (emulator-5554, one swipe then 10s idle): `capturedDurationMs: 6736` against `ffprobe duration=6.736144`, warning naming `6.7s of the 13.0s recording window`, and `durationMs: 15402` — the 2.4s between the host window and the reported duration is the export, exactly where it should be. `pnpm check:affected --run` at this head passes every gate except `mutation-model`, which fails identically at `04052fdcd2`.

**3.** Left the function-scoped `await import` as you asked. When the ceiling change lands, the two-site follow-up is `video.ts` importing the walker statically and `readMp4DurationMs` moving onto `recording-video`; the `recording-mp4-duration` and `recording-mp4-fixtures` subpaths added here would fold back at the same time.

**4.** Agreed, and the count is worse than the comment says for the next field: the daemon metadata encoder, both hand-written validators, the stop-recovery reader, the response builder, two contracts and the MCP schema. A `ScreenRecordingCompletion` codec in contracts shared by the session resource and the Android manifest is the fix; leaving it out here.

PR #2567, opened 2026-09-13
fix(cli): hand device-auth URLs to the browser without a shell

### LINKED ISSUE #2526: fix(cli): open the device-auth URL on Windows without cmd /c
## Defect

`src/cli/auth-session.ts:559` spawns `cmd /c start "" <url>` with `url = start.verificationUriComplete ?? start.verificationUri` (`:229`), taken verbatim from the device-auth response. `cmd /c` re-tokenizes its argument string even when passed as argv, so `&`, `^`, `|` in the URL are interpreted. This is the only shell-interpreting spawn left in the tree; all four `exec.ts` spawn paths are `shell: false`.

## Fix

Launch via `rundll32.exe url.dll,FileProtocolHandler <url>` or `explorer.exe <url>` as inert argv. Validate the scheme is `http`/`https` and the string has no control characters before dispatch. No host allowlist: verification URIs can legitimately differ in origin from the API.

## Done when

Unit test: a URL containing `&calc` never reaches a `cmd` spawn. Cannot be exercised on macOS; test-only.

thymikee:
## Summary

`agent-device auth login` opened the device-auth verification URI on Windows with `cmd /c start "" <url>`. `cmd /c` re-tokenizes its argument string even when spawned as argv, so `&`, `^`, or `|` in a URL taken from the auth response could run an extra command.

Windows now launches `rundll32.exe url.dll,FileProtocolHandler <url>`, and every platform passes the URL as one inert argv entry — no shell anywhere. The verification URIs are checked as the start response is validated: `http`/`https` only, and no control or format characters, so neither the launcher nor the terminal sees anything else. No host allowlist.

The launcher decision and the URL guard moved out of `src/cli/auth-session.ts` into `src/cli/browser-launch.ts`, which the auth flow reaches at its call sites so CLI startup keeps evaluating what it did before, and which stops reading a launcher's exit status as a handoff report on Windows, where a successful handoff can exit non-zero. Tests record the argv a flow would have spawned through the existing command-executor seam. Five files: the two sources, their tests, and a `host-execution.ts` test util for platform mocking and spawn capture.

Closes #2526

## Validation

Head `5c20b12f11`: `pnpm check:affected --run` passed, `pnpm test:unit` passed (1283 files, 9886 tests), as did `check:quick` (lint + typecheck), `check:layering`, `check:di-seams`, `check:production-exports`. The `scripts/__tests__/eager-closure-budgets.test.ts` entry probe for `src/cli.ts` passes again — its failure at `7cca9db4eb` is what the second commit fixes.

Each new assertion was planted-and-caught: `cmd /c start` on win32 fails the two Windows launch tests plus the flow test; dropping the `verificationUriComplete` guard, the `\p{Cf}` class, the launcher's own URL check, or the Windows exit-status tolerance each fails exactly its test. The Windows branch is exercised on macOS by mocking `process.platform` and recording argv, which is what the issue calls for; no Windows host ran the real `rundll32` handoff, which is why its exit status is ignored rather than trusted.


thymikee:
No findings at 5c20b12. The auth flow validates the URLs before printing or launching them, and Windows passes the URL without cmd parsing. The platform-mocked coverage satisfies #2526; checks are green and this is ready for human review.

PR #2568, opened 2026-09-13
fix(android-helper): build with the pinned build-tools version, fail on CI when unset

### LINKED ISSUE #2527: fix(android-helper): build with the pinned build-tools version, fail on CI when unset
## Defect

`scripts/build-android-helper.sh:59-65` picks build-tools with `find … | sort -V | tail -1`, whichever is newest on the image, while every workflow that builds the helper installs exactly `build-tools;36.0.0` (`release-android-snapshot-helper.yml:49`, `size.yml:61`, `test-app-build-cache.yml:180`, `.github/actions/setup-android-replay-host/action.yml:51-62`). The selection feeds `d8` (`:107`) and `aapt2` (`:114-128`), so a drift changes the compiled bytecode and resources, not just packaging.

## Fix

The script reads the build-tools version from an explicit input (env `AGENT_DEVICE_ANDROID_BUILD_TOOLS` or a third positional). On CI (`CI=true`) an unset value is a hard failure. Locally the `sort -V` fallback may remain with a stderr line. The draft plan's "stderr line only, even on CI" reproduces the silent-mismatch mode this issue closes; reject it. No shell-level test file.

## Done when

The `android-helpers` CI gate (`localRunnable: false`) is green with the pinned value passed, and fails on a branch that unsets it.

thymikee:
## Summary

`scripts/build-android-helper.sh` compiled with whichever build-tools directory was newest on the image while every lane that builds a helper installs exactly `build-tools;36.0.0`. That choice feeds `d8` and `aapt2`, so drift silently changed helper bytecode and resources.

The script now takes the version as an input — last positional or `AGENT_DEVICE_ANDROID_BUILD_TOOLS` — resolves it under `$SDK_ROOT/build-tools`, and checks the four tools the build runs instead of `aapt2` alone. Unpinned on CI is a hard failure; locally the newest-installed fallback stays, announced on stderr.

The three helper-building lanes interpolate the version into both their `sdkmanager` line and the build environment, so install and build cannot drift inside a lane; the replay-host cache keys name it too. Docs updated. 7 files, no shell test per the issue. A repo-wide version registry was rejected: each lane installs its own packages.

Closes #2527

## Validation

Commit `6078e44a83`, macOS, SDK with build-tools 30.0.3–36.0.0.

- `CI=true AGENT_DEVICE_ANDROID_BUILD_TOOLS=36.0.0 pnpm build:android` — the gate's own command — exits 0 with both APKs, checksums, and manifests.
- `env -u AGENT_DEVICE_ANDROID_BUILD_TOOLS CI=true pnpm build:android` exits 1 with `AGENT_DEVICE_ANDROID_BUILD_TOOLS must name a build-tools version under …` — the issue's fail-on-unset condition.
- Pinned, fallback, positional-override, missing-version, partial-install, and malformed-value paths ran under `/bin/sh` and `/bin/dash` against a fake SDK.
- `pnpm format` changed nothing; `lint`, `gate-manifest`, `gate-manifest-model`, `affected-selector`, `agent-guidance`, `command-docs` gates green. `pnpm gate unit`: 9871 passed — `client-metro.test.ts` timed out under parallel load and passes in isolation; no TypeScript touched.
- `android-helpers` is `localRunnable: false`, so its green run lands in CI on this head.


thymikee:
No findings at 6078e44. Each helper build lane installs and passes the pinned build-tools version, and an unset CI value fails explicitly. The reported build checks and current CI are green; ready for human review.

Issue #2458, opened 2026-09-10
Same published version installed twice replaces the daemon and drops the live session


Popalay:
## Summary

Two installs of the **same** published version produce different code signatures, so the second one replaces a running daemon and the open session is lost. The trees are byte-identical; only their mtimes differ, because npm stamps fresh mtimes on every install.

I am not suggesting `size:mtime` is the wrong stamp. `code-signature.ts` documents it as a deliberate stand-in for contents, and `code-signature-cache.ts` depends on replaying from `statSync` alone. The question is narrower: whether the signature should be consulted at all for an **installed** tree, where the version is already authoritative.

## Repro

With one device attached:

```bash
npm install --prefix /tmp/a agent-device@0.20.10
npm install --prefix /tmp/b agent-device@0.20.10

A=/tmp/a/node_modules/.bin/agent-device
B=/tmp/b/node_modules/.bin/agent-device

"$A" open com.android.settings --relaunch
"$A" is visible 'text="Search Settings"'   # Passed

"$B" is visible 'text="Search Settings"'   # same version
```

Observed:

```
Replacing daemon (pid 75790, v0.20.10) in /Users/me/.agent-device: code-signature mismatch
Error (SESSION_NOT_FOUND): No active session. Run open first.
```

Returning to `$A` replaces the daemon a second time. The session is unrecoverable.

The two trees are identical in content. All 319 files hash equal:

```bash
(cd /tmp/a/node_modules/agent-device && find . -type f | sort | xargs shasum) > /tmp/a.sums
(cd /tmp/b/node_modules/agent-device && find . -type f | sort | xargs shasum) > /tmp/b.sums
diff /tmp/a.sums /tmp/b.sums   # no differences
```

The only difference is the timestamp, two seconds apart because the installs ran back to back:

```bash
stat -f '%m' /tmp/a/node_modules/agent-device/package.json   # 1789044620
stat -f '%m' /tmp/b/node_modules/agent-device/package.json   # 1789044622
```

Repeating a command from the same install never prints `Replacing daemon`, and the session survives.

## Why an installed tree seems different from a source tree

`resolveDaemonTakeoverReason` in `src/daemon-client/daemon-client-lifecycle.ts` checks the version first, then the signature:

```ts
if (info.version !== readVersion()) return `version mismatch (client v${readVersion()})`;
if (info.codeSignature !== (await resolveLocalDaemonCodeSignature())) {
  return 'code-signature mismatch';
}
```

[...]

Issue #1457, opened 2026-07-28
gate(mutation): ratchet graduates on stability alone — a 28.75% score can lock in as the permanently gated floor


thymikee:
## Context

The mutation gate's graduation logic in `scripts/mutation/ratchet.ts` promotes a decision kernel from advisory to gating when its score is *stable* for `requiredStableRuns: 2` consecutive weekly sweeps (`mutation-baselines/decision-kernels.json`, `mutation-weekly.yml`, `mutation-affected.yml`; umbrella #1412 / gate issue #1415). There is **no minimum-score condition**. Verified consequence (2026-07-28 audit, adversarially confirmed): `scroll-edge-state` sits at 28.75% (228 survived mutants) — if that number is merely *stable* twice, it graduates and the ratchet then enforces 28.75% as the protected floor forever. Stability of a bad score is not quality; the gate would begin blocking unrelated PRs to defend a floor that detects almost nothing.

Companion issues #1455 / (kernel-errors) raise the actual scores; this issue fixes the policy so graduation can't outrun them.

## Proposed change (needs maintainer sign-off on the threshold)

Add a minimum-score condition to graduation: a kernel becomes gating only when `score >= FLOOR` (proposal: **70%**) AND stability is met. Below-floor kernels stay advisory and are reported (in the weekly summary output) as "stable but below graduation floor".

## Acceptance criteria

- [ ] `ratchet.ts` graduation requires both conditions; floor is a named constant in the baseline JSON schema (per-kernel override allowed, default 70) so raising/lowering it is a data change with an audit trail, not code.
- [ ] Unit tests in `scripts/mutation/*.test.ts` (the existing `mutation:test` lane) cover: below-floor+stable stays advisory; at-floor+stable graduates; already-graduated kernels are unaffected by the new rule (no regression of existing gating).
- [ ] Weekly summary output distinguishes "advisory: unstable" from "advisory: below floor" so the state is visible, not silent.
- [ ] Docs touched: `docs/agents/testing.md` section describing graduation criteria updated in the same PR.
- [ ] `pnpm mutation:test`, `pnpm check:tooling` green. No baseline regeneration needed (policy change only).

thymikee:
Recommended policy (2026-07-28, revised after maintainer review — the earlier 5-part version over-built the justification mechanism; superseded):

**One small PR:**

[...]

PR #2569, opened 2026-09-13
fix(daemon): let an installed version stand as its whole code identity

thymikee:
## Summary

Two installs of one published version signed their identical bytes differently: the fingerprint is `size:mtime`, and an installer stamps a fresh mtime on every file. The second client read `code-signature mismatch`, replaced a daemon already running exactly its code, and dropped the live session on it.

The bypass is pairwise, not one-sided. A running daemon now publishes `codeOrigin: 'installed' | 'checkout'` beside its signature, and a client declines the comparison only where **both** sides are installs of the version they report — the only pair whose shared version already fixes the bytes both run. Every other pair is judged: checkout against checkout by signature as before, and a pair whose trees cannot match at all restarts the daemon rather than run code it cannot identify. A daemon too old to carry the field is judged by its signature, so nothing is trusted by silence.

Users see `Replacing daemon` disappear between two installs of one version; `security-trust.md` states what remains undetected (an install hand-patched under its own version).

16 files, +498/-68 across two commits, one command family.

## Validation

Commit `edff9074af`.

- `pnpm check:quick`, `check:layering`, `check:package`, `check:production-exports` pass. The packaging assertion proves a clean install ships no `src/daemon.ts`.
- `pnpm gate vitest-related <changed files>`: 4228 passed; the one failure is the known `runner-client.test.ts` subprocess contention timeout (50/50 in isolation).
- Field evidence, two clean installs of one packed version sharing `--state-dir`: install B reuses A's daemon pid silently; a checkout client against the installed daemon prints `code origin mismatch (daemon installed, client checkout)`; an installed client against a checkout daemon prints `code origin mismatch (daemon checkout, client installed)`.
- Regression proof: the one-sided bypass fails the two installed-client cases; a naive `info.codeSignature !== await ...` compare fails the #2458 reuse case.

Unresolved: `mutation-model` fails at merge-base `ab3d11e069` too (`gestures.test.ts` no longer reaches `scroll-edge-state`) — advisory-only per #1457, and its workflow is path-filtered away from this diff.


thymikee:
The signature bypass at daemon-launch-spec.ts:113 only checks the client. With an explicit shared --state-dir, a modified source checkout can start a daemon under the published version, and an installed client now reuses that modified code. #2458 scopes the bypass to both sides being installed. Record the daemon’s code origin or equivalent package identity and bypass fingerprints only for two matching installed versions; add the source-daemon → installed-client case. Checks are green, but this remains a code blocker at 5fef822.

thymikee:
Fixed in `edff9074af`. The bypass is pairwise now, and the daemon says what it is.

A running daemon publishes `codeOrigin: 'installed' | 'checkout'` beside its signature (`server-lifecycle.ts`, from `resolveDaemonCodeOrigin()` in `@agent-device/host-kit/code-signature`), and `resolveCodeIdentityMismatch` in `daemon-launch-spec.ts:150` holds the bypass only where both sides report `installed`. A checkout still judges a checkout by signature; a pair whose trees cannot match at all restarts the daemon instead of running code it cannot identify — which is also the rule a daemon too old to carry the field is judged by, so no published daemon is trusted by silence.

Proof against two clean installs of one packed version, one shared `--state-dir`:

```
installed A -> installed B (dist mtimes shifted)   reuses pid 38232, stderr empty
checkout client -> installed daemon               code origin mismatch (daemon installed, client checkout)
installed B -> checkout daemon                    code origin mismatch (daemon checkout, client installed)
```

The third line is the case you flagged. Two tests pin it (`daemon-launch-spec.test.ts`), and restoring the one-sided bypass fails exactly those two. `daemon-client-metadata.test.ts` is the new mirror for the round trip through `daemon.json`, including a value no writer produces.

Local: `check:quick`, `check:layering`, `check:package`, `check:production-exports` green; `gate vitest-related` 4228 passed with the known `runner-client.test.ts` contention timeout (50/50 in isolation).

thymikee:
The previous blocker is fixed at edff907: fingerprint bypass now requires both client and daemon to be installed copies of the same version. The origin round trip, cross-origin rejection tests, and reported two-install field checks cover the failure case. Review is clean, all checks pass, and there are no merge conflicts.

PR #2572, opened 2026-09-13
fix(apple): carry the runner's sparse verdict through a viewport refusal

thymikee:
## Summary

A capture that no backend could read failed with an internal engine invariant — `regular iOS snapshot presentation requires a valid viewport` — naming neither the backend that was asked nor the screen that defeated it. The runner already declares such a payload sparse (backend, `reasonCode`, its own reason); the daemon read that verdict and then discarded it while reconstructing a viewport from the synthetic root every sparse payload carries.

The refusal now carries `error.details.snapshotQuality`, and the hint composes the shared sparse-capture advice with the presented surface host. `replay`/`test` keep the verdict in `REPLAY_DIVERGENCE` details instead of truncating to four cause keys. Same command over an `ASWebAuthenticationSession` sheet, before: `details { reason: "invalid-viewport", field: "viewport" }` plus a generic hint. After: the same message, plus `snapshotQuality: { state: "sparse", backend: "private-ax", reasonCode: "requested-backend", reason: "the private AX backend was selected because this capture asked for accessibility custom actions" }` and a hint naming `com.apple.SafariViewService`.

Refs #2560 — diagnosis, not the flow fix. That flow still fails on this surface until the capture plan stops routing presented out-of-process surfaces to private AX, which ADR 0004 already diagnosed for #2403; that capability declaration is the follow-up. Six files.

## Validation

Commit `c37e17d3f7`: `pnpm check:affected --run` passed (1874 tests); `check:quick`, `check:layering`, `check:daemon-wire-compat` passed.

Live: iPhone 17 Pro simulators on iOS 26.2 and iOS 18.6, UIKit app presenting `ASWebAuthenticationSession`, driven by this checkout's CLI — plain `snapshot` stayed healthy (35 nodes, `backend: tree`), `snapshot --actions` produced the quoted pair before and after the change. Manual sessions closed, created simulators deleted.

Unresolved risk: whether the reporter's run armed the XCTest-channel penalty needs their runner log (`SNAPSHOT_XCTEST_CHANNEL_PENALIZED` / `_DEFERRED`).


thymikee:
No code findings at c37e17d. The runner’s sparse verdict survives presentation refusal and replay error wrapping, with reported live evidence on two iOS versions. The remaining smoke failure is the same automation-longpress scroll assertion seen on #2565 before its successful retry, so it appears unrelated; rerun that check before merge. Ready for human review.

PR #2574, opened 2026-09-13
refactor(move): settle a device claim in its own module

thymikee:
## Summary

Moves claim settlement out of `src/daemon/device-claims.ts` into `src/daemon/device-claim-settlement.ts`: what the claim file says before an acquisition writes its own record, how a foreign claim is settled, the ownership predicates, and the conflict and supersede diagnostics. The resolution gains a `held` arm so the settlement module stops naming the public acquire result, and `claimHeldDevice` maps it back to `acquired`. `DeviceClaimSessionOwnership` and `ownershipFromClaim` move to `device-claim-record.ts`, the record they read.

Tests move with the code they cover, which also leaves `device-claims.test.ts` at 767 lines instead of 940. 772 gross lines, no behavior change: the 6 moved tests pass unchanged and the public surface keeps its existing re-exports.

Part 1 of 3 for #2538; #2575 and #2570 build on it.

## Validation

Tested at `5749359a00`. `pnpm check:layering` passes, including the committed enumeration of the new module. `pnpm check:fallow --base origin/main` reports no issues in the changed files. Typecheck and lint pass. `vitest related` across all six changed files: 254 files, 1,629 tests passed. No device-facing behavior changes, so no simulator or device run applies.


thymikee:
No blocking findings at 5749359. Settlement behavior and the held-to-acquired mapping are preserved, and the moved tests and current CI are green. Ready for human review as the first part of the stack.

PR #2575, opened 2026-09-13
feat(devices): answer when a device's current boot began

thymikee:
## Summary

Adds one question to the platform boundary — when did this device's current boot begin — as `@agent-device/contracts/device-boot`. An answered probe returns host-clock `bootedAtMs`; a device family with no honest answer returns `unsupported-device` rather than a guess.

An Apple Simulator reports the start time of the `launchd_sim` process naming that device, rejecting a stamp that lies in the future. Android converts `/proc/uptime` on the host clock, bounded by the instant the probe began so response latency cannot move the answer forward, and a guest clock that disagrees with the host cannot move it at all. `src/platform-runtime-device-boot.ts` is the root hub that dispatches by family; nothing consumes it yet.

Part 2 of 3 for #2538, on top of #2574. 489 gross lines.

## Validation

Tested at `6fae37be00`. `pnpm check:layering` passes, which enumerates the two new package subpaths and the contracts export. `pnpm check:fallow --base origin/main` reports no issues. Typecheck and lint pass. `pnpm check:package` imports all 13 published entry points from a clean install and runs the published CLI.

The probes are host-process reads on the same commands the platform packages already use for boot state, so no new device behavior is exercised here; #2570 drives them against a live simulator and emulator reboot.


thymikee:
No findings at 6fae37b. The Android probe now uses a conservative pre-request timestamp, so response latency cannot falsely invalidate a newer claim. Checks are green; ready for human review after #2574. The consuming takeover behavior remains in #2570.

PR #2585, opened 2026-09-14
refactor(tests): share the router-open fixtures between its suites

thymikee:
## Summary

Follow-up cleanup from #2570. Splitting the routed claim tests out of `request-router-open.test.ts` copied three fixtures into the new suite: the simulator fixture, the `open` request builder, and the reader that lifts a claim's stamp off disk. `storedClaimUpdatedAt` had already become the place two suites re-read the same production state, and a change to the claim record now has two call sites to find.

`src/daemon/__tests__/request-router-open-harness.ts` owns those fixtures. Each suite keeps its own `createOpenHandler`, because that composition is where each one installs its mocks; sharing it changed which module instances the router saw and broke four routing assertions in this branch's first commit.

141 gross lines, three files, no production change.

## Validation

Tested at `d8374a819f`. `pnpm check:affected --run` passed every runnable check, including layering, fallow, and related vitest (18 tests across both suites), plus `pnpm check:layering`, `pnpm lint`, and `tsc --noEmit` directly. Behavior is unchanged: the same 18 tests pass before and after, and the `test-file-size-ratchet` gate stays green with `request-router-open.test.ts` at 834 lines.


thymikee:
Clean at d8374a8. The shared fixtures preserve both suites’ mock composition and test discovery; the same 18 tests remain covered. All checks pass, with no merge conflicts.

Issue #2580, opened 2026-09-14
Sessions are keyed by cwd and bound to one device, so driving iOS and Android from one checkout needs hand-named --session on every command


JKobrynski:
agent-device 0.21.0, macOS, Node 24.18. Found while driving iOS and Android from one checkout for a visual regression PoC (callstack/react-native-paper#5115).

Sessions are keyed by cwd and bound to the first device they touch. After `open … --platform ios --udid …` from a directory, running `boot --platform android --device Pixel_10_Pro` from the same directory fails:

```
INVALID_ARGS: … already bound to apple device "iPhone 17 Pro" … but this request selected --platform=android
```

The hint (pass a different `--session`) is correct and `--session android` works. But a repo that tests both platforms then has to invent and thread session names by hand through every command, and a script that forgets one flag silently talks to the wrong device. Two smaller consequences of the same design:

- a leftover session from an earlier run in a repo root kept the wrong simulator bound; the workaround was to run from a different directory rather than `close` a session another process might own;
- when this script itself ran under `npx -p node@20 node …`, the nested `npx agent-device` inherited `npm_config_package` and installed node instead. Unrelated to sessions but the same class: environment leaking into what the CLI resolves.

Suggestion: key sessions by cwd + platform by default, or make `--platform` select a per-platform default session so the common two-platform case needs no naming.

JKobrynski:
Closing: filed prematurely out of a PoC. Will re-file properly if still relevant.

PR #2587, opened 2026-09-14
fix(daemon): key the implicit session by workspace and platform

thymikee:
## Summary

An implicit session was addressed by `cwd:<workspace>:default` — one slot per checkout — so a repo driving both platforms could not open the second: after `open --platform ios`, `boot --platform android` from the same directory failed with `INVALID_ARGS` against a session it never asked for.

`--platform` now selects the implicit session, so one workspace holds both:

```
agent-device open --platform ios --device "iPhone 17 Pro"      # cwd:<hash>:ios
agent-device boot --platform android --device Pixel_9_Pro_XL   # cwd:<hash>:android
agent-device snapshot       # AMBIGUOUS_MATCH, naming both + --platform hint
agent-device session list   # still routes, so the ambiguity stays resolvable
```

A platform-naming request joins a session it genuinely agrees with, so a session opened without `--platform` stays reachable and keeps its artifact directory; a same-platform device or target mismatch keeps the existing refusal instead of forking a twin. Ambiguity is refused, never resolved by open order. Also fixes a settling interaction republishing the session under `SessionState.name`, which `session list` already showed as a phantom row.

12 files, all session-routing plus docs. References #2580, which was self-closed as premature but still reproduces on 0.21.1.

## Validation

Rebased on `origin/main` at `1b6515b`. Head `9ffcb69`: `pnpm check:affected --run`, `check:fallow --base origin/main`, `tsc --noEmit` and `oxlint` pass; `src/daemon` is 2757 tests.

Live, built CLI, fresh checkout per case:

- iOS opened without `--platform`, then `open --platform android`: `session list` shows `:default` and `:android`, bare `close` refuses with `AMBIGUOUS_MATCH` naming both, and `close --platform ios` / `--platform android` each close their own.
- `AGENT_DEVICE_PLATFORM=android AGENT_DEVICE_SESSION_LOCK=reject` against an existing iOS session routes `boot`/`open` to `cwd:<hash>:android` and the locked `snapshot` returns an Android tree instead of running on iOS.
- `open --platform android` → `snapshot` → `press @e2 --settle` held `session list` to one address; reverting only the interaction fix reproduced the phantom `default` row.

Follow-up, not in scope: `--session` accepts address-shaped values like `cwd:<hash>:ios`, which can squat an implicit key (pre-existing).


thymikee:
Two routing blockers at 9ba53df.

In session-routing.ts:213, the default-address return bypasses the ambiguity check. Open iOS without --platform, then Android with --platform android: a bare press or close now silently selects iOS instead of AMBIGUOUS_MATCH. Check all implicit candidates before preferring the default leaf, and cover default + android through the attaching-command path.

In session-routing.ts:157, candidate matching uses req.flags even when the platform came from meta.lockPlatform. With an existing iOS session and an Android lock, empty flags make iOS appear to agree. The later lock policy leaves the platform unset because a session is already bound, so the command runs on iOS. Include the effective platform in candidate matching and cover an existing opposite-platform session, not only an empty store.

Checks are green and there are no merge conflicts. The reported live runs do not cover these two combinations; both need regression proof before this is ready.

thymikee:
Both fixed at `9ffcb69`, rebased on `origin/main` at `1b6515b`.

**session-routing.ts default-leaf bypass** — `resolveUnselectedSessionAddress` no longer short-circuits on `sessionStore.get(defaultAddress)`. The `default` leaf is now one candidate among the workspace's implicit sessions: a lone session still lands there, but a workspace holding `default` (iOS, opened without `--platform`) beside `:android` refuses with `AMBIGUOUS_MATCH` instead of silently selecting the iOS device. Covered through the attaching-command path (`close`) in the default+android shape.

**session-routing.ts lockPlatform matching** — candidate matching now reads the effective platform, not just `req.flags`, via `withImplicitPlatform`. With an existing iOS session and an Android lock on empty flags, the iOS session is seen to disagree on platform and the request resolves its own `:android` leaf rather than joining iOS where the lock policy would leave the platform unset. Covered with the opposite-platform session present.

Both also proven live on the built CLI:
- iOS without `--platform` + `open --platform android` → `session list` shows `:default` and `:android`, bare `close` refuses with `AMBIGUOUS_MATCH`, and `close --platform ios`/`--platform android` each close their own.
- `AGENT_DEVICE_PLATFORM=android AGENT_DEVICE_SESSION_LOCK=reject` against an iOS session routes `boot`/`open` to `cwd:<hash>:android` and the locked `snapshot` returns an Android tree — not iOS.

thymikee:
Both routing fixes look correct at 9ffcb69. A bare command now refuses with `AMBIGUOUS_MATCH` when the workspace holds `default` next to `:android`, and candidate matching uses the lock platform, so an Android lock no longer joins the iOS session. The new tests fail on the old code, and your live runs cover both cases.

The failing Smoke Tests job looks unrelated. The iOS Settings replay stopped at its `screenshot` step after a runner restart, and `open` had already worked in the same session. The same job passed on main at 730b201. Please rerun it before merge. No conflicts.

PR #2590, opened 2026-09-14
refactor(cli-schema): orient the schema layer above commands (#2543 step 3)

thymikee:
## Summary

Step 3 of #2543. Orients the CLI/MCP schema layer **above** commands (not below) — the honest direction: a package cannot import `src`, and `cli-schema` renders what the command facets author. Both zones are rank 3, so the gate never ordered them; R2 declares it now.

- Move `CommandSchema` / `CommandSchemaOverride` from `src/cli-schema/types.ts` into `packages/command-registry` (`@agent-device/command-registry/command-schema`), beside the flag grammar they already import. Retires the 22 forward `commands → cli-schema` type edges.
- `cli-config.ts` takes `projectConfigFlagKeys` from the registry directly instead of through `command-schema.ts`.
- Relocate `command-explain.ts` from `src/commands` into `src/cli` (it renders the schema and already types against the daemon route union) — the last `commands → cli-schema` value edges, and it also retires the last `commands → daemon-server` type inversion.
- Gate: `R2 commands-floor` now rejects any `commands → cli-schema` import of any kind.

After the change `commands → cli-schema` is **0 of any kind**; `cli-schema → commands` (reading the facet aggregate) is the declared direction. No CLI/help/MCP behavior change. 35 files, +79/-44.

## Validation

Head `9e61949`. Green on the commit: `check:layering` (235/235 + check OK; R4=0, R5=0, **R6 4→3**, R9 6, R11 301 subpaths), `typecheck`, `check:package`, `check:mcp-metadata`, `check:production-exports` (parity 68), `check:fallow`, `eager-closure-budgets` (612), help-conformance, `command-doc-coverage`, and `check:affected`'s `vitest-related` lane (1258 tests). `lint` + `format:check` clean.

Pre-existing, unrelated: `check:affected`'s `mutation-model` stage fails on `scripts/mutation/ownership.test.ts` (`gestures.test.ts → scroll-edge-state`). Verified it fails **identically at the merge-base `730b20149e`** — the same failure #2562 recorded. A parallel `lint` lane flaked under concurrent subprocess load; standalone `pnpm lint` is exit 0.


thymikee:
Looks good at 9e61949. The schema types move into `@agent-device/command-registry/command-schema` with no shim left behind, `command-explain` moves to `src/cli` with its test, and the new R2 row rejects every kind of import from `commands` into `cli-schema`. I found no CLI, help or MCP behavior change. CI is green, the size report shows no growth, and there are no conflicts.

PR #2591, opened 2026-09-14
fix(record): keep a failed capture's evidence for the retry

thymikee:
## Summary

Implements wave 0 of `docs/adr/0024-screen-recording-provable-signal.md` rule 6: an export failure is no longer a disposal. A durable kind now says what its retry needs with `failedFinishPolicy`, and the shared coordinator stops force-cleaning on a failed finish for preserving kinds (`screen-recording`, `perf-capture`); `app-log` and `audio-probe` say `dispose-on-failed-finish` explicitly.

A finish also says why it was asked for. `capture` may preserve retry material; `disposal` — session teardown, the only such caller — disposes whatever a failed finish left. Recording and perf-capture would otherwise have left a closing session unable to hand its recorder back.

`createCaptureHandle.finish` no longer memoizes a refused finish, so a second `record stop` re-enters the recorder. 29 files, one command family plus the coordinator it sits on.

```
agent-device record stop   # adb pull failed: error returned, device artifact + manifest kept
agent-device record stop   # same session: exports the MP4 the first stop left behind
```

## Validation

Tested at `4e1e17084b`. `pnpm check:affected --run` passed locally (372 files, 2,560 tests), along with `pnpm typecheck`, `pnpm lint` and `pnpm check:layering`.

Live proof on Android emulator `emulator-5554` (Pixel 7 CI) with the pull path patched to fail its first three attempts: `record start` / `record stop` returned the pull error and wrote no local file, while `adb -s emulator-5554 shell ls -l /sdcard/agent-device-recording-*` showed the 90,338-byte chunk and `agent-device-recording-active.json` intact and the session manifest `open`. A second `record stop` exported those bytes and cleared the device. Patch reverted before the pushed head.

Known follow-up, not fixed here: on that retry Android infers "stopped before record stop, likely the 180s limit" from its own earlier stop — ADR step 3 recorder observation owns it.


thymikee:
Keeping a failed stop's evidence fixes the Android `adb pull` case you proved. But the preserve policy now applies to every screen-recording backend, and on the Apple backends the retry cannot succeed.

In `packages/platform-apple/src/recording/runtime.ts`, `stopRunner` stores its promise with `runnerStop ??=`. After one rejected runner stop, every later `record stop` gets the same rejection without calling the runner again. The simulator path has the same shape: the `simctl recordVideo` process has already exited, so the next finish reads the same non-zero exit code and throws again. In both cases `record stop` fails every time, and `record start` keeps refusing with "recording already in progress" until the session closes. Before this change the failed stop cleaned up, so a new `record start` worked. Please make the retry real, or dispose when nothing can be retried: clear `runnerStop` when it rejects, and dispose (or point the user at closing the session) when the recorder process already exited with an error. A test with a repeated Apple stop failure would cover it.

The iOS and macOS recording paths change behavior here, but the only live run is on Android. A simulator run of a failed and then retried `record stop` is still needed. Smoke Tests were still running at 03cfd29, and there are no conflicts.

thymikee:
Update: after #2587 landed, the branch conflicts with main in `CHANGELOG.md`. Please resolve it when you address the Apple stop-failure test and the simulator run.

thymikee:
Rebased on `main` (`CHANGELOG.md` conflict resolved, #2587's entry kept) and made the Apple retry real. Head: `7466adbb`.

**Runner stop** — `stopRunner`'s `runnerStop ??=` cached the rejection, so the runner was asked once and every later stop got that cached answer. Only an in-flight or completed stop is shared now, which is what `createScreenRecordingLiveHandle` already does for a refused finish. `a refused runner stop is asked again by the next record stop` asserts two `stop` requests and a completed second attempt.

**simctl exit** — you're right that a settled exit can't be re-driven, so I stopped treating it as a failure. An exit is an observation about the recorder; the finalizer is what answers whether the file is a video (ADR 0024 2.2). The exit is disclosed (`exited with code 1`, `was killed by SIGTERM`) and collection proceeds, so recordings the old code threw away now export. Covered by `a simulator recorder that exited early is collected and its exit is disclosed` and `a refused simulator finish is re-driven by the next record stop` (finalizer called twice).

**When nothing can be retried** — a recorder killed mid-write leaves zero bytes, and no retry fixes that. That error now carries `recording-output-unplayable`, stays retriable for the case a retry does fix, and names session close as the way out.

**Simulator run** (iPhone 16 Pro, booted):
- finalizer patched to fail its first call: stop 1 returned the error with the 61,935-byte recording untouched and the manifest `open`/`completing`; stop 2 exported those bytes.
- recorder killed by SIGTERM: `recording was not finalized into a playable video` + the new hint; the retry re-drove it; `close` terminalized the manifest (`completed`, `cleanupStatus: cleaned`) and reported the lost export.
- healthy start/stop on this build: 119,935 bytes, no warning.

Killing the recorder left the CoreSimulator host recording lock (simctl exit 16) — a simulator shutdown/boot clears it. Nothing left behind: no `recordVideo` processes, no stale claims, throwaway state dir removed.

`pnpm check:affected --run` at `7466adbb`: 405 files, 2,795 tests, plus `typecheck`, `lint`, `check:layering`.

thymikee:
The Apple retry works now at 7466adb. A refused runner stop is asked again, and a simulator recorder that exited early goes to the finalizer, which still refuses a file that is not a playable video. Two gaps remain in the simulator path.

When simctl exits with an error and the file is unplayable, the error no longer carries the exit code, signal or stderr. Before this change the error named the exit code. Now the user gets only the generic `recording-output-unplayable` hint and must guess whether a retry can help, and each retry waits again for a file that will not change. Please add the recorder's exit details to that error in the simulator `finish` in `packages/platform-apple/src/recording/runtime.ts`, and point the hint at closing the session when the recorder has already exited.

The new warning can blame the wrong side. When our own stop escalates from SIGINT to SIGTERM or SIGKILL, `src/platform-runtime-screen-recording-apple-simulator-host.ts` resets the exit code to 0 but keeps `signal`. `describeSimulatorRecorderExit` checks `signal` first, so it reports `was killed by SIGTERM before record stop` for a kill that record stop sent. Please skip the warning, or change its wording, when the owner terminated the recorder.

Also, `a refused simulator finish is re-driven by the next record stop` uses `exitCode: 0`, and main already re-drives that case, so the test passes without this change. A non-zero exit there would cover the new behavior. Your simulator run covers the retry path. Smoke Tests were still running, and there are no conflicts.

thymikee:
All three addressed at `c9ed057d`.

**Exit details** — the simulator finish now wraps the export failure instead of replacing it:

```
recording was not finalized into a playable video: /tmp/x.mp4; simctl recordVideo exited with code 1: recordVideo lost its stream
```

`stderr`, `exitCode` and `signal` ride in `details` under `processExitError`, which is what appends the recorder's first stderr line, `recording-output-unplayable` is preserved, the retriable flag is cleared — a settled exit cannot be retried into new bytes — and the hint points at closing the session. `retriable: true` and the retry hint stay for the backends where the recorder may still be finalizing. Covered by `an unreadable recording names the exit that made it permanent and the way out`.

**Wrong blame** — you read the host right. `src/platform-runtime-screen-recording-apple-simulator-host.ts:200` normalizes a termination record stop *asked for* to `exitCode: 0` and keeps the signal, so checking `signal` first turned our SIGINT→SIGTERM escalation into "the recorder was killed". The exit code answers first now, and `a recorder that record stop terminated itself is collected without a warning` pins it.

**Test** — the re-drive test uses `exitCode: 1`, asserting the wrapped refusal on the first stop and the completed second one.

Went back over the whole change while doing this: three finishes that only make room were labelled `capture` — `audio start` over a running probe, `audio status` on a sampler that already exited, `logs clear --restart` before it deletes the log files. Nobody reads those completions and an open record left behind would refuse the start that follows, so they are `disposal` (`ddeb9ebf`). No behavior today — both kinds dispose on a failed finish — but the labels were a trap for any kind that later preserves.

`pnpm check:affected --run` at `c9ed057d`: 405 files, 2,797 tests, plus `typecheck`, `lint`, `check:layering`. One earlier run of the same head failed four `provider-integration` files at the 5s timeout under machine load; all four pass in isolation and on the clean re-run.

thymikee:
All three points from the last review are fixed at c9ed057: the unplayable error carries the recorder's exit, our own stop no longer gets blamed, and the re-drive test fails on main. Two small issues remain in the new wrap.

[`recorderExitLostTheExport`](https://github.com/callstack/agent-device/blob/c9ed057d213780c02f31e7cdc5cb7d39ff7132f1/packages/platform-apple/src/recording/runtime.ts#L129-L139) wraps every error from `completion`, not only an unplayable file. A transient telemetry write or finalize failure also gets `retriable: false` and the "close this session" hint, and the original `hint`, `diagnosticId` and `logPath` are lost. So a failure that a retry would fix tells the user to give up. Please wrap only when `reason === 'recording-output-unplayable'`, or keep the original details and add the exit facts to them.

The exit details are built by hand. `execFailureDetails` from `@agent-device/host-kit/command` already builds `stderr`, `exitCode` and `processExitError`, and `core/devicectl.ts` uses it.

Smoke Tests are still running, and there are no conflicts.

thymikee:
Both fixed at `2a8e3306`.

**Only the unreadable file is final** — the wrap is now gated on the reason, and keeps what it was handed:

```ts
const original = asAppError(exportError, 'COMMAND_FAILED');
if (original.details?.reason !== RECORDING_OUTPUT_UNPLAYABLE_REASON) return exportError;
return new AppError(original.code, `${original.message}; ${exit}`, execFailureDetails(result, {
  ...(original.details ?? {}),
  ...(result.signal === undefined ? {} : { signal: result.signal }),
  retriable: false,
  hint: '…Close this session to release the device, then record again.',
}));
```

Anything else the export path raised returns untouched, so a transient telemetry or finalize failure keeps its own code, `retriable`, `hint` and whatever `diagnosticId`/`logPath` it carried. `an export failure the recorder did not cause keeps its own verdict` asserts that with `rejects.toBe(transientError)`, and the re-drive test now throws a realistic unplayable error instead of a bare `Error` — which is why it caught this.

The reason both layers match on is now one exported `RECORDING_OUTPUT_UNPLAYABLE_REASON` in `packages/contracts/src/screen-recording-runtime.ts`, used by the finalizer that raises it and the simulator that reads it.

**`execFailureDetails`** — used, and the hand-built block is gone. Its parameter needed one widening: `exitCode: number | null`, because a process killed by a signal reports no code — the null `AppErrorDetails.exitCode` already documents. Callers passing `ExecResult` are unaffected.

`pnpm check:affected --run` at `2a8e3306`: 762 files, 5,753 tests, plus `typecheck`, `lint`, `check:layering`.

thymikee:
Reviewed at 2a8e330. The retry-path evidence issue from the c9ed057 review is fixed, and I found no new problems in the delta.

One question, not a blocker: `execFailureDetails` spreads `original.details` after the recorder's `stdout`, `stderr` and `exitCode`. If a later unplayable error carries its own `stderr` or `exitCode`, it will override the recorder's values instead of adding to them. Today the finalizer's error carries none of these fields. Is that order intended?

The iOS simulator smoke job fails with `id="automation-longpress" did not become visible after scrolling`. That step does not go through the record-stop path this PR changes, and #2491 tracks rotating iOS smoke failures on main, so it looks unrelated.

thymikee:
Reviewed at 2a8e330. Both points from the last review are addressed, and the delta raises nothing new. There are no conflicts.

Smoke Tests failed in the iOS simulator fixture run because `automation-longpress` did not become visible after scrolling. This change only shapes the error when the simulator recorder exits before `record stop`, which that scroll step does not reach, so the failure is likely unrelated. A rerun should confirm it.

Issue #1658, opened 2026-08-07
iOS on BrowserStack: snapshot fails SESSION_NOT_FOUND on live sessions; fill silently no-ops or mis-navigates on WebView inputs


TilakVattikuti:
## Summary

Two related bugs found testing an iOS app against BrowserStack App Automate real devices via `agent-device` 0.20.5, both isolated to the cloud-webdriver iOS path (local iOS and Android+BrowserStack were not affected).

## Bug A: `snapshot` fails with SESSION_NOT_FOUND on a genuinely live, verified session

```bash
CONNECT_OUT=$(agent-device connect browserstack --platform ios --device "iPhone 16" \
  --provider-os-version 18 --provider-app bs://<app-id> --force --json)
SESSION_NAME=$(echo "$CONNECT_OUT" | jq -r .data.session)
agent-device open com.example.app --session "$SESSION_NAME"   # "Opened: ..."
agent-device snapshot -i -c --session "$SESSION_NAME"
```

```
Error (SESSION_NOT_FOUND): iOS snapshot requires an active app session on the target device.
Run open first (for example: open --session browserstack:adc-xxxxxx --platform ios --device "<name>" <app>).
```

This happens instantly (0ms in the request log) with no `runner.log` ever created for the session — looks like a local pre-check, not a real driver round-trip. At the same moment:
- `agent-device connection status --session "$SESSION_NAME" --json` reports `connected: true`, `liveSession.status: "created"`.
- `agent-device artifacts --json --session "$SESSION_NAME"` returns a real `providerSessionId` with a working BrowserStack video URL.
- `press`, `find`, `fill`, `screenshot`, `appstate` all work normally against the exact same `--session` on the exact same live device.

Reproduced 100% of the time across 4 independent device/OS combinations: iPhone 16/iOS 18, iPhone 17/iOS 26, iPhone 16 Pro/iOS 18, iPad Pro 11 2025/iOS 26.

Not a duplicate of #1596 — that issue is about a daemon-*replacement* crash path (already fixed via #1603). This reproduces with no daemon replacement involved, against a session that was never anything but healthy.

## Bug B: `fill` silently no-ops or mis-navigates on WebView inputs

Target: a Safari view controller (`ASWebAuthenticationSession`) hosting a web-based OAuth/SSO login page, opened from the native app.

```bash
agent-device fill 'editable=true' "user@example.com" --session "$SESSION_NAME"
# -> "Filled 13 chars"
[...]

PR #2592, opened 2026-09-14
fix(limrun): witness the field a fill taps before typing

### LINKED ISSUE #2578: Limrun iOS: Flutter text entry fails with “No focused accessibility element found” after tapping an editable field
## Reproduction and observed behavior

Using agent-device 0.21.0 against a Flutter login screen on a Limrun iOS cloud simulator, snapshot exposes Email and Password as editable text fields. Filling Email, or tapping the field and using type, fails with:

```text
No focused accessibility element found — Unexpected internal error. Retry with --debug and report the diagnostics log if it persists.
```

The first attempt may display the iOS slide-to-type tutorial. Dismissing it and taking a fresh snapshot does not resolve subsequent typing failures. Repeated ref/coordinate focus and type attempts still fail.

The same Flutter app accepts input locally. A React Native app tested on Limrun does not exhibit the failure. This suggests a compatibility problem in this path, not proof of a particular Flutter engine defect. Exact Flutter/iOS versions and a standalone public reproduction are still needed.

## Current implementation

At e3880f10c826fea373cc48153996b372ef39136e, `packages/provider-limrun/src/ios.ts` sends both `type` and `fill` through `client.typeText`; `fill` first taps the supplied coordinates. The repository resolves `@limrun/api` 0.24.5, whose `typeText` signature has no focus-check option.

## Confirmed downstream workaround

We patched the bundled iOS client’s `typeText` request sender in agent-device 0.21.0:

1. Send the normal request first.
2. Only on the exact server error `No focused accessibility element found` (without the CLI’s appended diagnostic hint), retry once with the same `text` and `pressEnter` and `requireFocus: false`.
3. Propagate all other failures and any failure of the recovery request; never replay on a timeout or unknown failure.

We confirmed that this restores typing in the affected Flutter app. The workaround stays inside agent-device; the QA application does not call Limrun directly. This is a tested downstream workaround, not validation of a patch against current main.

## Proposed upstream direction

Use a supported SDK exposing `typeText(text, pressEnter, { requireFocus: false })`, and agree on a typed pre-input focus rejection or an explicit focus-bypass policy. Current upstream guidance forbids branching on error-message text; the temporary workaround should not be copied unchanged.

Do not bypass the focus check globally or retry after potentially partial writes. Preserve type/fill semantics and verify the intended field’s value. Keep keyboard tutorial handling separate.

Validation should include normal input, focus-check rejection, timeout/no replay, unsuccessful recovery, delayed character entry, existing-value fill, password fields, and live Flutter/React Native verification on Limrun. An SDK upgrade must also be validated against the provider’s existing lifecycle and recording API usage.

Would maintainers prefer a typed SDK error prerequisite or an explicit adapter option for input after a resolved field tap?

thymikee:
## Summary

`fill` on Limrun iOS failed with "No focused accessibility element found" on apps that expose fields without publishing a globally focused element - Flutter forms are the common case (#2578). The iOS interactor now witnesses text entry itself: it records which elements cover the target point, taps, and types only once an element carrying the per-node `IsEditing` trait covers that point or was among those that did. An editing element elsewhere is refused as `text_entry_focus_not_observed` rather than typed into (#1658). Replace and clear ride the field's own select-all, so `fill <target> ""` clears. `@limrun/api` moves to 0.49.3 for the typed element tree whose traits and frames make the witness possible. 16 files, +658/-75.

```
$ agent-device fill 'label="Email"' "ada@example.com" --json
{"ok":true,"data":{"message":"Filled 16 chars","textEntryReadiness":"focused-element"}}
```

Addressing a static text instead refuses with `reason: "text_entry_focus_not_observed"` and `editingElementObserved: true` instead of typing into whichever field happens to hold focus.

Closes #2578. Refs #1658.

## Validation

Size on `03e227d406`: +23.1 kB unpacked, +15.4 kB download, startup flat; accounted for in the size comment.

`check:package`: `check:package`, `check:layering`, `check:fallow --base origin/main`, lint, typecheck, docs/MCP/gate checks, and the unit and provider-integration lanes pass. `check:affected` halts at `scripts/mutation/ownership.test.ts`, which fails identically at `origin/main` (`730b20149e`).

Live on a Limrun iOS instance at `03e227d406`: the #2578 Flutter form fills, replaces and clears, each reporting `focused-element`, and `fill` on static text while a field holds focus is refused. A rebuilt variant with two unlabeled identical fields fills and clears each one, witnessed by geometry, with the text staying in the field that was aimed at. A React Native form fills, replaces, clears, and fills a second field with the keyboard up via `id=` selectors.

Left open: replace and clear are not value-verified; the refusal vocabulary is duplicated in `provider-webdriver`, whose eager-import budget forbids a shared module.


thymikee:
Pre-existing bug found while collecting evidence on this device, unrelated to this PR and left unfixed (different command family): `agent-device screenshot` on a Limrun iOS instance fails although the bytes are written.

```
$ agent-device screenshot /tmp/shot.png
Error: Screenshot file is not a valid PNG: /tmp/shot.png   (diagnostic id mu1aosza-7b8b6170)
$ file /tmp/shot.png
JPEG image data, JFIF standard 1.01, ... 402x874, components 3
```

`@limrun/api` documents `ScreenshotData.base64` as JPEG in both 0.24.5 and 0.49.3, and the provider writes those bytes verbatim, so `readScreenshotResultMetadata` in `packages/capture-kit/src/screenshot-density.ts` rethrows from `readPngSize` whenever `supportsScreenshotPixelDensity(device)` is true. The published 0.21.1 write path is identical, so this does not come from this diff.

thymikee:
The fill check can still type into the wrong field. `limrunTextEntryIdentity` in `packages/provider-limrun/src/text-entry-focus.ts` builds an identity from pid, type, `AXUniqueId` and `AXLabel`, and `witnessesThisTap` accepts an editing element when that identity was under the tap point. Two Flutter text fields with no id and no label have the same identity. If field B already has focus and `fill` taps field A, the first poll sees B editing, finds its identity among the elements under the point, and returns `focused-element`. Cmd+A then replaces B's text, which is the #1658 misdelivery this change should refuse. Please trust identity only when it is unique in the pre-tap tree and fall back to the geometry check otherwise, and add a test with two unlabeled fields where the other field holds focus.

The Size Report shows +77.2 kB unpacked. It looks like most of that is the larger bundled `@limrun/api` 0.49.3 plus the `@limrun/xdelta3-wasm` it brings. Please explain in the PR why the witness needs this SDK version (for example, whether 0.24.5 already returned `traits` in its tree) and why a smaller option did not work.

The failing Smoke Tests job looks unrelated: it is the native runner test `testAlertAcceptDoesNotActivateAReplacementWithASharedButton`, and this PR does not touch the Apple runner. No conflicts.

thymikee:
## Package size accounting

Measured with `pnpm pack` locally, matching the Size Report:

| | `origin/main` | this PR | diff |
|---|---:|---:|---:|
| unpacked | 4.53 MB | 4.61 MB | +77.2 kB |
| download | 1.34 MB | 1.38 MB | +38.2 kB |
| `CLI --version` startup | 24.8 ms | 25.0 ms | +0.2 ms |

The whole delta is the `@limrun/api` bump, which the fix needs rather than prefers: `requireFocus: false` - the option that skips the server-side focus scan #2578 dies on - does not exist before 0.47, and the witness reads the typed element tree's per-node `traits`. The SDK's own reachable source grows 199 kB (0.24.5) to 497 kB (0.49.3), and it newly pulls `yauzl` (app-archive), `ignore`, and `@limrun/xdelta3-wasm` (folder-sync; its base64 wasm is 51 kB of the shipped delta).

Checked and rejected as smaller:

- Pinning 0.47.0: reachable source 387 kB, but `^0.47.0` floats back up to 0.49.x and 0.47 predates the typed tree this reads.
- Importing `@limrun/api/client` and `@limrun/api/core/error` instead of the kitchen-sink index: **byte-identical bundle**, so the tunnel/HAR/xcode-cache/rbe/maestro surface was already tree-shaken away.
- Treating `yauzl`/`ignore`/`xdelta3` as dead weight: they sit behind `createInstanceClient().installApp` (`prepareAppBundlePath`, `watchAppArchive`), not beside it.

Sourcemap attribution of the shipped bundle: 2465 kB of third-party source, of which undici 1197 kB, `@limrun/api` 496 kB, yaml 281 kB, ws 129 kB, `@limrun/xdelta3-wasm` 61 kB, pngjs 61 kB, yauzl 51 kB. Nothing in the bundler setup is wasteful for this delta: `minify: true`, no source maps, no test or fixture code in `dist`, and the flat startup diff says the SDK stays behind the lazy provider import.

Two larger levers are outside this PR because they change the published-install shape - the package declares zero `dependencies` today, which is why the SDK is vendored:

- `undici` is 1197 kB of source / 500 kB shipped (~10% of `dist`, ~130 kB download) and is already in the published 0.21.2, bundled only for the SDK's proxy transport.
- the entire Limrun SDK (~1.4 MB shipped) ships to every user although only `connect limrun` needs it.

Both mean a real or optional dependency rather than a dev bundle, so they need a decision rather than a patch. A budget on the Limrun vendor chunk is the cheap middle ground: the Size workflow is report-only, so +77 kB could not fail anything here.


### Update on `210e5e89`

`@limrun/xdelta3-wasm` turned out to be droppable: the SDK loads it through a dynamic import used only by `client.syncApp`, which no agent-device command reaches. Aliasing the specifier to a chunk that names the omission keeps the bundle self-contained and the chunk lazy, and moves the numbers to:

| | `origin/main` | `210e5e89` | diff |
|---|---:|---:|---:|
| unpacked | 4.53 MB | 4.55 MB | +23.1 kB |
| download | 1.34 MB | 1.35 MB | +15.4 kB |
| `CLI --version` startup | 28.1 ms | 27.5 ms | -0.6 ms |

What is left is the SDK's own reachable source: 199 kB at 0.24.5 against 497 kB at 0.49.3, for an option (`requireFocus: false`, new in 0.47) and a typed tree the witness has no other way to read.

thymikee:
Thanks, the size notes explain the +77.2 kB: the fix needs `requireFocus: false` and the typed `traits` tree from the newer `@limrun/api`. The smaller options you tried did not help. The unlabeled-field identity problem from my earlier comment is still open at 11f8561 and still blocks.

thymikee:
## Identity collision: fixed in `4d6dfce0`

Confirmed against the code. `pid` is the app's process id, so two fields that expose neither `AXUniqueId` nor `AXLabel` share every part of the identity, and the twin holding focus satisfied the pre-tap set.

`readLimrunTapTargetIdentities` is now `readLimrunUnambiguousTapTargets`: an identity the pre-tap tree reports more than once is left out of the trusted set, so a twin is witnessed only by geometry. Two tests pin both directions - `a shared identity does not witness the twin that was not tapped` (the misdelivery you described, now refused) and `geometry still witnesses the twin that was tapped` (twins stay fillable).

Re-run live on a Limrun iOS instance after the change: the #2578 Flutter form fills, replaces and clears, each reporting `focused-element`; the React Native form fills, replaces, clears, and fills a second field with the keyboard up via `id=` selectors; aiming `fill` at a static text while a field holds focus is refused with `text_entry_focus_not_observed`.

## `@limrun/xdelta3-wasm`: dropped in `9a420837`

You were right that we do not need it. The SDK loads it through a dynamic import inside `folder-sync.mjs`, and the only caller is `client.syncApp`'s folder-watch path (`syncFolder`, `ios-client.mjs:1163` and `:1173`) - no agent-device command reaches it. The alias in `tsdown.config.ts` resolves the specifier to `src/vendor/limrun-delta-sync-omitted.ts`, which names the omission, so the bundle stays self-contained (the published-closure audit still passes) and the chunk stays lazy: the emitted call site is still `await import("./limrun-delta-sync-omitted.js")`.

Saves 52 kB unpacked and 23 kB download. Reaching that path now fails with a reason naming the package instead of loading a wasm blob nothing ever exercised; restoring the feature is one `deps.onlyBundle` line. Because a new `src/<folder>` must be classified rather than slip past the drift guard, the zone is declared unranked in `210e5e89`.

## Why 0.24.5 cannot host this witness

`0.24.5` declares `elementTree(point?): Promise<string>`, and its node type has no `traits` field - the per-node `IsEditing` is exactly what the witness reads. Decisively, its `typeText(text, pressEnter?)` has no options bag, so the client cannot send the flag that skips the server-side focus scan, which is the scan #2578 dies on. So the version question is not about size: 0.24.5 cannot express the fix. `0.47.0` is where `requireFocus` arrives and it already pulls `yauzl` plus `@limrun/xdelta3-wasm`, so no version buys the fix cheaper; with xdelta3 aliased out, the remaining growth is the SDK's own reachable source (199 kB at 0.24.5, 497 kB at 0.49.3).

thymikee:
At ba66380 the twin case I reported is fixed. An identity the pre-tap tree reports more than once no longer witnesses focus, and `a shared identity does not witness the twin that was not tapped` would fail on 11f8561. The lazy `import()` of the xdelta3 stand-in also looks right.

One gap remains in the same check. [`readLimrunUnambiguousTapTargets`](https://github.com/callstack/agent-device/blob/ba66380eaf7989dd8e324f62bdba95297d8a1aca/packages/provider-limrun/src/text-entry-focus.ts#L71-L93) counts duplicates only before the tap. Say the tapped field has no identifier or label and is the only one of its type. The tap does not focus it, but it shows a second unlabeled field of the same type elsewhere (a new row, a lazily built item), and that field takes focus. Its identity is in the trusted set, so the fill types into it and reports `focused-element`. Please trust an identity match only when the post-tap read also reports that identity once, and add a test with one such field before the tap and two after.

thymikee:
## Second gap closed in `03e227d4`

Confirmed: duplicates were counted only before the tap. The fix moves the rule into the type instead of adding a second check at the call site - `LimrunTextEntryFocus.identity` is now optional, and `readLimrunTextEntryFocus` populates it only when that read reports the identity once. A shared identity can no longer reach the trusted set from either side of the tap, leaving geometry as the only thing that can clear such a fill.

`a field that appears after the tap does not borrow the identity it now shares` is your case: one unlabeled field before the tap, two after with the newcomer editing. It fails on `ba66380` - verified by removing the new guard and watching it go red - and passes now.

Both readers now share one framed-node list and one `countIdentitySightings`, so pre-tap and post-tap uniqueness are the same rule by construction rather than two copies of it. `readLimrunTextEntryFocus` still returns an element whose identity is shared, just without an identity, so `editingElementObserved: true` keeps separating "focused the wrong field" from "nothing is editing".

Re-run live on a fresh Limrun iOS instance at this head: the #2578 Flutter form fills, replaces and clears at `focused-element`; the React Native form fills, replaces, fills a second field with the keyboard up through `id=` selectors, clears, and both forms refuse a `fill` aimed at static text while a field holds focus. 130 provider unit tests and 210 provider-integration tests pass, as does `check:layering` and `check:fallow`.

thymikee:
Reviewed at 03e227d. An identity now witnesses focus only when both the pre-tap and the post-tap reads report it once, through one shared counting helper, and the new test for a field that appears after the tap would fail on ba66380. That closes the gap from the last review, and I found no new problem in the delta.

The Limrun iOS live run in the PR body was on 11f8561. Commits 4d6dfce and 03e227d narrow which fills pass on that same route, so an unlabeled Flutter field with a twin now has to clear by geometry. Could you rerun the #2578 Flutter fill, replace and clear on 03e227d and confirm that they still report `focused-element`?

Smoke Tests failed in the iOS Settings replay on a local simulator (step 3: the runner was already restarted and the snapshot still failed). This PR changes only the Limrun provider, a type-only contract, the bundle config and a vendor stub, none of which are on that route, so the failure is likely unrelated; a rerun should confirm it. There are no conflicts.

thymikee:
## Flutter rerun at `03e227d`

`pnpm build` at this head, fresh state directory (so a fresh daemon), fresh Limrun iOS instance, the #2578 form:

| case | result |
|---|---|
| `fill @e3 "live@example.com"` | `textEntryReadiness: "focused-element"`, tree reads `"live@example.com"` |
| `fill @e3 "replaced@corp.dev"` | `focused-element`, tree reads the replacement |
| `fill @e3 ""` | `focused-element`, `Filled 0 chars`, placeholder back |
| `fill` on static text while the field holds focus | refused: `text_entry_focus_not_observed`, `editingElementObserved: true` |

The narrowing costs the shipped form nothing: its fields expose labels, so identity still witnesses them.

## The route the narrowing actually changed, run live too

That app has no twins, so it does not exercise the geometry-only path. I rebuilt the repro form with two extra `TextFormField`s carrying no label and no hint - the tree reports both as bare `[text-field]` nodes sharing one identity - and filled them on the same head:

| case | result |
|---|---|
| twin #1 | `focused-element` |
| twin #2, keyboard already up | `focused-element` |
| labeled email, cross-field from a twin | `focused-element` |
| twin #1 cleared | `focused-element`, `Filled 0 chars` |

Each twin was witnessed by geometry and the text stayed where it was aimed: after clearing twin #1 the tree shows it empty while twin #2 still holds its value. So the guard refuses the misdelivery without refusing a legitimate fill of an unlabeled field.

## Smoke Tests

Agreed it was off this PR's route - I reran the failed job and it is green now. Every check on `03e227d` is green: 20 passing, none failing.

thymikee:
Thanks for the rerun at 03e227d. The #2578 Flutter fill, replace and clear still report `focused-element`, and the extra run with two unlabeled twin fields shows the geometry-only route this head narrowed also works, with each value staying in the field it was aimed at. That covers the evidence gap from the last review. All checks on 03e227d pass, and there are no conflicts.

PR #2593, opened 2026-09-14
refactor(daemon): resolve device interactors through a composed capability

### LINKED ISSUE #2555: Resolve device interactors behind a composed capability

thymikee:
## Summary

`src/daemon/snapshot-interactor-capture.ts` reached interactor mechanics through `await import('../core/interactors.ts')` — the one production edge leaving the daemon zone dynamically, invisible to the ranked spine and recorded as `leaked-platform-mechanics` by R76.

Interactor resolution now arrives as a capability. `src/daemon/interactor-resolution.ts` holds the daemon seam, and `startDaemonRuntime` installs it beside `installProviderDeviceAdmission`, so one composition site names both request-scoped capabilities (ADR 0019 §1/2). An un-composed process fails closed with `reason: interactor-resolution-missing` instead of reaching for a lookup of its own.

R76 swaps the dynamic leaked edge for the composition edge, so no classified leaked-mechanics edge is reached dynamically, and `src/core/interactors.ts` keeps exactly one inbound daemon-zone edge — from composition. 10 files, +208/−38.

Closes #2555

## Validation

Head `771de8f510`: `pnpm check:affected --run` is green except `mutation-model`, which fails identically at base `730b20149e` (`pnpm mutation:test`: 39 pass / 1 fail on both) — pre-existing, unrelated to this diff.

Also green locally: full `unit-core` (10,024 tests), `pnpm test:integration:provider` (210 tests), `pnpm check:tooling` (format, lint, typecheck, layering, di-seams, build, package), `pnpm check:fallow`.

The fail-closed default earned its keep: `interaction-settle-private-ax-route.test.ts` and the provider-scenario `harness.ts` had to compose the capability the way the process root does, exactly as they already do for the provider-device admission. `src/daemon/server/daemon-runtime-interactor-composition.test.ts` boots the real runtime and resolves through the seam, so the root install line is covered: deleting it fails that test with `interactor-resolution-missing`. No device-facing behavior changed, so no device run is claimed.


thymikee:
Looks good at fdcb981. `startDaemonRuntime` installs the interactor resolution before it builds the request handler, and it is the only production entry that builds one, so the fail-closed default cannot trip in a real daemon. The R76 change removes the dynamic edge instead of relabeling it.

One small gap, not a blocker: the harness installs its own copy, so no test fails if the install in `daemon-runtime.ts` is removed. Smoke Tests is the only check that covers that line today. A short test that `startDaemonRuntime` installs it would close that gap.

thymikee:
Good catch — added `src/daemon/server/daemon-runtime-interactor-composition.test.ts` (771de8f5): it boots the real `startDaemonRuntime` with the usual runtime/provider mocks, then resolves a provider-owned device through `interactorResolution()` and asserts the provider interactor comes back. Verified it kills the mutant: removing the install line from `daemon-runtime.ts` fails it with `reason: interactor-resolution-missing`. `check:affected --run` re-run at that head is green up to the pre-existing `mutation-model` failure (identical at base).

thymikee:
Still good at 771de8f. The new test boots the real `startDaemonRuntime` and resolves through `interactorResolution()`, and the un-composed default throws `interactor-resolution-missing`, so the test fails if the install in `daemon-runtime.ts` is removed. That closes the gap from my last comment. Checks are green and there are no conflicts.

PR #2594, opened 2026-09-14
refactor(daemon): stop the client value-importing daemon internals (#2559)

### LINKED ISSUE #2559: Declare the daemon entry surface and stop daemon-client importing daemon internals
Parent: #2542. The classifier half landed in #2557; this is the entry half.

## Problem

`src/daemon-client/**` reaches into daemon internals on the value side, and nothing declares which
root module is allowed to import the daemon zone in the first place. Measured at the #2557 tip: 13
`daemon-client -> src/daemon/**` pairs, of which 8 are value edges.

| target | LOC | inbound files | what the client needs |
| --- | --- | --- | --- |
| `src/daemon/config.ts` | 113 | 42 | state dir / port / token resolution |
| `src/daemon/daemon-process.ts` | 122 | 4 | pidfile + liveness discovery |
| `src/daemon/session-repair-tombstone.ts` | 91 | 2 | tombstone path + read |
| `src/daemon/request-progress-protocol.ts` | 48 | 3 | progress event framing |
| `src/daemon/daemon-request.ts` | — | — | 4 type-only request/response shapes |

Three of the four value targets have tiny inbound sets and no daemon-only dependencies, so they are
shared on-disk/protocol contract rather than daemon internals. `config.ts` is the odd one: 42 inbound
files, so the client's need has to be cut down to the subset it actually reads instead of moving the
whole module. The `daemon-request.ts` type edges belong to the wire half that #2318/#2322 already split
into `daemon-request-wire.ts`.

## Desired behaviour

- One declared entry surface names which modules may import the daemon zone, and it is the process
  root composition — not `daemon-client`.
- The four value targets above are reached through a location that is not daemon internals: relocated
  where the inbound set is small, or narrowed to the subset the client reads for `config.ts`.
- The residue is ratchet-declared shrink-only, and every entry in it names a measured edge, not a
  directory.

## Acceptance

- `daemon-client -> src/daemon/**` value edges: 0.
- Type-only `daemon-request.ts` edges resolved against `daemon-request-wire.ts` or recorded with a
  rationale for keeping them.
- A planted `daemon-client` import of a daemon-internal module is red, and a stale declaration entry is
  red.

thymikee:
## Summary

`src/daemon-client/**` composed the daemon over the network but still **value-imported** its module graph — `config.ts` (state-dir), `daemon-process.ts` (pidfile liveness), `session-repair-tombstone.ts`, and `request-progress-protocol.ts`. This moves those process-root contracts out from under `src/daemon/` so both sides reach them at the composition root and repoints the client at the relocated modules. The client keeps reading the canonical `DaemonRequest`/`DaemonResponse` vocabulary from `daemon-request.ts` as a **type-only** import.

Adds gate **R78 `daemon-client-entry`**: a runtime (value/dynamic) import of a daemon module by the client is red, an unrecorded type import is red, and a recorded edge the tree no longer imports is stale — so the residue (five type-only `daemon-request.ts` reads) can only shrink. `daemon-modularity` now skips `src/daemon-client/` so the two gates never own one edge.

Review follow-ups: `src/daemon/config.ts` no longer re-exports the shared leaf and every importer reads `src/daemon-resolution.ts` directly (the re-export grew `src/cli.ts`'s eager closure 295→296 against ADR 0019), and the mutation lane's subprocess exclusion follows `daemon-process-takeover.test.ts` to `src/__tests__/`.

62 files touched. Production sources (`src` + `scripts`, excluding tests and fixtures, rename-aware): **+194 / −389**, net **−195**. No behavior change; renames are content-preserving. Part of #2542. Closes #2559.

## Validation

Head `3e7bd6f15a`, merge-base `730b20149e`. Green: `eager-closure-budgets` (610, `src/cli.ts` back to 295), `check:layering` (incl. R78 planted + stale proofs), `check:fallow` (0 issues in changed files, 0 stale), `check:daemon-wire-compat` + wire-compat vitest (protocol unchanged, 8 moved, 0 changed — digests identical), full `test:unit` (1299 files / 10033 tests), `typecheck`, `lint`, `build`. The CI Size report shows +320 B unpacked and slightly faster startup.

`pnpm check:affected --run` fails on two steps, both external to this change: `lint` emits no findings and aborts under the parallel fan-out (`pnpm lint` is exit 0), and `mutation-model` ("a kernel is owned by tests that reach it indirectly") is already red at merge-base `main` — a capture-kit `scroll-edge-state` reachability issue unrelated here. No rebase: the only file upstream also touched is `check.ts`, in a comment block. A serialized CI run is authoritative.


thymikee:
The moves at 5a350b4 keep content and wire digests the same, and R78 covers every client-to-daemon edge that `daemon-modularity` used to check. Two things need a fix before merge.

Coverage fails because of this change. `src/daemon/config.ts` now re-exports from the new `src/daemon-resolution.ts`, and `src/cli.ts:41` still imports `resolveDaemonPaths` from `daemon/config.ts`, so CLI startup loads one more module (296 vs 295, see `eager-closure-budgets.test.ts` in the [Coverage job](https://github.com/callstack/agent-device/actions/runs/34853452803/job/104006901577)). Point the importers at `daemon-resolution.ts` directly and drop the re-export, so `config.ts` keeps only the daemon's own rules.

`vitest.config.ts:14` still lists `src/daemon/__tests__/daemon-process-takeover.test.ts` in `MUTATION_EXCLUDED_SUBPROCESS_TESTS`. The test moved to `src/__tests__/`, so the mutation lane runs its real per-case daemon spawn again (#1545). Please update the path.

Also, the body says net −336 production lines, but the diff is about +294/−129 without tests and docs. Please correct it after the fix.

thymikee:
All three addressed at `3e7bd6f15a`.

- **Eager closure** — `ebb1ea0ac6` drops the re-export and points the CLI, client, daemon, scripts, and test importers at `src/daemon-resolution.ts` directly, so `src/daemon/config.ts` keeps only `resolveSessionIsolationMode`/`normalizeTenantId`. `eager-closure-budgets` is green again (610 passed; the `src/cli.ts` case is back at 295 against the merge-base).
- **Mutation lane** — `3e7bd6f15a` moves the `MUTATION_EXCLUDED_SUBPROCESS_TESTS` entry to `src/__tests__/daemon-process-takeover.test.ts`, so its per-case daemon spawn stays excluded (#1545).
- **Size figure** — body corrected. Production sources (`src` + `scripts`, excluding tests and fixtures, rename-aware) are **+194 / −389**, net **−195** across 28 files; 62 files touched overall.

Full `test:unit` (1299 files / 10033 tests), `check:layering`, `check:fallow`, wire-compat (protocol unchanged, 8 moved, digests identical), `typecheck`, `lint`, and `build` are green on the new head. Not rebased: the only file upstream also changed is `scripts/layering/check.ts`, in a comment block that does not overlap these hunks.

thymikee:
All three points are fixed at 3e7bd6f. `src/daemon/config.ts` no longer re-exports anything, no importer reaches `resolveDaemonPaths` or `resolveDaemonServerMode` through it, and the mutation exclusion now names `src/__tests__/daemon-process-takeover.test.ts`.

One small thing: the production figure in the body still does not match the diff. `git diff -M --numstat 730b201 3e7bd6f` without tests, fixtures and docs gives +303/−144 (net +159), not −195. Please correct the number. Smoke Tests were still running, and there are no conflicts.

PR #2595, opened 2026-09-14
refactor(shell-quote): one implementation, reached through the runner host port

### LINKED ISSUE #2525: refactor: one shellQuote implementation instead of three
## Duplication

Three byte-identical `shellQuote` bodies (verified at `eefe37b51e`):

- `packages/host-kit/src/internal/shell-quote.ts` (canonical; exported via `@agent-device/host-kit/command`, `command.ts:25`)
- `packages/platform-android/src/perf-native-artifacts.ts:124` (exported; used at `:143`, `:208`; imported by `perf-native-simpleperf.ts:8`)
- `packages/platform-apple/src/runner/runner-lease.ts:292` (used by `formatEnvAssignment`, a human hint string)

`platform-android` already imports `shellQuoteIfNeeded` from `@agent-device/host-kit/command` in three sibling files, so the Android copy is a direct import swap. R77 (`scripts/layering/apple-runner-host-port-policy.ts`) forbids `runner/**` from value-importing host-kit, so the Apple copy goes through the `AppleRunnerHost` port (`runner/host.ts`, bound in `core/runner-host.ts`). Do not take the "runner-local const" shortcut.

Note: the adb-shell injection gap from the 2026-08-06 audit is already closed (`057f0e6eb1e`, `6dd7d406c04`); this issue is duplication only.

## Done when

`grep -rn "replaceAll(\"'\"" packages src` finds one implementation. Unit tests only; no device run needed.

thymikee:
## Summary

Closes #2525. Two byte-identical `shellQuote` bodies went away, leaving
`packages/host-kit/src/internal/shell-quote.ts` as the only implementation. 5 files, no behavior
change and no new module.

- `platform-android/perf-native-artifacts.ts`: the exported copy is deleted and both call sites,
  plus `perf-native-simpleperf.ts`, import `shellQuote` from `@agent-device/host-kit/command` —
  the subpath three sibling Android modules already use for `shellQuoteIfNeeded`.
- `runner/runner-lease.ts`: goes through the `AppleRunnerHost` port (`runner/host.ts`, bound in
  `core/runner-host.ts`), not a runner-local const, because R77 (`apple-runner-host-port`) forbids a
  value import of host-kit from `runner/**` and `runner/` sits in the eager closure of seven Apple
  façades.

`grep -rn "replaceAll(\"'\"" packages src` now returns the canonical implementation plus
`packages/xml/src/internal/entities.ts`, which escapes `&apos;` for XML and was never a shell quoter.

## Validation

Tested at `37b1eca498`. `pnpm check:affected --run` passed (format, lint, typecheck, layering,
fallow, 420 related test files / 2,899 tests), so no device run applies. `runner-session.test.ts`
already pins the quoted hint text `AGENT_DEVICE_STATE_DIR='/tmp/agent-device-owner'` for the Apple
path, and it passes unchanged; `shell-quote.test.ts` pins the canonical output. GitHub stays
authoritative for coverage and provider integration.

First gate run failed on `runner-client.test.ts :: ensureXctestrunArtifact aborts only the
disconnected request build` (5.0s timeout). Per `docs/agents/testing.md` it was host contention:
that test passes alone in 113ms at the same commit and the full rerun above is green.


thymikee:
Looks good at 37b1eca. Both deleted copies match `host-kit`'s `shellQuote` exactly, `AppleRunnerHost` has one implementer and it binds the new member, and nothing still imports the removed Android export.

thymikee:
The Smoke Tests failure at 37b1eca is the native runner test `testAlertAcceptDoesNotActivateAReplacementWithASharedButton` ([job](https://github.com/callstack/agent-device/actions/runs/34853833362/job/104008178200)). This change does not touch the runner or alert handling, so it looks unrelated; a rerun should confirm.

thymikee:
Rerun confirms it. The failed iOS job on the same head `37b1eca` is green: [Smoke Tests](https://github.com/callstack/agent-device/actions/runs/34853833362/job/104058486518) passed, `testAlertAcceptDoesNotActivateAReplacementWithASharedButton` among them, and every other Smoke Tests lane is green too. Nothing was changed for this.

Worth noting for whoever owns the runner suite: `#2600`, a TypeScript-only change, failed Smoke Tests in the same file one test over — `testAlertCannotProveAnIdenticalReplacementAndDoesNotActivateIt` at `RunnerTests+AlertObservationTests.swift:75`, twice with `("First actions: 0; replacement actions: 0") is not equal to ("First actions: 1; replacement actions: 0")`. Two unrelated branches landing in adjacent tests of one alert-observation file, both reporting an action log that recorded nothing, reads like the fixture's own observation window rather than either change. Happy to file it as a flake issue if nobody else is on it.

PR #2596, opened 2026-09-14
perf(scroll-until): answer every candidate from one visibility index

### LINKED ISSUE #2528: perf(scroll-until): build the snapshot visibility index once per pass
## Waste

`src/daemon/scroll-until.ts:169` evaluates `.some((node) => evaluateIsPredicate(...))` per pass; `predicates.ts:77` builds a fresh `createSnapshotVisibility(nodes)` per candidate. The rule in `packages/contracts/src/snapshot-visibility.ts:135-141` (#1970) says N candidates share one index. `resolveSelectorPipeline`, called just above, builds an equivalent index and discards it (`resolve.ts:229`). N is the matched selector set (1-5), so this is not a measurable wall-clock win; it is a correctness-of-shape fix. Added in #2436.

## Fix

Hoist the index out of the `.some()` (one per pass). Thread the pipeline's index into the predicate if the seam is already there; otherwise leave that for the H2 capture-kit move. Three lines; no measurement gate, no perf claim.

thymikee:
## Summary

`evaluateIsPredicate` derived its visibility index from the `nodes` it was handed, so one `scroll --until` pass rebuilt the node map and viewport-rect scan once per candidate its selector matched. It now takes the capture's `SnapshotVisibility` and holds no tree of its own: the caller that owns the capture owns the index, which is #1970's rule that N candidates share one whole-tree pass. The `--until` pass builds one index and asks every candidate of the selector against it.

No behavior change for `is <predicate> <selector>` or `scroll <dir> --until <selector>`. `@agent-device/selectors` is private, so no CLI, MCP, help or wire surface moves. 6 files.

Closes #2528, correcting its premise: the `wait` row is `first-match`, so it resolves through the existence-only scan and never builds the index the issue pointed at. Nothing was left to thread on this path; `is` (`readUnique`) is where two indexes meet.

## Validation

`pnpm check:affected --run` passes at `bd319ae7`: format, lint, typecheck, layering, `check:fallow --base origin/main`, build, and 568 related test files / 4129 tests including provider integration. The first run failed only in `android-lifecycle.test.ts` frame-action correlation; that file passes alone and on the rerun, at load average 40+ on 12 cores.

Evidence: the probe test failed on unmodified `predicates.ts` (`TypeError … reading 'map'`); the two twin tests were green on unmodified code first; and `a pass materializes one visibility index for the candidates it asks` counts index builds through the contracts facade — 1 for a selector matching two rows, 2 and red when that build moves back inside the candidate loop.

Device lanes and Coverage stay GitHub-authoritative for this diff.


thymikee:
The shared index gives the same verdicts at 3dbf9dc. `createSnapshotVisibility` depends only on the nodes of the current pass, `decideUntilPass` builds it from each new capture, and `evaluateIsPredicate` has no callers outside this repo.

The tests do not lock in the change, though. If `createSnapshotVisibility(nodes)` in `src/daemon/scroll-until.ts` moves back inside the candidate loop, every test still passes: the twin tests check only verdicts, and the probe test in `predicates.test.ts` failed on the old code only because of the signature. Please add a scroll-until test with two matching candidates that counts index builds in one pass and expects one, or say in the PR that this shape does not need a test.

The failing Smoke Tests job is likely unrelated. The failure is the native runner test `testAlertCannotProveAnIdenticalReplacementAndDoesNotActivateIt`, this diff does not touch the Apple runner, and a sibling alert test failed the same way on #2592 today. There are no conflicts.

thymikee:
Pushed `bd319ae7` with the counting test, since the shape is what the child is about and prose would have rotted.

`a pass materializes one visibility index for the candidates it asks` wraps `createSnapshotVisibility` on the contracts facade (the `vi.hoisted` + `importOriginal` idiom `platform-runtime-app-log-process.test.ts` uses), runs one pass over `twinTree(-400, 200)` — two candidates, verdict visible on the second — and expects one build. Red proof: moving `createSnapshotVisibility` back inside the `.some()` makes it read 2 and fail; the twin verdict tests stay green either way, exactly as you found. 1 is also the true baseline: `wait` is `first-match` and `occlusion: 'ignore'`, and the pass returns before the edge analyzer, so nothing else indexes that tree.

Thanks for the alert-runner triage: agrees with the log. `RunnerTests+AlertObservationTests.swift:75` fails on `First actions: 0; replacement actions: 0`, one of four alert tests in that job, and no Swift or Apple path is in this diff.

`pnpm check:affected --run` green at `bd319ae7` — format, lint, typecheck, layering, `check:fallow --base origin/main`, build, 568 related files / 4129 tests.

thymikee:
The counting test at bd319ae locks in the change. It wraps the same `@agent-device/contracts/snapshot` import that `scroll-until.ts` uses, and `twinTree(-400, 200)` makes `.some()` reach the second candidate, so a build inside the loop would count 2 and fail. No code findings remain, CI is green, and there are no conflicts.

Issue #2589, opened 2026-09-14
Treat interaction targets behind the visible keyboard as occluded/off-screen (parity with scroll)


thymikee:
## Purpose

When the software keyboard is up, elements it visually covers still report `hittable: true` with an on-screen `rect`, so `press`/`click`/`fill`/`longpress` tap *through* the keyboard and silently activate whatever the keyboard's own hit-testing routes the point to (or nothing). Keyboard occlusion is a normal, expected iOS/Android pattern — apps are not going to move their tab bar above the IME — so the fix belongs in agent-device: an interaction target whose tap point falls under the visible keyboard should be treated as occluded/off-screen and refused with a typed reason and a recovery hint, exactly the way `scroll` already does.

This came out of dogfooding MaintainX on a physical iPhone (`thymikee-iphone`, v5.39-311017).

## Context / evidence

On the Messages tab of MaintainX, focusing the search field raised the keyboard over the bottom tab bar. The tab-bar nodes stayed in the app snapshot with their layout geometry and were still `hittable`:

```
Button "More"     center=(340,822)  rect={"x":303,"y":795,"width":74,"height":54}
Button "Messages" center=(274,822)  ...
```

The keyboard occupies roughly `y≈540..874`, so the More tab is fully behind it. `press 340 822` (and `press @more`) did **not** navigate; the touch landed on the keyboard. The AX overlay below shows the tab-bar refs drawn *under* the keys:

The annotated overlay (tab-bar refs `@e40`–`@e46` drawn *under* the keys) and a reproduction video are in the dogfood evidence bundle (`bug3-keyboard-covers-tabbar.png`, `bug3-keyboard-tabbar.mp4`); they need a manual web upload to attach here.

(Reproduction video: local `bug3-keyboard-tabbar.mp4` — More tab works with no keyboard, then is dead once the search field is focused. Same class of silent misfire on Android, whose IME is also a separate window.)

## Why the current guards miss it

- `occlusion` (ADR 0011) refuses targets "covered by another visible element" — but occlusion is computed from same-window drawing order in the snapshot tree. The keyboard is a **separate window** (a system surface), so it never appears as a covering sibling of app elements.
- `offscreen` refuses targets whose tap point is outside the **root Application/Window viewport** (`isTapPointInsideViewport` / `isVisibleOnScreen` in `packages/contracts/src/snapshot-visibility.ts`). The covered tab bar is still inside the app window rect, so it passes.

[...]

PR #2601, opened 2026-09-14
feat(contracts): name the keyboard band that owns a tap point

thymikee:
## Summary

One new vocabulary module for the interaction guarantee contract: `@agent-device/contracts/tap-keyboard-occlusion` names the band a visible software keyboard owns over a captured tree, and answers whether a tap point belongs to that band, to a keyboard control, or to app content behind it.

Nothing said this today. A software keyboard is its own system surface, so it is never a covering sibling of app content (`occlusion` cannot see it) and it leaves the app window rect intact (`offscreen` passes a covered tab bar) — the tap that reported success in #2589 was not covered by any existing guarantee.

Three rules decide whether reported geometry may be measured at all. Docking: the keyboard's own rects must reach within 80 units of the viewport's bottom edge (measured: iPhone 17 Pro portrait stops 58 pt short, iPad Pro 11-inch 3 pt short, Gboard lands on the display bottom with a 126 px three-button bar inside its reported region). Width: geometry taller than it is wide arrived in the keyboard's own rotated space (iOS reports its landscape keyboard that way) and is refused. Tiling: the keys must form one unbroken run of columns, because a split iPad keyboard docks both clusters on the bottom edge while its container still spans the screen and the app content between the clusters is live. Everything unmeasurable — floating, split, rotated, sparse — reads `undetermined` and fails open, the same stance the scroll keyboard policy takes: a keyboard the platform cannot measure is not evidence that a surface is blocked. Ownership reads accessibility type and input-method provenance, never a locale-dependent label.

No behavior change lands here — the one deletion is a retired key in an Android test fixture, repairing `main`'s red `typecheck` (Design, last paragraph). The acting paths adopt the decision in the stacked `fix/tap-keyboard-occlusion-tap`. 6 files, +4,452/-1, of which the golden table is 3,997 across 35 cases; the production module is 339 lines.

## Design

Asked for, on record, since this module's first consumer is #2602.

**Inline in the guard.** The decision has three readers across two zones — the refusal, the coordinate disclosure, and the golden table — and `R2 commands-floor` stops `daemon/` reading `commands/`, which #2622 hit for real when a four-line `find` guard failed the layering scan. Inlining also orphans the table: `AGENTS.md` keeps a golden-table-backed rule with its table under `contracts/fixtures/`, and the matrix `via:` cells need one shared symbol to name.

**Make it a covering-surface kind in the occlusion classifier.** Smallest diff available and structurally wrong: that classifier is same-window and ancestry-based and answers with a rect, while the keyboard is a separate surface whose verdict is a band running to the viewport bottom. It has no outcome for geometry that arrived rotated, floating, split or sparse, and the coordinate path needs a warning where the classifier only knows refusal.

**Extend `ios-system-surface.ts`.** The keyboard is a system surface, so this is the best candidate and it still does not fit: that module is a closed bundle-id registry keyed by process, with the invariant "activation destroys the session". This rule is keyed by accessibility type and IME provenance inside one captured tree, and has no activation semantics to join.

**Reuse `snapshot-visibility.ts`, or fold into `interaction-guarantees.ts`.** The first answers "inside the viewport the app can act in"; the keyboard is invisible to every app-window rule by construction, which is the misfire. The second declares which rules a path commits to, not what a rule decides — and either way a new rule would land as a branch inside a module half the interaction stack already imports.

**Fewer outcomes.** `no-keyboard` and `undetermined` both fail open and must stay distinguishable from `clear`: "this tree has no keyboard" and "this keyboard's geometry cannot be measured" are different answers, and collapsing them is what left #2589's success-looking tap with nothing to say about which tree it read.

**Land it inside #2602.** That PR is the guard, four matrix cells and the device lane. Split apart, the rule can be reviewed with no behavior change in the diff.

The place this could still get smaller is enforcement, not the decision: once the node-stage contract can carry an action, a label and a resolved tap point, the guard becomes one stage declared in the pipeline policy instead of four call sites, and `find` gets its cell for free. That is #2622's shape, and it shrinks the callers without shrinking this module. The one cosmetic cut available now is folding `TAP_KEYBOARD_OCCLUDES_TARGET_REASON` into the details const, five exports to four.

For scale, the zone already carries fixture-backed modules at 69, 88 and 269 lines (`snapshot-scope.ts`, `ios-system-surface.ts`, `snapshot-visibility.ts`); this is 339 lines with five exports, 136 of them prose.

## Validation

Tested at `dafab14379`, rebased on `origin/main` `2cafab3ad0`.

- `pnpm check:affected --run` — `format`, `lint`, `typecheck`, `layering`, `di-seams`, `fallow`, `mcp-metadata`, `build`, `package`, `integration-node` and `macos-coverage` passed; `vitest-related` then timed out one Apple runner test (`ensureXctestrunArtifact aborts only the disconnected request build…`, 5s budget) at 82s wall across 330 files, and passed unchanged on rerun: 330 files / 2,380 tests, and 50/50 in 920 ms for that file alone. `mutation-model` did not get to run because the gate stops at the first failure; its ownership assertion fails identically on `origin/main`.
- `pnpm typecheck` — clean. The `installArgs` breakage is repaired in this stack: `test(android): drop the retired installArgs from the install-timeout fixture` deletes one dead key from a typed literal in `packages/platform-android/src/__tests__/snapshot-helper-install.test.ts`, a field #2619 retired and the parser already ignores (`snapshot-helper.test.ts` pins that), and nothing read it. That un-reds `Typecheck & Package` here and on `main`. Happy to lift that commit into a standalone fix for `main` and drop it from this PR if you would rather it merge first.
- Stages run individually: `lint`, `check:layering`, `check:di-seams`, `check:fallow --base origin/main`, `build`, `check:package`, `check:command-docs`, `check:agent-guidance`, `test:integration:node`, `test:integration:macos-coverage` — clean. `check:production-exports` reports the same 68 findings here and on detached `origin/main` in this worktree (local fallow drift; CI's copy is green). `mutation-model`'s ownership assertion fails identically on `origin/main`.
- `npx vitest run packages/contracts` — 61 files / 282 tests green. Every case reaching a band declares `expected.frame` and `expected.controlRects`.
- Geometry captured on iPhone 17 Pro (iOS 26.2), iPad Pro 11-inch (M4, iOS 26.2) and Pixel 7 (Android 16, real Gboard); key rows adjoin across 395 pt, 743.5 pt and 1070 px with no seam above a point. The split, floating and sparse-emoji cases are modelled and say so in their names — the emoji row's tiling behaviour stays unverified on device, which costs coverage, not correctness.

Device evidence for the guard belongs to the stacked PR, where it runs.

The `installArgs` repair is a one-line test-fixture deletion with no production reader, so it is deliberately its own commit and trivially liftable: cherry-pick `test(android): drop the retired installArgs from the install-timeout fixture` to `main` and this PR drops it.


thymikee:
At 128cab9 the contract is well scoped: #2602 is its only consumer, the Android side reuses `isAndroidInputMethodNode`, and most of the size is the golden table. Two keyboard layouts get the wrong answer.

iPhone landscape can turn the guard off. [`resolveVisibleKeyboardSurface`](https://github.com/callstack/agent-device/blob/128cab9fb54948716b9887e17e6fc2cbf917fe80/packages/contracts/src/tap-keyboard-occlusion.ts#L170) returns null when the top of the keys is above half the viewport height. A landscape iPhone is about 402 pt tall, and the keyboard with its suggestion bar can start near that line, so the result is `undetermined` and the #2589 misfire still happens. Could the rule require the band to reach the bottom edge instead? A landscape case in the table would pin it.

A floating or split iPad keyboard gives false refusals. The band always runs from the top of the keys to the bottom edge and across the full width of the keys ([line 175](https://github.com/callstack/agent-device/blob/128cab9fb54948716b9887e17e6fc2cbf917fe80/packages/contracts/src/tap-keyboard-occlusion.ts#L175)). App controls below a floating keyboard, or between the two halves of a split one, then read as occluded, and #2602 refuses taps on them. Building the band only from keys that reach the bottom, and failing open otherwise, would avoid it. Please add both layouts to the table.

Two smaller things. `resolveVisibleKeyboardSurface` is exported, but only its own test calls it and the table already covers its null cases, so it can be module-private. An Android keyboard that is not in the known package list reads as `no-keyboard` rather than `undetermined`, so the "could not check" disclosure never shows; passing the active input-method package when the capture has it would help.

A smaller-design question: could `resolveKeyboardTapOcclusion` live next to `isTapPointInsideViewport` in `snapshot-visibility.ts`, which already owns the tap-point visibility rule, instead of adding a new contracts subpath?

thymikee:
Both layouts were wrong, and landscape had a second problem underneath. Rebased onto `main`-tip-free stack and force-pushed as `b7fdd71a10`, `6e13b1ccd7`, `342248ff0a` over `128cab9`.

Measured first, on iPhone 17 Pro (iOS 26.2) with Safari raising the system keyboard, because both findings are claims about geometry:

| orientation | reported keyboard geometry | viewport | distance from reported bottom to viewport bottom |
| --- | --- | --- | --- |
| portrait | key plane 402 x 233 at y 583..816 | 402 x 874 | 58 pt |
| landscape | key plane 162 x 327 at x 2..164, y 75..402 | 874 x 402 | 0 |

**Landscape (finding 1).** Confirmed, and worse than predicted. The landscape key plane starts at y 75 of a 402 pt viewport, so the midline rule declined to measure it — but the geometry itself is unusable: iOS reports the landscape keyboard's rects in the keyboard's own rotated space. The screenshot shows the keyboard full width across the bottom 327 pt, while the tree says the key plane is 162 pt wide, QWERTY advances down y at fixed x, successive rows step x down instead of y up, and the globe dock button sits at y 8. A band from that refuses app content near the top-left that the keyboard is nowhere near *and* lets a tab-bar item at (264, 360) tap straight through into a key, which I watched happen before the rule changed. Filed as #2612 (https://github.com/callstack/agent-device/issues/2612) for the capture-side fix.

Your suggested rule is in, with the measurement that makes it work: the midline heuristic is gone, and eligibility is two geometry questions against the reported rects — keyboard geometry must come within `KEYBOARD_BOTTOM_ANCHOR_TOLERANCE` (80 pt; the measured portrait gap is 58) of the viewport's bottom edge, and it must be wider than it is tall, because a docked keyboard is 402 x 233 in portrait and 874 x 327 in landscape. Rotated geometry gets no measurement and the tap fails open, the same treatment an unresolvable viewport already gets. Height-as-fraction cannot express this, which is what the table now says out loud: the same keyboard is 27 % of the portrait viewport's height and 81 % of the landscape one.

Landscape is pinned twice: the measured rotated tree as `undetermined`, and a wide docked landscape keyboard measured against the landscape viewport's own bottom edge as `occluded`.

**iPad floating and split (finding 2).** The docking budget is the fix you described: geometry that stops more than 80 pt above the bottom edge is not docked, its band stays unmeasured, and #2602 has nothing to refuse with. Both layouts are in the table as `undetermined`. One thing I should not let look stronger than it is: those two trees are modeled from the layout's gap below the keys, not captured, because there is no iPad simulator in this environment. The iPhone numbers above are device measurements; if you want the iPad cases captured too, say so and I will spend the device time.

**`resolveVisibleKeyboardSurface`** is module-private now. The derivation assertions that only its own test could see moved into the table — `expected.frame` pins the band — and a zero-area viewport case joined the null cases.

**Android package unknown reading `no-keyboard`.** The capture does not carry the fact you are asking me to pass. `activeInputMethodPackage` lives in `packages/platform-android/src/fill-verification.ts`, read over `dumpsys input_method` for fill verification, and Android snapshot nodes carry no IME-window provenance beyond package and resource id: an IME whose package we cannot name is indistinguishable from no IME in that tree. `no-keyboard` is then the honest reading of *the tree*, and the fix belongs where the fact would be added — the Android capture tagging IME windows with the package the platform already reads. I can file that if you want it tracked in this shape.

**Living next to `isTapPointInsideViewport`.** Kept separate, for two reasons. `snapshot-visibility.ts` answers one node against one window rect; this decides which system surface owns a point across the whole tree, which needs keyboard ownership — types, `UIKeyboardDockItemButton` / `UIAccessibilityElementKBKey`, IME package provenance, and the plane-vs-leaf flattening that decides whether a reported rect may excuse a bare coordinate. And its two consumers choose different consequences for one verdict: #2602's acting paths refuse while its coordinate path discloses. That module also carries a cross-language golden table, which `snapshot-visibility.ts` has no fixture for, and the ADR 0011 cells name this subpath as their enforcement site.

Also from #2602's review, landing here because the text is owned here (`6e13b1ccd7`): the refusal hint now names the snapshot as the source of the verdict and `snapshot -i` as the re-measurement.

Gate on this branch: Gate run on the stack top (`fix/tap-keyboard-occlusion-tap` at `09856dc1b3`, which contains this branch): every stage `check:affected` selects passes, `vitest related` green at 328 files / 2375 tests. `mutation-model` still fails `a kernel is owned by tests that reach it indirectly`, which fails identically at detached `origin/main` `b1c068da4e`.

thymikee:
Simplification pass pushed as `839fcc7ceb`, no behavior change: the width rule was taking a second span over the surface rects while the band was about to be built from the anchor span. Measured trees place the dock's buttons inside the key columns, and the landscape read the rule exists to catch (key plane 162 x 327) fails on either span, so the pair only widened the distance between what gets checked and what gets returned. The module doc also stopped calling the point rule "the center rule", which described the version #2602 no longer ships.

`packages/contracts/src/tap-keyboard-occlusion.test.ts` green over all 30 cases with no case rewritten.

thymikee:
Reviewed at 839fcc7. The band now docks to the bottom edge, so iPhone landscape no longer turns the guard off, and the measured rotated tree gives `undetermined`. Two layouts and one test gap are still open.

A split iPad keyboard docked at the bottom can still refuse a visible target. The docking rule at [tap-keyboard-occlusion.ts#L187](https://github.com/callstack/agent-device/blob/839fcc7ceb70e526e3938d15a77b9d8c77416d24/packages/contracts/src/tap-keyboard-occlusion.ts#L187) checks only the lowest bottom, and the frame spans from the leftmost to the rightmost anchor. Both key clusters of a docked split keyboard reach the bottom, so the frame covers the gap between them, and #2602 would refuse a tap on app content that is visible in that gap. The split fixture's keys stop more than 80 pt above the bottom, so the table does not catch it. Could disjoint anchor columns return `undetermined`, or build one band per cluster, with a table row for this case?

Could the Android guard fail open above 3-button navigation? `KEYBOARD_BOTTOM_ANCHOR_TOLERANCE = 80` at [#L170](https://github.com/callstack/agent-device/blob/839fcc7ceb70e526e3938d15a77b9d8c77416d24/packages/contracts/src/tap-keyboard-occlusion.ts#L170) is sized in iOS points, but Android rects are in pixels. 80 px is about 30 dp at 420 dpi, less than a 48 dp navigation bar, and every Android fixture sits flush with the bottom. A measured Android IME rect with 3-button navigation would settle it; a tolerance relative to the viewport would also avoid the unit mismatch.

[tap-keyboard-occlusion.test.ts#L79](https://github.com/callstack/agent-device/blob/839fcc7ceb70e526e3938d15a77b9d8c77416d24/packages/contracts/src/tap-keyboard-occlusion.test.ts#L79) compares `surface.frame` only for occluded cases and never compares `surface.controlRects`. #2602 consumes this surface, so a regression in plane exclusion, or in the frame of a clear verdict, would pass. Could cases declare expected `controlRects`, with the frame compared whenever a case declares one?

The size question from the last round also stands: this adds about 251 production lines in a new contracts subpath. Could it sit beside `isTapPointInsideViewport` in snapshot-visibility.ts instead?

CI is green on 839fcc7, and there are no conflicts. The next step is a table row and a fix for the docked split iPad keyboard, and a check of the tolerance against a measured Android IME rect.

thymikee:
Status against `839fcc7ceb`. The geometry asks landed and are pinned by the golden table: the band now has to be docked to the bottom edge (`KEYBOARD_BOTTOM_ANCHOR_TOLERANCE`), which is what makes the iPad floating and split cases fail open instead of refusing, the rotated landscape keyboard is measured nowhere (`the landscape iPhone keyboard reports its rects in its own rotated space`), landscape reported wide and docked is measured against the landscape viewport, and `resolveVisibleKeyboardSurface` is module-private with `resolveKeyboardTapOcclusion` the only export.

One case from the review is still open, and it is this PR's to carry:

**An Android keyboard outside `KNOWN_ANDROID_IME_PACKAGES` reads `no-keyboard`, not `undetermined`.** `isKeyboardAnchorNode` (`packages/contracts/src/tap-keyboard-occlusion.ts:86`) calls `isAndroidInputMethodNode(node)`, and that wrapper (`packages/contracts/src/android-input-ownership.ts:73`) forwards only `packageName` and `resourceId` — it drops the `activeInputMethodPackage` argument that `classifyAndroidInputOwnership` already accepts and would answer with `source: 'active-input-method'`. With no anchor node the resolver takes the `:226` branch and reports the one outcome that is a confident negative: nothing is on screen. On Gboard and SwiftKey that is right; on a third-party IME the tap behind the keys is measured against a keyboard the rule decided does not exist.

The seam is one optional argument threaded from `resolveKeyboardTapOcclusion` down through `resolveVisibleKeyboardSurface` → `isKeyboardAnchorNode` → `isAndroidInputMethodNode`. The fact itself is already read on this platform (`packages/platform-android/src/device-input-state.ts:156`, `readAndroidActiveInputMethodPackage` from the same `dumpsys input_method` that gives visibility), so what is undecided is only where the read belongs: recorded beside the capture so the guard costs nothing at interaction time, or read when the guard runs. The two other users of that wrapper have the same blind spot and would want the same answer — `packages/capture-kit/src/snapshot-chrome.ts:242` and `snapshot-state.ts:103` both classify IME nodes through it, so an unlisted IME is neither an overlay there nor a band here.

Happy to take either placement; the capture-side one is the smaller diff for the tap guard and fixes all three call sites at once.

thymikee:
All four addressed. Pushed as `65eb2b044c` (stack top `69c67c7459`).

**Docked split keyboard.** The band has a third measurement rule now: the keys must form one unbroken run of columns. I took the `undetermined` option deliberately. A per-cluster band needs the cluster model that just produced this bug class — nothing in the tree says where either cluster ends — and guessing it wrong refuses visible content instead of disclosing. So a gapped band reads like a rotated keyboard or a floating one: unmeasurable, fails open.

Seams are measured rather than assumed: iPhone 17 Pro (26.2) reports a key row adjoining across 395 pt, iPad Pro 11-inch (M4) across 743.5 pt, and Gboard on Pixel 7 across 1070 px, none with a gap above a point, so the allowance is for projection rounding and not layout. Two new rows: a docked split keyboard whose container spans 0..834 while its clusters reach 1207 of a 1210 pt viewport and an app control sits in the gap, and a measured docked iPad keyboard, `occluded` at frame 0,925 834x285.

The fallout is worth naming: eight table cases modelled a keyboard as `Q` plus `return` with 257 pt of nothing between them, which is exactly the shape the new rule refuses, so they now carry the measured Q..P row, and the shared runtime fixture in #2602 gained the measured iPhone bottom row for the same reason. A reduced model was hiding a layout claim. The residual risk is a projection reporting a gapped *subset* of keys, which would read unmeasurable and fail open; both projections I measured report every key with zero gaps.

**Android budget.** Measured on Pixel 7 (Android 16, 1080x2400 at 420 dpi) with three-button navigation: the bar is 126 px at 2274..2400 — 48 dp, your arithmetic exactly — and Gboard's reported region runs to the *physical display bottom* at 2400, so the bar sits inside what the tree reports. The interactive projection even attributes `Back` and `Switch input method` to the IME package at that rect. Classified against that live tree: band `x 5, y 1517, 1070x883`, `(540, 1800)` occluded, `(540, 1400)` clear. So the guard does not fail open above gesture navigation, and the case is a table row now.

I did not scale the budget to the viewport, because what that budget is load-bearing against is the iPad floating keyboard, which sits a visible slice of the screen above the bottom edge — and this environment has no iPadOS runtime whose float/split toggle I could reach, so I have no measured floor for a larger budget. Covering a 48 dp bar on a 2400 px Android tree needs ~240 px, which is the same number that would let a floating keyboard dock and then refuse the content below it. A dp conversion would need a density the contracts layer does not have. If an IME does ship reporting above its own bar, it reads `undetermined` and fails open, which is today's behaviour on main rather than a false refusal.

**Test gap.** Closed at the strongest form I could state it: every case that reaches a band declares `frame` *and* `controlRects`, the frame compares on `clear` verdicts too, a case declaring geometry that lands on `undetermined`/`no-keyboard` fails, and a non-surface case declaring either fails. 24 of 33 cases carry a band. Frames re-checked against node rects by hand, including the three measured ones.

**Placement.** Numbers first, since the size question has come round twice: the module is 339 lines, of which 145 are documentation and 171 are code — the 251 production lines figure is mostly the measurements that justify each constant. `snapshot-visibility.ts` is 269, so merging produces a 559-line file where the repository targets at most 300 and requires extraction past 500. It is also two questions: `isTapPointInsideViewport` asks whether a rect is inside the window that shows it, per node, with no keyboard in its vocabulary; this module asks which system surface owns a point across the whole tree. The module in this package that already reads keyboard geometry to decide what an interaction may touch is `scroll-gesture.ts` with `clipScrollViewportAboveKeyboard`, which is why the doc calls this its tap twin. Placing it beside `isTapPointInsideViewport` is where those two questions would meet in the middle of neither — but that is your call to make and I will move it if you still want it there.

Live checks after the rule change, on a restarted daemon: iPad Pro 11 with the real 33-key docked keyboard — a coordinate inside the band that lands on no reported control taps and discloses with `keyboard frame 925..1210`, a coordinate on a key taps silently. Android with three-button navigation and the live Gboard tree classifies as above. The four acting paths are the ADR 0011 cells, green.

Gate on the stack top: lint, typecheck, layering, di-seams, `check:fallow` (`✓ No issues in 22 changed files`), mcp-metadata, build, package, integration-node and macos-coverage all green; the related vitest stage passes at 328 files / 2375 tests, though one attempt tripped the wall-clock gate on `runner-client`'s xctestrun case at 5.19 s against a 2.5 s budget — that test measures 77 ms standalone. `mutation-model` still fails its pre-existing assertion, unchanged at `origin/main`.

thymikee:
Reviewed at 65eb2b0. The four findings from 839fcc7 are addressed, and one small cleanup is left.

[tap-keyboard-occlusion.ts#L134-L143](https://github.com/callstack/agent-device/blob/65eb2b0/packages/contracts/src/tap-keyboard-occlusion.ts#L134-L143) has two `/** */` blocks back to back on `collectKeyboardPlaneIndices`. The first one still says the walk climbs from every surface node, but that walk moved into `collectAncestorSurfaceIndices`. Tools read only the last block, so the first is dead and wrong. Please delete lines 134-138.

Two questions about the new column-run rule. A keyboard whose keys are naturally gapped, like the iOS emoji keyboard with only a few corner keys, would read as undetermined and turn the guard off; is that an accepted gap, and should a table row pin it? And when a projection reports only a full-width keyboard container and no keys, does `keysFormOneColumnRun` still put a band across a docked split keyboard's gap, and have you seen a projection that reports it that way?

All checks pass on 65eb2b0, and there are no conflicts. The next step is to remove the stale doc block.

thymikee:
Squashed both asks into the tiling commit, now `ceff9c3d34`.

**Stale block** — deleted, so `collectKeyboardPlaneIndices` says one thing. The module is 339 lines; the trade-off note the two rows below need took back most of what removing it gave.

One correction against my earlier placement comment: I cited a 300-line file target, and `#2620` deleted those tiers. `AGENTS.md` now flags only past 1,000 lines, so that argument was stale — the case for the module's size is cohesion against the golden table that consumes it, not a budget it was never under.

**Sparse layouts, e.g. the emoji panel.** Accepted gap, yes — `undetermined`, and both callers fail open. Pinned:

> `a sparse layout such as the iOS emoji panel reports corner keys over a wide container, does not tile, and reads undetermined instead of guessing a band`

Corner keys at each edge and a wide delete across the bottom leave the same middle gap that a split cluster leaves, and the only thing that tells the two apart is knowing which keys a given layout is supposed to have. The rule refuses to know that, so it refuses to measure — which is how rotated and floating already behave. It can only lose coverage, never invent a refusal.

Provenance, since it matters for the row: I did not capture that tree. Two attempts on iPhone 17 Pro this round (Settings' search field, Safari's address field, both in `kb2589b`) came back with no keyboard anchors at all, so the row is modelled from the layout you described at 402 x 126 with corner keys and the wide delete, over the iPhone viewport I did measure, and its name says modelled. Captured here is the opposite half: the QWERTY row tiles with zero seams on iPhone (395 pt), iPad (743.5 pt) and Gboard (1070 px). If someone captures the real emoji tree and it turns out to tile, the row changes and coverage goes up.

**Container-only projection.** Yes — with no keys reported there is nothing to run the rule against, the fallback keeps all anchors, and the band is whatever the container claims. For a split keyboard reported as one full-width container that band runs across the gap and the element path can refuse content sitting in it. Measured so far, no projection does that: iPhone and iPad report their keys (`Keyboard` plane plus 29–33 `Key` nodes, in the regular and the interactive projection both) and Gboard reports ~129 IME nodes including keys. The container-only cases in the table are harness shapes, and the new row says as much:

> `a projection that reports only the keyboard container falls back to the container width it claims`

That row also exposes something worth having written down: a container with no descendants is not a plane, so it lands in `controlRects`, and a bare coordinate inside that band goes unwarned as "on a keyboard control". Elements behind it still refuse, so the asymmetry is confined to the coordinate disclosure. If a real container-only split keyboard shows up, the fix is to retire the fallback to `undetermined`, not to start clustering.

Gate state for `ceff9c3d34` is in the body: `check:affected --run` stops at the `typecheck` stage on `installArgs`, which is upstream (`origin/main`'s own `Typecheck & Package` is red on `2cafab3ad0`); the stages past it pass run individually.

thymikee:
Quote correction: that first line was the fixture *description* sentence, not the case name. The row itself is

> `a sparse emoji layout is not one run of columns, so the guard abstains instead of guessing a band`

with the emoji-panel geometry (402 x 126 container, two corner keys, wide delete) in its `nodes`, `expected.kind` `undetermined`, and no `frame`/`controlRects` — which the table's own assertion forbids on a non-band verdict. The container-only quote was exact.

thymikee:
Reviewed at ceff9c3, as a follow-up to [the review at 65eb2b0](https://github.com/callstack/agent-device/pull/2601#issuecomment-5675796016). The earlier findings are addressed: the stale block is gone, and the two table rows and the doc paragraph are in. I have no new code findings.

One question before this is ready for review: this PR adds a new contracts module and golden table (339 production lines) with no consumer until #2602. I did not find a smaller shape, and a module-private classifier with one export already looks minimal. Can you note in the PR why a smaller design was rejected, so the added abstraction has its justification on record?

The emoji-panel row is built from geometry, not captured from a device, so whether a real iOS emoji tree tiles is still unverified; the classifier fails open there, so this affects coverage, not correctness.

Typecheck & Package fails with TS2353 on `installArgs` in `packages/platform-android/src/__tests__/snapshot-helper-install.test.ts`. This PR does not touch that file, and `main` fails the same way at 2cafab3 because #2618 removed the field while that test still sets it, so the failure is unrelated to this PR. A rebase will not clear it until `main` is fixed. Smoke Tests was still queued.

thymikee:
Gate state settled on `ceff9c3d34`:

- `Smoke Tests` failed its first attempt on `testAlertAcceptDoesNotActivateAReplacementWithASharedButton` (`RunnerTests+AlertObservationTests.swift:72`, `ALERT_DEADLINE_EXCEEDED`) and passed the rerun (`34949483945` attempt 2). Nothing in this PR reaches the runner — the diff is one contracts module, its test, the golden table, an export entry and the guarantee matrix — and #2602's copy of that lane passed on its first attempt at a head carrying the same Swift, so the file's deadline assumptions are the thing to look at, not this diff.
- `Typecheck & Package` fails at `snapshot-helper-install.test.ts(29,3): error TS2353 … 'installArgs'`, byte-identical to `origin/main`, whose own copy of that job is red on `2cafab3ad0` (run `34944609119`).
- Everything else green.

thymikee:
Both answered, at `dafab14379` (stack top `73a7aec338`).

**Why not smaller** — it is a `## Design` section in the body now, covering the five shapes that were rejected and the reason each fails: inline in the guard (two zone readers, and `R2 commands-floor` was not hypothetical — #2622 is a four-line `find` guard that died on it), a covering-surface kind in the occlusion classifier (same-window, rect-shaped, no outcome for unmeasurable geometry, and the coordinate path needs a warning rather than a refusal), extending `ios-system-surface.ts` (closed bundle-id registry keyed by process, whose invariant is activation), reusing `snapshot-visibility.ts` or folding into the guarantee matrix (a viewport rule and a path-declaration table respectively), and shipping it inside #2602. For scale, the zone carries fixture-backed siblings at 69, 88 and 269 lines; this is 339 with five exports, 136 of them prose.

The smallest shape I would still take is enforcement-side rather than module-side: a node stage declared in the pipeline policy replaces four call sites with one declaration and hands `find` its cell at the same door. That is #2622, and it needs a wider stage contract than any current stage asks for, so it does not ride in here.

**Typecheck** — repaired on this branch: `test(android): drop the retired installArgs from the install-timeout fixture`. One dead key out of a typed literal; #2619 retired the field, the parser ignores it on older blobs, `snapshot-helper.test.ts` pins that, and nothing read it. `pnpm typecheck` is clean repo-wide for the first time since `2cafab3ad0`, which also un-reds `Typecheck & Package` on `main` once this merges. It is its own commit and trivially liftable if you want it on `main` first — say so and I will open the standalone and drop it here.

**Gate** at `73a7aec338`: `format`, `lint`, `typecheck`, `layering`, `di-seams`, `fallow`, `mcp-metadata`, `build`, `package`, `integration-node` and `macos-coverage` passed. `vitest-related` then timed out `ensureXctestrunArtifact aborts only the disconnected request build…` at its 5s budget during an 82s/330-file run, and passed unchanged on rerun: 330 files / 2,380 tests, and 50/50 in 920ms standalone. `mutation-model` never got to run because the gate stops at the first failure; it fails identically on `origin/main`. Body's Validation section carries the same.

Emoji row: agreed — still modelled, coverage rather than correctness. Capturing it needs a keyboard-panel switch on a simulator that has not cooperated twice, so I left the row named as modelled instead of letting it read as a capture.

Issue #2546, opened 2026-09-13
iOS runner smoke: alert-observation XCTests are red or flaky on main


thymikee:
## Purpose

`Smoke Tests` in the `iOS` workflow is not deterministic on `main`. Two alert-observation XCTests fail on the runner host without any Apple source changing, so every PR that reaches that step has to be re-run before a reviewer can trust it.

## Observed on `main`

- `main@973b74cc14` (push run [34741577232](https://github.com/callstack/agent-device/actions/runs/34741577232), job 103682270067, step `Run targeted iOS runner XCTest regressions`), `RunnerTests+AlertObservationTests.swift:72`:

  ```text
  XCTAssertEqual failed: ("false") is not equal to ("true")
  - Optional(… code: "ALERT_DEADLINE_EXCEEDED", message: "alert dismiss exhausted its deadline")
  ```

  Failing test: `testAlertDismissDoesNotActivateAReplacementWithTheSameTitle`.

- PR #2533 head `a16821b661` (run [34742468875](https://github.com/callstack/agent-device/actions/runs/34742468875) attempt 1, docs-only diff), same file at `:75` and `:80`:

  ```text
  XCTAssertEqual failed: ("First actions: 0; replacement actions: 0") is not equal to ("First actions: 1; replacement actions: 0")
  ```

  Failing test: `testAlertCannotProveAnIdenticalReplacementAndDoesNotActivateIt`. Attempt 2 of the identical run and commit passed at 06:46 UTC, so the second case is a flake rather than a regression.

Both symptoms share one shape: a presented alert reports no actions, and the assertion times out through `ALERT_DEADLINE_EXCEEDED`.

Separate symptom on the same step family, worth tracking apart from the two above: `main@8e8eeb2ced` (run [34712516058](https://github.com/callstack/agent-device/actions/runs/34712516058)) failed `Run fixture-backed iOS simulator E2E smoke` with an assertion raised at `test/integration/ios-simulator-e2e/live-automation-scenario.ts:33` (`observeFixtureHome`).

## Required behavior

- The alert-observation assertions in `apple/runner/AgentDeviceRunner/AgentDeviceRunnerUITests/UnitTests/RunnerTests+AlertObservationTests.swift` pass or fail on their own terms, not on simulator alert presentation timing.
- A presented system alert that the test can see must expose its button list to the probe, or the probe must report the empty-button case as a distinct typed outcome instead of consuming the whole deadline.
[...]

PR #2603, opened 2026-09-14
fix(android): install the snapshot and IME helpers as plain, non-test-only APKs

### LINKED ISSUE #2364: Android: ColorOS blocks the test-only snapshot helper as a "PC install attack" (-99), removing all snapshot-based commands
## Summary

On ColorOS (OPPO), the bundled Android snapshot helper cannot be installed. The device's package installer classifies the install as a **"PC install attack"** and fails the session with vendor code `-99`, so every `snapshot`-dependent command is unusable on these devices.

This is **not** a blanket adb-install restriction: normal (non-test-only) APKs install over the same adb connection without any prompt. The block is specific to APKs carrying `android:testOnly="true"`, which the snapshot helper does.

Filing this because #1284 / #1285 deliberately chose hard-fail over the stock-UIAutomator fallback. That decision is sound, but it means an OEM that blocks test-only packages removes the entire Android capture surface with no supported escape hatch.

## Environment

| | |
|---|---|
| agent-device | 0.20.10 (global npm install) |
| Host | macOS 27.0, Node 22+ |
| Device | OPPO Find N6 (`PLP110`), ColorOS, `targetSdk=36`, physical device over USB |
| adb | 1.0.41 / 37.0.1 |

Screen was on and unlocked throughout. USB debugging authorized (`adb devices` reports `device`, not `unauthorized`).

## Reproduction

```
$ agent-device snapshot -i --json
Android snapshot helper failed: Failed to install Android snapshot helper:
failed to install .../agent-device-android-snapshot-helper-0.20.10.apk:
Failure [INSTALL_FAILED_TEST_ONLY]
```

Installing the same artifact by hand reproduces it and shows the vendor code:

```
$ adb install -r  <helper.apk>     # no -t
Failure [INSTALL_FAILED_TEST_ONLY]

$ adb install -r -t <helper.apk>   # with -t
Failure [-99]

$ adb push <helper.apk> /data/local/tmp/ && adb shell pm install -r -t /data/local/tmp/helper.apk
Failure [-99]
```

## Root cause (logcat)

```
E OplusCustomizePackageManagerService: getAdbInstallUninstallDisabled start
D PackageManager: send package verification broadcast
D PackageManager: installStage allowInterceptAdbInstallInInstallStage pkg:com.callstack.agentdevice.snapshothelper
I PackageInstaller: PC install attack detected!com.callstack.agentdevice.snapshothelper
D PackageInstallerSession: Marking session as failed: -99
D PackageInstallerSessionExtImpl: uploadInstallFailInfos:{code=20010,
   initiatingPackageName=com.android.shell, targetSdk=36,
   app_pkg=com.callstack.agentdevice.snapshothelper, statusCode=-99}
```

`-99` is not an AOSP `PackageManager.INSTALL_FAILED_*` constant; it comes from OPPO's interception layer (`OplusCustomizePackageManagerService`, `allowInterceptAdbInstallInInstallStage`).

## What I ruled out

| Hypothesis | Test | Result |
|---|---|---|
| Corrupt/truncated APK | `unzip -t` on the bundled artifact | Passes, no errors |
| Screen locked | `dumpsys window` during install | `mScreenOn=true`, `mScreenLocked=false` |
| AOSP adb-install verification | Set `verifier_verify_adb_installs` 1 → 0, retry | Still `-99` (restored to 1 afterwards) |
| **adb installs blocked wholesale** | **Pulled an already-installed, release-signed app off the device and reinstalled it with `adb install -r`** | **`Success`** |

That last row is the decisive one: the same adb channel installs a normal APK fine. Only the `testOnly` flag triggers the vendor block.

## What still works without the helper

Verified on the same device with a session open: `open` / `close`, `install` / `reinstall`, `screenshot` (produced a 1140x2616 PNG), `logs`, `apps`. Only `snapshot` and everything built on it (`click`, `fill`, `find`, `wait`, `diff`) are unavailable.

## Suggestions

1. **Document it.** A line in the Android installation docs — "some OEM builds (ColorOS observed) reject test-only packages from adb; the snapshot helper cannot be installed on those devices" — would save the diagnostic time. The current hint from #1285 is already good at saying *what* failed; naming this specific OEM class would say *why*.
2. **Consider an opt-in non-test-only helper build.** If the helper did not set `android:testOnly="true"`, it would install on these devices. I understand the flag is protective (it keeps the helper off Play and out of normal install flows), so this would need to be an explicit, clearly-labelled variant rather than the default.
3. **Consider an env var for a user-supplied helper artifact.** `adbProvider.snapshotHelperArtifact` already exists for cloud providers; exposing an equivalent for local runs would let affected users supply a locally-repackaged helper without patching `node_modules`. I checked the `AGENT_DEVICE_*` surface and did not find one.

Happy to test any of these on the device — I have it available.

thymikee:
## Summary

Fixes #2364.

Both Android helpers set `android:testOnly="true"`. That attribute was added in #464 purely to keep the `adb install -t` step from the `debuggable` era working; nothing depends on it. Instrumentation can only be started from shell via `am instrument`, and the helper's session server binds loopback, so the flag protects nothing here. What it does do is trip OEM adb-install interception: ColorOS classifies a test-only package as a "PC install attack" and fails the session with vendor code `-99`, which removed every snapshot-backed command on those devices.

Changes:
- Drop `android:testOnly` from the snapshot and IME helper manifests.
- Stop passing `-t` from the local install path (`helper-package-install.ts`) and from the release manifest's `installArgs`.
- Keep `-t` in the consumer allowlist so previously published release manifests still parse.
- Update both helper READMEs; the snapshot README now names why the flag is absent.

## Validation

- `pnpm build:android`; `aapt2 dump xmltree` confirms neither APK carries `testOnly`; release manifest `installArgs` is `["install", "-r"]`.
- Fresh API 36 emulator (Pixel 9 Pro XL, helpers uninstalled first): `adb install -r` succeeds for both helpers without `-t`; one-shot `am instrument -e mode viewport` returns `ok=true`.
- Through the daemon on the same emulator: `open` → `snapshot -i` (installs both helpers without `-t`) → `click text="Search settings"` → `fill @e2 wifi` → snapshot shows the field as `"wifi"`.
- `pnpm typecheck`; unit tests for `snapshot-helper`, `snapshot`, `touch-helper`, `adb-transfer`, and `request-router-android-snapshot-helper` pass.

Not verified: ColorOS itself. We have no such device; the reporter offered to test and has been asked to build from this branch.


thymikee:
Reviewed at 8d08bf2. I found no problems in the diff, and your fresh-install run on API 36 covers the changed install path.

The Smoke Tests failure is `RunnerTests.testAlertCannotProveAnIdenticalReplacementAndDoesNotActivateIt` in the iOS runner, which #2546 tracks as red or flaky on main. This PR does not touch the iOS runner, so it looks unrelated.

Two paths are still not checked live: an upgrade over an existing testOnly helper install, and a real ColorOS device. The reporter's confirmation on #2364 is the open item.

thymikee:
Reviewed at 8d08bf2. Dropping `android:testOnly` and the `-t` install flag matches the ColorOS rejection in #2364, and older release manifests that still list `-t` keep installing with it.

Does any test fail if `-t` comes back? In [`ime-helper.test.ts#L114`](https://github.com/callstack/agent-device/blob/8d08bf22f2a6ff3cd86f0e74e65af2141e3b7057/packages/platform-android/src/__tests__/ime-helper.test.ts#L114) the fake `install()` records only `apkPath` and `replace`, so the `deepEqual` passes with or without `allowTestPackages: true`. Could the fake record the full options object and assert that `allowTestPackages` is undefined? Then the fix has a regression guard.

Smoke Tests is still running on this head, and there are no conflicts. The ColorOS confirmation you asked the reporter for is still open.

PR #2611, opened 2026-09-14
fix(device-shell): one typed device-shell boundary for adb, exec-out, and hdc

### LINKED ISSUE #2026: Typed device-shell execution boundary (ShellSafe) — mechanism + complete migration, one PR
> **Status (2026-09-14):** rewritten. The original text pointed at `src/platforms/*` and `src/utils/exec.ts`, which no longer exist, and at a WIP branch (`claude/shell-safe-boundary-wip`) that is 340 commits behind and touches deleted paths. It also planned an AST backstop gate, which conflicts with the "fix-shaped, not detector-shaped" direction of #2531. Neither is a starting point. The design below replaces both; the WIP branch can be deleted.

## Threat

`adb shell`, `adb exec-out <cmd>`, and `hdc shell` join their argv after the subcommand into one string that the *device's* `sh` evaluates (adb deliberately does not escape: "just like ssh(1)"). The host never runs a shell (`runCmd` spawns with `shell: false`), so the surface is purely device-side: any unquoted dynamic element in a `['shell', …]` argv is a command injection.

State on `main` at `ad9b906140`:

- 165 raw `['shell'|'exec-out', …]` argv builds across 57 non-test files, with no funnel and no guard.
- Android quoting is ad hoc: `shellQuoteIfNeeded` is hand-applied in five files. The audited sites (`input text`, `cmd clipboard set text`, `am start` targets and deep links, launch args) are quoted.
- HarmonyOS has 35 `hdc shell` sites and **zero** quoting. `uitest uiInput text <text>` passes user text unquoted, and bundle ids, pids, and paths are inlined everywhere. This is the one live gap.

## Design: make the raw argv unrepresentable

One kernel module, `@agent-device/kernel/device-shell` (kernel because `platform-android`, `platform-harmonyos`, and `provider-limrun` all depend on it and nothing else in common):

- `deviceShellArgv(subcommand, words, prefix?)` is the only producer of a device-shell argv. It passes **every** string word through `shellQuoteIfNeeded` (identity on `[A-Za-z0-9_@%+=:,./-]+`, so a migrated site is byte-identical to today unless the word was an injection vector), stringifies numbers, and passes a `shellFragment(script)` verbatim. `prefix` carries transport options (`['-s', serial]`, `['-t', target]`).
- `shellFragment(script)` is the one escape hatch for genuine shell syntax (pipes, `&` background, redirects). Every interpolated value inside a fragment is `shellQuote`-d at the call site. It is greppable; that is its whole enforcement.
- `assertDeviceShellArgv(args, boundary)` refuses an argv containing `shell`/`exec-out` that `deviceShellArgv` did not build. It checks the value, so a variable-built or indirect argv fails the same as a literal one. No static gate, no allowlist.
- `shellQuote`/`shellQuoteIfNeeded` move here from `host-kit/command` (one implementation, as #2595 established; host-side hint text keeps importing them from the new path).

Guarded boundaries: `runAndroidAdb`; every executor handed out by `resolveAndroidAdbExecutor` / `resolveAndroidAdbProvider` / `createDeviceAdbExecutor` / `createLocalAndroidAdbProvider`; `runAndroidHostAdb`; `runHarmonyHdc`; and the generic host command port (`host.commands.run`) when the executable is adb or hdc. Derived arrays inside the provider scope (serial stripping, prefixing) are built after the check and are not re-checked, so the limrun executor needs no guard of its own.

Funnels: `runAndroidShell` / `runAndroidExecOut` (device-scoped, next to `runAndroidAdb`), `runAdbShell` / `runAdbExecOut` (for the injected-executor `…WithAdb` helpers, no `DeviceInfo` needed), `runHarmonyShell`. Host-port callers mint the argv with `deviceShellArgv` directly.

Call-site rule: drop the `'shell'` element, drop any hand-applied `shellQuoteIfNeeded`, split whole-command-line strings into words, and wrap only real shell syntax in `shellFragment`.

## Behavior changes to expect

- A word containing a space or metacharacter that was previously passed unquoted now arrives single-quoted. On Android this only affects values that were injection vectors; on HarmonyOS it fixes `uiInput text`.
- An empty word now renders as `''` (an explicit empty argument) instead of vanishing from the joined command line.
- `sh -c '<script>' name path` sites now hand `sh` the script as one argument, which is what `-c` requires.

## Delivery

One PR: mechanism, guards, all call sites, and the tests that assert argv shape (most stay identical because safe words render identically). No compatibility route is kept; a raw `['shell', …]` reaching any boundary throws `INVALID_ARGS` with reason `unguarded-device-shell-argv`.

thymikee:
## Summary

Closes #2026. `adb shell`, `adb exec-out`, and `hdc shell` join the words after the subcommand into one string the device's `sh` re-parses — adb escapes nothing on purpose, "just like ssh(1)". Any unquoted dynamic element was a device-side command injection. Android quoted a few sites by hand; HarmonyOS quoted nothing (`uitest uiInput text` passed user text raw).

`@agent-device/kernel/device-shell` is the only producer of a device-shell command:

- `deviceShellArgv(transport, subcommand, words, prefix?)` renders every word for the transport that carries it and returns a **frozen** command the funnel remembers by identity. The transport is an argument rather than a caller's choice because the two transports do not quote alike (see below); on `adb` the rendering is identity on the safe charset, so a migrated site is byte-identical unless its word was an injection vector.
- `shellFragment(script)` is the one greppable escape hatch for genuine shell syntax (pipeline, redirection, `& … & echo $!`), each of its sites interpolating through `shellQuote`.
- `assertDeviceShellArgv(args, boundary)` refuses a `shell`/`exec-out` command the funnel did not build, at every dispatch boundary: the adb executors and spawners the cluster hands out, scoped provider `exec`/`spawn`, the root adb host (`runAndroidHostAdb`, exec **and** spawn), `runHarmonyHdc`, and the generic host command port for adb/hdc (`guardedHostCommandArgv`). Acceptance is by identity, checked on the value, so a variable-built command is caught as fast as a literal one and a copy is refused rather than trusted. Refusal is `INVALID_ARGS` with `details.reason: 'unguarded-device-shell-argv'`.
- The one rebuild a transport may perform is taking **its own addressing** out of a minted command: the `-s <serial>` pair a provider scope adopts, or the option run `parseAndroidAdbArgv` moves into a typed target. That is `relayDeviceShellArgvWithoutOptions(args, removedIndex, removedLength)`, and it mints the result only while the removal stays inside that command's own option run — a removal that reaches the device command, or an argv that was never minted, comes back unminted and is refused. There is no general relay: an array the funnel did not build cannot be minted by copying it.

Addressing stays where #2632 put it. An Android device command is minted as the **payload** and travels in `AndroidAdbInvocation.command`; a command that must carry addressing in the same argv — an `hdc` command over the generic host command port, or an adb route spelling `-s`/`-P`/`wait-for-*` inline — mints it with `prefix`.

Bridge-facing surface (`agent-device/android-adb`): `runAdbShell` / `runAdbExecOut` over an `AndroidAdbExecutor`, `runAndroidShell` / `runAndroidExecOut` over a `DeviceInfo`, and the `ShellWord` type. HarmonyOS gets `runHarmonyShell` inside its package. `shellQuote` / `shellQuoteIfNeeded` now live in the kernel module (one implementation); the Apple runner still reaches it through its host port, whose `Pick` now derives from the kernel module.

A host command's argv is `readonly` from `runCmd` down, which is what keeps the array a port checked the same array a transport dispatches — a copy would no longer be the array the dispatch guard recognizes. `AndroidAdbExecutor`, `AndroidAdbProvider.exec`, `AndroidAdbSpawner`, and `CommandExecutorOverride` all take `readonly string[]`.

Limrun's install path now asks the runtime adapter to stop the app instead of spelling `am force-stop` itself, which also keeps the kernel funnel out of `packages/provider-limrun`'s eager closure (no closure-budget growth).

### The two transports do not quote alike

Verified on a physical nova 14 rather than reasoned about: `hdc` **wraps every element it forwards in double quotes**, so the device shell parses our word inside a double-quoted context where a single quote is inert while `$`, a backquote, and `"` stay live. Single-quoting HarmonyOS words — the obvious reading of the adb rule — was therefore both corrupting and still injectable:

| element sent as one argv element | device answered |
| --- | --- |
| `'a;b $(id) …'` (our quote) | `'a;b uid=2000(shell) gid=2000(shell) …'` — quotes arrived as data, `$(id)` expanded anyway |
| `x"; echo INJECTED; "b` | `INJECTED` on its own line + `/bin/sh: b: inaccessible or not found` — a `"` escaped the argument |
| `` a;b \$(id) "d" `tick` … `` (escaped, unquoted) | the payload byte-identical |

So `adb` (which forwards words verbatim) single-quotes, and `hdc` escapes the characters its own double quotes leave live. Two HDC limits follow from the same measurement and are now refused instead of silently misaligned: HDC **drops** an empty element (`echo A '' B` prints `A B`), so an empty word is refused with `details.reason: 'hdc-empty-word-unsupported'`; and a `shellFragment` could only arrive there as literal text, so it is refused with `'hdc-shell-fragment-unsupported'`. Neither is reachable on the current HarmonyOS call sites, which use no fragments and never type an empty word — they are the boundary, not a fallback.

Behavior changes are in the CHANGELOG and `website/docs/docs/client-api.md`: metacharacter words arrive quoted (`adb`) or escaped (`hdc`), an empty word renders as `''` on adb instead of vanishing while HDC refuses one, an `sh -c` body reaches `sh` as one argument, and a raw `['shell', …]` argv from a custom executor or provider is refused. 180 files: a typed seam that existing callers can bypass is false confidence, so mechanism, guards, and call sites land together.

## Validation

Tested at `3128f26bc0` (`5ddc1882d5` plus the HDC quoting correction), rebased onto `0ebd2540a3`.

- `pnpm format` / `lint` / `typecheck` / `build`, `pnpm check:layering`: pass.
- `pnpm test:unit`: 1349 files, 10,579 tests pass. `pnpm test:integration:provider`: 66 files, 210 tests pass.
- `pnpm check:fallow`: no issues in the 180 changed files.
- `pnpm check:production-exports` reports the same 67 findings as the merge-base `0ebd2540a3` (confirmed by running the same gate in a clean worktree of that commit); this PR adds none.
- Live Android 16 emulator (`emulator-5554`, fresh `pnpm build` + helper artifacts from `pnpm build:android`, isolated state dir), zero `unguarded-device-shell-argv` refusals:
  - snapshot helper installed and served an interactive snapshot through the persistent session (version read, `pidof … || echo` runtime probe, `am force-stop` all minted);
  - `fill` of ``a;b $(id) 'q' |wc > /tmp/pwned`` into the Settings search field landed **byte-identical** in the field and created no file;
  - `screenshot` through `exec-out screencap -p`; `logs start`/`stop` streamed 3,647 logcat lines through the guarded spawner; `record start`/`stop` exported a playable 198 KB MP4 through the `screenrecord … & echo $!` fragment plus the `test -e` / `stat` / `kill` probes;
  - `clipboard read` on Android 16 still refuses with `UNSUPPORTED_OPERATION` and the substitute hint rather than reporting an empty clipboard (#2674 verdict preserved over the new funnel).
- Guard coverage now includes the background/`spawn` transport: `adb-executor.test.ts` asserts an unminted `['shell','logcat']` never reaches a scoped provider's process, and that a minted command forwards unchanged.
- Live HarmonyOS device (`4UQ9K25710015363`, nova 14, HDC `Ver: 3.1.0e`, fresh `pnpm build`, isolated state dir), zero `unguarded-device-shell-argv` refusals:
  - `open com.huawei.hmos.settings --foreground` and an interactive ArkUI snapshot (21 nodes, ref frames and the expired-frame refusal both exercised);
  - `fill` of ``a;b $(id) 'q' "d" `tick` > /tmp/harm-pwned`` into the Settings search field read back **byte-identical** from `snapshot --json`, with no `uid=` expansion anywhere in the tree and no file created — on `5ddc1882d5` the same call produced `'a;b uid=2000(shell) …`, which is what forced the HDC correction;
  - `screenshot` wrote a 1084×2412 capture through `snapshot_display -f` + `file recv` + `rm -f`;
  - `logs start` / `logs mark` / `logs stop` streamed 519 real hilog lines (`backend=harmonyos`, 76 KB) through the minted `hilog -P <pid>` command with its `-t <target>` prefix; `logs clear`/`settings clear` deliberately not run on a personal device.
  - Not covered: app launch of a sideloaded bundle — this device refuses to start unsigned samples with AppGallery `10106105`, so `bm dump -n` / `aa start` were confirmed to reach the device but not to complete.


thymikee:
Reviewed at 1392bf3. The funnel closes the injection paths #2026 describes, but relayed Android providers can start refusing commands after a long session.

The limrun route is accepted by content, and that record expires. `mintedDeviceShellCommands` is capped at 1024 entries, FIFO, and not refreshed on re-add ([`device-shell.ts:75`](https://github.com/callstack/agent-device/blob/1392bf324dc01328c666c2c7d0836c602f3582ac/packages/kernel/src/device-shell.ts#L75)). The limrun path copies the argv (`executeLimrunAndroidAdb` builds `['-s', serial, ...args]` before `assertDeviceShellArgv`), so it passes only by content match. `WINDOW_DUMPS` and `ACTIVITY_DUMPS` in `observation.ts` are minted once at module load. After about 1024 distinct device-shell commands (each tap and text chunk is distinct), their keys are evicted, and `readAppState` and `readBlockingDialog` on a limrun device throw `INVALID_ARGS unguarded-device-shell-argv`. The PR's own test shows `[...constant]` throwing after 2000 mints. Could the brand travel through the relay instead, for example a `withTransportPrefix(argv, prefix)` helper that re-mints, with a regression test that sends a module constant through the limrun relay after more than 1024 mints?

A design question: could the invalid state be unrepresentable instead of detected? If `deviceShellArgv` returned a branded readonly type and the executors took `NonShellAdbArgs | DeviceShellArgv`, with a brand-keeping prefix helper for relays, the content-keyed set, its eviction and the runtime refusals at four boundaries would go away. A cheap identity check could stay for untyped callers such as MCP and the host command port. That needs the `AndroidAdbExecutor` and `HostCommandRequest.args` contract types to accept a branded argv. If you rejected this shape, what blocked it?

The new `hdcShell` closure in [`platform-runtime-screen-recording-harmony-host.ts:63`](https://github.com/callstack/agent-device/blob/1392bf324dc01328c666c2c7d0836c602f3582ac/src/platform-runtime-screen-recording-harmony-host.ts#L63) repeats `runHarmonyShell` from `packages/platform-harmonyos/src/hdc.ts`, because the package entry exports only `runHarmonyHdc`. `deployment/runtime.ts` and the Android recording host re-spell `runAdbShell` the same way. Could the shell helpers be exported instead, so each platform has one funnel?

Moving `shellQuote` and `shellQuoteIfNeeded` into `device-shell.ts`, whose header says the host never runs a shell, pulls host-side callers (CLI hints, daemon client metadata, the Apple runner host env export) into the mint registry's module. That reverses the ownership #2595 just settled. Could they live in a neutral `kernel/shell-quote` module that `device-shell` imports?

HarmonyOS `type` and `fill` text with spaces or metacharacters now arrives as one single-quoted argument ([`input-actions.ts:39`](https://github.com/callstack/agent-device/blob/1392bf324dc01328c666c2c7d0836c602f3582ac/packages/platform-harmonyos/src/input-actions.ts#L39)). That is user-visible and needs a CHANGELOG line. The PR body also counts 7 `shellFragment` sites; I count 6.

Live evidence is still missing for two changed routes: a relayed Android provider such as limrun, and HarmonyOS `type`/`fill` with spaces and metacharacters on a device.

Smoke Tests was still running at review time. The diff touches every Android `adb shell` and `exec-out` call site, so a failure there would need its own look.

thymikee:
Reviewed at 1392bf3. One funnel for device-shell argv is the right fix for #2026, but the guard design has a failure mode, and a few pieces are duplicated.

[`device-shell.ts#L44`](https://github.com/callstack/agent-device/blob/1392bf324dc01328c666c2c7d0836c602f3582ac/packages/kernel/src/device-shell.ts#L44): a copied argv is accepted only while its content key is still in `mintedDeviceShellCommands`, which holds 1024 entries, evicts the oldest first, and does not refresh a key that is minted again. `observation.ts` mints `WINDOW_DUMPS` and `ACTIVITY_DUMPS` once at module load. As far as I can trace, a limrun-scoped Android session copies that argv in `executeLimrunAndroidAdb` (`provider-limrun/src/android.ts#L203`) and `runAndroidHostAdb` then checks it by content. After about 1024 other distinct mints (taps, typed chunks), the dumpsys read would throw `INVALID_ARGS`, so foreground and dialog observation starts failing in long sessions. The test "keeps a long-lived minted argv accepted…" asserts this refusal for a copied constant. Can these be minted at read time, like `app-state.ts` and `window-state.ts` already do?

[`platform-runtime-android-tool-host.ts#L36`](https://github.com/callstack/agent-device/blob/1392bf324dc01328c666c2c7d0836c602f3582ac/src/platform-runtime-android-tool-host.ts#L36): because the guard is identity-based, the PR drops the `[...args]` copies, widens `AndroidObservationHost.runAdb` from `readonly string[]` to `string[]`, and adds an `args as string[]` cast. A later defensive copy would silently fall back to the evictable path above. Could a branded argv type keep `readonly` without the cast?

[`platform-runtime-screen-recording-harmony-host.ts#L63`](https://github.com/callstack/agent-device/blob/1392bf324dc01328c666c2c7d0836c602f3582ac/src/platform-runtime-screen-recording-harmony-host.ts#L63): `hdcShell` repeats `runHarmonyShell` because the HarmonyOS entry exports only a deferred `runHarmonyHdc`. Could the entry export a deferred `runHarmonyShell` instead, so there is one HarmonyOS shell funnel?

[`platform-runtime-host-device-shell.ts#L9`](https://github.com/callstack/agent-device/blob/1392bf324dc01328c666c2c7d0836c602f3582ac/src/platform-runtime-host-device-shell.ts#L9): the executable normalization repeats `isAdbCommand` in `adb-provider-scope.ts#L249`. Could both use one shared helper?

[`daemon-client-metadata.ts#L263`](https://github.com/callstack/agent-device/blob/1392bf324dc01328c666c2c7d0836c602f3582ac/src/daemon-client/daemon-client-metadata.ts#L263): host-shell quoting for CLI and daemon recovery hints now imports from `@agent-device/kernel/device-shell`. Those strings quote for the user's shell, not the device, and this moves the code #2595 just placed. Could `shellQuote` stay in a neutral kernel module, with `device-shell` importing it?

There is no CHANGELOG entry, but users will see changes: HarmonyOS type/fill text is now quoted, an empty word renders as `''`, `sh -c` bodies reach `sh` as one argument, and a raw `['shell', …]` argv from a custom executor is now refused with `INVALID_ARGS`. Could an Unreleased entry name the refusal and the funnels?

Would a smaller design do? If `deviceShellArgv` returned a branded `DeviceShellArgv` and the adb/hdc executors rendered it themselves, relays would pass it without copying. That removes the WeakSet, the 1024-entry FIFO, the `readonly` widening and cast, and most of `assertDeviceShellArgv`; a cheap check for a literal `shell` at `args[0]` could still guard plugins. The `AndroidAdbExecutor`/`HostCommandRequest` types and the provider relays would have to accept the branded value first.

The live Android run in the description covers open/click/type/fill/screenshot. Could you also run a limrun or provider-scoped Android session, the native perf and screen-recording paths, and the snapshot-helper output-file read whose `sh -c` quoting changed? For HarmonyOS, typing text with a space would show whether `hdc` passes the quotes through literally.

Smoke Tests is still running, and the diff sits on every Android shell call it makes. There are no conflicts.

thymikee:
Addressed at `1006591fee`.

**Eviction.** The content-keyed window is gone. Acceptance is identity-only, and every relay that rebuilds the array carries the identity across with `relayDeviceShellArgv`: the limrun serial prefix (`executeLimrunAndroidAdb`), the provider-scope serial strip, the managed server-port rewrite in the root adb host, and the host command port copy. A plain copy is refused loudly at the next boundary rather than falling back to anything. `device-shell.test.ts` sends a module constant through prefix and strip relays after 5,000 other mints, and `limrun-runtime.test.ts` already drives `am start` through the real limrun relay into `runAndroidHostAdb`. `observation.ts` also mints at read time now, though nothing depends on that anymore.

**Branded type instead of a runtime guard.** What blocked it: `AndroidAdbExecutor` is the provider contract, implemented by limrun, the SDK's `limrun-runtime-dependencies`, and every fixture provider, and the host side spawns from `runCmd(cmd, string[])`. A branded *object* argv would force every implementer and relay to render, and `HostCommandRequest` is shared with xcrun/hdc. A branded *array* type is compile-time only, so a relay copy would still need the runtime identity to survive; the guard is what makes a variable-built argv fail. So the split is: types carry `readonly string[]` end to end (no cast, no widening; the observation contract is readonly again), identity carries the mint at runtime, and the four boundaries stay.

**Funnels on the entries.** `runHarmonyShell` is a deferred export of the HarmonyOS entry and `runAdbShell`/`runAdbExecOut` are exported from Android mechanics; both recording hosts use them. `deployment/runtime.ts` keeps a local helper because it runs through the generic host command port (`commands.run`), where the argv has to carry the `-t <target>` prefix; that helper is `deviceShellArgv` with a prefix, which is the host-port funnel.

**`shellQuote` home.** I tried the neutral `kernel/shell-quote` module with `device-shell` importing it. The eager-closure gate refuses it: five entries (`src/cli.ts` and three Apple facades on the host side, Android mechanics on the device side) grow by one module because the split replaces one module with two. The gate allows no growth on an existing entry. One implementation therefore has to live in `device-shell.ts`; the header now says the same POSIX rule serves host-side hints. If you would rather rename the module to something neutral (`kernel/shell`), that is a mechanical follow-up.

**CHANGELOG** has an Unreleased entry naming the refusal, the funnels, and the three visible quoting changes. Fragment count corrected to 6.

**Live evidence** added to the PR body: record start/stop, perfetto, perf memory/frames, simpleperf (reaches the tool; the emulator refuses perf events), and the `sh -c` read shape run on the device, where the old unquoted shape fails with `exit: too many arguments`. No limrun account or HarmonyOS target is available here; those remain unit-only.

thymikee:
Addressed at `1006591fee`; the details are in my reply on the first review.

- `device-shell.ts#L44`: the FIFO window is gone. Identity-only acceptance, relays carry it with `relayDeviceShellArgv`, `observation.ts` mints at read time, and the regression test sends a module constant through prefix and strip relays after 5,000 other mints.
- `platform-runtime-android-tool-host.ts#L36`: no cast. `AndroidAdbExecutor`, `AndroidAdbSpawner`, the host port, `runAndroidAdb`, `runAndroidHostAdb`, `runHarmonyHdc`, and the limrun executor take `readonly string[]`; `AndroidObservationHost.runAdb` is readonly again.
- `platform-runtime-screen-recording-harmony-host.ts#L63`: the HarmonyOS entry exports a deferred `runHarmonyShell`, and the host uses it. Android mechanics exports `runAdbShell`/`runAdbExecOut` for the Android host.
- `platform-runtime-host-device-shell.ts#L9`: `deviceShellExecutableOf` in the kernel module is the one normalizer; the provider scope's `isAdbCommand` uses it.
- `daemon-client-metadata.ts#L263`: the neutral `kernel/shell-quote` split fails the eager-closure gate on five entries (one module replaced by two). One implementation stays in `device-shell.ts`; a rename to a neutral name is a mechanical follow-up if you want it.
- CHANGELOG: Unreleased entry names the refusal, the funnels, and the visible quoting changes.
- Live runs: recording, perfetto, perf memory/frames, simpleperf (emulator refuses perf events; the command line parsed), and the `sh -c` output-file read shape on the device (old shape fails, new shape works). No limrun account or HarmonyOS target here.

thymikee:
One more commit, `f97c77e169`: the readonly executor argv is a visible SDK type change, so the `client-api.md` snippet now annotates its custom executor as `(args: readonly string[])` and the CHANGELOG entry says so. `check:affected` on this head: three 5 s timeouts under contention (test-file-size ratchet, ios-record-trace, apple runner-client), each green in isolation and none on a path this PR touches.

thymikee:
Rebased onto `main` (two CHANGELOG conflicts, both entries kept) and trimmed at `a3a0ea2163`: `inventory.ts` now has one `runAdbShell(context, serial, words)` helper instead of five repeated `deviceShellArgv(..., ['-s', serial])` blocks, and the boundary lost a helper and a narration comment. I measured widening `runCmd` to `readonly string[]` to drop the three `relayDeviceShellArgv(x, [...x])` copies; it ripples into ~30 override mocks across capture-kit, platform-apple, and platform-web, so those three lines stay. After rebase: typecheck, lint, format, layering, fallow, unit (1306 files), provider-integration (66 files) all green.

thymikee:
Reviewed at a3a0ea2. Minting is now by identity in a WeakSet, which has no capacity, so the eviction failure from the last review cannot happen again, and every relay that rebuilds an argv carries the identity. Three problems remain.

Two long-lived process routes skip the guard. `guardProviderDeviceShell` at [adb-provider-scope.ts#L320](https://github.com/callstack/agent-device/blob/a3a0ea216356813323b0008642c05f6f5dc57fcc/packages/platform-android/src/adb-provider-scope.ts#L320) wraps `exec` but leaves `provider.spawn` raw, and `resolveScopedAndroidAdbBackgroundTransport` also hands out `scoped.provider.spawn` unguarded. `launchLocalAppLogCommand` (src/platform-runtime-app-log-process.ts:173) runs `runCmdBackground` without `assertHostDeviceShellRequest`, and HarmonyOS `hilog` uses that route. Every argv there is minted today, so nothing is open now, but #2026 asks for the invariant by construction. Could `spawn` and the background transport go through the same assertion, and could `launchLocalAppLogCommand` call `assertHostDeviceShellRequest` first?

[CHANGELOG.md#L10](https://github.com/callstack/agent-device/blob/a3a0ea216356813323b0008642c05f6f5dc57fcc/CHANGELOG.md#L10) tells custom adb executors and providers to use `deviceShellArgv` from `@agent-device/kernel/device-shell`, or `runAndroidShell` / `runAdbShell` / `runHarmonyShell`. The kernel package is private, no published entry exports these, and SDK providers never reach the guarded boundary. Could the entry describe the in-repo refusal instead, or should the funnel be exported if external callers need it?

[inventory.ts#L243](https://github.com/callstack/agent-device/blob/a3a0ea216356813323b0008642c05f6f5dc57fcc/packages/platform-android/src/inventory.ts#L243) adds a private `runAdbShell(context, serial, words)` beside the exported `runAdbShell(adb, words, options)` in the same package, plus a second spelling of the argv through `adbShellArgv`. Could the local helper get its own name, with `adbShellArgv` inlined at its one call site?

Is `relayDeviceShellArgv(received, rebuilt)` as strict as it needs to be? It re-mints any array without checking that the device-side tail matches `received`, so a relay could change the command and still pass. A relay that builds the array itself from `received`'s subcommand onward would make that impossible and drop the three `[...x]` copies.

The live Android runs in the comments were on 1006591, before the rebase and the inventory change. Please add a live Android run on a3a0ea2 that covers device listing, `open`, `snapshot` and `fill` through the provider-scope relay.

Smoke Tests failed on iOS only: the Settings replay `wait` timed out once and passed on retry, and the fixture step `wait text Agent Device Tester` timed out after runner modal-probe timeouts. For an executable other than adb or hdc, the host port returns early and passes the same array as before, so the failure is likely unrelated. There are no conflicts.

The next step is guarding `spawn` and the app-log background route, fixing the CHANGELOG note, and the live Android run on a3a0ea2.

thymikee:
At 5ddc188 the three findings from the last round are closed. The scoped provider now hands out a guarded `spawn`, and `adb-executor.test.ts` refuses an unminted `['shell','logcat']` before it reaches a process. The limrun app-log relay passes the minted argv by reference instead of copying it. The CHANGELOG now points SDK authors at `runAndroidShell` / `runAdbShell` on the public `android-adb` entry, which resolves for an npm consumer. CI is green here, 20 checks and none failing, and there are no conflicts.

Two guard gaps are left. Neither is reachable today, so both can be taken here or left as follow-ups.

`launchLocalAppLogCommand` ([platform-runtime-app-log-process.ts#L173](https://github.com/callstack/agent-device/blob/5ddc188/src/platform-runtime-app-log-process.ts#L173)) is the one `HostCommandRequest` consumer that still calls `runCmdBackground` with a copy of `request.args` instead of `guardedHostCommandArgv(request)`. The other two consumers are guarded. An unminted hdc or adb shell argv arriving on this long-lived route is passed to the device rather than refused, and the copy would make a guard added later refuse minted traffic instead. Passing `guardedHostCommandArgv(request)` closes it and drops the copy.

`resolveAndroidAdbTransferProvider` ([adb-provider-scope.ts#L217](https://github.com/callstack/agent-device/blob/5ddc188/packages/platform-android/src/adb-provider-scope.ts#L217) and L220) lost the `guardProviderDeviceShell` wrap it had on both branches at a3a0ea2; the squash looks to have dropped it. That makes it the one adb provider the cluster hands out unguarded, against what the comment on `guardDeviceShell` claims. It is inert because `adb-transfer.ts` only sends `pull` and `install`. Re-wrapping both branches is the smaller fix: hoisting the guard into `normalizeAndroidAdbProvider` is not safe, because `createAndroidCommandExecutorOverride` feeds it a payload that `serializeAndroidAdbInvocation` rebuilds unminted.

I ran nothing myself and took your reported runs as given: the live Android 16 emulator session at 5ddc1882d5 (the injection payload landing byte-identical in the Settings field, `exec-out screencap`, 3,647 logcat lines through the guarded spawner, the `screenrecord` fragment, and `clipboard read` still refusing per #2674), plus the unit, integration and `check:production-exports` gates. One question that run leaves open: did that daemon boot from dist? That is what would show the identity registry stays a single instance across bundle chunks, since a second instance would refuse every minted argv at a boundary loaded from the other chunk. A live limrun Android session and a HarmonyOS type or fill with spaces are still unrun.

PR #2613, opened 2026-09-14
feat(record): report whether the recorder stopped beside the export

thymikee:
## Summary

`record stop` answers the question a video cannot: whether its recorder actually stopped. `recorder` and `nativePathDisposition` arrive on CLI `--json`, the Node client, and MCP (ADR 0024 2.2/2.3). Additive — no stop changes its outcome, and older manifests and device markers parse unchanged.

Today a stop reports `confirmed`, or `lost` with `owner-session-lost` for an Apple recording whose session was invalidated, and names the recorder's own path `retirable` or `retired`. The wider vocabulary (`unconfirmed`, the identity-mismatch reasons, `pending`) is declared ahead of the ADR 0024 steps that gain those probes, and docs, help, MCP descriptions and the two vocabulary types now say which states have producers.

35 files, additive.

## Validation

Tested at `0f9b1138e7`, with the mutation-pin repair in `c69438a5fb`.

- `pnpm check:affected --run` at `0f9b1138e7`: every lane green (465 Vitest files, 3309 tests) except the mutation-model indirect-ownership pin, which has been red on `main` since #2499. That repair is `c69438a5fb`, where `pnpm run mutation:test` passes 40/40, and the restacked child passes `check:affected --run` in full.
- `pnpm lint` exit 0 and `pnpm format` clean at this head; `pnpm check:tooling` exit 0 at `bb7752c536`.
- Android emulator: `record stop --json` answers `recorder: "confirmed"` / `nativePathDisposition: "retired"` with the chunk gone from `/sdcard`, and a stop whose daemon was killed after its completed marker replays `retired` from the device instead of the `retirable` the marker froze. The failed-probe cases and both runs are written up in the review reply below.
- One drive-by: the `installArgs` fixture left by #2618 made `origin/main` fail its own Typecheck job (CI run `34944609119`); this branch drops it so the gate can run.


thymikee:
Reviewed at f0a8e1b. The new fields are additive on the CLI, Node client, MCP and daemon wire, older manifests and markers still parse, and no stop changes its outcome. Three reported details do not match what the backends do yet.

The docs, CHANGELOG and PR body say `recorder` is `unconfirmed` when a probe fails or no exit arrives, and `lost` when identity proves the writer is not ours ([commands.md#L969](https://github.com/callstack/agent-device/blob/f0a8e1b7cda08c6b0520bb41d6f0f1bce88677e9/website/docs/docs/commands.md#L969)). No backend reports `unconfirmed` or an identity mismatch today: Android ownership loss, a failed forced stop in `stopChunk`, and a refused runner stop RPC all still throw. A reader expects an export with `recorder: unconfirmed` and gets an error instead. Could the docs say these states are reserved for a later ADR 0024 step, and that today only `confirmed`, and `lost` for an invalidated Apple runner recording, are reported?

On Android the same recording can report two dispositions. The device-side completed marker is saved with `nativePathDisposition: 'retirable'` before `cleanupChunks` ([finalize.ts#L55](https://github.com/callstack/agent-device/blob/f0a8e1b7cda08c6b0520bb41d6f0f1bce88677e9/packages/platform-android/src/recording/finalize.ts#L55)), while the live result and the daemon manifest say `retired`. After a daemon restart, reattach reports `retirable` for chunks that are already gone, and `recovery.test.ts:126` pins that. A later step that reads the marker could try to retire them again. Could the marker be saved again as `retired` after cleanup succeeds, or could reattach report `retired` when the chunks are absent, with the test expecting one value?

The runner finish branch that sets `retirable` when `remotePath` is set ([runtime.ts#L237](https://github.com/callstack/agent-device/blob/f0a8e1b7cda08c6b0520bb41d6f0f1bce88677e9/packages/platform-apple/src/recording/runtime.ts#L237)) has no test; the existing `retirable` checks cover capture-kit, Android and HarmonyOS only. Could a runner runtime test with `remotePath` assert the disposition on the normal and on the invalidated finish?

The PR has a conflict with main in CHANGELOG.md. Because of it only 5 checks ran; they pass, but the full suite has not run on this head. The next step is to resolve the conflict, fix the Android disposition mismatch and the docs, and add the runner test.

thymikee:
Rebased onto `main` (CHANGELOG only) and took all three at the owning rule. `bb7752c536`.

**Docs overstated the states.** Fixed at the source rather than at the three quoted sites: the producer inventory now lives on the two vocabulary types (`recording-stop-observation.ts`, `recording-native-path.ts`), and docs, `--help`, CHANGELOG and the MCP field descriptions say what a stop emits today and that the rest of the vocabulary waits for the steps that gain those probes. Proof for the inventory — the only production constructions are `platform-apple/src/recording/runtime.ts:135,142,235`, `platform-apple/src/recording/completion.ts:24`, `platform-android/src/recording/finalize.ts:49`, `platform-harmonyos/src/recording/runtime.ts:184`, `platform-web/src/recording/runtime.ts:134`, i.e. `confirmed` plus `lost`/`owner-session-lost`; `unconfirmed`, `identity-not-ours`, `native-artifact-absent` and `pending` have no producer, exactly as you read it (Android ownership loss, a failed `stopChunk` and a refused runner stop still throw).

**Android reported two dispositions.** The fix is the observation, not the two values: `nativeChunksDisposition` (recording/chunks.ts) reads the device, and it is now the only decider — the stop answers with it after disposal, and the completed-evidence replay re-reads that one field instead of replaying a value frozen before the removal. One value from both paths, which is what `recovery.test.ts` now asserts. The marker keeps its pre-disposal `retirable` on purpose: publishing before cleanup is what carries an earned completion across a crash, and every reader now re-reads the one field that can go stale. Per the new guidance, the double fails the way a device does — `fixtures.ts` tracks remote files, so `exists` reports what a removal did or failed to do; a device that answers `rm` with success and keeps the file is now a case, and it stays `retirable` from both paths. Mutations run: serving the marker value unchanged fails the crash replay and both pid-reuse replays; hardcoding `retired` after cleanup fails the kept-artifact case.

**Runner disposition had no test.** Added for the `remotePath` branch on the normal and the invalidated finish (asserting `retirable` with `confirmed`, and with `lost`/`owner-session-lost` plus the overlay warning). Mutation run: `retirable` → `retired` at runtime.ts:238 fails exactly those two tests and nothing else.

**Conflict and gate.** Rebased; CHANGELOG was the only conflict. `pnpm check:tooling` exit 0 at this head. `check:affected --run` unit lane passes; five provider-integration/daemon-entrypoint cases timed out at their 5s budget under load average 60+ (another worktree compiling) and pass with `--testTimeout=30000` — rerunning on an idle host, noted in the body rather than papered over. Separately, `origin/main` fails its own Typecheck job (run 34944609119) because #2618 left `installArgs` in an annotated fixture; dropped in `f53e44a2b7` so the gate can run here — happy to split it out if you would rather it land alone.

thymikee:
Reviewed at bb7752c, as a follow-up to [the review at f0a8e1b](https://github.com/callstack/agent-device/pull/2613#issuecomment-5673820557). CI is green.

`nativeChunksDisposition` returns `'retired'` unless `exists` is truthy ([chunks.ts#L168](https://github.com/callstack/agent-device/blob/bb7752c/packages/platform-android/src/recording/chunks.ts#L168)). The production `exists` ([platform-runtime-screen-recording-android-host.ts#L58-L59](https://github.com/callstack/agent-device/blob/bb7752c/packages/platform-android/src/platform-runtime-screen-recording-android-host.ts#L58-L59)) runs `adb shell test -e` with `allowFailure` and returns `exitCode === 0`, so a failed probe (adb timeout, device drop) reads as absent and becomes `'retired'`. On reattach the marker is written before `cleanupChunks`, so after a daemon crash before cleanup, one failed probe reports `'retired'` for chunks that are still on the device. The fixture `exists` returns only a boolean, so no test covers a failed probe. Can the Android disposition be `'retired'` only when the device proves every chunk absent, the way `size()` already returns `'uncertain'` on a failed probe, and `'retirable'` otherwise? A recovery test with a failed probe after a crash before cleanup, expecting `'retirable'`, would cover it. How often the probe fails in practice is not known, so this is likely rather than confirmed.

The delta changes the Android device route, but the PR lists unit suites only. On an Android emulator with the repo CLI: `record start`, then `record stop --json` showing `recorder: 'confirmed'` and `nativePathDisposition: 'retired'`, with `adb shell ls` showing the chunk path gone. Then a second recording where the daemon is killed after the export and before disposal, restarted, and `record stop --json` returns the completed replay with a disposition that matches the device.

Next: the probe-failure disposition and the Android emulator runs.

thymikee:
Fixed at the probe, in `0f9b1138e7`, and rebased onto `main` (CHANGELOG was the only conflict). Head `c69438a5fb`.

**The transport was collapsing a failed probe into an absence.** `test -e` answers with an exit code, so only two answers are evidence: `0` is present, `1` with nothing on stderr is absent. Anything else — a killed probe, `adb` timing out, a device dropped mid-call — now returns `'uncertain'` from one `probeRemotePath` helper that `exists` and `size` share, so `exists` returns `false` only for a proven absence and `size` no longer reports "no file" for a probe that never ran. `nativeChunksDisposition` states the same rule in code: `retired` needs `false` for every chunk.

**The same over-claim sat one call away.** A reattach with a dead recorder asked whether the artifact was there and treated "could not ask" as "not there" (`!== true`), returning `transport-not-reattachable` with "ended before its artifact could be recovered". It is `=== false` now: an unanswered probe leaves the recording finishable so the caller can pull it once the device answers.

**Tests, and what they actually pin.** `credits a path as absent only from a probe that answered, not from one that failed` (host) — reverting `exists` to `exitCode === 0` fails it with `expected false to be 'uncertain'`, which is the reported bug. `keeps a recording finishable when a reattach cannot ask the device about its artifact` — reverting to `!== true` fails it with `expected 'unreattachable' to be 'active'`. `a probe that cannot answer after a crash leaves the retirement owed` covers your case: crash before disposal, replay's probe fails, both the stop and the replay answer `retirable`.

One disclosure about that last one: it passes with the disposition loop back on truthiness too, because the fixture's failed probe already answers `'uncertain'` and truthiness owes a retirement for that. The over-claim you traced lived in the host, not in the loop, so the loop change is the rule written down rather than the fix — the host test is what has teeth.

**Android emulator runs.** `record start` / `record stop --json`: chunk `/sdcard/agent-device-recording-1789478949114.mp4` present during, gone after, `recorder: "confirmed"`, `nativePathDisposition: "retired"`, 58,526-byte export. Then a recording whose daemon was killed by `SIGKILL` once the device showed a completed marker and before the stop returned: that stop failed with the daemon, the restarted one answered `recorder: "confirmed"`, `nativePathDisposition: "retired"` — the marker had frozen `retirable` and the device had no chunk, so the replay answered from the device. What did not reproduce: the marker itself stays on `/sdcard` after `close` until the next `record start` retires it, which `forceCleanup` skips once cleanup is confirmed. That is pre-existing and untouched here.

**Gate.** `check:affected --run` at `0f9b1138e7`: every lane green except the mutation-model indirect-ownership pin, which has been red on `main` since #2499 — that repair is now its own commit here (`c69438a5fb`) and `pnpm run mutation:test` passes at that head (40/40). The restacked child passes `check:affected --run` in full at its head.

thymikee:
Reviewed at c69438a, as a follow-up to [the review at bb7752c](https://github.com/callstack/agent-device/pull/2613#issuecomment-5677954396). The probe finding is fixed: `test -e` now counts only exit 0 as present and exit 1 with empty stderr as absent, every other result is `uncertain`, and both `nativeChunksDisposition` and the reattach check fail closed on it. The host test fails if `exists` goes back to `exitCode === 0`. The author-reported emulator runs cover the normal stop and the killed-daemon replay.

The mutation pin change points the indirect-ownership assertion at `back-runtime.test.ts`, which still reaches `scroll-edge-state`. There are no conflicts with main. Ready for human review.

PR #2618, opened 2026-09-15
refactor(android): drop the snapshot helper manifest installArgs

thymikee:
## Summary

Follow-up to #2603 / #2364. The reporter confirmed on ColorOS that `0.21.3` installs both helpers with plain `adb install -r`, which leaves the snapshot helper manifest's `installArgs` as a constant `["install", "-r"]`. The field only ever existed to carry `-t`. Its consumer chain parsed the array, validated it against a four-flag allowlist, lowered it to semantic options, and lowered those back to flags at the provider. None of that carried information any more.

Changes:
- Drop `installArgs` from the manifest type, the parser, `scripts/package-android-helper.sh`, and the README contract. The parser ignores the field on older manifests (test added).
- Install the snapshot helper with a fixed `replace: true`, the same path the IME helper already uses. The `readAndroidSnapshotHelperInstallOptions` export and its option type are gone.
- Narrow the adb provider `install` capability to `replace`. The `-t`, `-d`, `-g` options had no producer left; the type is internal to the platform package, not SDK-exposed.
- CHANGELOG entry under Unreleased.

Other helpers: the IME helper manifest never had `installArgs`, and the iOS runner manifest has no install-args analogue, so nothing to do there.

Not in this PR: the reporter's second finding (first helper install on ColorOS blocks on a system confirmation dialog, and the adb timeout hint then blames a wedged server) is a separate hint change.

## Validation

- `pnpm typecheck` passes.
- Unit tests: `snapshot-helper`, `adb-transfer`, `adb-executor`, `touch-helper`, `touch-helper-session`, `ime-helper`, `request-router-android-snapshot-helper`, `android-snapshot-capture-failure-reason` — 97 tests pass.
- `pnpm check:production-exports` reports no hits in touched files.
- The locally built `android/snapshot-helper/dist` manifest (untracked) still carries `installArgs`; the parser ignores it, so no rebuild is needed for source checkouts.


thymikee:
Reviewed at 2a14a6d, and I found no problems. `installArgs` has no reader left in the manifest, install path or adb transport, and the packaging script still emits the same `adb install -r <apk>` command that #2603 validated live, so the real install does not change.

Smoke Tests was still running at review time, not failed. There are no conflicts.

thymikee:
Smoke Tests has now failed on 2a14a6d in the live iOS simulator fixture run: `id="automation-longpress" did not become visible after scrolling` ([job](https://github.com/callstack/agent-device/actions/runs/34935298088/job/104271796449)). This PR changes only the Android snapshot helper install path, which that iOS scenario does not use, so the failure is likely unrelated; a rerun should confirm it.

PR #2619, opened 2026-09-15
fix(android): name the OEM install dialog on helper install timeouts

### LINKED ISSUE #2364: Android: ColorOS blocks the test-only snapshot helper as a "PC install attack" (-99), removing all snapshot-based commands

thymikee:
## Summary

The first `adb install` of each Android helper package can stall behind an OEM install-confirmation
dialog (ColorOS is one), so an unattended first `snapshot -i` or test-IME activation failed with the
generic "adb server may be wedged" advice. Helper installs now name the dialog and the fix.

- `installAndroidHelperPackage` is the shared install seam for the snapshot helper and the test IME
  helper. Timeouts there raise the install-confirmation hint; every other adb call keeps the
  wedged-server advice, and app-under-test installs are unchanged.
- Curated hints from provider/cloud install errors survive: only the generic timeout hint is replaced.
- `activateAndroidTestIme` helper-unavailable outcomes and the `android_test_ime_activate_failed`
  diagnostic now carry the curated hint.
- `isCommandTimeoutError` is now a type predicate (`error is AppError`).

Closes #2364. 15 files, +428/−16.

## Validation

Commit `6eb81957cd`. `pnpm check:affected --run` passed on that head (756 files / 5659 tests; fallow
dead-code and health baselines clean). `pnpm typecheck`, `pnpm lint`, `pnpm format:check` pass.

Regression proof: reverting each production edit fails its own test — snapshot-helper install timeout,
IME-helper install timeout, curated-hint passthrough, `adb-failure` hint policy, and the runtime warn
diagnostic. `app-deployment` pins the generic hint for app installs.

Unresolved risk: this is device-facing and no ColorOS/OPPO hardware is available here, so there is no
live run of the changed path. Confirming it needs `agent-device --platform android snapshot -i` on a
fresh install on an OPPO Find N6. Integration Tests and Coverage jobs are not observable locally.

Follow-up: the platform-apple host port still declares `isCommandTimeoutError(error): boolean`, which
loses narrowing there but does not change runtime behavior.


thymikee:
Reviewed at 6eb8195. The code looks correct: helper install timeouts now carry the install-confirmation hint through the shared helper install path, and the unit tests cover the classification with synthetic timeouts.

One validation step is left. This is a device-facing error path, and the tests do not show the hint on a real stall. As the PR body notes, the check is `agent-device --platform android snapshot -i` (or another helper install) on a device that shows the OEM install dialog, such as the OPPO ColorOS device from #2364, confirming the error carries the install-confirmation hint and not the wedged adb server hint. If no such device is available, a run that makes the real helper install time out on any Android device would still show that the hint reaches the user.

Smoke Tests failed in the iOS runner XCTest `testAlertDismissDoesNotActivateAReplacementWithTheSameTitle` (ALERT_DEADLINE_EXCEEDED). This diff touches no Apple or alert code, and the other 93 targeted XCTests passed, so the failure is likely unrelated. There are no conflicts.

Issue #2615, opened 2026-09-15
refactor(apple-runner): own close retention behind one runner operation


thymikee:
# refactor(apple-runner): own close retention behind one runner operation

## Purpose and evidence

PR #2605 (`44f78a986d51aeb0aab3e031700a96c2302e299b`) fixes reuse of a busy runner. Application lifecycle now combines `stopRunnerSessionIfBusy`, `scheduleRunnerIdleStop` and `stopRunnerSession`, although the runner owns occupancy and disposal state. Adding the busy primitive also changes shared contracts, forwarding modules and four replay fixtures.

Evidence: [application close orchestration](https://github.com/callstack/agent-device/blob/44f78a986d51aeb0aab3e031700a96c2302e299b/packages/platform-apple/src/lifecycle.ts#L328), [runner state owner](https://github.com/callstack/agent-device/blob/44f78a986d51aeb0aab3e031700a96c2302e299b/packages/platform-apple/src/runner/runner-session.ts#L588), [host contract](https://github.com/callstack/agent-device/blob/44f78a986d51aeb0aab3e031700a96c2302e299b/packages/contracts/src/application-lifecycle-runtime.ts#L315).

## Required behavior

Expose one awaited close-finalization operation through the existing runner host seam, conceptually `finalizeRunnerClose(deviceId, { retainForReuse: boolean }): Promise<void>`.

The application lifecycle supplies reuse intent; `runner-session.ts` chooses using its own occupancy state. Preserve this truth table:

| Caller intent / state | Required result |
| --- | --- |
| Retain + busy runner | Await disposal and lease release; never schedule reuse |
| Retain + idle runner | Schedule the existing idle stop; retain healthy warm reuse |
| Do not retain | Await the existing unconditional stop |
| Retain + no in-memory runner | Preserve existing idle-stop scheduling/cancellation behavior; do not create a runner |
| Do not retain + no in-memory runner | Still await the existing stop path, including idle-timer cancellation and owned-lease cleanup |
| Daemon shutdown | Preserve the separate shutdown/alert-cleanup path; do not introduce the ordinary close-finalization call |

Preserve alert-dismissal order, request behavior, error propagation, idle-stop timing and all occupancy update semantics from #2605. Keep lifecycle's shutdown decision outside this operation. Do not return early merely because the session map has no runner: unconditional stop also owns durable lease cleanup under its existing locks.

## Scope and exclusions

[...]

PR #2623, opened 2026-09-15
test(apple-runner): pin the close release truth table at the lifecycle seam

thymikee:
## Summary

Follow-up to #2605 for #2615. The seam the issue asks for already landed in this stack at `029c9e97`: `finalizeApplicationClose` issues one awaited `releaseRunnerOnClose(deviceId, { retain })`, `runner-session.ts` decides with its own `runnerMainThreadBusy`, and the busy-only operation plus its forwarding exports are gone. The issue still owed the proof, so this PR adds it: one test file, plus four daemon mock entries for the removed `scheduleIosRunnerIdleStop` export.

`runner-close-finalization.test.ts` drives the real `finalizeApplicationClose` over the real runner module, composed as the root's lazy tools compose it, one case per truth-table row: idle-retain warm reuse (no relaunch), busy-retain disposal plus a fresh boot on reopen, non-retain stop with idle-stop cancellation, owned-lease release with no runner in memory, retain-with-no-runner starting and claiming nothing, daemon-shutdown deferral, disposal rejection before alert dismissal, and a drain that lands after close is issued (runner exit held open; no microtask counting). Occupancy is produced by a real `RUNNER_BUSY` refusal and cleared by a real stamped reply, never by a mocked method call.

## Validation

Tested `455f9ec5c3`. `pnpm check:affected --run` green: format, lint, typecheck, layering, di-seams, fallow, build, package, integration-node (95), macos-coverage (6), and `vitest related` (2920 tests, 421 files). Its first pass timed out 8 unrelated daemon-spawning cases at host load 60+ with sibling worktrees' gates running; re-run alone passed.

Mutation probes kill the exact case: busy check removed (3), release no longer awaited (6), lease cleanup removed, idle cancel removed. Temporary `ready`-based reuse condition: 2 files touched, no contract method, adapter or replay-fixture edit; removed.

No live run: no production line changed. The wedge scenario and device lanes belong to #2605.


thymikee:
Reviewed at 455f9ec. This is test-only (no production lines), the cases drive the real close lifecycle, and each truth-table row would fail if its branch in the production decision were removed. Checks are green. It stays stacked on #2605.

Issue #2616, opened 2026-09-15
refactor(capture-kit): share the bounded provider app-log poller


thymikee:
# refactor(capture-kit): share the bounded provider app-log poller

## Purpose and evidence

PR #2112 (`8dad6500a31178e4f11ec32fe981ce069aa75041`) adds an approximately 200-line app-log poller with the same acquisition rollback, polling state, tail overlap, finish memoization and disposal mechanics as Limrun. Fixes to these mechanics would require coordinated provider edits.

Evidence: [Doublespeed poller](https://github.com/callstack/agent-device/blob/8dad6500a31178e4f11ec32fe981ce069aa75041/packages/provider-doublespeed/src/app-log-poller.ts#L24) and `packages/provider-limrun/src/app-log-poller.ts` (compare the exact merged base before implementation). The PR acknowledges a shared-capture follow-up but no issue currently owns it.

## Required behavior and boundary

Extract the shared polling mechanics into existing `packages/capture-kit`, consistent with ADR 0019. A narrow input should contain a disposable reader, backend identity, app bundle ID, output path and the existing clock/output ports. Conceptually the reader provides `readLogs(appBundleId, lineLimit, signal): Promise<string>` plus async disposal.

Providers keep lease/session identity, authentication, descriptor codecs, connection verification, reconnect and exact-owner admission. Adapt provider-specific cancellation at their boundary: Limrun currently settles an uncancelable read through `awaitLimrunOperation`. Doublespeed's poller supplies a signal; its reconnect reader forwards it, while `currentAppLogReader` currently drops it. Characterize both reader paths and do not silently change cancellation semantics during extraction. Do not imply abort proves the underlying request ended. Preserve current behavior for late results and rejections.

Keep the current polling interval, read timeout, line limit, recovery states, output append/marker behavior, backend selection and user-visible errors. Provider-specific cleanup messages remain explicit input or adapter-owned formatting; shared policy must not name Limrun or Doublespeed. Correcting the reader cancellation difference is separate behavior-changing work. Do not turn this into a generic provider framework or broaden it to recording/perf.

## Completion and validation

- First compare both implementations and list intentional differences. Stop if one common lifecycle would erase a provider-specific ownership or cancellation guarantee.
[...]

PR #2112, opened 2026-08-28
feat: add direct Doublespeed provider runtime

hassantsyed:
## Summary

Adds `agent-device connect doublespeed`, a direct provider runtime for [Doublespeed](https://mac.doublespeed.ai) iOS simulators (a Mac mini fleet exposed over HTTPS, billed per second). It follows the Limrun provider's shape end to end so operators get the same workflow:

```bash
export DOUBLESPEED_API_KEY=...
agent-device connect doublespeed --platform ios
agent-device install com.example.app ./build/Example.app   # allocates the simulator, uploads the bundle once
agent-device open com.example.app --relaunch
agent-device snapshot -i
agent-device click 'label="Continue"'
agent-device close                                         # releases (and stops billing) the simulator
agent-device disconnect
```

What the provider owns:

- **Lease lifecycle**: `allocate` creates a labelled simulator session and waits until its session API is up; `release` ends it; `recoverExpiredLease` finds orphans by the `provider`/`leaseId` labels and ends them, exactly like the Limrun label-selector recovery.
- **Interactor** over the session's JSON API: open (bundle id or deep link), close, tap, long press, double tap, selector taps (`id`/`label`/`text`/`value`), type/fill, scroll, home, portrait/landscape orientation, screenshot, and accessibility-tree snapshots stamped `{ backend: 'xctest', producer: 'doublespeed-ios-tree' }` (new producer added to the kernel provenance table).
- **App deployment**: `.app` directories are zipped by the host adapter, published once per organization through a content-addressed asset endpoint (sha256; a second install of the same build skips the upload), then installed and verified against the post-install inventory.
- **Platform-runtime owner (ADR-0019)**: facts + bound operations for app logs (durable `doublespeed-log-poller` descriptor with reattach/cleanup), network dump from app logs, app inventory, app state (foreground bundle id), lifecycle, deployment, and every interaction cell above. Unsupported cells are stated as facts with hints — back, alerts, clipboard, keyboard, settings, app switcher, gesture plans, port reverse, recording — and refuse at admission rather than mid-execution.
- **Connect verification**: a read-only list call; `401/403` → `UNAUTHORIZED` without echoing the key, `402` → a credits hint. Generated profiles store no secret; `DOUBLESPEED_DEVICE` (simulator model) and `DOUBLESPEED_API_URL` are read from the environment.

[...]

PR #2624, opened 2026-09-15
refactor(capture-kit): share the bounded app-log poller (#2616)

thymikee:
Refs #2616

Extracts the bounded app-log poller into `@agent-device/capture-kit/app-log-polling` and makes Limrun its first consumer. Deduplicates acquisition rollback, the poll state machine, the bounded read with its own settle-on-timeout, mark filtering, KMP tail-overlap de-duplication, finish memoization, disposal, and `cleanup-pending` semantics.

**Repurposed.** This PR previously based on the #2112 Doublespeed stack to convert both providers at once. That base predates the released `v0.21.x` wire ledger, so `Compatibility & Provenance` could never go green here (its "removing wire surface a released peer still sends … an ack cannot cover it" group is exactly the behind-a-release case). Doublespeed is not merged, so this now lands on `main` alone and Doublespeed's 30-line adoption follows #2112.

## Scope

- `packages/capture-kit/src/app-log-polling.ts` — the shared poller, published as the `./app-log-polling` subpath.
- `packages/provider-limrun/src/app-log-poller.ts` — 182 lines become a 32-line adapter. Provider identity stays where it belongs: `backendForReader` and the cleanup wording are adapter-local.
- `packages/provider-limrun/src/request-cancellation.ts` — `awaitLimrunOperation` loses `export`; this poller was its only external caller, and it is still used by the request-operation drain in the same module.

## Two deliberate narrowing decisions

- **Subpath export, not the façade.** `capture-kit/src/index.ts` falls under ADR-0019's merge-base NO-GROWTH rule, so an eager façade export would grow a locked closure. The subpath matches the existing `./ios-snapshot-acquisition` convention, and all 668 `eager-closure-budgets` cases pass with **no pin edit**.
- **`AppLogPollerReader.readLogs` takes no `signal`.** `main`'s Limrun reader is uncancelable and is currently called without one, so the shared contract forwards nothing and stays honest. The `AbortController` is internal: it drives `clock.sleep` and settles the bounded read on timeout, which is what makes teardown safe against a hung provider read. Widening the reader contract belongs with the first provider that can actually honour it.

The already-aborted path keeps the abandoned read's rejection handled, preserving `awaitLimrunOperation`'s behaviour rather than the leakier shape this branch had on the old base.

## Checks

`check:affected --run` passes. Locally: typecheck, `check:layering` (new subpath declared in the `package-boundaries` enumeration), `pnpm gate fallow --base <main>` → *no issues in 8 changed files*, `lint`, 18 app-log/request-cancellation tests, 668 eager-closure cases, and `daemon-wire-compat` → *178 declarations, 0 changed, 0 removed, 0 added* against `v0.21.6`.

## Follow-up

Doublespeed converts on top of #2112; #2616 stays open until both providers read one implementation.


thymikee:
Reviewed at e5f4096. This is already the smaller design: two ~200-line copies become one shared module and two ~20-line adapters, for -37 net production lines, and the extraction matches both providers by code comparison.

One question: could the Limrun adapter use `AppLogPollerReader` and drop its narrower `LimrunAppLogReader` type, or does it need to stay narrower? The two follow-ups the PR body lists (the `disposeAll` duplication and the `MARK_PREFIX` literal) are fine to leave for later.

Compatibility & Provenance fails in the daemon wire-compat gate on files the #2112 base stack touches (`http-server.ts`, `daemon-artifacts.ts` and others), none of which this diff touches, so it looks unrelated to this PR. Smoke Tests was still running.

Next: #2112 lands, then this retargets to `main` with the wire-compat gate green.

thymikee:
Answering the review at e5f4096 and the CI status.

**`LimrunAppLogReader` vs `AppLogPollerReader`** — it has to stay. `AppLogPollerReader` is only `{ readLogs, asyncDispose }`; `LimrunAppLogReader` additionally carries `platform`/`leaseId`/`instanceId`, which `app-log-runtime.ts` reads when building the descriptor and which `reconnect.ts`/`runtime.ts` construct. The adapter can't hand the poller those identity fields through the bare contract. The call-site `options.reader satisfies AppLogPollerReader` already states the subtype relationship without widening the provider type, and keeping `readLogs(appBundleId, lineLimit)` (no signal) is what encodes Limrun's uncancelable read at the type level — intersecting with `AppLogPollerReader` would launder the optional `signal` back in and erase that. So the `satisfies` at the boundary is the narrowest honest link; leaving as is.

**CI — `Compatibility & Provenance` (the only red gate).** Reproduced locally. It is entirely inherited from the stack base and cannot be greened in this diff:
- `git diff --name-only <base> HEAD` is exactly the 9 PR files; none are wire-surface files. The gate compares the tree's wire digests + `test/wire-compat/ledger.json` against the newest released tag (`v0.21.3`) — it ignores the PR base.
- The stack base (`8dad650`, #2112's fork head) is **not** a descendant of `v0.21.3` and is ~303 commits behind `main`. The flagged declarations (`LEASE_BACKENDS`, `daemonRuntimeSchema`, `SessionRuntimeHints`, `DaemonArtifactKnownType`, `http-server.ts`, `daemon-artifacts.ts`, `session-tenant-scope.ts`, `request-progress-protocol.ts`) are `main`-side wire changes the stale base simply predates.
- The gate's own output lists a "removing wire surface a released peer still sends … an ack cannot cover it" group. Being *behind a release* is exactly the case a `compatibleChanges` entry or a protocol bump cannot acknowledge — only a base at/after the release fixes it. #2112 itself doesn't touch any of those files, so this is `main`-vs-`v0.21.3` wire debt carried by the fork snapshot, not this PR's.

I checked the two ways to green it now and both mean doing #2112's own rebase (merge `origin/main` — or even `v0.21.3` — into the fork head): ~7–14 real conflicts including `package.json`/`pnpm-lock.yaml`, `snapshot.ts`, `cli-help.ts`, and `src/__tests__/eager-closure-budgets.ts` which **main deleted/restructured** (so my budget-pin edit lands at a dead path there too). That's #2112's landing conflict resolution, not work to fold into this PR.

Net: everything else is green on this head (Lint, Typecheck & Package, Repo Guards, Coverage, Integration, all Smoke, Bundle). This matches the plan in your note — retarget to `main` once #2112 lands; the wire-compat gate and the eager-closure budget pin both resolve then (Doublespeed hunks become pure edits, budget moves to main's restructured registry). Not marking merge-ready while base-only wire-compat is red.

The two follow-ups (`disposeAll` reuse, `MARK_PREFIX` shared with the daemon writer) stay out of scope as agreed.

thymikee:
Thanks, that answers the question. `LimrunAppLogReader` carries `platform`, `leaseId` and `instanceId`, and `backendForReader` and `app-log-runtime.ts` read them, so the narrower type has to stay. `satisfies AppLogPollerReader` at the call site is the right link. No code questions remain at e5f4096.

Compatibility & Provenance still fails on wire files from the #2112 base, which this diff does not touch. Next: #2112 lands, then this retargets to `main`.

thymikee:
Rebased onto `main` and repurposed, per review: `6322069` replaces (one no-op amend to re-trigger CI after the base change; tree unchanged) `e5f4096`, base is now `main`, and the two `packages/provider-doublespeed/**` files are out of this PR.

Supersedes my previous comment on the CI question: rather than document `Compatibility & Provenance` as base-only debt, the base is gone. `pnpm gate daemon-wire-compat` now reports *178 declarations, 0 changed, 0 removed, 0 added, 0 moved* against `v0.21.6`.

What the port changed relative to the old stack base, beyond dropping Doublespeed:

- Shared poller ships as the `./app-log-polling` **subpath** instead of a `capture-kit/src/index.ts` export. `main` replaced numeric closure pins with the ADR-0019 merge-base NO-GROWTH rule, so an eager façade export would grow a locked closure; the subpath keeps all 668 `eager-closure-budgets` cases green with no pin edit.
- `AppLogPollerReader.readLogs` no longer declares `signal`. `main`'s Limrun reader is uncancelable and is already called without one, so forwarding a signal was an affordance with no implementer. The internal `AbortController` still drives `clock.sleep` and the bounded read's settle-on-timeout. Doublespeed can widen the contract when it lands.
- Ported `awaitLimrunOperation`'s already-aborted path, which keeps the abandoned read's rejection handled; the shape on the old base could leave it unobserved.
- `awaitLimrunOperation` is no longer exported — this poller was its only external caller.

The `LimrunAppLogReader` question keeps its previous answer: it stays narrower, since `platform`/`leaseId`/`instanceId` are read by `app-log-runtime.ts`, `reconnect.ts`, and `runtime.ts`.

Locally green: typecheck, `check:layering`, `check:affected --run`, `gate fallow --base main` (no issues in 8 changed files), lint, 18 app-log/request-cancellation tests, 668 eager-closure cases.

thymikee:
Reviewed at 6322069. The move matches main: rollback, tail overlap, finish memoization and disposal in the new `app-log-polling.ts` are the Limrun code with named constants, and `readLogs` can drop `signal` because no Limrun reader takes one.

The retarget changes the size question from the first review. That review called this the smaller design because two poller copies became one. With the Doublespeed commit gone and #2112 still open, main has only the Limrun poller. So this PR now adds a public capture-kit subpath, a provider-neutral reader type and a second copy of the abort-settle logic, for about +60 production lines, and removes no duplication. Would it be smaller to keep the poller in provider-limrun and extract it in the PR that adds Doublespeed, when the second consumer shows what is really shared? If you prefer to land the extraction first, can [`settleOnAbort`](https://github.com/callstack/agent-device/blob/6322069/packages/capture-kit/src/app-log-polling.ts#L145) and the signal branch of [`awaitLimrunOperation`](https://github.com/callstack/agent-device/blob/6322069/packages/provider-limrun/src/request-cancellation.ts#L18) share one helper, so this PR does not add a copy?

Not blocking: the Limrun test keeps its own copies of the poller-host and deferred-sleep fixtures, and the title at [app-log-poller.test.ts#L66](https://github.com/callstack/agent-device/blob/6322069/packages/provider-limrun/src/app-log-poller.test.ts#L66) names a poller signal the reader no longer gets.

Checks were still running when I reviewed, and none had failed. There are no conflicts. Next: an answer to the one-consumer question.

thymikee:
Answer to the one-consumer question: **your size critique is correct, and I could not engineer around it cheaply.** Marking this draft rather than landing it.

**On net size.** Reproduced your number. Production-only (tests/fixtures excluded), this head is **+224 / −169 = net +55**: `app-log-polling.ts` +209, `app-log-poller.ts` +14/−168, and the `export` keyword on `awaitLimrunOperation`. With one consumer on `main`, that removes no duplication — it relocates it and adds a subpath plus a provider-neutral reader type. Deferring puts those +55 lines in the PR that also converts Doublespeed, where they pay for themselves.

**Why the two helpers cannot just share one.** I built it rather than guess: `settleOnAbort` moved to `@agent-device/contracts/async-lifecycle` next to `AsyncCleanupStack`, widened with the `onSettled` hook so it is the exact union of both shapes — Limrun's drain needs `onSettled` tied to the *source* finishing, which can be long after the caller stopped waiting. `awaitLimrunOperation` then delegates; 5 new cases in `async-lifecycle.test.ts` cover listener release, reason-vs-message precedence, and the already-aborted path. Net production falls to **+44** with no copy.

It fails ADR-0019's NO-GROWTH pin:

```
packages/provider-limrun/src/index.ts evaluates 30 modules on import; the merge-base evaluated 29.
  … → provider-limrun/src/runtime.ts → android.ts → request-cancellation.ts → contracts/src/async-lifecycle.ts
```

The structural reason: **provider-limrun's eager closure is 30 modules and none of them are foreign** — every cross-package import in its eagerly-evaluated subtree is type-only, and `request-cancellation.ts` imports nothing at runtime. So no shared runtime helper can be reached statically from there; only a dynamic `await import()` (the gate's own second suggestion) keeps it at 29, verified green at 670 eager-closure cases. That trades a startup metric for an import on every Limrun request operation. I don't think that's a better trade than waiting, so it is not what this PR ships.

Worth recording for whoever shapes the seam: the poller itself costs **zero** closure growth — `app-log-poller.ts` is reached lazily through `app-log-runtime.ts`, which is why this PR passes the pin at all. The placement was never the problem; the single consumer is.

**On your two non-blocking notes.** Both are fixed on the experiment branch: the third Limrun case was pure poller behaviour (bounded read never settles → `failed` → finish) already covered upstream, so deleting it removes the stale title naming a signal the reader no longer receives, and the adapter's 64 lines of hand-copied `pollerHost`/`deferredSleeps` become an import of the shared fixtures — following the existing `capture-kit/recording-artifact-fixtures` convention, which five packages already consume cross-package. That takes the Limrun test 156 → 68 lines. One more finding while testing it: for a real `AbortController`, `abort()` always installs a `DOMException` reason, so the `?? new Error(abortMessage)` fallback both helpers carry is only reachable from a signal that reports no reason. It is `main`'s shipped behaviour so I kept it, but it is now pinned by a test rather than assumed.

**Next.** Draft, no conflicts, `6322069` kept as the seed alongside `backup/2624-doublespeed-stack` (the both-provider version). Revisit when #2112's landing shape is known: if it lands, this rebases onto it and the extraction is justified by two readers with a seam both can satisfy; if it stalls, the smaller move is keeping the poller in `provider-limrun` and taking only what a single provider needs.

thymikee:
Agreed. With one consumer on `main`, the shared poller only relocates the code and adds a subpath, so keeping this on hold until the Doublespeed conversion gives it a second consumer is the right call. No open code questions at 6322069.

Issue #2614, opened 2026-09-15
refactor(settings): derive permission vocabulary from one declaration


thymikee:
# refactor(settings): derive permission vocabulary from one declaration

## Purpose and evidence

Adding `all` in PR #2363 (`e656ff158a5c7641564106c73ea4e9b787244b9f`) updates two `PermissionTarget` unions, a target array, parser comparisons, usage text and CLI membership. These independently maintained declarations make one vocabulary change span unrelated projections.

Evidence: [contract declarations](https://github.com/callstack/agent-device/blob/e656ff158a5c7641564106c73ea4e9b787244b9f/packages/contracts/src/settings.ts#L4), [public client type](https://github.com/callstack/agent-device/blob/e656ff158a5c7641564106c73ea4e9b787244b9f/packages/contracts/src/client-settings.ts#L5), [CLI validation](https://github.com/callstack/agent-device/blob/e656ff158a5c7641564106c73ea4e9b787244b9f/src/commands/capture/settings.ts#L158).

## Required behavior

1. In the existing settings contract, declare readonly collections for the current permission vocabularies and derive their union types. Preserve the current mobile/macOS subsets: a shared declaration must not widen what any surface accepts.
2. Derive parser membership, public client permission types, CLI membership and permission-name usage fragments from those collections. Public client derivation must use type-only imports: do not introduce runtime loading of `settings.ts` and its `AppError` dependency through the client type surface. Preserve existing exports, normalization, accepted values, errors and ordering of help text.
3. Keep platform-native permission mappings and device support checks in their existing owners. A name existing in the vocabulary does not imply platform support.

Illustrative declaration shape (names are not prescribed): `type MobilePermissionTarget = (typeof MOBILE_PERMISSION_TARGETS)[number]`; public client types may compose the existing mobile and macOS subsets.

## Scope and exclusions

Own `packages/contracts/src/settings.ts`, `client-settings.ts`, `src/commands/capture/settings.ts`, and their direct tests/consumers. No new package, registry, schema framework, compatibility re-export, platform behavior change or Maestro support-policy redesign. #2544 already tracks moving the Maestro adapter; that move is unrelated.

## Completion and validation

- Characterize the existing accepted/rejected names per surface, case/whitespace normalization and error/help strings before refactoring. Use fixed expected data, not expectations derived from the new collection.
- Both exported types retain their current assignability, including the macOS-only names where currently accepted.
[...]

PR #2625, opened 2026-09-15
refactor(settings): derive the permission vocabulary from one declaration

thymikee:
## Summary

`packages/contracts/src/settings.ts` declares the `settings permission` vocabulary once — actions, modes, app-scoped targets, macOS targets — and the two parsers, the permission fragments of `settings` help and its invalid-args message, the public client permission types (via `import type`, so `AppError` stays off the client's runtime path) and the CLI's membership sets are built from it.

Deleted: two `PermissionTarget` unions, a module-local target array, 16 parser comparisons, two hand-written usage fragments, the CLI's 19 literal permission names, a duplicated `PermissionState` alias, and the macOS fragment restated in `cliDetail`. Accepted names, case and whitespace normalization, error strings, help ordering and daemon positionals are unchanged; the mobile parser still refuses the macOS-only names, so nothing widened. Actions and modes are derived too — same duplication class, same owned files. The macOS `<grant|reset>` stays a literal: which actions macOS serves is a support fact owned by `platform-apple`.

Part of #2614. #2363 is still open, so this adds no `all`; when it lands, `all` is one name in `MOBILE_PERMISSION_TARGETS` plus the fixed test literals.

## Validation

`3aa6b46fa0`: `pnpm check:affected --run` green — 3017 tests, with format, lint, typecheck, layering, fallow and build. `check:production-exports` and `check:command-docs` report nothing in the changed files.

Characterization landed first, from fixed literals duplicated per test file so one edit cannot silence every witness. Issue replay: adding `face-time` to `MOBILE_PERMISSION_TARGETS` alone moved the parser, CLI membership, help fragment, `cliDetail` and the client type, and the fixed pins then failed (three runtime, two `toEqualTypeOf`) rather than accepting it silently. Removed before the commit. A planted value import of `./settings.ts` in `client-settings.ts` fails its canary.

Coverage, Integration and the device replay lanes stay CI-authoritative on this head; no device evidence is claimed, as no platform or runner code changed.

Rebased onto main, which is itself red at `2cafab3ad0`: `installArgs` in `snapshot-helper-install.test.ts:29` no longer exists on `AndroidSnapshotHelperManifest` (#2619 copied the fixture after #2618 deleted the field). The second commit drops that stale field — nothing parses it, and the leniency case stays covered in `snapshot-helper.test.ts:913`. It is outside the #2614 family and can be split out; without it the typecheck gate here cannot run.


thymikee:
Reviewed at ac6fc28. The code looks good: the permission vocabulary comes from one declaration, and the parsers, help fragments and client types derive from it without changing accepted names or errors. Smoke Tests runs `01-settings.ad` on real Android and iOS targets, and 3 of 4 matrix jobs were green at review time.

PR #2627, opened 2026-09-15
test(android): read the install-timeout fixture from the helper manifest fixture

thymikee:
## Summary

`pnpm typecheck` is red on `main` at `2cafab3ad0`, so **Typecheck & Package** fails for every open PR. #2618 removed `installArgs` from `AndroidSnapshotHelperManifest` and updated the shared helper fixture, the packaging script and the parse test — but `snapshot-helper-install.test.ts` kept a second hand-written manifest literal with the field still in it, which is an excess property on a type that no longer declares it.

Both tests in that file re-hash the manifest for the APK they write, so the identity fields carried no test intent. They now come from `ANDROID_SNAPSHOT_HELPER_FIXTURE_ARTIFACT` the way every other helper test reads it, which is also the reason a second copy cannot drift from the type again. The one case that needs a manifest carrying `installArgs` is the parse path that ignores it, and that case already spells the key inline in `snapshot-helper.test.ts:913`.

One file, +4/-15, test-only. This unblocks the gate for the open PRs, including #2607 and #2610.

## Validation

Tested at `f737dee6c6`.

- `pnpm check:affected --run` green: `format:check`, `lint`, `tsc -b` across all 24 projects (the check that is red on `main`), `check:layering`, `check:fallow --base origin/main`, `vitest related`.
- `snapshot-helper-install.test.ts` and `snapshot-helper.test.ts`: 26 tests passed.
- No production code or behaviour touched, so no runtime or device validation applies.


thymikee:
Closed: #2625 landed the same fix on main (`snapshot-helper-install.test.ts` no longer carries `installArgs`) and **Typecheck & Package** is green at `aba54b4cc6`, so there is nothing left for this PR to unblock.

PR #2631, opened 2026-09-15
feat(record): export a recording from a copy of what its recorder wrote

thymikee:
Stacked on #2613. Implements the write path of ADR 0024 2.3 for the first two backends.

## Summary

A stop now collects what its recorder wrote and produces the export from that copy, so the caller's path is never a file a recorder is still writing, and a stop that dies partway is driven again from the copy instead of re-signalling a recorder that already exited.

- `ScreenRecordingOutputHost` gains the four local steps a copy needs: `collectFromRecorder`, `writeExportFromCollected`, `retireRecorderFile`, `discardCollectedFile`. Only the host can see a path, so `retireRecorderFile` is what answers `retired` rather than `retirable`. `ScreenRecordingFinalizer.validatePlayable` exposes the playability rule for a collected copy.
- iOS Simulator: `simctl` records to `<name>.native.mp4`; the stop copies to `.collected.mp4`, checks playability there, writes the caller's path once, then retires the recorder's file.
- Android: chunks are pulled to `.collected` siblings and copied out; chunk and `--client-output-path` naming is unchanged. The 180s-limit disclosure now rides the stop's journal, so a retried stop keeps it.
- 44 files, 707 net production lines — past the size note. Android can come out as its own layer if you want it.

## Validation

`pnpm check:affected --run` and `pnpm check:tooling` pass at `17e7a67961` (restacked on #2613 `c69438a5fb`, which now also fixes the Android presence probe this branch's dispositions read).

Live, on the current head: `record start`/`record stop` on iPhone 17 simulator and `emulator-5554` each produced a playable MP4 with `recorder: confirmed`, `nativePathDisposition: retired`, and no `.native`/`.collected` sibling left on host or device; a stop whose daemon was killed after the export answered `retired` from the device, not from the marker it had frozen.

Mutations each fail a named test: recorder writing the export, disposition hardcoded, collect skipping playability, marker published after disposal.


thymikee:
Reviewed at e2ad807, stacked on #2613. Exporting from a collected copy is the right idea, but a retried stop can lose data, and the design is larger than it needs to be.

The finalization checkpoint drops fields. `encodeFinalization` at [stop-checkpoints.ts#L62](https://github.com/callstack/agent-device/blob/e2ad807/packages/capture-kit/src/recording/stop-checkpoints.ts#L62) does not carry `chunks` or `capturedDurationMs`. On Android, if `persistNativeManifest` or `cleanupChunks` throws after that checkpoint, the retry takes the `learned.finalization` branch ([stop-sequence.ts#L91](https://github.com/callstack/agent-device/blob/e2ad807/packages/capture-kit/src/recording/stop-sequence.ts#L91)) without them, so a chunked recording over 180 s returns only `outPath` and silently loses part-002 onward. Can the codec derive its keys from `ScreenRecordingFinalization`, so every served field round-trips? A stop-sequence test that replays a journaled chunked finalization and asserts `result.chunks` would catch this.

The stop clock is not journaled. `finalizeAndroidRecording` reads `stoppedAtMs = Date.now()` on every call ([finalize.ts#L63](https://github.com/callstack/agent-device/blob/e2ad807/packages/platform-android/src/recording/finalize.ts#L63)). A retry that resumes with `collectedPath` journaled skips the stop, so `exportCollectedChunks` measures up to the retry time and can report an idle-tail warning sized by the gap between attempts. Journaling `stoppedAtMs` with the stop observation would fix this.

Could an abandoned stop leave full-size files behind? `cleanupAppleRecording` and `cleanupSimulator` in recovery.ts do not appear to remove the `.native.mp4` or `.collected.mp4` siblings ([runtime.ts#L104](https://github.com/callstack/agent-device/blob/e2ad807/packages/platform-apple/src/recording/runtime.ts#L104)), and restart reattach is unreattachable, so session close, daemon restart or forced cleanup after a failed stop would leave them next to `--out`. Should every terminal path dispose of both files, or should the CHANGELOG say leftovers are expected?

Is playability now checked up to three times on the same bytes? `collect` validates the copy in completion.ts, `finalize.complete` validates the byte-identical export again, and on Android `pullChunks` already requires each chunk playable. That adds waits and a validator spawn to every iOS and Android `record stop`. Could validation happen once, on the collected copy?

The chunk-naming rule now has two owners: `chunkFilePairs` at [finalize.ts#L249](https://github.com/callstack/agent-device/blob/e2ad807/packages/platform-android/src/recording/finalize.ts#L249) re-implements what `pullChunks` ([chunks.ts#L118](https://github.com/callstack/agent-device/blob/e2ad807/packages/platform-android/src/recording/chunks.ts#L118)) owns, while `pullChunks`' `clientOutputPath` parameter has no production caller left. Can one of them own the rule and the dead branch go? Smaller cleanups: duplicate JSDoc above `chunkFilePairs` (L239, L244), the repeated comment in [runtime.fixtures.ts#L100-L108](https://github.com/callstack/agent-device/blob/e2ad807/packages/platform-apple/src/recording/runtime.fixtures.ts#L100-L108), a re-export between imports at [stop-sequence.ts#L13](https://github.com/callstack/agent-device/blob/e2ad807/packages/capture-kit/src/recording/stop-sequence.ts#L13), and hand-written output stubs in the HarmonyOS and web fixtures instead of `recordingFileStore().outputs`.

Resume is the headline behavior, but `createStopProgress` ([transitions.ts#L305](https://github.com/callstack/agent-device/blob/e2ad807/packages/capture-kit/src/durable-capture/transitions.ts#L305)) has no test, and every resume test hand-builds `DurableCaptureProgress`. Can one test run a failing `record stop` and then a second one through the real fence and lease, asserting one SIGINT, one collect copy, and the served result?

At about 787 net production lines, could the host interface be two steps instead of four: `collect` that copies and validates once, and `publish` that renames the collected copy onto the export, removes the recorder file and returns the disposition? That would remove `discardCollectedFile` and the second playability check, and Android could land as its own PR. Does ADR 0024 2.3 first need to say whether `finalize.complete` still owns playability, and whether the collected copy may be renamed into the export on the same volume?

Smoke Tests fails while waiting for the fixture home screen in the iOS input scenario. That scenario runs no `record` command and does not reach the recording routes this diff changes, so it looks unrelated.

A live resume run is still needed on an iOS simulator and an Android emulator: make the first `record stop` fail after the collect checkpoint (for example an `adb` failure during `persistNativeManifest` on a recording over 180 s), then run a second `record stop`. It should show no second SIGINT in `--debug` output, `recorder: confirmed` with the right `nativePathDisposition`, all chunks and `capturedDurationMs` on Android, and no `.native.mp4` or `.collected.mp4` left after an abandoned stop. The happy-path runs in the PR body do not reach this route. Next: fix the checkpoint round-trip and the stop clock, then add the resume test and live run.

thymikee:
Restacked on #2613 after its probe fix; new head `17e7a67961`, `check:affected --run` green, Validation above is the current head.

Answering the size report directly: +6.3 kB unpacked is over the 3 kB note, and 707 net production lines is over the size guideline, so say the word and I'll cut Android (`c6f3af2731`, plus its scenarios) into its own layer under this one — it is a separate backend and shares only the stop sequence and the output host verbs. Left as one PR otherwise, since the copy rule reads better with both backends beside it.

thymikee:
Answered the [review at e2ad807](https://github.com/callstack/agent-device/pull/2631#issuecomment-5680258767) in `b64efa7`, on top of `17e7a67`.

**Checkpoint round-trip.** `ScreenRecordingFinalization` now lives in `@agent-device/contracts/recording-stop-progress`, and the codec in `stop-checkpoints.ts` has one reader and one writer per field, keyed over every field of that type. A new field does not compile until it round-trips. `chunks` and `capturedDurationMs` are journaled; a finalization whose chunks or length cannot be read is redone, not served short. Tests: `serves every field a journaled finalization carries` (sequence) and `a retried commit serves every chunk and the captured length its first attempt finalized` (Android, the completion marker write fails once). Making the chunk and length readers return nothing fails the Android test with `expected undefined to deeply equal [...]`.

**Stop clock.** The sequence reads the clock before the signal and journals `stoppedAtMs` with the observation; `finalize` gets the journaled value. Android no longer calls `Date.now()`. Test: `a stop resumed after its collect measures the window its first attempt signalled`; reading `now()` instead of the journaled value fails it with the idle-tail warning.

**Leftover siblings.** Kept, and the CHANGELOG now says so. After an abandoned stop the `.native`/`.collected` files are the only copy of the video (the recorder used to leave it at `--out`), so deleting them on session close would lose the recording.

**Playability.** Checked once, on the export. `validatePlayable` is gone from the finalizer contract, the host and both `collect` steps. simctl has exited before `collect`, so the copy is final bytes, and Android's pull already retries until each chunk plays.

**Host surface.** `collectFromRecorder`, `writeExportFromCollected`, `retireRecorderFile` and `discardCollectedFile` are now two verbs, `copy({ from, to })` and `remove(path)`. The HarmonyOS and web fixtures use `recordingFileStore().outputs` instead of throwing stubs.

**Why not rename the collected copy into the export.** The overlay is not idempotent. If the copy is renamed and the stop dies after the overlay but before the finalization checkpoint, the retry cannot tell a raw export from an overlaid one and would overlay twice or serve a raw file. The copy is what makes that step safe to redo.

**Chunk naming.** One rule, `chunkPathAt` in `recording/chunk-path.ts`, used by `pullChunks`, the served/collected pairs and the completion-marker check in `manifest-validation.ts` (a third copy). `pullChunks` lost its unused client-path branch. The duplicate JSDoc, the repeated fixture comment and the re-export between imports are fixed.

**Resume through the fence.** `a record stop that fails after collecting resumes through the fence without a second signal` runs the real sequence behind the real live handle through `finishLiveScreenRecording` twice: one signal, one collect, the checkpoints on the open manifest after the failure, and the served result after the retry. Making `createStopProgress().record` a no-op fails it.

**Size.** Net production lines go from about 696 to 625 against #2613.

**Live runs** on a build whose source differs from `b64efa7` only in one integration-test regex (dedicated iOS 26.2 simulator and a `Pixel_7_review` emulator; the first stop was made to fail after collect by putting a directory at `--out`):
- iOS: the first `record stop` failed with `EISDIR` on the export copy, and the manifest stayed `open` with `stopObservation: confirmed`, `stoppedAtMs: 1789481017533` and `collectedPath`. No `recordVideo` process was left. After removing the directory, the second stop returned `recorder: confirmed`, `nativePathDisposition: retired`, and a 13.3 s playable `ios.mp4`. Both siblings were gone, and the completed manifest still held `stoppedAtMs: 1789481017533`, so the recorder was not signalled again.
- Android, 197 s recording: the first stop failed the same way, with both collected chunks on the host and both chunks still on the device. 60 s later the second stop returned `chunks` `and.mp4` (188.6 s) and `and.part-002.mp4` (47.9 s), `capturedDurationMs: 236566`, `recorder: confirmed` and `nativePathDisposition: retired`. The device chunks and the collected siblings were gone, and `stoppedAtMs` was unchanged. There was no idle-tail warning, although the retry came 60 s after the signal.

A failure after the finalization checkpoint (the completion marker write) cannot be provoked on demand without fault injection, so that route is covered by the Android unit test above, not by a device run.

Gate at `b64efa7`: `pnpm check:affected --run` passed format, lint, typecheck, layering, fallow, build, package, the node and macOS integration lanes, vitest-related (3,410 tests) and the model gates. `scripts/node-test-tmpdir.test.ts` timed out once on a busy host and passed on rerun; this PR does not touch it. The remaining gates after it (production-exports, mutation-model and the others) pass when run one by one.

thymikee:
Follow-up in `e41d66e`, after checking `b64efa7` against ADR 0024 2.3 and an independent read of it.

**The caller's path never holds a refused export.** In `b64efa7`, the export was copied to `--out` before `complete` checked it, so a killed simctl left an unplayable file at `--out` next to a failed stop. When the export copy or `complete` fails, both backends now remove the paths they served before the error returns, and the collected copy stays for the retry. Tests: `a simulator export the finalizer refuses leaves the caller path empty and the copy for the retry` and the chunked Android version. Each fails if its removal is dropped.

**The collected checkpoint follows the sniff again.** ADR 0024 2.3 records `collected` only after the copy passes the container sniff. `b64efa7` dropped the check from the simulator collect along with the full validator. The new `finalize.sniff` checks `ftyp` + `moov` (EBML for WebM) without spawning the validator, so the export still gets the full check exactly once. Android's pull already checks each chunk.

**The copy is kept until its finalization is journaled.** Removing it inside `finalize` left a crash window in which the checkpoint named a copy that was gone, so every later stop would fail on it. A `discard` step now runs after the finalization is journaled, and again on a replay. Test: `discards the collected copy only once its finalization is journaled, on a replay too`. It fails if `discard` runs before the journal write.

**Live, on `e41d66e`**, with the same injection as before (a directory at `--out` makes the first stop fail after collect):
- iOS: the first stop failed with `EISDIR`. The manifest was `open` with `stopObservation: confirmed`, `stoppedAtMs: 1789482459123` and `collectedPath`, and no recorder was alive. The second stop returned `recorder: confirmed`, `nativePathDisposition: retired` and a 20.7 s playable `ios.mp4`. The siblings were gone, and `stoppedAtMs` was unchanged in the completed manifest.
- Android, 197 s: the first stop failed after collecting both chunks. The second stop, 60 s later, returned `and.mp4` (215.6 s) and `and.part-002.mp4` (27.3 s), `capturedDurationMs: 242880`, `recorder: confirmed` and `nativePathDisposition: retired`. Device chunks and collected siblings were gone, `stoppedAtMs` was unchanged, and there was no idle-tail warning.

Gate at `e41d66e`: `pnpm check:affected --run` passed end to end.

thymikee:
Reviewed at e41d66e, the head that merged, as a follow-up to the [review at e2ad807](https://github.com/callstack/agent-device/pull/2631#issuecomment-5680258767). The earlier points are fixed: the finalization checkpoint round-trips `chunks` and `capturedDurationMs`, `stoppedAtMs` is journaled, the collected copy stays until its finalization is journaled, playability is checked once, and a refused export leaves nothing at `--out`. One defect remains and is now on main.

An abandoned stop can still lose its video. After an abandoned iOS stop, the video exists only in `<out>.native.mp4` and `<out>.collected.mp4`. A later `record start --out <same path>` runs `outputs.prepare(nativePath)`, and its forced `rmSync` deletes the kept native file before simctl starts. That recording's stop then copies over the kept `.collected.mp4` ([runtime.ts#L105](https://github.com/callstack/agent-device/blob/e41d66e/packages/platform-apple/src/recording/runtime.ts#L105), [#L166](https://github.com/callstack/agent-device/blob/e41d66e/packages/platform-apple/src/recording/runtime.ts#L166)). On Android, `pullChunks` in `chunks.ts` overwrites the kept collected chunks the same way. The CHANGELOG says these files are the only copy of the video, so nothing deletes them. A user who records again to the same `--out` after a failed stop silently loses the earlier recording. Should the start path refuse with a typed error that names the leftover sibling when no open recording owns it? If losing it is intended, the CHANGELOG should say that a later recording to the same `--out` replaces the kept files.

Do the refused-export tests guard the retry route? The iOS test in [runtime.test.ts#L287](https://github.com/callstack/agent-device/blob/e41d66e/packages/platform-apple/src/recording/runtime.test.ts#L287) and the Android one in `finalize.test.ts` stub `complete` with a plain `Error` and match its message. Production throws `COMMAND_FAILED` with reason `recording-output-unplayable` and `retriable: true`. A change that drops `retriable` would still pass. Throwing the production error from the fixture and asserting those fields would catch it.

On size, about 749 net production lines are over the guideline. Most of it is the Android move, and the copy, complete and remove-on-refusal blocks in `completion.ts` and `finalize.ts` have the same shape. Could a follow-up share one "export from collected" helper in capture-kit?

All checks passed at this head. The live resume runs in the PR body cover the copy-failure route; a finalizer refusal of a real unplayable export has unit coverage only. Next: a small follow-up that makes a new recording refuse, or document replacing, an abandoned recording's kept files.

PR #2632, opened 2026-09-15
refactor(android): separate adb transport addressing from the command payload

thymikee:
## Summary

Separates ADB transport addressing from the device-command payload, the production path traced in #2617. An invocation carries `{ target: { selector, server, waitFor?, hostGlobals? }, command, rawArgv? }`; flat argv comes from one serializer, and the argv grammar — option arity, the 20 `wait-for[-TRANSPORT][-STATE]` forms, managed-forbidden commands — is one table read by parser, policy and tests alike. `command` is appended and never re-parsed; a request nobody rewrote is emitted as the argv it arrived in.

Four rules are now explicit:

- A private adb server travels in the addressing that owns it. One rule covers both channels, argv `-P` or `serverPort` option: name no port and the transport uses its own, name this one and it is accepted, name another and the call is refused with `managed-device-transport-mismatch` **before dispatch**. Main rewrote a conflicting `-P` onto the transport's port, which let a caller read a zero exit as an answer about the server it had named; this tightening is intentional. Ambient adb is untouched — with no private server named, argv runs as typed.
- Under a private server, globals that transport cannot restate (`-t`, `-H`, `-a`, `-d`, a second `-s`, an unparseable `wait-for-…`) are refused, not answered with addressing dropped.
- A forwarding provider receives the caller's argv minus the scope's own `-s` pair, so readiness tokens survive the hop. It is checked against the same server rule, because a provider would otherwise hand the caller's `-P` to an adb it does not address.
- One typed refusal answers every mismatch: device, target, or server.

`provider-limrun` drops its structural type copy and its own argv projection, and hands an ADB failure the invocation it addressed. The root host lowers a typed request in one function, which the test host now shares instead of imitating. The provider builds no `AndroidAdbTarget` at all: the composition root hands it `deviceAdbInvocation` and `hostAdbInvocation` built by `androidAdbInvocation` over `androidAdbSerialTarget` and the new `androidAdbHostTarget`, because ADR-0019 keeps a provider's eager closure off the mechanics facet.

Route table and cost: https://github.com/callstack/agent-device/issues/2617#issuecomment-5681630852. 27 files; gross churn exceeds the 1,000-line budget: one contract moves across platform, provider and root binding. Not closing #2617; the published-options change stays in #2640.

## Validation

`0688fd0e7c` on `origin/main` `9173d3c797` (0.21.4), with #2599, #2604, #2629, #2636 and #2639 integrated. `pnpm test:unit` (1326 files / 10345 tests), provider-integration (66 files / 210 tests), typecheck, lint, layering (243 pass / 0 fail) and package gates pass locally. CI at head is green apart from one `Smoke Tests` red on `RunnerTests.testAlertCannotProveAnIdenticalReplacementAndDoesNotActivateIt`, which arrived with `27a97ee619 fix(ios): confirm alerts without repeating activation (#2326)`, touches no Apple code here, and passed on `main` at `fffa01e0`; the rerun of that job passed on the same head, so the red was environmental. `check:affected`'s `mutation-model` ownership assertion fails identically at `origin/main` here.

Refusals are pinned at the process boundary, not just at the thrown error: the fake adb appends every call it actually serves to a log, and the refused requests appear nowhere in it. Deleting the server check from `requireManagedAndroidAdbAddressing` fails four tests across the provider branch, the device route, the reachability run, and the root host. The user-visible part of the tightening is in `CHANGELOG.md`, including that ambient adb is unchanged.

Live at head, with helper assets built for 0.21.4 (`pnpm build:android`; snapshot and IME APKs in `android/*/dist`) on `emulator-5554`: `open` returning interactive `@refs`, `press @e18 --settle` settling +18/−10, `close`. (Pixel 7 CI, SDK 36) through the built CLI: `devices`, `open com.callstack.agentdevicelab --foreground` returning an interactive snapshot with `@refs`, `press @e18 --settle` settling with a +18/−10 diff onto the Settings screen, `snapshot -i`, `apps`, `keyboard status`, `screenshot` (1080x2400), `logs start`/`mark`/`stop` (15 KB), `close`.

Live against a real second adb server on port 15037 (`adb -P 15037 start-server`, separate daemon state dir): `createLocalAndroidAdbProvider(device, { serverPort: 15037 })` answered `shell id` (`uid=2000(shell)…`) and `getprop ro.serialno` (`EMULATOR36X2X12X0`); while a call was in flight the OS process table showed exactly `adb -P 15037 -s emulator-5554 shell sleep 3`, which is external proof of the lowered argv; a background `logcat` spawn through the same route returned 961 lines; `exec(['-P','9999','shell','id'])` threw `managed-device-transport-mismatch` and polling `lsof -iTCP:9999 -sTCP:LISTEN` for two seconds showed no server ever forked there, so nothing dispatched; the ambient host route kept working unchanged. Driving the whole CLI with adb's server on 15037 (`ANDROID_ADB_SERVER_PORT=15037`) also completed `devices`, `open`, `press --settle`, `screenshot`, `logs start`/`stop`, `close`.

Live in process, on the two shapes the review asked to see, against a real adb server on 15037 and the real emulator:

- **Private-port scope** (`withAndroidAdbProvider(provider, { serial, serverPort: 15037 })`, the shape a lease installs): `runAndroidHostAdb(['-s', serial, 'shell', 'id'])` answered `uid=2000(shell)…`; with a call in flight the process table read `adb -P 15037 -s emulator-5554 shell sleep 3`; a server-level `devices` was answered by the lease transport; `-P 9999` threw `managed-device-transport-mismatch` and added zero calls to the provider; a forced failure rendered `adb -s emulator-5554 shell definitely-not-a-command exited with code 127`.
- **Ambient provider scope** (the shape a forwarding provider installs, which is Limrun's): the provider received `[["shell","id"], ["wait-for-device","shell","getprop","ro.serialno"], ["shell","definitely-not-a-command"]]` — the scope's `-s` pair removed, the readiness token preserved — while real adb answered underneath (`uid=2000(shell)`, `EMULATOR36X2X12X0`), and the forced failure surfaced through the same addressed shape.

Worth naming while doing that run: under a port scope, `runAndroidHostAdb` is answered by the lease transport, so a forwarding provider installed alongside a private server receives nothing on that channel; it is reached through the executor-override channel instead, where the same server rule applies. Pre-existing routing, unchanged here, and no shipped combination uses both.

Live on a real Limrun Android instance (`connect limrun --platform android --session apex-2632-lane --force`), at head: the uploaded-asset catalog listed; `open "com.callstack.agentdevicelab.apk" --foreground` leased in 6.9s and opened in 13.1s into an interactive snapshot (9 visible / 29 total nodes, `@refs`); `snapshot -i` green; `press @e2 --settle` settled `+0 -0 (~29 unchanged)` and `press @e26 --settle` navigated with `+33 -24 (~5 unchanged)`; `logs start`/`stop` captured 1273 lines, 38 naming the package, so logcat ran over the tunnel through the changed reconnect/poller path; `close` then `disconnect` released the session. The session request log records the adb commands the provider served — `["-s","127.0.0.1:65049","shell","am","instrument","-w",…]`, `["-s","127.0.0.1:65049","reverse","--list"]`, and cleanup as `["disconnect","127.0.0.1:65049"]`, a device-less server command. A forced adb failure was injected by dropping that tunnel serial mid-session: `keyboard status` returned `COMMAND_FAILED` / `device '127.0.0.1:65439' not found` with `exitCode: 1`, the hint intact, and the request log shows the addressed command `["-s","127.0.0.1:65439","shell","dumpsys","input_method"]`. `APP_NOT_INSTALLED` for a missing package and the service's `Given archive does not contain AndroidManifest.xml` for a truncated archive were also observed, both answered before any adb command. See the evidence comment for the full lane.

## Blocked validation, with what each needs

- **Simlock-granted lease.** No lease can be granted from the shipped CLI in 0.21.3: `agent-device host` does not exist in this build (the ADR 0021 front-end is unshipped), and `connect --lease-id <id> --lease-backend android-instance` needs a Host plus co-resident Simlock to mint one and hold an Android slot. Per review, the in-process path stands in, and it now runs both ways: the reachability test drives a real fake-binary adb through the lease port and asserts the refused request appears in no dispatch log, and the private-port scope above runs against a real adb server. What still stands unverified is the allocator's grant, renewal and release around that transport.


thymikee:
Reviewed at 1dc9dea. One contract for the invocation is a good direction, but a leased executor may now accept a caller's server port, provider-scoped calls may lose addressing, and the PR closes an investigation issue without its output.

On main, `createSerialAdbExecutor` and `createSerialAdbSpawner` replace `options.serverPort` with the lease port. On head the caller's options pass through ([adb-provider-scope.ts#L62](https://github.com/callstack/agent-device/blob/1dc9dea/packages/platform-android/src/adb-provider-scope.ts#L62)), and the root lowering gives them priority: `serverPort ?? androidAdbOwnedServerPort(invocation)` ([platform-runtime-android-adb-host.ts#L144](https://github.com/callstack/agent-device/blob/1dc9dea/src/platform-runtime-android-adb-host.ts#L144)). Can a caller that passes `serverPort` in `AndroidAdbExecutorOptions` move a managed lease off its private adb server this way? Main closed this for the `-P` argv spelling. If so, only the owned port should reach lowering under a lease, either by stripping `serverPort` from options or by refusing a mismatch, with a test that passes `serverPort: 9999` to a leased executor and spawner.

The provider serial branch now passes `invocation.command` ([adb-provider-scope.ts#L222](https://github.com/callstack/agent-device/blob/1dc9dea/packages/platform-android/src/adb-provider-scope.ts#L222)), the tail after all parsed addressing. On main, `stripAdbSerialArgs` removed only the matching `-s` pair, so `adb -s S wait-for-device shell getprop` reached the provider with its `wait-for-device`. Does head drop that argument, and globals such as `-d` or `-e` without a lease, so a probe could run before the device is ready? The provider should get the caller's argv minus the scope's own `-s` pair, or a refusal, and a scope test with a wait-for argument would cover it.

Managed argv is not byte-identical to main for a few edge forms: a repeated `-s` with the same serial is pushed into `hostGlobals` and refused ([adb-transport.ts#L414](https://github.com/callstack/agent-device/blob/1dc9dea/packages/platform-android/src/adb-transport.ts#L414)), a second wait-for argument overwrites the first, and a non-integer `-P` is now refused. No internal production caller builds these forms, so either restore parity or state the change in the PR body.

`limrunAdbArgv` ([provider-limrun/src/android.ts#L195](https://github.com/callstack/agent-device/blob/1dc9dea/packages/provider-limrun/src/android.ts#L195)) re-implements `serializeAndroidAdbInvocation` for failure text without `rawArgv`, `hostGlobals` or `waitFor`, the limrun target builders repeat `androidAdbInvocation(androidAdbSerialTarget(serial), args)`, and app-log-reconnect.ts writes the same target literals a third time. That is a second serializer to keep in sync, which the PR set out to remove. Can the failure text come from the real serializer, and the target builders be shared? Smaller cleanups: the back-to-back JSDoc blocks at [adb-transport.ts#L212](https://github.com/callstack/agent-device/blob/1dc9dea/packages/platform-android/src/adb-transport.ts#L212), `androidHostAdbInvocation` as a pure alias of `parseAndroidAdbArgv` (L341), and `const requested = installed;` at [adb-provider-scope.ts#L287](https://github.com/callstack/agent-device/blob/1dc9dea/packages/platform-android/src/adb-provider-scope.ts#L287).

#2617 asks for a design trace, a route table and a continue/stop recommendation before any migration, and says the issue should not close as an implementation. Can "Closes #2617" be removed, the route table and recommendation posted on the issue, and the body corrected where it claims a single serializer and no re-slicing?

Would a smaller change do: keep the three host methods and flat argv, and move the three hand-written grammars (`findAdbSerialIndex`, `withServerPort`, `assertManagedAdbCommand`) into one table and parse helper in adb-transport.ts? That removes the duplicate parsing without new invocation types, `rawArgv`, the options/target double channel for the server port, or a limrun dependency on platform-android. If typed addressing stays, should the server port travel only on the target, with `AndroidAdbExecutorOptions.serverPort` removed from the device routes first?

All checks pass at 1dc9dea and there are no conflicts. A live run is still needed on the routes this changes: a managed-lease Android session on a private adb server whose `--debug` output shows `adb -P <lease> -s <serial> ...`, and a limrun-provider session showing `adb -s <tunnel-serial>` for exec and disconnect on cleanup. The run in the PR body covers only the ambient local route. Next: settle the lease port question and the provider argv, then add those runs.

thymikee:
Addressed in `25a8564cab` (platform + root lowering), `9cbb59c3fd` (provider), `8af64ef06e` (comment move).

**Lease port.** Yes, a caller could move a leased route: options passed through and root lowering preferred them. `serverPort` is now destructured out of both device routes and reconciled, not forwarded — `deviceServerPort` refuses a construction or per-call port that differs from the lease with `managed-device-transport-mismatch`, and without a lease the built-with port wins, as on main. Root lowering takes the single channel through `requireAndroidAdbServerPort`, so an owned port can never be moved by an option; scoped routes carry the lease port in the target and stop setting `options.serverPort`, which is what makes the second channel unrepresentable rather than merely unused. Pinned by a test that passes `serverPort: 9999` to a leased executor and spawner, and one that pins the no-lease precedence.

**Provider argv.** It did drop `wait-for-device`, and `-d`/`-e` with it. The serial branch now hands over `androidAdbPayloadWithoutSerial(args, scope.serial)` — the caller's argv with only this scope's `-s` pair removed, byte-order kept, no rebuild. Under a lease, globals the private transport cannot restate are refused; without one they travel to the provider, which is what its adb can honor. Scope tests cover the wait-for case, the no-lease `-d` case, and that a server-level command stays on the host route.

**Grammar parity.** Restored where adb itself is preserved, stated where it is not. A repeated `-s` naming the same serial is now dropped as redundant instead of refused. A second `wait-for` token keeps the first in the typed slot and the rest as host globals, so a managed route refuses rather than answering two waits with one. Main forwarded both verbatim on every route; head keeps that on the ambient route and refuses it on a managed one. A malformed `-P` stays an unowned global: the ambient route emits the caller's argv unchanged and adb decides, as on main; a managed route refuses it along with other unrestatable globals. The body states the rule change. Each form has a parse test.

**Second serializer.** `limrunAdbArgv` is gone. `adbError` now takes the `AndroidAdbInvocation` and the root adapter renders `command` with `serializeAndroidAdbInvocation`, asserted in `src/sdk/limrun-runtime-dependencies.test.ts`. Target builders are shared and used by `app-log-reconnect.ts`. I first put them in `provider-limrun/src/adb-addressing.ts`; that added one eagerly evaluated module to `provider-limrun/index`, which ADR-0019's no-growth rule refuses for an entry the merge base carries, so they live in `android.ts`, inside the closure both callers already sit in. `androidHostAdbInvocation` is deleted, `const requested = installed` went with `scopedServerPort`, and the transcription note is now the table's own doc block.

**#2617.** `Closes` removed, route table plus continue recommendation posted (https://github.com/callstack/agent-device/issues/2617#issuecomment-5681630852), and the two overclaims corrected: the parse direction does re-slice once — that is the only direction that flattens — and the provider no longer keeps a projection of its own.

**Would a smaller change do?** I tried that half first. The table plus one parse helper removes the duplicate parsing but not the double channel: `adbInvocation` still stitches `-s`/`-P` into argv at six sites, the provider arm still has to strip what it just read to hand the payload over, and `provider-limrun` still needs a structural copy of a type it cannot import by name. The invocation is what lets the provider branch pass the caller's array through untouched. I did take your narrower answer on the port: `serverPort` no longer travels as a live option on device routes. Removing the field from published `AndroidAdbExecutorOptions` is a public break that wants its own decision — say the word and I will open it as a follow-up.

**Live runs still owed.** Both remain unavailable here, so no evidence is claimed for them: a managed lease needs a Simlock host and token (`agent-device --debug` on a leased Android device, expecting `adb -P <lease> -s <serial> …`), and a Limrun session needs `LIMRUN_API_KEY` (`logs`/`exec` plus cleanup, expecting `adb -s <tunnel-serial>` and `disconnect`). `snapshot`/`press` additionally need `pnpm build:android-snapshot-helper` and `package:android-snapshot-helper:npm`, whose assets this worktree lacks; the recorded live run stays the ambient local route.

thymikee:
Reviewed at 8af64ef, as a follow-up to the [review at 1dc9dea](https://github.com/callstack/agent-device/pull/2632#issuecomment-5680259164). The server-mismatch rule now has more than one owner, and the managed-lease and Limrun routes still have no live run.

`requireSameServer` at [adb-provider-scope.ts#L127](https://github.com/callstack/agent-device/blob/8af64ef/packages/platform-android/src/adb-provider-scope.ts#L127) builds the same error as `requireAndroidAdbServerPort` at [adb-transport.ts#L474](https://github.com/callstack/agent-device/blob/8af64ef/packages/platform-android/src/adb-transport.ts#L474): `COMMAND_FAILED`, the same message, and reason `managed-device-transport-mismatch`. `transportMismatch()` already covers the device and target variants. If one copy drifts, the device route and root lowering answer the same misuse differently. Can `transportMismatch` get a server variant, with one exported check such as `requireSameAndroidAdbServer(owned, requested)` used by both, and the local copy deleted?

Under a lease, doesn't `withScopedTransportOptions` ([adb-provider-scope.ts#L304](https://github.com/callstack/agent-device/blob/8af64ef/packages/platform-android/src/adb-provider-scope.ts#L304)) drop a caller's `serverPort: 9999` without an error? `deviceServerPort` refuses the same option on device routes, and a `-P` in argv is overwritten. The route never moves, but the same misuse gets a refusal, a silent drop, or a silent overwrite depending on the route. Can every lease site use the one check above? If host-transport calls should ignore the option, can the PR body say so, with a 9999 case for `createScopedHostTransport`?

Your answer on the smaller design holds: the table plus parser alone keeps six `-s`/`-P` stitch sites and the provider's type copy. The extra surface left is the second port channel. Would removing `serverPort` from `AndroidAdbExecutorOptions` make a mismatch unrepresentable and delete `deviceServerPort`, `requireSameServer`, `withScopedTransportOptions` and `requireAndroidAdbServerPort`? That is a public SDK change, so it needs a maintainer decision, and the follow-up you offered is a good place for it. The continue-or-stop decision on #2617 should also land before merge.

Only the local emulator route ran live, and this PR changes two more. A `--debug` run on a Simlock-leased Android device (open, snapshot or press) should show every adb call as `adb -P <leasePort> -s <serial> ...`, plus one exec with a mismatched `serverPort` that returns `managed-device-transport-mismatch`. A Limrun Android session should show `adb -s <tunnel-serial> ...` on exec, `adb disconnect <tunnel-serial>` on close, and a forced failure that prints the command from `serializeAndroidAdbInvocation`. If either run is blocked, can the PR body name the device needed and the remaining risk?

One iOS Smoke Tests job is still running; this diff does not touch the iOS route, so it looks unrelated. There are no conflicts. Next: one server-mismatch check, the two live runs, and the #2617 decision.

thymikee:
Rebased on `origin/main` `2ad819ca06` and pushed as `d3162a0ade` (the previous history was tree-equivalent and had too many conflict hunks to replay, so it went in as one commit for one contract; `backup/pre-rebase-8af64ef` holds the old history).

Point by point:

**1 — one server check.** Agreed, and taken further than a call. `transportMismatch` now takes `'device' | 'target' | 'server'` and reads one message table, and `requireSameAndroidAdbServer(owned, requested)` is the only place that compares ports: `requireAndroidAdbServerPort`, `deviceServerPort`, and provider scope all go through it. The local `requireSameServer` and its duplicate message are deleted.

**2 — a lease dropping `serverPort`.** Fixed at the owning type rather than at each call. Every lease site now answers through that one check: `deviceServerPort` refuses a mismatched construction port and a mismatched per-call port for the executor and spawner routes, and `execOnScopedTransport` — where a lease builds host options — refuses before either arm runs. The `9999` case is pinned for `createScopedHostTransport` as well, which was the arm that silently swallowed it.

**3 — remove `serverPort` from `AndroidAdbExecutorOptions`.** This is your call, not mine: it narrows a published input type on `agent-device/android-adb` and the semver answer decides whether the deletion is allowed. I filed it as #2640 with the shape split (host options keep a per-call port, device executors and the `*WithAdb` helpers stop declaring one) and the exact deletions it buys — `deviceServerPort`'s three-way precedence, the option-channel removal in lowering, and the per-call callers from item 1. Say the word and I land it here instead.

While doing that round I cut more, since the contract allowed: `detached` is out of `AndroidAdbExecutorOptions` (it is a spawn detail; both the root host and the test host apply `process.platform !== 'win32'` where they fork), `lowerAndroidAdbInvocation` lives beside the grammar so root and the test host stop maintaining two lowerings, the executor and spawner share `deviceAdbRouteRequest`, and four internal helpers in `adb-transport` are no longer exported — their tests assert parse and managed behavior instead of internals. Full suite 1325 files / 10326 tests, provider-integration 210, layering 243/0, package gate and typecheck clean.

**4 — live runs.** Ambient route now has more evidence than a snapshot: on `emulator-5554` (Pixel 7 CI, SDK 36) through the built CLI I ran `devices`, `open com.callstack.agentdevicelab --foreground`, `apps`, `keyboard status`, `clipboard write`/`read`, `screenshot` (1080x2400 PNG), `logs start`/`mark`/`stop` (15 KB), `close`. The two lanes you asked for are still owed and the body now names what each needs: a managed-lease run needs a Simlock host and token (this machine has neither), a Limrun session needs an API key, and `snapshot -i`/`press` additionally need `package:android-snapshot-helper` assets that are absent here — `open` reports the missing artifact and every other command succeeds without it. If you can point me at a Simlock host with an Android slot I will run the lease lane before merge.

**5 — #2617 disposition.** Understood as yours. The PR does not close it and says so; the route table and the cost claim are on the issue. If the answer is "stop here", the useful residue is the grammar table and the provider handoff, which are the parts a different design would still want.

**6 — rebase.** Done; #2599's process-group handling and #2636's chunking are intact (`CHANGELOG.md`, `recording/chunks.ts` and the overlap in `adb-host-test-setup.ts` are main's, not my earlier copies).

One unrelated observation from the live pass: `cmd clipboard get text` on this SDK 36 image prints `No shell command implementation.` and exits 0, so `clipboard read` returns success with empty text. Pre-existing, and `isAndroidShellCommandUnsupported`'s comment says a zero exit is not evidence of failure — flagging it only because it made my own live check ambiguous, not as a request in this PR.

thymikee:
Reviewed at d3162a0, as a follow-up to the [review at 8af64ef](https://github.com/callstack/agent-device/pull/2632#issuecomment-5682081303). The lease refusal through `options.serverPort` is now one check, and #2640 tracks removing `serverPort`. Two things are left.

Isn't the argv channel still different? Under a lease, a caller passing `-P 9999 -s <serial>` gets the port overwritten, not refused ([adb-transport.ts#L527](https://github.com/callstack/agent-device/blob/d3162a0/packages/platform-android/src/adb-transport.ts#L527)). The override at adb-provider-scope.ts#L259 and `execOnScopedTransport` at #L299 both reach it through `applyManagedAndroidAdbServer`. `androidAdbOwnedServerPort` ignores a port from argv, so `requireSameAndroidAdbServer` never sees it, and on the serial-override branch the argv with `-P 9999` reaches `scope.provider.exec` for a non-local scope. This is not a regression, since main also overwrote `-P`, but the same misuse is refused or overwritten depending on the channel. Can `applyManagedAndroidAdbServer`, which every lease route passes through, run `requireSameAndroidAdbServer` on the argv port before it rewrites? A `-P 9999` test under a 15037 lease that expects `managed-device-transport-mismatch` would pin it. If overwrite is intended, can the PR body and #2640 say so? It is not confirmed that any production lease call passes `-P` today.

On size: removing `AndroidAdbExecutorOptions.serverPort` would also delete `deviceServerPort` and the per-call checks. Keeping this PR as one contract move is fine if a maintainer decides the semver question on #2640 first.

Live evidence for the two routes is still missing. For a managed lease: a Simlock lease with a server port other than 5037 running `open`, `snapshot` and `logs` start/stop, with `--debug` output showing `-P <leasePort> -s <leaseSerial>` and `ANDROID_ADB_SERVER_PORT=<leasePort>`, and a call naming another serial or server failing with `managed-device-transport-mismatch`. For Limrun: a session running `open`, `snapshot -i`, `press` and `logs`, showing the provider gets argv without the scope's `-s` pair and an adb failure reports the addressed invocation. The emulator-5554 run covers neither route.

All 20 checks pass and there are no conflicts. Next: the two live runs, and a decision on argv `-P` under a lease.

thymikee:
`e795aa32ca` implements the port rule as specified. CI green, unit 1325 files / 10327 tests, provider-integration 210, layering 243/0, package gate sound.

**`-P` under a private server.** One rule, both channels, checked in `requireManagedAndroidAdbAddressing(target, managedPort)` before anything is dispatched: no port named → the transport's own; matching → accepted; different → `managed-device-transport-mismatch`. The provider-forwarding branch parses the argv and applies the same check, so a caller's `-P` no longer reaches a provider that cannot address it. Ambient is untouched: `runAndroidHostAdb(parseAndroidAdbArgv(['-P','9999','devices']))` with no private server named still runs that argv, pinned in the root host test. Zero-executor-calls is pinned twice over — the provider-scope and device-route tests assert an untouched call recorder, and the root host and reachability fakes now append every call they *actually serve* to a log, with the assertion that no refused request appears in it.

One consequence worth your eye: the rule also covers a host call whose two channels disagree, so `runAndroidHostAdb(parseAndroidAdbArgv(['-P','9999','devices']), { serverPort: 15_038 })` is now refused instead of quietly preferring the option. Same trap, same answer; pinned.

The tightening is documented in the code doc on `requireManagedAndroidAdbAddressing` and in the PR body. No versioned help or user doc described the old overwrite — `serverPort` reaches it only through the published SDK types, and #2640 owns that surface.

**Live evidence gathered here.** Built the missing helper assets (`pnpm build:android`; snapshot + IME APKs land in `android/*/dist`, both installed on the emulator) and ran the full lane on `emulator-5554`: `open com.callstack.agentdevicelab --foreground` returning interactive `@refs`, `press @e18 --settle` settling +18/−10 onto Settings, `snapshot -i`, `screenshot`, `logs start`/`mark`/`stop`, `close`.

Then against a **real second adb server** (`adb -P 15037 start-server`, separate daemon state dir), driving the exact route a lease installs:

- `createLocalAndroidAdbProvider(device, { serverPort: 15037 })` answered `shell id` (`uid=2000(shell)…`) and `getprop ro.serialno` (`EMULATOR36X2X12X0`);
- with a call in flight, the process table read back `adb -P 15037 -s emulator-5554 shell sleep 3` — external proof of the lowered argv, not a mock's word for it;
- a background `logcat` spawn through the same route returned 961 lines;
- `exec(['-P','9999','shell','id'])` threw the typed mismatch and two seconds of polling `lsof -iTCP:9999 -sTCP:LISTEN` showed no server ever forked there, so nothing dispatched;
- the ambient host route kept answering unchanged;
- and the whole CLI through a private server (`ANDROID_ADB_SERVER_PORT=15037` + separate state dir) completed `devices`, `open`, `press --settle`, `screenshot`, `logs start`/`stop`, `close`.

**Still blocked, stated as blocked rather than waived.**

- *Simlock-granted lease:* not runnable on this machine at all. `agent-device host` does not exist in 0.21.3 — the ADR 0021 front-end is unshipped — and `connect --lease-id <id> --lease-backend android-instance` requires a Host plus co-resident Simlock to mint the lease and hold an Android slot. The private-server mechanics a lease installs are now live-verified; the grant, renewal, release and serial-ownership around them are not. I need a Host+Simlock host with an Android slot, or you run this row.
- *Limrun Android:* `connect limrun --platform android` → `INVALID_ARGS: connect limrun requires LIMRUN_API_KEY`. With a key (`LIMRUN_API_KEY`, optional `LIMRUN_REGION`) and an available Android instance the lane is `connect limrun --platform android` → `install` → `open` → `snapshot -i` → `press` → `logs start`/`stop` → `close` → `disconnect`, checking tunnel addressing, cleanup `disconnect`, and the addressed failure output.

Keeping merge on those two rows as you set it. #2617's typed-addressing direction continues here, and #2640 stays out of this branch.

thymikee:
Reviewed at e795aa3. The argv `-P` check now lives in `requireManagedAndroidAdbAddressing`, which every managed route already calls, so the fix is in the right place. The new test does not reach it, though.

The lease-override case at [adb-provider-scope.test.ts#L181](https://github.com/callstack/agent-device/blob/e795aa3/packages/platform-android/src/adb-provider-scope.test.ts#L181) passes `'-P', '9_999'`. `Number('9_999')` is `NaN`, so `parseAndroidAdbArgv` keeps the pair in `hostGlobals`, and the older `hostGlobals` check refuses the call before the new server check runs. If the new check were removed, this test would still pass. Can you use `'9999'`, assert the server refusal message (for example `/cannot select another server/`), and add a `'-P', '15037'` case on the same branch that asserts what the provider receives?

Under a private adb server, a conflicting argv `-P` used to be rewritten to the managed port. Now it fails with `managed-device-transport-mismatch`. This reaches `createLocalAndroidAdbProvider(device, { serverPort })`, `runAndroidHostAdb(invocation, { serverPort })` and the Limrun runtime dependency. Could CHANGELOG.md get an Unreleased entry for this that also says ambient adb is unchanged?

On design: is removing `AndroidAdbExecutorOptions.serverPort` (#2640) still the plan? It would delete `deviceServerPort`, `requireAndroidAdbServerPort` and the per-call `requireSameAndroidAdbServer` checks, and leave argv as the only addressing channel. That needs a decision on the semver question in #2640 first.

Live evidence is still missing for two routes. Limrun Android ships changed code in provider-limrun/src/android.ts and app-log-reconnect.ts, so it needs a run of `connect limrun --platform android`, `open`, `snapshot -i`, `press`, `logs start`/`stop`, `close` and `disconnect`, with `--debug` output that shows the provider gets argv without the scope's `-s` pair, plus one forced adb failure. The runs in the PR body do not reach a managed lease on a non-5037 port through `withAndroidAdbProvider`. If no lease can be granted from the shipped CLI in 0.21.3, the in-process reachability test can stand in for that route.

Smoke Tests is still running. This PR changes `lowerAndroidAdbInvocation` and adb-host.ts, so a failure there would not be clearly unrelated. There are no conflicts. Next: fix the test literal, add the changelog entry, and run the Limrun route.

thymikee:
`0bf8b3e424` fixes the test, `1de97dd415` adds the changelog entry.

**Test literal.** `'-P', '9_999'` became `'-P', '9999'`, the assertion is `.rejects.toThrowError(/cannot select another server/)`, and the same branch now has the matching-port case: `['-P','15037','-s',DEVICE.id,'shell','ls']` under a `15037` lease reaches the provider as `['-P', '15037', 'shell', 'ls']` — argv minus the scope's `-s` pair, port left where the caller wrote it. I then checked the reachability claim directly: deleting the `requireSameAndroidAdbServer` call from `requireManagedAndroidAdbAddressing` now fails four tests — this provider branch, the device route, the reachability run, and the root host — instead of none.

**CHANGELOG.** Added under Unreleased as `Changed (android)`: the refusal replaces the argv rewrite, names the three surfaces it reaches (`createLocalAndroidAdbProvider(device, { serverPort })`, `runAndroidHostAdb(invocation, { serverPort })`, the Limrun runtime dependency), and states what does not change — a request naming no server or the one its route holds, and ambient adb, whose argv still runs as typed `-P` and all.

**#2640.** Still the plan, and still gated on your semver call. Removing `AndroidAdbExecutorOptions.serverPort` deletes `deviceServerPort` and its three-way precedence, `requireAndroidAdbServerPort`, and the per-call `requireSameAndroidAdbServer` callers in `execOnScopedTransport`, leaving argv as the only addressing channel; the host port keeps a per-call port on its own options type. I am not folding it in here — say the word after the semver decision and it lands as its own change.

**Live, on the two shapes you named.** Ran them in process against a real adb server on 15037 and the real emulator:

- *Private-port scope* (`withAndroidAdbProvider(provider, { serial, serverPort: 15037 })`): `runAndroidHostAdb(['-s', serial, 'shell', 'id'])` → `uid=2000(shell)…`; process table during a call → `adb -P 15037 -s emulator-5554 shell sleep 3`; `devices` answered by the lease transport; `-P 9999` → `managed-device-transport-mismatch` with zero calls added to the provider; forced failure → `adb -s emulator-5554 shell definitely-not-a-command exited with code 127`.
- *Ambient provider scope*, which is Limrun's forwarding shape: the provider received `[["shell","id"], ["wait-for-device","shell","getprop","ro.serialno"], ["shell","definitely-not-a-command"]]` — `-s` pair gone, readiness token preserved — with real adb answering underneath (`uid=2000(shell)`, `EMULATOR36X2X12X0`), and the same forced failure addressed the same way.

One routing fact I ran into while building that, pre-existing and unchanged here: under a port scope `runAndroidHostAdb` is answered by the lease transport, so a forwarding provider installed *with* a private server gets nothing through that channel and is reached via the executor-override channel instead, where the same server rule applies. No shipped combination uses both; noting it because it decides which arm a future lease-plus-provider route would exercise.

Per your fallback, the in-process reachability test now stands in for the lease route on a non-5037 port: it drives a real adb binary through the lease port and asserts the refused request appears in no dispatch log. The allocator's own grant/renew/release around that transport remains unverified — no `host` command ships in 0.21.3, and `connect --lease-backend android-instance` needs a Host plus co-resident Simlock to mint a lease.

**Limrun** is unchanged and still owed: `connect limrun --platform android` → `INVALID_ARGS: connect limrun requires LIMRUN_API_KEY`. Key plus an available Android instance are the only missing pieces; the lane and the `--debug` check are written into the PR body so it is copy-pasteable for whoever has them.

Smoke Tests passed; CI was 0 non-green at `e795aa3`, and the two new commits are lint/format/typecheck/layering clean with unit 1325 files / 10327 tests.

thymikee:
Rebased on `origin/main` `9173d3c797` (0.21.4) and force-pushed; head is `b246b9546c`.

The only conflict was `CHANGELOG.md`, where main's three iOS entries from #2639 and this PR's Android entry both want the top of Unreleased — both are kept. `mechanics.ts` auto-merged: #2604 removed two snapshot-helper exports and this branch trimmed unrelated ones, and neither side touched the other's line.

Integration was not just textual, so it was re-verified rather than trusted. #2604 moved device-ownership reads off adb onto the device shell, which adds `adb shell pidof` traffic through the same provider this branch re-typed; #2629 re-routed test scratch through `mkdtempForTest` and forbade `node:os` in tests, which the tests added here already satisfy. At head, after `pnpm install --frozen-lockfile` and `pnpm build`: unit 1325 files / 10340 tests, provider-integration 210, layering 243 pass / 0 fail, lint, typecheck, and the package gate ("the package npm would publish is sound") all clean.

Live re-run at head with helpers rebuilt for 0.21.4 (`pnpm build:android`): `open com.callstack.agentdevicelab --foreground` returning the interactive snapshot, `press @e18 --settle` settling +18/−10, `close`.

The two owed rows are unchanged by the rebase and still open as stated above: Limrun needs `LIMRUN_API_KEY` plus an Android instance, and a Simlock-granted lease needs a Host plus co-resident Simlock, with the in-process reachability path standing in per your note.

thymikee:
Reviewed at b246b95, as a follow-up to [the review at e795aa3](https://github.com/callstack/agent-device/pull/2632#issuecomment-5684917709). Both points are fixed. The lease-override test now passes `'9999'` and asserts `cannot select another server`, with a `'15037'` case that checks the argv the provider receives, and CHANGELOG has an Unreleased entry for the `-P` refusal that says ambient adb is unchanged. The rebase keeps both CHANGELOG sides, and the `mechanics.ts` auto-merge touches separate exports.

Removing `serverPort` from device-scoped executors now lives in #2640, because it is an SDK break that needs its own semver decision, so it does not block this PR.

CI is green, and there are no conflicts. The Limrun Android route is still not live-validated, and this PR changes provider-limrun/src/android.ts and app-log-reconnect.ts. It needs a `connect limrun --platform android`, `open`, `snapshot -i`, `press`, `logs start`/`stop`, `close`, `disconnect` run with `--debug` showing the provider argv without the scope's `-s` pair, plus one forced adb failure. The Simlock lease route stays covered by the in-process reachability test, as agreed. Next: that Limrun run, or a maintainer decision to accept this PR without it.

thymikee:
Head is `164a9fde6a`. CI was 0 non-green at `b246b9546c`; unit 1326 files / 10344 tests, provider-integration 210, layering 243 pass / 0 fail, lint and typecheck clean.

**The Limrun run is still not done, and I could not get it done here.** I looked for a way rather than restating the blocker: no `LIMRUN_API_KEY` in the environment, shell rc files, `.env`, `~/.config`, or the keychain; the stored `~/.agent-device/remote-connections/generated/limrun-*.json` profiles carry session and tenant identity only — `leaseProvider`, `session`, `tenant` — no credential, so `connect limrun --platform android` still returns `INVALID_ARGS: connect limrun requires LIMRUN_API_KEY`. `connect limrun` verifies an existing instance service and does not create one, so a key alone is not sufficient: it also needs an Android instance to be available. So this needs a key plus an instance from someone who has them, or your call to merge without the run.

To make the run cost nothing but typing, it is scripted below, and to keep the changed code from being only unit-tested by inference I pinned the two shapes that run would have checked (`164a9fde6a`):

- `packages/provider-limrun/src/android.test.ts` (new, mirroring `android.ts`): the tunnel serial rides on `target.selector` and never enters the command array (`serializeAndroidAdbInvocation` adds `-s <serial>` on the way out, `command` stays `['shell','pm','list','packages']`); `limrunHostAdbInvocation` addresses no device, so `disconnect <serial>` stays in the command; `cleanupLimrunAndroidAdbTunnel` issues exactly `disconnect <serial>` with `allowFailure: true, timeoutMs: 10_000`, closes the tunnel, and clears `adbSerial`/`adbTunnel`/`adbTunnelPromise`.
- `app-log-reconnect.test.ts`: the reconnect path drives `readLogs` through the same builders and the dispose path is asserted at the same boundary — the reader's own command travels unchanged, and the cleanup command selects no device.

Those cover the tunnel addressing, the disconnect on cleanup, and the "provider passes the caller's argv, not a rebuild of it" claim. They do not cover the WebSocket tunnel, a real instance's adb handshake, or the real `disconnect` semantics, which is what the live run is for, so I am not treating them as a substitute.

```sh
export LIMRUN_API_KEY=…            # optional: LIMRUN_REGION
agent-device connect limrun --platform android --json
agent-device install path/to/app.apk --json
agent-device open <package> --foreground --json --debug
agent-device snapshot -i --json --debug        # provider argv must lack the scope's `-s` pair
agent-device press <ref> --settle --json --debug
agent-device logs start && sleep 5 && agent-device logs stop --json
agent-device close --json
agent-device disconnect --json                 # cleanup: tunnel close + adb disconnect <serial>
# one forced adb failure, addressed output must name the command it addressed:
agent-device open com.example.definitely-not-installed --json --debug
```
If you would rather have this as a lane in the repo than as a comment, say so and I will land it as a script.

**Decision requested:** run it with credentials you have, or tell me to treat the PR as accepted without the Limrun run. The Simlock lease row stays covered by the in-process reachability path plus the real-private-server run, as agreed.

thymikee:
## Live Limrun Android lane: ran it

Credentials are available in my environment, so the second merge-gating route is no longer blocked. Everything below ran at head `164a9fde6a` against a real Limrun Android instance (session `apex-2632-lane`, lease via `connect limrun --platform android --session apex-2632-lane --force` → `Credentials and Android instance access verified.`, device `deferred`).

**Happy path on the remote instance**

- `apps --session apex-2632-lane` → uploaded-asset catalog returned (`android-app.apk`, `com.callstack.agentdevicelab.apk`, …).
- `open "com.callstack.agentdevicelab.apk" --foreground` → `lease_allocate` 6.9s, `open` 13.1s, `Opened: com.callstack.agentdevicelab`, interactive snapshot `9 visible nodes (29 total)` with `@refs`.
- `snapshot -i` → green, `@e1…@e29` with labels.
- `press @e2 --settle` → `Tapped @e2 (361, 743)`, `settled after 3269ms: +0 -0 (~29 unchanged)`.
- `press @e26 --settle` (Catalog tab) → `Tapped @e26 (216, 1504)`, `settled after 3354ms: +33 -24 (~5 unchanged)` — a real navigation diff, not a no-op.
- `logs start` → `app.log`, 6s capture, `logs stop` → **1273 lines, 38 naming `com.callstack.agentdevicelab`**: real logcat through the tunnel, i.e. the changed `app-log-reconnect` / `app-log-poller` path on a live provider rather than a fake.
- `close` → `Closed: apex-2632-lane`; `disconnect` → `Disconnected remote session "apex-2632-lane".`

**Tunnel addressing, from the session request log (`exec_command` adb records)**

The provider receives payload and the serial travels as transport addressing, which is exactly the claim in the second commit's tests:

```
["-s","127.0.0.1:65049","shell","am","instrument","-w", …16 omitted]
["-s","127.0.0.1:65049","shell","cmd","package","query-activities", …5 omitted]
["-s","127.0.0.1:65049","shell","pm","list","packages"]
["-s","127.0.0.1:65049","reverse","--list"]
["disconnect","127.0.0.1:65049"]
```

The last line is the cleanup shape `cleanupLimrunAndroidAdbTunnel` now builds: a server-level command whose serial sits in the command itself, with no `-s` pair in the payload. After `close` + `disconnect`, a fresh `connect` allocated a new tunnel serial (`127.0.0.1:65439`), which `devices --platform android` lists next to `emulator-5554`, and `clipboard write` succeeded on it.

**Forced adb failure**

Injected at the transport boundary mid-session by dropping the tunnel serial from the local adb server (`adb disconnect 127.0.0.1:65439`), then asking the remote session for keyboard state:

```
keyboard status → exit 1
  code:    COMMAND_FAILED
  message: Failed to query Android keyboard state: device '127.0.0.1:65439' not found
  hint:    The device disconnected or is restarting — verify it is listed in adb devices, then retry.
  exitCode: 1
```

The matching request-log record shows the command that failed with its addressing intact: `["-s","127.0.0.1:65439","shell","dumpsys","input_method"]`. So a real nonzero adb exit on a provider tunnel is still classified, hinted, and attributed to the addressed command.

Two failures observed on the lane that are *not* adb-level, recorded so the negative space is explicit: `open com.example.definitely-not-installed` → `APP_NOT_INSTALLED` (`No package found matching …`) answered during package resolution, and `install` of a deliberately truncated archive → `Given archive does not contain AndroidManifest.xml: android-app.apk`, rejected by the Limrun asset path before any adb command. The shipped CLI has no arbitrary-adb surface, so the serial-drop injection above is the only way to get a genuine adb exit code out of a remote session.

**Cleanup**: remote session closed and disconnected, connection profile gone; local emulator verified unchanged (`pm list packages -3` → snapshot helper, ime helper, `com.callstack.agentdevicelab` only).

## Remaining validation status

| Route | Status |
| --- | --- |
| Limrun Android | **Live-verified** (above) |
| Managed lease | still not runnable here: `agent-device host` is not shipped in 0.21.x, so no lease can be granted to this checkout. Standing in per your call: in-process reachability tests plus the real private adb server on `-P 15037` (device-scoped argv `adb -P 15037 -s emulator-5554 shell sleep 3`, `-P 9999` refused with no listener forked on 9999, logcat spawn 961 lines, ambient host unaffected). |

## CI note

`Smoke Tests` went red at `164a9fde6a` on `RunnerTests.testAlertCannotProveAnIdenticalReplacementAndDoesNotActivateIt` — 96 tests executed, that one failing (twice, its own retry), everything else green. That test came in with `27a97ee619 fix(ios): confirm alerts without repeating activation (#2326)`, this branch touches no Apple code, and `main` ran the same suite green at `fffa01e0` 45 minutes earlier. I reran the failed job rather than pushing a no-op commit; it is in progress.

thymikee:
Rerun of the failed `Smoke Tests` job passed on the same head `164a9fde6a` (run 35084621336): all 19 real checks green, only the `matrix.name` placeholder skipped, `MERGEABLE` / `CLEAN`. That confirms the earlier `testAlertCannotProveAnIdenticalReplacementAndDoesNotActivateIt` red was environmental, not this branch.

thymikee:
Reviewed at 164a9fd. The change since b246b95 adds only tests. The Limrun Android run you posted reaches the changed route, and the argv and error shapes it reports match the code.

One reuse point remains in code that the b246b95 review did not flag. `limrunDeviceAdbInvocation` at [android.ts#L145](https://github.com/callstack/agent-device/blob/164a9fd/packages/provider-limrun/src/android.ts#L145) writes the target object by hand, but it is the same value that `androidAdbInvocation(androidAdbSerialTarget(serial), args)` returns, and this PR exports both builders from `@agent-device/platform-android/mechanics`. Nothing is wrong today, but if `AndroidAdbTarget` gains a field or changes a default, this copy can drift without an error. Can it call the shared builders? `limrunHostAdbInvocation` can use `androidAdbInvocation` in the same way.

The forced adb failure ran with `allowFailure: true`, so the typed refusal in `requireSuccessfulLimrunAndroidAdb` has unit coverage only.

All 20 checks pass after the Smoke Tests rerun. The earlier failure was an iOS runner test that this diff does not touch. There are no conflicts. Next: the builder reuse above.

thymikee:
## Builder reuse: yes, but not through the import you pointed at

`0688fd0e7c`. I first did exactly what you described — `import { androidAdbInvocation, androidAdbSerialTarget } from '@agent-device/platform-android/mechanics'` in the provider — and two existing gates refused it:

```
FAIL scripts/__tests__/eager-closure-budgets.test.ts > 'packages/provider-limrun/src/index.ts' evaluates no more modules than at the merge-base
FAIL scripts/__tests__/eager-closure-budgets.test.ts > 'packages/provider-limrun/src/index.ts' never evaluates a concrete platform implementation
```

That import is a runtime edge, so the provider's eager closure starts evaluating the Android mechanics facet, which ADR-0019 keeps behind discovery or binding. The gate's own advice — a function-scoped `await import` — would make a pure data constructor async, so I used the seam the provider already has for exactly this shape of platform value, `android.adbError`.

`LimrunAndroidRuntimeAdapter` gained two sync members:

```ts
deviceAdbInvocation(serial: string, command: readonly string[]): AndroidAdbInvocation;
hostAdbInvocation(command: readonly string[]): AndroidAdbInvocation;
```

and the composition root implements them with the shared builders (`src/sdk/limrun-runtime-dependencies.ts`, which already statically imports `runAndroidHostAdb`, so no new edge):

```ts
deviceAdbInvocation: (serial, command) => androidAdbInvocation(androidAdbSerialTarget(serial), command),
hostAdbInvocation: (command) => androidAdbInvocation(androidAdbHostTarget(), command),
```

`limrunDeviceAdbInvocation` and `limrunHostAdbInvocation` are deleted — both were introduced by this PR, so nothing outside loses them — which means the provider no longer writes an `AndroidAdbTarget` anywhere. There is no second copy to drift, which was the actual concern. The provider still owns the *decision* (the tunnel serial goes on the target, the command array stays untouched); it just no longer owns the syntax.

`androidAdbHostTarget()` is new in `adb-transport.ts`, and `parseAndroidAdbArgv` now seeds its addressing read from it, so "selects no device, names no private server" is stated in one place rather than in two literals.

## The typed refusal is worse than unit-only

I chased live coverage for it, and the answer is that it cannot be reached in the shipped wiring. Dropping the tunnel serial mid-session and running `screenshot` — a device command that sets no `allowFailure` — gives:

```
code:    COMMAND_FAILED
message: adb -s 127.0.0.1:64570 exec-out screencap -p exited with code 255: device '127.0.0.1:64570' not found
hint:    The device disconnected or is restarting — verify it is listed in adb devices, then retry.
details: { exitCode: 255, adbFailure: "device_not_found", processExitError: true, … }
```

That message is `runAndroidHostAdb`'s (`packages/platform-android/src/adb-host.ts:144`), which throws on the same condition the limrun guard tests, and it always wins because the shipped `host.runAdb` *is* `runAndroidHostAdb` (`src/sdk/limrun-runtime-dependencies.ts:82`). So `requireSuccessfulLimrunAndroidAdb`'s `adbError('Limrun Android ADB command failed', …)` only fires if an injected host adapter *returns* a nonzero result instead of throwing — which no shipped adapter does; `LimrunHostAdapter.runAdb` is typed to return a result, so the guard is the provider enforcing its own declared seam. I left the branch in place for that reason, but it is a contract detail and not a user-visible path, and deleting it so the host owns the single failure contract is a one-line change if you prefer that.

## Where coverage went

- `src/sdk/limrun-runtime-dependencies.test.ts` (new case): the port builders equal the shared builders and serialize to `-s 127.0.0.1:62001 shell pm list packages` / `disconnect 127.0.0.1:62001` with no device option.
- `adb-transport.test.ts`: `androidAdbHostTarget` emits no device option, and an argv naming neither device nor server reads as exactly that target.
- `app-log-reconnect.test.ts` and `android.test.ts`: the provider forwards the invocation the port addressed, unchanged — the logcat command travels with no `-s` stitched into the payload, and cleanup hands over the `disconnect` with `allowFailure`/`timeoutMs`.
- The provider-side device path is still asserted end to end by `runtime-dependencies.test.ts`, which checks the host receives `-s 127.0.0.1:62001 reverse tcp:8081 tcp:8081` from a real tunnel serial.

Gates at `0688fd0e7c`: `pnpm test:unit` 1326 files / 10345 tests, provider-integration 66 files / 210 tests, `check:layering` OK, format/lint/typecheck clean, package gate sound. `check:production-exports` reports the same 68 pre-existing findings it reports without this commit.

thymikee:
Reviewed at 0688fd0. This answers the builder reuse point. The provider no longer writes an `AndroidAdbTarget` itself. Both invocations now come from the shared builders through the runtime adapter, the same way `adbError` already does. The new case in `limrun-runtime-dependencies.test.ts` checks that they equal the shared builders and serialize to the same argv. The earlier Limrun Android run still covers this route, because this commit builds the same values.

`requireSuccessfulLimrunAndroidAdb` is already on main, so its removal can be a separate change.

Smoke Tests is still running on 0688fd0, and the other checks pass. There are no conflicts. Next: human review.

thymikee:
Re-ran the Limrun lane at `0688fd0e7c`, since this commit moves invocation construction behind the injected port and that is a live path: `open "com.callstack.agentdevicelab.apk" --foreground` into a 9-visible/29-total interactive snapshot, `press @e26 --settle` navigating with `+33 -24 (~5 unchanged)`, `logs start`/`logs stop`, `screenshot` (720x1616), `close`, `disconnect`. The request log at this head still records the device command with the tunnel serial as addressing and the helper's payload intact: `["-s","127.0.0.1:53931","shell","am","instrument","-w", …16 omitted]`. All CI checks green at this head, `MERGEABLE` / `CLEAN`.

PR #2633, opened 2026-09-15
test(ios): resolve the host app before stubbing its tree snapshot

okwasniewski:
## Summary

Follow-up to #2621. Running the iOS suite on a freshly booted simulator, with `testAbandonedTreeCaptureSkipsQuerySweepAndHonorsWarmupExemption` as the first test to launch the host app, the test runner was killed with `SIGTERM` about 30 s in (`Test crashed with signal term`; the console showed `Find the Target Application` and then nothing for 28 s). A probe build confirmed the stub is entered only from `captureSnapshotRoot`, so the intended block is the tree XPC; the cold first launch adds a slow resolution of the application element inside the plan.

The test now resolves the element before installing the stub (`wait(for: .runningForeground)` plus one `frame` read) and gives the plan wait 60 s instead of 30, so a slow cold path cannot tear the test down while the plan is still running. 1 file, +5/-1.

## Validation

Tested commit `09d30bfbe`.

- The test alone from a cold boot, three times (shutdown, boot, run): passed in 12.0 s, 11.8 s, 15.0 s, no runner restart.
- Full iOS runner suite from a fresh boot with this test first to launch the app: 248 tests, 0 failures, no runner restart.
- xctest-selection, packaged-runner-swift, format pass.


thymikee:
Clean at 09d30bf. Resolving the app element before the stub is installed, and giving the plan wait 60 s, keeps the cold first launch out of the blocked tree snapshot without changing what the test asserts. The change stays inside one XCTest file, all checks pass and there are no conflicts; the cold-boot runs in the PR body cover the changed setup.

PR #2636, opened 2026-09-15
fix(android): warn about the 180s record limit only when the last chunk ended early

thymikee:
## Problem

Every chunked Android recording (over 170 s) returned the truncation warning:

> Android adb screenrecord stopped before record stop, likely after reaching the 180s platform limit. The MP4 may be truncated; final interactions after the limit are not in the video.

The next chunk recorded those interactions, so the warning was wrong. Seen live on 2026-09-15: a 197 s recording returned chunks `and.mp4` (188.6 s) and `and.part-002.mp4` (47.9 s), plus both the chunked warning and this limit warning. The bug is on main; #2613 and #2631 did not add it.

## Cause

`stopOwnedChunks` combined every chunk's `stopChunk` result with OR. Rotation (`runtime.ts`) starts the next chunk and then stops the previous one. So at `record stop`, every earlier chunk always reports `already-missing`, and `reachedLimit` was always `true` for a chunked recording.

## Fix

`stopOwnedChunks` now returns the result for the active (last) chunk only. That is the only chunk that must still record at stop. Earlier chunks end because of rotation. The reattach path already used this rule: `recovery.ts` inspects only `chunks.at(-1)`.

## Tests

In `chunks.test.ts`, both tests go through the live `finish` path:
- `omits the 180s limit warning when only a rotated-out chunk exited before record stop`: two chunks, chunk 1 exited (rotation stopped it), chunk 2 still runs at stop. The result has the chunked warning and no limit warning.
- `warns about the 180s limit when the last chunk exited before record stop`: the only chunk exited before stop, and in a second case chunk 2 of 2 exited before stop. Both results have the warning.

Mutations I ran by hand:
- Put back the OR over all chunks (`activeAlreadyExited = alreadyExited || activeAlreadyExited`): **the omits test fails**.
- `const active = chunks[0]`: **the omits test fails**.
- `return false`: **the warns test fails**. The recovery tests still pass under this mutation because recovery passes its own `reachedLimit`. The warns test is the only test of the live path.

## Live validation

I booted a separate emulator (`Pixel_9_Pro_XL_API_37`, port 5590, not `Pixel_7_CI`) and used a source CLI with an isolated state dir. I opened Settings, started `record start`, scrolled every 10 s, and ran `record stop` after 218 s.
- Chunks: `live.mp4` 171.6 s and `live.part-002.mp4` 39.4 s (ffprobe).
- `warning`: *"Android adb screenrecord is capped at 180s, so this recording was split into multiple MP4 chunks. Android screenrecord encodes a frame only when the screen changes, so this video ends at the last frame it encoded: it covers 211.0s of the 222.6s recording window."* The limit warning is not in it.

## Checks

`pnpm format` done. `pnpm check:affected --run` stops at lint only because of `.claude/workflows/agent-device-pr-review.mjs`. That file is gitignored, the local harness makes it, and it is not in this diff. I ran each selected gate by hand. All pass: `oxlint . --deny-warnings --ignore-pattern '.claude/**'`, `typecheck`, `check:layering`, `check:fallow --base origin/main`, `build`, `vitest related`. The Android recording unit suite passes (85 tests).


thymikee:
Reviewed at e736fdd. Clean. `stopOwnedChunks` now returns the result for the last chunk only, so a chunk that rotation stopped no longer adds the 180s limit warning. The new `chunks.test.ts` cases cover both the dropped and the kept warning.

The live emulator run in the PR body (218 s, two chunks, no limit warning) exercises this route. All checks pass and there are no conflicts.

PR #2637, opened 2026-09-15
test(scripts): keep the indirect-ownership example on a test that still reaches the kernel

thymikee:
## Summary

`pnpm check:affected` stops at its `mutation-model` step on today's `main`: `scripts/mutation/ownership.test.ts` asserts that `src/commands/interaction/runtime/gestures.test.ts` indirectly owns the `scroll-edge-state` kernel, and the gesture test's import closure no longer reaches the scroll runtime at all. Any change touching a lane file runs that step, so this blocks more than it looks like it does — an edit to `.github/workflows/ios.yml` trips it.

The derivation it guards is intact: 147 tests own `scroll-edge-state`, every one of them reaching it through `src/daemon/scroll-runtime.ts` rather than naming it. What is stale is which test is used as the example, so the fix moves the example rather than relaxing the assertion. `src/daemon/__tests__/scroll-runtime.test.ts` sits one import away from `scroll-edge-state.ts` and never mentions it, which is exactly the omission the assertion exists to catch; the `kernel-errors` half of the test is unchanged.

One file, +6/-6, no runtime code.

## Validation

- Reproduced the failure at `5fb2860119` before touching anything: `node --test scripts/mutation/ownership.test.ts` → 4 pass, 1 fail (`a kernel is owned by tests that reach it indirectly`).
- At `5a8ba8e81e`: `node --test scripts/mutation/*.test.ts` → 40 pass, 0 fail.
- `pnpm check:affected --run` green at this head, including `mutation-model`, `gate-manifest`, `depgraph` and `tmpdir-leaks`.
- Test-only change, so no runtime or device validation applies.


thymikee:
Reviewed at 5a8ba8e. The change is correct: `scroll-runtime.test.ts` still reaches `scroll-edge-state` without naming it, so the assertion still proves indirect ownership.

This PR now conflicts with main. #2613 (208e3d3) already fixed the same assertion by moving the example to `back-runtime.test.ts`. Is this PR still needed? If main's fix is enough, it can be closed. If you prefer this example or the updated comment, rebase onto main and keep one of the two.

Smoke Tests fails in the live iOS scenario (the long-press element did not become visible after scrolling). This PR changes only `scripts/mutation/ownership.test.ts`, so that failure looks unrelated.

thymikee:
Closed: `main` already moved the example (in #2613, onto `src/daemon/__tests__/back-runtime.test.ts`, which is likewise an indirect owner of `scroll-edge-state`), so there is nothing left here to land. Verified at `354d3516ec`: `node --test scripts/mutation/ownership.test.ts` passes.

Issue #2638, opened 2026-09-15
iOS Simulator: Maestro engine and agent-device is/get/alert disagree depending on the bound snapshot backend (Simulator AX vs XCTest)


vynci:
## Summary

On the iOS Simulator, agent-device's **own** `is` / `get` / `alert` commands and the **bundled Maestro engine** (`test --maestro`) disagree about what is on screen, and which answer Maestro gives depends on which snapshot backend the session happens to bind. When the banner reads `Simulator AX snapshot inapplicable (system-surface-host-lingering); used XCTest, which read app content`, every flow passes. On a freshly booted simulator (Simulator AX bound) the same flows, same commit, fail on the same five kinds of element every run. The app is not involved: `agent-device is visible 'id="members.count"'` passes and `agent-device alert` prints `Sign out?` at the same second Maestro's `assertVisible` / `tapOn` against the same surfaces find nothing.

Filed as a concrete symptom for the convergence work in #2188 / #2274 (the "no fallback producer" note there matches what we see).

## Environment

- agent-device 0.21.3 (`npx agent-device@latest`), headless, no OAuth
- macOS 26.6.2, Xcode 27.0, iOS 26.5 simulator (iPhone 17 Pro), also seen on iOS 26.3
- Expo SDK 57 development build (React Native 0.86, New Architecture, Hermes), `@expo/ui` (SwiftUI) lists and sheets, `expo-router`
- Flows run with `agent-device test --maestro <dir> --platform ios`; assertions/taps by `testID` only

## What vanishes from the Simulator-AX tree (and not from XCTest)

1. A native `Alert.alert`'s buttons. `agent-device alert` prints the alert (`Sign out?`), Maestro `- tapOn: "Sign Out"` matches nothing. Under XCTest the same step passes.
2. An out-of-process system surface (`PHPickerViewController`): the picker is plainly on screen ("Collections" and all), `extendedWaitUntil visible: "Collections"` times out.
3. Fields inside a pushed `presentation: "formSheet"` route (expo-router native stack) — intermittently.
4. A non-interactive leaf `Text` inside a `FlatList` `ListHeaderComponent`: `agent-device get text 'id="members.count"'` returns `4 members`, Maestro `assertVisible: id: "members.count"` fails.
5. Anything inside a native map annotation (MapLibre).

A SwiftUI `BottomSheet`'s contents are **not** affected — they resolve under both backends — so this is about surfaces presented outside the app's own host, not sheets as such.

## Repro (minimal)

[...]

PR #2643, opened 2026-09-15
fix(ios): keep an identifier hook the fold cannot prove inert

thymikee:
## Summary

`collectIosStructuralIdentifierSuppression` deleted a childless `Other` wrapper carrying only an
identifier, assuming its content would carry that identifier onward. A childless wrapper has nothing
to delegate to, so the identifier left every canonical view while `is`, `get`, and `click` still
resolved it from the same capture. Whether a wrapper arrives childless is decided by a fact the
producer may not report at all: the Apple runner answers `hittable` for every node, the Simulator AX
bridge for none. Suppression now requires the node's own declared `hittable: false` (ADR 0004).

On one screen of one installed app, `snapshot -i` over both producers now yields identical 18-hook
sets: `home-title`, `gesture-lab-card`, `drag-gesture-fixture` and `gesture-metrics` are addressable
under Simulator AX, as they already were under XCTest. Refs #2638. Out-of-process surfaces in that
report (PHPicker, MapLibre annotation content) are not this rule and still need raw captures. 7
touched files.

## Validation

Tested at `35d1a35f77087f0412b644d49df073d9d45bbe4f`.

- `pnpm check:affected --run`: all runnable checks passed (2163 tests).
- Reverting the rule turns exactly two tests red: the rule case in `noise.test.ts` and
  `daemon-runtime-port-identifier-hook.test.ts`. The delegation case and both negatives stay green;
  they pin unchanged behavior.
- Live, iPhone 17 Pro, iOS 26.2, `examples/test-app`: `snapshot -i --json` identifier sets identical
  across producers (18 vs 18, none one-sided); a Maestro flow asserting `home-title` then
  `gesture-lab-card` reports `1 passed`; `is visible 'id="home-title"'` passes.
- A live off-screen negative was unreachable because this Home list ignores synthesized scrolls, so
  it is pinned at the adapter seam.

Risk: a bridge capture has no occlusion signal, so a hook behind an in-app modal can read visible —
predates this change, covers every bridge node. Sessions closed; simulator `2638-verify` left booted.


thymikee:
Reviewed at 35d1a35. The fold now suppresses a childless identifier-only wrapper only when the node itself declares `hittable: false`, so a producer that reports no hittability keeps the hook addressable. The new daemon-route test goes red without the guard, because a suppressed wrapper maps to no presented node and reads as not visible. The live `snapshot -i` comparison over both producers in the PR body covers the changed path.

Smoke Tests fails on `RunnerTests.testAlertAcceptDoesNotActivateAReplacementWithASharedButton`, a runner XCTest for alert activation. This diff changes no Swift or alert code, so the failure looks unrelated. There are no conflicts. Next: human review.

PR #2642, opened 2026-09-15
feat(limrun): long press on iOS direct sessions via HID touch primitives

okwasniewski:
## Summary

`LimrunIosInteractor.longPress` refused every call with `UNSUPPORTED_OPERATION` ("Limrun iOS direct sessions do not expose long press yet") even though the Limrun SDK exposes unpaired HID primitives for exactly this composition. The interactor now holds one touch as a single `performActions` batch of `touchDown`, `wait`, `touchUp`, so the hold duration runs on the device rather than across the wire. An omitted duration defaults to 800 ms, matching the Android and Linux interactors.

The `longPress` touch fact in `interaction-operations.ts` is now admitted on the iOS leg as well as Android, so `longpress` is accepted at admission instead of failing mid-execution. Gesture tiers, hover, home, and keyboard on Limrun iOS are unchanged.

Example on a `connect limrun --platform ios` session:

```sh
agent-device longpress @e7 1200
```

Independent of #2641; both touch adjacent lines in `ios.ts`, so whichever lands second needs a trivial rebase.

Touched files: 5 (`ios.ts`, `interaction-operations.ts`, its test, new `ios-interactor-long-press.test.ts`, `CHANGELOG.md`).

## Validation

- Tested commit: fbad6861 (on main at 9173d3c7)
- Live: verified on a Limrun iOS simulator against the repo test app. `longpress` flips the app's long-press canary (`Last input: longpress`, `Long presses: 1`) while a plain `press` reads `tap` and leaves the count alone. Clip, stills, and timeline in the PR comment below.
- `pnpm vitest run --project unit-core packages/provider-limrun/src/ios-interactor-long-press.test.ts packages/provider-limrun/src/interaction-operations.test.ts`: 14 passed. Regression evidence: with `ios.ts` and `interaction-operations.ts` reverted, all four new tests fail (the interactor throws the old unsupported error; the iOS fact reads `available: false`).
- `pnpm check:quick`: lint and typecheck clean.
- `pnpm check:affected --run --base upstream/main`: all runnable checks passed (114 test files, 722 tests).
- Remaining risk: the live run used this head merged locally with #2641 (`417df160`), not this head alone; the two changes touch different methods.

PR #2641, opened 2026-09-15
fix(limrun): batch the iOS double tap so the inter-tap gap stays on-device

okwasniewski:
## Summary

Limrun iOS direct sessions issued two independent `tap` requests for `press --double-tap`, so a network round trip sat between the taps. That exceeds the iOS double-tap recognition window, and every double tap registered as two slow single taps.

`LimrunIosInteractor.doubleTap` now sends both taps in one `performActions` batch with an 80 ms on-device `wait`, the same inter-tap interval the WebDriver interactor already uses. No other interactor or fact changes.

Example on a `connect limrun --platform ios` session:

```sh
agent-device press @e3 --double-tap
```

Touched files: 3 (`packages/provider-limrun/src/ios.ts`, new `packages/provider-limrun/src/ios-interactor-double-tap.test.ts`, `CHANGELOG.md`).

## Validation

- Tested commit: f5446df82 (rebased onto main at 6dd57d9ae, after #2642 merged)
- Live: verified on a Limrun iOS simulator against the repo test app. On a focused text field holding "hello world", a plain `press` shows the caret menu (`Select`, `Select All`, `AutoFill`) while `press --double-tap` selects the word and shows `Cut`, `Copy`, `AutoFill`. Clip, stills, and timeline in the PR comment below.
- `pnpm vitest run --project unit-core packages/provider-limrun/src/ios-interactor-double-tap.test.ts`: 2 passed. Regression evidence: both tests fail against the unpatched interactor (`performActions` never called; a rejected batch does not propagate).
- `pnpm check:quick`: lint and typecheck clean.
- `pnpm check:affected --run --base upstream/main`: all runnable checks passed (111 test files, 653 tests).
- Remaining risk: the live run used this head merged locally with #2642 (`417df160`), not this head alone; the two changes touch different methods.

PR #2645, opened 2026-09-16
fix(limrun): write the iOS screenshot as PNG instead of Limrun's JPEG bytes

okwasniewski:
## Summary

Limrun serves its iOS simulator capture as JPEG, and `LimrunIosInteractor.screenshot` wrote those bytes straight to the `.png` path. Every `screenshot` on a Limrun iOS direct session then failed in `readPngSize` with "Screenshot file is not a valid PNG". Observed live while verifying #2641 and #2642.

The interactor now sniffs the container and transcodes a JPEG to PNG before writing. A PNG passes through byte for byte, and anything else is refused with a labeled `COMMAND_FAILED`. The transcode runs as a PNG worker job (`transcodeScreenshotToPngAsync` in `png-worker-client.ts`, with the usual in-process fallback), so a full-resolution capture never blocks the event loop. The worker client is the only public route to the transcoder, and both share the one PNG signature check in `png.ts`.

New dependency: `jpeg-js` (pure JS, zero dependencies, BSD-3), bundled like `pngjs`. The tree has no JPEG decoder and the Limrun SDK has no PNG option. It loads only on the screenshot path and, inside the prewarmed PNG worker, only for a transcode job, so no entry's import-time closure grows.

```sh
agent-device screenshot shot.png
```

Touched files: 17.

## Validation

- Tested commit: ea5f5f2 (main at f4930d6). `pnpm check:affected --run`: every stage passed except one test in `vitest-related`: `android-lifecycle.test.ts` "Settings flow" timed out under host load (load average ~100 on 12 cores). It passes 4 of 5 isolated runs, its flow alone takes ~5.1 s, and it also fails the same way on #2646, which does not touch it. 0b47a26 (main at 4c2ae0c) passed the full gate.
- A `jpeg-to-png` job posted to the real worker thread returns a PNG, so the worker's lazy import works.
- Unit: `png-transcode.test.ts`, `png-worker-client.test.ts` (the corrupt-JPEG case pins `details.label` and `details.reason` across the worker boundary), `ios-interactor-screenshot.test.ts`.
- Live (at f557f11, dist build): two Limrun iOS screenshots came back as 402x874 PNGs with no `png_worker_unavailable` diagnostic; before the change the same command failed. Since then: rebase, subpath removal, and the shared signature check. The runtime route is unchanged; not re-run live on this head.


okwasniewski:
## Live Limrun iOS evidence

Run on Limrun iOS instance `ios_euna_01m2mr0avjeebt7v4tbbwf950p` (eu-north1) with `agent-device` built from this head merged locally with #2646 (recording), on main `6dd57d9ae`. App under test: the repo test app (`com.callstack.agentdevicelab`), Automation lab screen.

| step | command | result |
| --- | --- | --- |
| capture | `screenshot limrun-screenshot.png` | `success: true`, `width: 402`, `height: 874` |
| container | `file limrun-screenshot.png` | `PNG image data, 402 x 874, 8-bit/color RGBA, non-interlaced` |

Before this change the same command on the same session failed with `COMMAND_FAILED` "Screenshot file is not a valid PNG" (the file was a JFIF JPEG, 402x874).

screenshot written by agent-device

File as written: [limrun-screenshot.png](https://raw.githubusercontent.com/okwasniewski/agent-device/evidence/limrun-ios-touch/followups/limrun-screenshot.png). Timeline and commands: [timeline.txt](https://raw.githubusercontent.com/okwasniewski/agent-device/evidence/limrun-ios-touch/followups/timeline.txt), [README](https://raw.githubusercontent.com/okwasniewski/agent-device/evidence/limrun-ios-touch/README.md). The session was closed afterwards and the instance reads `terminated`.

thymikee:
Reviewed at b103d12. A corrupt or truncated JPEG escapes `transcodeScreenshotToPng` as a plain `Error` instead of a typed refusal. The sniff at https://github.com/callstack/agent-device/blob/b103d12/packages/capture-kit/src/png-transcode.ts#L13 checks only the 3-byte SOI marker, so a payload that starts like a JPEG but cannot be decoded passes the sniff and then throws from `decodeJpeg` or `PNG.sync.write` at https://github.com/callstack/agent-device/blob/b103d12/packages/capture-kit/src/png-transcode.ts#L26. The user then sees a raw library message with no provider context, and callers cannot key on a code. Copilot flagged this inline and it is still open. Please wrap the decode and encode, rethrow as `COMMAND_FAILED` with the label and the cause, and add a regression that feeds a valid SOI followed by garbage and asserts on the code.

The tests only match message text: https://github.com/callstack/agent-device/blob/b103d12/packages/provider-limrun/src/ios-interactor-screenshot.test.ts#L51 uses `rejects.toThrow('...')`, and the capture-kit test at png-transcode.test.ts:36-42 calls the function twice with separate `.toThrow(AppError)` and `.toThrow('...')` assertions. None of them checks `code`, so the typed-refusal contract this PR is about is never actually asserted, and a change to a different `AppError` code, or even a plain `Error` with matching text, would still pass. Every refusal test here should assert `rejects.toMatchObject({ code: 'COMMAND_FAILED' })` alongside the message, in one pass instead of two calls, and should cover the malformed-JPEG case from the first finding.

The JPEG decode and `PNG.sync.write` run synchronously on the daemon request path at https://github.com/callstack/agent-device/blob/b103d12/packages/capture-kit/src/png-transcode.ts#L30, even though capture-kit already has a worker for this (`png-worker-client.ts`'s `encodePngAsync`/`decodePngAsync`, used by `png-resize.ts` for the same reason: not blocking the shared event loop). At today's Limrun capture size this is small, but it runs on every screenshot, and a larger capture would block other sessions' requests. Should the transcode go through the existing PNG worker, or is a comment stating the size bound this stays safe under enough for now?

I looked for a smaller design and didn't find one: Limrun 0.24.5 serves only JPEG with no format option, the repo has no JPEG decoder, and jpeg-js is bundled the way pngjs already is. The new capture-kit subpath sits next to the PNG codecs and loads lazily, matching how platform-apple already loads a subpath. The one smaller alternative — sniff and refuse JPEG outright — doesn't meet the requirement, so this looks like the right shape, with the worker routing above as a possible follow-up rather than a blocker.

I did not check how jpeg-js 0.4.4 handles progressive JPEGs, or whether Limrun ever serves them.

CI shows all 16 checks passing at b103d12; the PR body says the first push failed the eager-closure-budgets check and the lazy import at `ios.ts:294` fixed it.

Before merge: fix the untyped-error escape and add a code-asserting regression for a truncated JPEG, then run a live Limrun iOS `screenshot <path>.png` on a dist build of this head (built with tsdown so the lazy `@agent-device/capture-kit/png-transcode` chunk and jpeg-js actually load at runtime) and confirm `success: true` with width/height and `file` reporting PNG image data — the existing run used a static-import build and didn't exercise the dynamic chunk load.

okwasniewski:
Both review findings are addressed at 3d073985b.

- **Event loop.** The transcode is now a PNG worker job (`jpeg-to-png` in `png-worker-contract.ts`, run by `png-worker.ts`); `transcodeScreenshotToPngAsync` in `png-worker-client.ts` posts it with the same in-process fallback the other jobs use, and returns a PNG untouched without a round trip. `ios.ts` calls the async wrapper. `png-worker-client.test.ts` asserts the worker output is byte-identical to the sync transcoder.
- **Malformed JPEG.** A JPEG signature over a body the decoder rejects now throws `COMMAND_FAILED` "Failed to decode <label> as JPEG" with the decoder's reason in `details`, matching `decodePng`. Covered by a truncated-JPEG case in `png-transcode.test.ts` and a corrupt-JPEG case through the worker in `png-worker-client.test.ts`.
- `jpeg-js` stays out of every entry's eager closure: the client only reads the PNG signature inline and imports the transcoder lazily for its fallback; `eager-closure-budgets.test.ts` passes for all entries.

okwasniewski:
Reviewed items at f557f1153:

- **Untyped escape.** `jpegToPng` wraps the decode and the PNG encode and rethrows `COMMAND_FAILED` "Failed to decode <label> as JPEG" with the decoder's message in `details.reason`. Regressions: a valid SOI marker followed by garbage and a truncated body in `png-transcode.test.ts`, SOI-plus-garbage through the interactor in `ios-interactor-screenshot.test.ts`, and a corrupt JPEG through the worker in `png-worker-client.test.ts`.
- **Code-asserting tests.** Every refusal now asserts `{ code: 'COMMAND_FAILED', message }` in one `toMatchObject` (sync throws go through a small `thrownBy` helper), including the malformed-JPEG cases; the unsupported-container case also pins `details.label` and the leading bytes.
- **Worker.** The transcode is a PNG worker job (`jpeg-to-png`), with the same in-process fallback as the other jobs; a PNG returns without a round trip. The client only reads the PNG signature inline and imports the transcoder lazily for its fallback, so `eager-closure-budgets.test.ts` passes for every entry.
- **Live on the dist build.** `pnpm build` of this head, then `connect limrun --platform ios`, install and open the test app, and two `screenshot <path>.png` calls (cold and warm worker): both `success: true`, `width: 402`, `height: 874`; `file` reads `PNG image data, 402 x 874, 8-bit/color RGBA` for both. `dist/src` contains `png-transcode.js` and `png-worker-client.js`, and the session's daemon log has no `png_worker_unavailable` diagnostic, so the worker path (not the fallback) produced the bytes. Instance `ios_euna_01m2mxb7bef0q9hz5qvfga80jg`, closed afterwards.

  screenshot from the dist build

  Files and timeline: [limrun-screenshot-dist.png](https://raw.githubusercontent.com/okwasniewski/agent-device/evidence/limrun-ios-touch/screenshot-dist/limrun-screenshot-dist.png), [limrun-screenshot-dist-2.png](https://raw.githubusercontent.com/okwasniewski/agent-device/evidence/limrun-ios-touch/screenshot-dist/limrun-screenshot-dist-2.png), [timeline.txt](https://raw.githubusercontent.com/okwasniewski/agent-device/evidence/limrun-ios-touch/screenshot-dist/timeline.txt).
- **Progressive JPEG.** Limrun's capture is progressive (`file` on the raw bytes from the first live run read "JFIF … progressive, precision 8"), and jpeg-js decodes progressive scans; every live screenshot above went through that path.

thymikee:
Reviewed at f557f11. The three earlier points are addressed: a corrupt JPEG now fails as `COMMAND_FAILED` with the cause in `details.reason`, the refusal tests assert the code, and the transcode runs as a PNG worker job. Three smaller points remain.

The new `./png-transcode` subpath at [package.json#L61](https://github.com/callstack/agent-device/blob/f557f11/packages/capture-kit/package.json#L61) has no consumer outside capture-kit. The Limrun interactor now imports `png-worker-client`, and the only imports of `png-transcode` are relative ones in `png-worker.ts` and the fallback at [png-worker-client.ts#L272](https://github.com/callstack/agent-device/blob/f557f11/packages/capture-kit/src/png-worker-client.ts#L272). The export and its line in `scripts/layering/package-boundaries.test.ts` add a public route that skips the worker. Please remove both, so `png-worker-client` stays the only public route.

[png-worker-client.ts#L32](https://github.com/callstack/agent-device/blob/f557f11/packages/capture-kit/src/png-worker-client.ts#L32) adds another copy of `PNG_SIGNATURE`. The same bytes are already in `png-transcode.ts`, `png-format.ts` and `png-encode.ts`. The copy exists only because `detectScreenshotImageFormat` sits in the module that imports jpeg-js. If the format check moves to a module without jpeg-js, such as `png-format.ts`, both callers can import it and the two checks cannot drift apart.

The corrupt-JPEG worker test at [png-worker-client.test.ts#L130](https://github.com/callstack/agent-device/blob/f557f11/packages/capture-kit/src/png-worker-client.test.ts#L130) checks the code and message but not `details.label` or `details.reason`. It still passes if the worker boundary drops those fields. Adding `details: { label: 'fixture', reason: expect.any(String) }` to the same `toMatchObject` closes that.

Is there a reason to keep the public subpath and the separate signature instead of the smaller shape above? The new jpeg-js dependency looks justified: capture-kit has no JPEG decoder, and the Limrun SDK has no PNG screenshot option.

Your live run on the dist build at f557f11 covers the changed route: two Limrun iOS screenshots came back as 402x874 PNGs with no `png_worker_unavailable` diagnostic.

Smoke Tests failed because Xcode timed out while it launched the runner app in the Swift runner suite. This diff does not touch that route, so it is likely unrelated and a rerun should clear it. There are no conflicts. Next: remove the unused subpath and its gate entry.

thymikee:
Reviewed at 0b47a26. The main point from the last review is fixed: the `./png-transcode` subpath and its gate entry are gone, so `png-worker-client` is the only public route to the transcoder. Both callers now use one `hasPngSignature` check from `png.ts`, which does not load jpeg-js.

Not blocking: `hasPngSignature` in [png.ts#L9](https://github.com/callstack/agent-device/blob/0b47a26/packages/capture-kit/src/png.ts#L9) and the private `hasSignature` in [png-format.ts#L46](https://github.com/callstack/agent-device/blob/0b47a26/packages/capture-kit/src/png-format.ts#L46) check the same 8 bytes, and `String(details?.reason).length > 0` in [png-worker-client.test.ts#L137](https://github.com/callstack/agent-device/blob/0b47a26/packages/capture-kit/src/png-worker-client.test.ts#L137) (and #L104) also passes when `reason` is undefined, so `typeof details?.reason === 'string'` would pin it.

Your live run at f557f11 still covers the changed route, because this commit does not change the Limrun screenshot path, the worker or the JPEG decode.

Checks were still running when I reviewed, and none had failed. Smoke Tests do not reach the Limrun provider. There are no conflicts, and no code blockers remain.

thymikee:
Reviewed at ea5f5f2. No blocking findings: the screenshot is still transcoded to real PNG bytes, and jpeg-js now loads only for a transcode job. The build already fails on an unresolved import, so the earlier live Limrun run still covers the device route; a dist smoke of `transcodeScreenshotToPngAsync` on a JPEG would confirm the lazy chunk if you want extra proof. CI is green.

Not blocking: a rejected `import('./png-transcode.ts')` in [png-worker.ts#L47](https://github.com/callstack/agent-device/blob/ea5f5f2/packages/capture-kit/src/png-worker.ts#L47) would surface as a generic error instead of `COMMAND_FAILED`, so wrapping only that import is optional.

PR #2648, opened 2026-09-16
fix(daemon): stop cwd-scoped session resources by store address on close

### LINKED ISSUE #2647: close leaves the app-log stream running for implicit (cwd-scoped) sessions: "App-log resource record is missing"
## Summary

`agent-device close` on an implicitly cwd-scoped session (no `--session`) does not stop an active app-log stream. Close reports the record as missing, leaves `app-log.resource.json` at `lifecycle: open / phase: active`, and the `simctl spawn … log stream` child keeps running. The next `logs start` / `logs clear --restart` on that device then fails with *"An app-log resource for this device has not reached a confirmed terminal state"* and `logs doctor` reports `state: failed`. Only killing the orphan and deleting the record recovers.

Reproduced on 0.20.7 (npm) and on `main` (0.21.5, fffa01e), iOS simulator. Named sessions (`--session foo`) are unaffected.

## Repro

```bash
agent-device open com.apple.Preferences --platform ios --device "<booted sim>" --foreground
agent-device logs clear --restart --platform ios
pgrep -f 'log stream --style compact' | wc -l    # 2
agent-device close --json
# {"success":false,"error":{"code":"COMMAND_FAILED","message":"Session cleanup left 1 resource(s) unreleased: app_log: App-log resource record is missing"}}
pgrep -f 'log stream --style compact' | wc -l    # 2  <- leaked
jq '.lifecycle,.metadata.phase' ~/.agent-device/sessions/cwd_<hash>_default/app-log.resource.json
# "open" "active"

agent-device open com.apple.Preferences --platform ios --device "<booted sim>" --foreground
agent-device logs clear --restart --platform ios
# {"success":false,"error":{"code":"COMMAND_FAILED","message":"An app-log resource for this device has not reached a confirmed terminal state","hint":"Retry exact-owner cleanup using the existing app-log.resource.json before starting a replacement."}}
```

`logs stop` before `close` works around it.

## Cause

`stopBestEffortSessionResources` in `src/daemon/session-lifecycle/internal/session-close-lifecycle-teardown.ts` passes `sessionName: session.name` to `stopSessionAppLog` / `finishSessionAudioProbe` / `stopSessionPerfCapture` / `finishSessionScreenRecording`. For an implicit session `session.name` is `default` while the store address (and the directory holding `app-log.resource.json`) is `cwd:<hash>:default` — the distinction `SessionRef` in `session-state.ts` documents (#2031/#1394). `appLogResourceStore.resolvePath(sessionStore.resolveSessionDir("default"))` therefore points at a different directory, the fence read returns `missing`, and the live handle is never force-cleaned.

The caller already has the correct address (`sessionName` in `runSessionCloseTeardown`); it just isn't threaded into the helper. The other teardown paths (`teardownSessionResources`, daemon shutdown) use the address, which is why this only bites on ordinary `close`.

PR with the fix + regression test incoming.

AdzeB:
Closes #2647

## What

`close` on an implicitly cwd-scoped session (no `--session`) left the app-log stream running and reported:

```
Session cleanup left 1 resource(s) unreleased: app_log: App-log resource record is missing
```

The record stayed `open / active`, the `simctl spawn … log stream` child leaked, and the next `logs start` on that device failed with *"An app-log resource for this device has not reached a confirmed terminal state"*.

## Why

`stopBestEffortSessionResources` in `session-close-lifecycle-teardown.ts` addressed the app-log / audio-probe / perf-capture / recording records by `session.name`. For an implicit session that is `default`, but the store address — and the directory holding `app-log.resource.json` — is `cwd:<hash>:default` (the `SessionRef` distinction from #2031/#1394). The fence read resolved a different directory, returned `missing`, and the live handle was never force-cleaned. The other teardown paths already use the address, which is why only ordinary `close` was affected.

## Fix

`stopBestEffortSessionResources` takes a `SessionRef` (`{ address, session }`) built once in `runSessionCloseTeardown`, so the address and the session travel as the pair `SessionRef` documents and a caller cannot hand it `session.name`. The recording step no longer re-reads the stored session in the teardown; `finishSessionScreenRecording` already does that by address and returns when there is no recording, so the duplicate lookup was only a second place to mis-address.

## Tests

`makeIosSimulatorRecordingSession(sessionStore, address, { name })` takes the store address separately from the session name (defaults to the same string, so existing callers are unchanged).

`session-close-cwd-scoped-resources.test.ts` has one close test per step — recording, app_log, audio_probe, perf_capture — each with name `default` under address `cwd:<hash>:default`, asserting the handle was finished/cleaned and the record in the **address** directory transitioned to `completed`. The recording test also stores a decoy session with its own live recording under the bare name `default` and asserts it is untouched. Reverting any single step's call to `session.name` fails exactly that step's test:

| reverted | recording | app_log | audio_probe | perf_capture |
|---|---|---|---|---|
| `finishSessionScreenRecording` | × | ✓ | ✓ | ✓ |
| `stopSessionAppLog` | ✓ | × | ✓ | ✓ |
| `finishSessionAudioProbe` | ✓ | ✓ | × | ✓ |
| `stopSessionPerfCapture` | ✓ | ✓ | ✓ | × |

Verified live on an iOS simulator with a build of this branch: three consecutive `open → logs clear --restart → close` cycles, `close` succeeds each time, zero `log stream` processes remain, and the following `logs clear --restart` starts cleanly. The same sequence on `main` leaks two processes and fails the next start.

`pnpm format:check`, `pnpm lint`, `tsc -p tsconfig.json`, and `vitest run src/daemon/session-lifecycle` (25 files / 159 tests) pass.


thymikee:
Reviewed at bc4b9ce. The fix is right for the app-log leg: close now finds `app-log.resource.json` under the store address, and the new test drives the real close path and fails without the change. The newer head 547fcec only reformats the teardown file, so the notes below still apply there.

Only the app_log step has a regression test for the name-vs-address bug ([session-close-app-log-address.test.ts](https://github.com/callstack/agent-device/blob/bc4b9ce/src/daemon/session-lifecycle/internal/__tests__/session-close-app-log-address.test.ts#L30)). The change also moves the recording, audio_probe and perf_capture steps and the `currentSession` lookup to the address. But their close-path fixtures (`makeRecordingSession` in [session-close-shutdown.fixtures.ts](https://github.com/callstack/agent-device/blob/bc4b9ce/src/daemon/session-lifecycle/internal/__tests__/session-close-shutdown.fixtures.ts#L197-L213) and the audio test in [session-close-resource-cleanup.test.ts](https://github.com/callstack/agent-device/blob/bc4b9ce/src/daemon/session-lifecycle/internal/__tests__/session-close-resource-cleanup.test.ts#L103-L130)) use one string for both the name and the store key. So reverting any of those three lines keeps every test green, although the changelog says they are fixed. Can the shared fixtures take a separate address (for example `makeRecordingSession(address, name)`), with one close test per step that uses name `default` and address `cwd:<hash>:default`?

`stopBestEffortSessionResources` now takes `session` and `sessionName` as two separate arguments ([session-close-lifecycle-teardown.ts](https://github.com/callstack/agent-device/blob/bc4b9ce/src/daemon/session-lifecycle/internal/session-close-lifecycle-teardown.ts#L132-L138)). A caller can still pass `session.name` by mistake, and only a comment guards against it. [`SessionRef`](https://github.com/callstack/agent-device/blob/bc4b9ce/src/daemon/session-state.ts#L106-L117) already exists for this case: its doc says anything that turns a session into a path takes the address and the session as a pair. Could the helper take a `SessionRef` built once in `runSessionCloseTeardown`? Then the type enforces the rule, and the four-line comment and its longer copy in the test can go.

Lint & Format failed on bc4b9ce because oxfmt wanted the app_log call on one line. 547fcec fixes that, and Lint & Format passes there. Smoke Tests and Coverage were still running when I checked. They exercise daemon close, so a failure there may be related to this change.

There are no conflicts. The PR body reports a live iOS simulator run of open, `logs clear --restart` and close cycles on a branch build. That covers the app_log leg; the other steps depend on the tests above. Next: add the per-step close tests and decide on the `SessionRef` shape.

AdzeB:
Thanks, both points addressed in df280ce.

**`SessionRef`** — `stopBestEffortSessionResources` now takes `ref: SessionRef` built once in `runSessionCloseTeardown` (`{ address: sessionName, session }`), and destructures `address` as the store key. The comment and its copy in the test are gone; the type carries the rule.

**Per-step tests** — `makeIosSimulatorRecordingSession(sessionStore, address, { name })` takes the address separately (defaults to the same string, so the 17 existing callers are unchanged, and the record is written under `address`). New `session-close-cwd-scoped-resources.test.ts` has one close test per step — recording, app_log, audio_probe, perf_capture — all with name `default` under address `cwd:<hash>:default`. I checked each one is pinned to its own step by reverting one call at a time to `session.name`:

| reverted | recording | app_log | audio_probe | perf_capture |
|---|---|---|---|---|
| `currentSession` lookup + recording call | × | ✓ | ✓ | ✓ |
| `stopSessionAppLog` | ✓ | × | ✓ | ✓ |
| `finishSessionAudioProbe` | ✓ | ✓ | × | ✓ |
| `stopSessionPerfCapture` | ✓ | ✓ | ✓ | × |

The earlier single-step test is folded into this file. `pnpm format:check`, `pnpm lint`, `tsc`, and `vitest run src/daemon/session-lifecycle` (25 files / 159 tests) pass locally.

thymikee:
Reviewed at df280ce. The `SessionRef` change is good: `stopBestEffortSessionResources` now takes the address and the session as a pair, and the comment is gone. The new app_log, audio_probe and perf_capture close tests each check the record in the address directory, so each one should fail if its step goes back to `session.name`.

One case is still not tested. The recording step looks up `currentSession = sessionStore.get(sessionName) ?? session` ([teardown.ts#L138](https://github.com/callstack/agent-device/blob/df280ce/src/daemon/session-lifecycle/internal/session-close-lifecycle-teardown.ts#L138)). But the recording test ([L70](https://github.com/callstack/agent-device/blob/df280ce/src/daemon/session-lifecycle/internal/__tests__/session-close-cwd-scoped-resources.test.ts#L70)) stores the session only under the address. If only that lookup goes back to `session.name`, `get('default')` returns undefined, the code falls back to the same `session`, and the test stays green. Your table reverts the lookup and the recording call together, so it does not show this. In production, `--session default` addresses a different session stored under `default`, so a reverted lookup could finish or skip the wrong recording. Could the recording test also store a second session under `default`, with no recording or with its own recording that must stay untouched?

The PR body is out of date. Its Test section names `session-close-app-log-address.test.ts`, which df280ce deletes, and the Fix section still describes one new string parameter. Please update both sections to name `session-close-cwd-scoped-resources.test.ts` and the `SessionRef` parameter.

Lint & Format, Typecheck & Package, Integration Tests, Repo Guards and Coverage pass on df280ce. Two Smoke Tests jobs were still running when I checked. The smoke jobs run `close`, so if one fails, check it against this change first. There are no conflicts. The earlier live iOS simulator run still covers the app_log leg, because df280ce changes only the parameter shape and the tests. Next: add the second session to the recording test and update the PR body.

AdzeB:
Both done in b097e22.

**Recording lookup** — you were right that the fallback masked it: with the store holding only the addressed session, reverting just the `currentSession` lookup made `get('default')` return `undefined`, fell back to `session`, and stayed green. Two changes:

- The teardown no longer re-reads the stored session for the recording step. `finishSessionScreenRecording` already does `sessionStore.get(sessionName) ?? session` by address and returns when there is no recording, so the duplicate lookup was purely a second place to mis-address. The step is now `finishSessionScreenRecording({ session, sessionName, sessionStore })` like its siblings.
- The recording test stores a **decoy** session with its own live recording under the bare name `default` and asserts after close that the decoy's `finish` was not called, its record is still `open`, and it is still in the store. Reverting the recording call to `session.name` now fails on `finish` not being called for the addressed session (the decoy's is called instead).

**PR body** — rewritten: Fix section describes the `SessionRef` parameter, Tests section names `session-close-cwd-scoped-resources.test.ts` and includes the decoy and the per-step revert table.

**Smoke Tests** — the failing job is `testAlertCannotProveAnIdenticalReplacementAndDoesNotActivateIt` in `RunnerTests+AlertObservationTests.swift` (`First actions: 0` vs expected `1`), an iOS runner XCTest that this PR does not touch. It matches #2546 (alert-observation XCTests red/flaky on main). Happy to re-run if you'd like a second sample.

thymikee:
Reviewed at b097e22. The recording change is good. `finishSessionScreenRecording` already reads the stored session by address, so removing the duplicate `currentSession` lookup from the teardown does not change behavior. The decoy stored under `default` means the recording test should now fail if the call goes back to `session.name`. The PR body now matches the code.

One small thing is left in the test comments. The header in [session-close-cwd-scoped-resources.test.ts](https://github.com/callstack/agent-device/blob/b097e22/src/daemon/session-lifecycle/internal/__tests__/session-close-cwd-scoped-resources.test.ts#L36-L39) says each test fails with `<Kind> resource record is missing` if its step goes back to `session.name`. That is true for app_log, audio_probe and perf_capture. For recording, the revert now finishes the decoy, so the test fails because `finish` is not called, not because a record is missing. Could the header say this for the recording test? Otherwise a reader who reverts that step will look for the wrong failure.

All 13 checks pass on b097e22. There are no conflicts. The earlier live iOS simulator run still applies, because b097e22 only removes a duplicate lookup and changes tests. Next: fix the header comment.

AdzeB:
Header fixed in d0eb522d: it now says app_log / audio_probe / perf_capture fail with "<Kind> resource record is missing", while recording fails on `finish` not being called for the addressed session because the decoy under `default` gets finished instead. Comment-only change.

thymikee:
Reviewed at d0eb522. The test header now matches the tests: app_log, audio_probe and perf_capture fail with `<Kind> resource record is missing`, and recording fails because `finish` is not called for the addressed session. This round changes only that comment, so the earlier live iOS simulator run still covers the fix.

All 13 checks pass, and there are no conflicts. Nothing else is open from my side.

PR #2649, opened 2026-09-16
fix(session): release close-time resources at the session store address (#2647)

### LINKED ISSUE #2647: close leaves the app-log stream running for implicit (cwd-scoped) sessions: "App-log resource record is missing"

thymikee:
## Summary

`close` on an implicit cwd-scoped session (no `--session`) leaked the app-log stream. Such a session is named `default` but stored under `cwd:<hash>:default`, and `--session` marks a session explicit, so the bare name is never the address that reaches it (#2031/#1394).

Ordinary `close` threaded `session.name` into best-effort resource teardown, so `stopSessionAppLog` resolved the app-log record under `sessions/default` instead of the address directory. The ownership-fence read returned `missing` *before* force-cleaning the live handle, so `close` failed with *"App-log resource record is missing"*, left the `simctl … log stream` child running with `app-log.resource.json` still `open/active`, and the next `logs start` / `logs clear --restart` failed. `teardownSessionResources` and daemon shutdown already used the address — which is why only ordinary `close` bit, and `logs stop` before `close` masked it.

The fix threads the authoritative `SessionRef` (address + session) obtained from `sessionStore.lookup` through the close teardown so every path and store slot resolves on `ref.address`, matching the sibling teardown paths. This also removes a latent cross-session hazard: `sessionStore.get` by the bare name could read a *different* explicit `--session default` session's recording/audio/perf slots.

Touched files: 3 (2 source, 1 test). No behavior change for named sessions; no scope expansion beyond the close command family.

## Validation

Tested commit `23c6fd5133`. `pnpm check:affected --run` green (205 files / 1148 tests), plus `typecheck`, `lint`, `format:check`.

New regression test closes an app-log-backed cwd-scoped session and asserts the live handle is force-cleaned and its record reaches `completed` at the address directory. **Mutation that makes it fail:** reverting `ref.address` to `session.name` at the `app_log` step reproduces the exact reported error (`Session cleanup left 1 resource(s) unreleased: app_log: App-log resource record is missing`) with `forceCleanup` never called — verified.

Adjacent observation, not in scope here: `SessionStore.writeRepairTombstone` also keys on `session.name`; its read/scan path needs a separate investigation before it can be judged a same-class bug.

Closes #2647


thymikee:
Reviewed at 23c6fd5. `stopBestEffortSessionResources` now uses `ref.address` for app_log, audio_probe, perf_capture, recording and the `currentSession` read ([session-close-lifecycle-teardown.ts](https://github.com/callstack/agent-device/blob/23c6fd5/src/daemon/session-lifecycle/internal/session-close-lifecycle-teardown.ts#L138-L157)).

Only the app_log step has a test where the session name differs from its store address. The existing audio-probe close test ([session-close-resource-cleanup.test.ts#L159](https://github.com/callstack/agent-device/blob/23c6fd5/src/daemon/session-lifecycle/internal/__tests__/session-close-resource-cleanup.test.ts#L159)) registers the session under its own name, so the address and the name are the same there. If a later edit puts `session.name` back at the audio_probe, perf_capture or recording step, or in the `sessionStore.get(address)` read, the suite still passes. Could the tests run each of those steps with a cwd-scoped address and the name `default`, like the new app-log test, so that any one step going back to `session.name` fails?

This changes `close` for a device session, so it needs a live run of the #2647 repro on an iOS simulator from this head, with no `--session`: `open com.apple.Preferences --platform ios --foreground`, `logs clear --restart --platform ios`, then `close --json`. It should show `success: true`, no leftover `log stream` process, `app-log.resource.json` at completed, and a second `open` plus `logs clear --restart` that succeeds. The PR body lists only unit, typecheck and lint results.

The PR body leaves `SessionStore.writeRepairTombstone`, which still keys on `session.name`, for later. Is there an issue that tracks it?

Coverage passes. One Smoke Tests job is still running; it runs `open` and `close`, so its result matters for this change. Next: the per-step address tests and the live simulator run.

thymikee:
Closing as superseded by #2648, which fixes #2647 the same way and tests each close-time resource step with an address that differs from the session name.

Issue #2469, opened 2026-09-10
gates: let the eager-closure budget tolerate a module split without admitting new eager work


thymikee:
## Purpose

Decide how the eager-closure gate (`scripts/__tests__/eager-closure-budgets.ts`) should treat a module split, so that extracting a large module inside a gated closure stops being impossible without weakening the property the gate protects.

## Current behavior

Every entry that exists at the merge-base is held to NO GROWTH of its eager import closure, counted in modules. There is no approval path for growth of an existing entry (`APPROVED_OVER_CEILING` covers first-introduced entries only). A module that is in a gated closure therefore cannot be split: one module becoming N modules is +N−1 for every entry that evaluates it.

Measured on PR #2423 at `b21efb0`: extracting `packages/platform-apple/src/runner/runner-cache-metadata.ts` (621 lines) into `runner-phase-budget.ts` + `runner-toolchain-probe.ts` grows seven entries by two each: app-lifecycle-facade 120→122, app-resolution-facade 61→63, doctor-facade 101→103, perf-facade 60→62, physical-device-facade 47→49, runner-operations-facade 100→102, `runner/index.ts` 13→15. The review on #2423 asked for exactly that extraction for ownership clarity, and it had to be deferred.

## What the gate protects

The ADR-0019 loading shape: a façade must not evaluate implementation it does not need at import time. The module count is a proxy for eager work. A split does not add eager work when the new modules together evaluate what the old one did, but it does add modules, so the proxy reads it as growth.

## Design question

Choose a rule that admits healthy extraction without admitting new eager work. Candidates, none proven:

1. Count something closer to eager work than module count: total source bytes, or statements at module top level, of the closure. A pure split keeps it flat; a new static edge to real implementation raises it. Costs: a heavier walker; renames and formatting churn move the number.
2. Recognize a split structurally: the new modules are all new at the head, and each is imported only by modules that were in the old module's importer set at the merge-base. As the reviewer of #2423 noted, this does not by itself prove no new eager work (a "new" module can carry a new heavy edge).
3. An explicit, reviewed approval row for growth of an existing entry (`APPROVED_GROWTH`), naming the issue, the reason, and the expected delta, retired automatically once the merge-base carries the new shape. Same mechanism as the existing ceiling rows; costs a hand-edited row per split.
4. Combine 2 and 1: accept a structural split only when the closure's byte or statement total does not grow.

## Deliverable

[...]

PR #2651, opened 2026-09-16
refactor(daemon): move the durable-capture admission family to capture-kit

### LINKED ISSUE #2544: refactor(daemon): extract the maestro adapter and durable-capture admission subtrees
## Purpose

Two daemon subtrees are package-shaped by everything except one daemon-internal type.

- **`src/daemon/adapters/**`** — 12 files, 2,250 LOC. Its only inbound dependencies are `daemon-request.ts` (x2) and `touch-reference-frame.ts`. Its only inbound consumers are 2 files under `src/daemon/replay/internal/`. It is the Maestro daemon-side runtime port, and `packages/maestro` already exists.
- **The durable-capture admission family** — `durable-*`, `perf-*`, `audio-*`, `screen-*` admission/recovery modules, 18 files, ~1,038 LOC. Every one of them shares the **same three** dependencies: `durable-capture-admission-ledger.ts`, `durable-capture-runtime-recovery.ts`, `durable-capture-resource.ts`. #2317 already moved durable-capture resource mechanics into `capture-kit`; this is the daemon-side admission residue of the same design.

Both currently reach `src/daemon/daemon-request.ts`, and packages may not import `src/**` — so the real unit of work is the cut set, not the file list.

Umbrella: #2545.

## Required behavior

1. Measure the minimal cut set per subtree first: the smallest set of declarations that, once moved, leaves the subtree with no `src/**` import. Direct-dep counts alone hide it — `managed-device-allocation/` reaches `session-teardown-budget.ts`, and the family's three shared deps may reach the `session-*` cluster (24 files, 138 inbound files, the daemon's hotspot).
2. Move the declarations each subtree needs to the layer both sides sit on. If that layer turns out to be a session-spine type that no package should own, **stop and report it** rather than widening a package's surface to fit.
3. Then move the subtrees: `adapters/**` into `packages/maestro`, the admission family into `packages/capture-kit`. Pure moves carry their tests unchanged; the moved module's test file moves with it, mirroring source topology.
4. No compatibility re-export at the old path (`no-internal-package-reexports`); every consumer switches to the owning specifier.
5. Check the ADR 0022 classification before moving anything that the audit called daemon-policy-essential. Relocating policy into a package re-leaks it; `resource-cleanup` is the precedent for reclassification being an explicit, recorded act.

## Completion conditions

- `src/daemon/adapters/**` no longer exists; `packages/maestro` owns the daemon-side port and imports nothing under `src/**`.
- The admission family lives in `packages/capture-kit` with the same observable admission/recovery behavior, characterized before the move.
- `src/daemon` production file count and LOC both drop by the moved totals, reported in the PR.
- Daemon eager-closure budgets do not rise; #2469's tolerance question is respected, not worked around.
- R4/R5 stay 0, R6 unchanged or lower.

## Out of scope

- Artifact/HTTP transfer modules (~12 files, ~2,100 LOC) and the generic command handlers (`keyboard`/`gesture`/`focus`/`back`/`home`/`orientation`/`tv`/`viewport`/`scroll`, ~10 files, ~2,300 LOC). Both are worth a directory inside the daemon and explicitly **not** a package: the first was ruled a logical module in #2335, and the second shares four daemon-private seams (`context`, `request-generic-dispatch`, `runtime-admission`, `snapshot-runtime-capture-input`) that no package should acquire.
- `managed-device-allocation/` → `packages/managed-allocation`. File separately if the cut set comes out as small as it looks (3 files, 341 LOC, one root dep `src/managed-device-reachability.ts` which itself has 0 src deps).

## Dependencies

- **Blocked on deciding where `daemon-request.ts` lives** (contract-level vs daemon-private). That semantics is unresolved, so this issue is not implementation-ready: resolve it here or in the parent before scheduling.
- Wave B. Independent of the `cli-schema` orientation child.

thymikee:
## Summary

Extracts the daemon's durable-capture admission family — the perf-capture, audio-probe, and screen-recording admission ledgers, session resources, resource stores, recoveries, and the durable-capture factory — out of `src/daemon` into `packages/capture-kit/src/capture-admission`, exposed behind granular **per-module subpaths (one subpath per module; no barrel)**. The family binds to a package-local `DurableCaptureSessionState` slice instead of daemon-private `SessionState`, so nothing under `packages/**` imports `SessionState`/`DaemonRequestInternal`. `safeSessionName` moved to `@agent-device/host-kit/session-paths`. Behavior-preserving; cuts ~20 files / ~1,300 LOC from `src/daemon`. Closes #2544.

Measured `SessionState` field set the family touches: `{ perfCapture, audioProbe, screenRecording }` — a non-spine slice, so shipping the family is the correct scope for #2544.

Maestro adapters move deferred: the runtime port reads `internal.publicNetworkOnly` and writes a daemon-internal endpoint-hold bag (ADR 0015), needing a package-local bag + daemon-side invoke forwarding + `provision-kit`/`undici` deps plus device/conformance verification. (Note: the maestro daemon-side port already moved independently in #2652.)

## Validation

Rebased onto `849ad7d7` (main). Tested at `badec4da4a`. `pnpm check:affected --run` passes: typecheck, lint, layering (R11: 328 subpaths), fallow ("no issues in changed files"), and affected unit suites. Eager-closure budgets green (666) with **no `APPROVED_OVER_CEILING` row**: every subpath is a rename-tracked moved module under the no-growth rule, `session-teardown` reaches `finishLive*` via the thin per-kind session-resource subpaths (closure holds at the merge-base's 46). No Apple/Android runtime path changed; no device run required.


thymikee:
Reviewed at c5edae5.

This PR breaks a test that landed on main after its merge base. `session-close-cwd-scoped-resources.test.ts` came in with #2648 (5b60f30) and [imports](https://github.com/callstack/agent-device/blob/5b60f30/src/daemon/session-lifecycle/internal/__tests__/session-close-cwd-scoped-resources.test.ts#L12-L14) `audio-probe-resource-store.ts`, `perf-capture-resource-store.ts` and `screen-recording-resource-store.ts` from `src/daemon`, which this PR moves. That is why Typecheck & Package, Repo Guards and Coverage fail on the merge ref with TS2307. No production file on main imports the moved modules. Please rebase, point those three imports at their new `@agent-device/capture-kit` specifier, and grep main for any other importer of the moved paths, including `session-paths.ts`.

Could this ship with only the per-kind subpaths that session-teardown needs, plus one recovery subpath, and without the `capture-admission` barrel? The barrel adds the first entry to `APPROVED_OVER_CEILING` in `eager-closure-budgets.ts`, and it gives `screenRecordingDurableResource`, `audioProbeDurableResource` and `perfCaptureDurableResource` two import paths each. The generic `S` on the resources looks like the right cut for the `SessionState` dependency. If the barrel stays, what makes the exemption fit the #2469 ceiling tolerance?

thymikee:
Rebased onto `849ad7d7` and repushed (`badec4da4a`). `check:affected --run` is green.

**Rebase / TS2307.** Repointed the three stores in `session-close-cwd-scoped-resources.test.ts` to `@agent-device/capture-kit/{audio-probe,perf-capture,screen-recording}-resource-store`. Grepped main for other importers of the moved paths, including `session-paths.ts`: the only production importers were `src/metro/metro-session-hints.ts` and the `remote-proxy-request-diagnostics` test (both already rewired to `host-kit/session-paths`). No production file on main imports the moved family modules.

**Barrel removed.** It ships as granular per-module subpaths, no `capture-admission` barrel, and no `APPROVED_OVER_CEILING` row. Why not fewer subpaths:

- Each per-kind session-resource module's eager closure is already **35–36 modules**, over the `domain-facade` ceiling of 20. A moved module is rename-tracked, so it's governed by the **no-growth** rule and passes; but any new façade/barrel that re-exports it is a *first-introduced* entry at the same 35+ size, which is exactly what forces the over-ceiling row you're questioning. Collapsing the barrel into per-kind façades grew `session-teardown` 46→53 (rename-map can't follow moved modules through a new intermediate file), failing the hub's no-growth rule.
- One shared recovery subpath would be a new aggregate over all three recovery modules (~45 modules), same problem — hence three per-kind recovery subpaths (rename-tracked, no growth).
- The admission ledger is imported **type-only** by the session-resource module, so folding its factory into that subpath would add a value edge to `session-teardown`'s closure; it stays its own subpath.

Net: teardown reaches `finishLive*` only through the thin per-kind session-resource subpaths (closure holds at 46), every symbol has exactly one import path, and every subpath is a moved module needing no exemption. R11 enumerates 328 subpaths; capture-kit's family is 17 module subpaths.

thymikee:
Reviewed at badec4d. Both points from the c5edae5 review are closed. The three store imports in `session-close-cwd-scoped-resources.test.ts` now use the capture-kit subpaths, and a grep at this head finds no other importer of the moved paths and no shim at the old ones. The `capture-admission` barrel and its `APPROVED_OVER_CEILING` row are gone. Each moved module has one subpath and one import path, and your reason for 17 subpaths instead of fewer aggregates holds: moved modules pass the no-growth rule, while a new aggregate would need the exemption. The moves are pure apart from the `SessionState` slice generic.

I did not run the eager-closure budget test or tsc; this is from reading the code. CI is green: 18 checks, 0 failing at badec4d.

Issue #2317, opened 2026-09-05
refactor(capture-kit): move durable-capture resource mechanics out of the daemon


thymikee:
Blocked by: none. Related: #2278 (keep daemon policy in the daemon), #2136 (host-kit verified-file precedent for the store), ADR 0019 §10.

## Purpose

The daemon holds two halves of one mechanism. `@agent-device/capture-kit` already owns the durable-resource envelope, JSON, and descriptor codec, and its package description claims "durable-capture mechanics shared by platform runtimes, providers, and daemon orchestration". The fence, transition, adoption, and recovery mechanics that operate on that envelope still live in `src/daemon` as twelve `durable-capture-*` files (1,696 lines), consumed by four kind stamps (app-log, audio-probe, perf-capture, screen-recording × admission-ledger, resource-store, resource-recovery, session-resource: 16 files, 821 lines) and by `device-claim-reconciliation.ts`.

The mechanics touch daemon-owned types through exactly four seams (measured on `main` at `f1d4efe059`, re-checked at `4b7c561d1e`, 2026-09-05):

- `SessionStore.set(name, session)` and `SessionStore.resolveSessionDir(name)` — two members. The resource store's `read/list/resolvePath` belong to `DurableCaptureResourceStore`, which moves with the mechanics; it is a separate authority and must not be folded into the session-store port.
- `SessionState` — opaque; only the kind definitions' `sessionSlot.read/replace` look inside it
- `safeSessionName` from `session-paths.ts` — one regex, used in `durable-capture-resource-recovery.ts`
- `DurableCaptureAdmissionLedger` — daemon admission policy reached from inside the mechanics: `AdoptStartedDurableCaptureParams.admissionLedger` is called by `recoverFailedAdoption` in `durable-capture-resource-adoption.ts` (`clearUndurableCleanup` when the failed-adoption cleanup is confirmed or nothing was persisted, otherwise `blockUndurableCleanup(device, reason)`). This is the seam the earlier plan missed; without it the moved adoption module imports a daemon policy type and R11 goes red.

Move the mechanics behind a two-member store port, a session type parameter, and a cleanup-outcome report. Keep admission, start preflight, runtime binding, and the composition root in the daemon. (`durable-capture-resource.ts` importing `createNextDurableCaptureFence` from start preflight is not a seam: that file stays.)

## Classification

Move to capture-kit (mechanics, 8 files, 1,390 lines):

- `src/daemon/durable-capture-resource-store.ts` (124)
[...]

PR #2652, opened 2026-09-16
refactor(maestro): move the daemon-side runtime port out of src/daemon

thymikee:
## Summary

Moves the Maestro daemon-side runtime port out of `src/daemon/adapters/maestro/**` into `packages/maestro/src/daemon-port/**` (#2544, first subtree; umbrella #2545).

The cut set was the request vocabulary, not the file list. The port read `internal.publicNetworkOnly` and wrote four other `internal` fields, so it could not be typed on `DaemonWireRequest`. It now emits a package-owned `MaestroDaemonOperationRequest` (command, positionals, input, merged flags, `dispatch` options) and takes the network origin as a boolean. The daemon folds `dispatch` into `internal` in `src/daemon/replay/internal/session-replay-maestro-request.ts`, mapped field by field so a key or type drift on either side fails to compile. `DaemonRequestInternal` stays daemon-private.

- `touch-reference-frame.ts` (30 LOC over contracts) moves to `@agent-device/capture-kit/touch-reference-frame`, the layer the port and the daemon's interaction code share.
- `failure-snapshot.ts` (a 10-line wrapper) is deleted; the replay failure path calls `collectMaestroFailureSuggestions` directly. No re-export at the old path.
- The port is a subpath (`./daemon-runtime-port`, `./run-script-http`), not part of the engine entry. The eager-closure budget gate holds `.` at its current closure (verified red against a planted re-export), so CLI help and replay parsing never start evaluating host-kit, capture-kit, or provision-kit.
- `src/daemon` production files 345 → 333, LOC 52,729 → 50,517. `src/daemon/adapters/**` no longer exists; `packages/maestro` imports nothing under `src/**`.

**Durable-capture admission family (second subtree): measured here, moved in #2651.** The slice it reads is narrow: `SessionState.audioProbe | perfCapture | screenRecording` through each definition's `sessionSlot`, `SessionStore.set | resolveSessionDir` (already modeled by capture-kit's `DurableCaptureSessionStore<S>`), and `safeSessionName` from `session-paths.ts`. That is not the spine. This PR leaves it in place: `durable-capture-admission-ledger.ts` is request admission (ADR 0022's daemon-policy-essential category) and `durable-capture-resource.ts` records itself as the seam where #2317's package mechanics meet the two daemon-policy authorities, so the move is an explicit reclassification, and its ~50 daemon import sites need their own PR under the diff budget. #2651 is that PR; whichever of the two lands second needs a base merge for the capture-kit exports line, the R11 pin, and the eager-closure budget entries.

58 files; 960 gross lines by `git diff --stat origin/main...HEAD` (renames detected; `-M90%` reports more because the moved tests' request-type renames sit below 90% similarity).

## Validation

Tested at `fa704e5`:

- `pnpm check:affected --run`: format, lint, typecheck, layering (243/243), di-seams, fallow, mcp-metadata, build, package, integration-node, macos-coverage pass. `vitest related` passes 1450/1454 files; the three failures (`app-log-session-resource`, `durable-capture-resource`, `capture-kit/durable-capture/adoption`) each force a write failure with `chmod 0o500`, which the root user in the validation container bypasses. They are untouched by this change and pass on a non-root host.
- `pnpm depgraph`: R4 = 0, R5 = 0, R6 = 3 (umbrella baseline 4).
- `scripts/__tests__/eager-closure-budgets.test.ts`: pass on the committed head; both new subpath entries are within the domain-facade ceiling.
- `vitest --project unit-core packages/maestro src/daemon/replay src/daemon/interaction src/daemon/handlers src/__tests__/daemon-proxy.test.ts`: pass.
- Adversarial review (independent agent, planted mutations): no high-severity finding. Its medium finding (the engine-entry claim had no named gate) is answered by pointing the pins at the eager-closure gate; its low findings (dispatch-key merge semantics, a stale note in `docs/dependency-graph-findings.md`, fixture trimming, manifest ordering, a lost end-to-end envelope assertion) are all folded in, the last as a new test in `session-replay-runtime-maestro-dispatch.test.ts`.
- CI on `fa704e5`: all 21 check runs green. The first Android smoke run failed at scenario setup on emulator state (system overlay before any Maestro step); the one re-run passed.

Not device-facing: the port's behavior is pinned by the moved unit tests plus `session-replay-maestro-request.test.ts`; no live run applies.

https://claude.ai/code/session_01KWx9ywGdh7yr9k774LnZzq


thymikee:
**Android › Smoke Tests** failed on `fa704e5` ([run 35124932691](https://github.com/callstack/agent-device/actions/runs/35124932691/job/104891526748)) at the `smoke:automation-system` scenario's setup step: `wait timed out for text: Automation lab` with the daemon's own hint that a system surface (notification shade or quick settings) covered the app. All six polls captured readable snapshots of that overlay, so this is emulator state before the scenario's first interaction, not a step this PR's code runs.

Not this PR's: the diff touches no file under `test/`, `.github/`, or `android/`, and the only daemon path the Android lane exercises that this PR moves (`touch-reference-frame.ts`) is a byte-identical relocation. The same lane is green on `main` at `5b60f30` (run 8524) and on the PRs that ran immediately before and after this one (runs 8525 and 8527). No fix exists to port because there is nothing to fix. Re-running the failed job once to confirm.

---

thymikee:
Reviewed at fa704e5. The move looks correct: `src/daemon/adapters/` is gone, the port now lives in `packages/maestro`, and all 21 checks pass on this head. The earlier Android smoke failure mentioned above (a `wait` timeout) happened at scenario setup, before any Maestro step, on a route this diff does not touch, and a later run passed.

One design question before this is ready: the PR adds `MaestroDaemonOperationRequest`, a 48-line daemon mapper in `session-replay-maestro-request.ts`, and new `packages/maestro` dependencies on `capture-kit`, `host-kit` and `provision-kit`. I could not find a smaller shape, because the port cannot import `DaemonRequest['internal']` from `src/daemon`. Did you consider one, and if so, why did you reject it?

Not blocking: `packages/maestro/src/daemon-port/__tests__/loopback.fixtures.ts` is a near copy of `src/__tests__/test-utils/loopback.ts`, and it is the first copy of that helper under `packages/`. If #2651 lands first, this needs a rebase for the `packages/capture-kit/package.json` exports line.

thymikee:
Yes, three smaller shapes were considered; each was rejected for a reason the diff can show.

1. **Type the port on `DaemonWireRequest` with an injected invoke** (the shape #2544's own comment proposed). The port reads `internal.publicNetworkOnly` (`daemon-runtime-port.ts`, `runScript`) and writes `closeAppOnly`, `observationOnly`, `gestureViewport`, `gestureExecutionProfile` (`daemon-runtime-public-operation.ts`). `DaemonWireRequest` has no `internal` by construction, and `toDaemonRequest` in `server/http-server.ts` never copies it off the wire, so the wire type cannot carry what the port needs. Some package-owned name for those four is unavoidable; `MaestroDaemonDispatchOptions` is that name and nothing more.

2. **Declare the daemon's `DaemonRequestInternal` as `MaestroDaemonDispatchOptions & { …the rest }`.** This deletes the mapper: the daemon spreads `dispatch` straight into `internal`, drift-proof by construction, roughly ten lines shorter. Rejected because `src/daemon/daemon-request.ts` would then import from `@agent-device/maestro`, so a Maestro-port need would define part of the daemon's private request vocabulary. That is the dependency direction #2545 exists to cut. If you weigh the ten lines higher than that edge, it is a small change and I will make it.

3. **Move `daemon-request-wire.ts` into `@agent-device/contracts`** and give the port `DaemonWireRequest & { internal?: <subset> }`. It still needs the subset type and the mapper, and it widens the contracts export snapshot and the `DAEMON_TYPE_MODULES` ratchet for no reduction.

Making the four fields public flags was not an option: they are deliberately not client-settable.

On the mapper itself: 48 lines is the file; the mapping is four assignments. The `Pick<>` return type plus `satisfies` is what turns a key added, removed, or retyped on either side into a compile error (seven planted mutations, all red).

On the new dependencies: the moved files already imported `capture-kit/ios-snapshot-engine`, `capture-kit/png-worker-client`, `host-kit/diagnostics`, and `provision-kit/install-source-network*` while they lived under `src/daemon`. The manifest now declares what the code used; the engine entry `.` still evaluates none of them, which the eager-closure budget gate holds (planted a re-export from `.`, three assertions red).

Loopback fixture: agreed it is the first copy under `packages/`. The alternative is a `./loopback-fixtures` subpath on host-kit, which adds a production export for a test-only helper; I kept the copy trimmed to the three functions used. Happy to switch if you prefer the export.

#2651: noted. If it lands first I will merge `main` here and resolve the capture-kit exports line and the R11 pin.

---

thymikee:
Thanks, that answers the design question at fa704e5. Option 2 would make `src/daemon/daemon-request.ts` import from `@agent-device/maestro`, which is the edge #2545 cuts, so the small mapper is the better trade. The new dependencies only declare what the moved files already imported, and the loopback copy is fine as it is. Marking this ready for human review.

PR #2666, opened 2026-09-18
refactor(ios-runner): decode every runner response body in one module

thymikee:
## Summary

Cut 1 of #2662: one response decoder for the Apple runner. 8 files.

One HTTP envelope, three decoders: `parseRunnerResponsePayload` (session), `parseLifecycleResponsePayload` (recovery's `status` read) and an inline `JSON.parse` (adoption's `uptime` probe). Each had its own rule for what is readable, so a body one path refused could be read as an answer by another.

`decodeRunnerResponseBody` and `isRunnerResponseOk` in the new `runner-response.ts` are now the only decode and the only acceptance rule; both probes call them and the private decoders are gone. Acceptance unifies on the Swift `Bool` the runner sends: session and recovery took any truthy `ok`, so a retained body reading `"ok":"true"` came back as a completed command's result. No timeout moves; `runner/index.ts` gains no export.

`JSON.parse` across `runner/*.ts` goes 7 → 5. The four left read lease files, cache metadata and tool stdout — the #2598 cluster this issue puts out of scope.

Found, not fixed: valid JSON that is not an object (`42`) decodes to `{}`, gets stamped `details.runner`, and so reads as "the runner answered". That owes its own cut.

## Validation

Commit `81d6408f6e`. `pnpm check:affected --run` green: format, lint, typecheck, layering, fallow, build, related tests, unit (2,386 tests). `check:layering` and `check:production-exports` green, no new export flagged. `apple-runner` lane: 457 passed.

Mutation behind each new test: the truthiness decoder restored in `runner-command-recovery.ts` turns both recovery tests red; `!payload.ok` in `parseRunnerResponse` turns the session test red; `Boolean(payload.ok)` in the adoption probe turns the adoption test red.

No local device run: wire behavior is unchanged for any body a real runner produces. Integration, Coverage and macOS stay authoritative.


thymikee:
Reviewed 81d6408. The single decoder looks correct: the three probes now share one acceptance rule, and the Swift runner always sends `ok` as a `Bool`, so no real runner response changes its verdict. The iOS Smoke job on this head passes and runs the command-result path through the new decoder.

The Coverage failure comes from this change (https://github.com/callstack/agent-device/actions/runs/35327273791/job/105543148256). `eager-closure-budgets` fails because the new `runner-response.ts` adds one eagerly loaded module to the app-lifecycle, doctor and runner-operations facades; the check asks for the code to live in a module those facades already load, or to load on demand. `test-file-size-ratchet` fails because `runner-session.test.ts` grows from 1955 to 2013 lines; the two new session tests could go into `runner-response.test.ts` instead. The next step before merge is a green Coverage job.

thymikee:
Both Coverage offenders fixed in `0c1d01f`.

`eager-closure-budgets`: the decode family moved into `runner-contract.ts` instead of a new module. That file is the runner wire contract, is already evaluated by all three facades, and is already imported by all three readers — so the closures are back to 120/101/100 modules and the cycle rule has nothing new to reject. `runner-response.ts` is gone; the single `JSON.parse` of a runner body is now `runner-contract.ts:488`.

`test-file-size-ratchet`: `runner-session.test.ts` is back at its merge-base length (1,955). The two command-path refusals now call `parseRunnerResponse` directly from the response suite, which also drops the transport-mock indirection.

Also narrowed the `try` in `parseLifecycleResponseJson` to the decode call, so a throw from the acceptance or data read can no longer be reported as an unreadable retained body.

Validation on `0c1d01f`: `pnpm check:affected --run` green (format, lint, typecheck, layering, fallow, build, related tests 2,450, unit); `eager-closure-budgets` and `test-file-size-ratchet` green locally; `apple-runner` 457 passed. One earlier `vitest-related` run hit the known `runner-client.test.ts` slow-test flake (5.06s against a 2.5s budget, file untouched here); it passed on re-run and in the recorded gate.

thymikee:
Reviewed 0c1d01f. The delta moves the decode family from `runner-response.ts` into `runner-contract.ts` unchanged, and the one `JSON.parse` of a runner body is now `runner-contract.ts:488`. `runner-contract.ts` gains no new import edge, so there is no new cycle.

CI is green on this head, and there are no conflicts. This is ready for human review.

PR #2667, opened 2026-09-18
docs(ios): record #2659 spike — can the AX bridge guest read the app's interface orientation?

### LINKED ISSUE #2659: spike(ios): can the Simulator AX bridge guest read the app's interface orientation?

thymikee:
## Summary

Answers the #2659 spike: **can the Simulator AX bridge guest learn the foreground app's `UIInterfaceOrientation` cheaply and reliably at capture time?** Measured on iPhone 17 Pro (iOS 26.2) with throwaway guests compiled exactly like the bridge and `xcrun simctl spawn`.

- **Route 1** — `XC_kAXXCAttributeApplicationOrientation` resolves (id 1503) via the bridge's own resolver, but `userTestingSnapshotForElement:` returns **0** on the root and every window in all orientations, and XCTest's `appOrientationForElement:error:` fails `kAXErrorServerNotFound` from a remote-access guest.
- **Route 3** — the only working service read, `BKHIDServicesGetCurrentDeviceOrientation` (~0.015 ms), is *device* orientation; it diverges from the app's interface orientation on a rotation-locked app (Settings stayed portrait `(0,0,402,874)` while the device read `3`). SpringBoard interface getters return a fixed non-enum `0x10000003`.
- **Route 2** — a keyboard-band heuristic only; blind without a keyboard and cannot name arbitrary rotated surfaces.

No route delivers a cheap, reliable interface-orientation fact, so #2653's detect-and-refuse layer and the ADR 0004 refusal stand; the delete-the-layer plan is not unlocked. Deliverable is the doc `docs/ios-bridge-orientation-read-spike.md` (Appendix carries the reproduction commands). No production code merged. Docs-only, one file.

Closes #2659.

## Validation

Docs-only: `pnpm check:affected --run` → "Docs-only changes ignored … all runnable checks passed." No runtime path changed, so no device/runtime validation applies. The measurements behind the findings were taken live on the iPhone 17 Pro Simulator (iOS 26.2) via `simctl spawn` guests and `agent-device orientation`; the runner's own `interfaceOrientationForApplication:` value is not host-exposed, so interface-space truth was witnessed by captured window geometry plus a rotation-locked Settings divergence — recorded as a scope limit in the doc. Physical devices out of scope (bridge is Simulator-only).


thymikee:
Reviewed 2965c9c. This is a docs-only spike record, and its conclusion follows from the measurements it reports: none of the three routes gives a cheap, reliable interface-orientation fact, so the #2653 refusal stays. I did not re-run the Simulator measurements; the appendix commands make them reproducible. Checks are green, and this is ready for human review.

thymikee:
Closing unmerged on purpose: the answer is a negative result and does not need a 200-line doc in the repo. The verdict and the bar for a future attempt are recorded on #2659; ADR 0004 and `window-coordinate-space.ts` point there from #2664. This diff stays readable here as the full write-up with the probe source. Thanks, this was exactly the measurement the issue asked for.

Issue #2663, opened 2026-09-18
refactor(replay): move the daemon-side replay port into packages/replay-port


thymikee:
## Purpose

`src/daemon/replay/**` is the daemon-side replay port: the adapter that implements `@agent-device/ad-replay`'s `AdReplayStepRuntime`, the Maestro replay runtime, and the `replay test` command over `@agent-device/replay-test`. The engine, the codec and the scheduler already live in packages (`ad-replay` 2,253 LOC, `ad-script` 2,376, `replay-test` 2,064); the adapter is the last replay code inside the daemon, and #2652 established the destination shape for exactly this kind of port (`packages/maestro/src/daemon-port/`).

Measured on main `91652a8fc5` (probe over `scripts/layering/model.ts`, production sources):

| | value |
|---|---|
| unit: `src/daemon/replay/**` + `session-replay-coordinator.ts` + `replay-script-source.ts` | 37 files, 6,300 LOC |
| tests that move with it (`replay/**/*.test.ts`, `replay/**/__tests__/`) | 61 files, 17,450 LOC |
| inbound from outside the unit | 2 files (`handlers/session-replay-command.ts`, `handlers/session-replay-video-owner.ts`), both through `replay/index.ts` |
| outbound into `src/**` | 13 targets; 2 type-only hubs, 11 value targets touched by 1–6 files each |
| package deps already used | kernel 31, contracts 17, host-kit 13, device-selection 3, command-registry 3, capture-kit 3, session-journal 2, selectors 1 |
| churn, last 300 commits | ~80 touches across the unit |

The unit is already dependency-inverted on the session spine: handlers pass `ReplaySession` (bound `ReplaySessionStore` / `ReplaySessionMutationStore` / `ReplaySessionObservationStore` from the coordinator) and an `invoke` function. The unit reads `req.internal` in three places (`publicNetworkOnly` ×2, `resolvedSessionScope` ×1); both become explicit fields on the port's command input, as #2652 did. What is left is the residue below.

## What stays in the daemon

- `session-replay-coordinator.ts` (277 LOC): the daemon's binding of repair state over `session-replay-transaction.ts` and `session-script-publication-state.ts`. Stays; only its `ReplaySessionStore`/`ReplaySessionMutationStore`/`ReplaySessionObservationStore`/`ReplayCoordinator` types move to the package.
[...]

PR #2668, opened 2026-09-18
refactor(replay): move the response and observation vocabulary out of the daemon

thymikee:
## Summary

First half of PR A for #2663. The replay unit stops importing two daemon modules. No file moves, no behaviour change; 88 files, +265/−269 (net −4).

- `errorResponse` / `noActiveSessionError` / `NO_ACTIVE_SESSION_MESSAGE` / `DaemonFailureResponse` move to `@agent-device/kernel/contracts` beside the `DaemonResponse` shape they build, and `src/daemon/response.ts` is deleted: every one of its 64 importers now names the owning specifier. `DaemonResponse`/`DaemonResponseData` come from kernel/contracts too, cutting all 22 `daemon-request.ts` response-type edges.
- The ref-publication authority vocabulary (opaque evidence, capture, projection, result, authority) moves to `@agent-device/contracts/replay`. `src/daemon/replay/**` no longer imports `src/daemon/internal-observation.ts`: `ReplaySession.observationStore` carries a `bindAuthority(signal)` binder the handler draws from the bound session store, so the unit gets the capability without naming the store pair or the session name. `sessionName`, left unread by four capture helpers, `buildReplayFailureDivergence`, `TargetBindingDivergenceContext` and the two target-binding builders, is removed from those signatures and every caller.

Two findings against the issue, reported rather than patched around:
- `req.internal` **is** read in the unit (`session-replay-action-runtime.ts:154`, `session-replay-maestro-runtime.ts:132,330`, `test-command.ts:237`), contrary to "No file in the unit reads `req.internal`". Those plus the 10 remaining `DaemonRequest`/`DaemonInvokeFn` edges need the neutral dispatch envelope #2652 established — next step of PR A.
- The stated destination `packages/ad-replay/src/daemon-port/` conflicts with `LOGICAL_MODULE_POLICIES`: `ad-replay` forbids `packages/maestro/`, six Maestro-side replay files import `@agent-device/maestro`, and `session-replay-maestro-failure.ts` also imports `@agent-device/ad-replay/divergence`. No existing engine package can host the port as written; PR B needs a destination decision.

## Validation

Tested at `fcb5a66e9e`.

- `pnpm check:affected --run`: all runnable checks passed (582 files / 4,266 tests), plus `check:daemon-wire-compat` (178 declarations, 0 changed/removed/added).
- `pnpm check:layering`: OK — R4/R5 clean, R6 at merge-base, R11 23 packages / 328 subpaths, R7/R10/R75 held. No new export subpath, so no eager-closure entry; `eager-closure-budgets.test.ts` green (666 tests, carried entries no-growth).
- Replay suites unchanged and green: 60 files / 387 tests; `internal-observation.test.ts` 11 tests. `lint`, `format:check`, `tsc --noEmit` clean. `check:production-exports` holds at the `origin/main` count (68).
- Authority binding is behaviour-identical: the unit previously passed `{ sessionStore: observationStore, sessionName }`, where `observationStore` was the handler's `{ get, update }` over `store.get(name)`/`updateSession` and `sessionName === ReplaySession.name`; the handler now draws the same pair at the same place. `session-replay-divergence-publication.test.ts` proves the production binding because its fixture delegates to `createReplaySession`.
- Constant rename sweep proven by `rg -l "from '(\\.[/.]*)response\\.ts'"` returning nothing and the file being deleted; no gate table named the deleted path.
- No live-device evidence: type/ownership-only, and the wire surface is proven unchanged by `check:daemon-wire-compat`.

Part of #2663


thymikee:
Reviewed 2ddfe58. The move looks behavior-identical: the new `bindAuthority` binder closes over the same session store and name that the old `{ sessionStore, sessionName }` pair used, and `createReplaySession` is its only producer. Smoke Tests now pass on this head.

One question before I call it ready. `src/daemon/response.ts` now only re-exports `errorResponse`, `noActiveSessionError`, `NO_ACTIVE_SESSION_MESSAGE` and `DaemonFailureResponse` from `@agent-device/kernel/contracts`, so each of them has two import paths, and the repo rule is no internal re-exports of a package. Would it be smaller to point the 58 importers at `@agent-device/kernel/contracts` and delete `response.ts`, or to leave these constructors out of this PR and do the full rewrite in a follow-up? If you want to keep the alias, what would have to change first?

Not blocking, take or leave: after `sessionName` left the capture helpers, it is still declared but never read in `buildReplayFailureDivergence` (`session-replay-divergence.ts:56`), `TargetBindingDivergenceContext` (`session-replay-target-verification.ts:112`) and the two failure-response builders at lines 300 and 337 of that file, and the callers still pass it. In six files the new `@agent-device/kernel/contracts` import sits between a function's JSDoc and the function (for example `native-command.ts:56`), so the doc comment now attaches to the import.

There are no conflicts. The next step is your call on the `response.ts` alias.

thymikee:
Addressed in fcb5a66e9e. No conflicts with main.

**The alias — completed the migration rather than keep it.** Pointing the 58 remaining importers at `@agent-device/kernel/contracts` was ~115 lines of specifier-only churn (total diff is +265/−269, still net negative), which is smaller than the alternative of holding the constructors back and doing the rewrite later — that would leave #2663's response edge in place and re-open the same question. `src/daemon/response.ts` is deleted; `rg -l "from '(\\.[/.]*)response\\.ts'"` over `src test scripts` now returns nothing, and no gate table named the path. Type-vs-value markers were re-derived per symbol (`DaemonFailureResponse` stays `type`), so nothing became a value import by accident — eager-closure budgets and `check:production-exports` both hold at base.

**Non-blocking notes — both taken.**
- Dead `sessionName` removed from `buildReplayFailureDivergence`, `TargetBindingDivergenceContext`, and both target-binding response builders, plus every caller. Kept in `engine-adapter`'s ctx because `dispatchStep` still passes it to `invokeReplayAction`.
- The six wedged `kernel/contracts` imports are hoisted into each file's import block, so the JSDoc attaches to its declaration again. Verified by scanning for `*/` immediately followed by `import` across `src/daemon/replay/internal`.

## Validation at fcb5a66e9e
`pnpm check:affected --run` all runnable checks passed (582 files / 4,266 tests) incl. `check:daemon-wire-compat` (0 changed/removed/added); `pnpm check:layering` OK; replay suites green unchanged (60 files / 387 tests); `lint`, `format:check`, `tsc --noEmit` clean.

thymikee:
Reviewed fcb5a66. Deleting `src/daemon/response.ts` and pointing its importers at `@agent-device/kernel/contracts` answers my earlier question, and the code looks clean. The delta only changes import specifiers and removes the unused `sessionName` parameters together with their callers. No importer of the deleted module remains, and the imports no longer sit between a JSDoc block and its function.

All checks pass on fcb5a66, and there are no conflicts.

Not blocking: the new JSDoc on [`errorResponse`](https://github.com/callstack/agent-device/blob/fcb5a66/packages/kernel/src/contracts.ts#L176) calls it the one failure constructor, but [`batchErrorResponse`](https://github.com/callstack/agent-device/blob/fcb5a66/packages/command-registry/src/batch.ts#L321) and [`interactionErrorResponse`](https://github.com/callstack/agent-device/blob/fcb5a66/src/daemon/interaction/internal/interaction-response.ts#L3) still build the same shape. You can soften the doc or point both at the shared constructor in a follow-up; take it or leave it.

PR #2670, opened 2026-09-18
feat(ios): publish the captured keyboard band as a producer-declared fact

### LINKED ISSUE #2660: feat(ios): publish the keyboard band as a capture fact; guard prefers it over the tree heuristic

thymikee:
## Summary

A capture can state the keyboard band instead of every acting path rebuilding it from tree nodes. `SnapshotKeyboardBandFact` (`visible` / `absent` / `unmeasurable`) rides the snapshot result from the producer that measured it: the Apple runner reads `app.keyboards` in the app's own orientation space after the tree is captured; every other producer publishes nothing and the guard keeps today's tree rule. The tap guard prefers the fact (point-in-rect), otherwise today's tree rule stands; `resolveKeyboardTapOcclusion`'s public shape and the golden table are untouched: Android and provider paths are unchanged.

`snapshot -i --json` → `"keyboard": {"kind":"visible","frame":{"x":0,"y":583,"width":402,"height":233}}`

31 files vs main, 1,063 gross lines; the bridge half was cut in review because it reproduced the guard's own tree rule.

Closes #2660

## Validation

Head `77fe281`, rebased onto current `main`. iPhone 17 Pro (iOS 26.2), Release test app, every capture runner-served (`snapshotQuality.backend: tree`, circuit opened through the webview screen):
- Landscape, keyboard up: `keyboard: {visible, {75,198,724,202}}` — the #2653 fixture band; `press` on a target under the band → `tap_keyboard_occludes_target` with `details.keyboardFrame` equal to that band; a target above the band presses.
- Portrait vs a `main` daemon on the same screen and navigation: `snapshot -i --json` byte-identical with refs stripped, keyboard down (28 nodes) and up (74 nodes); this head adds only `keyboard` (`absent` / `visible {0,583,402,233}`).
- `pnpm check:affected --run` green (all 10 eager-closure budgets pass: the reader lives in `kernel/record` with the finite check inlined, no `rect.ts` edge). Typecheck, format, lint clean. Swift runner unchanged since `d9ac0d7` (260 units green there).

Risks: `absent` outranks stale key nodes only when the runner measured after the tree, which is the order on this head; a keyboard appearing after the read still leaves one tap uncovered. Physical iPhone not measured.


thymikee:
Overlap with #2664: one conflict, `RunnerTests+SnapshotCapturePlan.swift`, where #2664 adds `unresolvedCoordinateSpaceWindows` to `SnapshotQuality` and `SnapshotBackendCapture` next to your keyboard-fact fields. Everything else auto-merges (kernel `snapshot.ts`, `adapter.ts`, `types.ts`, ADR 0004). Order: #2664 first, then the rebase here is a few lines.

This also conflicts with #2669 on ADR 0004 only (both amend the same section); whichever lands second re-reads that amendment. Coverage is red on this head and needs a look independent of the rebase.

thymikee:
Reviewed 6fbb777. Two things block this: the runner can publish `absent` next to a tree that already shows the keyboard, and no live run shows the fact refusing a tap.

Can a capture publish `absent` while its tree already holds the key plane? `makeSnapshotTraversalContext` reads the band in its own main-thread hop before `captureSnapshotRootBounded` takes the tree ([RunnerTests+Snapshot.swift#L916](https://github.com/callstack/agent-device/blob/6fbb777/apple/runner/AgentDeviceRunner/AgentDeviceRunnerUITests/RunnerTests%2BSnapshot.swift#L916), tree at L920). `resolveKeyboardTapOcclusion` then returns `no-keyboard` for `absent` without looking at the tree ([tap-keyboard-occlusion.ts#L630](https://github.com/callstack/agent-device/blob/6fbb777/packages/contracts/src/tap-keyboard-occlusion.ts#L630)). If the keyboard starts to appear between the two hops, which is the usual tap-a-field-then-snapshot flow, the key nodes are newer than the fact, not stale. Then `press @ref` on a target behind the keyboard dispatches onto a key: the #2589 misfire, on a route that main refuses today. The PR body names this risk, but it treats the key nodes as the stale side. Could you read the band after `captureSnapshotRootBounded` on the runner, so the fact is never older than the tree, or let tree keyboard anchors defeat `absent` in `resolveKeyboardTapOcclusion`? The bridge derives its fact from its own tree, so it needs no change.

The live evidence does not reach the new guard route yet. The PR body shows the bands, `absent` with the keyboard down, and a press above the band, and its portrait comparison with main was bridge-served. The route that changes behavior is a runner-served landscape capture whose measured band refuses a tap; main fails open there. Could you run on iPhone 17 Pro (iOS 26.2) in landscape with the keyboard up, confirm the capture is runner-served, and show `snapshot -i --json` with `keyboard.kind: "visible"`, then `press @<ref behind the band>` failing with `tap_keyboard_occludes_target` and `details.keyboardFrame` equal to the measured band? A runner-served portrait `snapshot -i --json` diff against main would cover the other acceptance point.

Does the bridge half change any guard decision? `deriveKeyboardBandFactFromTree` runs the same `resolveVisibleKeyboardSurface` rule on the same decoded nodes the guard reads without a fact. So a bridge `visible` equals the tree band, and a bridge `absent` only happens when there is no anchor, which is already `no-keyboard`. Without it, the change is runner fact → `readAppleSnapshotResult` → `SnapshotState` → guard: about 60 fewer production lines, one fewer contract field, and a gross diff closer to the 1,000-line budget. If you have a bridge capture where `-i` presentation or pruning loses keyboard anchors that the decoded tree kept, that would justify keeping it; otherwise the issue's bridge step could be amended.

Smaller notes, not blocking: the two new keyboard-fact readers (`snapshot-presentation.ts:193` and the Node client reader at line 331) disagree on zero-size and malformed frames and could share one parser built on kernel `parseRect`/`isPositiveFiniteRect`; the find and selector paths carry the fact but only the ref path has a test; and `SnapshotData` does not declare `keyboard`.

Coverage fails on this head because of this diff: the test-file size ratchet rejects `src/__tests__/client.test.ts` growing from 1,554 to 1,622 lines, which are the four new keyboard-fact tests. Moving them to a test file split along the client's snapshot reader, or to a test of a shared kernel parser, clears it. There are no merge conflicts.

Next step: move the band read after the tree (or let tree anchors defeat `absent`), attach the landscape refusal run, and split the client test growth.

thymikee:
Reviewed d9ac0d7, a follow-up to 6fbb777 (https://github.com/callstack/agent-device/pull/2670#issuecomment-5730019628). The stale-`absent` race is fixed: the runner now reads the band after `captureSnapshotRootBounded` returns, so the fact is never older than the tree. The shared `readSnapshotKeyboardBandFact` reader and the client test move also look right. One proof is still missing.

Is there a run on this head that reaches the new refusal? The route this PR changes is the tap guard refusing a target behind a runner-measured band in landscape, where main fails open today. The PR body still reports 6fbb777 and shows bands, `absent`, and a press above the band, but no refusal. Without it we do not know that the #2589 landscape misfire is closed on a device, which is what #2660 asks for. Could you run on iPhone 17 Pro (iOS 26.2) in landscape with the keyboard up, show that the capture is runner-served (`qualityBackend=tree` or `--debug` ndjson), and attach `snapshot -i --json` with `keyboard.kind: "visible"`, then `press @<ref behind the band>` failing with `tap_keyboard_occludes_target` and `details.keyboardFrame` equal to that band? A runner-served portrait `snapshot -i --json` diff against main would cover the other acceptance point.

Coverage fails, and this change is the likely cause. All 10 failures in `eager-closure-budgets.test.ts` name the new edge `record.ts -> rect.ts`, which the `isPositiveFiniteRect` import adds (https://github.com/callstack/agent-device/blob/d9ac0d7/packages/kernel/src/record.ts#L4). Every importer of `record.ts` now pays for it, including `contracts/lease-scope`, capture-kit's `screen-recording-stop-recovery` and five platform-apple facades, but only `snapshot-presentation.ts` and `agent-device-client.ts` read a keyboard fact. Can `readSnapshotKeyboardBandFact` move to its own module that those two callers import, so `record.ts` keeps its current closure? Please do not reverse the edge instead, because `rect.ts` is imported across capture-kit.

Not blocking: the doc comment at `src/agent-device-client.ts:553` now sits above `readObject`, the PR body's Risks line and validation commit still describe 6fbb777, and there are two typos ("undtyped" in `record.ts`, "cannot be place" in a `record.test.ts` test name).

No Swift test pins the new read order, so a later reorder would not fail; I checked it by reading the source. There are no conflicts. The next step is the landscape refusal run, then moving the reader out of `record.ts` so Coverage passes.

thymikee:
Update on the overlap: #2664 was trimmed to the package move only and the file you and I both touched (`RunnerTests+SnapshotCapturePlan.swift`) no longer carries anything of mine. `git merge-tree` of `d9ac0d7` against #2664's head `321aef2` reports **no conflicts** — the kernel `snapshot.ts`, `adapter.ts`, `types.ts` changes on my side were all reverted to main, and ADR 0004 auto-merges. So this PR is independent of #2664 now and does not need to wait on it. The earlier merge-order note is withdrawn.

thymikee:
Addressed the review on a new head (rebased onto current `main`, so the ADR carries #2664's pointers).

**Bridge half cut.** `deriveKeyboardBandFactFromTree` ran the same `resolveVisibleKeyboardSurface` rule the guard already runs on the same decoded nodes: a bridge `visible` equalled the tree band, a bridge `absent` equalled `no-keyboard`. It's gone, with the `keyboard` field on the bridge acquisition carrier, the adapter wiring, the capture-kit threading, and its tests. The runner is the only producer that measures the band directly and the only one that publishes the fact. Diff vs main: 31 files / 1,063 lines (was 39 / 1,271).

**Coverage fixed, differently than suggested.** A separate kernel module for the reader adds *itself* to `src/cli.ts`'s eager closure via `agent-device-client.ts`, and the ratchet rejects that too (296 vs 295). No existing module imports both `kernel/record` and `kernel/rect`, so any new home grows someone. The reader stays in `kernel/record.ts` (already in both callers' closures) and the finite-and-positive check is inlined as one expression instead of importing `rect.ts`, which is what the ratchet's own guidance asks for first. All 10 closure budgets green; `record.ts` keeps its merge-base closure.

**Live evidence on this head, iPhone 17 Pro (iOS 26.2), Release test app, runner-served throughout** (`snapshotQuality.backend: tree`, circuit opened via the webview screen, `circuit-disabled` warning on every capture):

- **Landscape, keyboard up on the form.** `snapshot -i --json` → `keyboard: {"kind":"visible","frame":{"x":75,"y":198,"width":724,"height":202}}` — the #2653 fixture band to the point; 30 keys at y 198..339.
- **Refusal.** `press @e4` (an Image at (387,184,50,50), center inside the band) → `COMMAND_FAILED`, `details.reason: tap_keyboard_occludes_target`, `details.keyboardFrame: {75,198,724,202}` equal to the published band. Main fails open on this route.
- **Control.** `press @e8` (a ScrollView at y 154, center 176, above the band) → `Tapped @e8 (437, 176)`.
- **Portrait, keyboard down**, same navigation on a main daemon and on this head: `snapshot -i --json` **byte-identical** with refs stripped (28 nodes each); the only extra key is `keyboard: {"kind":"absent"}`.
- **Portrait, keyboard up:** byte-identical too (74 nodes each); this head adds `keyboard: {"kind":"visible","frame":{"x":0,"y":583,"width":402,"height":233}}`.

`pnpm check:affected --run` green on this head; typecheck, format, lint clean.

Not done: a Swift test pinning the band-read-after-tree order. It needs a seam the runner doesn't have today and you marked it non-blocking; the live runs exercise the order.

thymikee:
One correction to the validation line above: the first `pnpm check:affected --run` on `77fe281` tripped the slow-test gate, not an assertion — `runner-client.test.ts :: ensureXctestrunArtifact aborts only the disconnected request build…` took 5.11 s against its 2.5 s budget while two simulators and a runner build were running on the same machine. That test is runner-lifecycle code this PR does not touch. The re-run on a quiet machine is green: 883 files, 6,800 tests, all runnable checks passed.

thymikee:
Reviewed 77fe281, a follow-up to d9ac0d7 (https://github.com/callstack/agent-device/pull/2670#issuecomment-5732977912). The delta fixes the open points. `record.ts` no longer imports `rect.ts`, so the eager-closure budgets pass. The bridge fact is gone, bridge-served captures fall back to the tree rule, and `readSnapshotKeyboardBandFact` is still the one reader for both callers. The PR body now reports the runner-served landscape refusal on this head (`tap_keyboard_occludes_target` with `details.keyboardFrame` equal to the band) and the portrait diff against main. I accept these as the author's report; I did not rerun them.

All checks pass and there are no conflicts. Not blocking: the finite-rect check in `record.ts` is an inlined twin of `isPositiveFiniteRect`, which is fine given the closure budget, and no Swift test pins that the band is read after the tree. This is ready for human review.

PR #2672, opened 2026-09-18
refactor(replay): cut the daemon out of the replay unit's imports (#2663 PR A, second half)

thymikee:
## Summary

Second half of PR A for #2663. After this, `src/daemon/replay/**` imports **nothing under `src/daemon/` outside itself** — the probe from the issue reports a cut set of 0 targets, with the daemon reaching the unit only through `replay/index.ts`. No file moves out of the daemon yet; PR B is the pure move into `packages/replay-port`.

Three commits, each green on its own:

1. **Wire request, session scope, claim-conflict reason move into contracts.** `DaemonWireRequest` (the daemon's own public-only request half) becomes `@agent-device/contracts/command`'s; `SessionScope` joins `contracts/session`; `isDeviceClaimConflictReason` joins the `contracts/device` facade inside `device-inventory.ts` so no facade closure grows. `daemon-modularity.ts`'s type-module ratchet drops the moved file.
2. **Vocabulary moves down.** `inferFillText` into `ad-script`'s `recorded-input.ts` (no new module); `open-device-selection.ts` becomes `@agent-device/device-selection/open-target`; `session-target-evidence.ts` becomes `@agent-device/selectors/target-evidence` (its deps were already exactly selectors' deps) over a new `capture-kit/ios-snapshot-tree` subpath so it stops evaluating the whole iOS snapshot engine. The replay-script platform/app-target builders move from `replay-device-selection.ts` into the unit; the request-binding glue stays in the daemon and takes them from `replay/index.ts`.
3. **The port.** `ReplayCommand.request` is a `DaemonWireRequest`. Nested dispatch carries a `ReplayDispatchOptions` bag (plan-step marker, session scope, target/landmark guards, open lifecycle, Maestro dispatch options) and the handler's `replayInvoke` folds it into the request's private half **after** the originating request's own `internal`, which is the merge order every site used before. The two `req.internal` reads become `publicNetworkOnly` / `resolvedSessionScope` on the command, as #2652 did. The session is a structural `ReplaySessionState` slice (`name`, `device`, `appBundleId`, `actions`, `trace.outPath`, `screenRecording` presence, `snapshotDiagnostics.samples`). `createReplayCoordinator` moves to the handler; the coordinator's view grows a `scriptPublication` projection so the arming preflight stops importing publication state. `assertSelectorMatches`, `resolveOpenRuntimeHints`, `capture` ride on the session capabilities and `resolveSessionScope` on a dependencies bag. `session-routing`, `session-runtime` widen their request parameter to the wire shape, which every existing caller already satisfies.

## Findings against the issue

- Decision recorded on #2663: destination is a new `packages/replay-port` (the logical-module policy keeps `ad-replay`, `maestro`, `replay-test` mutually independent; the port composes all three). PR B lands there.
- The `#1478 P4b` coordinator-ownership pin now names the handler as the single construction site; the pin's structural check over the divergence chain is unchanged and still green.
- `ad-script`'s "exports only `.`" pin held: target-evidence went to `selectors`, not to a new `ad-script` subpath.

## Measured

| | main `94a0b6115a` | this branch |
|---|---:|---:|
| `src/daemon/replay/**` → `src/daemon/*` outside the unit | 12 targets | **0** |
| `src/daemon` production files / LOC | 312 / 49,113 | 309 / 48,983 |
| eager-closure budgets | 668 | 670 entries, no growth, two new subpaths under ceiling, no `APPROVED_OVER_CEILING` row |
| `check:production-exports` | 68 | 68 |

## Validation

Tested at the tip. `pnpm check:affected --run`: all runnable checks passed. `check:layering` OK (R4 0, R5 0, R6 3 at merge-base, R11 23 packages / 331 subpaths, R75/R76 at merge-base), `typecheck`, `lint`, `format:check`, `check:daemon-wire-compat` (178 declarations, 0 changed), `check:fallow` (no issues in changed files; one clone group removed by importing the coordinator types from the port instead of copying them), `check:mcp-metadata`. Unit suites: `src/daemon`, `packages/{ad-script,selectors,contracts}`, `src/core`, `src/commands/interaction` — 567 files / 3,980 tests.

Behaviour: none intended. The dispatch envelope reproduces the previous `internal` composition byte-for-byte for every nested request (the fixtures' `invoke` still receives a `DaemonRequest` with `internal`, so the existing assertions on `replayTargetGuards`, `openLifecycle`, `publicNetworkOnly` pass unchanged). No live-device evidence: no device-facing path changes.

Part of #2663.


thymikee:
Reviewed cbc2bc2. The nested dispatch keeps main's merge order: `replayInvoke` folds the dispatch bag in after the originating request's own `internal`, and `publicNetworkOnly` and `resolvedSessionScope` come from the same handler values as before. `src/daemon/replay/**` no longer imports anything under `src/daemon/` outside the unit, and the vocabulary moves are pure moves. I found no bug that blocks this.

One design question first. `ReplayDispatchOptions` is a hand-kept copy of ten `DaemonRequestInternal` keys (https://github.com/callstack/agent-device/blob/cbc2bc2/src/daemon/replay/internal/command-types.ts#L170), and the `satisfies` check that the deleted `dispatchInternal` had is gone, because `withDaemonPrivateHalf` spreads the bag into `internal` and TypeScript does not check a spread for extra keys. If the daemon renames a key such as `replayTargetGuard`, replay would keep sending the old one, the build would stay green, and the guard would be silently ignored. Could the dispatch-option type live once in contracts, with `DaemonRequestInternal` extending it and replay importing it? That removes the copy and brings the compile-time check back. If you would rather keep the bag in the port, what would have to change first? It is worth settling before PR B, so the new package does not ship its own copy.

Not blocking: `session-state.ts` still keeps `SessionScope` as an alias with six daemon importers, and `DaemonInstallSource` is re-aliased on `contracts/command`; a few doc blocks in `test-command.ts` and `replay-script-selection.ts` now sit above the wrong function; and a fixture test could pin that a dispatch key overrides the base `internal` value while the other base keys survive.

Smoke Tests was still running when I looked. This diff touches the replay dispatch route those flows exercise, so a failure there needs a look before it is called unrelated. There are no conflicts. The next step is an answer to the dispatch-type question above.

thymikee:
Addressed in `1a695a82`.

**Dispatch options declared once.** `ReplayDispatchOptions` (and its `ReplayOpenLifecycle`) now live in `@agent-device/contracts/replay`. `DaemonRequestInternal` is `ReplayDispatchOptions & { publicNetworkOnly, executionPlan, retainDeviceExecutionLock, admittedLease, findResolvedTarget }`, and the port imports the same type, so a renamed key fails to compile on whichever side did not follow. The Maestro mapping got its `satisfies Record<keyof MaestroDaemonDispatchOptions, unknown>` guard back (`maestroDispatchOptions`), so a port-side option without a counterpart fails too. Nothing else had to change first: the only daemon-typed member was `openLifecycle.beforeDispatch(session: SessionState)`, and the one caller (`session-open-execution.ts`) passed a session no hook ever read, so the contract hook takes no argument. That also means `DaemonRequest` no longer reaches `SessionState` at all; the wire-shape test's positive control now uses `ExecutionPlan`, which the private half still carries. PR B ships no copy.

**Not-blocking items, all done.** `SessionScope` alias removed from `session-state.ts` (six importers take `contracts/session`); `DaemonInstallSource` alias removed from `contracts/command` (the two daemon importers take `kernel/contracts`, which already owned it). The stacked doc blocks in `test-command.ts` (three above one function, two of them pre-existing) and `replay-script-selection.ts` sit above the functions they describe. `handlers/__tests__/session-replay-command.test.ts` pins `replayInvoke`: a dispatch key overrides the base `internal` value, the other base keys survive, `dispatch` never reaches the daemon, and a request with neither side carries no `internal`.

**Validation at `1a695a82`.** `check:layering` OK, typecheck, lint, format, `check:daemon-wire-compat` (178 declarations, 0 changed), `check:production-exports` (68, unchanged), `check:fallow` (no issues in 23 changed files), eager-closure budgets (672, no growth). Suites `src/daemon`, `src/core`, `src/commands/interaction`, `packages/{contracts,selectors,ad-script}`: 3,976 passed; one 5 s timeout in `platform-apple/core/__tests__/apps.test.ts` (`closeIosApp on macOS`) under full-suite contention, untouched by this branch and green alone (25/25). Smoke Tests passed on the previous push; the new push re-runs them.

thymikee:
Reviewed at c551347. The dispatch options now live once in `@agent-device/contracts/replay`, which answers the design question from the last review. Nested replay dispatch still reaches the daemon with the originating request's private half plus the replay keys, as on `main`, and dropping the session argument from `beforeDispatch` changes nothing because no hook read it. No blocking findings.

The iOS Smoke Tests failure (`typed email suffix should be observable` in the live simulator fixture run) looks unrelated: this PR only changes import paths in the interaction files, and the same live simulator test also fails on `main` at f4930d6, at a different step.

thymikee:
The code verdict at c551347 is unchanged, but the branch now conflicts with `main`. I removed `ready-for-human` until it is rebased; #2686 is stacked on this branch, so it will need the same rebase.

thymikee:
Rebased onto `main` at `8d5bc8bae3`. Main grew the same seam from the other side while this was open, so three files conflicted and two auto-merges were wrong; here is each resolution.

**`session-replay-maestro-runtime.ts`.** Main threads `publicNetworkOnly` from the handler into `createMaestroReplayPort` instead of reading `command`. Kept main's explicit threading and this PR's `const { request: req, invoke } = command` unpacking plus the `dispatch` fold-in; `trustedScripts` still reads the threaded value, so the `node:vm` trust comment keeps its meaning.

**`session-replay-command.ts`.** Main passes `invoke: invokeReplayAction ?? invoke` plus a separate `publicNetworkOnly` field; this PR folds the private half into the invoke and draws `publicNetworkOnly` from the request split. Took this PR's side, which supplies the same field through `replayCommandEnvelope`/`splitReplayCommandRequest`, so main's line is not lost.

**The one real bug the auto-merge created:** `ReplayCommand` ended up with both this PR's `publicNetworkOnly?: true` and main's `publicNetworkOnly?: boolean`. Kept `?: true` — the contract is declared once in `@agent-device/contracts/replay` now, `splitReplayCommandRequest` emits `true` or nothing, and every reader tests truthiness.

That shape difference then needed three assertions translated, from `false` to absent, all in main's own new tests: `session-test-command-trust.test.ts` (`forwardedCommands()` for the trusted case) and two in `session-replay-command-trust.test.ts`. Absent is how the envelope says "local caller, trusts its own scripts", which is what those tests assert. `session-replay-command-trust.test.ts` also needed `vi.hoisted` for its façade doubles: this PR adds a daemon importer of the replay façade, so the mock factory now runs before a plain `const` initializes. Its two "maps `internal.publicNetworkOnly`" assertions, which are the security-relevant ones, are untouched.

Validation at `8d5bc8bae3`: `typecheck`, `lint`, `format`, `check:layering` OK; `vitest --project unit-core src/daemon/replay src/daemon/handlers` 668 passed. The earlier iOS Smoke failure (`typed email suffix should be observable`) is being re-run.

thymikee:
Reviewed at 8d5bc8b. The rebase keeps the code verdict from c551347. The conflict resolutions keep main's `publicNetworkOnly` threading and this PR's import cut, and the branch no longer conflicts with `main`.

The rebase commit adds 16 files of local test output under `.agent-device/test-artifacts/req-maestro-test-wire-{true,undefined}/` ([example](https://github.com/callstack/agent-device/blob/8d5bc8b/.agent-device/test-artifacts/req-maestro-test-wire-true/01-flow.yaml/attempt-1/result.txt#L1)), with `/tmp` paths and timestamps from one run. They come from main's `runWithTestNetworkFlag`, which calls `handleReplayTestCommand` without an `artifactsDir`, and `.gitignore` does not cover `.agent-device/`. Please remove them before merge. Giving that test an `artifactsDir` under its temp root, or ignoring the directory, can be a separate change.

Not blocking: [test-command.ts#L251](https://github.com/callstack/agent-device/blob/8d5bc8b/src/daemon/replay/internal/test-command.ts#L251) still sets `publicNetworkOnly: command.publicNetworkOnly` after the `...nestedReplayCommand(command, …)` spread already forwards it, an auto-merge leftover you can drop.

One real CI failure: the iOS Smoke job's `RunnerTests.testAbandonedTreeCaptureSkipsQuerySweepAndHonorsWarmupExemption` ([RunnerTests+SnapshotCapturePlanOccupancyTests.swift#L111](https://github.com/callstack/agent-device/blob/8d5bc8b/apple/runner/RunnerTests/RunnerTests+SnapshotCapturePlanOccupancyTests.swift#L111), occupancy count 0 vs 1). This diff touches no Swift or runner file, so it is likely unrelated. Repo Guards, Coverage and two Smoke runs were cancelled by newer runs. Repo Guards and Coverage exercise this diff, so they need a green rerun.

Next: remove the 16 test-artifact files and get green reruns of Repo Guards, Coverage and Smoke Tests.

PR #2675, opened 2026-09-19
refactor(replay): pass the Maestro script-trust flag as a command input (#2663 PR A)

thymikee:
## What

Next step of **PR A** for #2663 (narrows the replay unit's cut set with zero file moves). The replay unit stops reading `req.internal` for the Maestro script-trust decision; the daemon now hands that policy as an explicit command input, mirroring the #2652 `MaestroDaemonOperationInvoke` pattern.

No file moves, no behavior change.

## Change

- `ReplayCommand` gains `publicNetworkOnly?: boolean`. The daemon handler (`session-replay-command.ts`) computes it from the request's daemon-private half (`req.internal?.publicNetworkOnly === true`) at both entry points and passes it in.
- `session-replay-maestro-runtime.ts` reads the field **once** and derives both engine inputs from it — `trustedScripts: !publicNetworkOnly` for `executeMaestroFlow`, and the plain `publicNetworkOnly` option on the Maestro runtime port — replacing the two `req.internal?.publicNetworkOnly` reads.
- The flag threads into the per-test-case `runReplayCommand` in the replay-test runner, so Maestro flows under `replay test` keep the identical trust decision.
- One direct-entry test (`session-replay-maestro-remote-evalscript`) sets the field the way the handler does; its remote-vs-local `trustedScripts` assertions still exercise the engine input.

`internal` is still **written** into nested dispatch (a daemon mechanic, not a unit read), so the Maestro request envelope and the `maestroOperationDaemonRequest` fold are unchanged.

## Scope note (honest boundary)

This retires **2 of the 3** `req.internal` reads #2668 flagged in the unit. The remaining read — `resolvedSessionScope` in `session-replay-action-runtime.ts` — folds an `open` dispatch's implicit caller scope **and** its `?? resolveSessionScope(req)` fallback, and the `SessionScope` union it carries is shared with the session-lifecycle and snapshot families. Moving it means injecting `resolveSessionScope` and relocating `SessionScope`, which crosses module groups, so it lands with the dependencies bag in the **next PR A slice**, not here.

## Verification

Static + unit (all green):
- Characterization: `src/daemon/replay` (59 files / 376 tests) and `src/daemon/handlers/__tests__` (46 files / 280 tests) — identical results.
- `pnpm typecheck` clean.
- `pnpm check:layering`: OK — R6 still 3, R10 "2 external daemon request/session-state importers" unchanged (change is type + threading only, no new edges).
- `pnpm check:daemon-wire-compat`: protocol unchanged (178 declarations, 0 changed/added/removed).
- `pnpm check:replay-compat`: verified. `pnpm format`: clean.

Live device characterization (#2663 req 6), base (`94a0b6115a`, this commit's parent) vs head, same worktree/daemon/app — Android emulator `emulator-5554`, `com.callstack.agentdevicelab`:
- **`.ad` replay lane** (`pnpm test:replay:android`): **6 passed (6)** on head. The suite is non-deterministic on system-UI replays regardless of commit — I observed the pass count swing 2 → 5 → 6 across repeated runs of the *same* code; the flaky cases are Android Settings / Notifications / keyboard / system-swipe, which don't use the Maestro path this PR touches, and the native `.ad` path reads `publicNetworkOnly` nowhere. Base and head show the same flaky set; a clean head run is fully green.
- **Maestro flow** (`pnpm test-app:maestro:android`): base and head diverge at the **identical** step — step 3, `scrollUntilVisible "home-open-form"`, same screen (20 refs), same message. The `home-open-form` scroll target is simply not reachable in the current app/metro state on *either* commit, so this is a pre-existing app-content precondition, not a behavior delta. Launch + `assertVisible` (steps 1–2) pass and the engine reaches step 3 identically on both.

Conclusion: live `.ad` is green on head, and the live Maestro path is byte-identical base↔head — same observable replay behavior.

## Out of scope

`.ad` codec, replay semantics, heal, divergence classification; `session-replay-transaction.ts` and the repair lifecycle; the file moves (PR B → `packages/replay-port`).

Part of #2663


thymikee:
Reviewed at 1bb8b52. The production wiring is correct: both handler entry points and the `replay test` per-case run set `publicNetworkOnly` from `req.internal`, and a client cannot set it. But the trust test no longer covers the route that now does the mapping.

Before this change, the test fed `req.internal.publicNetworkOnly` into `runTypedMaestroReplay` and the runtime read it from there, so the test covered the whole path to `trustedScripts: false`. Now the test passes `publicNetworkOnly` to the runtime itself (https://github.com/callstack/agent-device/blob/1bb8b52/src/daemon/replay/internal/__tests__/session-replay-maestro-remote-evalscript.test.ts#L44). No test exercises the handler mapping in `session-replay-command.ts` (https://github.com/callstack/agent-device/blob/1bb8b52/src/daemon/handlers/session-replay-command.ts#L26 and https://github.com/callstack/agent-device/blob/1bb8b52/src/daemon/handlers/session-replay-command.ts#L67) or the forwarding in `test-command.ts` (https://github.com/callstack/agent-device/blob/1bb8b52/src/daemon/replay/internal/test-command.ts#L253). If a later change, such as the PR B move, drops one of those lines, the unset field reads as trusted and the suite stays green. A remote HTTP Maestro flow could then run evalScript with `trustedScripts: true`. Can the existing assertion go through `handleReplayCommand` and `handleReplayTestCommand` with `req.internal.publicNetworkOnly = true` and a spied `executeMaestroFlow`, so both `replay` and `replay test` are covered?

Not blocking: `ReplayCommand.publicNetworkOnly` is optional, and an absent field means trusted. A new construction site that forgets it type-checks and fails open. A required field, or an explicit input such as `scriptTrust: 'trusted' | 'untrusted'`, would make that state impossible and let the compiler find every site. It can go in the same follow-up or later.

This is a type and threading change with no device-facing behavior change, so I don't think this slice needs a live Maestro replay. CI was still running on all checks at review time, with no failures. There are no conflicts. The next step is the handler-level test for both routes.

thymikee:
**Correction + live evidence.** My initial description said the live `.ad`/Maestro lanes weren't run — that was wrong (a device was available). I ran them now (Verification section updated).

Base (`94a0b6115a`) vs head, same worktree/daemon/app on `emulator-5554`:
- `.ad` lane: **6/6 green** on head on a clean run. The suite is flaky on system-UI replays (pass count swung 2→5→6 across repeated runs of the *same* code) — those cases don't touch the Maestro path this PR changes, and the native `.ad` path reads `publicNetworkOnly` nowhere.
- Maestro flow: base and head diverge at the **identical** step (`scrollUntilVisible "home-open-form"`, same screen/message). Launch + `assertVisible` pass and the engine reaches step 3 identically on both → no behavior delta from this change. The `home-open-form` target isn't reachable in the current app/metro state on either commit (app-content precondition, not replay behavior).

Env note for whoever reruns these lanes in a fresh worktree: the Android snapshot lane needs `pnpm build:android` first (ships the `android/snapshot-helper` dist), or step-1 snapshots fail with "bundled helper artifact was not found".

thymikee:
Fixed in `7509689`.

The engine assertion now drives both real routes:

- `replay` through `handleReplayCommand` with `req.internal.publicNetworkOnly = true` and a spied `executeMaestroFlow`, asserting `trustedScripts: false` reaches the engine.
- `replay test` through `handleReplayTestCommand` with two Maestro flows, asserting both scheduler attempts receive `trustedScripts: false`.

Two narrower tests also pin the exact fail-open lines:

- `session-replay-command-trust.test.ts` captures the command produced by both handlers and asserts `internal.publicNetworkOnly` maps to `command.publicNetworkOnly`.
- `session-test-command-trust.test.ts` drives the suite command and captures the per-case `runReplayCommand` command to pin forwarding.

Local replay and handler suites, typecheck, lint/format, layering, and compatibility checks pass. CI is now green on `7509689`, with the PR clean and mergeable.

Agreed on the non-blocking follow-up: PR B should make this state impossible through a required field or an explicit trust enum.

thymikee:
Reviewed at 7509689. The new tests close the gap from the last review: `replay` and `replay test` now run through their handlers with `req.internal.publicNetworkOnly = true` down to the engine, so removing a mapping line fails a test. No blocking findings, and CI is green.

Not blocking: the mocked-argument check in [session-replay-command-trust.test.ts#L95](https://github.com/callstack/agent-device/blob/7509689/src/daemon/handlers/__tests__/session-replay-command-trust.test.ts#L95) repeats what the engine-routed tests already cover, so you could drop it.

PR #2676, opened 2026-09-19
fix(android): classify the clipboard shell response, not its exit status

### LINKED ISSUE #2674: Android: `clipboard read` returns empty string (success:true) on Android 10+ — read is denied, not surfaced
## Summary

`clipboard read` on a modern Android device returns an **empty string with `success: true`**, indistinguishable from a genuinely empty clipboard. Android 10+ (API 29) restricts clipboard reads to the foreground app or the active IME, and `agent-device`'s reader is neither — so the read is silently denied rather than surfaced as an error. This makes `clipboard read` unusable as written on physical devices, and worse, silently wrong: an agent testing "did the app copy this token to the clipboard?" gets `""` and concludes nothing was copied.

Issue #92 (which added the command) listed *"Supported on iOS simulator and Android emulator/device (document any limits)"* as acceptance criteria; this is that undocumented limit.

## Steps to reproduce (stock Settings app, no custom app needed)

```bash
agent-device open com.android.settings --serial <serial> --foreground
agent-device clipboard write "SECRET-MARKER-42" --serial <serial> --json
# → {"success": true, "data": {"message": "Clipboard updated"}}
agent-device clipboard read --serial <serial> --json
# → {"success": true, "data": {"platform":"android","action":"read","text": ""}}
```

Expected `text: "SECRET-MARKER-42"`; got `""`.

The write genuinely lands (the OS clipboard service holds the value — `adb shell dumpsys clipboard` confirms a primary clip exists); only the read comes back blank.

## Cause (hypothesis)

Since Android 10 (API 29), `ClipboardManager.getPrimaryClip()` returns `null`/empty unless the caller is the current default IME or the app with input focus (see the platform's clipboard access notes). Whatever process `agent-device` uses to read the clipboard does not satisfy that gate on a physical device, so the framework returns an empty clip and the command reports it verbatim.

## Impact

- `clipboard read` cannot verify copy/"Copy link"/OTP-to-clipboard flows on a real device — the primary use cases from #92.
- It fails **silently** (`success: true`, empty string), which is worse than an error: callers can't distinguish "denied" from "empty".

## Suggested fix

Route the read through a component that satisfies the API 29 gate — e.g. read via the bundled test IME (which is the active IME under `--test-ime` and is therefore permitted `getPrimaryClip`), or via a small foreground helper activity. Failing that, at minimum return a typed error/warning (e.g. `reason: clipboard_read_denied`) instead of an empty string, and document the limit on `help clipboard`.

## Environment

- agent-device 0.21.6 (npm latest)
- Device: Pixel 8a, **Android 16** (API 36), physical device over USB
- Host: macOS (Apple Silicon)

thymikee:
## Summary

`clipboard read` answered `text: ""`, `clipboard write` answered "Clipboard updated" (both `success: true`), and `capabilities` advertised `clipboard` on devices where adb cannot reach the clipboard at all. The cause in #2674 is not the Android 10 clipboard gate: Android 16's `ClipboardService` implements **no** shell command, so `Binder.handleShellCommand`'s framework default answers `No shell command implementation.` on stderr with **exit status 0** — and both legs plus the capability probe settled on the exit status first.

One typed verdict now owns that answer (`clipboard-shell-response.ts`), consumed by both legs and the probe: the service's own sentence is evidence at every exit status, because a service that never ran wrote no payload; `stdout` stays payload until the call has failed; and adb's generic `unknown command` wording never overrides a clean exit, since adb prints it about itself. Affected devices refuse with `UNSUPPORTED_OPERATION` plus a hint naming a device-side substitute. 13 files.

Closes #2674. Reading through a component that satisfies the Android 10 focus/IME gate — the other half of what that issue asked for — is #2685.

## Which route does a physical device take?

No physical device was reachable here, so this answers from the source that decides the route plus the reporter's own evidence, reproduced.

The route is decided by whether the clipboard **service class** implements a shell command; `cmd` finds the `clipboard` service and hands it the arguments, and the class is the same one a Pixel ships. Counting `ShellCommand` in `frameworks/base/services/core/java/com/android/server/clipboard/ClipboardService.java` per AOSP ref (line counts show the fetch landed):

| Ref | Lines | `ShellCommand` hits |
| --- | ---: | ---: |
| `android13-release` | 1182 | 0 |
| `android14-release` | 1540 | 0 |
| `android15-release` | 1617 | 0 |
| `android16-release` | 1690 | 0 |
| `main` | 1634 | 0 |

And the answer those builds produce is framework code, not a build variant — `core/java/android/os/Binder.java`, identical at `android16-release` and `main`:

```java
public int handleShellCommand(in, out, err, args) {
    FileOutputStream ferr = new FileOutputStream(err.getFileDescriptor());
    PrintWriter pw = new FastPrintWriter(ferr);
    pw.println("No shell command implementation.");
    pw.flush();
    return 0;
}
```

So the sentence arrives on the **err** descriptor with a clean exit on a physical Android 16 device too, which is the case this classifier keys on.

What the report cites as proof that the write landed does not evidence a working `cmd clipboard`, and both halves reproduce on this build:

```
$ npx -y agent-device@0.21.6 clipboard write "SECRET-MARKER-42" --serial emulator-5554 --json
{ "success": true, "data": { "action": "write", "textLength": 16, "message": "Clipboard updated" } }

$ adb -s emulator-5554 shell dumpsys clipboard | wc -c     # Android 16, ro.build.type=user
0
$ adb -s emulator-5554 shell dumpsys window | wc -c
58847
```

The published write is the false success this PR deletes, and `dumpsys clipboard` emits nothing at all to `shell` on an Android 16 user build (while other services dump normally), so it cannot have shown a primary clip on the reporter's device either — a clip there is an in-app copy, which an adb read never observes either way.

What would reopen this: a build that answers with something other than that sentence. Asked the reporter for `adb shell cmd clipboard set text X; echo rc=$?` and `adb shell cmd clipboard get text; echo rc=$?`. On such a build the blank read is Android 10's focus/IME gate, not a missing shell command, and that belongs to #2685 rather than to a prose rule invented for output this repo has never seen — which is why the classifier leaves a clean exit with no payload as `executed`.

## Shape

The façade carries one lazy function rather than two: `androidClipboardShellSupportForResult(result)`, because the probe holds an adb result and wants an admission verdict, and the verdict vocabulary stays package-side. That keeps one function-scoped `await import` in `mechanics.ts`, which is what holds its evaluated closure at the merge-base size.

`isAndroidShellCommandUnsupported` stays in `adb.ts`. It is call prose for the whole family, `settings airplane` reads it today, and moving it into the clipboard module would point a settings module at clipboard protocol while re-growing the closure that wrapper exists to hold down. Only clipboard protocol moved: the verdict, its admission mapping, and the refusal.

## Validation

`b73d0031b7`: `pnpm check:affected --run` passed every runnable check (442 files, 3221 tests, provider-integration and command-docs included), and the coverage lane's eager-closure probe passes with the façade at its merge-base closure size. GitHub lanes are 19/19 on this head: iOS, Android, macOS and Linux smoke, Coverage, Lint & Format, Typecheck & Package, Repo Guards, command-docs and provider integration. The iOS smoke lane needed three attempts here, and both reds are existing #2491 signatures — the cold-capture `wait_capture_stalled` stall (#2343) and the `automation-longpress` scroll — on a lane that `main`'s own `f4930d6346` also fails; the matrix is in the review thread.

Live, Android 16 (API 36) Emulator (Pixel 7 CI):

```
$ adb shell cmd clipboard get text; echo rc=$?     # the device, unchanged
No shell command implementation.                   # printed on stderr
rc=0

$ agent-device clipboard write "SECRET-MARKER-42" --serial emulator-5554 --json
- {"success":true,"data":{"action":"write","message":"Clipboard updated"}}
+ {"success":false,"error":{"code":"UNSUPPORTED_OPERATION","hint":"…focused field…"}}
```

`clipboard read` went `text: ""` → that refusal, and `capabilities` no longer lists `clipboard`. The positive leg holds: the provider lane's build answering `clipboard text: hello` still reads and writes.

Out of scope: the focus/IME-gated read route (#2685) and pre-`shell_v2` merged streams, where the sentence lands on `stdout` and is indistinguishable from payload — the residue `settings airplane` shares.


thymikee:
Reviewed b00e309. The emulator case is fixed, but I'm not sure the reported device is.

The fix keys only on the framework's `No shell command implementation.` sentence (https://github.com/callstack/agent-device/blob/b00e309/packages/platform-android/src/clipboard-shell-response.ts#L37). #2674 reports a physical Pixel 8a on Android 16 where `cmd clipboard set text` works and `dumpsys clipboard` shows the clip. If that is right, that build implements the shell command, and the read comes back blank because of the API 29 focus/IME gate: exit 0, empty stdout, no sentence. That classifies as executed, so the read still returns `text: ""` with `success: true`, which is the reported failure. The new known-limitations text already says an adb read "can still come back blank there", and the live run here used an emulator, not the reported route. Can you show which route a physical Android 16 device takes, with `adb shell cmd clipboard get text; echo rc=$?`, `cmd clipboard set text X; dumpsys clipboard`, and `agent-device clipboard read --json` on the same device (or from the reporter)? If stderr carries the sentence, this PR already covers it. If not, should any read with no `clipboard text:` payload fail instead of reporting success, whatever the exit code? Otherwise "Closes #2674" should become "Part of #2674", with the docs and CHANGELOG scoped to the builds actually observed.

Is there a smaller shape for the façade? `mechanics.ts` adds two async wrappers (https://github.com/callstack/agent-device/blob/b00e309/packages/platform-android/src/mechanics.ts#L98) for two sync functions that the tool host always calls as a pair. One lazy `androidClipboardShellSupportForResult(result)` called from `platform-runtime-android-tool-host.ts` would do the same job, and `isAndroidShellCommandUnsupported` could move into `clipboard-shell-response.ts`, since `settings airplane` is its only other caller. If you prefer the current split, what would have to change first?

Smoke Tests was still running at review time. It runs the changed clipboard probe through capabilities admission, so a failure there would need a look. There are no conflicts. The next step is the physical-device evidence above.

thymikee:
Reviewed at a42b496. The question from the last review is still open. The classifier at [clipboard-shell-response.ts#L42](https://github.com/callstack/agent-device/blob/a42b496/packages/platform-android/src/clipboard-shell-response.ts#L42) still treats exit 0 without `No shell command implementation.` on stderr as executed. On a build where `cmd clipboard` runs but the API 29 focus/IME gate blocks the read, `clipboard read` still returns `text: ""` with `success: true`, which is the symptom in #2674. The PR uses AOSP sources and an Android 16 emulator run to say the Pixel 8a takes the no-implementation route, but the reporter saw the clip in `dumpsys clipboard`, and the emulator transcript does not match that.

Could this say `Part of #2674` instead of `Closes #2674`, as #2685 does? Then the emulator evidence is enough. To keep `Closes #2674`, the PR needs output from a physical Android 13-16 device, ideally the reporter's Pixel 8a: `adb shell cmd clipboard get text; echo rc=$?` showing the sentence with rc=0, and `clipboard read --json` returning `UNSUPPORTED_OPERATION` with reason `android_clipboard_shell_command_unavailable`.

The branch now conflicts with `main` and needs a rebase. Smoke Tests were still running at review time.

thymikee:
Both asks addressed at `a42b4966a1` (rebased onto `main`).

**Which route a physical device takes.** No physical device was reachable here, so this is answered from the code that decides the route, plus the report's own evidence reproduced. The route turns on whether the clipboard *service class* implements a shell command — `cmd` finds the `clipboard` service and hands it the arguments — and that class is the same one a Pixel ships. `ShellCommand` occurrences in `services/core/java/com/android/server/clipboard/ClipboardService.java`:

| Ref | Lines | `ShellCommand` |
| --- | ---: | ---: |
| `android13-release` | 1182 | 0 |
| `android14-release` | 1540 | 0 |
| `android15-release` | 1617 | 0 |
| `android16-release` | 1690 | 0 |
| `main` | 1634 | 0 |

What such a service answers with is framework code, identical at `android16-release` and `main` (`core/java/android/os/Binder.java`): it writes `No shell command implementation.` to the **err** file descriptor and `return 0`. So the reported Pixel 8a takes the sentence route, which is the case the classifier keys on.

The report's two pieces of physical-device evidence also reproduce as the symptom rather than as a working `cmd clipboard`, on this build:

```
$ npx -y agent-device@0.21.6 clipboard write "SECRET-MARKER-42" --serial emulator-5554 --json
{ "success": true, "data": { "action": "write", "textLength": 16, "message": "Clipboard updated" } }

$ adb -s emulator-5554 shell dumpsys clipboard | wc -c     # Android 16, ro.build.type=user
0
$ adb -s emulator-5554 shell dumpsys window | wc -c
58847
```

The published write is the false success this PR deletes, and `dumpsys clipboard` emits nothing at all to `shell` on an Android 16 user build while other services dump normally — so that command cannot have shown a primary clip on the reporter's device. A clip there is an in-app copy, which an adb-side read cannot observe either way.

On failing any read that carries no `clipboard text:` payload: kept it as `executed` deliberately. On a build that *does* implement the command, a blank read is Android 10's focus/IME gate, and a rule for prose this repo has never captured would be guessing at a denial — and it cannot distinguish that from a genuinely empty clipboard, which is a real answer. That gate is the read route in #2685, filed from the issue's suggested fix. Asked the reporter for `cmd clipboard set text X; echo rc=$?` and `cmd clipboard get text; echo rc=$?` on their device; if it prints anything other than that sentence, the classifier should widen on the observed case and I would rather reopen #2674 than pre-empt it. Kept `Closes` with that stated in the body.

**Smaller façade.** Taken: one lazy `androidClipboardShellSupportForResult(result)` replaces the two wrappers. The probe holds an adb result and wants an admission verdict, so the verdict vocabulary stays package-side and the pair no longer has to be composed at the call site.

On moving `isAndroidShellCommandUnsupported`: left in `adb.ts`. It is prose about a call for the whole family, `settings airplane` reads it today, and the move would have a settings module depend on clipboard protocol while putting the eager edge back into the façade closure that the wrapper exists to hold down. Only clipboard protocol moved — the verdict, its admission mapping, the refusal. If the shared constant still bugs you, the fix is a lazily-reached prose module shared by both families, which I'd take in a follow-up rather than smuggle in here.

Smoke Tests did need a look, but not here. This branch's `b00e3097e5` run died inside `smoke:automation-input` at `wait text Agent Device Tester` with `wait_capture_stalled` and zero readable captures ([run 35428534467](https://github.com/callstack/agent-device/actions/runs/35428534467)) — an iOS-simulator capture stall, in a lane `main`'s own `f4930d6346` fails the same way at `wait text Automation lab` ([run 35443296005](https://github.com/callstack/agent-device/actions/runs/35443296005)). My first head `d3a98b7afd`, carrying the same Android change, passed it, and nothing in this diff reaches Apple capture. It is green on the rebased head ([run 35445370019](https://github.com/callstack/agent-device/actions/runs/35445370019)): iOS, Android, macOS and Linux smoke lanes, plus Coverage, Lint, Typecheck & Package and Repo Guards, all pass at `a42b4966a1`.

thymikee:
Reviewed at b73d003, a rebase of a42b496 where only the CHANGELOG position changed. Thanks for the answer on the device route. The AOSP count shows that `ClipboardService` has no shell command on Android 13-16, and your run shows that `dumpsys clipboard` prints nothing on an Android 16 user build, so the report's clip evidence does not point to a different route on the Pixel 8a. Keeping a blank read as `executed` is also right: with no captured denial text it cannot be told apart from an empty clipboard, and #2685 owns the gated read. That clears the open question. No physical device has run this yet, so a Pixel check is still a useful follow-up, but it does not block review.

The earlier Smoke Tests failure on this head was in the iOS simulator fixture (`automation-longpress` did not become visible after scrolling), a route this Android-only change does not touch, so it looks unrelated. A rerun is in progress.

thymikee:
Smoke lane update, since it went red twice here and #2491 asks that each current signature be confirmed against main rather than waved off.

| run | tree | iOS smoke |
| --- | --- | --- |
| [35428534467](https://github.com/callstack/agent-device/actions/runs/35428534467) | `b00e3097e5` | fail — `wait for Agent Device Tester`, `wait_capture_stalled`, `readableCaptures: 0` |
| [35445370019](https://github.com/callstack/agent-device/actions/runs/35445370019) | `a42b4966a1` (same tree, base `f4930d63`) | pass |
| [35446623933](https://github.com/callstack/agent-device/actions/runs/35446623933) a1 | `b73d0031b7` | fail — `id="automation-longpress" did not become visible after scrolling` |
| 35446623933 a2 | same commit | fail — same assertion, same call sites |
| 35446623933 a3 | same commit | **pass** |
| [35445832157](https://github.com/callstack/agent-device/actions/runs/35445832157) a2 | main's `0cd679fc` — this PR's base, without this commit | pass |

Both signatures are rows in #2491: the cold-capture stall (#2343) and the `automation-longpress` scroll. What argues against "the branch changed iOS behavior" is the same commit passing the lane on a later attempt, having failed two before it — a deterministic regression does not move its symptom or its verdict per attempt. And the reach is narrow: of the 13 files, two are loaded on an iOS run at all — `src/commands/system/index.ts`, where the change is one sentence of `help clipboard` text, and `src/platform-runtime-android-tool-host.ts`, whose change sits inside a probe body that runs only when something calls this Android tool host. The other 11 are `packages/platform-android` and its tests, this command family's own root test, two docs pages and the CHANGELOG.

Rollup at `b73d0031b7` is 19/19 with `MERGEABLE` / `CLEAN`. I left both red attempts visible rather than re-running until the lane went quiet. The lane is still the #2491 problem it was: three of six iOS runs on this branch were red, across trees that differ only by base, and `main`'s `f4930d63` is red on it too.

PR #2684, opened 2026-09-19
refactor(ios): drop the flat sweep's discarded hittable computation

thymikee:
Follow-up to #2669, from the reviewer's non-blocking note there: `flatSnapshotNode` (the query-sweep tier) still computed `hittable` from the reported frame, but since #2669 the one normalization pass in `captureWithBackend` calls `replacing(rect:hittable:)` on every acquired node, so that value was overwritten before anything read it.

## What changed
- `flatSnapshotNode` passes `hittable: false`, the same placeholder the recursive walker and the private-AX path already pass, and its `viewport` parameter (only used by the dropped computation) is gone. The sweep's own `viewport` stays: it still sizes the synthetic root and the result.
- 3 insertions, 14 deletions, one file. No presented output changes: `hittable` on every node is decided in `SnapshotGeometrySpace.normalized`, which is unchanged.

## Validation (`2463a74`)
- iOS runner built with unit tests compiled in (`AGENT_DEVICE_XCUITEST_INCLUDE_UNIT_TESTS=1`, arm64, iOS 26.2); `testAbandonedTreeCaptureSkipsQuerySweepAndHonorsWarmupExemption`, `testRegularDepthCutsPresentationNotAcquisition`, `testVisibilityFoldKeepsIndependentChildPastClippedParent` pass on a throwaway iPhone 17 Pro simulator (3 executed, 0 failures).
- `pnpm check:affected --run --head 2463a74`: `xctest-selection` and `packaged-runner-swift` pass; the Swift build lanes are GitHub-authoritative. `oxlint` is clean on the repo (a local, git-ignored `.claude/` file fails the working-tree lint on my machine; it is not in the diff).
- No changelog entry: the #2661 entry already states that the normalization pass recomputes `hittable`.


thymikee:
Reviewed at 2463a74. No blocking findings: `normalized()` recomputes `hittable` for every node, so the value the flat sweep computed was never used, and removing it changes no snapshot field. CI is green.

Not blocking: the three tests named in the PR body do not call `flatSnapshotNode` or `querySweepSnapshotAcquisition`, so the validation note could say that correctness rests on the recompute in `normalized()`.

PR #2686, opened 2026-09-19
refactor(replay): bind the session and dispatch envelope inside the unit (#2663 PR A′)

thymikee:
## What this is

The seam that #2663 part B needs, split out of the move. Stacked on #2672; it does not move the unit.

Part B is "pure move `src/daemon/replay/**` into `packages/replay-port`". Measured against this branch, 29 of the unit's 68 test and fixture files still reach daemon modules (`SessionStore`, the handler's session factory and command envelope, the capture seam, `ALLOCATOR_CLAIM_MISSING`, `test-utils`), and `scripts/layering/package-boundaries.ts` (R11) refuses exactly that from inside a package. Moving first would have meant either exporting port internals for tests or leaving 29 files behind.

So this commit moves the last two things the *production* unit was missing to stop naming daemon mechanics, and leaves the test topology for the next slice.

## Shape

`src/daemon/replay/internal/replay-session-binding.ts`
- `ReplaySessionContainer`: the live record reached through four reads (`get`, `lookup`, `getRuntimeHints`, `ensureSessionDir`).
- `ReplaySessionPolicy`: what the port consults but does not own — the repair gateway, selector admission, open runtime hints, the ref-publication authority binder, and snapshot capture.
- `bindReplaySession(name, logPath, container, policy)`: the assembly that `createReplaySession` used to hand-build, including the `store` / `observationStore` split.

`src/daemon/replay/internal/replay-dispatch-envelope.ts`
- `splitReplayCommandRequest`: the wire half plus the two admission facts (`publicNetworkOnly`, `resolvedSessionScope`), typed over `DaemonWireRequest` and a caller-named private half.
- `replayInvokeOverDispatch`: nested dispatch that folds the private half back in, merge order unchanged. The private merge stays module-private, as `withDaemonPrivateHalf` was.

`createReplaySession` in the handler is now an adapter: it names the daemon's store and policy and delegates. The coordinator keeps its own full-`SessionState` store view rather than the narrowed container, because the repair transaction writes daemon fields the port never sees.

Net: the handler loses ~80 lines and the unit gains no new daemon import. Facade and `DAEMON_REPLAY_FACADE.exports` name the two new values.

## Gates

- `pnpm typecheck`, `pnpm lint`, `pnpm format`, `pnpm check:layering` — clean (R13 needs the new modules committed; that is why they are in this commit and no earlier).
- `pnpm exec vitest run --project unit-core src/daemon/replay src/daemon/handlers` — 107 files, 660 tests.
- Four test files change by rename only: `replayCommandEnvelope` → `splitReplayCommandRequest`, `replayInvoke` → `replayInvokeOverDispatch`, now imported through the facade. No assertion moved or was dropped.
- `check:production-exports`: neither new module is reported; the 68 findings it prints are pre-existing in packages this branch does not touch.

## Deviations and what is next

#2663 part B asked for the move in one slice. Measured, it is not one slice, and I did not pretend otherwise:

- 8 of the 29 daemon-coupled test files also depend on daemon policy modules, and 15 files assert repair-transaction or ref-frame state that only the daemon's `SessionState` carries. A port-local harness can double the coordinator's *interactions*, not the daemon's transaction outcomes, so those files owe a split (unit test over a double + daemon-side test of the real transitions) rather than a relocation.
- The unit's eager-closure entries evaluate 47–285 repo modules each against the `domain-facade` ceiling of 20, so part B also owes an `APPROVED_OVER_CEILING` row (precedents: #2544, #2317, #2198).
- Fallow baselines are path-keyed, so the move carries `fallow-baselines/health.json` entries with the files.

Follow-ups, in order: a port-local session harness plus the ~21 files whose daemon coupling is only container, capture and command construction; the 8 daemon-coupled files and `ALLOCATOR_CLAIM_MISSING`; then the move itself with ownership and closure entries.

## Adversarial review outcome

Reviewed the move for behavior drift after the first commit. Two findings were fixed in `cd6304cb71`:

- The handler wrote the four session reads twice (once for the port's container, once for the coordinator's store). It now names them once, so scoping a read on one view without the other is no longer reachable.
- The private-half merge carried an `as unknown as` cast that muted the convertibility check and bought nothing; the plain `as` compiles.

Both open findings from that review were closed by the rebase, on #2672's side rather than here:

- **The private-half key vocabulary is checked again at its declaration.** `DaemonRequestInternal` is now `ReplayDispatchOptions & { ... }` in `src/daemon/daemon-request.ts`, and `ReplayDispatchOptions` is declared once in `packages/contracts/src/replay.ts`, so a key renamed on one side fails to compile on the other without any cast in this unit. `replay-dispatch-envelope.ts` reads the type from that contracts declaration.
- **The merge precedence and the empty-private-half branch are covered.** `session-replay-command.test.ts` grew both tests while this branch was open; `950ce4c412` carries them into the unit beside `replay-dispatch-envelope.ts`, assertions unchanged.

Rebased on `refactor/2663-replay-cut-set` at `c551347036`. The rebase was clean at patch level but the new upstream tests referenced the two helper names this commit removes, so they moved with the behaviour instead of being renamed in place.

## Rebased on `main` (second pass)

#2672 took a rebase onto `main` underneath this branch, which brought main's `publicNetworkOnly` threading and two new trust tests. Both branches are rebased and green; the follow-up commits here are the integration:

- `74cefa8` — the Compatibility & Provenance gate failed: `ReplayDispatchRequest` and `ReplayInvoke` were unused type re-exports on the replay façade. Binding the invoke moved those types inside the unit, which imports them from `internal/command-types.ts`, so the façade lost its only consumers when the handler stopped naming them. Removed with their two pin-table entries; the façade is a hand-curated surface, not one the exhaustiveness rule covers. `pnpm gate fallow --base <#2672 head>` now reports no issues in the 11 changed files.
- `c323646` — main's `session-replay-command-trust.test.ts` doubles the replay façade with a factory naming only the two entry points. This branch adds the session binding and request split to that façade, and the handler calls them on the way to the command input the test observes, so the mock now spreads the real module and doubles just those two. Its four assertions are unchanged.

Gates at `c323646`: `typecheck`, `lint`, `format`, `check:layering`, `gate fallow` clean; `vitest --project unit-core src/daemon/replay src/daemon/handlers` 668 passed.

## Why the rebase's CI went red, and what the last two commits are

Three gates failed on `c323646` — Bundle Size, Compatibility & Provenance, Coverage — all with the same cause, none of it in the diff: each compares this branch against the base commit recorded in the `pull_request` payload, and #2672's rebase force-pushed `refactor/2663-replay-cut-set` off `c551347`, so that object was unreachable. `fatal: unable to read tree (c5513470…)`, and `could not determine changed files for base ref 'c5513470…'`. A re-run replays the same payload, so only a fresh `synchronize` event carries a readable base. That is what `fe4159f`/`b3cbae6` are for, and all three gates pass on them.

Those two commits are the review points that were still open on this branch, not filler:

- `fe4159f` — `ReplayCommandEnvelope` was a hand-copy of three `ReplayCommand` members, including a second copy of the public-network doc sentence. It is now `Pick`ed from the command, so the envelope cannot drift from what the command reads.
- `b3cbae6` — the binding passes the container's four reads straight through, so a read written as a method needing a receiver would type-check and throw on first call. Stated on `ReplaySessionContainer`, and the container spreads into the store view instead of each read being named twice.

Local at `b3cbae6`: `typecheck`, `lint`, `format`, `check:layering`, `gate fallow --base <#2672 head>` clean; `vitest --project unit-core src/daemon/replay src/daemon/handlers` 668 passed.


thymikee:
Reviewed at 950ce4c. The session binding and the dispatch envelope match the handler code they replace, so replay behavior does not change.

The Compatibility & Provenance failure comes from this diff. The handler was the last consumer outside the unit of the `ReplayDispatchRequest` and `ReplayInvoke` re-exports. This change removes that import but keeps both names in [src/daemon/replay/index.ts](https://github.com/callstack/agent-device/blob/950ce4c/src/daemon/replay/index.ts#L7-L8) and in `DAEMON_REPLAY_FACADE.exports` in [architecture-ownership.ts](https://github.com/callstack/agent-device/blob/950ce4c/scripts/layering/architecture-ownership.ts#L19-L20). Removing them from both places should make fallow pass.

Does `bindReplaySession` earn its seam? The daemon handler still writes all nine members, and the only logic the binding adds is the shared `get` alias, so the `ReplaySessionContainer`/`ReplaySessionPolicy` split does not seem to move any responsibility into the port. Could part B move the unit with `createReplaySession` left in the handler (it already names only `ReplaySession`) and keep only the dispatch envelope? If a port-local session harness in part B will bind a session itself, that justifies the seam, but could it then land with that harness?

The branch also conflicts with its base, `refactor/2663-replay-cut-set`, so it needs a rebase. Next step: remove the two re-exports and answer the seam question.

thymikee:
Reviewed at c323646. The earlier finding is fixed. `src/daemon/replay/index.ts` and `DAEMON_REPLAY_FACADE.exports` now list the same 16 names, and `ReplayDispatchRequest` and `ReplayInvoke` are used only inside `src/daemon/replay/internal/`. The moved session binding and dispatch envelope still match the handler code they replace.

The PR body answers the earlier `bindReplaySession` question. The unit's tests still reach the handler's session factory and command envelope, and R11 refuses that from inside a package, so part B cannot move the unit with `createReplaySession` left in the handler.

The branch no longer conflicts with its base. Bundle Size, Compatibility & Provenance and Coverage failed because they ran with `--base c551347`, the old #2672 head that no longer exists after the force-push, so they never compared this PR's content. They need a rerun against `8d5bc8b` to confirm that fallow, coverage and size pass.

Issue #2682, opened 2026-09-19
Silent target re-activation lets a screenshot and the next snapshot describe different apps


thymikee:
## Why

The iOS runner already repairs foreground loss. Before a non-lifecycle command it re-activates the session app — `activateTarget(bundleId:reason:)` (`RunnerTests+Lifecycle.swift:203`) called with `reason: "stale_target"` (`RunnerTests+CommandExecution.swift:1527`) and again with `"interaction_foreground_guard"` (`:1551`). Lifecycle commands are exempt (`RunnerTests+Models.swift:56`), and `screenshot` is one of them.

So the sequence an agent can actually observe is:

1. Something hands off to another app (a `mailto:`/`prefs:` link, Settings, a user gesture).
2. `screenshot` shows that app, because it skips activation.
3. The next snapshot or tap silently brings the session app back to the foreground and answers as if nothing happened.

The only record of that repair is an `NSLog`. The agent's own evidence contradicts itself — a screenshot of one app and a tree of another — and nothing in the response says the runner moved the screen. An external report describes the symptom exactly (pixels showing Settings while semantics stayed with the original app) and instructs its agents around it. The report is secondhand; the activation call is not, and it is the most likely cause.

Known system hosts are the honest case: `presentedSystemSurfaceHost()` serves them in place instead of activating (`RunnerTests+CommandExecution.swift:1501`, `:1577`, #2438). The gap is that an *unknown* foreground app gets repaired in silence.

## Task

Decide the policy, then make the repair visible.

1. **Policy decision first, in this issue.** Either (a) the runner keeps reactivating and discloses it, or (b) it refuses when the bound app is not foreground and the caller decides. Today it is (a) with no disclosure, which is the one option that is not acceptable: ADR 0012 retired silent state rewriting, and an undisclosed `XCUIApplication.activate()` is the same category of decision made on the agent's behalf.
2. **Emit the fact where it is decided.** Attach the activation to the response of the command that triggered it, e.g. `targetActivation: { reason, priorState, foregroundPid? }`. Decode it once in `platform-apple` and let the daemon append a warning through the shared response builder — response-level metadata, appended, never replacing earlier warnings (AGENTS.md).
[...]

PR #2688, opened 2026-09-19
feat(ios-runner): key runner startup failures on typed reasons

### LINKED ISSUE #2680: Key runner build failures on typed reasons instead of signing-substring hints
## Why

A `build-for-testing` failure is the one Apple failure shape with no typed reason. `runner-artifact.ts:521` throws `COMMAND_FAILED` `'xcodebuild build-for-testing failed'` with `error`, `details`, `logPath`, and a prose `hint` from `resolveSigningFailureHint` (`runner/runner-contract.ts:601-620`), which lowercases `error.message` + `JSON.stringify(error.details)` and substring-matches: `failed registering bundle identifier`, (`app identifier` + `not available`), `requires a development team`, `no profiles for` / `provisioning profile`, `code signing`. Nothing downstream can switch on it, which contradicts AGENTS.md: "Key behavior on typed reasons and details, never error text."

This is a completion, not a new subsystem. Two typed classifiers already exist and the build path is wired to neither:

- `RunnerErrorRule` — "the one declaration of runner error classes (#1631)" (`runner/runner-contract.ts:205+`), declarative match + verdicts.
- `BootFailureReason` (`packages/provision-kit/src/boot-diagnostics.ts:5`, mirrored at `runner/host.ts:146`), already covering `IOS_RUNNER_DEVICE_NOT_PROVISIONED` from CoreDevice `0xe8008012` and `BOOT_COMMAND_FAILED` from provisioning/embedded-profile text, with hints in `bootFailureHint` (`:125`).

So early-exit and install failures get reasons; the build itself does not.

## Task

1. Extend the existing tables — do not add a third classifier. Note `BootFailureReason` is declared **twice**, in `packages/provision-kit/src/boot-diagnostics.ts:5` and `packages/platform-apple/src/runner/host.ts:146`, both extending `InfrastructureBootFailureReason` from contracts; growing it means both declarations move together. The alternative is `RunnerErrorRule` rows only. Pick one and say why. Either way, the `runner-artifact.ts:517-527` catch — which today calls the exported `resolveRunnerBuildFailureHint` (`runner-contract.ts:622`), a thin wrapper over `resolveSigningFailureHint` — must end up calling exactly one classifier that returns `{ reason, hint }`.
2. Reasons to cover: `signing_no_development_team`, `signing_provisioning_profile_missing`, `signing_style_conflict` (automatic vs manual), `bundle_identifier_already_registered`, plus the host `DevToolsSecurity` refusal, which today throws `COMMAND_FAILED` with no reason at all (`runner/runner-session.ts:724-736`). Name the catch-all `build_failed_unclassified` and keep `RUNNER_CACHE_RECOVERY_HINT` as its hint — do not claim a cause you did not prove.
3. `runner-session.ts` is 1,181 lines. Per AGENTS.md, adding behavior there owes a split first; the Developer Mode/DDI **probe** is #2683, not this issue. Only the missing reason at the existing throw site belongs here.
4. Keep the neighboring sniffs honest: `matchesRunnerErrorMessage` (`runner-contract.ts:321-322`) and `resolveRunnerEarlyExitHint` (`:537-538`) are also text matchers. Either fold them into the table or state at the table why they stay.
5. Fixture provenance: each fixture records the Xcode version and the command that produced it. These are TypeScript-only fixtures next to the classifier's test — `contracts/fixtures/` golden tables are for cross-language rules and are the wrong home.

## Acceptance criteria

- [ ] Exactly one classifier is reachable from the build-failure catch, and `resolveRunnerBuildFailureHint` is gone rather than renamed or kept as a pass-through wrapper.
- [ ] Each fixture is driven through the build-failure catch, not called directly against the classifier, and asserts `details.reason` on the thrown error.
- [ ] One fixture per reason in the union, checked by count so a new reason without a fixture fails.
- [ ] Assertions are on the normalized envelope as callers see it: top-level `code`, `hint`, `logPath`, `diagnosticId`, and `details.reason`. This is a regression guard, not a feature claim — `normalizeError` **deletes** `hint`, `logPath`, and `diagnosticId` from `details` (`packages/kernel/src/errors.ts:417-429`), so an AC written against `details.hint` would assert a field that cannot exist.
- [ ] Error-code assertions alone prove nothing — every case here is `COMMAND_FAILED`. The reason is the assertion.
- [ ] Device evidence with commit SHA for at least: unset `AGENT_DEVICE_IOS_TEAM_ID`, and a taken bundle identifier.
- [ ] Versioned CLI help updated where it describes runner prepare failures; `pnpm check:affected --run`, `pnpm check:layering`, `pnpm check:production-exports` green.

## Non-goals

Signing, keychain unlocking, automatic profile fetching, and any new fallback. Unknown failures keep today's hint. Device-side Developer Mode and developer disk image classification, and the stale-log-tail defect, are #2683.

Related: ADR 0010 (error system), ADR 0005, #2662 (same one-decoder instinct one layer down). Blocks: #2683, which adds its two device-readiness reasons to the vocabulary created here.

thymikee:
## Summary

`build-for-testing` was the one Apple failure shape that reached callers as prose: the catch in `runner-artifact.ts` threw `COMMAND_FAILED` with a hint chosen by substring-matching the message plus a JSON dump of details, so nothing downstream could switch on the cause.

`classifyRunnerStartupFailure` is now the only classifier reachable from that catch, and it extends `RUNNER_ERROR_RULES` instead of adding a third table. `BootFailureReason` was rejected: it is `provision-kit`'s boot vocabulary, shared with Android boot, and a build that never produced a binary has no boot to classify. One table is also the only place a row and its reason cannot disagree.

Rows carry `buildFailure: { reason, hint }`, so reason and advice cannot disagree — `signing_no_development_team`, `signing_provisioning_profile_missing`, `bundle_identifier_already_registered`, `signing_unspecified`, `devtools_security_developer_mode_disabled`, and `build_failed_unclassified`, which keeps `RUNNER_CACHE_RECOVERY_HINT` instead of claiming a cause. Matching reads only what xcodebuild published — message, stdout, stderr — never the argv we were handed, and never this PR's own emitted reason. `signing_style_conflict` is withdrawn: conflicting provisioning settings stay unclassified until a capture names the disagreeing setting. `resolveRunnerBuildFailureHint` is deleted.

A named profile has to earn its reason, and in the same line (#2688 review). The shipped sniffer's bare `provisioning profile` trigger is a phrase a failing build prints while talking about something else — the codesign invocation, a settings dump, a note about the profile it used — and a second phrase anywhere in the log is no better, because a profile note above an unrelated expired-certificate warning is two lines that never met. So the table gains `toolTextLineIncludesAll`, the profile rows require Xcode's own complaint on the profile's line (`IDEProvisioningErrorDomain` naming the profile, `doesn't include`, `has expired`), and a mention that says nothing keeps the cache-recovery advice. `signing_no_profiles_for_bundle_id` shares the `PROFILE_UNUSABLE` advice constant rather than spelling the same hint out a third time.

```
$ agent-device prepare ios-runner --platform ios --json
{"hint":"Configure signing in Xcode or set AGENT_DEVICE_IOS_TEAM_ID …",
 "details":{"reason":"signing_no_development_team"}}
```

Closes #2680. 86 files, 443 production lines.

## Validation

At `bcfed35b4c` (rebased onto `main` `fc59620932`): `build`, `format`, `check:quick`, `check:layering`, `check:gate-manifest`, `check:command-docs`, `check:production-exports` and `apple-runner` (504) pass, as does the `unit-core` cli-schema, help-conformance and eager-closure ratchet lane and `fallow audit --base origin/main` over the stack. Restoring whole-details matching fails 3 tests — the argv fixture, its focused negative, and self-match; reverting the expired row to a whole-log `has expired` match fails `profile-note-above-an-expired-certificate`. `check:affected --run` stays with CI.

Risk: live evidence pending — commands in `.device-evidence/CHECKLIST.md`. Fixtures declare `captured`, `shipped-sniff-trigger` or `invented-shape` with `xcodeVersion: unobserved`, so nothing transcribed reads as observed; the three narrowed profile rows are `invented-shape` until Phase B records which line xcodebuild prints, and the checklist now asks for the line breaks because which line carried which phrase is the evidence.


thymikee:
Reviewed at 6342a18.

The typed reasons change the error envelope on the physical-device `build-for-testing` route, but every new fixture is transcribed, not captured, and the PR body says live evidence is still pending. #2680 lists device evidence with a commit SHA as a done criterion. Until then we do not know that real Xcode output matches these rows and does not fall through to `build_failed_unclassified`. Could you run `agent-device prepare ios-runner --platform ios --device <iPhone> --json` on the pushed head twice: once with `AGENT_DEVICE_IOS_TEAM_ID` unset (expect `details.reason: signing_no_development_team`), and once with a bundle id that another team owns (expect `bundle_identifier_already_registered`)? Please paste both JSON envelopes with the hint, `logPath` and `xcodebuild -version`, and mark the matching fixtures as captured.

Could `.device-evidence/CHECKLIST.md` go? It adds a top-level directory that is not on `main`, and a doc comment in `runner-startup-failure-fixtures.ts` points to it; the commands fit in the PR body.

I did not find a smaller design: the new rows reuse `RUNNER_ERROR_RULES`, and the change is about 198 net production lines. Optional cleanups: put the duplicated bundle-id and profile hint strings in constants, and let [`buildDevToolsSecurityRefusal`](https://github.com/callstack/agent-device/blob/6342a18/packages/platform-apple/src/runner/runner-dev-tools-security.ts#L35) throw once with the reason the probe already knows, not build an `AppError` only to classify and rebuild it.

CI is green. GitHub reports a conflict with `main`, so the branch needs a rebase. Next step: the two live device runs on the rebased head.

thymikee:
Reviewed at 98f2870. This follows up on the earlier review at 6342a18 (https://github.com/callstack/agent-device/pull/2688#issuecomment-5742670852).

The two live physical-device runs asked for at 6342a18 are still missing, and #2680 lists them as a done criterion. Every build-for-testing fixture in [runner-startup-failure-fixtures.ts](https://github.com/callstack/agent-device/blob/98f2870/packages/platform-apple/src/runner/__tests__/runner-startup-failure-fixtures.ts#L133) still has provenance `shipped-sniff-trigger` or `invented-shape`, with `xcodeVersion: 'unobserved'`. This delta also cut the fixture sentences down to the bare trigger substrings. So nothing shows that real Xcode output reaches these rows instead of falling through to `build_failed_unclassified`. If current Xcode words a signing failure differently, the physical-device prepare route would publish `build_failed_unclassified` and a generic hint for it.

Can you run this on a physical iPhone at the pushed head: `node --experimental-strip-types src/bin.ts --json prepare ios-runner --platform ios --device <iPhone>`, once with `AGENT_DEVICE_IOS_TEAM_ID` and `AGENT_DEVICE_IOS_PROVISIONING_PROFILE` unset, and once with `AGENT_DEVICE_IOS_BUNDLE_ID` set to a bundle id that another team owns? The first should fail with `COMMAND_FAILED` and `details.reason: 'signing_no_development_team'`, plus a hint and `logPath`. The second should fail with `details.reason: 'bundle_identifier_already_registered'`. Please paste both JSON errors with the commit SHA and `xcodebuild -version`, then update the two matching fixtures with the captured text and mark them `captured`.

Not blocking: the PR description says 86 files and 443 production lines, but this diff touches 11 files with +264/-43 production lines, and the earlier question about `.device-evidence/CHECKLIST.md` is still open because `main` has no `.device-evidence` directory.

CI is green on all 5 checks. The branch conflicts with `main` and needs a rebase.

Next: post the two physical-device `prepare ios-runner --json` captures with reason, hint, `logPath`, SHA and `xcodebuild -version`, mark those fixtures captured, and rebase onto `main`.

thymikee:
At 85bd4de this is still the same rebase plus a test-helper extraction; it adds no new captures. Every build-for-testing fixture in packages/platform-apple/src/runner/__tests__/runner-startup-failure-fixtures.ts#L56 is still shipped-sniff-trigger or invented-shape with xcodeVersion 'unobserved', so nothing shows real Xcode output actually reaches these rows. Two rows are unproven in opposite directions: `signing_provisioning_profile_unusable` matches the bare substring "provisioning profile" anywhere in message/stdout/stderr, and no fixture drives it, so a build log that merely mentions a profile can now publish `signing_provisioning_profile_missing`, which installation.md tells consumers to branch on; meanwhile real signing prose that differs from the trimmed trigger text falls through to `build_failed_unclassified` with the generic recovery advice. On the physical-device prepare route this means a caller can get a typed reason that names the wrong cause, or no cause, while the docs invite branching on it.

Can you run, on this pushed head, `node --experimental-strip-types src/bin.ts --json prepare ios-runner --platform ios --device <iPhone>` twice: once with AGENT_DEVICE_IOS_TEAM_ID and AGENT_DEVICE_IOS_PROVISIONING_PROFILE unset (expecting COMMAND_FAILED with details.reason `signing_no_development_team`), and once with AGENT_DEVICE_IOS_BUNDLE_ID set to an identifier another team owns (expecting `bundle_identifier_already_registered`)? Please paste both full JSON envelopes (reason, hint, logPath), the commit SHA, and `xcodebuild -version`, mark the two matching fixtures as captured with the real sentences, and say which of the two profile rows the captured text actually hits so the bare "provisioning profile" row is either narrowed or backed by a fixture.

Not blocking: the rebase onto 0ebd254 (#2689) left the PR body's "declared twice" rationale stale now that BootFailureReason has one declaration in provision-kit/src/boot-diagnostics.ts#L5 with runner/host.ts only delegating, so the PR body could drop that clause and keep the vocabulary argument, but this can be taken or left.

I did not run any test locally, so the claim that the moved assertions still fail without the fix rests on reading the hunks, not a run. I could not verify real Xcode wording for any rule since no capture exists in the branch. Repo Guards job 105956422148 shows CANCELLED with no failed-step log, which looks like a run superseded by the rebase push, but the diff does touch what that job exercises (a new top-level .device-evidence/ directory, a new platform-apple module, and docs), so the cancellation cannot be read as unrelated; it simply carries no result and needs a re-run on 85bd4de.

Before this can merge: post the two physical-device prepare captures with reason, hint, logPath, commit SHA, and `xcodebuild -version`, mark those fixtures captured, and get Repo Guards to run to completion on 85bd4de.

thymikee:
Both points addressed, rebased onto `main` `d688926167`. New head: `9621905975`.

**The bare `provisioning profile` row is gone.** You were right that it is the shipped sniffer's trigger, and that a name is not a complaint: `xcodebuild` prints that phrase in a `note:` about the profile it used, in a settings dump, and in the codesign invocation. Three rows now require the profile *plus* the complaint Xcode attaches to it — `IDEProvisioningErrorDomain`, `doesn't include`, `has expired` — sharing one hint constant so the rows cannot drift into three fixes for one lever. `profile-mentioned-while-compiling` is the negative fixture: a benign mention above a compile error keeps `build_failed_unclassified` and the cache-recovery hint. The `argv-names-a-provisioning-profile` fixture still proves the argv half.

All three narrowed rows are `invented-shape`, not `captured`: I know the wording families, not this Xcode's sentence for them, so `xcodeVersion` stays `unobserved` and `.device-evidence/CHECKLIST.md` §#2680 now asks the coordinator to record which of those lines actually prints. Un-captured stays un-claimed.

**PR body corrected.** The `BootFailureReason` "declared twice" clause is removed — that was true of the `runner/host.ts` mirror before #2689, not of `main` today. The argument left standing is the one that survives: it is `provision-kit`'s boot vocabulary shared with Android, and a build that never produced a binary has no boot to classify.

Gates at the new head: `build`, `format`, `check:quick`, `check:layering`, `check:gate-manifest`, `check:command-docs`, `check:production-exports`, `apple-runner` (503) and the `unit-core` help-conformance/eager-closure lane pass.

Not done, and not runnable from here: the live `prepare ios-runner` captures in §#2680 (items 1–4) are still owed by the device-evidence coordinator, so no fixture in this PR says `captured`.

thymikee:
This is a follow-up on the earlier review (85bd4de, https://github.com/callstack/agent-device/pull/2688#issuecomment-5745832538). Reviewed at 9621905.

Row 471 in https://github.com/callstack/agent-device/blob/9621905/packages/platform-apple/src/runner/runner-contract.ts#L471 only requires the substring 'expired', but toolTextIncludesAll scans the whole captured build log (execFailureDetails attaches full stdout/stderr, see packages/host-kit/src/internal/exec.ts#L703), so the pair can match across unrelated lines. A log with the benign `note: Using provisioning profile "X"` line plus any unrelated expiry sentence, such as an expired signing certificate or `Your session has expired`, satisfies both substrings and publishes signing_provisioning_profile_missing. A caller on the physical-device `prepare ios-runner` route that branches on details.reason would then be told to reinstall a profile for a certificate or session problem, which is the same wrong-lever failure this PR set out to remove. Can a profile row require a phrase that can only be a complaint about the profile itself, since the matcher is not line-scoped? Concretely, can row 471 require "has expired" (keeping "doesn't include" and the IDEProvisioningErrorDomain pair), with a negative fixture that pairs the benign profile note with an unrelated expired-certificate line and expects build_failed_unclassified?

All build-for-testing fixtures, including the three new profile rows, carry provenance 'invented-shape' or 'shipped-sniff-trigger' with xcodeVersion 'unobserved' (packages/platform-apple/src/runner/__tests__/runner-startup-failure-fixtures.ts#L151), so nothing at 9621905 shows real xcodebuild output reaching these rows instead of falling through to build_failed_unclassified. Can this be validated on a physical iPhone at 9621905975 by running `node --experimental-strip-types src/bin.ts --json prepare ios-runner --platform ios --device <iPhone>` for each case, pasting the full JSON envelope (details.reason, hint, logPath), `xcodebuild -version`, and the verbatim xcodebuild error lines: no AGENT_DEVICE_IOS_TEAM_ID/AGENT_DEVICE_IOS_PROVISIONING_PROFILE set, expecting signing_no_development_team; AGENT_DEVICE_IOS_BUNDLE_ID owned by another team, expecting bundle_identifier_already_registered; and AGENT_DEVICE_IOS_PROVISIONING_PROFILE set to a profile that doesn't cover the bundle id, reporting which of IDEProvisioningErrorDomain, "doesn't include", or "has expired" actually printed? Once these are captured, the matching fixtures can be marked captured.

Not blocking: the signing_no_profiles_for_bundle_id row at runner-contract.ts#L438-441 spells out the same reason+hint pair as literals instead of using the PROFILE_UNUSABLE constant added at line 276, so it could point at PROFILE_UNUSABLE too, but this can be taken or left.

No tests or device commands were run for this review; the claim that a profile-mentioned-while-compiling log matched the pre-delta bare row rests on reading the removed code, not on a run. The real Xcode wording for all three narrowed rows is unverified, so whether "doesn't include" arrives with an ASCII apostrophe, versus a typographic one that would fall through row 463 to the catch-all, is unknown. The author's gate list (build, format, check:*, apple-runner 503, unit-core) was not re-verified at this head.

Smoke Tests failed at test/integration/ios-simulator-e2e/live-assertions.ts#L52, with id="automation-longpress" not visible after scrolling, after the runner built and 14 other assertions passed; this looks unrelated, since the head commit touches only the build-failure rule rows, fixtures and CHECKLIST, which run only when xcodebuild build-for-testing throws, and the runner-session.ts step rename plus the DevToolsSecurity preflight import return early for kind !== 'device', with no hunk touching snapshot, scroll, or selector resolution.

The path to merge is narrowing row 471 to "has expired" with a cross-line negative fixture, then posting the three physical-device prepare captures at 9621905975 and marking the matching fixtures captured.

thymikee:
Both items in, rebased onto `main` `fc59620932`. New head: `bcfed35b4c`.

**The expiry row no longer reads a bare `expired`, and one line is now the unit of evidence.** Narrowing `expired` → `has expired` alone would not have fixed what you pointed at, so I went one step further: `toolTextIncludesAll` is a whole-log AND, and a whole-log AND cannot express "this complaint is about this profile". A `note:` about the profile the build used, plus `The certificate "Apple Development: …" has expired.` three lines below it, still satisfies `['provisioning profile', 'has expired']`. The rules table therefore gained a sibling axis, `toolTextLineIncludesAll` (`packages/platform-apple/src/runner/runner-contract.ts:188`), matched line by line over the same tool-text haystack, and all three profile rows use it: `IDEProvisioningErrorDomain` naming the profile, `doesn't include`, `has expired`. Nothing else in the table changed its axis.

**The cross-line negative is in and it is the proof, not decoration.** `profile-note-above-an-expired-certificate` pairs a benign profile note with an unrelated expired-certificate warning and asserts `build_failed_unclassified` with cache-recovery advice. Reverting just that row to `toolTextIncludesAll` makes it fail (`× a build-for-testing failure publishes build_failed_unclassified for profile-note-above-an-expired-certificate`); the row as shipped passes. `profile-xcode-signing-error` was reworded while I was there: it now puts the domain and the profile on the same line, because Xcode prints the profile inside the domain's own description and two lines that never met are two facts, not one complaint.

**Constant reuse done.** `signing_no_profiles_for_bundle_id` points at `PROFILE_UNUSABLE` (`runner-contract.ts:276`); the reason and hint literals are gone. That row keeps `toolTextIncludesAll` deliberately — "No profiles for 'com.example' were found" is the complaint and the subject in one sentence.

Fixtures stay honest: all three narrowed rows are still `invented-shape` with `xcodeVersion: 'unobserved'`. `.device-evidence/CHECKLIST.md` §#2680 now tells the coordinator to paste the line breaks, since which line carried which phrase is what would promote them.

Gates at the new head: `build`, `format`, `check:quick`, `check:layering`, `check:gate-manifest`, `check:command-docs`, `check:production-exports`, `apple-runner` (504) and the `unit-core` help-conformance/eager-closure lane pass; `fallow audit --base origin/main` is clean across the stack.

Still owed, and outside what I can run: the §#2680 prepare captures (no-team, foreign bundle id, missing profile, pinned-profile pairing). No fixture here says `captured`.

thymikee:
bcfed35 looks good and the code verdict is clean. Narrowing the profile rules so a name and its complaint must share one line is the right answer to the cross-line match from the last round, and it reuses the existing haystack instead of adding a second source of truth. CI is green with 20 checks passing, and the unit lane the new fixtures drive covers the changed classification path. No conflicts are known.

Not blocking: `bundle_identifier_unavailable` still pairs its two phrases across the whole log rather than one line (https://github.com/callstack/agent-device/blob/bcfed35/packages/platform-apple/src/runner/runner-contract.ts#L415), so one row is left on the old axis even though the rule table is what enumerates those sites; and the `profile-xcode-signing-error` fixture was edited to repeat the profile name inside the `IDEProvisioningErrorDomain` line (https://github.com/callstack/agent-device/blob/bcfed35/packages/platform-apple/src/runner/__tests__/runner-startup-failure-fixtures.ts#L161) rather than proved against a capture. Both can be taken or left.

One limit worth recording: the new fixture's pre-fix red state comes from reading the 9621905 rule table against the fixture output, not from a run, and the real xcodebuild wording for the `IDEProvisioningErrorDomain`, "doesn't include" and "has expired" diagnostics is still unobserved in this repo. Whether those phrases really share one line on a live signing failure is what the Phase B device capture in the checklist would settle, and that capture is also what would promote the three profile fixtures from invented shape to captured.

thymikee:
Rebased onto `main` @ `921a2cb3ac` (post-#2682). New head: **`15808ae228`** (was `bcfed35b4c`).

**The one nit — the bundle-identifier row was the last whole-log AND.** `bundle_identifier_unavailable` moved from `toolTextIncludesAll: ['app identifier', 'not available']` to `toolTextLineIncludesAll`, so the identifier and its availability have to meet in one sentence. It is the only multi-phrase row left on the whole-log axis; a grep for `toolTextIncludesAll: [...,...]` in `runner-contract.ts` now comes back empty.

A cross-line negative rides along, `app-identifier-and-availability-in-different-lines`: one line faults the identifier, another says something else is not available, and neither pairs them. Red with the row back on the whole-log axis, green as shipped, and marked `invented-shape` / `xcodeVersion: 'unobserved'` like the other un-captured shapes.

**The `.device-evidence/CHECKLIST.md` collision is resolved in #2682's favour.** That file is the merged lane's sheet and this PR no longer touches it; the runner-failure capture sheet lives beside it as `.device-evidence/CHECKLIST-runner-failures.md`, and the fixture provenance comment points there.

Gates at this head: `check:quick`, `check:layering`, `check:production-exports` clean; `vitest --project apple-runner` 54 files / 543 tests; `test:unit` 1363 files / 10767 tests.

Still owed, and unchanged in that: the physical-device captures in `.device-evidence/CHECKLIST-runner-failures.md` are coordinator Phase B on the connected iPhone. Nothing here was verified on hardware at this head.

thymikee:
At `15808ae` the code delta is clean and all 20 checks pass, so nothing in the code is holding this back.

What is still missing is the Phase B physical-iPhone capture, unchanged since the `bcfed35` round — and this delta is what makes it load-bearing. The four invented-shape rows still carry `xcodeVersion: 'unobserved'`, while `bundle_identifier_unavailable` and the two profile rows now require their phrases on a single line. If real xcodebuild output wraps `App Identifier` away from `not available`, a failure that used to be classified falls through to `build_failed_unclassified`, and only a real run settles which way it wraps. Please run `agent-device prepare ios-runner --platform ios --device "<iPhone>"` for each invented-shape row in `.device-evidence/CHECKLIST-runner-failures.md`, then paste the raw xcodebuild stdout with its line breaks intact plus `xcodebuild -version`, showing that `App Identifier` + `not available`, and `Provisioning profile` + (`IDEProvisioningErrorDomain` | `doesn't include` | `has expired`), each land on one line. A re-wrapped or paraphrased transcript will not prove it.

For that reason I have taken `ready-for-human` back off. My round-1 comment asked for the same captures and should not have applied it ahead of them; it goes back on with the captures, not with a code change.

Two things I did not check here: the apple-runner suite was not run, so the claim that the new cross-line fixture is red under the pre-delta rule is read from the old and new match functions rather than executed; and the single-phrase whole-log rows and the profile rows themselves are outside this delta, so they keep their earlier verdict.

Not blocking: the PR body still points at `.device-evidence/CHECKLIST.md`, which is now #2682's sheet rather than this stack's `CHECKLIST-runner-failures.md`, and still says "86 files, 443 production lines" where this head measures 11 files and +320.

thymikee:
## Phase B — signing-failure captures on a physical host (coordinator-run)

Host: `thymikee-iphone` (iPhone 17 Pro), `Xcode 26.2 / Build 17C52`, Apple Development identity team `7XSDF3H789`. Head `15808ae228`. Runner build path forces a device build with `-allowProvisioningUpdates`.

### Result: only `signing_provisioning_profile_missing` is reachable with a real build on this account
With a clean derived path (fresh `AGENT_DEVICE_IOS_RUNNER_DERIVED_PATH`, no cache reuse), `prepare ios-runner --platform ios` fails and the classifier publishes:

```
reason: signing_provisioning_profile_missing
hint:   Install/select a valid iOS provisioning profile, or set AGENT_DEVICE_IOS_PROVISIONING_PROFILE.
```

Verbatim `xcodebuild` errors (the matched source text):

```
AgentDeviceRunner.xcodeproj: error: No Accounts: Add a new account in Accounts settings. (target 'AgentDeviceRunner')
AgentDeviceRunner.xcodeproj: error: No Accounts: Add a new account in Accounts settings. (target 'AgentDeviceRunnerUITests')
AgentDeviceRunner.xcodeproj: error: No profiles for 'com.callstack.agentdevice.runner' were found: Xcode couldn't find any iOS App Development provisioning profiles matching 'com.callstack.agentdevice.runner'. (target 'AgentDeviceRunner')
AgentDeviceRunner.xcodeproj: error: No profiles for 'com.callstack.agentdevice.runner.uitests.xctrunner' were found: ... (target 'AgentDeviceRunnerUITests')
```

So the `signing_provisioning_profile_missing` row is now proven against **real Xcode 26.2 output** (`No profiles for '<id>' were found`), and the reason it emits matches the classifier.

### Happy path (control): signing succeeds
With the default derived path and automatic signing against the signed-in account, the same command **builds and starts successfully** (`com.callstack.agentdevice.runner` + `.uitests.xctrunner` signed with team `7XSDF3H789`). So signing works end-to-end on this machine.

### Two rows are NOT inducible on this account, and why
- **`signing_no_development_team`** — unsetting `AGENT_DEVICE_IOS_TEAM_ID` does **not** fail: automatic signing resolves team `7XSDF3H789` from the installed identity/profile and the build succeeds. A bogus `TEAM_ID`/account-less build does not surface a team-specific error; it short-circuits earlier at `No Accounts`, which classifies as `signing_provisioning_profile_missing`.
- **`bundle_identifier_already_registered`** — `AGENT_DEVICE_IOS_BUNDLE_ID=com.apple.TestFlight` (and a fresh derived path) does **not** reach the App-ID-registration stage: either the build succeeds under `-allowProvisioningUpdates`, or, without account context, it dies earlier at `No Accounts` / `No profiles ... were found`. The registration-conflict text this row keys on is never produced here.

**Implication for the two `invented-shape` fixtures:** these two rows cannot be upgraded from synthetic to captured on a valid-signing host with an active account. Reproducing them needs either an Apple account that is signed in but has *no* development team (for the no-team row) or a bundle id registered to a *different* team with automatic signing able to reach registration (for the already-registered row). I can capture both if pointed at such an account/machine; otherwise I recommend documenting them as environment-gated with the mechanism above and keeping the synthetic fixtures for the classifier unit coverage.

thymikee:
Thank you — these captures close the gap I held this on at `15808ae`. The real Xcode 26.2 output proves the `No profiles for '<id>' were found` row, and it also answers the worry behind my request: that diagnostic arrives as one long `error:` line rather than wrapped, so the sibling rows in the same provisioning family are far less likely to split the way I feared.

Your account-level explanation for `signing_no_development_team` and `bundle_identifier_already_registered` is the right call, and I would take the option you suggest. Please record that environment gating beside those fixtures so the rows read as blocked rather than unexamined, and include `bundle_identifier_unavailable`, `doesn't include` and `has expired` in the same note, since each of them needs the same kind of differently-configured account to induce.

I put `ready-for-human` back on, as I said it would go back with the captures.

thymikee:
Reviewed at 12de0a8. The delta does what I asked at 15808ae. The `No profiles for '<id>' were found` fixture now carries the real Xcode 26.2 output. The rows that need a differently configured account stay marked as not captured, and the account gating is written beside them. The classifier does not change in this delta.

The Smoke Tests failure is in the iOS simulator scroll check (the long-press test element did not become visible after scrolling). This PR changes only runner startup-failure classification, so the failure looks unrelated.

There are no conflicts, and `ready-for-human` stays on.

PR #2689, opened 2026-09-19
refactor(platform-apple): derive the runner host port instead of mirroring host-kit

### LINKED ISSUE #2658: refactor(platform-apple): derive the runner host port instead of mirroring host-kit
## Purpose

`packages/platform-apple/src/runner/host.ts` is the Apple runner host port: R77 keeps `runner/**` from value-importing `@agent-device/host-kit/*` so the seven Apple façade eager closures stay fixed. The port is right. Its shape is not: it is a hand-maintained **mirror** of host-kit, so every host-kit change is re-typed by hand.

Measured over the last 300 commits (main `91652a8fc5`): 403 LOC, 63 exports, 28 importers inside the package, 11 touches. Nine of the eleven were not moves; six of those were a host-kit signature change or one new capability, each costing 3 to 16 lines here plus the binding in `core/runner-host.ts` (#2598, #2599, #2595, #2470, #2423, #2160).

Four hand-kept layers, and each host-kit change touches two or three of them:

1. ~100 lines of copied type declarations (`ExecResult`, `RetryPolicy`, `ProcessLockOwner`, `DiagnosticEventInput`, …). Seven of the names do not exist in host-kit under that name (`ExecStreamOptions`, `RetryPolicy`, `RetryOptions`, `RetryAttemptContext`, `RetryTelemetryEvent`, `DiagnosticEventInput`, `DefinedEnvMap`) and are re-typed from scratch.
2. A 50-member `AppleRunnerHost` interface that re-declares every function signature.
3. Fifty one-line delegators (`export const runCmdSync: AppleRunnerHost['runCmdSync'] = (…) => requireHost().runCmdSync(…)`).
4. `core/runner-host.ts`, which lists every name a third time.

## Required behavior

Derive the port from the modules it fronts; keep the seam and the rule.

1. **Types.** Delete the copied type block. The 15 runner files that import those types take them from the owning package with `import type` (R77 exempts type-only imports; typecheck erases them). Export the seven missing names from the host-kit subpath that owns the function (`command`, `retry`, `diagnostics`) — a host-kit fix, not a platform-apple one.
2. **Interface.** Replace the re-declared signatures with a name list over the real modules:
   ```ts
   import type * as HostCommand from '@agent-device/host-kit/command';
   import type * as HostProcess from '@agent-device/host-kit/process';
   // one per subpath; ../core/tool-provider.ts, simctl.ts, physical-device-control.ts,
   // plist-xml.ts, runner-owner-state.ts for the Apple-owned members
   export type AppleRunnerHost =
     Pick<typeof HostCommand, 'runCmdStreaming' | 'runCmdSync' | 'runCmdBackground' | 'requireExecSuccess' | 'isCommandTimeoutError' | 'shellQuote'>
     & Pick<typeof HostProcess, 'isProcessAlive' | 'isProcessGroupAlive' | 'readProcessStartTime' | 'readProcessCommand' | 'signalPidsBestEffort' | 'signalProcessGroupBestEffort' | 'classifyOwnerLiveness'>
     & /* … */
     & { deadlineFromTimeoutMs: typeof HostRetry.Deadline.fromTimeoutMs };
   ```
   `import type * as` is a type-only declaration: R77 and the eager-closure gate see nothing. A signature change in host-kit reaches the runner with zero edits here.
3. **Delegators.** Keep the named exports so the 28 importers do not change, typed by derivation:
   ```ts
   function delegate<K extends keyof AppleRunnerHost>(name: K): AppleRunnerHost[K] {
     return ((...args: unknown[]) =>
       (requireHost()[name] as (...a: unknown[]) => unknown)(...args)) as AppleRunnerHost[K];
   }
   export const runCmdSync = delegate('runCmdSync');
   ```
   Keep the `Deadline` object shim as is (seven files call `Deadline.fromTimeoutMs`).
4. **Binding.** `core/runner-host.ts` spreads the real modules instead of listing names (`{ ...HostCommand, ...HostProcess, …, deadlineFromTimeoutMs: Deadline.fromTimeoutMs, ...appleCore }`). The `Pick` list becomes the single place that says what the runner may reach. `test-host.ts` is a Proxy and needs no change.

After this, adding a capability is one name in the `Pick` and one `delegate` line.

## Completion conditions

- Replayed as a thought test, #2598, #2599 and #2595 each become a zero-line change in `host.ts`.
- `pnpm check:layering` green with R77 and R13 untouched; eager-closure budgets for the seven Apple façade entries unchanged (no new value import enters `runner/**`).
- `typecheck`, `check:production-exports`, and the `apple-runner` vitest project green.
- `host.ts` ≈ 120 LOC (from 403). Report the number in the PR.

## Caveat

Spreading whole modules into the binding puts every export of those modules on the host object at runtime; the `Pick` type hides them. If `check:production-exports` or fallow objects, list the names explicitly in the binding instead. That keeps two lists, still with no hand-written signatures.

## Out of scope

- Retiring the port or R77. The eager-closure property still binds the runner subtree.
- Changing any host-kit behavior. Only type exports are added there.

thymikee:
## Summary

`runner/host.ts` was a hand-maintained mirror of host-kit. It is now derived from the modules it fronts, keeping the R77 seam and the eager-closure property. Closes #2658. 34 files touched.

- **Types.** The ~100-line copied block is gone. The 23 runner files that used those types take them from the owning package with `import type` (R77 exempts type-only imports). `DefinedEnvMap` was never a host capability — it moves to its only caller.
- **Interface.** `AppleRunnerHost` is `Pick`s over `import type * as` namespaces of host-kit, kernel, provision-kit and the Apple-owned core modules. No signature is written here.
- **Delegators.** The 41 named exports stay, typed by derivation through `delegate(name)`; the 28 importers are unchanged. The `Deadline` object shim stays.
- **Binding.** `core/runner-host.ts` keeps naming each import rather than spreading the modules whole — see *Rejected* below. It holds names only; it re-types nothing.

Host-kit gains three type exports the port needs by name and no behavior change: `ExecStreamOptions`, `DiagnosticEventInput`, and `DeadlineClock` — the read side of `Deadline`, which the class now `implements` and which `RetryOptions.deadline` accepts, so a test host can keep substituting its own clock now that the port no longer declares a structural copy.

The runner-facing slice of physical-device control moves to `core/physical-device-routing.ts`, below both sides. Without it the port reaches `physical-device-control.ts`, whose launch/terminate members are typed against the runner's own command executor — a 19-file type cycle that fails R9/R10.

**Replay.** #2599 (`ExecBackgroundOptions = Omit<ExecOptions, 'timeoutMs'>`) becomes a zero-line change here. #2598 and #2595 each drop from 11 and 3 lines in `host.ts` to 2: their type and signature changes are free, and what remains is the new capability itself — one `Pick` name plus one `delegate` line, as the issue predicts.

**`host.ts`: 403 → 187 LOC** (≈137 code, ≈50 doc). The issue's ≈120 estimate did not count the preserved R77 header.

**Known consequence.** Deriving from `typeof HostCommand` drops the old `ExecOptions` narrowing, so `cwd`, `binaryStdout` and `stdin` are now reachable through the port. That narrowing existed to keep the mirror small ("narrowed to what the runner uses"), and restoring it means re-declaring the exec signatures by hand — exactly what this PR deletes. The runner modules now import `ExecOptions` from host-kit directly anyway, so the port could not enforce it either.

**Rejected.** The issue's binding spreads the modules whole. Measured, that puts every export of all sixteen on the host object (41 members → 128), ships +1.8 kB including the vitest-only `resetAllProcessMemosForTests`, and makes a name shared by any two of them silently rebind a capability — a tripwire across four packages, catchable only at runtime. The issue's caveat licenses the named binding; tsc stays the sole enforcer of completeness.

## Validation

Tested at `5b29c32` (pre-push amend of the pushed head; tree identical).

Green: `pnpm typecheck`, `pnpm check:layering` (R77 and R13 untouched; R9 back to the merge-base's 6-file max), `pnpm check:production-exports`, `pnpm check:fallow` (no issues in 35 changed files), `pnpm lint`, `pnpm format:check`, `pnpm build`, `pnpm check:package`, `pnpm check:bundle-owner-files`, `pnpm check:gate-manifest`, `pnpm depgraph:test`, the `apple-runner` project (49 files / 480 tests), and the eager-closure budget suite (666 tests) — the seven Apple façade closures do not grow.

`pnpm check:affected --run` fails only its vitest lane, on three tests that fail identically on a clean `origin/main` checkout in this container (root user, so their `chmod`-based unwritable-path setup is a no-op): `app-log-session-resource.test.ts > unwritable tombstone plus rejecting cleanup blocks same-process replacement`, `capture-kit durable-capture/adoption.test.ts > a failed terminal transition…`, `capture-kit durable-capture-resource.test.ts > an unconfirmed failed-adoption cleanup…`. Everything else in `pnpm test:unit` passes (10470 tests). CI on Linux runners is the authority here.

No device-facing behavior changes: the port's runtime shape is the same object bound to the same functions.

https://claude.ai/code/session_01PPceQTJdXsEJaZsBM8ycek

---
_Generated by [Claude Code](https://claude.ai/code/session_01PPceQTJdXsEJaZsBM8ycek)_


thymikee:
Reviewed at 172f58e. No blocking findings: the port is now derived from the modules it fronts, the mirror is gone, and the change removes about 175 net production lines. The whole-module spread binding from the issue was measured and rejected in the PR body (41 to 128 members, +1.8 kB, and silent rebinding when two modules export the same name), and that reason holds. Not blocking: [`delegate()`](https://github.com/callstack/agent-device/blob/172f58e/packages/platform-apple/src/runner/host.ts#L135) casts through `unknown[]`, so forwarded arity is checked only at call sites, which tsc already covers. CI is green.

PR #2691, opened 2026-09-19
feat(ios-runner): disclose target re-activation on the command that caused it (#2682)

thymikee:
## Summary

Settled policy: **(a) disclose** — the runner keeps repairing foreground loss and now says so (`.device-evidence/DECISION.md`). (b) refuses the flow every caller already uses and cannot ship alone; ADR 0012 makes the status quo — an undisclosed `activate()` — the one unacceptable option.

The runner stamps the repair as response-level metadata on the command that paid for it: `targetActivation: { reason, priorState, otherActiveApplicationPid? }`, where `priorState` is the bound app's `XCApplicationState` read **before** `activate()` ran, and the pid is stated only when exactly one application other than the session app held an active accessibility session. `@agent-device/platform-apple`'s `runner/target-activation.ts` is the single decoder; the capture path appends the disclosure alongside existing warnings. Ships useful alone; #2693 adds the daemon seam.

No new runner command, so `CommandTraits` is untouched. `screenshot` stays lifecycle, #2438 in-place system-host serving and ADR 0005 `targetReset` are unchanged.

## Review round 1 (adversarial)

- **Identity over-claim (P1).** The wire fact proves "exactly one other application had an active accessibility session", not "this app held the foreground" — `activeApplications` exposes no ordering, and `DECISION.md` recorded that. So `foregroundPid` is now `otherActiveApplicationPid`, the shared sentence says the only other AX-active application rather than "held it", and the runner doc comments state the liveness limit. `.device-evidence/CHECKLIST.md` no longer pre-commits that a wrong pid is a bug; it scopes the claim to liveness. `priorState` is untouched — that is `XCUIApplication.state` read before `activate()`.
- **Unwired ACs (P1).** Both runner-side acceptance criteria are now pinned at the owning type in `UnitTests/RunnerTests+LifecycleCacheTests.swift`, which already swizzles `state` and `activate` and drives both branches: the `.runningForeground` call asserts `pendingTargetActivation == nil` (no happy-path work, nothing to disclose) and the `.runningBackground` call asserts the stamped `priorState` equals `XCUIApplication.State.runningBackground.rawValue` (captured BEFORE `activate()`). Verified on `apex-2682-proto`: `xcodebuild test-without-building … -only-testing:AgentDeviceRunnerUITests/RunnerTests/testActivateTargetSkipsForegroundAndActivatesNonForegroundApplication` → passed.
- **Live lane was in no CI lane (P1).** `ios.yml` and `replays-manual.yml` use an explicit file list, so the live lane is now added to both `--test` lists inside the existing `AGENT_DEVICE_IOS_E2E=1` env blocks. It also follows the shared harness conventions: `createContext()` tier/session (`ios-e2e-<tier>-<pid>`), artifacts under the harness `artifactDir` instead of `/tmp/agent-device-2682-*.png`, `--json` assertions on parsed `data` instead of scraped text, `runCmd('xcrun', ['simctl','openurl',…])`, and a bounded poll for the handoff instead of `spawnSync('sleep', ['6'])`.
- **Raw enum decoded by array index (P2).** `XCApplicationState` is not in any header this repo compiles against, so the decode is an explicit `{0:'unknown',1:'notRunning',2:'runningBackground',3:'runningBackgroundSuspended'}` table. An unmapped raw value no longer drops the fact back into a silent activation: the repair is disclosed with `priorState: 'unknown'` and the gap is reported through an injected `onUnmappedPriorState` sink that `snapshot-presentation.ts` wires to `emitDiagnostic` (kept off the decoder itself, which stays pure and host-free — R77).

## Validation

Head `daaf445c17`, simulator `apex-2682-proto` (iPhone 17 Pro, iOS 26.2), fixture `com.callstack.agentdevicelab`:

- BEFORE: `open` → handoff (`simctl openurl https://example.com`) → `screenshot` shows Safari "Example Domain" → `snapshot -i` answers the fixture tree, silently.
- AFTER: the same `snapshot -i` carries `…(prior state runningBackground), so the runner activated it before answering (reason stale_target). Any capture taken earlier in this session described the only app other than the session app with an active accessibility session (pid 33878)…`; host `ps` maps 33878 to `MobileSafari.app`. The next `snapshot -i` discloses nothing.
- Runner unit lane: the lifecycle-cache activation test passes with both new assertions (build with `AGENT_DEVICE_XCUITEST_INCLUDE_UNIT_TESTS=1`, run with `-only-testing:`).
- Green: `format`, `check:quick`, `check:layering`, `check:production-exports`, `check:xctest-selection`, `test:unit` (10529 passed, 1 skipped), targeted decoder/contracts tests, and the coverage classifier lane. `check:affected --json` plans the full set (fails open on the contracts manifest, as documented); `--run` and the physical lane (`thymikee-iphone`, `.device-evidence/CHECKLIST.md`) stay with the coordinator/CI.

Commands that consume no capture (`press <x> <y>`, `press @ref` off a live ref frame) still stamp without disclosing — the interaction projection has no decoder for the stamp. Tracked as #2694, and #2693's docs state that silence proves nothing.

## Round 2: the live lane, and why it is manual instead of CI-gated

Adding `smoke-ios-target-activation.test.ts` to the iOS Smoke job surfaced three real defects in the lane, all fixed and kept:

- **Nothing was installed.** The job fetches the fixture with `install: 'false'` and the fixture E2E installs it through the public CLI as its first step; this lane opened it instead. It installs the same way now.
- **Two lanes, one simulator.** `node --test` gives every file its own process and runs them concurrently, so the activation lane and the fixture lane were relaunching and closing under each other — the fixture lane died at "cold launch fixture" while this lane waited for a repair nobody owed.
- **A render gate that only one build variant could pass.** Waiting for a home-screen card (`Automation lab`, then `Gesture lab`) fails against a development build, which draws the server menu instead. The gate is now a string the app draws on any of its own surfaces.

The remaining CI failure is not a product bug and is not chased further: on the CI simulator the runner reads its own target as `.runningForeground` while a foreign app is demonstrably on screen, so nothing is activated and no disclosure is owed. The uploaded artifact of [run 35459451230](https://github.com/callstack/agent-device/actions/runs/35460535722/job/105943468492) pairs a rendered app screenshot with a Settings screenshot and 31 captures of the bound app at ~450 ms each. Asserting the disclosure there measures `XCUIApplication.state` on a headless simulator, so the lane is env-gated/manual and its allowlist entry is reverted. Filed: [#2696](https://github.com/callstack/agent-device/issues/2696).

The AC "assert the field on a real device response" is satisfied by the physical-device run the coordinator performs next, plus the local simulator proof in the Validation section above and the rendered capture cited in #2693.

Also fixed from the CI audit: the lifecycle-cache assertions above, the raw-value table's coverage assertion, and `UnmappedPriorStateDetail.reason`, which now carries the reason type the decoder had already validated (`IosTargetActivationReason` had lost its only consumer).

## Round 3: human review, before merge

Head `81dd06c8d9` on `main` @ `d688926167`. Three items, all in-commit:

- **The lane asserted a field this head does not publish.** `smoke-ios-target-activation.test.ts` polled `data.targetActivation`, which #2693 introduces; at this head the interactor keeps the fact on its own result and publishes only the warning sentence, so the lane would have spent its 90 s deadline on a field that never appears. It now asserts the `data.warnings` sentence, and the helper (`:174`) parses the `prior state`/`reason` out of it, gates them against `IOS_TARGET_ACTIVATION_PRIOR_STATES` and `isIosTargetActivationReason`, requires both agent routes, and re-renders the fact through `iosTargetActivationDisclosure` — so a paraphrase fails instead of passing a substring grep. Lane re-run green (32.9 s).
- **`.device-evidence/CHECKLIST.md` quoted a shape PR1 does not emit** (`data.targetActivation: {…}`); it now quotes the rendered sentence and names the typed object as PR2's seam (`:89`).
- **Declared-reasons predicate.** `isIosTargetActivationReason` sits beside the declared reasons so a new reason cannot silently skip the decoder's predicate, and the decoder's `as IosTargetActivationReason` cast is gone (`runner/target-activation.ts:49`).

Simulator captures at this head, gate counts, and the handoff's before/after screenshot hashes: https://github.com/callstack/agent-device/pull/2691#issuecomment-5750019870. CI green (17 pass, 1 skipping).

## Round 4: the nicety — the lane recognises the disclosure by rebuilding it

Head `65e02e3c70`. `activationFactFrom` matched the disclosure with a hard-coded copy of the builder's opening sentence, so a reword of `iosTargetActivationDisclosure` would have returned `undefined` forever and the lane would have gone back to timing out at 90 seconds for something it could no longer see. It now asks the builder for every sentence a declared fact can produce — bare, and with an impossible pid (`88888888`) opened into a capture group — and only a response carrying one of those exact sentences counts. `disclosureStem()` derives the shared opening from the same calls, and the timeout text now separates "the sentence changed shape" from "the device never repaired anything" (`missingDisclosureEvidence`, `smoke-ios-target-activation.test.ts:171`).

Lane re-run on `apex-2682-proto` twice at this head: 30.1 s and 13.6 s, both passed. Details: https://github.com/callstack/agent-device/pull/2691#issuecomment-5750518716


thymikee:
Reviewed at daaf445. Two issues block this.

The live lane at [smoke-ios-target-activation.test.ts#L57](https://github.com/callstack/agent-device/blob/daaf445/test/integration/smoke-ios-target-activation.test.ts#L57) polls `snapshot -i --json` until `json.data.targetActivation` appears. On this head that fact does not reach daemon response data: neither `buildSnapshotState` ([snapshot-state.ts#L42-L93](https://github.com/callstack/agent-device/blob/daaf445/packages/capture-kit/src/snapshot-state.ts#L42-L93)) nor `captureSnapshotAttempt` ([snapshot-capture.ts#L131-L150](https://github.com/callstack/agent-device/blob/daaf445/src/daemon/snapshot-capture.ts#L131-L150)) copies it, and that plumbing is in #2693. The fact is one-shot, so the first poll consumes it and `captureUntilDisclosed` likely runs out its 90 s budget and fails. The test is now wired into the ios.yml and replays-manual.yml smoke steps, so the iOS Smoke job is likely to go red. Can the lane assert what this PR emits, the shared disclosure sentence in `data.warnings`, with a passing run recorded on daaf445 or later? Otherwise the `data.targetActivation` assertions and the CI wiring belong in #2693.

`IOS_TARGET_ACTIVATION_PRIOR_STATES` at [ios-target-activation.ts#L30](https://github.com/callstack/agent-device/blob/daaf445/packages/contracts/src/ios-target-activation.ts#L30) has no consumer outside its own file now that the decoder reads `PRIOR_STATE_BY_RAW_VALUE` ([target-activation.ts#L30](https://github.com/callstack/agent-device/blob/daaf445/packages/platform-apple/src/runner/target-activation.ts#L30)), and the four states are listed twice. This likely explains the Compatibility & Provenance failure on this head. Please derive `IosTargetActivationPriorState` from the raw-value table, or make it a plain literal union, and drop the unused export. `UnmappedPriorStateDetail` also looks exported only for a test.

Is 277 production lines across runner, contracts and platform-apple the smallest shape? One literal-union type, with decoding in the shared runner response reader instead of the snapshot reader, could give the same result with less surface and without a second decoder later. #2693 also declares the same vocabulary again in `kernel/snapshot.ts`, so it is worth choosing one owner (contracts) before that PR lands.

Not blocking: the decoder runs only from `readAppleSnapshotResult`, so `interaction_foreground_guard` cannot surface and a `press` or `fill` right after a handoff drops its fact silently, which is worth stating in the PR body with a pointer to #2694; also, `.device-evidence/` is a new committed root, not an existing repo convention.

Live evidence is still needed. On a simulator at daaf445 or later, open the fixture, take a screenshot, run `simctl openurl https://example.com`, confirm Safari is on screen, then run `snapshot -i`. It should show the disclosure sentence with a non-foreground prior state and the matching `AGENT_DEVICE_RUNNER_ACTIVATE_FACT` line in runner.log, and the next `snapshot -i` should show nothing. Record the commit SHA. The simulator pass in the PR body was at 458dad3, where the test asserted stdout. The physical-device checklist run is also still pending.

Compatibility & Provenance failed on this head and is likely related, for the export reason above. The Smoke Tests job was still running.

thymikee:
Reviewed at 122c67c. This follows up on the earlier review at daaf445 (https://github.com/callstack/agent-device/pull/2691#issuecomment-5743116863). The delta adds a fixture install step and a doc comment to the smoke test, but it does not change what the test checks, so the earlier finding still applies.

`captureUntilDisclosed` in [smoke-ios-target-activation.test.ts#L101-L109](https://github.com/callstack/agent-device/blob/122c67c/test/integration/smoke-ios-target-activation.test.ts#L101-L109) polls `data.targetActivation`, and line 59 reads `repaired.json!.data.targetActivation`. On this head the fact does not reach the daemon response. `interactor.ts` stamps `targetActivation` on the raw result, but `snapshotCaptureAnnotationsFrom` ([snapshot-capture-annotations.ts#L31-L40](https://github.com/callstack/agent-device/blob/122c67c/packages/contracts/src/snapshot-capture-annotations.ts#L31-L40)) reads only analysis, androidSnapshot, freshness, quality and warnings, and `snapshot-command-runtime.ts` rebuilds the response from that object. Only `data.warnings` carries the disclosure sentence. The iOS Smoke job at 122c67c failed on exactly this lane with `no foreground disclosure within 90000ms` ([job](https://github.com/callstack/agent-device/actions/runs/35453451684/job/105924495698)). The newer head 2dc9c83 switches the handoff to Settings and checks that the screen changed, but it still polls `data.targetActivation`, so the handoff change alone is unlikely to fix it. Can the lane assert that `data.warnings` contains `iosTargetActivationDisclosure(fact)`, as it already does at L68-L71? Or should this assertion and its CI wiring move to #2693, where the `data.targetActivation` plumbing lands?

`IOS_TARGET_ACTIVATION_PRIOR_STATES` ([ios-target-activation.ts#L29](https://github.com/callstack/agent-device/blob/122c67c/packages/contracts/src/ios-target-activation.ts#L29)) is still exported with no consumer. Can the export be dropped and the raw-value table stay module-private?

Smoke Tests is likely related, as shown above. Compatibility & Provenance fails on the same unused export, which this PR adds. There are no conflicts with `main`.

Next: make the live lane check `data.warnings` or move it to #2693, drop the unused export, and post a passing run of `smoke-ios-target-activation.test.ts` on a simulator at the current head.

thymikee:
Reviewed at a00162a. The newer head bd1091b only changes how the smoke test reports its runner-log note, so this review also applies there.

The earlier finding from the [122c67c review](https://github.com/callstack/agent-device/pull/2691#issuecomment-5743509339) still applies. `captureUntilDisclosed` ([smoke-ios-target-activation.test.ts#L146](https://github.com/callstack/agent-device/blob/a00162a/test/integration/smoke-ios-target-activation.test.ts#L146)) waits for `data.targetActivation`, but `snapshotCaptureAnnotationsFrom` ([snapshot-capture-annotations.ts](https://github.com/callstack/agent-device/blob/a00162a/packages/contracts/src/snapshot-capture-annotations.ts)) still drops that field on this head. The delta moves the handoff to Settings and copies the runner log on failure, but it does not change the route or the check. So the lane polls for 90 seconds and fails whatever the runner does, and it gives no live evidence for #2682. Can the lane assert only what this head publishes, the disclosure sentence in `data.warnings` of the first `snapshot -i` after the handoff, and leave the typed-field checks to #2693? Or can the lane and its CI wiring move to #2693?

No run at this head shows a command response with the disclosure yet. So it is still open whether the runner stamps the fact on a real simulator, or whether the lane fails only because of the check above.

The new test in [target-activation.test.ts#L25](https://github.com/callstack/agent-device/blob/a00162a/packages/platform-apple/src/runner/__tests__/target-activation.test.ts#L25) gives `IOS_TARGET_ACTIVATION_PRIOR_STATES` a real reader and clears the earlier unused-export point; typing `PRIOR_STATE_BY_RAW_VALUE` as a `Record<IosTargetActivationPriorState, number>` could make that check a compile-time rule instead, but this is optional.

Smoke Tests is likely related: it runs the test this PR adds and fails in `captureUntilDisclosed` for the reason above. Compatibility & Provenance now passes. There are no conflicts with `main`.

Next: make the live check read only what this head publishes, or move the lane to #2693. Then post a passing run of `smoke-ios-target-activation.test.ts` on an iOS simulator at the current head, where the first `snapshot -i` after the handoff carries the disclosure and the second carries none.

thymikee:
Reviewed at dff0b2e. The two new commits wait for the `Gesture lab` card before the handoff. That card is on the fixture's home screen, so the new wait itself looks fine.

The finding from the [a00162a review](https://github.com/callstack/agent-device/pull/2691#issuecomment-5743891622) still applies. `captureUntilDisclosed` ([smoke-ios-target-activation.test.ts#L150](https://github.com/callstack/agent-device/blob/dff0b2e/test/integration/smoke-ios-target-activation.test.ts#L150)) waits for `data.targetActivation`, but `snapshotCaptureAnnotationsFrom` ([snapshot-capture-annotations.ts](https://github.com/callstack/agent-device/blob/dff0b2e/packages/contracts/src/snapshot-capture-annotations.ts)) still does not publish that field on this head. So the lane waits 90 seconds and fails whatever the runner does, and it gives no live evidence for #2682. Can the lane check only the disclosure sentence in `data.warnings` of the first `snapshot -i` after the handoff, or can the lane and its CI wiring move to #2693?

Smoke Tests is still running at this head. It runs this test, so it is likely to fail for the reason above. There are no conflicts with `main`.

Next: make the live check read only what this head publishes, or move the lane to #2693. Then post a passing run of `smoke-ios-target-activation.test.ts` on an iOS simulator at the current head.

thymikee:
This is a follow-up on dff0b2e (https://github.com/callstack/agent-device/pull/2691#issuecomment-5744327888). The doc block for the manual lane is now a run-me-yourself recipe and the handoff moved back to Safari, but the check inside it still targets a field this head does not publish.

test/integration/smoke-ios-target-activation.test.ts polls `last.json?.data?.targetActivation` (https://github.com/callstack/agent-device/blob/7a43342/test/integration/smoke-ios-target-activation.test.ts#L167) and asserts on `data.targetActivation` at L95 and L118. On this head, `packages/platform-apple/src/interactor.ts#L245` puts `targetActivation` on the interactor result, not on the daemon response, and `packages/contracts/src/snapshot-capture-annotations.ts#L30-L41` only forwards analysis/androidSnapshot/freshness/quality/warnings. A grep for `targetActivation` under `src/` finds no reader that would put it back on the response. So running this lane as written will time out after 90s no matter what the runner stamps, because it is waiting on a field nothing publishes. What this head does publish is the sentence in `data.warnings` (the test already builds it at L106 as `iosTargetActivationDisclosure(fact)`). Can the lane assert on that sentence instead of `data.targetActivation`, and leave the typed-object assertions to #2693?

Not blocking: the CHECKLIST quote at .device-evidence/CHECKLIST.md#L89 shows the typed `targetActivation` object even though line 35 of the same file says that field ships in #2693 and this head only publishes the warnings sentence, and separately, is it worth adding an `isIosTargetActivationReason` predicate next to `IOS_TARGET_ACTIVATION_REASONS` in packages/contracts/src/ios-target-activation.ts so target-activation.ts#L55 doesn't need the `as IosTargetActivationReason` cast — take or leave both.

I read the head with git show and grep and did not run any test; I confirmed the disclosure sentence is appended, not swapped in, inside interactor.ts, but I did not trace the rest of the daemon response assembly to confirm warnings from different sources merge rather than overwrite, and that code is untouched by this PR. I could not check the CI artifact behind the #2696 claim (run 35460535722), so the "simulator reports runningForeground with a foreign app on screen" reasoning for making this lane manual comes from the description, not from something I verified.

The delta also removes the lane from ios.yml#L292 and replays-manual.yml#L173, restoring both files byte-for-byte to 0ebd254, so the iOS Smoke route this PR previously reddened is no longer exercised there. CI reports 18 checks green, and none of them run the smoke-ios-target-activation lane.

What's needed before merge: point the manual lane at `data.warnings` (what this head actually publishes) instead of `data.targetActivation`, fix the CHECKLIST quote to match, and post a two-capture run at this head — the first `snapshot -i` right after the `simctl openurl` handoff carrying the disclosure sentence with a non-foreground prior state, and the following `snapshot -i` carrying none. The PR body's BEFORE/AFTER at daaf445 shows that sentence, but the lane as written here asserts a field nothing publishes, so it can't produce that run yet.

thymikee:
All three before-merge items are in. New head: **`81dd06c8d9`** (base `main` @ `d688926167`; the stack's child is #2693).

**1. The manual lane now asserts the `data.warnings` sentence, not `data.targetActivation`.**
`test/integration/smoke-ios-target-activation.test.ts` polled `last.json?.data?.targetActivation` — right you are, at this head the interactor keeps the fact on its own result (`packages/platform-apple/src/interactor.ts:245`) and publishes only the sentence (`:277`), so the lane would have burned its 90 s deadline.

- poll: `test/integration/smoke-ios-target-activation.test.ts:157` — `activationFactFrom(last.json?.data?.warnings ?? [])`
- first capture: `:104`, second capture: `:113`
- helper: `:174` `activationFactFrom(warnings)` — finds the sentence, requires its `prior state` to be one of `IOS_TARGET_ACTIVATION_PRIOR_STATES` (a set that omits `runningForeground`, so "prior state claims the repair's outcome" cannot pass), requires its `reason` to satisfy `isIosTargetActivationReason`, requires both agent routes, and then **rebuilds it with `iosTargetActivationDisclosure(fact)` and requires that to be in `warnings`**. A paraphrase of the sentence therefore fails instead of passing a substring grep, and no typed field is needed to check it.

Lane re-run at this head on a local simulator: `✔ live iOS runner discloses a foreground repair on the command that paid for it (32959ms)`.

**2. CHECKLIST quote fixed.** `.device-evidence/CHECKLIST.md:89` now quotes the rendered sentence and says plainly that on this head the sentence *is* the whole disclosure, with the typed `data.targetActivation` object named as PR2's daemon seam.

**3. Non-blocking, done.** `isIosTargetActivationReason` lives next to the declared reasons (`packages/contracts/src/ios-target-activation.ts:25`) and the decoder dropped its cast (`packages/platform-apple/src/runner/target-activation.ts:49` — `if (!isIosTargetActivationReason(reason)) return undefined;`, no `as IosTargetActivationReason` left anywhere in the tree). Pinned by a new contracts test that accepts every declared reason and refuses an undeclared one.

**Simulator captures at this head** (`apex-2682-proto`, iOS 26.2, `session apex2682-pr1`; the lane's own Safari handoff):

```
$ xcrun simctl openurl C2748D97-B92D-4B3B-9C65-AF86494AD904 https://example.com   # Safari comes forward
$ agent-device snapshot -i
Page: com.callstack.agentdevicelab
App: com.callstack.agentdevicelab
Snapshot: 15 visible nodes
The session app was not foreground when this command arrived (prior state runningBackground), so the runner activated it before answering (reason stale_target). Any capture taken earlier in this session described the only app other than the session app with an active accessibility session (pid 33878), not the session app. Re-capture now that the session app answers, or drive the other app in its own session.
Simulator AX snapshot unavailable (foreground-owner-unverified); used XCTest for this app generation.
@e1 [application] "Agent Device Tester"

$ agent-device snapshot -i
Page: com.callstack.agentdevicelab
App: com.callstack.agentdevicelab
Snapshot: 15 visible nodes
Simulator AX snapshot unavailable (circuit-disabled); used XCTest for this app generation.
@e1 [application] "Agent Device Tester"
@e2 [other] "Tab Bar"
```

Handoff was visible, not inferred: the pre/post `screenshot` sha256s differ (`afe4e14027cc…` before, `16bddb634d57…` after), and `runner.log` carries `AGENT_DEVICE_RUNNER_ACTIVATE … state=2 reason=stale_target` plus `AGENT_DEVICE_RUNNER_ACTIVATE_FACT … priorState=2 otherActiveApplicationPid=33878` — which is what the sentence reports.

**Gates:** `pnpm format` no-diff, `check:quick`, `check:layering`, `check:production-exports` (exit 0), `check:xctest-selection` (0 methods reachable by no lane), `fallow audit --base origin/main` clean (20 files), `pnpm test:unit` 10649 passed | 1 skipped, targeted vitest for the decoder + contracts 10 passed, `node:test` lane 14 passed / 1 skipped (this file skips without `AGENT_DEVICE_IOS_E2E=1`). Physical `thymikee-iphone` untouched.

thymikee:
Reviewed at 81dd06c. This is ready to merge on the code.

The prior finding from 7a43342 is fixed: the disclosure now carries the command that caused re-activation, and the response builder's output matches what the smoke test asserts byte-for-byte.

Not blocking: activationFactFrom (test/integration/smoke-ios-target-activation.test.ts:175) matches the disclosure by a hard-coded copy of the builder's opening sentence, so a reword of iosTargetActivationDisclosure would make the finder return undefined and bring back the old 90s dishonest-timeout failure mode; deriving the probe from the builder's own output would close that gap, but it's fine to take or leave.

I did not run tsc or the test suite, and I did not exercise the manual or simulator lane, so the type-check and the new contracts test's pass are author-reported, not verified by me here.

Coverage, both Smoke Tests jobs, and the physical-device checklist run were still in progress at 81dd06c, so there's nothing to attribute from CI yet; note that smoke-ios-target-activation.test.ts isn't in the ios.yml or replays-manual.yml test lists at this head, so the Smoke Tests jobs don't cover this lane even once they finish.

thymikee:
Nicety done — the lane no longer quotes the builder, it asks the builder. New head: **`65e02e3c70`** (base `main` @ `fc59620932`; child #2693 rebased onto it).

**The probe is now derived, not copied.** `activationFactFrom` (`test/integration/smoke-ios-target-activation.test.ts:202`) enumerates `DECLARED_ACTIVATION_FACTS` — every declared prior state × reason pair (`:185`) — and asks `iosTargetActivationDisclosure(fact)` for the sentence each one produces: once bare, and once with `otherActiveApplicationPid: 88888888` (a pid no live process can have) whose escaped text is opened into a `(\d+)` capture group. A warning counts only if it matches one of those builder-produced sentences exactly, so:

- there is no opening-sentence literal left anywhere in the file, and no regex over prose;
- rewording `iosTargetActivationDisclosure` rewrites the probe with it instead of silently returning `undefined` and reviving the 90-second timeout;
- the pid still round-trips, and `warnings.includes(builder(fact))` is now structurally guaranteed rather than asserted after a paraphrase was accepted.

**And the timeout now says which failure it hit.** `disclosureStem()` (`:222`) derives the longest opening every builder sentence shares — again from the builder, not from a copy — and `missingDisclosureEvidence` (`:171`) puts that in the failure text: "A warning opened like the shared disclosure sentence but no declared fact rebuilt it" (the sentence changed shape, or the fact is undeclared) versus "No warning opened like the shared disclosure sentence (…)" (the device never repaired). The old message could not tell those apart, which is exactly how the dishonest timeout stayed hidden.

Lane re-run at this head on `apex-2682-proto` (iOS 26.2), twice: `✔ live iOS runner discloses a foreground repair on the command that paid for it` — 30.1 s, then 13.6 s after extracting the message helper (that extraction also cleared the last `fallow` complexity finding; `fallow audit --base origin/main` is now clean over 20 files).

Also in this head, unchanged in intent from before: the CHECKLIST quote fix, and `isIosTargetActivationReason` beside the declared reasons with the decoder's cast dropped.

`format`/`format:check` no-diff, `check:quick` (lint + full `typecheck`), `check:layering` 246 pass, `check:production-exports` exit 0, `check:xctest-selection` 0 reachable by no lane, `node:test` lane 14 passed / 1 skipped (this file skips without `AGENT_DEVICE_IOS_E2E=1`). Physical `thymikee-iphone` untouched, no merge.

thymikee:
This is ready at 65e02e3. Checks show all-green, but they run a lane that never touches the file this delta changes, so the passing state confirms nothing about this fix. I did not run the node:test file against a live simulator or device, since that needs AGENT_DEVICE_IOS_E2E=1 plus a fixture app and udid; I re-executed the extracted string-derivation logic (activationFactFrom, disclosureStem, escapeRegExp) against all 32 declared sentence forms plus adversarial inputs, and it held. Nothing here blocks landing. Not blocking: in `test/integration/smoke-ios-target-activation.test.ts` around https://github.com/callstack/agent-device/blob/65e02e3/test/integration/smoke-ios-target-activation.test.ts#L220, `activationFactFrom` no longer fails fast on a warning that opens like the stem but doesn't match a declared fact, so that case now waits out the full 90s HANDOFF_DEADLINE_MS before the final assert distinguishes it from "nothing arrived yet" — worth checking `missingDisclosureEvidence` inside the poll loop for an immediate `assert.fail`, but this can be taken up later since the lane is manual and excluded from CI.

PR #2693, opened 2026-09-19
feat(#2682): disclose the target re-activation on every capture-consuming command

### LINKED ISSUE #2682: Silent target re-activation lets a screenshot and the next snapshot describe different apps

thymikee:
## What

Carries the runner's activation fact (#2682, #2691) through the daemon, so the command that paid for a foreground repair is the command that reports it. Settled policy from the prototype lane: **disclose, do not refuse** (`.device-evidence/DECISION.md`).

- `IosTargetActivation` moves to `packages/kernel/src/snapshot.ts`, where `SnapshotState` can hold it; `@agent-device/contracts/ios-target-activation` re-exports it and owns the shared disclosure sentence, so the CLI and the daemon cannot drift apart in prose.
- `src/daemon/system-surface-disclosure.ts` → `src/daemon/capture-disclosure.ts`: the #2438 surface disclosure plus `withTargetActivationDisclosure`, and `withCaptureDisclosures` for routes that owe both. The disclosure is response-level metadata and its sentence is **appended** — staleness, quality, and surface warnings that came first are never replaced.
- `targetActivation` survives every response level, including `--level digest`.
- CLI JSON keeps the structured fact next to the warning (`serializeSnapshotResult` / capture-annotations projection), so scripted callers never parse prose.
- `appstate` on Apple refuses honestly and names the per-command disclosure in its hint; `appstate` help, `website/docs/docs/commands.md` and `sessions.md` say the same thing, including the new "Foreground repairs on iOS" section.

## Review round 1 (adversarial)

- **False attribution (P1).** Selector reads may answer from a tree an earlier command stored (`reusableLastSnapshot`, `reusableSessionSnapshot`). Stamping the disclosure off the *consumed* tree therefore told a command "the session app was not foreground when **this** command arrived" for a request that never looked at the device. The repair no longer rides the consumed tree at all: it travels on a per-request `activationProof` filled only by a capture the request actually took (`captureSelectorSnapshot`, the snapshot route's own `execute`, the interaction's target capture), first fact-carrying capture wins so a later fact-less re-capture in a poll cannot erase it. `consumedTree` keeps its #2438 job — it describes what the response actually shows. Pinned by `selector-capture-runtime.test.ts` (cache hit reports nothing; a later fact-less capture does not erase an earlier proof).
- **Snapshot route read stored state (P1).** `snapshot-runtime.ts` read `sessionStore.get()?.snapshot` unconditionally, so even a command that failed before capturing could have inherited the previous command's repair into `error.details.hint`. The fact is now bound to the capture inside `execute`; the only source is this request's own proof.
- **Silence presented as safety (P1).** The docs said a command that consumed no capture "stays silent", which reads as "no repair happened". That is ADR 0012's undisclosed activation wearing a disclaimer: the runner may well have re-activated for a coordinate `press`, and the interaction projection drops the stamp. `commands.md` now states that silence proves nothing, tells the caller to spend a `snapshot -i` when the distinction matters, and links #2694, which tracks closing the gap.
- **Identity (P1, with #2691).** Renamed here too: `otherActiveApplicationPid`, and the shared sentence claims only "the only app other than the session app with an active accessibility session". `grep -rn foregroundPid` is 0 across the tree.
- **Docs that were simply wrong (P2).** `sessions.md` still claimed iOS `appstate` works; the `appstate` bullet pointed at nothing ("described above"); `bundle_changed` was defined by feel; `missing_after_wait` and `interaction_foreground_guard` were never documented. All fixed, with the liveness-not-foreground limit on the pid and the "never `runningForeground`" note on `priorState`.

## Simulator evidence (`apex-2682-proto`, iOS 26.2, head `aa2f310189`)

`pnpm build && pnpm clean:daemon`, off-app handoff with `xcrun simctl openurl <UDID> https://example.com` before each command:

| command | outcome |
| --- | --- |
| `snapshot -i` | `targetActivation {reason, priorState, otherActiveApplicationPid}` + appended warning |
| `press 'label="INFO"'` | disclosed (`reason: stale_target`) |
| `get text 'label="INFO"'` | disclosed |
| repaired `snapshot -i`, then `is visible 'label="INFO"'` twice | first discloses, both followers report `null` — the false attribution is gone on the shipped path, not just in the unit test |

## Gates

- `pnpm format` (repo-wide), `pnpm check:quick`, `pnpm check:layering`, `pnpm check:production-exports`, `pnpm check:xctest-selection`
- `pnpm test:unit` → 10529 passed, 1 skipped
- New/updated tests: `capture-disclosure-target-activation` (incl. a failure that must not borrow the stored repair), `selector-capture-runtime` (cache hit + first-wins), `snapshot-runtime-target-activation`, `interaction-target-activation-disclosure`, `src/commands/capture/output.test.ts` now renders through the shared sentence instead of copying it
- `pnpm check:affected --json` (plan only) → fails open only on #2691's `packages/contracts/package.json` + `scripts/layering/contracts-exports.snapshot.json`

No Swift changed in this PR; #2691 owns the runner side. The physical-device lane stays with the coordinator (`CHECKLIST.md`).

Closes #2682

## Round 1b: the CI audit

`Compatibility & Provenance` failed on this branch for three findings, all fixed rather than suppressed:

- `IosTargetActivationReason` had no consumer, so the decoder's unmapped-prior-state sink now carries the reason type it had already validated (`#2691`) instead of a bare string.
- `dispatchInteractionCommand` had grown to nine commands plus a default. The touch cases stay exactly as ADR 0011's structural guard requires — one delegation each, still asserted by `interaction-response-construction-guard.test.ts` — and the five commands that aim no gesture at a captured target moved to their own dispatch.
- `snapshotView` had one conditional spread per optional signal; the signals a digest must carry are now a declared list.

`fallow audit --base apex/2682-target-activation-disclosure` is clean (28 changed files), `check:layering`, `check:production-exports` and `test:unit` (245 targeted + full suite) are green.

The iOS Smoke job is no longer expected to run #2691's activation lane: on the CI simulator the runner reads its own target as `.runningForeground` while a foreign app is demonstrably on screen, so it activates nothing and owes no disclosure there. The lane is env-gated/manual and the allowlist entry is reverted; the divergence is filed as [#2696](https://github.com/callstack/agent-device/issues/2696).

## Local capture, freshly built daemon (`apex-2682-proto`, iOS 26.2, head `38d6c9db8a`)

`pnpm build && pnpm clean:daemon`, `install` + `open com.callstack.agentdevicelab`, then
`xcrun simctl openurl <UDID> https://example.com`, then `snapshot -i --json` twice:

```
FIRST   fact    = {"reason":"stale_target","priorState":"runningBackground","otherActiveApplicationPid":33878}
FIRST   warning = The session app was not foreground when this command arrived (prior state runningBackground),
                  so the runner activated it before answering (reason stale_target). Any capture taken earlier in
                  this session described the only app other than the session app with an active accessibility
                  session (pid 33878), not the session app. Re-capture now that the session app answers, or drive
                  the other app in its own session.
SETTLED fact    = null      warning = NONE
```

pid 33878 is `MobileSafari.app`, resolved by host `ps`. The same sequence through the lane runner
(`AGENT_DEVICE_IOS_E2E=1 … --test test/integration/smoke-ios-target-activation.test.ts`) passes in
~12 s against a freshly built daemon. One caveat worth recording: a **stale** daemon from another
worktree's build answers those captures with the warning and no structured field, which is exactly
the confusion `pnpm clean:daemon` exists to prevent.

## Round 2b: human review, before merge

Head `7977596e28` (`6d8cd10030` daemon, `7977596e28` CLI text) on `apex/2682-target-activation-disclosure` @ `81dd06c8d9`. Five items, each with a test that is red without the fix:

- **One dedupe gate for both carriers.** `appendDisclosure` + `carriesDisclosure` (`capture-disclosure.ts:93`/`:128`) now gate on the sentence's presence across `data.warnings`, `data.warning` and `error.details.hint`, not on `response.ok`, so a failing command that passes through two disclosure wrappers names the repair once. `capture-disclosure-target-activation.test.ts:111`, `:148`.
- **The proof is recorded where the repair is paid.** Recording moved into `runCapture` (`selector-capture-runtime.ts:222-227`), so a recovery re-capture that replaces the first tree cannot discard what the first tree paid for. `find` holds one proof per request and uses it at all three exits, including the sparse one (`find.ts:132`, `:165-169`) — previously find's only exit that disclosed nothing. `wait`'s decoration capture records too (`wait-current-surface.ts:93`). `find-target-activation-disclosure.test.ts:106`, `:120`.
- **Every capture-consuming text formatter prints the `Warning:` line.** `snapshot` did; `press`, `wait`, `get`, `is` and `find` rendered only the singular `warning`, so the same repaired capture read aloud on one command and silently on the others. One renderer now (`output-common.ts:49`/`:69`, reused by `settle-output.ts:71`), and it skips a warning the text already holds so two wrappers cannot say it twice. `get --format attrs|text` stay parseable. `interaction/output.test.ts:324/334/343/353/358`, `capture/wait.test.ts:9`.
- **`snapshot` stopped reading stored session state** for its disclosure (`snapshot-runtime.ts:86`): the #2438 sentence already rides the capture's annotations, so this removed a double copy on success and an invented surface on failure.
- **The `storedRepair` route test exists** (`snapshot-runtime-target-activation.test.ts:89`) and is red without the change. Its sibling `:61` is green-only, because the route overwrites the stored snapshot before the wrapper reads it — recorded as a guard, not as a red test.

Also: `isIosTargetActivationReason` moved to kernel, where the declarations live on this branch; digest carry-set test (`response-views.test.ts:53`); the Apple `appstate` refusal hint no longer promises it "reads the active session state" (`runtime.ts:106`); stray line removed from `CHECKLIST.md`.

Two gaps that are **not** closed here, stated rather than papered over: `find`'s sparse-first capture cannot be forced on a healthy simulator, so that path rests on the two route tests; and a *satisfied* `wait text` still prints nothing, because the runner stamps the repair on its text-observation op and only the snapshot payload is decoded — the formatter side is covered (`capture/wait.test.ts:9`, and `wait stable` prints it on device). Named on #2694, whose fix shape already covers non-snapshot commands.

Gates at this head: `format`, `check:quick`, `check:layering`, `check:production-exports` (exit 0), `check:xctest-selection`, `fallow audit --base apex/2682-target-activation-disclosure` clean (48 files), targeted vitest 449 files / 3218 tests, `pnpm test:unit` **10677 passed**, 1 skipped. CI green (19 pass, 1 skipping). Device outputs for `press`/`is`/`wait stable`/`find exists`/`get text`: https://github.com/callstack/agent-device/pull/2693#issuecomment-5750019996.

## Round 2c: the carrier rule moved to the dispatcher

Head `05be687e76`. A reviewer found that `getCliOutput` appended the activation warning only when `format` was neither `text` nor `attrs` — and `get`'s format is a required two-value field on every surface, so that branch was unreachable, its test built an `input: {}` no caller can produce, and a successful `get text` after a repair printed the value alone. Four per-formatter wrappers and the settle wrapper were saying the same rule in four other ways.

The rule now lives in one place with one declared exception:

- `formatCliOutput` (`src/commands/cli-output.ts:27`) appends every response's `Warning:` lines after the command's own text, for every command. `routeResponseWarnings` (`output-common.ts:73`) owns the rendering and the once-only dedupe.
- The only opt-out is `parseableOutput?: true` on the command's descriptor (`command-contract.ts:58`), declared for `get`, `find`, `clipboard` and `record`, whose stdout *is* the value or artifact path a caller pipes: those print their warnings on **stderr**, so `value=$(agent-device get text …)` still captures exactly the value. `--json` writes neither line. MCP has one stream, so it merges the stderr half back (`mcp/command-tools.ts:522`).
- Per-formatter carve-outs removed, including the unreachable `get` branch and its dead test; `getCliOutput` is total over `text | attrs`; the settle wrapper renders only its diff, so two wrappers cannot say a warning twice.
- The enumeration is executable, not prose: `src/commands/cli-output.test.ts` pins the declared set, forbids the trait on any formatter-less command, and asserts the rule for all 47 formatter commands against each command's real payload shape. Red-without measured: 48 of 57 with the routing removed, 8 with the trait removed.
- What that rule exposed: 11 commands had no formatter at all — including the capture-consuming `type`, `focus`, `swipe` and `gesture` — and printed no warning whatever the response carried. The generic CLI path routes the same way now (`cli/commands/generic.ts:80`), asserted on stdout for `type`.
- Also: the first-wins activation recording at three capture paths collapsed into `recordActivationProof` (`capture-disclosure.ts:33`); `@agent-device/contracts/ios-target-activation` stopped re-exporting five kernel symbols so each declaration has one spelling and its test moved to kernel with the compile-time half of its claim; `commands.md` names the four parseable-output commands and adds a satisfied `wait <text>` to the "Silence is not proof" bullet.

Simulator captures for `get text` and `find … get text` at this head, red-without numbers, and the gate list: https://github.com/callstack/agent-device/pull/2693#issuecomment-5750518836


thymikee:
Reviewed at fba103a. The branch conflicts with its base, `apex/2682-target-activation-disclosure`, which moved to daaf445 and renamed `foregroundPid`, so it needs a rebase first.

`remember` overwrites `consumedSnapshot.state` on every capture ([selector-capture-runtime.ts#L81](https://github.com/callstack/agent-device/blob/fba103a/src/daemon/selector-capture-runtime.ts#L81)), and [interaction-runtime.ts#L58](https://github.com/callstack/agent-device/blob/fba103a/src/daemon/interaction/internal/interaction-runtime.ts#L58) does the same. Example: `wait 'label="X"' 8000` after an off-app handoff. Poll 1 re-activates the app and carries the fact, poll 2 is a clean capture and overwrites it, so the final response has no `targetActivation`. The command that paid for the repair then reports nothing, which is the silent case #2682 is meant to close. Could one request-scoped holder keep the first `targetActivation`, shared by selector, wait and interaction? A two-poll wait test through `dispatchWaitViaRuntime` would cover it.

`withTargetActivationDisclosure` writes the sentence to `data.warnings` ([settle-output.ts#L67](https://github.com/callstack/agent-device/blob/fba103a/src/commands/settle-output.ts#L67), [capture-disclosure.ts#L49](https://github.com/callstack/agent-device/blob/fba103a/src/daemon/capture-disclosure.ts#L49)). But the text formatters for press, click, fill, longpress, hover and scroll render only `data.warning`, and the formatters for wait, find, get and is never read `data.warnings`. With the default text output, only snapshot shows the disclosure; the other commands show it only with `--json` or MCP. Could every capture-consuming formatter render warnings through `messageWithWarningsText` / `readResponseWarnings` ([output-common.ts#L38](https://github.com/callstack/agent-device/blob/fba103a/src/commands/output-common.ts#L38))? A rendering test for press and wait would confirm it.

The snapshot route discloses against `sessionStore.get(sessionName).snapshot`, not its own capture ([snapshot-runtime.ts#L77](https://github.com/callstack/agent-device/blob/fba103a/src/daemon/snapshot-runtime.ts#L77)). When capture admission fails, or the Android timeout response returns, the stored tree is from an earlier command, so its `targetActivation` or surface sentence lands on the failure as if this command caused it. The selector fallback for get, is and find ([selector-runtime.ts#L110](https://github.com/callstack/agent-device/blob/fba103a/src/daemon/selector-runtime.ts#L110)) has the same issue when no capture ran. Disclosure should use only the request's own capture. Since the snapshot capture annotations already publish the fact, does the snapshot-route wrapper add anything?

`withCaptureDisclosures` now also applies `withSystemSurfaceDisclosure` to snapshot responses, which had none before. iOS in-place system sheets already put the surface sentence in `warnings[]` (interactor.ts#L274), and Android shade snapshots already get `formatAndroidSystemSurfaceWarning`. So this adds a second copy as `data.warning` and changes #2438 JSON and MCP snapshot responses without a docs or changelog note. Should it come out together with the snapshot-route wrapper?

Could one request-scoped capture holder with a sticky `targetActivation`, appended once in the shared response finalization, replace the five per-route wrappers and the separate interaction holder? That would fix the overwrite and stale-fallback problems by construction. The first step would be one owner of the consumed-capture holder, shared by selector, wait and interaction, without the stored-session fallback in `consumedSessionSnapshot` for this fact.

Not blocking: after the rebase, the kernel type, the docs JSON example, the commands.md wording and the test sentence in output.test.ts still use the old `foregroundPid` meaning ([snapshot.ts#L211](https://github.com/callstack/agent-device/blob/fba103a/packages/kernel/src/snapshot.ts#L211)); `responseWarnings` duplicates the kernel's `readResponseWarnings` with a weaker filter; and contracts re-exports the kernel's types under a second name, which fallow flags as unused.

Live evidence is still needed after the rebase and fixes, on an iOS simulator with `simctl openurl` handing off before each command: a text-mode `screenshot` then `snapshot -i` that shows the Warning line, with the SHA; a `wait 'label="X"'` where X appears only after two or more polls, showing `targetActivation` in `--json` and the Warning line in text; and a text-mode `press 'label=...'` that shows the Warning line. The simulator evidence in the PR body is from before the base moved.

Compatibility & Provenance is likely related: fallow flags the new unused type re-exports in contracts ([ios-target-activation.ts#L17-L18](https://github.com/callstack/agent-device/blob/fba103a/packages/contracts/src/ios-target-activation.ts#L17-L18)) and complexity in `dispatchInteractionCommand` and `snapshotView`, both changed here. Smoke Tests were still running.

thymikee:
Reviewed at 3bb640f. This follows up on the earlier review at fba103a (https://github.com/callstack/agent-device/pull/2693#issuecomment-5743117087). The new `RequestActivationProof` fixes the stale attribution on the snapshot route, but four issues remain, and the live evidence is from before the rebase. The newer head 40236d9 only rebases onto the updated #2691 branch and does not change this PR's own patch, so the points below still apply.

`captureSelectorSnapshot` records the first-wins proof on the tree it returns, not on each device capture ([selector-capture-runtime.ts#L117](https://github.com/callstack/agent-device/blob/3bb640f/src/daemon/selector-capture-runtime.ts#L117)). In `find` sparse recovery (`legacyIosSparse` or `sparseVerdictQueryScope`), the first `runCapture` re-activates the app, carries `targetActivation`, and is often sparse right after activation. `recoverLegacySparseIosSnapshot` or `recoverSparseVerdictWithQueryScope` then runs a second `runCapture` against an app that is now in front, and first-wins sees only that second tree, which has no fact. The comment at L115 says it covers a recovery re-capture, but it does not seem to. Then a `find` or another selector read that hits sparse recovery after an off-app handoff pays for the repair and reports nothing, which is the silent case #2682 is meant to close. Could the proof be recorded inside `runCapture`, the one place every selector device capture passes? One shared `recordActivationProof(proof, snapshot)` helper could then serve `capture-disclosure.ts`, `interaction-runtime.ts` and `snapshot-runtime.ts`. A test with a sparse first capture that has `targetActivation`, followed by a clean recovery capture, would pin it.

The text output is unchanged from the previous review ([settle-output.ts#L67](https://github.com/callstack/agent-device/blob/3bb640f/src/commands/settle-output.ts#L67)). `withTargetActivationDisclosure` writes to `data.warnings`, but `appendResponseNotes` renders only `data.warning` for click, press, fill, longpress, hover and scroll; `wait` renders through `messageOutput` only; and the get, is and find formatters in `src/commands/interaction/output.ts` never read warnings. With default text output, only `snapshot` shows the disclosure, so a `press`, `wait`, `get`, `is` or `find` that re-activated the app prints no Warning line. Please render `readResponseWarnings(data)` in every formatter for a command that can carry this disclosure (in `appendResponseNotes` and in the wait, get, is and find formatters, or through `messageWithWarningsText`), with rendering tests for `press` and `wait`.

[snapshot-runtime.ts#L85](https://github.com/callstack/agent-device/blob/3bb640f/src/daemon/snapshot-runtime.ts#L85) still passes the stored session snapshot into `withSystemSurfaceDisclosure`. This route had no surface disclosure at the base (122c67c), and the capture annotations already publish the surface warning ([snapshot.ts#L273](https://github.com/callstack/agent-device/blob/3bb640f/src/commands/capture/runtime/snapshot.ts#L273)), so this adds a second copy as `data.warning`. When a snapshot fails before it captures, the earlier command's surface sentence is appended to `error.details.hint`, so a failed snapshot can report a system surface it never saw. It also changes the #2438 output shape again without a docs note. Can this route call only `withTargetActivationDisclosure(response, activationProof.state)` and leave the surface sentence to the capture annotations?

No test uses the `storedRepair` branch in `scenario()` at [snapshot-runtime-target-activation.test.ts#L54](https://github.com/callstack/agent-device/blob/3bb640f/src/daemon/__tests__/snapshot-runtime-target-activation.test.ts#L54). If `snapshot-runtime.ts` went back to reading the stored snapshot as the proof, all new tests here would still pass. A route test with `storedRepair: true`, where the capture throws or returns no `targetActivation`, that asserts no `targetActivation` in the response and no disclosure in `error.details.hint`, would pin the stale-attribution fix.

The simulator evidence at [.device-evidence/CHECKLIST.md#L80](https://github.com/callstack/agent-device/blob/3bb640f/.device-evidence/CHECKLIST.md#L80) is from aa2f310, before the rebase and the fix commit. It also does not show the text-mode Warning line, a multi-poll wait, or a `screenshot` then `snapshot -i` pair. On an iOS simulator at the current head, with a handoff to another app before each command, please show: a text-mode `screenshot` then `snapshot -i` with the Warning line; a `wait 'label="X"' 8000` where X appears only after two or more polls, with `targetActivation` in `--json` and the Warning line in text; a text-mode `press 'label=...'` with the Warning line; and a `find` that hits sparse recovery and still reports `targetActivation`.

The new proof holder answers most of the earlier design question. Recording it in `runCapture`, as in the first point, would also remove the three copies of the first-wins code.

Compatibility & Provenance is likely related: fallow flags two unused re-exports in `packages/contracts/src/ios-target-activation.ts` (`IOS_TARGET_ACTIVATION_PRIOR_STATES` and `IosTargetActivationReason`), plus complexity in `dispatchInteractionCommand` and `snapshotView`, which this PR changes. The Coverage failure looks unrelated: it comes from unhandled rejections in `request-router-open-claim.test.ts` after test teardown, a file this PR does not change. There are no conflicts with the base branch.

Next: record the activation proof in `runCapture`, render `data.warnings` in every capture-consuming text formatter, drop the stored-snapshot surface wrapper from the snapshot route and add its route test, fix the fallow findings, then post the live simulator run at the current head.

thymikee:
This is a follow-up on 2d8e234 (https://github.com/callstack/agent-device/pull/2693#issuecomment-5743508849). The disclosure now fires on more routes, but at 5b39210 it can still print twice on one carrier, stay silent on the recovery path it was built for, and stay silent in default text mode.

`get` and `is` are dispatched from `handleInteractionCommands` with the shared activation proof (interaction.ts:30-35), and selector-runtime.ts already runs `withCaptureDisclosures` on that same proof. On an `!ok` response, `withTargetActivationDisclosure` takes the no-dedupe branch (https://github.com/callstack/agent-device/blob/5b39210/src/daemon/capture-disclosure.ts#L58), so a failing `get text 'label="nope"'` or `is visible ...` after a foreground repair gets the four-sentence disclosure appended twice to `error.details.hint`. Can the two carriers (`data.warnings` and `error.details.hint`) share one gate, so any wrapper skips a disclosure the response already contains, and the sentence is guaranteed to appear once per response regardless of how many routes it passes through?

The proof is recorded off what `captureSelectorSnapshot` returns, not off each device capture (https://github.com/callstack/agent-device/blob/5b39210/src/daemon/selector-capture-runtime.ts#L119). `recoverLegacySparseIosSnapshot` and `recoverSparseVerdictWithQueryScope` discard the first capture and return a fresh one, so the first capture's `targetActivation` never reaches the caller, and `find.ts` returns its sparse snapshot before any disclosure wrapper runs. What enumerates every selector device capture, and can the proof be recorded there once (for example inside `runCapture`) so the sparse-recovery path — the path an off-app handoff makes likely — still reports the repair?

`withTargetActivationDisclosure` writes to `data.warnings`, but `appendResponseNotes` only renders `data.warning` for press/click/fill/longpress/hover/scroll, `wait` uses `messageOutput`, and the get/is/find formatters never read warnings at all; only snapshot output renders `data.warnings`. What enumerates the capture-consuming text formatters, and can each of them render `readResponseWarnings(data)` through the existing `messageWithWarningsText`/`messageWithWarningsOutput`, so a default-mode `press`, `wait`, `get`, `is`, or `find` that re-activated the app prints the Warning line the way `snapshot` already does?

`snapshot-runtime.ts`'s `consumedTree` still reads `params.sessionStore.get(params.sessionName)?.snapshot` (https://github.com/callstack/agent-device/blob/5b39210/src/daemon/snapshot-runtime.ts#L86). On a failing snapshot that tree is the previous command's, so the #2438 surface sentence gets appended to `error.details.hint` for a surface this command never saw; on success it's a second copy next to the capture annotations' own warnings entry. Should this route call only `withTargetActivationDisclosure(response, activationProof.state)` and leave the surface sentence to the capture annotations, so a failed snapshot never claims a surface it didn't observe?

`snapshot-runtime-target-activation.test.ts`'s `storedRepair` branch is never exercised with `true` (https://github.com/callstack/agent-device/blob/5b39210/src/daemon/__tests__/snapshot-runtime-target-activation.test.ts#L54), so the stale-attribution fix this PR names as P1 has no test that fails without it. Can a route test cover `storedRepair: true` where the capture throws or returns no `targetActivation`, asserting neither a `targetActivation` field nor a disclosure appears in `error.details.hint`?

Not blocking: the digest carry-set has no test proving `targetActivation` survives it, a stray `D` line sits in `.device-evidence/CHECKLIST.md` near the new simulator-lane section, and the Apple runtime refusal hint still opens with "reads the active session state" while `commands.md` now says iOS appstate is unavailable — worth fixing but fine to leave for now.

I read the PR-head sources via `git show` for all of this; I ran no test or device command, didn't re-derive the exact error code a failing `is`/`get` returns, and couldn't verify the PR body's fallow-audit or test-count claims beyond the packet's check totals. The checks report all 20 passing, and this delta's `dispatchInteractionCommand`/`snapshotView` changes match the two functions the prior complexity findings named. I know of no conflicts on this branch. What's still needed before this is ready: record the proof inside the shared capture path and route find's sparse return through it, render warnings in every capture-consuming text formatter, stop reading the stored session snapshot in the snapshot route and add the missing route test, dedupe the disclosure on the error carrier, then post a run at the current head on an iOS simulator that runs `xcrun simctl openurl <UDID> https://example.com` before each command and shows: default-text `press`, `is visible`, and `wait` each printing the Warning line; `get text 'label="nope"'` after a handoff showing the disclosure once, not twice, in the error hint; and a `find` whose first capture returns a sparse verdict still reporting `targetActivation`.

thymikee:
All five before-merge items are in, each with a test that is red without the fix. New head: **`7977596e28`** (base `apex/2682-target-activation-disclosure` @ `81dd06c8d9`, which is on `main` @ `d688926167`). Two commits: `6d8cd10030` (daemon) and `7977596e28` (CLI text).

**1. One dedupe gate, shared by both carriers.** `src/daemon/capture-disclosure.ts:93` `appendDisclosure(response, disclosure, carrier)` is now the only door a disclosure goes through, and it opens once: `carriesDisclosure` (`:128`) checks `data.warning`, `data.warnings[]` and `error.details.hint`, and a response that already holds the sentence comes back untouched. `withTargetActivationDisclosure` (`:50`) still lands the typed fact even when the words were already spoken, because the field is that response's own claim.

- red-without / green-with: `src/daemon/__tests__/capture-disclosure-target-activation.test.ts:111` (two nested wrappers → `hint.split(sentence).length - 1 === 1`) and `:148` (capture already spoke it in `data.warning` → no copy in `data.warnings`, typed field still present). Reverting `capture-disclosure.ts` fails exactly those two.
- on the simulator, the failing `get text` case you named prints it once (below).

**2. The proof is recorded in the shared capture path, and `find`'s sparse return routes through it.** `src/daemon/selector-capture-runtime.ts:222-227` records inside `runCapture` — every capture a request takes, including a recovery re-capture whose tree then *replaces* the one that activated. `find` owns one `RequestActivationProof` for the whole request (`src/daemon/interaction/internal/find.ts:132`), passes it into its target capture, and uses it at all three exits — including the sparse one, `:165-169`, which was previously find's only exit that disclosed nothing. `wait`'s decoration capture records the same way (`src/daemon/wait-current-surface.ts:93`, wired from `src/daemon/wait-runtime.ts`).

- red-without / green-with: `src/daemon/interaction/internal/__tests__/find-target-activation-disclosure.test.ts:106` (first capture legacy-sparse + stamped, recovery tree clean → response must still carry `targetActivation`; without the change `data.targetActivation` is `undefined`) and `:120` (tree stays sparse → the failure's hint carries it). Reverting `selector-capture-runtime.ts` + `find.ts` fails both. Route-level wait coverage: `src/daemon/__tests__/capture-disclosure.test.ts:379`, unit-level: `src/daemon/wait-current-surface.test.ts:105`.

**3. Every capture-consuming text formatter prints the `Warning:` line.** One renderer owns it: `src/commands/output-common.ts:49` `appendWarningLinesText` (reads `readResponseWarnings` plus the pre-existing singular `warning`), reused by `messageWithWarningsText` (`:38`), by `withResponseWarnings` (`:69`), and by the settle wrapper (`src/commands/settle-output.ts:71`, which used to render only `data.warning` and would otherwise say it twice). Entries: `src/commands/interaction/output.ts:90-93` (`fill`/`longpress`/`hover`/`scroll`) and `:99-100` (`is`, `find`), `get`'s default branch in `getCliOutput`; `wait` is `messageWithWarningsOutput` (`src/commands/capture/wait.ts:71`). `get --format attrs`/`text` deliberately stay untouched — those outputs are piped.

- tests: `src/commands/interaction/output.test.ts:324` (press), `:334` (is), `:343` (find), `:353` (get), `:358` (attrs stays parseable), `src/commands/capture/wait.test.ts:9`. Reverting the four CLI files fails 5 of them.

**4. `snapshot` stops reading the stored snapshot.** `src/daemon/snapshot-runtime.ts:86` is now `withTargetActivationDisclosure(response, activationProof.state)` — no `consumedTree` from `sessionStore.get(sessionName)?.snapshot`. The #2438 surface sentence already rides the capture's own annotations, so this removes both the double copy on success and the invented surface on failure.

**5. The `storedRepair: true` red test exists.** `src/daemon/__tests__/snapshot-runtime-target-activation.test.ts:89` — a snapshot refused before it captured (unresolvable ref scope) with the stored tree carrying both `targetActivation` and `iosSystemSurfaceBundleId` asserts no `targetActivation` in details, no disclosure and no surface sentence in `error.details.hint`. Reverting `snapshot-runtime.ts` fails it (`expect(false).toBe(true)` on the leaked hint). `:61` covers the sibling case — a capture that answered with no repair must not borrow the stored one; I'll be straight with you: that one is green before the change too, because the route overwrites the stored snapshot before the wrapper reads it, so it is a guard rather than a red test.

**Non-blocking, all done:** digest carry-set test `src/daemon/__tests__/response-views.test.ts:53` (green-only: the field already survives collapse — it now cannot silently stop); stray `D` line removed from `.device-evidence/CHECKLIST.md`; the Apple `appstate` refusal hint no longer promises it "reads the active session state" — it says `Apple appstate is unavailable: the Apple target answers no sessionless foreground probe…` (`packages/platform-apple/src/runtime.ts:106`, assertion moved to `runtime.test.ts:101`).

**Simulator captures** (`apex-2682-proto`, iOS 26.2, head `7977596e28`, default text mode, `xcrun simctl openurl … https://example.com` + 7 s before each command):

```
$ agent-device press 'label="INFO"'
Tapped label="INFO" (372, 144)
Warning: The session app was not foreground when this command arrived (prior state runningBackground), so the runner activated it before answering (reason bundle_changed). Any capture taken earlier in this session described the only app other than the session app with an active accessibility session (pid 33878), not the session app. Re-capture now that the session app answers, or drive the other app in its own session.

$ agent-device is visible 'label="INFO"'
Passed: is visible
Warning: The session app was not foreground when this command arrived (prior state runningBackground), so the runner activated it before answering (reason stale_target). Any capture taken earlier in this session described the only app other than the session app with an active accessibility session (pid 33878), not the session app. Re-capture now that the session app answers, or drive the other app in its own session.

$ agent-device wait stable 4000
Warning: The session app was not foreground when this command arrived (prior state runningBackground), so the runner activated it before answering (reason stale_target). Any capture taken earlier in this session described the only app other than the session app with an active accessibility session (pid 33878), not the session app. Re-capture now that the session app answers, or drive the other app in its own session.

$ agent-device find 'label="INFO"' exists
Found: true
Warning: The session app was not foreground when this command arrived (prior state runningBackground), so the runner activated it before answering (reason stale_target). Any capture taken earlier in this session described the only app other than the session app with an active accessibility session (pid 33878), not the session app. Re-capture now that the session app answers, or drive the other app in its own session.

$ agent-device get text 'label="nope"'
Error (COMMAND_FAILED): Selector did not match: label="nope"
Hint: Selector text/label values match exactly (quote multi-word values: text="Sign in"). Run snapshot -i to see current elements and refs, or use find <text> for contains matching.
The session app was not foreground when this command arrived (prior state runningBackground), so the runner activated it before answering (reason stale_target). Any capture taken earlier in this session described the only app other than the session app with an active accessibility session (pid 33878), not the session app. Re-capture now that the session app answers, or drive the other app in its own session.

$ rg -c "was not foreground" gettext-miss.txt
1
```

**Two things I could not show on the simulator, said plainly:**

- **`find` whose *first* capture returns a sparse verdict.** A healthy iOS simulator never gives me a sparse tree — the AX/XCTest acquisition answers readably, and I could not force the verdict without faking it. That path is proven by the two route tests above (`…/find-target-activation-disclosure.test.ts:106`, `:120`, both red without the change); the ordinary `find` path is the simulator output above.
- **A satisfied `wait text` printing the line.** The formatter does print warnings (unit test + the `wait stable` capture above), but on the device the runner stamps the repair on its *text-observation* op, and the daemon's decoder only reads the snapshot payload — so a satisfied `wait text` still shows nothing. Same family as #2694 (a command whose device work isn't a capture owes a disclosure channel); I left the decoration-capture recording in place so a timed-out wait that re-activated during its surface capture does disclose. Say the word and I'll file it or extend #2694 instead of leaving it in a PR comment.

**Gates:** `format` no-diff, `check:quick`, `check:layering`, `check:production-exports` (exit 0), `check:xctest-selection` (0 unreachable), `fallow audit --base apex/2682-target-activation-disclosure` clean (48 files), targeted vitest 449 files / 3218 tests, `pnpm test:unit` 10677 passed | 1 skipped, `node:test` registration lane 14 passed / 1 skipped. Physical `thymikee-iphone` untouched; no merge.

thymikee:
This is a follow-up on the 5b39210 review. At 7977596, getCliOutput in src/commands/interaction/output.ts#L26 still only appends the activation warning when `format` is neither `'text'` nor `'attrs'`, but `get`'s format field is `requiredField(enumField(['text','attrs']))` (src/commands/interaction/metadata.ts#L171) and the CLI reader throws on anything else via readGetFormat (src/commands/cli-grammar/common.ts#L215). No surface — CLI, MCP, or SDK — can produce a value that reaches that branch, so a successful `ad get text 'label="INFO"'` after a foreground repair prints the value alone, and `get` stays silent in the default text mode even though it's one of the five routes #2682 names. The new test at src/commands/interaction/output.test.ts#L353 passes `input: {}`, which no real caller can construct, so it's green against dead code, and the posted simulator capture only shows the failing `get` on its error-hint path. Can you settle this at one place: either render the warning for `format === 'text'` (on stdout or stderr, your call, as long as stdout stays pipeable), or write down that `get` is a declared parseable-output exception and delete the unreachable branch and its test? Either way, something should enumerate which formatters render the warning and which are declared exceptions, so the next formatter can't silently miss it.

Not blocking, take or leave: find's text branches get the warning only because withResponseWarnings wraps the whole entry rather than the declared-exception rule from the point above (src/commands/interaction/output.ts#L100), the first-wins activation recording is hand-copied at three call sites with nothing enumerating them (src/daemon/selector-capture-runtime.ts#L222, wait-current-surface.ts#L91, interaction/internal/interaction-runtime.ts#L57) and could be one exported `recordActivationProof`, the ios-target-activation re-export block in packages/contracts/src/ios-target-activation.ts#L15 leaves two spellings of the same kernel symbols for three importers that could just import from `@agent-device/kernel/snapshot` directly, and the commands.md#L795 "Silence is not proof" bullet lists only commands that consume no capture even though a satisfied `wait <text>` reportedly carries no activation fact either.

Is the smaller shape here to append the warning once in the single dispatcher, formatCliOutput (src/commands/cli-output.ts#L17), for every command, with one declared opt-out for commands whose stdout is parsed by scripts — would that avoid both the `get` gap and the hand-wrapped `find` case in one move, instead of per-formatter carve-outs in interaction/output.ts? That would need a `parseableOutput`-style trait on the command descriptor first, rather than logic spelled out per formatter.

What's missing before this can merge is a run at 7977596 of a successful default-text read after a handoff — `xcrun simctl openurl <UDID> https://example.com`, then `agent-device get text 'label="INFO"'` against a selector that matches — showing whether stdout carries the Warning line or only the value, since the posted captures cover only the failing `get`. A capture of `agent-device find 'label="INFO"' get text` alongside it would confirm the `find` case in the same run. No tests or device commands were run for this pass; the route tracing above comes from reading the diff at head. The claim that a satisfied `wait <text>` carries no activation stamp, and MCP/SDK reachability of `get`'s default branch, are both inferred from the registry metadata and grammar code, not from a live call.

Smoke Tests was still in progress at 7977596, so there's no failing check to attribute yet, but the diff touches the same paths that job exercises (test/integration/smoke-ios-target-activation.test.ts, the daemon capture/disclosure code, and the CLI output paths), so a later failure there would be worth checking against this PR rather than waved off as unrelated. The next thing to settle is one carrier rule for parseable text output so `get text` and `find … get text` don't disagree, backed by a successful `ad get text` run at the current head.

thymikee:
All four items in, and I am endorsing the reviewer's shape explicitly: **the carrier rule is decided once, in `formatCliOutput`, from a declared descriptor trait — the per-formatter carve-outs are gone, including the unreachable branch and its `input: {}` test.** New head: **`05be687e76`** (base `apex/2682-target-activation-disclosure` @ `65e02e3c70`, which is on `main` @ `fc59620932`). Two commits: `fix(#2682): let the dispatcher decide where a warning lands`, `refactor(#2682): one activation proof recorder, one home for each fact`.

You were right on the diagnosis: `get`'s format is `requiredField(enumField(['text','attrs']))` (`interaction/metadata.ts:171`) and the CLI reader always produces it, so `getCliOutput`'s "neither text nor attrs" branch was unreachable from every surface, and `output.test.ts:353` was green against dead code.

**1. One rule, in the dispatcher, with one declared opt-out.**
- `formatCliOutput` (`src/commands/cli-output.ts:27`) renders each command's formatter output and then appends the response's warnings through `routeResponseWarnings` (`src/commands/output-common.ts:73`).
- The opt-out is `parseableOutput?: true` on the command descriptor (`src/commands/command-contract.ts:58`), declared through the same field-command path as every other trait, and derived into one set at module load (`cli-output.ts:14`). It is exactly `get`, `find`, `clipboard`, `record` — the commands whose default stdout *is* the value or artifact path a caller pipes.
- **Stream decision, stated:** a warning goes **on stdout after the command's own line** for every command, and **on stderr** for a declared `parseableOutput` command, so `value=$(agent-device get text …)` still captures exactly the value. `--json` writes neither line and keeps `data.warnings`. MCP has one stream, so `toolTextWithWarnings` (`src/mcp/command-tools.ts:522`) merges the stderr half back into the model-facing text — scoped to the parseable commands, so no formatter's pre-existing stderr note moves into MCP output that wasn't there before.
- Removed: `withResponseWarnings`, `messageWithWarningsText`, `messageWithWarningsOutput`, the `is`/`find` wrappers, the `fill`/`longpress`/`hover`/`scroll` copies, `wait`/`settings`/`replay`'s warning formatter (back to `messageOutput`), and the settle wrapper's warning render (it keeps only the settled diff, so it can no longer say it twice). `getCliOutput` is now total over `text | attrs` (`interaction/output.ts:18`) — no third branch, no dead test.
- **Enumeration, not prose:** `cli-output.test.ts:70` pins the declared set to a literal; `:81` asserts no formatter-less command may declare the trait; and the table above it asserts the rule for **all 47 formatter commands**, each against the payload shape that command really returns (`apps`, `devices`, `session`, `capabilities`, `batch`, `debug`, `artifacts` have fixtures, not a synthetic `{message}`), so a formatter cannot pass by ignoring a payload it would never see. `apps`/`devices` are marked in the table as having no warnings channel on a bare-array payload — named rather than skipped.
- **Red-without / green-with, measured at this head:** with the dispatcher's routing call removed → **48 of 57** tests in `cli-output.test.ts` fail; with the routing kept but the declared trait removed → **8** fail (the set-equality test, all four parseable commands, and the three reachable-input `get`/`find` cases). Green with both in: 57/57.
- **Reachable-input `get text`:** `cli-output.test.ts:127` builds `{ format: 'text', target: { ref: '@e1' } }` — the input the reader actually produces — and asserts `text === 'General'`, `stderr === 'Warning: …\n'` exactly once; `format attrs` asserts `JSON.parse` succeeds and exactly one `Warning:` on stderr; a third case asserts a repaired-but-silent read stays silent on both streams.
- **The hole your rule exposed, closed:** 11 commands had no formatter at all (`type`, `focus`, `swipe`, `gesture`, `screenshot`, `trace`, `diff`, `push`, `react-native`, `trigger-app-event`, `test`) and printed nothing whatever the response carried — four of them are capture-consuming interactions. The generic CLI path now routes through the same helper (`src/cli/commands/generic.ts:80`), and `src/cli/commands/__tests__/generic.test.ts` asserts a `type` response's warning reaches stdout.

**2. One `recordActivationProof`.** `src/daemon/capture-disclosure.ts:33` owns the first-wins rule and returns the capture, so the three hand-copied guards in `selector-capture-runtime.ts`, `wait-current-surface.ts` and `interaction/internal/interaction-runtime.ts` collapsed to one call each. Existing behaviour is pinned by the tests that were already red-without it (`find-target-activation-disclosure.test.ts`, `wait-current-surface.test.ts`, `capture-disclosure.test.ts`).

**3. Contracts re-export block gone.** `@agent-device/contracts/ios-target-activation` now exports only `iosTargetActivationDisclosure` — what the fact says to an agent. The declarations have one spelling each: `packages/platform-apple/src/runner/target-activation.ts`, `runner/snapshot-presentation.ts`, `runner/__tests__/target-activation.test.ts`, `contracts/src/interactor-types.ts` and the live lane import `@agent-device/kernel/snapshot`. The reason-predicate test moved to `packages/kernel/src/snapshot-target-activation.test.ts` with the declarations and gained its compile-time half: `// @ts-expect-error` on `runningForeground` as an `IosTargetActivationPriorState`, so "a disclosure may not name the repair's outcome" is now enforced by the compiler as well as the value assertion.

**4. Docs.** `website/docs/docs/commands.md` gains a bullet naming the four `parseableOutput` commands and the stderr rule (plus "`--json` keeps them in `data.warnings` and writes neither line"), and the "Silence is not proof" bullet now names a `wait <text>` answered by its text observation as a command whose stamp the daemon does not decode, while noting that a timed-out `wait <text>` does disclose via its surface capture. `check:command-docs` green (12 tests).

**Simulator captures at `05be687e76`** (`apex-2682-proto`, iOS 26.2, session `apex2682-pr4`, `xcrun simctl openurl C2748D97-B92D-4B3B-9C65-AF86494AD904 https://example.com` + 8 s before each command; stdout and stderr captured separately because they are now different streams):

```
$ agent-device get text 'label="INFO"'      # exit 0
STDOUT  INFO
STDERR  Warning: The session app was not foreground when this command arrived (prior state runningBackground), so the runner activated it before answering (reason bundle_changed). Any capture taken earlier in this session described the only app other than the session app with an active accessibility session (pid 33878), not the session app. Re-capture now that the session app answers, or drive the other app in its own session.

$ agent-device find 'label="INFO"' get text   # exit 0, fresh handoff
STDOUT  INFO
STDERR  Warning: The session app was not foreground when this command arrived (prior state runningBackground), so the runner activated it before answering (reason stale_target). Any capture taken earlier in this session described the only app other than the session app with an active accessibility session (pid 33878), not the session app. Re-capture now that the session app answers, or drive the other app in its own session.
```

Both files kept under `/tmp/apex2682-state/` (`get-stdout.txt`, `get-stderr.txt`, `find-stdout.txt`, `find-stderr.txt`). The two agree: same value on stdout, same single sentence on stderr. Two more, to show the other half of the rule:

```
$ VALUE=$(agent-device get text 'label="INFO"' 2>/dev/null); echo "VALUE=[$VALUE]"
VALUE=[INFO]

$ agent-device is visible 'label="INFO"' 2>&1      # not parseable → stdout, after its own line
Passed: is visible
Warning: The session app was not foreground when this command arrived (prior state runningBackground), so the runner activated it before answering (reason stale_target). Any capture taken earlier …
```

**Gates at this head:** `format`/`format:check` no-diff, `check:quick` (lint + full `typecheck`), `pnpm build`, `check:layering` 246 pass / 0 fail, `check:production-exports` exit 0, `check:xctest-selection` 0 reachable by no lane, `check:command-docs`, `fallow audit --base apex/2682-target-activation-disclosure` clean over 58 changed files, targeted vitest `src/commands`+`src/mcp`+`src/cli` 135 files / 1310 tests, and `pnpm test:unit` **10746 passed | 1 skipped** across 1362 files. `node:test` registration lane 14 passed / 1 skipped. No Swift touched; physical `thymikee-iphone` untouched; no merge.

Still open, unchanged from my last comment: a *satisfied* `wait <text>` carries no activation fact (named on #2694 and now in the docs), `find`'s sparse-first capture rests on its two route tests because a healthy simulator won't produce a sparse tree, and #2696 keeps the live lane manual.

thymikee:
This is ready. The fix from 7977596 (https://github.com/callstack/agent-device/pull/2693#issuecomment-5750123832) is in at 05be687, and I have no blocking findings left.

Not blocking: snapshot-runtime.ts:36 still writes the first-wins activation state by hand instead of calling recordActivationProof, so recordActivationProof isn't yet the single owner of the rule (https://github.com/callstack/agent-device/blob/05be687/src/daemon/snapshot-runtime.ts#L36); `record`'s parseableOutput trait doesn't stop recordCliOutput from also pushing its own warning lines onto stdout, so the dispatcher's stderr dedupe may leave a duplicate-looking line on a real warning (https://github.com/callstack/agent-device/blob/05be687/src/commands/recording/output.ts#L20); and `find ... click` prints its success message on stdout while its disclosure goes to stderr, unlike a direct `click`, worth a line in commands.md if that's expected — but take or leave any of these.

I did not run tests or device commands; the red/green counts and simulator captures in the PR are the author's, not reproduced here. Smoke Tests is still in progress at 05be687 with 18 green and 1 skipped so far, and its route overlaps this diff (the smoke-ios-target-activation import rewrite, the capture-disclosure code, and the CLI output dispatcher), so land once it goes green, checking any failure against these paths first.

PR #2702, opened 2026-09-19
feat(action-button): press the iPhone Action Button

### LINKED ISSUE #2699: feat(action-button): press the iPhone Action Button
## Purpose

Every other device-level system input is drivable (`home`, `back`, `app-switcher`, `orientation`, `tv-remote`), but the iPhone Action Button is not, so system Shortcut/App Intent delivery — including while the app is backgrounded or terminated — cannot be exercised. Requested in #2655.

## API

```sh
agent-device action-button
```

```ts
await client.command.actionButton();
```

No positional and no flags. Two research findings set this shape:

- `XCUIDevice.press(_:)` (`XCUIAutomation`, Xcode 16.3+) takes **no** `forDuration:` overload, unlike the `XCUIRemote.press(_:forDuration:)` that `tv-remote` uses. A `press|long-press` verb and `--duration-ms` therefore cannot be honoured for `.action`, and would ship a flag that is silently ignored.
- The Action Button has no slide gesture. Apple's Action Button guide documents exactly one input: "press and hold the Action button". The capacitive surface you slide a finger across is the **Camera Control**, a separate button exposed as `XCUIDevice.Button.camera`, and XCUITest offers no slide/light-press API for either.

So this is shaped like `home` (device-level, no arguments), not like `tv-remote`.

## Required behavior

- Public `action-button` command descriptor: `frameworkTier: 'extended'`, `daemon.route: 'generic'`, `refFrameEffect: 'may-invalidate'`, the generic mutating traits, and `platformExecution: { kind: 'device-runtime', uses: [actionButtonRuntimeUse] }`.
- New `actionButton` runtime operation on the ADR 0019 fact path: a `packages/contracts/src/action-button-runtime.ts` owner module (intent, operations, facts, binder), a `RUNTIME_OPERATION_NAMES` entry, an `INTERACTOR_OPERATIONS` row, an owner-stated `actionButton` cell in `UnavailablePlatformRuntimeFacts`, and an optional `actionButton?(): Promise<void>` `Interactor` member — optional because only Apple can press it.
- No `--settle` and no post-action observation. The Action Button is expected to deliver to a backgrounded or terminated app, so the press must preserve the app's starting state: the runner must not `activate()` the app first, and the command must not re-capture the previously-foreground app. This is what keeps it on `home`'s trait set rather than `press`'s `settle-and-verify`.
- Apple fact: available on iOS/iPadOS simulator and physical-device leaves; tvOS, macOS and watchOS refuse at the fact level. Within an admitted leaf, a device without the button refuses on `XCUIDevice.hasHardwareButton(.action)` reported as a typed reason — never error-text matching.
- Every other platform owner states its own refusal cell (Android, HarmonyOS, web, Vega, Linux, WebDriver, Limrun). The completeness gates make an unstated cell a compile error rather than an assumed default.

## Completion conditions

- `agent-device action-button` presses the Action Button on an iPhone simulator and returns `{ action: 'action-button', message: … }`.
- On a device whose model has no Action Button, it fails with a typed unsupported reason and a hint naming the hardware requirement — not a generic transport error.
- `agent-device action-button` on Android, web, tvOS, macOS, Vega, Linux and HarmonyOS is refused by fact admission before any runner request is issued.
- A press issued while the app is backgrounded does not foreground it, and the response carries no settle or observation payload.
- Node surface `client.command.actionButton()` and the MCP tool are present, with the MCP output schema derived from `CommandResultMap`.
- `pnpm check:affected --run` and the registry, owner-files, and generic-route completeness gates pass.

## Dependencies

None. The Camera Control (`XCUIDevice.Button.camera`) is deliberately out of scope: XCUITest can only click it, and a press-only `camera-control` command that cannot slide is a separate decision.

thymikee:
## Summary

Adds `action-button`: one press of the iPhone/iPad Action Button, shaped like `home` (no positional, no flags) rather than `tv-remote`.

```bash
agent-device action-button
```

```ts
await client.command.actionButton();
```

XCUITest exposes `XCUIDevice.press(.action)` with no hold-duration overload, so `long-press`/`--duration-ms` are not expressible, and the slide surface on the same edge is Camera Control, which this command does not drive. The press is dispatched without the runner's app-activation preflight and with no post-action observation, so the session app keeps the state the press found, and the recorded-failure conversion still applies to it. The runner asks the device for the button with `hasHardwareButton(.action)` and refuses with typed `UNSUPPORTED_OPERATION`; every other platform states its own refusal cell. 76 files, +649/−57.

What the system does with the press is **not** claimed: Simulators run no Shortcuts or App Intents, so delivery to an assigned Shortcut is verifiable only on a physical iPhone, and help text and docs say so.

Closes #2699

## Validation

Tested at `afc4c28` (rebased onto `d688926` after the foldable-capture and Limrun recorder lands; `CHANGELOG.md` was the only conflict).

- Local deterministic gate green on this head: `pnpm check:tooling` and `pnpm check:unit` (unit + tmpdir + smoke) both exit 0; `pnpm check:affected --run` passes every selected gate except `replay-compat` and `daemon-wire-compat`, which refuse to run here because this clone is shallow (`… needs full history and tags`), and are GitHub-owned history-backed checks.
- The two CI failures from the previous head are fixed and verified locally: `eager-closure-budgets` (672 tests) and `test-file-size-ratchet` now pass. The press declares itself through the existing zero-argument interactor group instead of a new contracts module and export subpath, and the runner-protocol fixture test checks itself against the trait manifest instead of a hand-maintained list.
- Swift: `pnpm build:xcuitest:ios` and `:macos` build; `check:xctest-selection` reports 0 methods reachable by no lane. Two new XCTest decisions pin the activation skip and the recorded-failure conversion.
- Live iOS Simulator evidence re-collected at the pre-rebase head `3f28086` (Swift sources unchanged since): iPhone 17 Pro returns `Pressed Action Button` and its `runner.log` shows `ACTIVATE … reason=bundle_changed` for the neighbouring `home` command and no activation line for `actionButton`; iPhone SE (3rd generation) refuses with `UNSUPPORTED_OPERATION`, repeatably, without invalidating the session.
- Unresolved risk: no physical iPhone available here, so Shortcut/App Intent delivery is unverified. That is now stated in the command description, `commands.md`, and the CHANGELOG instead of asserted.


thymikee:
Reviewed at 7f36e4d. This is findings, not a merge-ready change.

The lifecycle exemption looks wrong. `.actionButton` joins the `isLifecycle: true` group to skip the activation preflight, but `isLifecycle` also gates the recorded-failure conversion: `xctestRecordedFailureResponse` (https://github.com/callstack/agent-device/blob/7f36e4d/apple/runner/AgentDeviceRunner/AgentDeviceRunnerUITests/RunnerTests%2BCommandExecution.swift#L2517) returns nil for lifecycle commands, so the generic guard never turns a recorded XCTest issue into `XCTEST_RECORDED_FAILURE` for this command. If XCTest records a failure while `XCUIDevice.shared.press(.action)` runs, the command still returns `ok: true`, and since this command has no settle and no post-action observation, that conversion was the only evidence the press happened. `home`, `app-switcher`, and `alert` all keep the conversion, so this looks like an accidental regression rather than an intended change. Can `.actionButton` stay `isLifecycle: false` and get the activation skip through the existing `shouldSkipAppActivationPreflight(_:)` (https://github.com/callstack/agent-device/blob/7f36e4d/apple/runner/AgentDeviceRunnerUITests/RunnerTests%2BCommandExecution.swift#L2550), which already exempts `.alert` for the same reason, and get a case pinned in the existing Swift tests around line 579-782?

The Shortcut/App Intent claim in the command description (https://github.com/callstack/agent-device/blob/7f36e4d/src/commands/system/index.ts#L56) isn't backed by evidence simulators can produce. Simulators don't run Shortcuts or App Intents, so the only property this PR can show on the admitted path is that XCUITest accepted `press(.action)` and no activation was logged; the behavior the command exists for — a Shortcut or App Intent firing while the app is backgrounded or terminated — is untested. A new public command is shipping a stated effect that's unproven, while its help text asserts that effect to agents. Can this be validated on a physical iPhone 15 Pro or later with a Shortcut assigned to the Action Button: background the session app, run `agent-device action-button`, and show the command output plus evidence the Shortcut ran while the app stayed backgrounded? If no such device is available, the description and docs/commands.md should say only what's proven: the press reaches the system, and Shortcut execution is unverified and doesn't happen on simulators.

Not blocking: the availability check in RunnerTests+Interaction.swift:136 guards the hardware query but not the press itself, so on iOS 15 (deployment target 15.6) the press could fire unconditionally on hardware without an Action Button — `guard #available(iOS 16.0, *) else { return false }` would close that; the simulator caveat about Shortcuts not firing should also land in commands.md, not just the code; the new action-button-runtime.ts files duplicate app-switcher-runtime.ts nearly verbatim and are what the Coverage gate flags across four entry closures, so declaring the zero-argument press through the existing `bindNoArgumentInteractorOperation`-style path in interactor-operation-binding.ts would avoid growing those closures; the transport test asserting no `activate` runner command doesn't actually pin the preflight skip (it would pass even without the `isLifecycle` change) and a Swift unit test next to `shouldSkipAppActivationPreflight` would be the decidable place for that; the fixture at provider-device-runtime.fixtures.ts:272 re-derives the iOS/iPadOS admission rule instead of reading the existing fact; and CHANGELOG.md has no Unreleased entry for this new command. These can be taken or left.

Is the surface area proportional here? This adds a contracts module, a new package subpath, a required Interactor member across nine owners, a daemon route module, and a new trait meaning in the runner, and pushes npm unpacked size up 6.0 kB against the 3 kB threshold that asks for a smaller owning interface. A `bindNoArgumentInteractorOperation(name)`-style factory in interactor-operation-binding.ts, one `resolveBoundNoArgRuntime(...)` on the daemon side for home/app-switcher/action-button, and reuse of `shouldSkipAppActivationPreflight` in the runner would cover the same feature as one row in each table instead of three new files, and would also remove the lifecycle-exemption issue above.

I could not run a physical device, so the Shortcut/App Intent gap above is a documentation and design gap, not something I reproduced with a failing run. Coverage's four eager-closure failures and the test-file-size ratchet failure on runner-client.test.ts (1440 vs 1438 lines at the merge base) both point at exactly what this diff adds, so that CI failure looks related to this change. The Smoke Tests failure is in snapshot capture-plan occupancy handling, a path this PR's Swift changes don't touch, so that one looks unrelated. Closing the lifecycle exemption and availability-fallback gaps in the runner, resolving the Shortcut-delivery claim one way or the other, and moving the new contracts declaration into a module the existing entry closures already evaluate would get the code and the checks aligned.

thymikee:
All findings addressed at `3f28086`. No physical iPhone was available to me, so the second finding is resolved as a documentation and design correction rather than with device evidence.

**Lifecycle exemption (#1, blocking)** — fixed in 19af3bf. `.actionButton` is `isLifecycle: false` and skips activation through `shouldSkipAppActivationPreflight`, next to the `.alert` exemption, so the recorded-failure conversion stays live for it. Pinned where each decision is made: `testActionButtonPressSkipsAppActivationPreflightWithoutBeingRunnerLifecycle` (asserts the skip and `isRunnerLifecycleCommand(.actionButton) == false`) and `testXCTestRecordedFailureResponseFailsActionButtonSuccess` in the host-lane group beside the other conversion tests. Live re-check at this head: the same session logs `ACTIVATE … reason=bundle_changed` for `home` and no activation line for `actionButton`.

**Shortcut/App Intent claim (#2, blocking)** — no iPhone 15 Pro or later is reachable from here, so the claim is gone rather than proven. 47a15f5 rewrites the command description, the `commands.md` bullets, and the CHANGELOG entry to what a Simulator shows: the press is dispatched, the session app is not brought forward, and what the system triggers from it needs a physical iPhone to verify.

Non-blocking, taken:

- Availability guard closes the press on pre-iOS 16 instead of falling through to it.
- The simulator caveat lives in `commands.md`, not only in code.
- The duplicated module is gone: the press declares itself as the first member of `NoArgumentInteractorOperations`, bound by `bindNoArgumentInteractorOperations`, and owners state a plain `actionButton:` cell the way they state `ensureReady`. `packages/contracts/src/action-button-runtime.ts` and its export subpath are deleted, the contracts export snapshot is back at 119 specifiers, and the four eager-closure growths and the `runner-client.test.ts` size tripwire are green locally. One deviation from the suggested shape: the binder sits in `interactor-operation-catalog.ts`, not `interactor-operation-binding.ts` — a type-only import of `platform-runtime-operations` from the binding module pulled eight contracts files into R10's largest type cycle against a baseline of zero, so layering forced the leaf to stay a leaf.
- The Swift test is now the place that pins the preflight skip; the provider transport test stays because it proves a different claim (the press rides a provider's transport).
- The provider fixture reads the new kernel `hasAppleActionButton`, the same rule the Apple fact reads, instead of restating the iOS/iPadOS compare.
- `CHANGELOG.md` has an Unreleased entry.

Not taken, with reason: the daemon-side `resolveBoundNoArgRuntime(...)` shared by `home`/`app-switcher`/`action-button`. `src/daemon/__tests__/runtime-binding-conformance.ts` is keyed per command over per-command `<command>-runtime.ts` resolvers, and merging three command families here crosses the one-family boundary this PR should hold. Happy to take it as a follow-up on its own — the copy the review pointed at (module, subpath, facts helper) is already gone.

CI: the Coverage job's two failures are the ones above and both pass locally (`eager-closure-budgets` 672 tests, `test-file-size-ratchet`). The unrelated Smoke Tests failure you saw is in snapshot capture-plan occupancy, consistent with your read that it is not this diff.

thymikee:
Re-triggering CI: GitHub did not create workflow runs for the synchronize event on 3f28086 (only CodeQL ran). No code change; reopening immediately.

thymikee:
Reviewed at 3f28086. The code is clean. Your re-push landed afc4c28 while I was writing this: the range-diff shows the same seven commits with only the CHANGELOG.md hunk moved, so this reads on the current head too.

Since the earlier review at 7f36e4d, the raised issues are resolved in this delta. Not blocking: `hasAppleActionButton` in packages/kernel/src/device.ts:183 still answers true for a platform like Android because it only checks `resolveDeviceAppleOs`, not `isApplePlatform`, and the no-argument interactor binder in packages/contracts/src/interactor-operation-catalog.ts:31 is typed for the whole `NoArgumentInteractorOperations` group while only registered and gated for `actionButton`, so neither breaks anything today but both are worth a look whenever you're in that code.

The CHANGELOG.md conflict that was open at 3f28086 is gone, since afc4c28 rebased onto main and resolved exactly that hunk.

CI is still pending: only CodeQL ran on 3f28086, so the gates that matter here never reported. The delta removes a contracts export subpath and rewrites two test files, which is exactly what `eager-closure-budgets` and `test-file-size-ratchet` check, the same two gates that failed at 7f36e4d. The re-push on afc4c28 has those jobs running now, so their result is what to watch.

I could not inspect the live evidence you describe for 3f28086 (iPhone 17 Pro pressing the Action Button, iPhone SE returning UNSUPPORTED_OPERATION) since no log artifact is attached and I have no device to reproduce it on, but the claim names the route and the output the delta changed, so I'm taking it as stated. I did not run any gate myself, so the `pnpm check:*` results are your local report, and I did not re-review code outside this delta beyond confirming it left the registry descriptor and CLI/Node/MCP declarations consistent, since those were already covered at 7f36e4d.

Nothing is blocking on the code side, so this is ready for you. Watch the gates now running on afc4c28, and take the `hasAppleActionButton` note whenever you are next in that file.

PR #2704, opened 2026-09-20
fix(test-app): launch the lab on iOS 27.1 by adopting the scene lifecycle

thymikee:
## Summary

iOS 27.1 will not launch a legacy `UIApplicationDelegate` app at all: it traps at launch inside `___UIApplicationEvaluateRuntimeIssueForNoSceneLifecycleAdoption`, so `examples/test-app` crashed immediately on iPhone Duo and could not be used to verify anything on that runtime.

Expo SDK 58's template ships a `SceneDelegate` and the matching `UIApplicationSceneManifest`, which is the scene adoption iOS 27.1 requires. This moves the lab from Expo SDK 56 / RN 0.85 to Expo 58 preview / RN 0.88.

Draft on purpose: **this is a preview SDK** (`expo@58.0.0-preview.3`, `react-native@0.88.0-rc.0`) and only the iOS simulator path is validated here. Take it only if CI on the SDK preview is acceptable, or hold it until SDK 58 ships stable.

Two call sites needed the nullability RN 0.88 introduces: `AppState.currentState` is now a plain nullable string (narrowed instead of cast) and `useColorScheme()` can return `null`. The `expo-modules-jsi: 56.0.13` override was dropped because `expo-modules-core@58` needs `~58.0.2`; keeping it produced `no type named 'JavaScriptDecodable' in module 'ExpoModulesJSI'`.

## Validation

Tested at `e0377fba4`, and re-run at each head since — current head `ae8fad731` on `main` at `921a2cb3a`: `pnpm check:affected --run` passed every time, `examples/test-app` `tsc --noEmit` clean, and its `security:test` suite passes.

Two reds this branch carried, both resolved or explained at the last head:

- **Android smoke** (`wait text Catalog` on the launcher) was named by the crash buffer #2710 added: SDK 58's `NativeTabs` renders `md` icon names on Android through `expo-symbols` and throws inside the navigator when it is missing, which kills the JS runtime and leaves the emulator on the launcher. `ae8fad731` installs the package the error names, and the Android lane is green on this head (26m53s). It was never Release-specific; the pre-#2710 artifact simply carried no app-side log.
- **iOS smoke** failed once on an earlier head while waiting for `automation-longpress`, which is #2714: `scroll` stops moving the inset `ScrollView` on `/automation` once its unclamped gesture endpoint lands above the scrollable's top edge, and keeps answering with the requested distance. That lane is green here, so #2714 is intermittent in CI rather than resolved, but it reproduces on demand locally against `main`'s CLI with only this fixture under it; `/catalog` is the working control.

Live on a booted iPhone Duo (iOS 27.1): the app installs, launches without a crash report, and `snapshot -i` returns the 41-node home surface; earlier in the same build a tap dismissed the dev-menu sheet and a Catalog tap settled `+19 -15`.

Android smoke and the Release builds for both platforms now run green in CI on this head, which retires the earlier "unvalidated: Android, Release iOS builds on the preview SDK" note. The Maestro replay suites still run only in the device lanes.


thymikee:
Reviewed at e0377fb.

The Android Release APK this PR produces does not survive normal navigation. Smoke Tests builds examples/test-app into `app-release.apk` via `test-app-build-cache.yml`, keyed on this PR's own Expo fingerprint, and drives that exact binary. The E2E taps `automation-continue-catalog` on [AutomationLabScreen.tsx](https://github.com/callstack/agent-device/blob/e0377fb/examples/test-app/src/screens/AutomationLabScreen.tsx), a file this PR edits, and the app then drops to the Android home launcher instead of reaching Catalog. That points at the RN 0.85→0.88.0-rc.0 / Expo 56→58.0.0-preview.3 bump in [examples/test-app/package.json:23](https://github.com/callstack/agent-device/blob/e0377fb/examples/test-app/package.json#L23), alongside the react-native-gesture-handler 2.x→3.x, react-native-screens 4.25→4.27, and webview 13→14 jumps. The PR body itself flags Android and Maestro/e2e as unvalidated, so this failure isn't a surprise, but it does mean the bump hasn't been proven safe on Android yet.

Is there a smaller path here? The iOS 27.1 fix only needs a SceneDelegate and a UIApplicationSceneManifest in the generated iOS project. Could a config plugin inject those into the current Expo 56/RN 0.85 tree instead of taking the full SDK bump, so Android stays on its already-validated versions? If Expo has no supported way to backport a scene manifest to SDK 56, then the full bump is probably the only path and the current approach is reasonable, but that would need to be confirmed first.

This can't merge until a live Android run on this PR's own Release APK reaches the Catalog screen after tapping `automation-continue-catalog`, or until Android logcat/crash output pins down the specific incompatible dependency so it can be fixed or downgraded. Marking this as a draft is the right call for now.

thymikee:
### Launch evidence on iOS 27.1

Same app source, same booted iPhone Duo (`iPhone19,4`, iOS 27.1, Xcode 27.1 Beta), closed pose.

**Before this PR** the app never renders: launch traps in `___UIApplicationEvaluateRuntimeIssueForNoSceneLifecycleAdoption`, because the SDK 56 generated project has an `AppDelegate` but no `SceneDelegate` and no `UIApplicationSceneManifest`. iOS 27.1 enforces UIScene adoption.

**After this PR** the app boots and draws on the Duo's lit outer panel. `agent-device snapshot -i` returns 41 nodes, `tap @e189` at `(345, 301)` dismisses the dev-menu overlay, and tapping Catalog at `(128, 626)` settles with `+19 -15` nodes.

RN app running on iOS 27.1 on the Duo lit panel

Agreeing with the review: the Android Release APK is the blocker. Reproduced locally as the Smoke Tests failure at `wait for Catalog` after tapping `automation-continue-catalog`, which is a real regression from the SDK-preview bump and not something to merge through.

thymikee:
Tested the smaller path you asked about, and it does not exist inside supported Expo. Numbers below.

**SDK 57 stable does not adopt the scene lifecycle.** I downgraded this branch to `expo@57.0.24` / `react-native@0.86.3` (which is also `screens ~4.26.0` and `webview 13.16.1`, i.e. main's Android surface kept), ran a clean `expo prebuild --platform ios --no-install`, and inspected the output:

- `UIApplicationSceneManifest` in the generated `Info.plist`: **0 occurrences**
- `SceneDelegate.swift` in the app target: **absent**
- `AppDelegate.swift` is `class AppDelegate: ExpoAppDelegate` with **0** `configurationForConnecting` / `UISceneConfiguration` references

The 57 pod does contain `ExpoAppSceneDelegate.swift` and `SceneEventForwarder.swift`, but nothing wires them into the app target, so 57 hits the same `___UIApplicationEvaluateRuntimeIssueForNoSceneLifecycleAdoption` trap as 56. Only the 58 preview's prebuild emits the manifest and the delegate. (My first look at this appeared to say 57 was enough — that was a stale `ios/` directory from the 58 prebuild. Re-running `--clean`-style is what produced the numbers above.)

**A config plugin is a bigger risk, not a smaller one.** RN core implements no scene API in any of these versions — grepping `Libraries/AppDelegate` in `react-native@0.85.0`, `0.86.0`, `0.87.0`, and `0.88.0-rc.0` gives **0** occurrences of `UISceneConfiguration`, `configurationForConnecting`, or `windowScene`. So injecting a manifest into a 56/57 tree declares a scene delegate that nothing implements; you would hand-write the window and root-view bootstrap that the 58 template generates, i.e. fork the template and own that code. That leaves the app's launch path custom, which is worse for a sample app whose whole job is exercising the tool against a stock stack.

**The Android majors are not my choices.** `bundledNativeModules.json` shipped with `expo@58.0.0-preview.3` pins `react-native 0.88.0-rc.0`, `react-native-gesture-handler ~3.2.1`, `react-native-screens ~4.27.0`, `react-native-webview 14.0.1`, `react-native-safe-area-context ~5.9.1`. Overriding those back to main's versions fights the SDK's own mapping and is exactly the kind of half-migrated tree that produces unverifiable failures.

Which means your call on how to spend effort here. I see two real options:

1. **Bisect the Android regression.** My read of the evidence points at `react-native-gesture-handler` 2.x→3.x: the drop to the launcher happens on the tap of `automation-continue-catalog`, which is gesture-handler-owned routing through `expo-router`, and the `@expo/dom-webview` bump is the other 3.x-adjacent suspect. I have no local AVD, so this iteration happens on CI; if you want it, I will split the majors into separate commits so each CI run tests exactly one variable.
2. **Hold until 58 leaves preview.** The only preview-pinned component is what Expo itself pins — including an `rc` React Native — so waiting gets the same scene adoption with Android dependencies that have had a stable validation cycle. In the meantime the iOS 27.1 verification I did here stands as evidence, and I would rather leave this PR as a draft documenting the requirement than land a bump that breaks the Android smoke.

Say which you prefer and I will take it from there. If it is (1), I would also add a narrow CI job that runs only the Android release navigation, because the full smoke is a slow oracle for a one-variable question.

thymikee:
That settles the smaller path, thanks. Checking the 57.0.24 prebuild output directly, and grepping RN core for the scene API across 0.85 through 0.88, is the proof I was missing: if no released Expo version wires a SceneDelegate into the app target, a plugin would have to hand-write the window and root-view bootstrap that the 58 template emits, and this sample app is the worst place to own a forked launch path. I am dropping that question. The full bump is the right shape.

The Android Release APK is still the only blocker at e0377fb. Before you split the majors into separate commits, note that the failing Smoke Tests job already uploads `android-artifacts` on every run, including the session state directory with `app.log`. If the drop to the launcher is a native crash, that log names the library in one run and you can skip the bisect entirely. Worth reading before spending several CI rounds on one-variable runs.

If the log is not conclusive, I would take your option 1 with the narrow navigation-only job. The answer outlives this PR: if gesture-handler 3.x breaks release-build navigation under expo-router, users will hit it too, and that belongs upstream rather than parked behind an SDK wait. Holding for 58 stable is also defensible if you would rather not spend the CI rounds now. The iOS 27.1 evidence you posted stands either way, and the draft documents the requirement. Your call.

thymikee:
Two updates on the red checks, both from the artifacts rather than from guessing.

**The iOS failure is not this change.** Job `106068926059` at `95d07b67b` died 4m7s in, inside package verification, before any simulator interaction: `AppError: xcrun timed out after 30000ms` for `xcrun --sdk iphonesimulator --show-sdk-version`, which skipped its own retry because the shared snapshot-source deadline was already spent. Filed as #2712 with the deadline math. Nothing here touches Apple toolchain probes.

**The Android failure cannot be diagnosed from what CI uploads, and that is now fixable.** You were right that a Release-only drop to the launcher is the interesting kind: `failed-step.txt` is the `wait text Catalog` timeout, and `failed-step-11-snapshot.json` labels are the launcher itself (`At a glance`, `Play Store`, `Home`, `Phone`, alongside the fixture icon) — the app really is gone. But the artifact carries no `app.log`, and `daemon.log` is two lines, so the crash reason is unrecoverable from that run. #2710 adds exactly the three missing probes to failed-step evidence: `pidof` for the fixture app, the resumed activity, and the crash buffer, so a dead pid next to a launcher `mResumedActivity` names the library in one run instead of a bisect across majors.

Suggestion: land #2710, then re-run here. If the crash buffer names a library, the SDK-56-vs-57 bisect you were weighing becomes a single line of evidence. I have no local Android emulator on this host, so I cannot pre-empt that round.

thymikee:
One factual correction to the diagnosis, because the two readings imply different fixes. Both failures are the **app-drops-to-launcher** signature, not an emulator boot problem. From the current Android job on this head (`106089269585`, same shape on #2704's `106089267544`):

```
smoke:fixture-bootstrap: 9978ms
command: agent-device wait text Catalog 10000 --platform android --serial emulator-5554 --session android-e2e-48t --json
  screenshot: …/failed-step-11.png
  snapshot:   …/failed-step-11-snapshot.json
  device:     …/failed-step-11-device.txt
```

The emulator booted, the fixture installed, `smoke:fixture-bootstrap` completed in ~10s, and the run got all the way to step 11 — so the adb-offline-boot-loop reading doesn't match this job. Whatever happens between tapping `automation-continue-catalog` and the Catalog screen is the question, and the artifact still cannot answer it: no `app.log`, and `daemon.log` is two lines. That is precisely the gap #2710 closes, so one Android run on a head that carries it should name the library.

thymikee:
Rebased onto `main` at `921a2cb3a`; head is `e2404e6ee`, and `check:affected --run` passed on it. Marking this ready.

The reason it was held as a draft is resolved as of an hour ago: #2710 merged, so this head now carries the failed-step probes that were missing when the Android failure was undiagnosable. The next Android smoke run on this branch writes `pidof`, the resumed activity, and the crash buffer into `failed-step-11-device.txt` — so the run that reproduces the launcher drop should also name the library that dropped the process, instead of costing a bisect across Expo majors to find out.

The Android check will likely go red on this head again for that pre-existing reason. If it does, the artifact should now say why, and that answer decides the question the review kept open: whether the Release-only break is in the fixture app's startup or in something Expo 58 changed under it. Nothing else is open on this branch from the review rounds — the scene-lifecycle shape was accepted, and the smaller-path alternative was measured and ruled out.

Stacked #2705 is rebased on top at `129896d07`; it inherits this branch's Android check and changes nothing that reaches that step.

thymikee:
Rebase checked at `e2404e6`, no new code review: `git range-diff` against `c2ecc46` shows the single commit byte-identical, so the code verdict from that review stands.

Both Smoke Tests jobs are red at this head, and the iOS one is a new failure, not the `xcrun` probe timeout you filed as #2712. Job `106112615304` fails 107s into the live iOS simulator E2E with `id="automation-longpress" did not become visible after scrolling` — an interaction assertion, well past the toolchain probes. This head changes `AutomationLabScreen.tsx` and `theme.ts`, which is the screen that assertion scrolls, so the overlap is direct. Does the long-press row still reach the scrollable area at the simulator's width after the Expo bump, or has the lab grown past what the scroll view can reach? #2705 caps the lab's content column at regular width on top of this branch, which makes that pair worth comparing once its own run lands.

The Android job `106112615210` is the same app-drops-to-launcher signature as before, and it now carries the #2710 probes — `failed-step-11-device.txt` on this run should finally name what dropped the process.

Next blocker: the iOS smoke failure needs attributing before this is ready for human review, and the Android drop still needs its cause.

thymikee:
The Android failure is named now, and #2710 is what named it — the crash buffer it added landed in `failed-step-11-device.txt` on the first run that carried it (run `35523907560`, job `106112615210`):

```
## app process
(failed: Command failed: adb -s emulator-5554 shell pidof com.callstack.agentdevicelab)

## resumed activity
topResumedActivity=ActivityRecord{… u0 com.google.android.apps.nexuslauncher/.NexusLauncherActivity t6}

## crash buffer
FATAL EXCEPTION: mqt_v_native
com.facebook.react.common.JavascriptException: Error: NativeTabs.Trigger.Icon `md`
icons on Android require 'expo-symbols'. Install it with `npx expo install expo-symbols`
or use the `src` or `dra…` prop
    at NativeTabsNavigatorWrapper (…/index.android.bundle:1:945807)
    at TabsLayout (…/index.android.bundle:1:876503)
```

So the process was gone before the wait (no `pidof` match, launcher resumed) and it died in the JS runtime while mounting the native tab bar: SDK 58's `NativeTabs` renders the `md="home"` style names on Android through `expo-symbols`, which the lab never depended on, and throws inside the navigator. That kills the app on the tabs layout, so `wait text Catalog` never sees its text. Nothing about Release builds or the signing path — it was never Release-specific, and it was invisible because the pre-#2710 artifact carried no app-side log at all.

Fixed in `ae8fad731` by installing the package the error names; the `md`/`sf` icon pair the tab bar already declares is unchanged. `tsc --noEmit` and `check:affected --run` pass on the head. #2705 picks it up at `75380723f`.

Two things this does *not* explain, both still open:

- The iOS `Smoke Tests` failure on this head is unrelated and it is a genuine product defect, not this diff: `scroll` stops moving the scrollable while reporting the requested distance. I can reproduce it locally on the `/automation` route and it is being filed separately with measurements.
- The `expo-symbols` requirement is undocumented in the tab-bar guide as far as I could tell, and the error only fires on Android, so any SDK 58 app using `md` icon names crashes on Android under native tabs while looking fine on iOS.

thymikee:
Green across the board on this head: 19/19 checks, Android smoke included (26m53s), iOS Release and Android Release both pass, and the merge state is CLEAN.

The two lanes that had been red for most of this PR's life close out differently:

- **Android** was a real fixture bug, newly visible only because #2710's crash buffer finally printed it: `NativeTabs.Trigger.Icon` `md` icons need `expo-symbols` on Android, and its absence threw inside `TabsLayout`. Fixed in `ae8fad731`.
- **iOS** was #2714, a CLI `scroll` defect that reproduces against `main` with only this fixture underneath it. That lane passed this time, so #2714 is intermittent in CI rather than gone; the deterministic local repro is on the issue.

thymikee:
Reviewed at `ae8fad7`. The new commit adds `expo-symbols ~58.0.1` to the fixture app and nothing else, and it is the right fix: every `NativeTabs.Trigger.Icon` in `examples/test-app/app/(tabs)/_layout.tsx` uses `md=` icons, `expo-router` already lists `expo-symbols` as an optional dependency, and the crash text #2710's buffer printed asks for exactly that package. The lockfile delta stays confined — only `expo-symbols` and its transitive `@expo-google-fonts/material-symbols` change version, everything else is a peer-key rehash. The code verdict is clean.

All 19 checks pass on this head, Android Smoke Tests and Android Release included. The Android crash text is read from the CI artifact quoted in this thread, not from a build run here. The iOS lane's earlier `automation-longpress` failure has no overlap with this diff, which touches only `package.json` and the lockfile, and your #2714 repro builds the CLI from `main` with only the fixture moving under it. That one belongs to the CLI scroll path, not to this PR.

No conflicts. One small thing that does not block: `examples/test-app/README.md:61` still explains the `@expo/dom-webview` pin as keeping the development runtime "on the SDK 56 native module", but this PR moves that pin to SDK 58, so the sentence now names the wrong major.

Next: nothing here blocks human review. #2714 stays open on its own.

PR #2705, opened 2026-09-20
feat(test-app): fit the lab to regular-width panels

thymikee:
## Why

Stacked on #2704. The lab had no width handling of any kind: prose ran the full window and the catalog kept one column the entire way down. On the iPhone Duo's inner panel that is a 951pt-wide page with no measure limit, and on iPad the catalog stretches the same way.

## What

`AppFrame` already wraps every route and owns frame geometry, so the capped, centered column lives there: one rule, and every framed route inherits it — including `ProductScreen`, `AccessorySetupScreen`, `AutomationLabScreen`, and the `inert` route, which a per-screen pass would leave full-bleed. The webview route renders `WebViewScreen` outside `AppFrame`, so its arbitrary-width web content keeps the full-bleed layout it renders today; capping a `WebView` is a separate call from capping the lab's own cards.

`useReadableColumn` reports the width a route's content actually gets and whether the readable cap bound. The centered column view renders only when it does, so at compact width the tree is the one every e2e, smoke, and replay run uses today: no container, no cell, no wrapper.

The catalog takes two columns once a card clears 340pt, with cell width computed from the column width rather than a hand-tuned percentage. Cards keep their own 12pt bottom margin and the grid owns only the column gutter, so the vertical rhythm matches the compact list instead of doubling it. Every `testID`, `accessibilityLabel`, and `accessibilityRole` is byte-identical, and the child order that `stickyHeaderIndices={[1]}` depends on is unchanged.

Navigation stays on the stock tab bar in `automatic` mode. I measured `sidebarAdaptable` rather than argue it: expo-router maps it to `tabBarControllerMode='tabSidebar'`, which on iPadOS gives the sidebar this lab does not want, and the two things it was supposed to buy are already there without it — the strip docks to the trailing edge at regular width, and the active tab reports `[selected]`.

## Evidence

In the last comment, on iOS 27.1 / Xcode 27.1:

- **iPhone 17 sim, compact `402x874`**: catalog unchanged, and `product-card-*`, `favorite-*`, `details-*`, `add-*` all resolve at the top and bottom of the list.
- **iPad Pro 13-inch, 1032pt**: two product columns inside the capped column, each card keeping its badge, price, `Save`, `In cart`, and `View details`; the tab items flow horizontally rather than becoming a sidebar.
- **Product detail at 1032pt**: a route this change previously left full-bleed, now inside the same column.

## Verification

- Head `75380723f`, rebased on `main` at `921a2cb3a` on top of #2704, which now also carries the `expo-symbols` fix that names and clears the inherited Android red. All 19 checks are green on this head, Android smoke included. `pnpm exec tsc --noEmit` in `examples/test-app`: clean. `pnpm check:affected --run`: all runnable checks passed. `pnpm format`: no changes.
- Found on device while verifying, and worth the reviewer's skepticism about percentages: `flexBasis: '50%'` beside a 12pt `columnGap` renders **one card per row at half width**, because wrapping is evaluated on flex-basis before shrinking. No gate would have caught it; the arithmetic version is what ships.

## Two things this PR cannot show you

- **No CI coverage of the layout.** `.github/workflows/ios.yml` runs `macos-26` with no Xcode pin, and a current run's log points at `/Applications/Xcode_26.6.app` with `iphonesimulator26.5`. There is no foldable, iPad, or regular-width target in any workflow, so nothing here would catch a regression in this column. All measurement is local on Xcode 27.1.
- **No foldable captures driven by navigation.** `tap` and `scroll` against the iPhone Duo's lit inner panel report success and change nothing, filed as #2711, so the regular-width captures above come from iPhone 17 and iPad Pro 13-inch.


thymikee:
Evidence, all on the iPhone Duo simulator (iOS 27.1) with the fold open, so only the inner panel is lit and captures come back at `951x669 @1x`.

**Before** — the stock lab. Cards run edge to edge across the panel and every line of prose is a full 951pt measure:

before

**After** — same screen, same data. The column is capped and centered, so the measure is readable and the card edges no longer sit under the display's own rounded corners. The tab strip is in the same place in both: iOS had already moved it there.

after

**Catalog, two columns.** Each card keeps its own badge, price, `Save`, `In cart`, and `View details`, so the selectors the smoke tests use are unchanged:

catalog grid

The "Open in 'Agent Device Tester'?" sheet in that third image is worth recording because it cost me a wrong conclusion. It is a SpringBoard prompt, not an app alert:

- `alert get`/`alert accept` report `alert not found`, because iOS snapshots require an app session and the sheet belongs to SpringBoard.
- Coordinate taps at both plausible rot90 mappings do not reach it either.
- While it is up it **swallows every app-scoped tap and scroll**. It is queued by `--launch-url` / `simctl openurl`, survives terminating the app, and survives `killall SpringBoard`.

So for a while here I was reading tab taps and `scroll down` as broken on the rot90 panel, when in fact the prompt was in front of the app the whole time. I have not been able to reproduce a genuine tab-tap failure since, and I do not claim one. Worth its own issue: on a `rot90` panel there is currently no programmatic way to dismiss that prompt, and an agent driving the device gets no signal that input is being discarded — the tap returns success and the tree keeps reporting the app's nodes.

thymikee:
Reviewed at 895245f.

At compact width, `gridded` is false on every e2e, smoke, and replay device (iPhone 17 Pro sim, physical iPhone, Android emulator), but the grid container `View` and the per-product wrapper `View` in [CatalogScreen.tsx:103](https://github.com/callstack/agent-device/blob/895245f/examples/test-app/src/screens/CatalogScreen.tsx#L103) still render with `style` undefined, so the catalog tree gains one container plus one wrapper per card versus main. The rule this needs is: the extra wrapper should exist only where the code actually enumerates two columns, i.e. inside the `gridded` branch, so the compact tree stays byte-identical to main. Every capture in this PR is the 951pt foldable panel; there's no compact-width capture, and no suite touches `product-card-*`, `favorite-*`, `details-*`, or `add-*` (a grep over `test/` and `examples/test-app/replays` finds nothing), so if the extra untyped views do land in the iOS AX snapshot or survive the Android covered-subtree prune, nothing in CI will report it. Could this be fixed by returning the card directly at compact width and only wrapping it when `gridded` is true, or is there a reason the wrapper needs to exist unconditionally?

Not blocking, take or leave: `LabLayout.regular`/`width` are unread fields (layout.ts:16), `productCell`'s hand-tuned `'48.6%'` plus stacked `gap: 12` and `marginBottom: 12` puts 24pt between rows but 12pt between columns (CatalogScreen.tsx:178), `GRID_MIN_WIDTH` (760) is numerically identical to `READABLE_MAX_WIDTH` so the grid decision is really "capped column is full width" rather than a window-width rule (layout.ts:13), and `AppFrame` already owns frame geometry and wraps every route but the cap is threaded into five screens instead, leaving `ProductScreen`, `AccessorySetupScreen`, `AutomationLabScreen`, and `WebViewScreen` full-bleed at regular width (components.tsx:208).

Since `AppFrame` ([components.tsx:208](https://github.com/callstack/agent-device/blob/895245f/examples/test-app/src/components.tsx#L208)) already wraps every route and owns frame geometry, could the capped, centered column live there with `useWindowDimensions` inline, so `layout.ts`, the five screen edits, and their imports all go away, and `CatalogScreen` keeps only a local `columns = width >= 760 ? 2 : 1` with the wrapper rendered in the two-column branch? That would also cover `ProductScreen` and the other routes this PR leaves full-bleed, and nothing structural blocks it since `AppFrame` already exists.

The six named checks were in progress at review time with no failure log. Two of them overlap this change for real: `ios.yml` and `android.yml` install a fixture app built from `examples/test-app` and run `01-navigation-scroll.ad` plus the smoke tests against the Catalog screen this PR rewrites, but that replay only scrolls and matches "Catalog scroll: *" / "Seasonal footer target," so it covers traversal, not whether the tree shape changed. Repo Guards and Coverage don't touch `examples/test-app` runtime code.

No app or simulator was run for this review; the layout conclusions come from reading the head tree. Whether `sidebarAdaptable` changes tab hit targets at compact width, and whether RN's plain wrapper views actually show up in the iOS AX snapshot or the Android pruned tree, weren't checked — the runner's depth limit is far above this tree so truncation isn't the risk, node duplication is. `tsc`, `check:affected`, and `format` were not run either.

What would move this forward: make the compact-width tree byte-identical by wrapping the card only in the two-column branch, or post one compact-width Catalog capture from this head (iPhone 17 Pro sim or the Android emulator, Catalog tab) showing `product-card-*`, `favorite-*`, `details-*`, and `add-*` still resolve with card spacing unchanged from main.

thymikee:
Reworked at `5889b7c50`. Both of your structural points were right and are now the shape of the change.

**The wrapper existed unconditionally: fixed, and it was the wrong question.** The cap now lives in `AppFrame`, which already wraps every route and owns frame geometry, so `layout.ts` and the five screen edits are gone and `ProductScreen`, `AccessorySetupScreen`, `AutomationLabScreen`, and `WebViewScreen` get the column too instead of staying full-bleed. `useReadableColumn` returns the width the column actually gets *and* whether the cap bound, and the extra view renders only when `capped` is true — so at compact width there is no container, no cell, and no frame wrapper: the catalog renders the same `SectionCard` element main renders.

**Compact width verified, not asserted.** iPhone 17 sim, portrait `402x874 @1x`, Catalog tab:

compact-catalog-5889b7c

All four selector families resolve against this head, at the top and bottom of the list:

```
is visible id=product-card-citrus-kit     Passed        is visible id=product-card-noodle-pack   Passed
is visible id=favorite-citrus-kit         Passed        is visible id=favorite-noodle-pack       Passed
is visible id=details-citrus-kit          Passed        is visible id=details-noodle-pack        Passed
is visible id=add-citrus-kit              Passed        is visible id=add-noodle-pack            Passed
                                                        is visible id=catalog-footer             Passed
```

Card spacing is main's: the frame's 18pt padding and each card's own 12pt bottom margin, with no grid styles in the path.

**The regular-width grid, and the bug your percentage question was circling.** I kept a percentage at first — `flexBasis: '50%'` with `columnGap: 12` — and on device it rendered **one card per row at half width**: wrapping is computed on flex-basis before shrinking, so 50% + 12 overflows the line every time. That is a silent degradation no gate would have caught. Cell width is now arithmetic off the column width the hook reports, and the two-column decision is `cellWidth >= 340` rather than a window threshold that happened to equal the cap:

regular-grid-5889b7c

Gutters: the grid owns the 12pt column gap, cards keep their own 12pt bottom margin, so the rhythm matches the compact list instead of doubling.

**Routes this now covers for free** — Product detail at 1032pt, which the per-screen version left full-bleed:

product-regular-5889b7c

**Also taken:** `LabLayout.regular`/`width` (unread) went with `layout.ts`; `GRID_MIN_WIDTH === READABLE_MAX_WIDTH` is gone in favour of `MIN_GRID_CARD_WIDTH`; `'48.6%'` is gone.

**Two things still worth your eyes.** The tab bar stays stock `automatic` — `sidebarAdaptable` is gone, and I was wrong that it was what published `[selected]`: the active tab reports `[selected]` with no flag set, so the last reason for it is gone. And navigation-driven captures on the Duo's inner panel are still impossible because `tap`/`scroll` there report success and do nothing (#2711), which is why every capture above is iPhone 17 and iPad Pro 13-inch on iOS 27.1. `tsc`, `pnpm check:affected --run`, and `pnpm format` are clean at this head.

compact-catalog-5889b7c

regular-grid-5889b7c

product-regular-5889b7c

thymikee:
Noting which check is red so nobody reads it as a layout regression: job `106078928123` is **Android**, `agent-device wait text Catalog` timing out at step 11 (`test/artifacts/android-emulator/1789909319070-5124/failed-step-11.*`). That is the same failure #2704 carries at its own head — this branch is stacked on it and inherits it. Every other check here is green, including the two iOS smoke jobs, and the replay that scrolls the Catalog screen passed against the tree this PR changes.

thymikee:
5889b7c looks good. The compact-width tree-identity issue from the earlier review (895245f) is fixed on every route now, so the code is in good shape. It still targets #2704’s branch, so that one lands first.

Not blocking: the PR body and the rework comment credit WebViewScreen as one of the screens that picked up the AppFrame cap, but examples/test-app/app/webview.tsx#L8 renders WebViewScreen without AppFrame (https://github.com/callstack/agent-device/blob/5889b7c/examples/test-app/app/webview.tsx#L8), so that route still renders full-bleed at regular width — swapping WebViewScreen for the route that actually changed, inert, would make the write-up match the diff, but this can be taken or left.

The Android Smoke Tests failure doesn't look related to this diff. The change only touches examples/test-app/src/components.tsx and CatalogScreen.tsx, plain RN layout code, and the failing job times out in the emulator's adb-offline boot loop before it ever reaches the app, the same failure seen at the base PR #2704's own head, which this diff doesn't touch.

I didn't run a live device replay to recheck the Catalog-scroll tree byte-for-byte; that call rests on the green replay check (test/integration/replays/ios/fixture/01-navigation-scroll.ad) plus the static diff showing compact-width JSX output is unchanged from the merge base. I also did not check the behaviour right at the regular-width grid threshold (contentWidth about 692-728pt), which this round did not need to cover.

No conflicts are known and no new finding came out of this round, so there is nothing left to fix here.

thymikee:
Taken — the write-up was wrong, not the code. `app/webview.tsx#L8` renders `WebViewScreen` outside `AppFrame`, so that route never got the cap and I should not have named it. The body now lists the framed routes correctly (`ProductScreen`, `AccessorySetupScreen`, `AutomationLabScreen`, and the `inert` route) and says explicitly that the webview route renders outside the frame.

I left the route alone deliberately rather than wrapping it: `WebViewScreen` hosts arbitrary third-party pages, and capping a `WebView` at 760pt is a different decision from capping the lab's own cards — it would change what the web fixture lanes load and measure, which this PR has no business doing. If you want it framed too, that's a one-line change and I'd rather land it as its own thing.

No code changed; the branch is rebased onto current `main` as `7fd35c058` for that edit alone, on top of #2704 at `c2ecc46c8`. Your read of the Android failure matches mine — it is the base branch's, and it is what #2710 exists to make diagnosable.

thymikee:
7fd35c0 is the same patch as the reviewed 5889b7c, rebased onto #2704's new head, so the earlier result stands and the code is in good shape.

The body now matches the diff: `inert` is named as the route that picked up the cap, and the webview route is called out as rendering outside `AppFrame`, so the one note from the last round is closed. No conflicts are known. The Android and iOS checks are still running on this head and none has failed so far. The next step is #2704 landing first, since this branch still targets it.

thymikee:
Follow-up on the red check: the failing Smoke Tests job at 7fd35c0 is the inherited Android one, not this diff. It dies at `step: wait for Catalog` on the launcher surface, and the base PR #2704 fails the same step with the same message at its own head c2ecc46, which does not carry this change. The width cap only takes effect at regular width and the emulator runs compact, so nothing here should reach that step. The label stays; #2710 adds the pid, resumed-activity and crash-buffer probes that would name the real cause on the next failing run.

thymikee:
One factual correction to the diagnosis, because the two readings imply different fixes. Both failures are the **app-drops-to-launcher** signature, not an emulator boot problem. From the current Android job on this head (`106089269585`, same shape on #2704's `106089267544`):

```
smoke:fixture-bootstrap: 9978ms
command: agent-device wait text Catalog 10000 --platform android --serial emulator-5554 --session android-e2e-48t --json
  screenshot: …/failed-step-11.png
  snapshot:   …/failed-step-11-snapshot.json
  device:     …/failed-step-11-device.txt
```

The emulator booted, the fixture installed, `smoke:fixture-bootstrap` completed in ~10s, and the run got all the way to step 11 — so the adb-offline-boot-loop reading doesn't match this job. Whatever happens between tapping `automation-continue-catalog` and the Catalog screen is the question, and the artifact still cannot answer it: no `app.log`, and `daemon.log` is two lines. That is precisely the gap #2710 closes, so one Android run on a head that carries it should name the library.

thymikee:
Rebase checked at `129896d`, no new code review. Your lab commit is byte-identical to `7fd35c0`; the head moved only because #2704 was rebased under it. The clean verdict from that review stands and `ready-for-human` stays on.

One thing to watch rather than a finding: #2704's iOS smoke job at `e2404e6` now fails with `id="automation-longpress" did not become visible after scrolling`, which is new since the `xcrun` timeout. Because this PR reshapes the same screen's content column, it would be useful to say whether the cap moves that row back into reach or leaves it where the base run found it.

thymikee:
Rebased onto #2704 at `ae8fad731`; this branch's head is `75380723f` and the diff is unchanged from the reviewed `5889b7c`.

The inherited Android red that this PR kept wearing is now explained and fixed underneath it: the crash buffer #2710 added captured `NativeTabs.Trigger.Icon 'md' icons on Android require 'expo-symbols'` throwing inside `TabsLayout`, which is what dropped the app to the launcher. See #2704's last comment for the log. The iOS red is #2714, a `scroll` defect in the CLI that reproduces against `main` with only this fixture underneath it. Neither is this diff.

thymikee:
Also green end to end here: 19/19 checks on `75380723f`, Android smoke included (30m18s), so the inherited red this PR was carrying is gone with #2704's `expo-symbols` fix. No diff change from the reviewed `5889b7c`.

thymikee:
Rebase checked at `7538072`, no new code review. `git range-diff` against `129896d` shows your lab commit byte-identical; the head moved only because #2704 picked up its `expo-symbols` fix underneath. The clean verdict stands and `ready-for-human` stays on.

That also answers the question left open here. The iOS smoke lane is green on #2704 alone at `ae8fad7`, without this width cap, so the cap is not what moved the long-press row back into reach — the earlier failure tracks #2714's scroll defect rather than the lab's height. All 19 checks pass on this head too.

PR #2708, opened 2026-09-20
test(ios): size the alert command's deadline to the host's own reads

thymikee:
## Summary

The replacement-alert fixtures give the alert command `timeoutMs: 10000`. It starts at dispatch, and `resolveAlert` spends 24 accessibility reads before choosing a button, so the budget only means something in units of one read's cost: ~10 ms when healthy, ~1.7 s when contended — enough to expire mid-resolution, before the tap is synthesized. [Runs 35](https://github.com/callstack/agent-device/actions/runs/35307766846), [36](https://github.com/callstack/agent-device/actions/runs/35421818827), [37](https://github.com/callstack/agent-device/actions/runs/35489655382) needed 31.2–40.4 s; green run 34 spent 6.2 s.

The fixtures now time one read once the alert is up and derive `timeoutMs` from it, floored at production's default, capped at this lane's slowest test. Production keeps 10 s. Two files.

## Validation

`3a7e70d77c` + `6b1f0c1a00`. `pnpm check:affected --run --head 3a7e70d77c` exits 0; `check:fallow --base origin/main` clean. Hosted `iOS / Smoke Tests`: `Executed 98 tests, with 0 failures`, sizing test and fixtures included. Healthy local timings unchanged (14.81 s vs 14.73 s: budget stayed on the 10 s floor). Mutation: pinning the budget to 250 ms reproduces run 37 exactly. A contended host is not reproducible locally, so the rescue is arithmetic against the traced red nights. All checks green; `Repo Guards` flaked once on `scripts/node-test-tmpdir.test.ts:305`, unrelated, green on re-run. Refs #2546 — its typed empty-button outcome stays open.


thymikee:
This is ready to merge at 6b1f0c1. The .github/workflows/ios.yml -only-testing list matches the tests this PR changes, so the iOS Smoke Tests job (98 tests, 0 failures) is direct evidence for this diff, not an unrelated pass. Not blocking: the 24-read decomposition in apple/runner/AgentDeviceRunner/AgentDeviceRunnerUITests/UnitTests/RunnerTests+AlertObservationTests.swift#L51 is a hand-maintained constant that could silently under-count if resolveAlert/handleAlert grows more AX reads, the sampled static-text read at line 88 only approximates the real per-read cost it scales, and the confirmed=true fixtures at line 89 can now take up to 60s to report ALERT_DEADLINE_EXCEEDED instead of 10s — all fine to take or leave. #2546 stays open: its second symptom, the empty-button case needing a typed outcome, is out of scope here and this PR does not claim to fix it.

Issue #2707, opened 2026-09-20
Recording touch overlay exports at ~220x480 or all-black whenever touches are drawn


thymikee:
## Purpose

`record` with the default touch overlay produces an unusable export as soon as the recording contains an interaction: the track collapses to roughly `220x480` regardless of the device's native capture size, and on a long recording it comes out entirely black. Touch indicators are therefore unavailable for exactly the recordings that need them, and the caller gets exit code 0 either way.

This is the recording transport, not the display-selection path. The raw capture honors `--display` correctly since #2703.

## Observed

iPhone 17 simulator `0E0FCBDC-88E2-4CB6-A1DA-E148F77743F4` (iOS 27.0, non-rot90 panel), Xcode 27.1 beta, `agent-device` at `057dfb7c1`. Mean luma measured over sampled frames; content changes across frames are marked where they occur.

| Overlay | Interactions during capture | Exported size | Result |
| --- | --- | --- | --- |
| `--hide-touches` | none | `1206x2622` | correct |
| default | none | `1206x2622` | correct |
| default | two tab taps | `220x480` | bright, content changes present |
| `--hide-touches` | two tab taps | `1206x2622` | correct, content changes present |
| default | taps, scroll, 97s | `480x220` | `meanLuma=0.00` every frame |

Two things follow from the matrix. The trigger is the overlay drawing touch events, not the interactions themselves — `--hide-touches` stays full size while recording the same taps. And it is not orientation-dependent: the failing device here has a normal portrait panel.

The overlay does draw something, since the small export tracks the screen. Only the geometry and, in the long case, the frame content are lost.

## Required behavior

- A recording with touches exports at the same track size as the same recording without touches.
- A recording never exports all-black. If the overlay compositor cannot produce frames, fail the command or drop the overlay and say so in the result, rather than returning a black file with exit 0.
- The choice to drop the overlay must be reported on the response, not inferred from pixel dimensions by whoever notices later.

## Completion conditions

- Table above re-measured on a non-rot90 target and on the iPhone Duo inner panel, with the overlay rows matching the `--hide-touches` rows for size and non-blackness.
- A regression test asserts exported track size equals the raw path's for a scripted tap, since "file exists and is non-zero bytes" would have passed all four broken rows above.
[...]

PR #2709, opened 2026-09-20
docs(ios): scope the overlay defect to touch drawing, not rot90

thymikee:
## Summary

ADR 0025 attributes the touch-overlay export failure to the iPhone Duo inner panel's `rot90` track and cites an intact non-rotated iPhone 17 recording as the control. That control contained no touches to draw, so it proved nothing.

Measured on an iPhone 17 (iOS 27.0), which has no rotated panel: a 10s recording containing two taps exports `220x480`; the same two taps under `--hide-touches` export `1206x2622` with on-screen content changing; a 97s clip with touches exports `480x220` at mean luma 0.00 throughout. So the trigger is the overlay drawing touch events, on any panel, and touches are unavailable in exactly the recordings that need them. `docs/agents/device-verification.md` gains the matching guidance: make an interaction legible through its on-screen effect, not a cursor.

Two files, docs only. The code defect is recorded in #2707, which this does not close.

## Validation

Commit `c0e19da4c`. Runtime validation does not apply to a docs-only change; `pnpm check:affected --run` passed and `pnpm format` produced no changes.

The numbers above are reproducible in one command pair on any booted iPhone simulator:

```sh
agent-device record start /tmp/a.mp4 && agent-device tap 'label="Form"' && agent-device record stop
agent-device record start /tmp/b.mp4 --hide-touches && agent-device tap 'label="Form"' && agent-device record stop
```

`/tmp/a.mp4` comes back at a fraction of `/tmp/b.mp4`'s track size.


thymikee:
Reviewed at c0e19da. This is ready to merge. The rewrite scopes the overlay defect to touch drawing, and the numbers match issue #2707's measured table.

Not blocking: line 172 ("Feeding an untouched raw `simctl` capture straight into `recording-overlay.swift` reproduces a `0x0` zero-duration output") is carried over from the retracted rot90 paragraph and now lands right after the iPhone 17 sentence, so it reads as a no-touch case contradicting the control row above it — worth naming the Duo-inner-panel sample and its recorded touches, or dropping the sentence since #2707 owns the exporter defect, but take it or leave it. https://github.com/callstack/agent-device/blob/c0e19da/docs/adr/0025-foldable-apple-panels.md#L172

I did not re-run the record/tap repro commands myself; I relied on #2707's table being internally consistent, and I did not check whether #2707 or a sibling doc still carries the old rot90 attribution elsewhere in the repo. This is a docs-only change with all checks passing, so nothing here needs to hold up the merge.

PR #2710, opened 2026-09-20
test(android): name the crash in failed-step evidence

thymikee:
## Summary

An Android e2e step that ends on the launcher is ambiguous in CI. The `wait for Catalog` failure on #2704 uploaded a two-line `daemon.log` plus a snapshot showing Play Store, Gmail, and the fixture icon — identical whether the process crashed or merely navigated somewhere unexpected. With no logcat in the artifact, the only way to tell was to bisect majors across several CI rounds, which is exactly what the review asked me to avoid.

Failed-step device evidence gains three probes: `pidof` for the fixture app, the resumed activity, and the crash buffer. A dead pid next to a launcher `mResumedActivity` is a crash, and the crash buffer names the library in that same run. The resumed-activity selector keeps the `ActivityRecord` hash with the component name, so the same hash on a new component is a navigation and a new hash is a restart.

The adb read moves out of `live-harness.ts` into `device-evidence.ts`, where the seam is a single query function, so a test can replay a device. `deviceEvidenceTimeoutMs` becomes pass-through from the harness and rises with the probe count: the bound protects the screenshot, and losing the document whole would drop the rotation facts the existing probes carry. 4 files, +224/-73.

Two follow-ups landed on top after review: the suite is registered in the `unit-core` lane (it was running in nothing, since `test:integration:node` globs one level deep), and the probes are minted through `deviceShellArgv` because #2611 made that the only way a device-shell command may exist and `pidof` interpolates the fixture app id.

## Validation

Head `30a82efa7`, rebased on `main` at `3155aea61`. `pnpm check:affected --run` and `pnpm format` passed.

The four tests replay adb output through the probe seam rather than needing an emulator: crash, alive-process navigation, a failing probe, and app-id quoting. Mutations, each run against the named lane and then reverted:

- deleting the `['crash buffer', …]` probe → 2 of the first 3 go red;
- hand-building the `pidof` argv instead of minting it → the quoting case goes red.

`pnpm exec vitest run --project unit-core test/integration/android-emulator-e2e/device-evidence.test.ts` is the lane that catches both.

Remaining limitation, stated rather than papered over: the probes themselves were not exercised against a live emulator, because this host has no AVD (`~/.android/avd` is empty). The next Android smoke run on a failing head is the first real read; if a probe times out on a loaded runner, it records `(failed: …)` for that section and leaves the rest of the document intact, which is the behavior the third test pins.


thymikee:
Reviewed at c24a015.

test/integration/android-emulator-e2e/device-evidence.test.ts is not run by any lane. `test:integration:node` globs `test/integration/*.test.ts`, one level deep, so it misses this nested path (https://github.com/callstack/agent-device/blob/c24a015/package.json). `test:smoke` only matches `test/integration/smoke-*.test.ts`. No vitest project include covers this path either; the one nested file that is covered, `test/integration/support/daemon-leak-model.test.ts`, is listed explicitly in vitest.config.ts. And no workflow references `device-evidence`. The android smoke lane imports `device-evidence.ts` but never the `.test.ts` file, so today these three tests only run when someone runs the file by hand. Since this PR's whole value is the guard, an inert lane means a later change that drops the crash-buffer, app-process, or resumed-activity probe — the exact regressions this test targets — would merge green. Because the file uses `node:test`, adding a vitest include alone won't fix it. Can you register the file with a lane in this PR, either by adding its path to the `test:integration:node` script's `node --test` arguments alongside the existing glob, or by converting it to vitest and adding an explicit include in vitest.config.ts the way `daemon-leak-model.test.ts` was added in 5e48486ad8? Then confirm discovery by breaking one probe on purpose and watching the named lane fail, per docs/agents/testing.md.

The PR could not be run against a live Android emulator here, so the pidof/dumpsys/logcat selectors are unverified against real device output; the PR body already flags this and defers first live validation to the next Android CI run on a failing head. The PR body points at #2704, which is the iOS scene-lifecycle change rather than an Android crash-ambiguity one; that looks like a linking mistake rather than a code problem, so it should not hold up the code itself.

The only non-passing check, Smoke Tests (run 35509757986, job 106075495758), is the iOS workflow's job and was still in progress, not failing; its steps are all iOS-specific and this PR only touches Android test files and shared harness types in live-device-e2e/runtime.ts, so there's no overlap with what that job exercises. Once the test file is wired into a real lane and that lane is shown to fail on a broken probe, this is ready to merge, pending the unrelated iOS job finishing.

thymikee:
Both findings taken. Rebased on current `main`; head is `30a82efa7`.

**The suite was inert — registered, and proven to bite.** You were right that nothing ran it: `test:integration:node` globs `test/integration/*.test.ts` one level deep. I converted the file to vitest and added it explicitly to `unit-core`, following the `test/integration/support/daemon-leak-model.test.ts` precedent you pointed at — it replays adb output through the probe seam, so it needs no emulator, daemon, or subprocess and belongs in the fast lane beside the selectors it pins.

Discovery confirmed against the named lane, per `docs/agents/testing.md`:

```
$ pnpm exec vitest run --project unit-core test/integration/android-emulator-e2e/device-evidence.test.ts
Tests  3 passed (3)                      # baseline

# crash-buffer probe deleted from the probe table
× a dropped process with a launcher activity reads as a crash and names the library
× one failing probe records its failure without taking the rest of the document down
Tests  2 failed | 1 passed (3)           # probe restored → green again
```

I chose vitest over appending the path to `test:integration:node`'s `--test` arguments because that script is a one-level glob by design and there are 77 nested files under `test/integration/` that have never run in it — widening it here would either drag all of them in or leave a one-off exception in a shared script.

**One more finding from the rebase, and it was a real one.** `main` gained #2611 while this was open: `adb shell` forwards its words to the device's `sh`, and a device-shell command may now only come from `deviceShellArgv`. My `pidof` probe interpolated the fixture app id into exactly that position, so it was the shape that change retired — the three static probes I moved from `live-harness.ts` were innocent, but the one I added was not. It now goes through `deviceShellArgv` with `-s <serial>` as the documented prefix, and a fourth case pins both renderings: a safe app id reaches the device byte-identical, `com.lab'; reboot` comes back as one single-quoted word. Hand-building that argv again fails the new case. The two `logcat` queries are host-side and stay outside the boundary.

**On the #2704 reference:** not a mislink. #2704 is the scene-lifecycle change, but its Android smoke is also the failure this evidence targets — `wait text Catalog` at step 11 with a launcher snapshot and no `app.log`. I want one CI round there to name the library instead of bisecting Expo majors, which is the whole point.

Live-device caveat still stands and is in the body: no AVD on this host, so the `pidof`/`dumpsys`/`logcat` selectors are unverified against real device output until a failing Android head runs them. `pnpm check:affected --run` and `pnpm format` are clean at this head.

thymikee:
Reviewed at 30a82ef. This is ready to merge.

Not blocking: the new pidof case only asserts minting for one probe, so a later hand-built shell argv for the other four probes would still pass (test/integration/android-emulator-e2e/device-evidence.test.ts#L88) — could `assertDeviceShellArgv` cover every probe argv in the loop rather than just this one, and the moved addressing in the probe argv lines (test/integration/android-emulator-e2e/device-evidence.ts#L41) still hasn't run against a real device, but that gap was already accepted last round, so both can be taken or left.

Smoke Tests are still queued, so there's nothing failing yet; the touched files are Android test-harness code, live-device-e2e/runtime.ts types, and vitest.config.ts, so the iOS smoke route has no overlap with this diff. Merge once those queued jobs finish; the first live Android run against a failing head will be what actually exercises these probes.

PR #2713, opened 2026-09-20
test: derive the command-surface enumerations that restated a declaration

thymikee:
## Summary

Follow-up to #2702, which touched 72 files to add one command. Most of that fan-out is declaration sync the registry genuinely owns, but three gates were re-listing a fact that is already derivable at its declaration site. This PR removes those three, so the next public command costs zero edits in the seven files this PR touches.

- **Coverage summaries.** Each platform manifest already builds its classification summary from its own rows, and its smoke test then compared that derived value to a hand-incremented literal (`54` → `55`). `assertCoverageClassificationSummaryWiredToManifest` now recounts the manifest independently of the builder, ties `total` to the public catalog's size, and checks the buckets still sum.
  It guards **wiring and denominator, not per-row classification**, and says so in its name and doc comment: the recount and the published summary read the same rows, so a row moved between buckets moves both and stays consistent.
- **Row membership is gated by evidence instead.** `assertLiveCoverageMatchesEvidence` ties a platform's live bucket to the commands its own scenario sources execute, read independently of the manifest's `level` fields — macOS through `parseReplayScriptDetailed` plus the provider call shapes, web through the argv heads its smoke scenario passes. Linux already compared its live set against the parsed replay script and Android against each scenario's own command declaration; tvOS claims nothing live. A command executed but not claimed, or claimed but no longer executed, now fails and names its platform.
- **Daemon generic route.** The test enumerated eleven commands to prove the router's fallthrough is reached on purpose. Ten declare `daemon.route: 'generic'`, which is what `getDaemonCommandRoute` reads, so the list was a second source of truth. The expected set is that trait projected; what stays hand-written is the case the list guarded — a catalog command with no daemon facet at all, landing on the `?? 'generic'` fallback — asserted alone, with a message naming the declaration to add.
- **Fail-closed test facts.** `createUnavailableRuntimeFactsForTest` restated all 27 fail-closed cells plus nine lifecycle cells to say what `createFullyUnavailablePlatformRuntimeFacts` already builds. The forms are equal because the input's `network` cell carries the same sentinel the omitted optional cells inherited, so the returned facts are unchanged for all 23 importers.

Parity was checked before deletion, per ADR 0008: the five removed literals (macOS 22/15/18/55, web 42/1/12/55, Linux 29/9/17/55, tvOS 17/38/0/55, Android 14/0/41/55) each equal the derived rollup at this head. The web live roster likewise still resolves to the same 12 commands the pin counted, now read off the scenario rather than typed by hand.

## Deliberately not taken

- `unavailableDeploymentSnapshotAndShutdownOperationFacts` is a partial flat record spread into 23 fixture files, in mixed positions — after another spread in `src/platform-runtime-gateway.fixtures.ts`, before per-test claims elsewhere. Completing it would clobber claims at some sites and needs a per-site audit, not a mechanical fill.
- `REPLAY_SCRIPT_LINE_PLANS` in `property-arbitraries.ts` and the per-platform rows in `command-coverage/declarations.ts` are genuine per-command claims (a replay line shape; a platform's own evidence pointer and sentence), already colocated one block per command. Deriving them would invent a default that erases the judgment.
- Per-scenario (rather than per-platform) live equality on macOS. Its provider scenarios execute `open`, `appstate`, `snapshot`, `screenshot`, `wait` and `click` whose live claims the manifest attributes to the replay scenario, so equality holds only at platform level — which is what the gate asserts.

## Validation

Tested at `d950b44`.

- `pnpm check:affected --run`, `pnpm check:tooling` and `pnpm check:unit` (unit + tmpdir + smoke) all exit 0.
- The five coverage lanes run directly: 37 tests before the macOS gate, 15 across the macOS and web files after it. The daemon route test runs with the generic-route completeness test.
- Every changed gate was planted red, per `docs/agents/testing.md`: a mis-bucketed summary and a denominator drift each fail naming the platform; deleting `viewport`'s `daemon` facet fails the route test with `['viewport', 'install-from-source']` against the single expected fall-through; appending `callCommand('hover', …)` to a macOS live scenario fails with `macOS live claims are not the commands its own evidence executes`; a `['hover',` argv head in the web scenario fails the same way for web.
- No production module, no Swift, no device-facing behavior changed, so no runner rebuild or simulator evidence is claimed.
- CI note: the iOS `Smoke Tests` job failed twice on the first head at two unrelated steps (scroll visibility, then `ALERT_DEADLINE_EXCEEDED` with `IOS_TREE_CAPTURE_TIMEOUT`) and went green on a third attempt with no code change; `a371ad284` failed the same workflow in the same window. Logs are in the comments below.


thymikee:
Reviewed at d6c0d16.

`assertCoverageClassificationSummaryDerivedFromManifest` recomputes `recounted` from `Object.values(manifest)` and compares it to `summary`, but every platform's `..._CLASSIFICATION_SUMMARY` constant (for example `ANDROID_EMULATOR_COVERAGE_CLASSIFICATION_SUMMARY` at https://github.com/callstack/agent-device/blob/d6c0d16/test/integration/android-emulator-e2e/coverage.ts#L12) is itself built by `buildCoverageClassificationSummary(Object.values(sameManifest))`. Both sides are pure functions of the same manifest rows, so a command mislabeled with the wrong `level` moves `summary` and `recounted` together and the check still passes: only a bug inside `buildCoverageClassificationSummary` itself, or a summary wired to the wrong manifest, would fail it. The old hand-typed literal was independent of the manifest and would have caught that drift. Can the doc comment and PR body say plainly that this only guards wiring and denominator drift, not per-row misclassification, or should the test tie bucket membership to an independent signal such as the live-command evidence checks that already exist per platform? See https://github.com/callstack/agent-device/blob/d6c0d16/test/integration/support/coverage-classification.ts#L378.

`smoke-web-platform-coverage.test.ts` weakens `assert.equal(liveCommands.length, 12)` to `assert.ok(liveCommands.length > 0, ...)` at https://github.com/callstack/agent-device/blob/d6c0d16/test/integration/smoke-web-platform-coverage.test.ts#L55. `liveCommandsForWebSmoke()` filters `WEB_PLATFORM_COVERAGE` for `level === 'live'`, and that manifest is untouched by this PR, so the count is still exactly 12. Was this loosening needed, or can the exact-count assertion come back? As written, a future change that drops most of the web live-command roster would pass silently instead of failing.

Not blocking: `localCliCommands()` in `src/daemon/__tests__/request-handler-catalog.test.ts` filters for `catalog.group === 'local-cli'`, but `genericByAbsence` only iterates `CATALOG_COMMANDS` (public and internal groups, mutually exclusive from local-cli), so the `!localCliCommands().includes(command)` clause can never trigger — worth dropping as dead code or fixing the intent, but take it or leave it.

The diff touches the five smoke coverage test files and the daemon route test directly, and Smoke Tests (both) and Coverage were still queued or in progress at review time with no failure output captured, so there is no basis yet to call any of them unrelated if they do fail. The recount finding and the loosened web live-command count both need an answer, in the doc comment or the PR body, before this is ready to merge.

thymikee:
The iOS `Smoke Tests` job is red on this head across two attempts, at two different steps, and this diff changes no production TypeScript and no Swift — all 8 files are test or test-support modules:

```
src/__tests__/test-utils/runtime-operation-facts.ts          (test util)
src/daemon/__tests__/request-handler-catalog.test.ts
test/integration/smoke-{android-emulator,linux,macos,tvos-platform,web-platform}-coverage.test.ts
test/integration/support/coverage-classification.ts
```

- Attempt 1 failed `Run fixture-backed iOS simulator E2E smoke`: `id="automation-longpress" did not become visible after scrolling`, last probe `selector_not_found` on `is visible`.
- Attempt 2 passed that step and failed `Run targeted iOS runner XCTest regressions`: `testAlertAcceptDoesNotActivateAReplacementWithASharedButton` → `ALERT_DEADLINE_EXCEEDED`, with `AGENT_DEVICE_RUNNER_SNAPSHOT_BACKEND_FAILED backend=tree error=the XCTest tree capture exceeded its 8s time slice` and `AGENT_DEVICE_RUNNER_SYSTEM_MODAL_PROBE_ABORTED elapsedMs=4008 error=IOS_TREE_CAPTURE_TIMEOUT` in the same run.

Both signatures are capture/deadline exhaustion rather than a changed assertion, and `a371ad284` — a different pull request, same window — also failed this workflow. `Coverage`, `Integration Tests`, `Repo Guards`, `Bundle Size` and the other twelve checks are green here, including the macOS `Smoke Tests` job. A third attempt is running; I read these two as the iOS lane's timing instability rather than this diff, and will not code around them.

On the change itself: `docs/agents/testing.md` asks that a changed structural gate be planted red before it is trusted, and both new assertions were — a mis-bucketed summary and a denominator drift each fail naming the platform, and deleting `viewport`'s `daemon` facet fails the route test with `['viewport', 'install-from-source']` against the single expected fall-through. Parity also holds for the five deleted literals (macOS 22/15/18/55, web 42/1/12/55, Linux 29/9/17/55, tvOS 17/38/0/55, Android 14/0/41/55): each still equals its manifest's derived rollup, so removing the pins drops their maintenance and not an invariant.

thymikee:
Third attempt on the same head `d6c0d16` is green: all 17 checks pass, iOS `Smoke Tests` included, with no code change between attempts. That confirms the earlier two failures were the iOS lane's capture/deadline instability rather than this diff.

thymikee:
Addressed at `d950b44`, taking both halves of the first finding rather than only the wording.

**The recount is tautological for per-row misclassification** — correct, and it is now stated in the name (`assertCoverageClassificationSummaryWiredToManifest`) and the doc comment: it guards a summary wired to the wrong array, a mis-bucketed rollup, and a denominator that stopped matching the catalog. Membership moved to the independent signal you pointed at. `assertLiveCoverageMatchesEvidence` derives the commands a platform's evidence actually executes and compares both directions against the manifest's live set — macOS via `parseReplayScriptDetailed` for the replay scenario plus `command: '…'` / `callCommand('…'` for the provider ones, web via the argv heads its smoke scenario passes. macOS executes exactly the 18 commands it claims live and web exactly the 12, so the equality holds as written. Linux already did this against its parsed replay script and Android against each scenario's command declaration, and tvOS claims nothing live.

My first attempt at that macOS tie compared each scenario's partition and found the provider scenarios running `open`, `appstate`, `snapshot`, `screenshot`, `wait` and `click` with no live claim, which I recorded as a deliberate narrow claim. That reading was wrong — those rows are live, attributed to `replay:system-settings` — so the gate compares at platform level and the commit message owns the correction.

**The web live count** is not loosened anymore; the `> 0` was the wrong call. The roster is now the parsed argv-head set compared to the live set, which still pins twelve and additionally fails on a command the scenario runs without a claim, or a claim the scenario stopped honouring. Appending a line containing `['hover',` to the scenario fails it on `hover`.

**`localCliCommands()` was dead** — dropped. `CATALOG_COMMANDS` is the public and internal groups and a `catalog.group: 'local-cli'` descriptor is never projected into either; the reason now sits where the clause was rather than in a filter that cannot fire.

On evidence for the lanes: fair, at review time Smoke Tests and Coverage were queued and I had no basis to call anything unrelated. Coverage, Integration Tests, Repo Guards and macOS Smoke are green on the first head, whose iOS job went green on a third attempt with no code change between attempts. The new head's runs are in progress; I will not merge ahead of them.

thymikee:
Reviewed at `d950b44`. Both halves of the earlier finding are answered — the renamed summary guard now fails on a rollup wired to the wrong array, and the live buckets are closed from the evidence sources themselves. This is ready for human review.

Not blocking: `commandsExecutedInScenarioSource` and `sourceContainsCommand` in `test/integration/smoke-macos-coverage.test.ts` switch on the same three syntaxes and re-derive the same command tokens from the same evidence files, so `sourceContainsCommand` could check membership in the new enumeration's result instead of parsing again — worth doing or leaving.

Smoke Tests is still running on this head. The diff touches only test and test-support files and does not overlap the iOS runner capture path that lane exercises, so it is worth letting it and Coverage finish before merging.

thymikee:
Checks on `d950b44`: all 17 green. The iOS `Smoke Tests` job failed its first attempt on the same step as before, `live iOS simulator fixture E2E` → `id="automation-longpress" did not become visible after scrolling`, and passed on a re-run of the identical head with no code change. Nothing in this diff reaches that lane — no production module, no Swift, and none of the files it reads.

PR #2718, opened 2026-09-20
feat(diff): accept JPEG inputs for screenshot comparison

thymikee:
## Summary

`diff screenshot` decoded both inputs as PNG, so a capture exported by another tool had to be converted first and a HarmonyOS artifact — which the device serves as JPEG under whatever name the command was given — could never be compared. Each input is now decoded from its own bytes: the container is sniffed, so a `.png` name holding JPEG decodes as JPEG.

`png-transcode.ts` became `screenshot-image.ts`, the one owner of container sniffing for both the decode path and the provider transcode path, and the PNG worker gained a `decode-image` job that answers pixels instead of PNG bytes, so a full-resolution decode never blocks the daemon event loop. The `--out` diff image stays PNG, as do the crop, overlay, and resize passes that rewrite a screenshot in place.

13 files, ~80 net production lines. No new flag, no schema or wire change.

```sh
agent-device diff screenshot --baseline baseline.jpg --out diff.png
agent-device diff screenshot --baseline baseline.png harmony-capture.jpg --threshold 0.2
```

## Validation

`pnpm check:affected --run` passed (all runnable checks) at `46b0079daf`, after `pnpm check:quick`. `8d3a8719a9` only restores gate-generated `.agent-device/test-artifacts` files to their `origin/main` content; no source changed.

New coverage: worker↔sync decode parity for both containers, `decode-image` transfer ownership, a JPEG stored under a `.png` name comparing exactly at `--threshold 0`, decoded dimensions on a JPEG, corrupt-JPEG and neither-container refusals with their typed details, and the diff artifact pinned to a PNG signature. The screenshot-diff suite now runs the shipped worker client instead of a module mock.

Unresolved risks, both pre-existing and deliberately not folded in here:
- `packages/platform-harmonyos/src/screenshot.ts:15` still writes JPEG bytes into a `.png` artifact, so `--scale` and `--overlay-refs` fail on that platform. The provider-side transcode Limrun got in #2645 is the owning fix; this PR makes those artifacts diffable meanwhile.
- A hostile input file is fully decoded before `maxImagePixels` is consulted (already true for PNG); `jpeg-js` self-caps at 100 MP / 512 MB.


thymikee:
Reviewed at 8d3a871. This looks ready to merge. CI is green across all 20 checks, and I didn't find anything that blocks it.

I traced the JPEG decode and re-encode paths manually rather than running the suite, relying on CI green plus reading the code against the test assertions.

Not blocking: `transcodeScreenshotToPng` narrows the try/catch around `PNG.sync.write` compared to the old `jpegToPng` (https://github.com/callstack/agent-device/blob/8d3a871/packages/capture-kit/src/screenshot-image.ts#L43), so a re-encode failure would surface as a raw Error instead of the typed `COMMAND_FAILED` — worth a look, but fine to take or leave as a follow-up.

PR #2723, opened 2026-09-21
feat(settings): read and apply the preferred text size

### LINKED ISSUE #2695: feat(settings): support preferred text size / Dynamic Type
## Problem

`agent-device settings` does not expose the preferred text-size / Dynamic Type category. This makes accessibility and layout QA less reproducible: an agent can observe unexpectedly large text, but it cannot inspect or restore the simulator's content-size category through the `agent-device` command surface.

Observed with `agent-device` 0.21.0 on an iOS simulator. The current workaround is to leave `agent-device` and call CoreSimulator directly:

```sh
xcrun simctl ui <udid> content_size
xcrun simctl ui <udid> content_size large
```

Driving the Settings app is possible, but it is slower and depends on localized labels and slider coordinates.

## Proposed command

Add an iOS-simulator settings helper, for example:

```sh
agent-device settings text-size
agent-device settings text-size large
agent-device settings text-size accessibility-extra-large
```

The no-argument form should return the normalized current category. The setter should accept the categories supported by `simctl ui ... content_size`:

- `extra-small`
- `small`
- `medium`
- `large`
- `extra-large`
- `extra-extra-large`
- `extra-extra-extra-large`
- `accessibility-medium`
- `accessibility-large`
- `accessibility-extra-large`
- `accessibility-extra-extra-large`
- `accessibility-extra-extra-extra-large`

Exact naming can follow the existing settings-command conventions.

## Scope

- Implement the first slice for iOS Simulator by delegating to `simctl ui <device> content_size`.
- Respect the active session or explicit device selection (`--device` / `--udid`).
- Return `UNSUPPORTED_OPERATION` with a useful message on targets where agent-device cannot set the category.
- Android font-scale parity can be handled separately because its values and behavior do not map exactly to Apple's categories.

## Acceptance criteria

- The current preferred content-size category can be read through CLI and structured API/MCP responses.
- A supported category can be set on the selected iOS simulator.
- The result reports the applied normalized category rather than only command success.
- Invalid category values fail before mutating simulator state and list the accepted values.
- Help text and command documentation include the new setting.
- Tests cover parsing, device selection, success, invalid input, and unsupported-target behavior.

thymikee:
## Summary

`settings text-size` reads the preferred text size a target holds; `settings text-size <category>` applies one of 12 shared rungs. iPhone/iPad simulators serve it via `simctl ui <device> content_size`; Android maps rungs onto `system font_scale` multipliers (`large` = 1.0 … `accessibility-extra-extra-extra-large` = 3.2) and names the nearest rung for a multiplier it never wrote, keeping the exact value as `platformValue`.

```sh
agent-device settings text-size                      # category: large, platformValue: 1.0
agent-device settings text-size accessibility-large
```

The read is its own leg: `readSetting` gets a fact cell, its own admission, and its own recording/ref-frame classification, because a leaf can hold a setting nobody verified it can report. Off-ladder categories are refused before any device call — `simctl ui … content_size bogus` exits 0.

54 files, +1777/−106, 860 of them production. Reviewed by codex, a strict internal pass, and a second adversarial pass whose findings are the five commits on top of `13211c4e`: the leg rule is resolved once in `platform-runtime-operations.ts` instead of twice plus a parity pin; `ReadSettingInput.setting` is the readable list, so owner read legs are exhaustive maps rather than switches with a default refusal; an Apple simulator with no content size is refused before admission, so a write it will not perform no longer expires the session ref frame (ADR 0019 §"no stub throws after binding"); the read payload is keyed by `setting` and both response sentences come from the vocabulary. Rejected: a shared `settings` cell, which hides a leaf that writes without reading, and a separate Android module or eager vocabulary import, both of which grow a ratcheted closure.

Closes #2695

## Validation

`f58ef0545`: `pnpm check:affected --run` passed; `pnpm test:unit` 10,929 passed; `check:fallow`, `check:layering` (largest type cycle 6 files, R10 held at its merge-base counts), `lint`, `typecheck`, `check:mcp-metadata`, `check:command-docs` clean.

`node bin/agent-device.mjs` from this commit:

- iPhone simulator (iOS 27.1): read `large` → `{ setting, category, platformValue, message }`; wrote `accessibility-large`, read it back, `simctl ui <udid> content_size` agreeing both times; earlier passes confirmed `small` → `platformValue: "Small"` and `gigantic` → `INVALID_ARGS` naming the ladder.
- macOS host: read and write give the same `INVALID_ARGS: Unsupported macOS setting: text-size`.
- API 35 emulator: 12/12 rungs round-tripped against `adb shell settings get system font_scale`; unset reads `large`/`1.0`; `1.2` → `extra-extra-large`/`1.2`.

The tvOS/visionOS pre-admission refusal is pinned by tests (frame stays `active`, owner never bound); no such simulator is installed here, so it is not device-verified. Devices restored, emulator shut down, daemon cleaned. Fact cells only, no device: HarmonyOS, Vega, Linux, web, webdriver; limrun Android admits the read through the same `createAndroidInteractor` but was not driven through that provider.


thymikee:
Reviewed at 13211c4. The read and write legs route through their own fact cells, and I found no defect in the code. The live runs on the iOS 27.1 simulator and the API 35 emulator are as reported in the PR body; the HarmonyOS, Vega, Linux, web, webdriver and limrun refusals are checked from the fact code only.

One design question before this goes further. The PR body explains why a shared `settings` cell was rejected. A smaller option it does not cover: keep only the `setSetting` cell and add a per-setting read gate in the daemon, next to the existing macOS gate. That would skip the new operation, binder, interactor member, catalog row and seven fact producers. Would that do, or is the owner-fact refusal the reason the read needs its own cell?

All checks pass.

The PR has merge conflicts with main, so I could not check the code against the current main. Please resolve them and rerun the checks.

Not blocking: on the macOS host a text-size write fails with `INVALID_ARGS` ([snapshot-settings.ts#L262](https://github.com/callstack/agent-device/blob/13211c4ef715dd91070b47aee072d0f4437c2628/src/daemon/handlers/snapshot-settings.ts#L262)) while the read fails with `UNSUPPORTED_OPERATION`, the `resolveSettingsReadRequest` doc comment says the registry shares a rule it only copies, the parity test lists the settings subcommands inline, and limrun Android also admits the read but is not in the PR body's list of unverified targets.

thymikee:
Rebased onto `45e4c594a`; the only conflict was the app-id precedence, where Maestro's `req.internal.settingsAppBundleId` is kept and now has a name (`settingsWriteAppId`). `17fd8cee4` carries the follow-ups below. Checks rerun on the new head: `pnpm test:unit` 10,929 passed, `pnpm check:affected --run` all runnable checks passed, `check:fallow` clean (it excludes two findings inherited from `f58daa2ce` in `conformance-normalize.ts`).

**Why the read keeps its own cell.** The refusal is only half the reason. The two legs answer different questions, and the leaves disagree *inside one platform*: `appleSettingsReadFact` returns available for iPhone/iPad simulators and unavailable for tvOS/visionOS simulators, watchOS, attached devices and the macOS host — all of which the daemon already routes as one `platform`. A daemon gate would have to re-derive `device.kind` and the Apple OS to draw that line, which is the leaf knowledge ADR 0019 moved into facts. Producers state both cells today (`settingsRuntimeOperationFacts({ setSetting, readSetting })`): Android and limrun's Android available, webdriver/HarmonyOS/Vega/Linux/web unavailable, Apple narrowed per leaf. A gate replaces nine local declarations with a platform×leaf table the daemon must remember to edit, and it fails open — an unlisted provider would be admitted for a read nobody implemented and fail inside the owner.

Admission also binds an owner per operation. On one cell, a read binds the mutation runtime and travels in `setSetting`'s input, which carries a `state` the read has no value for while the answer is a `ReadSettingResult`. So either that input sometimes means "observe", or the interactor member exists anyway — and a member without a cell is unreachable, which is the shape that hides a leaf able to write but not to read.

The part of the smaller option that is right: where capability is per-setting on a target that does serve the operation, no cell can say it, and that is the daemon gate. macOS serves settings writes for `appearance` and `permission` but not `text-size`. That gate is now one helper both legs consult, which is also the fix for the asymmetry you saw: on the host, `settings text-size` and `settings text-size large` both answer `INVALID_ARGS: Unsupported macOS setting: text-size` (verified live).

The other three: the `resolveSettingsReadRequest` doc now says the daemon parser applies the rule while the registry keeps a copy pinned against `READABLE_SETTINGS`; the parity test derives its subcommand list from `SETTINGS_USAGE_OVERRIDE` instead of a hand-maintained copy; and the Validation section names the webdriver refusal and limrun, whose Android leg admits the read through the same `createAndroidInteractor` but was never driven through that provider.

thymikee:
Reviewed 17fd8ce. The rebase keeps Maestro's `settingsAppBundleId` precedence under `settingsWriteAppId`, both macOS legs now refuse `text-size` through one helper, and I found no blocking issue. Your answer on why the read keeps its own cell settles the design question from the last review.

Not blocking: `settingsRequestRefusal` now runs before `admitRuntimeUse` on the write leg ([snapshot-settings.ts#L255](https://github.com/callstack/agent-device/blob/17fd8cee4efbdba6f66dda7a2a7f10a6169b3a86/src/daemon/handlers/snapshot-settings.ts#L255)), so a sessionless macOS write of an unsupported setting such as `wifi` now returns `INVALID_ARGS` before the admission error it used to show first; one line in the PR body would cover it.

iOS Smoke Tests failed at the keyboard-dismiss step with `UNSUPPORTED_OPERATION` ("no dismiss key"). This looks unrelated: the diff does not touch keyboard code, the smoke script runs no settings step, and the same job passed on main at 45e4c59. Please rerun it.

thymikee:
Second round, `17fd8cee4..f58ef0545`, after an adversarial pass over the design rather than the diff. Five commits, each independently green.

**A live defect.** The Apple write fact is one claim covering every simulator, so `settings text-size large` aimed at a tvOS or visionOS simulator passed admission, expired the session ref frame at the mutation seam, emitted its diagnostic, and only then met the owner's leaf guard. A refused write took down a frame no device change had made stale, and that ordering is the stub-throws-after-binding shape ADR 0019 rules out. It now refuses in the shared per-setting gate both legs consult, ahead of admission. Tests pin both directions: the admitted iPhone write still expires an active frame, the refused tvOS request leaves it `active` and never binds. No tvOS simulator is installed here, so that half is pinned and not device-verified.

**Three copies of one leaf rule became one.** `isHandheldAppleSimulator` (kernel, pure taxonomy) is now read by the Apple read fact, the owner's guard, the daemon gate and the daemon fixture; `APPLE_TEXT_SIZE_LEAF_REFUSAL` is the one message and typed reason. The read-cell hint no longer claims every setting is readable on that leaf, and it never implied text-size was the only one.

**The leg rule is resolved once.** Measured with the eager-closure probe rather than assumed: the registry closure is 72 modules, `contracts/settings.ts` is absent from it and `platform-runtime-operations.ts` is inside it. The rule lives there as `resolveSettingsRuntimePlan`, the descriptor and the daemon both read its `kind`/`use`, and the duplicated set, the parity pin and three comments about a copy are gone. The plan also narrows the name, so `ReadSettingInput.setting` and `Interactor.readSetting` carry `ReadableSetting` and each owner's read leg is a map `satisfies Record<ReadableSetting, …>` — a readable name with no owner answering it is a compile error, and the two runtime default refusals it made unreachable are deleted.

Two consequences worth naming, because both were caught by gates and shaped the result: a *new* data-only vocabulary module costs the hub the same +1 module the settings import cost, so the list stays in `settings.ts` as a type with its value where the hub evaluates it, pinned equal in both directions by the compiler; and importing `platform-runtime-operations.ts` from `settings-runtime.ts`/`interactor-types.ts` closes a contracts type cycle through its forty operation imports, which R10 counted at 41 files against a merge-base of none.

**Response text and payload.** `ReadSettingResult` is keyed by `setting`, and `describeSettingRead`/`describeSettingWrite` compose both sentences from the vocabulary — the write leg's `Text size set to …` was hardcoded in the handler too. Nothing with this shape has shipped (`git tag --contains 673a142cb` is empty), so the key is free now and a wire-format change later.

**Deferred on purpose:** per-setting runtime facts (a device-parameterized support map would need `RuntimeOperationFact`, the gateway, admission and the operation-keyed managed-owner allowlist to change shape, and today only two leaves can read at all), and a `SettingsCatalog` that derives the CLI reader — the layering rule keeps CLI readers out of contracts, and ADR 0012 plus the case asymmetry between the CLI's exact match and the daemon's lowercasing mean a derived reader must preserve recorded argv aliases before it can replace the `if`-chain. Appearance read, where the leaf set and the payload shape both differ, is the change that should decide both.

Limrun's Android read cell is now pinned beside the write cell it shares.

thymikee:
Reviewed f58ef05, the delta since 17fd8ce. The tvOS/visionOS text-size refusal now runs in the shared gate before admission, so a refused write leaves the ref frame active, and `resolveSettingsRuntimePlan` is now the one leg rule that the registry and the daemon both read. I found no blocking issue. The tvOS/visionOS refusal is covered by the daemon handler test only, as the PR body says.

Not blocking: `READABLE_SETTINGS` is a separate list from the `Record` below it ([platform-runtime-operations.ts#L747](https://github.com/callstack/agent-device/blob/f58ef05454d8472d230130be01a7bf31a13953f1/packages/contracts/src/platform-runtime-operations.ts#L747)), so a readable name added to the type and the `Record` but not to the array still compiles and routes to the write leg. Deriving the array from the `Record` would make the "both directions" comment true.

The failing Smoke Tests step, "Verify clean-installed Simulator snapshot bridge preparation", looks unrelated. Its skip guard printed "no merge base" on the shallow fetch, so the step ran although this PR touches none of its guarded paths, and the call that timed out after 90 s was the packaged `devices --json` discovery call. Please rerun it.

PR #2732, opened 2026-09-21
chore(worktrees): stop copying .env into agent worktrees

### LINKED ISSUE #2530: chore(worktrees): stop copying .env into agent worktrees
## Hygiene

`.worktreeinclude:2-3` copies `.env` and `.env.*` into every agent worktree; 32 copies exist on the maintainer host. No repo code reads `.env` (no dotenv, no `--env-file`); the operator recipe sources it manually (`set -a && . ./.env && set +a`). The copies are on one laptop under one user, so this is hygiene, not an exposure: the risk is an agent tarring or `cat`-ing a worktree into a transcript.

## Fix

Delete the two lines. Update the operator recipe (memory/docs) to source the main checkout's `.env` by absolute path. No guard script, no rotation unless a copy is known to have left the machine.

Land last: it touches `.worktreeinclude` after the config-drift issue does.

thymikee:
## Summary

Stop copying `.env` and `.env.*` into agent worktrees. Nothing in this repo reads the file — no dotenv dependency, no `--env-file` flag — so every copy existed only to be present, and a worktree is exactly the directory an agent tars, `cat`s, or snapshots into a transcript. Hygiene, not exposure: the copies live on one laptop under one user.

`.worktreeinclude` keeps a short reason where the two entries were, so the row is not re-added by someone who reads it without the issue. `AGENTS.md` records the operator recipe in the worktree traps it already owns: source the main checkout's copy by absolute path.

No guard script, no rotation, per the issue. #2521 landed first, so the ordering the umbrella asked for holds.

Closes #2530

## Validation

Two non-source files: `.worktreeinclude` (3 changed lines) and `AGENTS.md` (3 added). No runtime path reads either, so no device or runtime validation applies.

`pnpm check:affected --run` at `af177f40ff`: **all runnable checks passed** (122 gate invocations), including `agent-guidance`, which budgets AGENTS.md ownership and context. `.worktreeinclude` has no derivable owner, so selection fail-opens to the full 59-check set rather than a narrowed run; the GitHub-authoritative checks run in CI on this head.

One diagnosis worth recording: the first full run failed `tmpdir-leaks-model` (`swift-toolchain-tmpdir.test.ts:113`, `kill ESRCH`). An earlier gate run I terminated left a straggler test child holding a run directory, and the descendant-kill assertion raced it. The check passes in isolation (24/24) and in the clean rerun above — contention on this host, not this change.

CI on this head: all checks green (Typecheck & Package, Lint & Format, Tests, Integration, Coverage, Repo Guards, Compatibility & Provenance, Bundle Size, Ownership and context budgets, and all four Smoke Tests lanes).


thymikee:
Reviewed af177f4. No tracked code, script or config loads `.env`, so removing it from `.worktreeinclude` breaks nothing, and the AGENTS.md note matches the change. Smoke Tests does not exercise `.env` or the worktree include list, so its result does not bear on this change.

PR #2733, opened 2026-09-21
fix(apple): harden Device Hub window discovery across host displays

thymikee:
## Summary

Follow-up to #2724: harden `fold` discovery after a Device Hub window on another monitor was reported missing.

- Search all matching Device Hub processes within a shared discovery budget. Commit to a window only after finding the requested UDID row; unrelated windows cannot prevent trying other candidates or reopening empty instances.
- Preserve AX read failures separately from successful empty reads. Record process, window bounds, and intersecting host display IDs.
- Reveal sidebars through each window's own toolbar button, replacing application-global menus. Bound traversal and cleanup; reuse the proven row with refreshed selection timeouts.
- Independent adversarial review identified and re-reviewed these corrections; no remaining blocking findings. Scope: 11 files in discovery, diagnostics, tests, and docs. Public fold output is unchanged.

## Validation

Tested commit: `9c5740b055bd4228a365f9c4ea87116261cfff04`.

- `pnpm check:affected --run`: passed.
- Native release build and 16 Swift tests passed. Restricting identification to the first candidate caused three regression assertions to fail; restored code passed.
- Live Duo/Xcode 27.1 through `setAppleFoldPose`: half-open 130°, open 180°.
- Live hidden-sidebar check proved the UDID row was absent, helper recovery/selection succeeded, and the row became hidden again afterward. Restored the original visible sidebar; Duo remains open.

CI pending on the updated head. Secondary-monitor geometry has unit coverage, but live secondary-monitor and simultaneous-Xcode scenarios remain unverified. Original monitor-incident cause remains unconfirmed. Duo coverage stays local-only; sidebar cleanup is bounded best-effort.


thymikee:
Reviewed 9c5740b. I found no blocking code defect. The open items are live evidence for the secondary-display route and one design question.

The live Duo runs cover the main display and the hidden-sidebar case, and the PR body says the cause of the monitor incident is still unconfirmed. The display-related code here only adds diagnostics. Can you run `fold open` and `fold closed` on the Duo with the Device Hub window fully on a secondary display, once with the sidebar visible and once hidden? The `apple_fold_pose_pressed` diagnostic should show only the secondary display's ID, the verified hinge angle, and the sidebar restored. That is the route the title names, and today it has unit coverage only.

A design question before this goes further: are a public generic `DeviceHubWindowInventory` type, a failure-reason enum and a new diagnostics payload needed for this fix? AX windows do not depend on the display, so could the display IDs stay in the error details only, with a per-candidate outcome kept in the existing inventory type? If not, what would have to change first?

Two questions on the helper, not blocking. If the sidebar press's 0.1 s reply times out but the press still applied, does `setDeviceHubSidebar` return before its cleanup `defer` is registered ([DeviceHubSidebar.swift#L22](https://github.com/callstack/agent-device/blob/9c5740b055bd4228a365f9c4ea87116261cfff04/apple/macos-helper/Sources/AgentDeviceMacOSHelper/DeviceHubSidebar.swift#L22))? `selectDevice` has the same pattern with its 0.25 s timeout. Also, `deviceMissing` ([DeviceHubWindowDiscovery.swift#L82](https://github.com/callstack/agent-device/blob/9c5740b055bd4228a365f9c4ea87116261cfff04/apple/macos-helper/Sources/AgentDeviceMacOSHelper/DeviceHubWindowDiscovery.swift#L82)) is reported for a missing Show Sidebar button, an exhausted budget and an early loop exit, not only for an absent row. The Swift tests cover the inventory and geometry helpers, but not the budget, reopen and reason decisions in the production loop.

Smoke Tests is still running; it does not run the Device Hub fold route.

thymikee:
Addressed the design and helper findings in 91062451e, rebased onto main (including the settled-pose and native-panel-coordinate changes).

- **Design:** the inventory is now concrete and package-private; the generic public interface and separate failure-reason enum are gone. Candidate outcomes stay on that inventory. Removed display geometry and all added success diagnostic fields; active host display IDs remain in discovery errors only. No prerequisite architecture change was needed. `apple_fold_pose_pressed` retains its existing shape, so future monitor evidence should pair a read-only window/display probe with the verified fold result.
- **Uncertain replies:** cleanup is registered before the sidebar press. A `cannotComplete` reply proceeds to bounded row observation, with cleanup still armed. Device selection similarly verifies `AXSelected` and the window title after an uncertain reply instead of treating the write acknowledgement as the outcome.
- **Reasons:** unknown identity is no longer reported as a missing device. Candidate outcomes distinguish sidebar unavailable, row unconfirmed, and timeout; shared budget interruption has its own reason. Unrelated reopen failures retain process attribution without masking a visible candidate’s uncertainty.
- **Tests:** production inventory tests cover reopen eligibility/once-only behavior, exhausted budgets, reason precedence, and candidate outcomes. Mutating the reason and budget guards produced failures; restored tests passed. These do not claim full AX-loop or timeout fault-injection coverage.

The requested **secondary-display run is still pending**. The host currently exposes only display 1; I asked for the external display to be connected. On that available display, the final helper passed closed/open with both visible and hidden sidebars, with row visibility and restoration checked, and rejected a wrong UDID. The PR title/body now describe discovery and sidebar recovery rather than imply proven secondary-monitor repair. I have not claimed that the original monitor incident is resolved.

PR #2734, opened 2026-09-21
fix(fold): require a settled hinge before reporting half-open

### LINKED ISSUE #2730: fix(fold): distinguish an observed half-open angle from a settled pose
Part of #2725.

## Decision: half-open success requires settled readings
`fold half-open` succeeds only after two consecutive readings both classify as half-open and differ by at most the existing 0.5-degree tolerance. A final half-open angle without that evidence is an unsettled failure, even though the requested angle category was observed. Open/closed endpoints retain their existing success rule.

This intentionally supersedes ADR 0025's budget-end exception. The prior rule confuses an observed category with a completed pose change. Do not hardcode the Book preset's measured 130 degrees: stability, not one device-specific angle, is the success condition.

## Exact failure contract
When the final sample is half-open but never settled, return `COMMAND_FAILED` with `details.reason: "fold-pose-unsettled"`, requestedPose, observedPose, hingeAngleDegrees, and the previous reading when available. Say that the hinge was observed half-open but did not settle within the sampling budget. Do not say it failed to reach half-open.

If the final sample is another pose, retain `fold-pose-unverified` and its observed pose/angle. Cancellation must propagate. Do not increase timeout or attempt count.

## Bounded implementation
Change `awaitHingePose()` in `packages/platform-apple/src/foldable/pose.ts` and its colocated tests. Remove the unconditional half-open budget-end success. Update ADR 0025 and owning command docs to match. No screenshot, display-dimension, or gesture changes.

## Acceptance and regression proof
- 175.1 -> 130 -> 130 succeeds.
- 170 -> 150 -> 120 -> 90 fails with `fold-pose-unsettled`, observedPose half-open, hingeAngleDegrees 90.
- A final open/closed sample for a half-open request fails with the existing wrong-pose reason.
- Two consecutive samples must both belong to the requested category; numerical proximity across a category boundary is insufficient.
- Open/closed endpoint checks and cancellation remain covered.
- Observe the regression fail before the fix, pass afterward; run focused tests and affected gates; verify the Book preset locally on the exact PR head and clean up its session.

Ready for implementation; no sibling prerequisite. The sequence above is an existing mocked counterexample, not a claim that this session observed a moving hinge live. Follow AGENTS.md and keep one issue per PR.

thymikee:
## Summary

`fold half-open` reported the pose whenever its four-read budget happened to end on a half-open angle. The interval is open, so a hinge travelling between the end stops passes through it on every fold: an observed *category* was being returned as a settled *pose*.

`half-open` now succeeds only on two consecutive readings that both classify as half-open and differ by at most the existing 0.5°. `closed` and `open` keep the end-stop rule; attempt count, per-read timeout, and cancellation are unchanged. A budget ending on an unsettled half-open angle fails with `details.reason: "fold-pose-unsettled"` plus `requestedPose`, `observedPose`, `hingeAngleDegrees`, `previousHingeAngleDegrees`, saying the hinge was observed half-open and did not settle; a budget ending on another pose keeps `fold-pose-unverified`. Supersedes ADR 0025's budget-end exception, recorded there as an amendment.

Six files: implementation, colocated tests, ADR 0025, versioned help, user docs, CHANGELOG. Closes #2730.

## Validation

`pnpm check:affected --run` passed on `4d3654a8b`; `pose.test.ts` 11/11.

Regression evidence: against the pre-fix `pose.ts`, the two new cases fail (`promise resolved "{ pose: 'half-open', ... }" instead of rejecting`) and pass after the fix. Mutations killed: dropping the previous-reading category check, restoring the budget-end success, routing `closed` through the settle rule, removing the loop's abort guard.

Live on an iPhone Duo (iOS 27.1) at this head: `fold open` reported 180°; `fold half-open` from that open pose reported 130° naming `LCD-1` at `669x951pt`; a manual `devicectl device motion hinge-angle` stream then read `Angle:130,0° Velocity:+0,0°/s`, so the reported angle is where the hinge rests. The `fold-pose-unsettled` path itself is proven only by the mocked counterexample (`170 → 150 → 120 → 90`) — it cannot be forced on hardware on demand. Manual session closed, daemon stopped, no stale device claims. Provider integration and coverage run on CI.


thymikee:
Reviewed 4d3654a. `half-open` now needs two consecutive half-open readings within 0.5°, `closed` and `open` keep the end-stop rule, and the attempt count, per-read timeout and cancellation are unchanged. I found no issue, all checks pass, and there are no conflicts. The `fold-pose-unsettled` failure is covered by the mocked test only; the live Duo run covers the settled path.

PR #2735, opened 2026-09-21
fix(ios-runner): decode targetActivation prior states in the SDK's order

thymikee:
## Summary

`targetActivation.priorState` named the wrong state. The decoder translated the runner's raw `XCUIApplicationState` integer from a handwritten table that had `RunningBackgroundSuspended = 2` and `RunningBackground = 3` reversed, so a suspended app was disclosed as merely backgrounded — shipped in v0.21.7 and v0.21.8. The runner stamps `state.rawValue`, so only this translation was wrong: `priorState: 2` now decodes to `runningBackgroundSuspended` and `3` to `runningBackground`. Wire shape, raw 0/1, and the unmapped-value `unknown` with its request-log note keep their behavior.

The claim that no public SDK header declares these values is gone — `XCUIAutomation/XCUIApplication.h` does. `RunnerTests+ApplicationStateRawValueTests.swift` pins each integer against `XCUIApplication.State` itself, guarded the way the SDK guards the suspended case, and the decoder test reads those pins back, so the two handwritten copies cannot drift unnoticed. The manual-evidence checklist's log pairing now matches the corrected decoder.

## Validation

Head `d634a83bb`, rebased on `08d977f14`; the only conflict was `CHANGELOG.md`, resolved by keeping both entries. `git diff e6b0c380d HEAD -- apple/` is empty, so the native evidence below applies unchanged. 8 files, +211/−24; a second commit lists the new pin for the iOS PR lane.

- `pnpm check:affected --run` on that head: all runnable checks passed; `check:xctest-selection` 260 methods, 0 dark; `format:check` clean.
- Regression seen first: the corrected tests failed with `raw 2: decoder says runningBackground, SDK pins runningBackgroundSuspended`, then passed.
- Native pin green on the macOS host lane at each head, and on an iPhone 18 Pro (iOS 27.0) simulator at `e6b0c380d` (`build-xcuitest-apple.sh` + `xcodebuild test-without-building`, Xcode 27.0 beta 6); red when a pinned integer changed, or a pin was commented out, renamed away from `test…`, or rewrapped.
- Risk: the macOS build compiles the suspended case out, so only simulator lanes pin it — now listed in `ios.yml`, not nightly-only.


thymikee:
Reviewed a1a4595. The decode table now matches `XCUIApplication.State`, and `priorState` only feeds the warning text, so no other behavior changes. I found no issue, and there are no conflicts. The iOS fixture Smoke Tests job timed out on its first `wait` for the fixture app's home screen; this diff only changes warning text, so the failure looks unrelated. Please rerun it.

PR #2736, opened 2026-09-21
fix(ios): serve a simulator recording whose recorder died with its daemon

thymikee:
## Summary

A retried `record stop` for an iOS simulator recording whose daemon died failed forever with `Screen recording is missing` (`reason: "resource-missing"`) while the video sat on disk: `reattachAppleRecording` answered `missing` once the `simctl recordVideo` process was gone.

Reattach now asks what a resumed stop would still have to read, through `readStopCheckpoints(envelope.metadata)`: a journaled finalization is replayed, a journaled copy is finalized again, and only a stop that journaled neither needs the recorder's file. The coordinates are the caller-facing keys of the live snapshot — a `Pick` of it, encoded and restored by spread — validated by a capture-kit `recording-facts` validator Android's descriptor now shares. A recovered handle refuses an overlay whose gesture events died with the daemon. Nothing reachable is reported as a loss; a manifest that cannot name an export is answered as before. No command, flag, error code or wire shape is added.

Part of #2281: the `needs-info` evidence request and the 90s `record` envelope are untouched.

## Validation

Head `e8dd93c`. `pnpm check:affected --run` passes; coverage, provider-integration and Apple device lanes stay GitHub-authoritative.

`recovery.test.ts` adds eight cases, including a journaled finalization that completes without re-running the finalizer, and an unreadable copy answered `missing`.

Live on iPhone 17 Pro with a 4s stop envelope over temporary reverted pauses: pausing after `nativePath` was retired left the manifest `open`/`completing` with `collectedPath`, and the retried stop returned `recorder: "confirmed"`, `nativePathDisposition: "retired"` and a playable clip carrying `overlayWarning`; pausing after the finalization journal replayed it and closed the record `completed`. Both failed with `resource-missing` before.

Risk kept: daemon-start claim reconciliation still settles an owed export through `forceCleanup`, as Android does.


thymikee:
Reviewed bfd1669. This fixes the retried stop when the recorder died with its daemon and the MP4 is still on disk, but one window is still open.

The reattach gate decides only from `simulatorRecordingRemains(descriptor.outputPath)` ([recovery.ts#L326](https://github.com/callstack/agent-device/blob/bfd1669b16466fd4247e3eae4b9c606f7d98bdf4/packages/platform-apple/src/recording/recovery.ts#L326)) and ignores the stop checkpoints in `envelope.metadata`. `finalizeAppleRecordingFromCollected` removes `nativePath` as its last step. After that, the stop sequence journals `finalization`, discards the copy, and `transitionFinishOutcome` closes the record. If the daemon dies in that span, a retried stop sees no recorder and no native file, answers `missing`, and `finishRecovered` throws `resource-missing`. That is the same symptom this PR fixes, although the manifest already holds `finalization` or `collectedPath` and the export is complete at `outPath`. Could reattach also accept a journaled finalization or a collected copy that passes the sniff, through `readStopCheckpoints(envelope.metadata)`? Please add tests for a progress with `finalization` and no native file, and for one with `collectedPath` and no native file.

A design question: 383 production lines is a lot for coordinates that are a `Pick` of `ScreenRecordingLiveSnapshot`. Could encode, decode and restore collapse to a spread plus one validator? Android's `isValidAndroidRecordingDescriptor` already checks the same scope, showTouches, recordOnlySession, activeSessionApp and exportQuality fields, so could one capture-kit validator serve both? Could the restore callback be a module-level handle factory in runtime.ts instead? Sharing the validator would first need Android's checks moved into capture-kit.

Not blocking: `reattachedSimulatorSnapshot` rebuilds what `snapshot()` and `simulatorExportCoordinates` already write, `encodeRecordingAppIdentity` duplicates capture-kit's private `encodeAppIdentity`, and the recovered handle finalizes with `showTouches: true` but no gesture events, so it runs an empty overlay pass before it reports the overlay as unavailable.

All checks pass, and there are no conflicts. After the checkpoint fix, a live retry after the native file was removed (stop the daemon after finalize removes `nativePath`) should return a completion, not `resource-missing`.

thymikee:
Rebased onto `main` and pushed `2b6ba36` (+ a `chore(gates)` commit for the boundary enumeration).

**The checkpoint window is closed.** `reattachSimulatorRecording` now asks what a resumed stop would still have to read, through `readStopCheckpoints(envelope.metadata)` (`resumedExportSource`): a journaled finalization needs no file at all, a journaled copy is sniffed as the copy, and only a stop that journaled neither reads the recorder's path. No fallback — reading a different file than the sequence will read would promise an export the retry cannot produce.

Three tests in `recovery.test.ts`, each passing the journaled metadata as `progress.learned` the way `createStopProgress` does:

- finalized export, native file gone → completes, and the finalizer double is **not** called;
- collected copy, native file gone → completes, and the only `copy` call is `from: collectedPath`;
- journaled copy that cannot be read → `missing` instead of a fresh collect.

Live evidence, two windows, with the pauses in a temporary reverted patch and a 4s `record stop` envelope:

- pause after `finalizeAppleRecordingFromCollected` retires `nativePath`: manifest left `open` / `phase: completing` with `collectedPath` and no native file; retried `record stop` returned `recorder: "confirmed"`, `nativePathDisposition: "retired"`, `overlayWarning` and a playable `clip.mp4`.
- pause inside `discard`, after the finalization journal: manifest `open` with `stopFinalization`; retried stop replayed it, closed the record `completed`, and discarded the copy.

**Your design question, answered by deletion.** The facet is now `Pick<ScreenRecordingLiveSnapshot, ...>`; encode and restore are spreads; decode is one predicate. The validator moved to `capture-kit/recording-facts.ts` and Android delegates in `descriptorRecordingIsValid`, `manifestRecordingIsValid` and `completionRecordingIsValid` (`manifest-validation.ts` −38 lines; android suites green — the shared rule requires a non-empty `bundleId`/`name`, which nothing was writing). `restoreSimulatorExport` is gone: recovery answers `restore-export` and runtime's module-level `simulatorExportHandle` builds the handle.

Nits: `reattachedSimulatorSnapshot` and `encodeRecordingAppIdentity` deleted; `finalizeAppleRecordingFromCollected` takes `overlayUnavailable`, so the recovered handle refuses the overlay on the way in (and stops claiming `showTouches: true` to the finalizer) and gets the same `overlay unavailable: <reason>` disclosure `invalidatedReason` uses.

`pnpm check:affected --run` passes at `e8dd93c`.

thymikee:
One check is red on `e8dd93c`, and it is not this change: `Smoke Tests` → *Run fixture-backed iOS simulator E2E smoke* fails in `smoke:automation-input` at `wait for Agent Device Tester` with `reason: "wait_capture_stalled"`, `readableCaptures: 0`.

The latest `main` push ([run 35643318229](https://github.com/callstack/agent-device/actions/runs/35643318229), the `targetActivation` decode fix this branch is rebased onto) fails that same job with the same assertion, scenario and reason. Nothing here touches snapshot capture; the Android smoke job that failed first was a corrupt SDK download (`Error on ZipFile unknown archive`) and passed on rerun.

Everything else is green at `e8dd93c`: `Repo Guards`, `Coverage`, `Integration Tests`, `Typecheck & Package`, `Lint & Format`, `Bundle Size`, CodeQL, and the other three `Smoke Tests` jobs.

thymikee:
Reviewed e8dd93c. The recovery, runtime and completion changes look correct and close the finding from bfd1669: reattach now reads the journaled stop checkpoints, and the live run in the PR body covers the window after `nativePath` was retired.

Not blocking: `isServedCompletion` ([screen-recording-stop-recovery.ts#L176-L186](https://github.com/callstack/agent-device/blob/e8dd93c4e948395a5790d0796535e49ad7bb65af/packages/capture-kit/src/capture-admission/screen-recording-stop-recovery.ts#L176-L186)) repeats what `recordingFactsAreValid` checks ([recording-facts.ts#L48](https://github.com/callstack/agent-device/blob/e8dd93c4e948395a5790d0796535e49ad7bb65af/packages/capture-kit/src/recording/recording-facts.ts#L48)), and the recovery tests hand-write journal keys instead of building them with `writeStopCheckpoint`.

CI: iOS Smoke fails at "wait for Agent Device Tester" with `wait_capture_stalled`. Main at a798792 fails the same job at the same step, and this change is not on the wait or snapshot route, so it looks unrelated. No conflicts.

Issue #2725, opened 2026-09-21
iOS reliability follow-ups from v0.21.5–v0.21.8 review


thymikee:
## Purpose
Track bounded follow-ups from the v0.21.5–v0.21.8 iOS source review and the measured iPhone Duo investigation in #2724. This is coordination, not a single implementation PR. Existing input bugs #2711 and #2719 remain linked to #2724; #2694 remains the separate disclosure-routing task.

## Execution contract
Each child is deliberately sized for Luna: concrete entry files, one outcome, explicit exclusions, acceptance tests, and a stopping condition. Work one issue per PR. Read blockers before starting; needs-triage children are not AFK-ready. Do not implement an unresolved API/behavior choice by guessing. Investigation success means evidence and a decision, not speculative production code.

## Child tasks
- [x] #2726 — fix(ios): correct targetActivation background-state decoding
- [x] #2727 — investigate(ios): establish a measured screen-capture source for the active window
- [ ] #2728 — fix(ios): use the resolved display for screenshots and visual verification
- [x] #2729 — fix(fold): make reported screen dimensions match their coordinate-space contract
- [x] #2730 — fix(fold): distinguish an observed half-open angle from a settled pose
- [x] #2731 — test(ios): automate semantic checks across Duo pose transitions

## Order and resolved decisions
The display-capture investigation #2727 is complete with measured open/half-open/closed, single-panel, and native-alert evidence. #2728 adopts resolved-window capture with upright raster encoding and actual scale; implement on #2724 or after its merge. #2729 retains released dimensions and marks them `coordinateSpace: "native-panel"`. #2730 requires stable half-open readings and distinguishes unsettled failure. These decisions are recorded in the child contracts.

#2731 is explicitly local-only by maintainer decision until GitHub Actions supports Duo. It can be implemented on #2724 and does not depend on the screenshot or metadata fixes. Automatic CI wiring is deferred, not a completion condition. All four implementation tasks are ready with their branch prerequisites explicit.

## Completion
Verify all child issue states and their implementation/evidence PRs, not just this checklist. Runtime changes require live semantic evidence; mocked unit tests and successful synthesis acknowledgements are insufficient. Keep the existing centralized response decoding, typed lifecycle/readiness work, and off-deadline bridge preparation; no broad rewrite is requested.

PR #2738, opened 2026-09-21
test(examples): manual iPhone Duo pose-transition semantic checks

### LINKED ISSUE #2731: test(ios): add local semantic checks across Duo pose transitions
Part of #2725.

## Decision: local-only Duo regression coverage for now
The maintainer explicitly approved keeping Duo coverage local until GitHub Actions supports it. Deliver a repeatable local scenario and evidence; automatic PR/scheduled selection is NOT a completion condition for this issue. Clearly label it local/manual coverage, never automatic regression coverage. No new self-hosted runner or hosted-toolchain installation is requested.

## Investigation of CI availability (2026-09-21)
- Current `.github/workflows/ios.yml`: `macos-26`, iOS 26.2, iPhone 17 Pro.
- Repository Actions runners API returned `total_count: 0` registered self-hosted runners.
- Official macos-26 image 20260907.0351.1 lists Xcode 26.x.
- Official xcode-27 preview image 20260912.0186.1 lists Xcode/iOS 27.0 and no Duo. Merely switching the label does not provision iOS 27.1/Duo.
- Sources: https://github.com/actions/runner-images/blob/main/images/macos/macos-26-arm64-Readme.md and https://github.com/actions/runner-images/blob/main/images/macos/xcode-27-arm64-Readme.md . Recheck these before future automatic wiring; this is a dated inventory, not a permanent limitation.

## Dependencies and scope
Prerequisite: PR #2724 must be merged, or explicitly base the test PR on its branch. Do not duplicate its input fix. #2728/#2729/#2730 are not prerequisites for a semantic interaction replay: do not assert their new metadata or forced-fallback behavior here.

Reuse the fixture app and existing replay/test harness; no new framework or CI/gate registration. Pin `DEVELOPER_DIR` to a Duo-capable Xcode, accept an explicit UDID/session, and document the runtime and Device Hub Accessibility prerequisites. If the fold helper reports `device-hub-window-missing`, report actionable setup failure rather than skipping. The investigation observed this with a real window on a second host display; moving it to the main display restored pose control. Include that workaround without claiming the window is absent.

## Required scenario
1. Verify the correct fixture surface and select a local test session.
2. Exercise closed -> half-open -> open -> closed. After each fold, capture fresh state, press a visible selector/ref, and assert the selected tab or another activation-dependent app counter changed.
3. On open, scroll and assert a newly visible canary; long-press and assert its dedicated count increments.
4. Run one multipointer pinch canary and assert recognized scale change, not onFinalize or transport success. Derive its coordinates from the current target bounds rather than portrait constants. Reuse the existing gesture-lab fixture.
5. Never reuse refs across pose changes or mutations. Save failure artifacts, clean up the session, and restore the starting pose where safely possible.

## Acceptance
- One documented local command reproduces the scenario using the existing harness.
- Publish the tested commit, toolchain/runtime, command, semantic results, and remote-accessible result summary in the PR.
- Demonstrate a wrong-display or geometry mutation makes a semantic assertion fail, then restore and pass.
- Run relevant harness tests and affected gates. Do not require an automatic Duo job or weaken the ordinary iPhone CI lane.
- Record future GHA enablement as deferred work, with the admission requirements: Duo runtime/device type, working Device Hub pose controls/permissions, and an actual successful lane run.

Implementation can start on #2724's branch with its PR base/dependency explicit. Do not merge ahead of the prerequisite. Follow AGENTS.md and manual-session hygiene.

thymikee:
## Summary

Adds a **local/manual-only** semantic interaction replay across iPhone Duo pose transitions (`examples/test-app/foldable/duo-pose-semantic.mjs`), closing the local-coverage ask in #2731.

It drives one booted iPhone Duo through **closed → half-open → open → closed** against the existing `agent-device` CLI/daemon and the Agent Device Tester fixture, asserting the app reacts after **every** pose change:

- a Catalog tap is **proved** to activate by an `absent→present` `catalog-title` transition, read from the **same snapshot that resolved the tap's ref**, with `home-title` gone after; the Home reset is **asserted** before the next fold,
- Add to cart moves the cart counter,
- a scroll reveals a canary and a long press increments its dedicated count,
- one multipointer pinch changes a recognized scale, origin derived from the **current** target bounds.

Refs/coords are re-resolved from a fresh snapshot after each fold — never reused across a pose change or mutation. Two demonstration flags prove the checks are load-bearing: `--demo-geometry-mutation` (mis-derived pinch origin) and `--demo-skip-home-reset` (skipped reset).

Local/manual only: GHA cannot schedule it (no hosted Duo runtime/runner); enablement is documented as deferred in `examples/test-app/README.md`.

Closes #2731 · Part of #2725 · Base is prerequisite #2724.

## Validation

- Tested head `aecdb46b6` (rebased onto current #2724; the earlier `RunnerTests+SynthesizedInteraction.swift` hunk now lives upstream and dropped from this PR), iPhone Duo, Xcode 27.1 Beta / iOS 27.1, iOS runner rebuilt (`TEST BUILD SUCCEEDED`).
- Clean run: **15/15 semantic checks passed**; every pose reports `catalog absent->present, home present->absent`.
- `--demo-skip-home-reset`: **10/15**, exit 1 — the closed **Home-reset** check goes red (`home=false catalog=true`) and the half-open/open **activation** checks go red (`catalog present->present`) instead of falsely passing.
- `pnpm check:affected --run` (incl. `xctest-selection`), `format:check`, `lint`, `typecheck` all pass.
- Command + admission requirements documented in `examples/test-app/README.md`. Smoke's `wait text "Automation lab"` timeout is pre-existing on the base and unrelated (this diff only adds an example + README that Smoke never runs).


thymikee:
Reviewed 6f6895a. The per-pose activation check can pass on a tap that did nothing, so this needs one change before it is ready.

`activateAndAssert` only checks that `catalog-title` is present after the Catalog tap ([duo-pose-semantic.mjs#L365](https://github.com/callstack/agent-device/blob/6f6895aff54ab633aec60af756dfa351281f3c16/examples/test-app/foldable/duo-pose-semantic.mjs#L365)). In the closed and half-open poses `selected` is `undefined`, and `selected !== false` accepts that. The Home reset tap at [L517](https://github.com/callstack/agent-device/blob/6f6895aff54ab633aec60af756dfa351281f3c16/examples/test-app/foldable/duo-pose-semantic.mjs#L517) is never asserted. So if the Home tap misses, or the Catalog tap lands off-target after a fold (the reference-frame bug #2724 fixes), `catalog-title` can still be in the tree and the check prints PASS. Can each activation check compare a before-state and an after-state, with the before-state from the same snapshot that resolved the tap's ref: `catalog-title` absent before and present after, in every pose, and the Home reset asserted before the next fold? Is it also possible that react-navigation keeps the inactive Catalog screen in the tree after switching to Home? The before-state check would catch that too.

To show the fix works, please post one clean Duo run where every pose's Catalog check shows an absent to present change, and one run where a skipped Home reset or a mis-aimed Catalog tap turns a closed or half-open check red. The 11/11 and 10/11 results in the PR body used the presence-only check.

Non-blocking notes: on `SESSION_NOT_FOUND`, `command()` re-sends the same `@ref` or coordinates after `open`, which breaks the script's own "never reuse a ref across a mutation" rule, and retrying on "may not have been performed" can fire a long press twice ([L176](https://github.com/callstack/agent-device/blob/6f6895aff54ab633aec60af756dfa351281f3c16/examples/test-app/foldable/duo-pose-semantic.mjs#L176)). Could the fold, tap and wait steps live in an `.ad` replay like `gesture-lab.ad`, with a small script only for the bounds-derived pinch origin? The Swift hunk in `RunnerTests+SynthesizedInteraction.swift` is identical to #2724's head commit f3d5387, so it drops out after a rebase onto the current base.

The Smoke failure looks unrelated. It has the same `wait text "Automation lab"` timeout as the base #2724 on every head that contains a2f0e2068, and this diff only adds an example script and README text that the Smoke route never runs.

thymikee:
Rebased onto the current #2724 (now `073b5687b`); the `RunnerTests+SynthesizedInteraction.swift` hunk is upstream and dropped from this PR. Head is `aecdb46b6`.

**Blocking (presence-only activation check) — fixed.** `activateCatalog` no longer asserts mere presence. It resolves the Catalog ref from a fresh snapshot via `actOnFresh`, which returns the *same* snapshot it resolved the ref from, and reads the before-state there: it requires `catalog-title` **absent** and `home-title` **present** before the tap. After the tap it waits for the route flip and requires `catalog-title` **present** while `home-title` is **gone** — in every pose. So a no-op tap, a mis-aimed tap, or a fold the app hadn't finished can no longer print PASS.

**Home reset — now asserted.** `resetToHomeAndAssert` runs before each subsequent fold and records a check that Home is the active route (`home` present, `catalog`/`settings` absent). A missed reset fails here, and the next pose's before-state (`home present`, `catalog absent`) would also fail, so a stale before-state can't be mistaken for a real activation.

**react-navigation persistence — checked empirically.** On this fixture the `-i` snapshot carries only the active screen: after Catalog→Home, `catalog-title` is absent; after Home→Catalog, `home-title` is absent. The absent→present + `home gone` invariant holds in all four poses, so the check does not rely on `selected` (which is unreliable: `undefined` for closed, laggy for open — I dropped it).

**Evidence (this head):**
- Clean: `15/15 semantic checks passed`, each pose printing `catalog absent->present, home present->absent`.
- `--demo-skip-home-reset`: `10/15`, exit 1 — closed **Home-reset** check `home=false catalog=true`, and half-open/open **activation** checks `catalog present->present, home absent->absent` (red, not a false PASS). This is the "skipped Home reset turns a closed/half-open check red" case.

**Non-blocking notes:**
- *`SESSION_NOT_FOUND` replaying an `@ref`:* fixed. `command()` now re-opens but only replays pose/observation commands (`open`/`fold`/`snapshot`/`close`); any interaction that carries an `@ref` or raw coordinates surfaces as transient so `actOn`/`snapshotNodes` re-resolve a fresh ref against a fresh snapshot.
- *Long-press retry double-fire:* the assertion is now a strict increment (`countAfter > countBefore`) instead of exact `+1`, so a transient retry that delivers two presses still proves registration while an unregistered press (zero) fails.
- *Fold/tap/wait into `.ad`:* I kept a single `.mjs`. The activation and count checks are host-side value assertions (before/after route-title and counter transitions, `absent→present`), the pinch origin is computed from the target's current bounds, and refs must be re-resolved per pose — none of which `.ad` can express, and splitting would duplicate the snapshot/ref state across two harnesses. Happy to revisit if a `.ad` value-assert/derive step is planned.
- *Smoke:* agreed — the `wait text "Automation lab"` timeout reproduces on base #2724 and this diff only adds an example + README the Smoke route never runs.

thymikee:
Reviewed aecdb46. The finding from 6f6895a is fixed: each pose now reads the before-state from the same snapshot that resolved the tap's ref, and the Home reset is asserted.

Not blocking: the `TRANSIENT` regex in `actOnFresh` ([duo-pose-semantic.mjs#L59](https://github.com/callstack/agent-device/blob/aecdb46b66baa0c6893959123b8e89e89035e7f1/examples/test-app/foldable/duo-pose-semantic.mjs#L59)) also matches "may not have been performed", so a retry after a tap that did land can give `activateCatalog` a before-snapshot that already shows Catalog and report a false FAIL; keeping the first attempt's snapshot as the before-state would avoid that.

CI: this change only touches the manual Duo script and its README, which Smoke does not run. Android Smoke failed on an SDK download error, and iOS Smoke fails at the same "wait for Automation lab" step as base #2724, so both look unrelated. The 15/15 and 10/15 runs are the results reported in the PR body. This PR lands after #2724.

thymikee:
Rebased onto #2724 `800151395`; head is now `cb6084784`.

**`TRANSIENT` / retry false-FAIL — fixed.** `actOnFresh` now also returns `firstNodes`, the *first* snapshot that resolved the target, before any tap attempt. `activateCatalog` reads its before-state (`catalog-title` absent, `home-title` present) from `firstNodes`, so a retried tap — including one that landed but reported `may not have been performed` — can no longer hand the assertion a post-tap snapshot that already shows Catalog. The post-action state is still read fresh after the action via `waitForActiveRoute`. Re-ran clean: `15/15` with every pose still printing `catalog absent->present, home present->absent`; `--demo-skip-home-reset` still fails the closed Home-reset and the half-open/open activation checks.

Agreed the CI Smoke failures are unrelated (this PR only touches the manual Duo script + README, which Smoke never runs; Android Smoke is an SDK download error and iOS Smoke is the base #2724 "wait for Automation lab" step).

thymikee:
Synced onto #2724 `b9748e264` (head now `e0e9514e9`). That commit already routes synthesized display IDs through `resolveRunnerWindow` (`RunnerResolveWindowDisplayID` + a `resolvedWindow:` parameter on the gesture entry points, with `firstMatch` reads folded onto the helper and new unit tests), so I did **not** re-implement it here — that would duplicate the upstream change. No merge conflicts on either PR.

Rebuilt the runner from `b9748e264` and re-ran the manual scenario: **15/15**, every pose still `catalog absent->present, home present->absent`; `--demo-skip-home-reset` still fails the closed Home-reset and the half-open/open activation checks. As an independent cross-check of #2724's routing on a real Duo, this harness's `runner.log` open-pose records show `displayID=3 interfaceOrientation=4` (lit inner panel) for taps / controlled-scroll / the pinch `gesture-plan`, and `displayID=1 interfaceOrientation=1` when folded closed — the resolved window's display, not main.

thymikee:
Reviewed e0e9514. The before-state now comes from the first snapshot that resolved the target, so the note from aecdb46 is fixed. I have no new findings.

CI: Smoke Tests was still running when I reviewed. No CI job runs this manual script, so that result does not reflect this change. This PR lands after #2724.

thymikee:
Re-synced onto #2724 `8cbac1592` (anchor fix + rebase onto `main`); head is `1a2ed3d64`, diff unchanged (the manual Duo script + README only). Both PRs are now mergeable and clean. No code review changes needed here — your `aecdb46` before-state finding was already addressed and `e0e9514` recorded "no new findings."

Issue #2728, opened 2026-09-21
fix(ios): use the resolved display for screenshots and visual verification


thymikee:
Part of #2725. Investigation evidence and reproducible probe: https://github.com/callstack/agent-device/issues/2727#issuecomment-5762778410.

## Resolved design / dependencies
Use the resolved app window's screen, not `XCUIScreen.main`, for iOS app-screen capture and visual verification. Resolve `app.windows.firstMatch.frame` before obtaining that window's private `screen`; screenshot the resulting `XCUIScreen`. Open, half-open, closed, a single-panel control, and a native fixture alert were measured successfully. Do not choose by a hardcoded display number or the first screen in `XCUIScreen.screens`.

Prerequisite: PR #2724; implement on its branch with an explicit PR base, or after it merges. No unresolved design decision remains. Consumer-level checks below remain required implementation validation; the spike proves the capture primitive, not the complete migration.

## Small owning interface
Introduce one iOS capture helper, conceptually `captureResolvedAppScreen(app:) -> CapturedAppScreen`, returning an upright image and actual display ID, encoded pixel width/height, and pixels-per-point. Keep the private window/screen lookup exception-safe in the native XCTest bridge. Reuse the window-resolution order established by #2724; do not maintain independently drifting display resolvers or introduce a session-wide display cache.

Normalize the image by drawing it in its oriented logical size with its actual scale before PNG encoding. Measured open Duo: window 951x669, UIImage logical size about 950.667x668.667, scale 3, raw CGImage/PNG 2006x2852, UIImage orientation right; upright drawing yields 2852x2006. Do not infer orientation from raw CGImage dimensions, rely on PNG metadata being respected by every consumer, or stretch the one-pixel XCTest discrepancy to CoreDevice's nominal 2853x2007 size.

## Wire and failure contracts
Carry screenshot-only metadata with the runner screenshot result (e.g. `screenshotMetadata: {displayID, pixelWidth, pixelHeight, pixelsPerPoint}`), alongside the existing file/base64 payload. Declare it at the owning cross-language contract, not by per-consumer casts. Host normalization uses those actual source facts, never `SIMULATOR_MAINSCREEN_SCALE` for an image from another panel.

An unresolved/invalid app window or screen fails required screenshot capture with a typed reason, without selecting main. Optional visual checks return their existing unavailable/unknown result; no sample is not evidence of stability or navigation change. Do not make an observation activate an app merely to obtain a screen; use the caller's intended app context.

## Bounded migration
[...]

PR #2739, opened 2026-09-21
docs(adr): record which runner capture surface sees the lit foldable panel

### LINKED ISSUE #2727: investigate(ios): establish a measured screen-capture source for the active window
Part of https://github.com/callstack/agent-device/issues/2725

## Purpose
Before replacing remaining `XCUIScreen.main` reads, establish which native screen-capture operation actually captures the active app window's display. PR #2724 established that resolving a window frame before reading its screen yields the correct gesture display identity. That does not yet prove screenshot behavior.

## Inputs and narrow scope
Read PR #2724, ADR 0025, `RunnerXCTestEventBridge.m`, `RunnerTests+Lifecycle.swift`, and the screenshot case in `RunnerTests+CommandExecution.swift`. This task is a bounded investigation, not a runtime migration.

Use a dedicated session and iPhone Duo/iOS 27.1-capable Xcode. Read the fixture app setup instructions. Test closed, half-open, and open; use visibly different app content to distinguish panels. Also check an ordinary single-panel iPhone. Resolve the window first, inspect its screen identity, and test the screen's screenshot operation. Include orientation, logical window size, pixel size/scale, and whether system overlays are included. Do not infer success from nonempty PNG bytes alone.

## Deliverable / stop condition
Post a remote-accessible evidence table on this issue, with tested commit, toolchain, commands, images or artifact links, and an explicit supported/unsupported verdict for each case. Propose one narrow native helper signature that returns the captured image with its actual display/geometry metadata. State how an unresolved window is handled without silently selecting main.

No permanent runtime code changes or new fallback chains in this task. Remove probes and clean up the session. If a required simulator is unavailable, record that blocker; do not substitute mocks for the measurement.

Dependency context: #2724 is open at issue creation; use its measured findings as evidence and check its latest status. Existing input bugs #2711/#2719 stay owned by that work, not this investigation.

## Working rules
Read AGENTS.md and its relevant task procedures. Keep this to one task/PR; do not absorb sibling issues. Before implementation, check current main and prerequisite PRs: the review baseline is v0.21.8 and findings may already be fixed. Do not change the active handoff worktree or use its simulator without checking ownership. Follow repository validation and manual-session cleanup requirements. Report an unavailable runtime or device as a blocker, not a passing validation.

thymikee:
## Summary

Records the #2727 measurement in ADR 0025 so #2728 migrates the runner's capture source without re-measuring it.

The resolved app window is the one surface measured that follows the fold: open, it captures the lit inner panel, while `XCUIScreen.main.screenshot()` and `XCUIApplication.screenshot()` — what the `screenshot` command's iOS branch, the recording frame source, and the navigation probe call — capture black. `XCUIScreen` does answer `displayID`, `scale`, and `bounds`, but only as keys no XCTest header declares, so CoreDevice still names the panel.

It also records rotated panel-native geometry, overlays, and XCTest's refusal on an unresolved window. The old gap becomes a migration scope of six call expressions in four files plus the host-side and user-doc text repeating them. A pose finding replaces a promise this ADR could not keep, and `device-verification.md` made the same one: Device Hub confirms selection by title, which same-named simulators defeat.

Two files, docs only, two commits (the second a re-wrap). Closes #2727

## Validation

`pnpm check:affected --run` selects no checks on either commit (head `7c2b93a28`, rebased onto main after #2734 landed); a docs-only diff takes no runtime or device validation. Evidence is a test-only runner probe on iPhone Duo (iOS 27.1) and an iPhone 17 control under Xcode 27.1 beta, since removed; method, tables, and the erratum are on #2727, and images below are resized. Residual risks: no physical foldable measured; the undeclared identity keys are a risk #2728 takes on; #2724 edits this same paragraph, so whichever PR lands second rebases onto this text; two review passes found eleven overstated claims, folded into this commit and recorded in #2727's erratum.

closed: lit outer panel, simctl oracle closed: dark inner panel capture, all black

open: XCUIScreen.main capture, all black open: resolved window capture, rotated content open: app.screenshot capture, all black


thymikee:
Two more from the same measurement run: the open-pose ground truth for the inner panel, and a closed-pose app-window capture with the software keyboard composited into it — one 1398x2034 surface, so an overlay is verified on the app panel rather than through the overlay's own window.

open-groundtruth-lcd-1

closed-window-with-keyboard

thymikee:
Reviewed 7c2b93a. The ADR's statements about the current capture call sites match `main`, and the text agrees with the #2727 evidence as reported on that issue. This is docs-only, CI is green, and there are no conflicts.

One small note: the reflow joined some `- **X.**` items onto the end of the previous bullet (for example `- **Geometry.**` near [line 303](https://github.com/callstack/agent-device/blob/7c2b93a2885ec9308d4f7f08ea1c6dd5be11bd05/docs/adr/0025-foldable-apple-panels.md#L303)), so they render as plain text instead of list items.

PR #2740, opened 2026-09-21
fix(ios): keep list rows after a selectable text row in interactive snapshots

okwasniewski:
- An interactive iOS snapshot (`snapshot -i`) of a list whose row holds selectable text dropped every row after that text. Regression since 0.21.0, where `collectIosScrollIndicatorPresentation` started deriving a scroll container's viewport from its indicator.
- A UITextView is a UIScrollView. XCTest publishes its `Vertical scroll bar, 1 page` indicator inside the `TextView`, and the rule attributed it to the nearest scroll-typed ancestor: the list. The list's viewport became the 22 px text row and `clipDescendantsToDerivedScrollViewports` suppressed everything below it.
- An indicator whose nearest scrolling ancestor is a `TextView` now belongs to that text view and derives no container. The list's own indicator, a direct child, still drives its viewport as before.

```mermaid
flowchart TD
  L[ScrollView list 0,110,402,764] --> R[Other row]
  R --> T[TextView 'Thread root' 16,180,370,22]
  T --> I1[Other 'Vertical scroll bar, 1 page' 353,180,30,22]
  L --> B[Button 'Reply (58 replies)' y=232]
  L --> I2[Other 'Vertical scroll bar, 5 pages' 369,110,30,702]
  I1 -. before: clipped L to y=180,h=22 .-> L
  I2 -. after: only this derives L's viewport .-> L
```

Bluesky post thread on an iPhone 17 Pro simulator, root post is selectable text, 58 replies:

| | `snapshot -i` nodes | replies listed |
| --- | --- | --- |
| 0.20.8 | 74 | yes |
| 0.21.8 | 20 | none, root post actions gone too |
| this PR | 67 | yes |

Full (non-interactive) snapshots were unaffected: 113 nodes on every version. The regression test in `runner-presentation.test.ts` is the thread's runner tree cut down to the list, the text row with its indicator, and the rows after it.

`pnpm check:affected --run` on 12f97ae: all runnable checks passed (353 unit files, 2156 tests, 0 skipped).


thymikee:
Reviewed 163d240. The fix looks correct: a scroll indicator inside a `TextView` now belongs to that text view, so it no longer shrinks the list's viewport, and the new regression test covers this route. The Bluesky numbers in the PR body come from your run; I traced the same mechanism through `findScrollIndicatorContainer` on the test fixture.

The new head 12f97ae is the same code rebased onto `main`, with the CHANGELOG.md conflict resolved, so there are no conflicts now. CI is still running on 12f97ae; the unit lane that covers `scroll.ts` is the one to check. Small note: the PR body has no `pnpm check:affected --run` result for this head.

okwasniewski:
Thanks. Ran `pnpm check:affected --run` on 12f97ae: all runnable checks passed, 0 skipped (353 unit files, 2156 tests, the engine suite with the new `runner-presentation.test.ts` case included). Added the line to the body.

On CI for this head, Coverage and every job but the iOS Smoke Tests are green; that one is still queued. Its earlier failure on 163d240 was the live simulator scenario missing `automation-longpress` after a scroll. That screen has no `TextView` (its one `TextInput` is single-line and the `ScrollView` hides its indicator), so the changed rule never fires there, and three other PRs failed the same live job in the same half hour at `wait for Automation lab` while main passed. I'll watch the rerun.

thymikee:
Reviewed 12f97ae. The code is the same as 163d240, which I reviewed earlier. The only change is the CHANGELOG.md conflict resolution, and it keeps the entry intact. My earlier verdict stands: the fix is correct, and your Bluesky thread run still covers this code. All CI checks pass on this head, including the iOS smoke tests, and the body now has the `pnpm check:affected --run` result. Ready for human review.

PR #2741, opened 2026-09-21
fix(ios): capture the display that hosts the app, not the main screen

thymikee:
## Summary

iOS screenshots and visual verification now capture the display that hosts the target app instead of `XCUIScreen.main`. On a foldable one panel is lit and the dark panel still answers with a valid all-black PNG, so a main-screen capture proved nothing.

The runner resolves the app window, reads the display off that same window instance, and encodes the capture upright, reporting display ID, encoded pixel size, and pixels-per-point. Density normalization reads those facts instead of applying a panel the host guessed. A capture that cannot name a display fails closed with a typed runner code (`APP_SCREEN_*`) and never falls back to the main screen. `screenshot` is a runner-lifecycle command and skips app activation, so it resolves the requested bundle id for observation without foregrounding anything.

One target-selection helper, `captureObservedScreen(app:)`, is shared by `screenshot`, `captureRunnerFrame`, keyboard settle and the navigation fallback: it asks the session app first, falls back to SpringBoard when the session app resolves no window, and fails closed only when neither resolves. `.unresolvedScreen` never triggers the fallback, so a display refusal is not masked by the system surface.

Stacked on #2724.

## Validation

Tested at `80079f743`, rebased onto the rewritten base: `pnpm check:affected --run` passed; `pnpm build:xcuitest:ios` reports `** TEST BUILD SUCCEEDED **`; the Apple runner unit lane passes 18 tests, including the three new target-selection cases; 75 touched TypeScript tests, `pnpm typecheck`, `lint`, `format`, `check:fallow` and `check:layering` pass.

The device evidence below was taken at `424b00928`. `git range-diff f3d538789..424b00928 073b5687b..HEAD` moves only this ADR, so every captured line below still describes the code at this head.

Live iPhone Duo, with a `xcrun` shim failing only `simctl io … screenshot` with the real `Timeout waiting for screen surfaces` so the runner route actually ran. One command, unchanged flags:

| Pose | Lit panel | Capture | Mean luma |
| --- | --- | --- | --- |
| Open 180° | `LCD-1` (3) | 951x669 @1x | 214 |
| Half-open 130° | `LCD-1` (3) | 951x669 @1x | 214 |
| Closed 0° | `LCD` (1) | 466x678 @1x | 177 |

The closed row is the point: the capture followed the app onto the other panel unprompted, and every row is content where a main-screen capture is black. The open row was re-run at this head; the other two are from the previous head, whose capture code is unchanged in the pose path.

Single-panel control, same forced-`simctl` route on an iPhone 17 Pro Max with Safari foregrounded: the runner capture returned 440x956 @1x and logged `AGENT_DEVICE_RUNNER_SCREEN_CAPTURE display=1 pixels=1320x2868 pixelsPerPoint=3`, so `pixelsPerPoint` equals the device scale on a device this change is not for.

Home screen, same route, session bound to `com.apple.springboard`: a real 951x669 image whose band luminance is 136.4 against 227.6 for the app screen on the same panel, logged `display=3 pixels=2852x2006 pixelsPerPoint=3`.

Fixture native alert kept in the capture: `open-automation-lab` → `Open automation alert`, then one `screenshot` before and one after. `alert get` reports `Automation confirmation`, the snapshot shows it as a separate `[window]` rather than app content, and the centre band luminance falls 227.6 → 189.1, so the overlay is in the captured pixels and not only in the element tree. Both captures logged `display=3 pixels=2852x2006 pixelsPerPoint=3`.

`runner.log` lines for the two consumers that sample rather than return an image, both on the lit panel:

```
AGENT_DEVICE_RUNNER_KEYBOARD_STABILITY samples=3 captured=3 settled=yes
AGENT_DEVICE_RUNNER_IN_APP_BACK_VISUAL_VERIFICATION beforeBytes=365716 afterBytes=365716 changed=no
```

The keyboard line came from dismissing the email field, whose `returnKeyType="done"` exposes the dismiss key the settle path waits behind; three samples, none nil. The `back` line is the no-change control: the fallback sampled 365 KB both sides and read no change, and the command then reported that no in-app back control exists.

Unresolved, called out rather than papered over: physical foldables unverified; `windows.firstMatch` picks an order-dependent window for a multi-window app; recording refuses to start with an untyped error when no window resolves; the navigation fallback turns an unresolvable display into a `back` failure; the shared resolver's new `exists` read costs one extra query per synthesized gesture; `home` does not background the app on this Duo lane, so the SpringBoard fallback branch is pinned by the runner unit test rather than by a device run.


thymikee:
Reviewed da8bf8a. There are two problems to fix before merge and one size question.

On iOS, `screenshot` now only calls `captureResolvedAppScreen(resolveAppWithoutActivation(command))` ([RunnerTests+CommandExecution.swift#L2088](https://github.com/callstack/agent-device/blob/da8bf8a8815b8fe4ec93e9729803a49d72fbe7f4/apple/runner/AgentDeviceRunner/AgentDeviceRunnerUITests/RunnerTests%2BCommandExecution.swift#L2088)). With no `appBundleId` and no current app, the target is the runner host, which your PR body shows answers `exists=NO` with a `0x0` frame. A session app that is not running or not in the foreground has no `windows.firstMatch` either. Do both cases now return `APP_SCREEN_WINDOW_UNRESOLVED` where they used to return the main screen? If so, home-screen screenshots and captures without a session app fail on every runner-routed screenshot path, and recording, keyboard settle and the navigation fallback lose their samples too. #2728 asks to fail closed when the intended app window is unresolved, not when there is no app. Could one target-selection helper, used by screenshot, `captureRunnerFrame`, keyboard settle and the navigation fallback, pick the session app when it resolves a window and SpringBoard otherwise, and fail closed only when neither resolves? A runner unit test for the no-bundle-id and not-running cases would pin it.

The new contracts `runner-screen-capture` subpath has no consumer outside platform-apple. Its static import in [runner-contract.ts#L11](https://github.com/callstack/agent-device/blob/da8bf8a8815b8fe4ec93e9729803a49d72fbe7f4/packages/platform-apple/src/runner/runner-contract.ts#L11) adds one module to the eager closure of `app-lifecycle-facade`, `doctor-facade` and `runner-operations-facade`, which is the Coverage failure (eager-closure-budgets). The PR is also at +717 net production lines. Would the smaller design do: the three `APP_SCREEN_*` codes in runner-contract.ts next to `SCROLL_KEYBOARD_OCCLUDES_SURFACE_RUNNER_CODE` (L693), `RunnerScreenCaptureMetadata` and its reader in screenshot.ts, and the subpath, its package export and its snapshot line deleted? That removes about 90 lines and clears the budget failure. If the subpath must stay, what needs it outside platform-apple? Also, does `RunnerResolveApplicationScreen` do more than forward to `RunnerResolveApplicationScreenInto`, or can the two merge?

#2728's acceptance also needs a single-panel control, a native overlay kept in the capture, navigation change/no-change controls, and keyboard settle seen on the lit display. The PR body shows the three Duo poses for `screenshot` only. Please add, at the PR head: a standard iPhone simulator with the forced `simctl` failure, where the runner capture succeeds with `pixelsPerPoint` equal to the device scale; the same route on the home screen with no session app, returning an image; a fixture native alert visible in a Duo capture; and runner.log lines from a `back` that reaches the navigation fallback and from a keyboard dismiss, both with non-empty samples from the lit panel.

Coverage is caused by this PR, through the import above. Smoke has the same `wait text "Automation lab"` timeout as the base #2724, and the merge-base already contains a2f0e2068, so it is most likely inherited. The only overlap is that `RunnerResolveApplicationDisplayID` now also reads `exists` on every synthesized gesture, before the tap that should navigate. The not-running app behavior above comes from the host-app measurement in your PR body, not from a device run on my side.

Next step: fix the no-app capture target and move the codes and metadata reader into platform-apple, or explain why the contracts subpath is needed, then add the missing #2728 evidence.

thymikee:
Rebased onto the rewritten base and pushed as `80079f743`. Range-diff against the previous head `424b00928` moves only the ADR; relative to the `da8bf8a` you reviewed, the code changes are the two fixes below.

**1. The no-app capture target.** You read it right, and both cases did fail closed before this fix: no `appBundleId` resolves the windowless runner host, and a session app that is not running has no `windows.firstMatch`, so each returned `APP_SCREEN_WINDOW_UNRESOLVED` where the old code returned the main screen. Home-screen captures lost their image and every sampling consumer lost its frames.

There is now one target-selection helper, `captureObservedScreen(app:)` in `RunnerAppScreenCapture.swift`, used by `.screenshot`, `captureRunnerFrame`, keyboard settle (which goes through `captureRunnerFrame`) and the navigation fallback. It asks the session app first, and asks SpringBoard only when the session app answers `.unresolvedWindow`. It deliberately does not ask SpringBoard on `.unresolvedScreen`: that is a resolved window whose display refuses to name itself, and routing it to the system surface would mask the refusal your `#2728` wording asks us to report. It fails closed only when neither resolves, so the rule is app-window-first rather than fail-on-no-app.

The rule is separated from the two queries as `selectObservedScreenCapture(resolving:fallingBack:)` so it is testable without a device. Three runner unit tests pin it, one per branch:

- `testObservedScreenCaptureKeepsTheSessionAppCaptureItResolved`
- `testObservedScreenCaptureAsksTheSystemSurfaceWhenTheSessionWindowIsUnresolved` — the no-bundle-id and not-running cases, both `.unresolvedWindow`
- `testObservedScreenCaptureReportsADisplayRefusalWithoutAskingTheSystemSurface` — the fallback must not swallow a display refusal

**2. The contracts subpath.** Nothing outside `platform-apple` needed it, so it is gone as you described: the three `APP_SCREEN_*` codes sit in `runner-contract.ts` beside `SCROLL_KEYBOARD_OCCLUDES_SURFACE_RUNNER_CODE`, `RunnerScreenCaptureMetadata` and its reader moved to `core/screenshot.ts`, and the subpath, its package export and its snapshot line are deleted with the module. The eager closure of the three facades loses a module and Coverage is green (`pnpm check:affected --run`). Net production size is +597/−71 across 10 files at this head, against your `+717`.

`RunnerResolveApplicationScreenInto` no longer exists. `RunnerResolveApplicationScreen` is the single function: it reads the window and its frame, then reads `screen` off that same instance — the two reads stay in that order because a window queried before its frame can still name main.

**3. The missing #2728 evidence.** All four routes, at this head's code, with `simctl io … screenshot` forced to fail by a `xcrun` shim so the runner route is the one that ran.

- Single-panel control — iPhone 17 Pro Max, Safari foregrounded: capture returned 440x956 @1x and logged `AGENT_DEVICE_RUNNER_SCREEN_CAPTURE display=1 pixels=1320x2868 pixelsPerPoint=3`. `pixelsPerPoint` equals the device scale, and the scale probe is no longer consulted on this route.
- Home screen, no app — same route with the session bound to `com.apple.springboard`: a real 951x669 image, `display=3 pixels=2852x2006 pixelsPerPoint=3`. Its centre-band luminance is 136.4 against 227.6 for the app screen on the same panel, so the capture is the home screen and not a stale app frame.
- Native overlay kept in the capture — the fixture's `Open automation alert` on the Duo. `alert get` returns `Automation confirmation`, the snapshot reports it as its own `[window]` rather than app content, and one `screenshot` either side of it moves the centre-band luminance 227.6 → 189.1. The overlay is inside the pixels, not only in the tree, and both captures logged `display=3 pixels=2852x2006 pixelsPerPoint=3`.
- Navigation change/no-change and keyboard settle — `runner.log` from the lit panel:

```
AGENT_DEVICE_RUNNER_KEYBOARD_STABILITY samples=3 captured=3 settled=yes
AGENT_DEVICE_RUNNER_IN_APP_BACK_VISUAL_VERIFICATION beforeBytes=365716 afterBytes=365716 changed=no
```

The keyboard line needed a field whose `returnKeyType="done"` exposes the dismiss key the settle path waits behind; the Form screen's email field does, and three samples came back non-nil. The `back` line is the no-change control: 365 KB of pixels on both sides, read as unchanged, and the command then reported that no in-app back control exists. Both consumers now say what they looked at, which is what made these two lines possible.

**4. Coverage, smoke, and the `exists` read.** Coverage was this PR and is fixed by the move above. I agree the smoke timeout is inherited; the automation screen behind that `wait text` is reachable and its alert is exercised above. On the `exists` read: the base already resolved `windows.firstMatch` and read its `frame` on every synthesized gesture, so the added cost is one attribute read on an element that was already resolved, and `testResolvedScreenRefusesAWindowTheQuerySaysIsNotThere` pins it — a windowless app answers `frame` with `(0,0 0x0)` without raising, which the frame check alone catches only by coincidence. Happy to split it if you would rather the gesture path keep its old reads exactly.

One gap I have not closed, now recorded in the ADR: `home` does not background the app on this Duo lane, so the SpringBoard branch is pinned by the unit test above and by a device run where SpringBoard *is* the session app, not by a run where the bound app went windowless.

thymikee:
Reviewed 80079f7, the follow-up to da8bf8a. I found no code problem in the two fixes, and your size answer closes that question. What is still missing is a live run of the new SpringBoard fallback in `captureObservedScreen` ([RunnerAppScreenCapture.swift#L97](https://github.com/callstack/agent-device/blob/80079f7431549be36ac7f0cefc889d52302221ae/apple/runner/AgentDeviceRunner/AgentDeviceRunnerUITests/RunnerAppScreenCapture.swift#L97)).

The "home screen, no app" run in the PR bound the session to com.apple.springboard, so `resolveAppWithoutActivation` returned SpringBoard and the capture succeeded on the primary branch without reaching the fallback. The ADR also notes that `home` does not background the app on the Duo lane. So the no-app case, and a session app that is backgrounded or terminated, are covered only by unit fakes that stub both queries. If a backgrounded or terminated app answers the window query differently (for example `exists=YES` with a stale frame, which gives `.unresolvedScreen` with no fallback), home-screen screenshots, recording and keyboard settle on the runner route would still fail with `APP_SCREEN_*`.

Can you run this head on a standard iPhone simulator with `simctl io screenshot` forced to fail, so the runner route runs? First, a fresh session with no `open`: `screenshot` should return an image, and runner.log should show `AGENT_DEVICE_RUNNER_APP_SCREEN_UNRESOLVED stage=window ... resolved=no` followed by `AGENT_DEVICE_RUNNER_SCREEN_CAPTURE display=1 ... pixelsPerPoint=3`. Second, `open <fixture>` then `home` (or `simctl terminate` the app) with the session still bound: the same two lines and a home-screen image.

CI: iOS Smoke fails at `wait text "Automation lab"` with `wait_deadline_exceeded`. That wait polls the snapshot route, not screenshot, and base #2724 fails at the same step, so it looks like it comes from the base branch. No conflicts.

thymikee:
Rebased again — the base moved twice under this PR (`800151395`, the interaction-window helper). `git range-diff 073b5687b..80079f743 origin/fix/ios-synthesized-reference-frame-capture-viewport..24cd0951e` reports `=`, so the capture code carried through unchanged, and the current head `940c20d78` adds only a measured-behaviour paragraph to the ADR. GitHub reports the PR `MERGEABLE`, so there were no conflicts to resolve beyond the ADR one in the first rebase.

Both runs you asked for, at this head, on an iPhone 17 Pro Max with the same `xcrun` shim failing only `simctl io … screenshot`.

**Fresh session, no `open`.** `open --platform ios --udid <sim>` binds a session with no app, and `screenshot` returned a 440x956 image:

```
AGENT_DEVICE_RUNNER_APP_SCREEN_UNRESOLVED stage=window frame=(0,0 0x0) raised=no resolved=no
AGENT_DEVICE_RUNNER_SCREEN_CAPTURE display=1 pixels=1320x2868 pixelsPerPoint=3
```

**Bound app, then backgrounded or terminated.** The iPhone ran Safari rather than the fixture: the fixture is installed on the iPhone 17 Pro, which another workspace's session holds, and installing it on the free sim needed a Metro build I could not justify for a capture check. With the session still bound to `com.apple.mobilesafari`:

- `simctl terminate`: the same refusal line, `resolved=no`, followed by `AGENT_DEVICE_RUNNER_SCREEN_CAPTURE display=1 pixels=1320x2868 pixelsPerPoint=3`, and the image is the home screen — band luminance 127.2 against 226.8 for Safari on the same session.
- `home` instead: the image is the home screen again (127.2), but there is no refusal line. A suspended app answers the window query with a usable window, so it takes the primary branch and captures the panel it is on.

Your stale-frame worry did not appear in either state: a terminated app answers `resolved=no` rather than `exists=YES` with a leftover frame, so the fallback engages, and a suspended app resolves cleanly rather than landing on `.unresolvedScreen` with no fallback.

I ran the same sequence on the Duo as well, because that is where this branch can go wrong — SpringBoard has a choice of panels there. Fixture bound, `simctl terminate`, session still bound:

```
AGENT_DEVICE_RUNNER_APP_SCREEN_UNRESOLVED stage=window frame=(0,0 0x0) raised=no resolved=no
AGENT_DEVICE_RUNNER_SCREEN_CAPTURE display=3 pixels=2852x2006 pixelsPerPoint=3
```

`display=3` is the lit inner panel, and the resulting image has `blackFraction=0.0000`. Had SpringBoard's window named the outer panel, that capture would be the black one this PR exists to prevent. `home` is still what I could not make work on that lane — it leaves the app frontmost on the Duo, which is why the Duo fallback was driven with `terminate`. The ADR now states the measured answers rather than the assumed ones.

On CI, I can add a data point that agrees with your read. Locally, `wait text "Automation lab"` does time out on a Duo session after the app is terminated and reopened through a deep link, and the surface the failure names is `Toolbar, Agent Device Tester, user-icon, Development Build` over a `DEVELOPMENT SERVERS` list — the dev client's own launcher screen, which does not contain that text and does contain it once you enter the app. So the text is genuinely absent rather than unreadable, and the snapshot and screenshot routes are not what loses it. That points at launch routing on the base branch rather than at anything here; say the word and I will take it as a follow-up.

thymikee:
That `CONFLICTING` came from the base moving a third time while I was running the runs above — `b9748e264`, "route synthesized display IDs through the resolved window", which edits the same bridge function this PR rewrote. Rebased over it; head is `00bec14be` and GitHub reports `MERGEABLE` again.

The conflict was `RunnerXCTestEventBridge.m` and its unit tests. How I resolved it, since it decides who owns what:

- Your `RunnerResolveWindowDisplayID(id window, NSUInteger *displayID)` entry point stays as the gesture path's API. Its frame check stays yours, and its screen read now goes through `RunnerReadResolvedWindowScreen`, so a window-level read that raises or answers no display fails with the same typed reason and the same `AGENT_DEVICE_RUNNER_APP_SCREEN_UNRESOLVED stage=screen` line as an app-level one, instead of a second hand-rolled `displayID` read.
- `RunnerResolveApplicationDisplayID` still delegates through `RunnerResolveApplicationScreen`, which is where `windows.firstMatch` is walked and checked for `exists` plus a usable frame. So the two entry points differ only in who resolves the window; the reads below them are one implementation.
- Both sides' bridge tests are kept, not merged away — 12 in `RunnerTests+XCTestEventBridgeTests.swift`, including your `testFirstUsableWindowSkipsAbsentAndEmptyWindows` and `testSynthesizedDisplayRefusesWindowWithoutQualifyingFrame` next to the six `testResolvedScreen*` cases.

At the new head: runner builds, the unit lane passes 20 tests, `check:affected --run`, `check:layering`, `check:fallow`, typecheck, lint and format are clean. I also re-took the no-app fallback capture so the log lines describe this head rather than the previous one — `AGENT_DEVICE_RUNNER_APP_SCREEN_UNRESOLVED stage=window frame=(0,0 0x0) raised=no resolved=no` then `AGENT_DEVICE_RUNNER_SCREEN_CAPTURE display=1 pixels=1320x2868 pixelsPerPoint=3`, with the home-screen image behind it. The PR body now states which captures came from which head.

thymikee:
Ran the forced-failure live verification the review asked for, on a standard iPhone simulator (iPhone 17 Pro, iOS 26.x) with `xcrun simctl io … screenshot` shimmed to fail with the fallback-eligible signature (`timed out waiting for screen surfaces`), so every capture below is the runner route. Host built from the restacked head; code identical to `00bec14be` apart from the ADR paragraph.

**1. Fresh session, no app fixture** (session opened device-only, runner host owns the session): `screenshot` returned a real image (402x874 @1x), and `runner.log` shows exactly the two predicted lines:

```
AGENT_DEVICE_RUNNER_APP_SCREEN_UNRESOLVED stage=window frame=(0,0 0x0) raised=no resolved=no
AGENT_DEVICE_RUNNER_SCREEN_CAPTURE display=1 pixels=1206x2622 pixelsPerPoint=3
```

**2. `open <fixture>` then `simctl terminate` with the session still bound**: same two lines verbatim, and the image is the home screen — SpringBoard icons, wallpaper and dock all present, no black region (449 KB PNG; the old failure mode was a near-black capture).

I deliberately did not expect the two lines from the `home` variant: per the ADR's own measurement a suspended app keeps answering with a usable window, so it takes the first branch — that leg is covered by `testObservedScreenCaptureKeepsTheSessionAppCaptureItResolved` plus the Duo evidence, and the terminate variant is the one that exercises `resolved=no` → system-surface fallback, which is what ran.

Also: the base moved under this PR twice today; I replayed `24cd0951e` onto `b9748e264` independently before noticing your own resolution — the two resolutions are byte-identical apart from your ADR note, and I restored your `00bec14be` as the head. Local validation of the restacked pair: `TEST BUILD SUCCEEDED` with unit tests enabled, and the 17 highest-risk runner tests pass on-device, including both selector-based synthesis-fallback regressions against the new `resolvedWindow:` selector shape and all six `RunnerResolveApplicationScreen` / capture-observed-screen tests.

thymikee:
Reviewed 00bec14. Thanks for the forced-failure run. It covers the `captureObservedScreen` fallback that I asked about.

The rebase onto b9748e2 made a test from #2724 weaker. At this head, `DisplayWindowFixture` resolves only on `exists`, and both window-path tests read `exists` before they call `RunnerResolveWindowDisplayID` ([RunnerTests+XCTestEventBridgeTests.swift#L88](https://github.com/callstack/agent-device/blob/00bec14bea4c2eb14715e436393e8f0e5c550b54/apple/runner/AgentDeviceRunner/AgentDeviceRunnerUITests/UnitTests/RunnerTests+XCTestEventBridgeTests.swift#L88) and #L98). On b9748e2, these tests proved that the window path reads `frame` before `screen`, because an unresolved screen answers display 1. Now the window is resolved before the call. So if a change reads the screen first, or drops the frame read, in [RunnerXCTestEventBridge.m#L231-L244](https://github.com/callstack/agent-device/blob/00bec14bea4c2eb14715e436393e8f0e5c550b54/apple/runner/AgentDeviceRunner/AgentDeviceRunnerUITests/RunnerXCTestEventBridge.m#L231-L244), the tests still pass. On a foldable, that change would send a synthesized tap to display 1 and no test would fail. Can you make the fixture resolve on the `frame` read again, as on b9748e2, and remove the `exists` pre-reads? Please also add window-path cases for display ID 0 and for a window with no screen. Each should expect "no resolved window display ID", as the application walk tests already do.

Not blocking: the comment at [RunnerXCTestEventBridge.m#L227](https://github.com/callstack/agent-device/blob/00bec14bea4c2eb14715e436393e8f0e5c550b54/apple/runner/AgentDeviceRunner/AgentDeviceRunnerUITests/RunnerXCTestEventBridge.m#L227) says that a window that raises fails closed with the same reasons as the application walk. But the frame read at L232 is outside the exception catcher. Either move that read inside `RunnerObjCExceptionCatcher`, or narrow the comment.

CI: nothing has failed. A newer push cancelled three Smoke Tests runs and Coverage, and one Smoke Tests run is still in progress. This change touches the route that iOS Smoke runs, so that result matters here.

Next step: restore the window-path order test and add the two refusal cases.

thymikee:
Rebased onto the moved base (`41164051c`) and took all three items in `7e1798877`:

- `DisplayWindowFixture` resolves on the `frame` read again (not `exists`), and both window-path tests dropped their `exists` pre-reads — `testSynthesizedDisplayRoutesByTheResolvedWindow` now fails if the window path ever reads `screen` before `frame`, since an unresolved fixture answers display 1.
- Two new refusals pin the window path to the typed screen reason: a window answering display ID `0` and a window whose `screen` read raises both expect `no resolved window display ID` in the error, not just a non-nil return.
- `RunnerResolveWindowDisplayID` now reads `frame` inside `RunnerObjCExceptionCatcher`, so a raising window gets the typed unresolved-window refusal plus the `stage=window raised=yes` log line — the comment matches the code.

All 13 bridge/capture/resolver unit tests pass locally (`TEST EXECUTE SUCCEEDED`).

thymikee:
Both asks are on the branch, and the branch is on the current base tip (`41164051c`); GitHub reports `MERGEABLE` at head `ea9855583`.

**Window-path order proof and the two refusals.** These landed in `7e1798877`, which came from the parallel lane rather than from me, and I want that on record: I had built the same change independently on the previous base, and my own force-push briefly displaced that commit while I was restacking. I restored it as soon as I saw it — net zero on the branch contents, and it is why three runs show as cancelled in the Checks list. Worth one lane preferring fast-forward pushes while two of us are pushing this branch.

I checked the restored guard rather than trusting the diff. Rebuilding the head with a mutation that reads `screen` before `frame` inside `RunnerResolveWindowDisplayID`, `testSynthesizedDisplayRoutesByTheResolvedWindow` fails with `("1") is not equal to ("3")` — exactly the display-1 routing you predicted for a foldable. On the pre-fix fixture that mutation ran green, which was the weakness you pointed at. The mutation is reverted; only the test addition below is pushed.

**The catcher.** `7e1798877` moved the frame read inside `RunnerObjCExceptionCatcher` and gave the refusal the same `stage=window` log line as the application walk, so the comment is now true. Nothing drove that branch, so `ea9855583` adds `testSynthesizedDisplayRefusesAWindowWhoseFrameLookupRaises`: a window whose `frame` key the runtime will not answer, which is how a renamed private read fails — KVC raises rather than returning nil. It expects `no resolved application window` and the caller's `displayID` left untouched, and it passes. A frame read that raises would otherwise land in the gesture path, which holds that window across a command boundary.

Validation at this head: 23 unit tests pass on-device across `RunnerTests+XCTestEventBridgeTests.swift` and the capture tests, `pnpm check:affected --run` is green including the packaged-runner line-numbering guard, and lint and format are clean.

CI: everything re-queued at `ea9855583`. iOS Smoke and Coverage are the two that bear on this route, so I would not call this done until those are green.

I also refreshed the Validation section so its head and test counts match this restack.

thymikee:
Reviewed ea98555. The earlier finding is fixed: `DisplayWindowFixture` resolves on the `frame` read again, both window-path tests now fail if the screen is read before the frame, and the new refusals check the typed "no resolved window display ID" reason. The caught frame read gives a raising window the typed refusal and leaves the success path unchanged. The forced-failure run you posted for 00bec14 still covers the capture fallback, because the capture commit did not change.

CI: Smoke Tests, Repo Guards, Coverage and Integration Tests were still running when I reviewed, with no failures so far. No conflicts. This PR lands after #2724.

thymikee:
ea98555 now conflicts with its base branch after #2724 was rebased onto main (b8fcbdb). The earlier clean review of ea98555 still stands; I removed ready-for-human until this is restacked without conflicts.

thymikee:
Restacked onto the current base tip again — the branch is `4d9229bac` on `24a6bbca0` and GitHub reports `MERGEABLE`. This one was a real conflict rather than context drift, so recording the judgment behind it.

The base's newest commit drops `RunnerResolveApplicationDisplayID`. Earlier revisions of this PR called that function from four tests, so the tree arrived half-deleted: the API gone, its callers still in the test file, and two copies of `testSynthesizedDisplayRoutesByTheResolvedWindow`. I kept the deletion. Nothing needs an application-level string display resolver any more — the capture route resolves through `RunnerResolveApplicationScreen` and branches on the typed reason, the gesture route resolves through `RunnerResolveWindowDisplayID`, and the deletion is what stops a third caller re-walking `windows.firstMatch` for a display.

Four app-level tests went with it, because each assertion is carried elsewhere:

- an empty frame refuses without naming a display → `testSynthesizedDisplayRefusesWindowWithoutQualifyingFrame` and `testResolvedScreenRefusesAnEmptyWindowFrameWithoutNamingAnyDisplay`
- a resolved window with no answerable display ID → `testSynthesizedDisplayRefusesWindowWithNoDisplayIDOrNoScreenAnswer`
- window resolved before screen → the `resolved` assertion in `testSynthesizedDisplayRoutesByTheResolvedWindow`, which now states why it is asserted, plus the identity assertion in `testResolvedScreenHandsBackTheScreenTheResolvedWindowReported`
- the `exists` refusal → `testResolvedScreenRefusesAWindowTheQuerySaysIsNotThere`, now the only thing that can refuse a window with a plausible frame

One history change worth flagging: the fixture that lets the window-path tests see a screen-first change had to move into the first commit, because with the old fixture the intermediate commit could not compile against the base's tests. No commit on the branch now carries the weakened order proof. The two test commits keep their messages.

Validation at this head: 21 runner unit tests pass on-device across the bridge and capture test files, `pnpm check:affected --run` is green including the packaged-runner Swift guard, and the runner and TypeScript builds and lint/format are clean. The unit count is lower than the previous 23 because of the four redundant tests above, not coverage lost to the rebase.

The base has been restacked five times today, so CI is fully re-queued at this head.

thymikee:
Reviewed 4d9229b. The restack is sound. It keeps the base's removal of `RunnerResolveApplicationDisplayID`, no reader of it is left, and no test name is duplicated. The tests you named still cover each check from the four removed tests, and the window-before-screen order proof still holds.

Not blocking: the capture path picks its window with `windows.firstMatch` and refuses it when it is empty ([RunnerXCTestEventBridge.m#L135](https://github.com/callstack/agent-device/blob/4d9229bac41477aa1323d4e80f2ef8f448c0adcb/apple/runner/AgentDeviceRunner/AgentDeviceRunnerUITests/RunnerXCTestEventBridge.m#L135)). `resolveRunnerWindow` instead picks the first window that exists and has a frame. If the first window is empty and a later one is lit, could capture refuse the window that interactions use? Routing both through `resolveRunnerWindow` would keep one rule.

CI: Smoke Tests fails in `testBareTypeUsesTappedInputWhenSoftwareKeyboardIsHidden`. This looks unrelated. The tap step passed, then the typing step posted 17 characters and read back 2, which is the #1874 dropped-keystroke pattern, and this diff does not change that typing route. A rerun of the job should confirm it.

No conflicts. This is ready for human review, and it lands after #2724.

thymikee:
#2724 landed as `ee9d203db` and its branch was deleted, which retargeted this PR at `main` and flipped it to `CONFLICTING` without any of my commits moving. Rebased the three commits onto `215229b4a`; `git range-diff` reports all three unchanged, so the merged bridge is byte-for-byte the tip this branch was already building against and the conflict was ancestry, not content. The diff against `main` is now 17 files, +1188/-119 — the stacked commits dropped out of the PR as intended.

Revalidated on `main`: 21 runner unit tests on-device across the bridge and capture test files, `pnpm check:affected --run` green including the packaged-runner Swift guard, `check:layering` and `check:fallow` clean over the 17 changed files, builds and lint/format clean. Pushed as `3e1955429`; GitHub reports `MERGEABLE` and CI is queued.

thymikee:
Rebased onto `215229b` after #2724 merged. The patch at `3e19554` is byte-identical to the reviewed `4d9229b` (same diff against its base), so the earlier clean verdict stands.

The [Coverage failure](https://github.com/callstack/agent-device/actions/runs/35720666162/job/106722558363) does not come from this PR. `scripts/__tests__/eager-closure-budgets.test.ts` reports the `APPROVED_OVER_CEILING` row for `packages/capture-kit/src/recording/contact-sheet.ts` as stale, because main carries that entry since #2747. Every PR based on current main will fail this test until that row is removed. Smoke Tests is still running.

PR #2743, opened 2026-09-21
refactor(capture-kit): share glyph drawing and recording script lookup

thymikee:
## Summary

Bottom layer of stack #2748 (`record contact-sheet`). Extracts two primitives that a second consumer needs, without changing behavior:

- `png-glyph-text.ts` owns the bitmap glyph table `screenshot-overlay-draw.ts` used to burn `@eN` refs, so frame timestamps can be drawn with the same font instead of a second copy.
- `swift-cache.ts` owns `resolveRecordingScriptPath` / `buildRecordingScriptPathCandidates`, previously private to `overlay.ts`, so any recording helper resolves its Swift source the same way through the cached compile seam.
- `png-pixels.fixtures.ts` and `resizePngBox` become shared pixel fixtures/helpers.

9 files, ~430 gross lines.

## Validation

Refactor layer: no behavior change, so the existing overlay surface is the test. `screenshot-overlay.test.ts`, `overlay.test.ts`, `swift-cache.test.ts` and the capture-kit unit tree pass (73 tests in the recording/capture subset); full typecheck clean. Evidence for the whole stack (live iOS simulator recordings, calibrated thresholds, adversarial review) is in #2747.

Risk: none beyond the moved code paths; every existing caller still resolves the same script candidates and draws the same glyphs.


thymikee:
Reviewed 6c1c562. Two things need a change before this layer is ready.

`screenshot-overlay-draw.ts` now imports `setPngPixel` for `fillRect` ([L79](https://github.com/callstack/agent-device/blob/6c1c5629c645369fe2a1115ca8196bbed6c1df70/packages/capture-kit/src/screenshot-overlay-draw.ts#L79)), but `drawHorizontalLine` and `drawVerticalLine` still call a local `setPixel` with the same code ([L108-L120](https://github.com/callstack/agent-device/blob/6c1c5629c645369fe2a1115ca8196bbed6c1df70/packages/capture-kit/src/screenshot-overlay-draw.ts#L108-L120)). The goal of this PR is one shared primitive, and two copies can drift on the next bounds fix. Can both functions call `setPngPixel`, and `setPixel` be deleted?

Moving `resolveRecordingScriptPath` into swift-cache.ts ([L3](https://github.com/callstack/agent-device/blob/6c1c5629c645369fe2a1115ca8196bbed6c1df70/packages/capture-kit/src/recording/swift-cache.ts#L3)) adds static imports of `fileURLToPath` and `findProjectRoot` there. video.ts and screenshot-overlay.ts import swift-cache.ts, so their eager closures now pull in host-kit/version, although video.ts never calls the new function. That is why the eager-closure budget check in Coverage fails (video.ts 27 modules vs 23, screenshot-overlay.ts 33 vs 32). Could the `findProjectRoot` wrapper live in a module that video.ts does not import, with only the pure `buildRecordingScriptPathCandidates` in swift-cache.ts?

Not blocking: png-pixels.fixtures.ts could call `setPngPixel` instead of inlining the RGBA write, the new png-resize.ts export has no consumer yet, and the scale loops in `drawPngGlyphText` are over the complexity threshold.

CI: Coverage and Compatibility & Provenance fail on code this PR adds or moves, so they are related. iOS Smoke fails at "automation-longpress did not become visible after scrolling"; main also fails the live iOS fixture job and this diff is not on that route, so it looks unrelated. No conflicts.

thymikee:
Both fixed, and the shape of this layer changed because of the second point.

**One pixel primitive.** `setPixel` is gone; every painter in the module — border lines, badge fill, glyph blocks — writes through `setPngPixel`. That work now lives in #2746, which is where the primitive gains its second consumer.

**The eager-closure growth.** You were right that `findProjectRoot` in `swift-cache.ts` was the cause. Rather than split the wrapper into a third module, `resolveRecordingScriptPath(scriptName, projectRoot, …)` now takes the root from its caller: `overlay.ts` already imported `findProjectRoot` at the merge base, so its closure is byte-identical, and `video.ts` never sees it. Eager-closure is green locally (`video.ts` back to its merge-base count, `screenshot-overlay.ts` unchanged).

The glyph-table extraction is withdrawn. `screenshot-overlay.ts`'s eager-closure entry has a no-growth budget, and a new module in its closure failed it before any second consumer existed — the same objection fallow raised as an unused export. Glyphs stay where they are and #2746 widens them in place.

Non-blocking items taken: the `png-resize.ts` export moves to #2746 where the renderer calls it; `drawPngGlyphText`'s scale loops move into #2746, where the per-glyph block is a `drawGlyph` helper and health is clean; and the pixel fixtures now write through `setPngPixel` (#2746), since the primitive only exists from that layer up.

CI: Coverage and Compatibility & Provenance both came from this diff and are fixed here. iOS Smoke also fails on `origin/main` at the same live-fixture step, so it is not this route.

thymikee:
Reviewed 73fec06. Both earlier findings are resolved: this PR no longer touches `screenshot-overlay-draw.ts`, and `resolveRecordingScriptPath` takes the project root from its caller, so `swift-cache.ts` adds no host-metadata import and the `video.ts` and `screenshot-overlay.ts` eager closures do not grow. The moved lookup keeps its candidate order, and its test moved with it. All checks pass and there are no conflicts.

PR #2744, opened 2026-09-21
feat(recording): decode contact-sheet frames from an exported clip

thymikee:
## Summary

Adds the decode half of a recording contact sheet: a runtime-compiled Swift helper plus the TypeScript that drives it.

`recording-frames.swift` asks an `AVAssetImageGenerator` for a list of presentation times, writes one PNG per answered sample into a scratch directory, and prints a manifest of the times the decoder *actually* returned — never the times that were requested. `extractRecordingFrames` compiles it through the existing cached Swift seam, bounds the run, and turns every failure into an `AppError` with a `details.reason` (`contact_sheet_frame_extraction_failed`, `contact_sheet_no_frames`), including a helper that fails to start or is killed at its deadline. Coverage loss is computed as requested minus delivered, so a decoder that stays silent about dropped samples cannot report full coverage.

Also refuses a non-macOS host (`contact_sheet_unsupported_host`) before any work, and adds the refusal vocabulary to `@agent-device/contracts/recording-contact-sheet`.

## Validation

`contact-sheet-frames.test.ts` (15 tests) covers grid pass-through, out-of-order answers sorted by actual time, declared and undeclared sample loss, malformed manifests, oversized answers, spawn failure, and cancellation forwarding. `recording-scripts.test.ts` typechecks the new Swift file. Stack-wide live evidence is in #2747.

Risk: `readVideoContainerKind` was renamed from an internal predicate; existing recording callers are unaffected.


thymikee:
Reviewed a3412ea. I have one scope question before this layer is ready.

Does this layer need more than the Swift helper, `extractRecordingFrames`, and the extraction and no-frames reasons? The new `@agent-device/contracts/recording-contact-sheet` subpath ships six reasons, a result type and `readVideoContainerKind` that nothing here reads yet, which is what the fallow unused-export and unused-type findings catch. Could the subpath wait for the layer that consumes it, or start with only these two reasons plus the unsupported-host reason? What would have to change first?

Not blocking: the error pass-through at [contact-sheet-frames.ts#L107](https://github.com/callstack/agent-device/blob/a3412eafc64b9447af44c7b8173167f72f29b9c0/packages/capture-kit/src/recording/contact-sheet-frames.ts#L107) (and L252) is wider than the cancel case and could narrow to `isRequestCanceledError`; the two catch-and-wrap blocks at L104 and L251 are the 51-line clone fallow flags, and one `toExtractionFailure` helper would cover both; an invalid `actualTime` seems to be written as 0 ms in [recording-frames.swift#L168](https://github.com/callstack/agent-device/blob/a3412eafc64b9447af44c7b8173167f72f29b9c0/apple/runner/AgentDeviceRunner/RecordingScripts/recording-frames.swift#L168) instead of marking the sample skipped; and `extractRecordingFrames` never calls `assertContactSheetHostSupport`, so a non-macOS host gets `contact_sheet_frame_extraction_failed` instead of the unsupported-host reason.

CI: Lint & Format fails on contact-sheet-frames.test.ts, which this PR adds, and Compatibility & Provenance fails on the unused exports and the clone above, so both are related. The Coverage eager-closure failure comes from the #2743 base. Smoke fails on a WebView step that this decoder is not wired to, so it looks unrelated. No conflicts.

thymikee:
Narrowed — the subpath now carries exactly three names: `contact_sheet_unsupported_host`, `contact_sheet_frame_extraction_failed`, `contact_sheet_no_frames`. Nothing else here reads anything more, so nothing more is declared.

Nothing has to change first for the subpath to exist; it lands with its first consumer, and every other name now arrives in the layer that raises it: the duration reason with the sample grid (#2745), the pixel-budget reason with the renderer (#2746), container/write/collision plus the result and cell types with the pipeline (#2752), and `contact_sheet_threshold_invalid` with the command that parses the flag (#2747). `readVideoContainerKind` moved the same way — the rename and its export now sit in #2752, where the pipeline sniffs the container before it spawns a decoder.

Non-blocking items taken:

- `throwExtractionFailure` now rethrows a canceled request via `isRequestCanceledError`, so cancellation is no longer riding on the "already has a reason" branch; the pass-through for typed failures stays because a compile failure that names its own reason should not be relabelled.
- The two catch-and-wrap blocks are one helper, which is the 51-line clone fallow flagged; Compatibility & Provenance is clean locally against this layer's base.
- `recording-frames.swift` no longer reports `0 ms` for an unusable `CMTime`. It refuses to name a presentation time it does not have: the sample is recorded as skipped before anything is written, so a frame cannot arrive labelled as the opening one.
- `extractRecordingFrames` answers the host question itself, before compiling or spawning, so a caller that only asks for frames gets `contact_sheet_unsupported_host` rather than a spawn failure. It takes an optional `hostPlatform` for the same reason the pipeline does, and the tests state their host — this suite is in the unit lane, which also runs on Linux.

thymikee:
Reviewed ccf6687. The scope question from the first review is answered: the contracts subpath now declares only the three reasons this layer raises, each has a reader here, and the unused-export and clone findings are gone. The cancellation rethrow, the single extraction-failure helper and the `actualTime` fix look right.

Not blocking: the cancellation test ([contact-sheet-frames.test.ts#L284](https://github.com/callstack/agent-device/blob/ccf6687ea74190495c78c40dda7c0a11016b894d/packages/capture-kit/src/recording/contact-sheet-frames.test.ts#L284)) aborts before `extract()` runs, so it does not cover a canceled error from a rejected `runCmd` or `compileSwiftSourceFile`; a case that rejects with `createRequestCanceledError()` would pin it.

CI: Smoke Tests was still running. This module has no production importer yet, so Smoke does not reach it. No conflicts.

thymikee:
Reviewed e88dd5b. Since 442d367, this only moves the three refusal reasons from the contracts subpath into `capture-kit/src/recording/contact-sheet-report.ts` and drops the contracts export and its snapshot line. Nothing still imports `@agent-device/contracts/recording-contact-sheet`, and the constants are unchanged apart from the file comment.

PR #2745, opened 2026-09-21
feat(recording): choose the frames a contact sheet shows

thymikee:
## Summary

The reading rules for a contact sheet: which moments of a clip get sampled, and which of those frames earn a cell.

- `contact-sheet-plan.ts` turns a duration into a bounded grid: one sample per 250ms until 48 samples, then those 48 spread evenly over the clip, always including both endpoints so the sheet can speak about how much it covered. A two-hour take costs the same work as a five-second one.
- `contact-sheet-selection.ts` keeps a frame when it moves enough of the picture relative to the **last kept** cell rather than its predecessor, so small changes accumulate (a list scrolling one row per frame still shows up). It always keeps the opening frame, appends the last one, and thins evenly across the kept sequence to 24 cells, reporting that it thinned.
- `png-changed-pixel-ratio.ts` is the shared per-frame comparison.
- The default threshold is calibrated, not guessed: measured iOS simulator captures put idle repaints (status-bar clock) at ≤0.0093 and the smallest real change at 0.030, so the constant sits at 0.02.

## Validation

17 unit tests over endpoints, sub-interval clips, accumulation, thinning and both-end preservation. Stack-wide live evidence is in #2747.

Risk: a transient that opens and closes entirely between two sample times is not in the returned frames; that is disclosed as coverage, not hidden.


thymikee:
Reviewed ace7f8f. I found no problem in the sampling and selection rules, but one gate failure comes from this diff.

`CONTACT_SHEET_FRAME_WIDTH` ([contact-sheet-plan.ts#L9](https://github.com/callstack/agent-device/blob/ace7f8fb41afa1df4286c79f88e3976a31de2875/packages/capture-kit/src/recording/contact-sheet-plan.ts#L9)) and `CONTACT_SHEET_CHANGED_PIXEL_THRESHOLD` ([contact-sheet-selection.ts#L16](https://github.com/callstack/agent-device/blob/ace7f8fb41afa1df4286c79f88e3976a31de2875/packages/capture-kit/src/recording/contact-sheet-selection.ts#L16)) are exported, but nothing consumes them yet, so the dead-code check in Compatibility & Provenance fails on this PR. Can they stay module-private until a reader lands, or can this PR land together with the layer that reads them?

Not blocking: `computePngChangedPixelRatio` is a third pixel-ratio loop next to `computePngRgbDifference` and `computeScreenshotDiffPixels`, and its length-mismatch branch reuses the `dimension_mismatch` status.

CI: Lint & Format (a base-layer test file), Smoke (an unrelated runner probe) and Coverage (import growth from #2743) fail on code this PR does not touch. I could not check the 0.02 threshold calibration from the PR body against a fixture in the diff. No conflicts.

thymikee:
Reviewed 200296b. The earlier finding is fixed: the frame-width and threshold constants are no longer exported without a reader, and the duration reason added here has a reader in this layer. The planner and selector have no caller yet; #2752 wires them in. All checks pass and there are no conflicts.

thymikee:
Reviewed 52aadc4. After the rebase onto the new #2744 head, the only logical change since 2f940b2 is the import path: the reasons now come from `./contact-sheet-report.ts`, and the two reasons this layer adds live in that module unchanged. No reference to the removed contracts subpath is left.

PR #2746, opened 2026-09-21
feat(recording): render a contact-sheet PNG from decoded frames

thymikee:
## Summary

Draws the sheet and publishes it. `buildRecordingContactSheet` reads the container, refuses anything that is not MP4, plans the grid, decodes, selects, renders one PNG with each cell titled with its elapsed time, and reports what it covered (`sampledFrameCount`, `decodedFrameCount`, `skippedSampleCount`, per-cell `changedPixelRatio`) plus warnings for anything the grid could not show.

Frames are decoded from the finished export, never from a parallel capture, so the sheet cannot describe a screen the delivered video does not contain. Layout shrinks cells before it ever refuses, and a refusal carries `contact_sheet_pixel_budget_exceeded`. Output is staged under a name unique to the call, created with `wx`, and renamed into place, so a caller never sees a half-written PNG and no pre-existing path is followed; writing over the recording itself is refused with `contact_sheet_output_collides_with_input`. A frame wider than the cells that were asked for, or bigger than the whole sheet budget, is refused before it is allocated.

Also adds the `screen-recording-contact-sheet` artifact type and the beside-the-recording output path.

## Validation

23 unit tests over container refusal, coverage warnings, thinning disclosure, budget refusal, oversized frames, output collision and publish/cleanup. Stack-wide live evidence is in #2747.


thymikee:
Reviewed 3950eab. The output-collision guard can overwrite the recording, so this needs a fix before merge.

The guard compares `fs.realpathSync` strings ([contact-sheet.ts#L150](https://github.com/callstack/agent-device/blob/3950eab279a4ee7e46a0ed5720664ba5a20e39e0/packages/capture-kit/src/recording/contact-sheet.ts#L150)). On the default case-insensitive APFS volume, `realpathSync` keeps the case it was given, so `/x/video.mp4` and `/x/VIDEO.MP4` compare as different paths although they are the same file (checked with Node's `fs.realpathSync`). The guard passes, the file sniffs as MP4, and `writeContactSheetFile` renames the PNG over the recording. macOS is the only supported host, so an `--out` that differs only in case silently destroys the input. Can the guard compare file identity when the output exists (`statSync` `dev` and `ino` on both paths), and keep the string check only for a new output? A test with a case-variant path, a hard link or a symlink would cover it.

The "honours an explicit output path" test does not pass `hostPlatform: 'darwin'` ([contact-sheet.test.ts#L141](https://github.com/callstack/agent-device/blob/3950eab279a4ee7e46a0ed5720664ba5a20e39e0/packages/capture-kit/src/recording/contact-sheet.test.ts#L141)), so `assertContactSheetHostSupport` throws on the Linux runner and Coverage fails there. The test file's build helper could supply it for every case.

The new `DaemonArtifactKnownType` member ([contracts.ts#L128](https://github.com/callstack/agent-device/blob/3950eab279a4ee7e46a0ed5720664ba5a20e39e0/packages/kernel/src/contracts.ts#L128)) changes an ADR 0006 wire declaration without a `compatibleChanges` entry or digest in `test/wire-compat/ledger.json`, so `wire-compat.test.ts` fails. Nothing here uses it yet, so moving it to #2747 with its ledger entry looks cleaner.

Not blocking: `CONTACT_SHEET_COLUMNS` in contact-sheet-render.ts is flagged as an unused export.

CI: the explicit-output test and wire-compat failures in Coverage come from this diff. The video.ts and screenshot-overlay.ts module counts come from #2743, the format failure from #2744, and iOS Smoke also fails on main on a route this code is not on. No conflicts.

thymikee:
All three taken. This layer is also smaller than the one you reviewed: it is now the grid and the glyph table only, with the pipeline split out to #2752.

**Collision guard.** You were right, and the case-variant path is worse than a mistyped name because the guard reads as passing. The guard now compares file identity — `statSync` `dev`/`ino` on both paths — whenever the output exists, so a hard link, a symlink, or a second spelling of the recording all resolve to the same file and are refused; the string comparison is kept only for an output that does not exist yet, where there is no file to open. Landed in #2752 with a test that plants both a symlink and a hard link at `--out` and asserts the recording is still whole behind both names and nothing was spawned.

**Explicit-output test.** The file now has one `build()` helper that supplies `hostPlatform: 'darwin'` unless a case names another, so no case can silently depend on the runner's OS. The decoder tests in #2744 do the same.

**`DaemonArtifactKnownType`.** Moved to #2752, which is where the type is used, together with its `test/wire-compat/ledger.json` digest and a `compatibleChanges` ack keyed to that digest (`DaemonArtifactType` already widens to `(string & {})`, so an unknown artifact type was always valid on the wire). It is that layer's `chore(gates)` commit.

**`CONTACT_SHEET_COLUMNS`** is module-private.

Also from this layer: `drawPngGlyphText` gained the integer scale the elapsed-time labels need, the glyph table gained `:` and `.`, and every write in the module goes through `setPngPixel` — with the pixel fixtures using it too, so there is one RGBA write path in capture-kit.

thymikee:
Reviewed 7b6ff78. The earlier findings no longer apply to this layer: the collision guard, the explicit-output test and the `DaemonArtifactKnownType` change moved to #2752, and this PR has no file-writing code. Every painter in `screenshot-overlay-draw.ts` now writes through `setPngPixel`, the badge geometry is unchanged, and the renderer refuses a canvas over the pixel budget.

Not blocking: the tests could pin more. The pixel-budget test checks only the message, not `details.reason` ([contact-sheet-render.test.ts#L76](https://github.com/callstack/agent-device/blob/7b6ff788215fd7b16cdc60bbfc766f01271d4e55/packages/capture-kit/src/recording/contact-sheet-render.test.ts#L76)); the label test accepts any white pixel instead of one inside each cell ([#L53](https://github.com/callstack/agent-device/blob/7b6ff788215fd7b16cdc60bbfc766f01271d4e55/packages/capture-kit/src/recording/contact-sheet-render.test.ts#L53)); and no overlay test pins a known badge fill pixel or the pixel past the badge's right edge.

CI: all checks pass. No conflicts.

thymikee:
Reviewed 62461e7. Since 814b2cc, the only logical change is that the pixel-budget reason comes from `./contact-sheet-report.ts` instead of the removed contracts subpath. The rendering code is unchanged.

PR #2747, opened 2026-09-21
feat(cli): read a recording with record contact-sheet

thymikee:
## Summary

Top of stack #2748. Exposes the feature to callers:

```bash
agent-device record contact-sheet ./take.mp4
agent-device record contact-sheet ./take.mp4 --out ./sheet.png --json
```

```ts
await device.recording.contactSheet({ video: { kind: 'path', path: './take.mp4' } });
```

The command is CLI-local like `diff screenshot`: no session, no device, no daemon, and it can rebuild a sheet for a recording that finished last week. Both file paths go through command policy, an uploaded recording must name an output, and a downloaded input is released on every exit path. Versioned help and `website/docs/docs/commands.md` state the coverage claim and the macOS-host limit.

## Validation

Tested at `bd30d3d71`. Live gate `test/integration/recording-contact-sheet.test.ts` records on a booted iPhone 17 Pro (iOS 27.0), navigates Settings, builds the sheet, and asserts the PNG header matches the reported dimensions, cell times increase inside the clip, and every cell clears the threshold — passes in 25s under `AGENT_DEVICE_RECORDING_E2E=1`. Two manual takes (10.2s, 33.6s) drove real `simctl` transitions: 42/42 and 48/48 samples decoded, sheets read back as the sequence they depict. Refusals checked live: `--out take.mp4`, WebM, non-macOS.

Stack-wide: 1,889 unit tests, full typecheck, `oxlint`, `check:layering` 246/0, `check:packaged-runner-swift`, `test:integration:progress:check` with 0 unclassified flags. Two adversarial reviews fixed: staging-file symlink escape, sheet over its recording, uncounted sample loss, untyped compile/spawn failures, missing endpoint sample, oversized frame, silent PNG-write failure.

Residual risk: a hard link or parent-symlink swap inside a caller-owned output directory is still possible; 48 cells stay decoded at once, each bounded by cell width and `maxImagePixels`.


thymikee:
Reviewed bd30d3d. One flag-handling bug needs a fix, and I have one design question about the whole stack.

`recordCliSchema.allowedFlags` now shares `out` across `record start`, `record stop` and `record contact-sheet` ([index.ts#L67](https://github.com/callstack/agent-device/blob/bd30d3d7122cd955f49607216e893aac9e714685/src/commands/recording/index.ts#L67)). Before this PR, `record start --out take.mp4` was refused. Now the parser accepts it, but `recordCliReader` ([L78-L86](https://github.com/callstack/agent-device/blob/bd30d3d7122cd955f49607216e893aac9e714685/src/commands/recording/index.ts#L78-L86)) never reads `flags.out` for start or stop, so the flag is dropped and the recording lands at the default path. The reverse also happens: `record contact-sheet v.mp4 --fps 5 --scope device` parses, but the contact-sheet handler reads only the positionals and `--out`, so the recording flags are ignored. Per-command `allowedFlags` exist to stop exactly this ([flag-groups.ts#L105-L110](https://github.com/callstack/agent-device/blob/bd30d3d7122cd955f49607216e893aac9e714685/packages/command-registry/src/flag-groups.ts#L105-L110)). Could `recordCliReader` refuse `--out` for start and stop, and the contact-sheet handler refuse scope, fps, quality and hide-touches, with parser tests for `record start --out x` and `record contact-sheet v --fps 5`?

Across #2743-#2747 (about 1.6k net production lines), does the new `@agent-device/contracts/recording-contact-sheet` subpath need to exist? Its only readers are capture-kit and this PR's runtime file. Moving the result type and reason constants into the capture-kit facade would drop a contracts subpath, its exports-snapshot row and a layering pin, and a lazy import of that facade would avoid the eager-closure ceiling. What would have to change first for that smaller shape to work? The decode and render work in the lower PRs looks like the real weight, and I did not find an existing primitive that already does it.

Not blocking: `recordCommandDescription` advertises contact-sheet to the MCP tool although `readRecordingAction` refuses it; the local-runtime setup in src/cli/commands/recording.ts nearly copies `diffCommand` in screenshot.ts; the ordering comment at contact-sheet.ts#L55 could become enforced order; and no CLI test drives a successful contact-sheet run.

CI: the Coverage eager-closure failure for the capture-kit contact-sheet facade and the fallow complexity finding in the new integration test come from this PR. The Repo Guards finding in contact-sheet-render.ts, the wire-compat change and the oxfmt failure come from #2746 and #2744. No conflicts.

thymikee:
The env-gated integration test here (`test/integration/recording-contact-sheet.test.ts`) was re-run on this stack's current tip against a booted iPhone 17 Pro simulator and passes in ~24 s:

```
AGENT_DEVICE_RECORDING_E2E=1 AGENT_DEVICE_IOS_UDID=<udid> \
  node --experimental-strip-types scripts/node-test-tmpdir.ts --test test/integration/recording-contact-sheet.test.ts
✔ record contact-sheet reads a live iOS simulator recording (24019ms)
```

It asserts the response and the bytes together: PNG magic, reported dimensions equal to the PNG header, a bounded sample grid with nothing declined, the sheet opening on frame 0, strictly increasing cell times inside the clip, and every cell clearing the reported threshold. Artifacts land in `test/artifacts/ios/record-contact-sheet/<runId>/` next to the `.mp4` they were drawn from.

What the resulting sheet actually looks like — 15 cells, why the clock ticking costs nothing, and where each transition landed — is in #2752's live-evidence comment.

thymikee:
Reviewed b61ddf7. The flag-routing finding from bd30d3d is still open after the rebase, and the PR now conflicts with its base.

`recordCliSchema.allowedFlags` ([index.ts#L66](https://github.com/callstack/agent-device/blob/b61ddf75681bffa933f4b3a30831fb167a9d27eb/src/commands/recording/index.ts#L66)) is still shared by start, stop and contact-sheet. `record start --out take.mp4` parses, but `recordCliReader` ([L75-L82](https://github.com/callstack/agent-device/blob/b61ddf75681bffa933f4b3a30831fb167a9d27eb/src/commands/recording/index.ts#L75-L82)) never reads `flags.out`, so the recording goes to the default path and the flag is dropped. Before this PR, `record start --out` was refused. `record contact-sheet v.mp4 --fps 5 --scope device` also parses, but `recordingCommand` ([recording.ts#L10-L41](https://github.com/callstack/agent-device/blob/b61ddf75681bffa933f4b3a30831fb167a9d27eb/src/cli/commands/recording.ts#L10-L41)) reads only the positionals and `--out`, so the other flags do nothing. Each record action should accept only the flags it reads and reject the rest with `INVALID_ARGS`: `recordCliReader` can refuse `--out` for start and stop, and `recordingCommand` can refuse scope, fps, quality and hide-touches. Parser tests for `record start --out x`, `record stop --out x` and `record contact-sheet v.mp4 --fps 5` would pin it.

Not blocking: the threshold validation and its reason are not reachable from a `--changed-threshold` flag, although the PR body lists one, so the threshold is Node API only today. Add the flag or say so in the PR body.

On the design question from the first review: this head adds a second reason to `@agent-device/contracts/recording-contact-sheet` and an `APPROVED_OVER_CEILING` waiver for the capture-kit facade. Its readers are still only capture-kit and `src/commands/recording/runtime/contact-sheet.ts`. Would keeping the result type and reasons in the capture-kit facade, with a lazy import from the runtime, remove the subpath and maybe the waiver? Is there a consumer outside those two, such as MCP or the wire-compat ledger, that needs the reasons without loading capture-kit?

CI: all 9 checks pass; the eager-closure check passes under the new waiver. The conflict with feat/contact-sheet-pipeline is a separate blocker. Next: fix the flag routing with parser tests, and resolve the conflict.

thymikee:
Reviewed 34436f6. The per-action flag table fixes the earlier finding for typed flags: `record start --out x` and `record contact-sheet v.mp4 --fps 5` are now refused, and the new tests fail on the old code.

One problem is likely left. `assertRecordFlagsMatchAction` checks the merged flags ([index.ts#L95](https://github.com/callstack/agent-device/blob/34436f6409511cbc1a77ffb36d53964fbeeedfad/src/commands/recording/index.ts#L95)), and those include config and env defaults, not only typed flags. `record` now supports `fps`, `quality`, `hideTouches`, `recordingScope` and `out`, so `finalizeParsedArgs` ([args.ts#L200-L205](https://github.com/callstack/agent-device/blob/34436f6409511cbc1a77ffb36d53964fbeeedfad/src/cli/parser/args.ts#L200-L205)) keeps those defaults for every record action. With `{"hideTouches": true}` in config, or `AGENT_DEVICE_FPS` set, I expect `record stop` and `record contact-sheet clip.mp4` to fail with INVALID_ARGS for a flag the user never typed, so an open recording cannot be stopped. Can you confirm with a test? The parser already keeps `providedFlags` apart from defaults. If the per-action table lives on the command schema and the parser applies it, a typed flag the action does not read is refused and an unread default is dropped. A `resolveCliOptions` test with `AGENT_DEVICE_HIDE_TOUCHES=1` and `record stop` would pin it, next to the existing refusal test.

CI: the two Smoke failures (`testFreshCoordinateTapContainsUnavailableTextInputProbe` and `testHardwareKeyboardResponderConfirmsItsOwnKeyboardFocus`) are iOS runner tests for tap probes and keyboard focus. This diff does not touch the runner, so they look unrelated.

No conflicts. Next step: apply the flag check where typed flags and defaults are still separate, and add the default test.

thymikee:
Both findings addressed.

**Action-scoped `record` flags.** The parser is command-granular, so the shared list let `record start --out take.mp4` and `record contact-sheet clip.mp4 --fps 30` through and drop them quietly — the caller is left holding a file nobody asked for. The family now declares which action reads which option (`RECORD_FLAGS_BY_ACTION`, `src/commands/recording/index.ts:72`), `allowedFlags` is derived from that declaration so the two cannot drift, and the CLI refuses an option its action never reads (`assertRecordFlagsMatchAction`, `src/commands/recording/index.ts:92`, called from the `record` handler at `src/cli/commands/recording.ts:14`, which every action reaches before the generic route does). `record stop` reads no recording option at all, and a flag that belongs to no action is refused by every action — so a new flag has to name the action that reads it rather than widen a list. #2756 adds `noDiffOverlay` to the `contact-sheet` row and tests both directions.

**Contracts vocabulary.** Collapsed, as you suggested. `packages/contracts/src/recording-contact-sheet.ts` is gone; the refusal reasons and the report shape live in `packages/capture-kit/src/recording/contact-sheet-report.ts` and are published by the `@agent-device/capture-kit/recording-contact-sheet` facade, which was already the only surface the one external consumer (`src/commands/recording/runtime/contact-sheet.ts`) needed. No daemon route, registry entry, or wire-ledger declaration references them, so they were never cross-layer vocabulary — the contract here is between a pipeline and the surface that calls it.

What that removed: the `packages/contracts/package.json` export and its `contracts-exports.snapshot.json` line, both added in #2744. That layer's publication-only `chore(gates)` commit now had nothing left to publish, so #2744 is a single commit.

What stays: the eager-closure waiver on `packages/capture-kit/src/recording/contact-sheet.ts`. Its weight is the sample grid, the frame decoder seam, and the PNG codec behind one surface — the collapse moved a vocabulary module, not the mechanics, so the entry is still above the façade p75 for the reason the waiver records.

thymikee:
Reviewed 5c9ea2f. The config and env default problem from 34436f6 is still there. This delta does not change `src/commands/recording/index.ts` or the parser.

`assertRecordFlagsMatchAction` ([index.ts#L92](https://github.com/callstack/agent-device/blob/5c9ea2f2ab1c425fe165d2f461df56ac85dc90ad/src/commands/recording/index.ts#L92)) checks the keys of the merged flags. `resolveCliOptions` fills those flags with config and env defaults through `finalizeParsedArgs` ([args.ts#L168-L205](https://github.com/callstack/agent-device/blob/5c9ea2f2ab1c425fe165d2f461df56ac85dc90ad/src/cli/parser/args.ts#L168-L205)), and `fps`, `quality`, `hideTouches`, `recordingScope` and `out` can all come from config or env. So with `AGENT_DEVICE_FPS=30` set, or `{"hideTouches": true}` in config, `record stop` and `record contact-sheet clip.mp4` fail with INVALID_ARGS for a flag the user never typed. A configured `out` breaks `record start` the same way. The per-action table covers typed flags only. Can the check use `providedFlags`, which `resolveCliOptions` already returns, and drop unread defaults instead of refusing them? A `resolveCliOptions` test with `AGENT_DEVICE_FPS=30` and `record stop` would pin it.

CI: the Compatibility & Provenance job failed in the Fallow audit because of this delta. It reported 8 unused re-exports and 1 unused type export in [contact-sheet.ts#L40-L53](https://github.com/callstack/agent-device/blob/5c9ea2f2ab1c425fe165d2f461df56ac85dc90ad/packages/capture-kit/src/recording/contact-sheet.ts#L40-L53). The newer head 6c3191a keeps only `CONTACT_SHEET_THRESHOLD_REASON` and `RecordingContactSheetResult`, which should clear it. It does not change the flag check.

No conflicts. Next step: check only typed flags in `assertRecordFlagsMatchAction` and add the env-default test.

thymikee:
Fixed, and moved to where the rule belongs — `5f8fd76..ad53cdd`.

The check no longer looks at the merged bag. It also no longer lives in the command family: refusing an option the parse cannot route is the parser's job, and it already had the right input. `CommandSchema` now carries `flagsByAction`, and `finalizeParsedArgs` refuses an option the selected action never reads in the same block that already refuses an option the command never accepts — off `parsed.providedFlags` (`src/cli/parser/args.ts:201`, `src/cli/parser/args.ts:397`).

Consequences worth naming:

- A default cannot refuse anything. Config, env, and remote-config values never enter `providedFlags`, so `AGENT_DEVICE_FPS=30` or `{"hideTouches": true}` in config fills `flags` and stays out of the check by construction rather than by filtering. Pinned in `src/cli/resolve-cli-options.test.ts`: with `AGENT_DEVICE_FPS=30`, `record stop` resolves `flags.fps === 30` and `providedFlags` empty.
- Options every action reads stay untouched. Only the options the table names in some row are candidates, so `--json`, `--session`, and `--no-record` cannot be turned into unread options — a first cut that filtered on `!reads.includes(key)` did exactly that to `--no-record`, and `src/cli/parser/__tests__/args-validation.test.ts` now pins both directions.
- An action the table does not list is left to the reader that owns the action list, and commands without a table (`perf`, `trace`) parse exactly as before.

Also fixed here: the `no-duplicates` lint error the vocabulary collapse left in `src/commands/recording/runtime/contact-sheet.test.ts`, and the Fallow finding you read off `5c9ea2f` — the façade now publishes only `CONTACT_SHEET_THRESHOLD_REASON` and `RecordingContactSheetResult`, the two names production actually reads, which is why `--gate` reports zero new findings on `6c3191a` and later.

